PR #1073, opened 2020-09-11
Clean up to tool-attributes, add annotations and add test for custom …

shedaniel:
…tool types.

This PR does not change any of the behaviour, this PR simply does 3 things:
- Add @Nullable and @ApiStatus.Internal (people are not supposed to touch them!! whoever using the internal events please stop) annotations
- Remove redundant VanillaToolsModdedBlocksToolHandler's Tag#contains, it is already checked beforehand.
- Add a custom tool type (tater) in the testmod, and blocks that are effectively broken by taters. 

Signed-off-by: shedaniel <daniel@shedaniel.me>


fabric-tool-attribute-api-v1/build.gradle
```diff
@@ -1,9 +1,9 @@
 archivesBaseName = "fabric-tool-attribute-api-v1"
-version = getSubprojectVersion(project, "1.2.2")
+version = getSubprojectVersion(project, "1.2.3")
 
 dependencies {
     compile project(path: ':fabric-api-base', configuration: 'dev')
     compile project(path: ':fabric-tag-extensions-v0', configuration: 'dev')
     testmodCompile project(path: ':fabric-object-builder-api-v1', configuration: 'dev')
-	testmodCompile project(path: ':fabric-events-lifecycle-v0', configuration: 'dev')
+	testmodCompile project(path: ':fabric-lifecycle-events-v1', configuration: 'dev')
```
i509VCB, line 8: Convert everything above to tabs

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/AttributeManager.java
```diff
@@ -23,10 +23,12 @@
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
+import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.entity.attribute.EntityAttribute;
 import net.minecraft.entity.attribute.EntityAttributeModifier;
 
+@ApiStatus.Internal
```
modmuss50: General question: Do we want this on all internal classes?

If so would some way to enforce it for the whole of fabric api be a good idea?

shedaniel: This could be applied via package info I assume

modmuss50: Yes, that would be better.

shedaniel: Fixed.

modmuss50 (approved):
LGTM

OroArmor (approved):
Looks fine, tbh most of this won't be used by most programmers, but I do like the cleanliness.

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/VanillaToolsModdedBlocksToolHandler.java
```diff
@@ -51,12 +54,13 @@ public ActionResult isEffectiveOn(Tag<Item> tag, BlockState state, ItemStack sta
...
 	}
 
+	@NotNull
 	@Override
 	public TypedActionResult<Float> getMiningSpeedMultiplier(Tag<Item> tag, BlockState state, ItemStack stack, LivingEntity user) {
 		if (!(stack.getItem() instanceof DynamicAttributeTool)) {
 			ToolManagerImpl.Entry entry = ToolManagerImpl.entryNullable(state.getBlock());
 
-			if (entry != null && entry.getMiningLevel(tag) >= 0 && tag.contains(stack.getItem())) {
+			if (entry != null && entry.getMiningLevel(tag) >= 0) {
```
i509VCB, line 63: Is there a reason the tag check is dropped here?

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/ToolManagerImpl.java
```diff
@@ -253,8 +258,9 @@ default ActionResult isEffectiveOn(Tag<Item> tag, BlockState state, ItemStack st
 		 * @param user  the user involved in breaking the block, null if not applicable.
 		 * @return the result of mining speed.
 		 */
+		@NotNull
 		default TypedActionResult<Float> getMiningSpeedMultiplier(Tag<Item> tag, BlockState state, ItemStack stack, @Nullable LivingEntity user) {
-			return null;
+			return TypedActionResult.pass(1.0F);
```
i509VCB, line 263: Was there a reason this was null?

shedaniel: There is actually an Objects#requireNonNull check originally anyways, this was just a wrong default value.

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/VanillaToolsModdedBlocksToolHandler.java
shedaniel: Mentioned in the first message

Player3324 (approved):
Looks fine, thanks. Not entirely sure about the package-info classes, they are rather in the way - could probably add the annotations to the first real class in the package?

i509VCB:
The reason to specify `@Internal` in package info is so you don't need to apply it to every class file

shedaniel:
Yep, it is to not apply internal to every impl class, I've seen people trying to register an internal handler thinking that was required to have a custom tool type. This should help a bit.

i509VCB (approved):

PR #1075, opened 2020-09-11
conditional resources

shedaniel:
Adds conditional JSON resource loading via the JSON. Conditions can be registered via ResourceConditions#register.
Here is an example:
```javascript
{
  "type": "fabric:mod",
  "condition": {
    "minecraft": ">=1.16.3",
    "yes": [
      "<1.0.0",
      ">=2.0.0"
    ]
  }
}
```

The above JSON can be put in a file named `<original file>.fabricmeta` to apply to individual files or `.fabricmeta` to apply to the whole folder.

Signed-off-by: shedaniel <daniel@shedaniel.me>


haykam821:
Why not allow this for all resources?

shedaniel:
maybe this could turn into something like libcd then? aka using .mcmeta to define conditions

fabric-conditional-resource-api-v1/src/main/java/net/fabricmc/fabric/api/conditionalresource/v1/ResourceCondition.java
```diff
@@ -14,13 +14,13 @@
...
+package net.fabricmc.fabric.api.conditionalresource.v1;
 
 import com.google.gson.JsonElement;
 
 import net.minecraft.util.Identifier;
 
 @FunctionalInterface
-public interface RecipeCondition {
+public interface ResourceCondition {
 	boolean process(Identifier recipeId, JsonElement element);
```
ghost: `resourceId`

i509VCB:
Conditions should be able to be applied on a per directory scale imo. So a condition would for example not load anything in `data/namespace/tags/blocks/conditional`. How you would do that is going to be another question itself

shedaniel:
We could have a `<file name>.mcmeta` for per file conditions + `.mcmeta` for the whole directory. (the extension doesn't have to be .mcmeta)

shartte:
As we discussed yesterday, it could be possible to just have these on a per-directory level, but allow them to define pattern-based conditions. We could then cache these on a per-resource-manager level, reducing the time-to-lookup to once per directory. I'd also not make them "recurse" into subdirectories for simplicity.

shedaniel:
I have decided to cache it in ResourcePack level since you can have "duplicate" files such as with tags, you don't want to accidentally disable the vanilla one.

shedaniel:
The current solution of using WrappedResourcePack is not ideal, but that saves me from using a lot of redirects, I will continue to see if there are better solutions

haykam821 (changes requested):
Needs more codec.

shartte:
Hmm, I am not sure about it, you might actually want to disable the Vanilla one in case of recipes for example.
Would this _not_ allow you to have a conditional resource declaration that i.e. disables a certain vanilla recipe?

shedaniel:
Nope, this current PR just hides your json or file if the condition is not satisfied, you can only disable resources in the scope of your resource pack.

ghost:
100% in favor of this idea, I'll come back to review it when it's more finalized

Prospector:
This is a very cool idea, would love to see it implemented.

i509VCB:
Superseded by #1264

i509VCB:
Sorry wrong PR :stuck_out_tongue:

modmuss50:
Looks like a resonable start, will need to make sure that it doesnt have an impact on performace too much. I like the idea of applying this to whole folders 👍

liach:
A resource's condition should be determined at the time a game instance is ready but there is no client or server that would need any resource pack yet (i.e. the blocks, items, etc. are determined, but no resource pack is). Your approach currently determines it in the resource manager, which will be inadequate if we decide to build a giant resource pack to enhance mod compatibility, like in the situation mentioned in https://github.com/FabricMC/fabric/pull/1371#issuecomment-798950347.

I hope this can become part of #1373, an improved resource api. Imo this should just become part of the renovated resource api/resource helper than stay in a separate module. If we use these conditions to define custom rules for building the builtin resource pack provided by fabric mods, it would be extremely helpful!

Technici4n:
Superseded by #1395.

PR #1069, opened 2020-09-09
Server side Block place event

hollandjake:
Completes Issue #970

PR #1078, opened 2020-09-12
Added client side block place and break events

hollandjake:
Raised from https://github.com/FabricMC/fabric/pull/1069#issuecomment-690337405

This is a collection of client side events which client side mods can use to detect block breaks and placements.

The added events are as follows:

```
ClientBlockPlaceEvent.BEFORE
ClientBlockPlaceEvent.CANCELED
ClientBlockPlaceEvent.AFTER

ClientBlockBreakEvent.ON_START
ClientBlockBreakEvent.ON_PROGRESS
ClientBlockBreakEvent.ON_CANCEL
ClientBlockBreakEvent.ON_BREAK
```

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockBreakEvent.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockBreakEvent.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.player;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ClientBlockBreakEvent {
+	private ClientBlockBreakEvent() { }
+
+	/**
+	 * Callback before a block is to begin breaking.
+	 *
+	 * <p>If any listener cancels a block breaking action, that block breaking
+	 * action is canceled and {@link #ON_CANCEL} event is fired. Otherwise the
+	 * {@link #ON_PROGRESS} event is fired with a progress of 0.
+	 */
+	public static final Event<OnStart> ON_START = EventFactory.createArrayBacked(
+			OnStart.class,
+			listeners -> (world, player, pos, state, progress) -> {
+				for (OnStart event : listeners) {
+					boolean result = event.onBlockBreakStart(world, player, pos, state, progress);
+
+					if (!result) {
+						return false;
+					}
+				}
+
+				return true;
+			}
+	);
+
+	/**
+	 * Callback during a blocks breaking stages
+	 *
+	 * <p>If any listener cancels a block progress action, that block
+	 * action is canceled and {@link #ON_CANCEL} event is fired.
+	 */
+	public static final Event<OnProgress> ON_PROGRESS = EventFactory.createArrayBacked(
+			OnProgress.class,
+			listeners -> (world, player, pos, state, progress) -> {
+				for (OnProgress event : listeners) {
+					boolean result = event.onBlockBreakProgress(world, player, pos, state, progress);
+
+					if (!result) {
+						return false;
+					}
+				}
+
+				return true;
+			}
+	);
+
+	/**
+	 * Callback after a block is broken.
+	 * At the point of invocation you can ensure the client block has been destroyed from the world
+	 */
+	public static final Event<OnBreak> ON_BREAK = EventFactory.createArrayBacked(
+			OnBreak.class,
+			listeners -> (world, player, pos, state, progress) -> {
+				for (OnBreak event : listeners) {
+					event.onBlockBreak(world, player, pos, state, progress);
+				}
+			}
+	);
+
+	/**
+	 * Callback when a block break has been canceled.
+	 *
+	 * <p>Can be used to stop a client from breaking a block if maybe a mod only wants to
+	 * let you break a certain block at a certain time.
+	 */
+	public static final Event<OnCancel> ON_CANCEL = EventFactory.createArrayBacked(
+			OnCancel.class,
+			listeners -> (world, player, pos, state, progress) -> {
+				for (OnCancel event : listeners) {
+					event.onBlockBreakCancel(world, player, pos, state, progress);
+				}
+			}
+	);
+
+	@FunctionalInterface
+	public interface OnStart {
+		/**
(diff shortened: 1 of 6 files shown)


Geometrically (changes requested):
To be honest, I would just add this to the main BlockPlace/BlockBreak events. Then the event subscriber can do a world.isClient check to see whether they are on the server/client. It's a lot more cleaner, and less messier

hollandjake:
> To be honest, I would just add this to the main BlockPlace/BlockBreak events. Then the event subscriber can do a world.isClient check to see whether they are on the server/client. It's a lot more cleaner, and less messier

The problem with that is that if the client side mod wants to cancel the removal/placing of a block the serverside event wont be able to do this, as the packet has already been sent to the server. 

Another option is to use the same event but have two points which invoke it. this would create an abstraction for the modders, but under the hood it would be functionally the same as having both

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockBreakEvent.java
```diff
@@ -0,0 +1,167 @@
...
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ClientBlockBreakEvent {
```
i509VCB, line 27: Mark the class as client only `@Environment(EnvType.CLIENT)`

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockPlaceEvent.java
```diff
@@ -0,0 +1,122 @@
...
+
+	/**
+	 * Callback before a block is placed.
+	 * Called on both client and server
+	 *
+	 * <p>If any listener cancels a block placing action, that block placing
+	 * action is canceled and {@link #CANCELED} event is fired. Otherwise, the
+	 * {@link #AFTER} event is fired.</p>
+	 */
+	public static final Event<Before> BEFORE = EventFactory.createArrayBacked(
```
i509VCB, line 38: Similarly to my review on server events, the Before phase needs to have cancelation logic split out to an Allow phase

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinClientPlayerInteractionManager.java
```diff
@@ -50,16 +55,25 @@
...
 	private ClientPlayNetworkHandler networkHandler;
 	@Shadow
 	private GameMode gameMode;
 
+	@Shadow
+	public abstract void cancelBlockBreaking();
+
+	@Shadow
+	private float currentBreakingProgress;
+	private BlockState blockBreakingState;
```
i509VCB, line 75: Mark field as `@Unique`

fabric-events-interaction-v0/src/testmod/java/net/fabricmc/fabric/test/event/interaction/ClientBlockBreakEvents.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import net.minecraft.block.Blocks;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.client.player.ClientBlockBreakEvent;
+
+public class ClientBlockBreakEvents implements ModInitializer {
```
i509VCB, line 27: Entrypoint class name should have test in the name. I thought this was more events at first.

fabric-events-interaction-v0/src/testmod/java/net/fabricmc/fabric/test/event/interaction/ClientBlockPlaceEvents.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import net.minecraft.block.Blocks;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.client.player.ClientBlockPlaceEvent;
+
+public class ClientBlockPlaceEvents implements ModInitializer {
```
i509VCB, line 27: Same as other test entrypoint

i509VCB, line 27: Should be a client initializer.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockPlaceEvent.java
```diff
@@ -0,0 +1,122 @@
...
+				for (After event : listeners) {
+					event.afterBlockPlace(world, player, pos, state);
+				}
+			}
+	);
+
+	/**
+	 * Callback when a block place has been canceled.
+	 *
+	 * <p>Called on both client and server. May be used on logical server to send packets to revert client-side block changes.
```
i509VCB, line 70: Client event called on both logical sides?

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientBlockPlaceEvent.java
```diff
@@ -0,0 +1,122 @@
...
+		 *
+		 * <p>Implementations should not modify the world or assume the block place has completed or failed.</p>
+		 *
+		 * @param world the world in which the block is placed
+		 * @param player the player placing the block
+		 * @param pos the position at which the block is placed
+		 * @param state the block state <strong>before</strong> the block is placed
+		 * @return {@code false} to cancel block placing action, or {@code true} to pass to next listener
+		 */
+		boolean beforeBlockPlace(World world, PlayerEntity player, BlockPos pos, BlockState state);
```
i509VCB, line 94: Can specify ClientWorld and ClientPlayerEntity as the parameter types.

i509VCB (changes requested):

fabric-events-interaction-v0/src/testmod/resources/fabric.mod.json
```diff
@@ -10,7 +10,9 @@
   },
   "entrypoints": {
     "main": [
-      "net.fabricmc.fabric.test.event.interaction.PlayerBreakBlockTests"
+      "net.fabricmc.fabric.test.event.interaction.PlayerBreakBlockTests",
+      "net.fabricmc.fabric.test.event.interaction.ClientBlockPlaceEvents",
```
oliviathevampire, line 14: Should be in "client" since they should be client-side and not on both sides

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinClientPlayerInteractionManager.java
```diff
@@ -50,16 +55,25 @@
...
+public abstract class MixinClientPlayerInteractionManager {
 	@Shadow
 	private MinecraftClient client;
 	@Shadow
 	private ClientPlayNetworkHandler networkHandler;
 	@Shadow
 	private GameMode gameMode;
 
+	@Shadow
+	public abstract void cancelBlockBreaking();
```
oliviathevampire, line 71: This does not seem to be used? If it's not used, then it should be removed

oliviathevampire, line 71: This does not seem to be used? If it's not used, then it should be removed

i509VCB:
Superseded by #1264

PR #247, opened 2019-06-05
Add events for network connections

killjoy1221:
Adds events for LOGIN, JOIN, LEAVE on client and server.

Usage:
```java
S2CPlayConnectCallback.EVENT.register((connection, packetListener) -> {
     // joined on client
    ClientPlayNetworkHandler handler = (ClientPlayNetworkHandler) packetListener;
 });
C2SPlayConnectCallback.EVENT.register((connection, packetListener) -> {
    // joined on server
    ServerPlayNetworkHandler handler = (ServerPlayNetworkHandler) packetListener;
});
```

PR #605, opened 2020-05-03
Add Fabric Networking API v1

liach:
This fabric networking API v1 is a low-level API (or more exactly, hooks) that covers basic custom-payload/plugin message packets offered by the vanilla Minecraft protocol.

Usage: Users are supposed to use stuff exposed in `ClientNetworking` or `ServerNetworking`, or utilities in `net.fabricmc.fabric.api.networking.v1.util` package.

Initial content based off FabLabsMC/networking-api-v1-draft@eb256236cdfa9cb3b0697ceb12a957c47582fec3.
Javadoc for draft available at https://fablabsmc.github.io/networking-api-v1-draft/

Will perform v0 deprecation/migration only after the new api is accepted, or this pr is going to be too large.

Supersedes #455. Issue coverage: #164, #187, #247, #345, #424

Thanks @i509VCB for previewing the draft before I submit this pull request.

Signed-off-by: liach <liach@users.noreply.github.com>

PR #164, opened 2019-04-27
add (currently draft/internal) networking handshake API

asiekierka:
Other changes:

* Mark the EntityTracker mixin in fabric-networking as optional (it only adds internal APIs, and the method using them has a fallback)
* Clean up fabric-registry-sync a little

PR #598, opened 2020-04-30
Add 2020 to license headers

YanisBft:
Probably the most controversial PR in this repo

PR #1081, opened 2020-09-13
Fabric Networking API V1

### LINKED ISSUE #187: ClientSidePacketRegistry.INSTANCE.register can be called on dedi server, resulting in crash
`ClientSidePacketRegistry.INSTANCE.register()` doesn't have an embedded environment check, so if the code gets called on a dedicated server, it will crash due to NoClassDefFoundError for MinecraftClient.

### LINKED ISSUE #345: Expose NetworkHandler or ClientConnection to clientside network events, fix player retrieval
MC creates and sets the player on the client thread by synchronizing the join packet, which happens after receiving other packets.

In the meanwhile `PacketContext.getPlayer()` returns null and the connection is not easily unobtainable for an immediate reply (only reflectively from `MinecraftClient.clientConnection`, `ConnectScreen.connection` or `RealmsConnect.connection` depending on the connection origin).

The potential null result and thread unsafety[1] of `PacketContext.getPlayer()` is problematic, maybe it's better to block its use on the network threads. Most of those uses are likely a bug, with the most significant exception being the serverside `NetworkThreadUtils.forceMainThread(..., player.getServerWorld())`, which could be replaced by exposing the server conveniently instead.

[1] There are no guarantees on the chronologically "correct" value propagating without proper synchronization. It may be significantly delayed or even sporadically revert to an outdated value between consecutive field reads (e.g. thread moved between cores and new core's cache has older state).

i509VCB:
This is partially based of off and supersedes #605. This pull request is based off the `feature/cut` branch on the fablabs drafting repository. This pull request was primarily written by @liach but has been adapted into a new pull request on the FabLabs fabric api fork to continue the pull request.

This pull request migrates the networking api to v1 and deprecates the old api (minus player streams which need to be migrated by a different pull request). The old networking api is re-implemented using the new networking api in a backwards compatible way.

Registration of custom packet handlers is now done through the client or server networking's play receiver. The interface used to read the packet and apply (originally PacketConsumer) is now much more ergonomic, as it exposes the client and serve respectively along with the network handler. This effectively resolves #345.

This pull request also adds the ability for mods to send and handle login query messages. This allows the client and server to communicate much earlier and delay login for the client/server until tasks have been completed on either side (async tasks included).

Events related to the connection/disconnection of the client or server were added to the respective `Client/ServerConnectionEvents`. This effectively resolves #247

Also resolves #187, and prepares the groundwork for a system described in #164.


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientConnectionEvents.java
```diff
@@ -0,0 +1,69 @@
...
+			callback.onPlayInitialized(handler, client, sender);
+		}
+	});
+
+	/**
+	 * An event for the disconnection of the client play network handler.
+	 *
+	 * <p>No packets should be sent when this event is invoked.
+	 */
+	public static final Event<PlayDisconnected> PLAY_DISCONNECTED = EventFactory.createArrayBacked(PlayDisconnected.class, callbacks -> (handler, client) -> {
```
RDIL: This might just be my computer but it looks like this indent is different then below?

i509VCB: I think GH is funky with tabs sometimes

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ChannelHandlerRegistry.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Associates packets to individual packet reception handlers by channel.
+ *
+ * @param <H> the channel handler type
+ */
+public interface ChannelHandlerRegistry<H> extends ChannelRegistry {
+	/**
+	 * Registers a handler to a channel.
+	 *
+	 * <p>If a handler is already registered to the {@code channel}, this method will return {@code false}, and no change will be made.
```
ghost: `this method should return {@code false}` because ultimately implementation details can't definitively be known by interface exposure, but we can suggest that would be the norm

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientChannelEvents.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Offers access to events related to the registration of network channels for a client-side network handler.
+ */
+@Environment(EnvType.CLIENT)
+public final class ClientChannelEvents {
+	/**
+	 * An event for the client play network handler receiving an update indicating the connected server's ability to receive packets in certain channels.
+	 *
+	 * @see PlayPacketSender#hasChannel(Identifier)
+	 */
```
ghost: This could be further clarified. At first glance understanding what this does is a little difficult. Like when exactly does this trigger.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientChannelEvents.java
```diff
@@ -0,0 +1,71 @@
...
+		for (Registered callback : callbacks) {
+			callback.onChannelRegistered(handler, client, sender, channels);
+		}
+	});
+
+	/**
+	 * An event for the client play network handler receiving an update indicating the connected server's lack of ability to receive packets in certain channels.
+	 *
+	 * @see PlayPacketSender#hasChannel(Identifier)
+	 */
```
ghost: Same here

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientNetworking.java
```diff
@@ -0,0 +1,151 @@
...
+	 * @return the associated packet sender
+	 */
+	public static PlayPacketSender getPlaySender(ClientPlayNetworkHandler handler) {
+		return ClientNetworkingDetails.getAddon(handler);
+	}
+
+	/**
+	 * Returns the packet receiver for channel handler registration on client play network handlers, receiving {@link net.minecraft.network.packet.s2c.play.CustomPayloadS2CPacket server to client custom payload packets}.
+	 */
+	public static ChannelHandlerRegistry<PlayChannelHandler> getPlayReceiver() {
```
ghost: I might change this to `getPlayReceivers` to clarify that this isn't returning one specific play receiver but a registry of play receivers

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientNetworking.java
```diff
@@ -0,0 +1,151 @@
...
+		return ClientNetworkingDetails.PLAY;
+	}
+
+	/**
+	 * Returns the packet receiver for channel handler registration on client login network handlers, receiving {@link net.minecraft.network.packet.s2c.login.LoginQueryRequestS2CPacket login query request packets}.
+	 */
+	public static ChannelHandlerRegistry<LoginChannelHandler> getLoginReceiver() {
```
ghost: Same here `getLoginReceivers`

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientNetworking.java
```diff
@@ -0,0 +1,151 @@
...
+		return ClientNetworkingDetails.LOGIN;
+	}
+
+	/**
+	 * Sends a packet to the connected server.
+	 *
+	 * @param channel the channel of the packet
+	 * @param buf the payload of the packet
+	 */
+	public static void send(Identifier channel, PacketByteBuf buf) {
```
ghost: Doesn't this have the opportunity to throw an exception? If so, that should be mentioned in the javadoc.

ghost: Also maybe there should be a method similar to `getPlaySender(ClientPlayNetworkHandler handler)` and `getPlaySender() throws IllegalStateException` that is `send(ClientPlayNetworkHandler handler, Identifier channel, PacketByteBuf buf)`

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ClientNetworking.java
```diff
@@ -0,0 +1,151 @@
...
+	@Environment(EnvType.CLIENT)
+	@FunctionalInterface
+	public interface PlayChannelHandler {
+		/**
+		 * Handles an incoming packet.
+		 *
+		 * <p>This method is executed on {@linkplain io.netty.channel.EventLoop netty's event loops}.
+		 * Modification to the game should be {@linkplain net.minecraft.util.thread.ThreadExecutor#submit(Runnable) scheduled} using the provided Minecraft client instance.
+		 *
+		 * <p>An example usage of this is to display a
```
ghost: Where did the rest of this sentence go. I'm dying to know what it says

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PlayPacketSender.java
```diff
@@ -0,0 +1,29 @@
...
+
+package net.fabricmc.fabric.api.networking.v1;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * Supports sending packets to channels in the play network handlers.
+ *
+ * <p>Compared to a simple packet sender, the play packet sender is informed if its connected recipient may {@link #hasChannel(Identifier) accept packets in certain channels}.
+ * <!--TODO: Implement use of system property-->
```
ghost: Just highlighting this as something that is on the to-do list for others looking at this review to get a synopsis

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerChannelEvents.java
```diff
@@ -0,0 +1,66 @@
...
+		for (Registered callback : callbacks) {
+			callback.onChannelRegistered(handler, server, sender, channels);
+		}
+	});
+
+	/**
+	 * An event for the server play network handler receiving an update indicating the connected client's lack of ability to receive packets in certain channels.
+	 *
+	 * @see PlayPacketSender#hasChannel(Identifier)
+	 */
```
ghost: Similar to the comment I made on the client side, more specifics about when this occurs would be good in the javadoc

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerChannelEvents.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Offers access to events related to the registration of network channels for a server-side network handler.
+ */
+public final class ServerChannelEvents {
+	/**
+	 * An event for the server play network handler receiving an update indicating the connected client's ability to receive packets in certain channels.
+	 *
+	 * @see PlayPacketSender#hasChannel(Identifier)
+	 */
```
ghost: Same here see below

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractChanneledNetworkAddon.java
```diff
@@ -0,0 +1,173 @@
...
+
+import net.minecraft.network.ClientConnection;
+import net.minecraft.network.PacketByteBuf;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.InvalidIdentifierException;
+
+import net.fabricmc.fabric.api.networking.v1.PacketByteBufs;
+import net.fabricmc.fabric.api.networking.v1.PlayPacketSender;
+
+// play
```
ghost: What's this comment for?

fabric-networking-v0/src/main/java/net/fabricmc/fabric/impl/networking/OldNetworkingHooks.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.networking;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.network.C2SPacketTypeCallback;
+import net.fabricmc.fabric.api.networking.v1.ServerChannelEvents;
+
+public final class OldNetworkingHooks implements ModInitializer {
```
ghost, line 23: This is definitely me missing the point, but what is this for exactly?

ghost (approved):
Everything looks good I mainly just had some minor concerns over some of the documentation and beefing that up. After looking through it for a while it actually wasn't that bad making sense of all of it, but for the average user this whole module is kind of overwhelming so having more details about exactly when certain events are occurring and that sort of thing might be helpful. Other than that I saw a couple places where there were TODOs floating around, some of them looked like integrations with future modules, some just looked like things that have yet to be done. Regardless I'm sure they're in the works and will be done before final pull. To me it looks pretty good I don't really have any concerns. The one design detail I was hesitant about was having only a single handler per channel but that seems best for simplicity and people who want multiple handlers per channel just need to fix their impl because the only way I could see you wanting that would be from trying to implement a badly designed idea.

fabric-networking-v0/src/main/java/net/fabricmc/fabric/impl/networking/OldNetworkingHooks.java
i509VCB: That class implements the old `PacketType` events using the new `ClientChannelEvents` and `ServerChannelEvents`

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ChannelHandlerRegistry.java
i509VCB: Imo specifying semantics are fine in javadoc for an interface

Player3324 (approved):
API looks good, thanks for this extensive effort!

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/client/networking/v1/ClientLoginConnectionEvents.java
```diff
@@ -0,0 +1,99 @@
...
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.network.ClientLoginNetworkHandler;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Offers access to events related to the connection to a server on the client while the server is processing the client's login request.
+ */
+public final class ClientLoginConnectionEvents {
```
shedaniel, line 31: This (and its inner classes) should be annotated with @Environment(EnvType.CLIENT)

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PlayerLookup.java
```diff
@@ -0,0 +1,193 @@
...
+	 * @param radius the maximum distance from the position in blocks
+	 * @return the players around the position
+	 */
+	public static Collection<ServerPlayerEntity> around(ServerWorld world, Vec3d pos, double radius) {
+		double radiusSq = radius * radius;
+
+		return world(world)
+				.stream()
+				.filter((p) -> p.squaredDistanceTo(pos) <= radiusSq)
+				.collect(Collectors.toList());
```
shedaniel, line 169: Would returning an `Iterable` or `Iterator` be sufficient in order to not creating a list?

Player3324: Having the size is quite nice and it needs to create something, the (Array)List should be one of the cheapest options.

sschr15 (commented):
The copyright in the headers does not include 2020

haykam821:
Don't worry, that isn't new: #598

i509VCB:
The license header year will be handled later when I make Fabric API use JDK style license headers

fabric-player-tracking-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,26 @@
+{
```
modmuss50: I dont think this player tracking stuff belongs here? The module is not included in gradle?

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/login/NetworkingLoginQueryTest.java
```diff
@@ -0,0 +1,76 @@
...
+		ServerLoginConnectionEvents.LOGIN_QUERY_START.register(this::onLoginStart);
+		ServerLoginConnectionEvents.LOGIN_QUERY_START.register(this::delaySimply);
+
+		// login delaying example
+		ServerLoginNetworking.registerGlobalReceiver(NetworkingPlayPacketTest.TEST_CHANNEL, (server, handler, understood, buf, synchronizer, sender) -> {
+			if (understood) {
+				FutureTask<?> future = new FutureTask<>(() -> {
+					for (int i = 0; i <= 10; i++) {
+						Thread.sleep(300);
+						NetworkingTestmods.LOGGER.info("Delayed login for number {} 300 milliseconds", i);
```
modmuss50: Disable this by default, its a fine test just dont really want this all the time when using the test mods :D

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/keybindreciever/NetworkingKeybindPacketTest.java
```diff
@@ -0,0 +1,54 @@
...
+	private static void receive(MinecraftServer server, ServerPlayerEntity player, PacketByteBuf buf, PacketSender responseSender) {
+		// TODO: Can we send chat off the server thread?
+		server.execute(() -> {
+			player.sendMessage(new LiteralText("So you pressed ").append(new KeybindText("fabric-networking-api-v1-testmod-keybind").styled(style -> style.withFormatting(Formatting.BLUE))), false);
+		});
+	}
+
+	@Override
+	public void onInitialize() {
+		// FIXME: DO NOT LET ME PR THIS BEFORE FIXING IT
```
modmuss50: Need fixing apparenly.

modmuss50 (changes requested):
Looking really good 👍 few very minor things around the test mods.

This will need a decent amount of testing, including vanilla/mutliversion servers (where we have had issues before).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/QueryIdFactory.java
```diff
@@ -0,0 +1,39 @@
...
+package net.fabricmc.fabric.impl.networking.server;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Tracks the current query id used for login query responses.
+ */
+interface QueryIdFactory {
+	static QueryIdFactory create() {
+		// todo incremental ids or randomized
```
FlashyReese: Is already done? Should remove

modmuss50 (approved):
Awesome 🍾 

Nice work, the API looks solid 👍

sschr15 (approved):

FlashyReese (approved):

marcus8448 (approved):

NeusFear (approved):

NeusFear:
Sorry, had stashed comments apparently.

LambdAurora (approved):

OroArmor (approved):
Ok its time to merge this in

modmuss50:
Lets go 🚀 

Thanks for this.

PR #1082, opened 2020-09-13
Generify more of registry-sync-v0 and remove unchecked casts where possible

shartte:
Cleaned up use of Generics in registry-sync-v0 and used Registry#getKey where possible instead of reverse-lookup of the registry key.

There's also a Registry#getKey function that returns the registries own registry key, which I used in a few places where previously the ID was being looked up in Registry.ROOT (this makes it easier to support registries that aren't in there, such as biomes later).

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdList.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdList.java
@@ -31,13 +31,13 @@ import net.minecraft.util.collection.IdList;
 import net.fabricmc.fabric.impl.registry.sync.RemovableIdList;
 
 @Mixin(IdList.class)
-public class MixinIdList implements RemovableIdList<Object> {
+public class MixinIdList<T> implements RemovableIdList<T> {
 	@Shadow
 	private int nextId;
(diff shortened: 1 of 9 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdList.java
```diff
@@ -86,15 +86,15 @@ public void fabric_remapId(int from, int to) {
 	@Override
 	public void fabric_remapIds(Int2IntMap map) {
 		// remap idMap
-		idMap.replaceAll((a, b) -> map.get(b));
+		idMap.replaceAll((a, b) -> map.get((int) b));
```
shartte, line 89: IDE was complaining about get(Object) being deprecated in the FastUtils collection,  hence the explicit cast here.

i509VCB (approved):

ghost (approved):

modmuss50 (approved):
LGTM

frqnny (approved):

Player3324 (approved):
LGTM, nice cleanup. Thanks!

oliviathevampire (approved):

i509VCB (approved):

PR #1085, opened 2020-09-14
Add FabricDamageSourceBuilder

JanThomas118:
I created this very simple PR to avoid Boilerplate code when creating custom DamageSources in my mod.  

The issue is that almost all setters in the vanilla DamageSource class are protected, so in order to create your own you need to subclass DamageSource and copy all the setter methods into your class.  

I have included methods which were already public in the superclass to make this more maintainable in case the access modifiers in vanilla code change.

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDamageSourceBuilder.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDamageSourceBuilder.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDamageSourceBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package net.fabricmc.fabric.api.object.builder.v1.entity;
+
+import net.minecraft.entity.damage.DamageSource;
+
+/**
+ * Allows creating custom Damage Sources without subclassing it.
+ */
+public class FabricDamageSourceBuilder extends DamageSource {
```
i509VCB, line 24: This isn't really a builder for say? `FabricDamageSource` is probably fine for this.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDamageSourceBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.object.builder.v1.entity;
+
+import net.minecraft.entity.damage.DamageSource;
+
+/**
+ * Allows creating custom Damage Sources without subclassing it.
```
i509VCB, line 22:
```suggestion
 * Allows creation of custom Damage Sources without subclassing.
```

i509VCB (commented):
Looks good overall, class name probably needs a bit of a discussion.

oliviathevampire:
Maybe make it an actual builder and have a method called build or create which returns a damage source?

ghost:
Maybe using transitive access wideners?

Until then i will copy this file into my mod, by crediting the source... xD

Technici4n:
Superseded by #2341.

PR #517, opened 2020-03-09
Migrate networking player tracking to new module

liach:
This player tracking module has a horrible name. Please suggest a better name for me (guess it can match the new api name pattern suggested in #575)

PR #1092, opened 2020-09-19
Player tracking api

i509VCB:
Supersedes #517. Splits off the player tracking aspect of v0 networking into a new module. Also adds events that are called when an entity starts or stops being tracked by a player.

`PlayerStream` isn't great tbf. A better name would be nice, possibly just going for `PlayerTracking`?


JamiesWhiteShirt:
We should also consider being consistent with "watching" vs. "tracking", as this API uses both. This is more of a problem in Yarn than anything, but this API can be consistent.

My opinions:
- I find "tracking" to be the most appropriate term.
- If we should preserve the term "player stream", it should be called `PlayerStreams`, not `PlayerStream`. But yes, it's not great because the term is too generic. +1 for `PlayerTracking`.

i509VCB:
I've decided on "tracking" for the naming purposes of this api.

fabric-player-tracking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/player/tracking/v1/package-info.java
```diff
@@ -14,12 +14,12 @@
...
-package net.fabricmc.fabric.impl.networking.server;
-
-import java.util.stream.Stream;
-
-import net.minecraft.server.network.ServerPlayerEntity;
+/**
+ * The fabric networking player tracking API, version 1.
+ *
+ * <p>This API allows mods to easily select player recipients of packets
+ * based on distance from a position, their watching status of server
```
JamiesWhiteShirt:
```suggestion
 * based on distance from a position, their tracking status of server
```
Missed one.

fabric-player-tracking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/player/tracking/v1/PlayerTracking.java
```diff
@@ -0,0 +1,180 @@
...
+		return tracking((ServerWorld) blockEntity.getWorld(), blockEntity.getPos());
+	}
+
+	/**
+	 * Gets all players tracking a block position in a server world.
+	 *
+	 * @param world the server world
+	 * @param pos   the block position
+	 * @return the players tracking the block position
+	 * @throws IllegalArgumentException if the world is not a server world
```
ghost: Is this true? I only see NullPointerExceptions, also this requires an instance of ServerWorld ( or subclass )

This comment also applies to the two methods below this one.

i509VCB: Those javadoc throws tags are not removed since they don't apply any more.

fabric-player-tracking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/player/tracking/v1/PlayerTracking.java
```diff
@@ -0,0 +1,182 @@
...
+import net.fabricmc.fabric.impl.networking.player.tracking.ThreadedAnvilChunkStorageTrackingExtensions;
+
+/**
+ * For example, a block entity may use the methods in this class to send a packet to all clients which can see the block entity in order notify clients about a change.
+ *
+ * <p>The word "tracking" means that an entity/chunk on the server is known to a player's client (within in view distance) and the (block) entity should notify tracking clients of changes.
+ *
+ * <p>These methods should only be called on the server thread and only be used on logical a server.
+ */
+public final class PlayerTracking {
```
Player3324: The name isn't great, particularly when combined with some methods, like "PlayerTracking.all" or "PlayerTracking.tracking". Maybe something like PlayerLookup?

fabric-player-tracking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/player/tracking/v1/PlayerTracking.java
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/**
+	 * Gets all players tracking a chunk in a server world.
+	 *
+	 * @param world the server world
+	 * @param pos   the chunk in question
+	 * @return the players tracking the chunk
+	 */
+	public static Stream<ServerPlayerEntity> tracking(ServerWorld world, ChunkPos pos) {
```
Player3324: There is some inconsistency between Stream and Collection use. I'd use Collection everywhere since partial consumption is unlikely and it's the simpler API.

Player3324 (commented):
Splitting such a small and stable API off is IMO a bad idea, each module comes with some complexity and overhead cost. The events are fine, the player lookup helper could use a few minor tweaks.

i509VCB:
Closing since this is being cherry-picked into #1081

PR #1094, opened 2020-09-20
Tweak entity type builder generics to eliminate explicit type arguments

i509VCB:
This is most definitely not a great solution as it allows people to over-specify the type argument and therefore cause class cast exceptions at creation time.

The intention here is to remove the need for explicit type arguments:

Before:
```java
// Omit type argument and this fails to compile
static EntityType<TestEntity> OLD_TEST = FabricEntityTypeBuilder.<TestEntity>createLiving()
		.entityFactory(TestEntity::new)
		.spawnGroup(SpawnGroup.CREATURE)
		.build();
```

After:
```java
// Changes make this compile
static EntityType<TestEntity> TEST_ENTITY_1 = FabricEntityTypeBuilder.createLiving()
		.entityFactory(TestEntity::new)
		.spawnGroup(SpawnGroup.CREATURE)
		.build();
```

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
@@ -136,10 +136,10 @@ public class FabricEntityTypeBuilder<T extends Entity> {
 		return this;
 	}
 
-	public FabricEntityTypeBuilder<T> entityFactory(EntityType.EntityFactory<T> factory) {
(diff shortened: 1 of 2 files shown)


ghost (approved):

frqnny (approved):
I've always used FabricEntityTypeBuilder#create instead of createLiving, even for living entities. why do we need createliving in the first place btw?

i509VCB:
The living and mob builders are to handle entity attributes for living entities and spawn restrictions for mobs.

shartte (approved):

YanisBft (approved):

Juuxel (approved):

modmuss50 (approved):

Player3324 (approved):
Good now, thanks!

PR #1095, opened 2020-09-22
Extra block settings

Shnupbups:
Adds new things to FabricBlockSettings to allow for overriding values from a block's material.
Currently includes Piston Behavior, Replaceable, and Solid.
A Test Mod is included, featuring 2 different blocks that use these new settings.

So if you want a block of the `WOOD` material to be immovable by pistons, you don't have to make a new Block class and override `getPistonBehavior` just to do it, you just slap it in the block settings.

As a side effect, also adds methods for `isReplaceable` and stuff (with a `BlockState` argument!) for easy checking of things that are otherwise annoying to check as well as simpler overriding for those things.

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricBlockInternals.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricBlockInternals.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.piston.PistonBehavior;
 import net.minecraft.item.Item;
 import net.minecraft.tag.Tag;
 
@@ -51,12 +52,27 @@ public final class FabricBlockInternals {
 			for (MiningLevel tml : data.miningLevels) {
 				ToolManagerImpl.entry(block).putBreakByTool(tml.tag, tml.level);
 			}
+
+			if (data.pistonBehavior != null) {
+				((AbstractBlockInternals) block).setPistonBehavior(data.pistonBehavior);
+			}
+
+			if (data.replaceable != null) {
+				((AbstractBlockInternals) block).setReplaceable(data.replaceable);
+			}
+
+			if (data.solid != null) {
+				((AbstractBlockInternals) block).setSolid(data.solid);
+			}
 		}
 	}
 
 	public static final class ExtraData {
 		private final List<MiningLevel> miningLevels = new ArrayList<>();
 		/* @Nullable */ private Boolean breakByHand;
+		/* @Nullable */ private PistonBehavior pistonBehavior;
+		/* @Nullable */ private Boolean replaceable;
+		/* @Nullable */ private Boolean solid;
 
 		public ExtraData(Block.Settings settings) {
 		}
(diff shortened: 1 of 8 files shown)


YTG1234 (approved):
Yes

Leo40Git (approved):
~~Might want to run the "format licenses" task (I... forgot what it's called).~~ edit: you did it the exact moment I submitted the review
Looks good otherwise, apart from the probably accidental reformatting in the mixins JSON.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricBlockInternals.java
```diff
@@ -51,12 +52,27 @@ public static void onBuild(Block.Settings settings, Block block) {
...
+
+			if (data.solid != null) {
+				((AbstractBlockInternals) block).setSolid(data.solid);
+			}
 		}
 	}
 
 	public static final class ExtraData {
 		private final List<MiningLevel> miningLevels = new ArrayList<>();
 		/* @Nullable */ private Boolean breakByHand;
```
ghost: These can be made `@Nullable` now as we have annotations

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/AbstractBlockMixin.java
```diff
@@ -0,0 +1,97 @@
...
+import net.minecraft.item.ItemPlacementContext;
+
+import net.fabricmc.fabric.impl.object.builder.AbstractBlockInternals;
+
+@Mixin(AbstractBlock.class)
+public abstract class AbstractBlockMixin implements AbstractBlockInternals {
+	@Shadow
+	@Final
+	protected Material material;
+	/* @Nullable */ private PistonBehavior pistonBehavior;
```
ghost: Same here, can be made `@Nullable`

ghost (commented):
Looks good other than the annotation comments.

i509VCB:
I personally don't see a real purpose for the additions here. Yes it may make things more convenient but this just adds duplicate registration logic where the (Fabric)MaterialBuilder or overriding methods could suffice.

ghost (approved):

Mysterious-Dev (approved):

ghost:
Could I convince you into making an extra block setting for modifying the addition of block state definitions? Currently the construction of the block state manager is the first thing that happens in the super constructor, meaning the only way to pass a static variable into the constructor of a block and have that modify the value for a block state declaration, is to use the block settings parameter as a mule and then mix into the AbstractBlock constructor (hey new mixin feature) to modify the block state before the block state manager is constructed. So while the appendProperties method is not inherently static it's effectively static by virtue of when it's called. The only other alternative is completely reconstructing the block state manager and just like totally borking some stuff that shouldn't be borked with. Potential applications include using the same constructor to make blocks very similar but with different block state variations. Crops are a good example. Making new crops with different ages requires making a new class declaration for every single crop (instead of reusing the same constructor) because the block state for the age is effectively static.

If not in this PR I might make a new PR for it.

Technici4n:
What's the point of this?

apple502j:
Yeah, also this is very outdated. Closing.

PR #203, opened 2019-05-22
Enchantment API

Siphalor:
Would fix #202.

This adds an api for adding custom enchantment targets.
Since vanilla is using an enum for them this pr heavily relies on `Redirect` and modifying the `EnchantmentTarget.ALL`.

It also adds an interface for blocks that provide enchantment power to the enchanting table.

I still don't know if I did the subproject right.

PR #1102, opened 2020-09-26
fabric-enchanting-api-v1 updated for 1.16.3

### LINKED ISSUE #202: API for EnchantmentTargets
Enchantment targets (so like different kinds of tools or the armor pieces) are in an ugly enum.

It would be nice to be able to easily register own ones without having to `@Redirect` or `@Overwrite` a lot of the games logic.

ghost:
Updated the proposed enchantment target api for 1.16.3. Seems like Mojang removed `EnchantmentTarget.ALL`, so this version uses a null value to indicate that functionality should be handled in another way. In the process I replaced the redirect mixin with a more compatibility forward inject. This should also fix the concerns about thread safety. Closes #203. Closes #202.


fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantedBookItemMixin.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// This target mixes in right at the end of the append stacks method
+	@Inject(method = "appendStacks", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+	public void appendStacks(ItemGroup group, DefaultedList<ItemStack> stacks, CallbackInfo callback) {
+		// Get an iterator for the enchantment registry. This mixin iterates over the enchantment
+		// registry a second time, which does increase initialization time but because this method
+		// is only called once in the initialization processs, and alternative mixins would be more
+		// invasive, this is how it is.
+		Iterator<Enchantment> iterator = Registry.ENCHANTMENT.iterator();
```
i509VCB: A Registry is Iterable, could just get away with an enchanced for loop

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantmentHelperMixin.java
```diff
@@ -0,0 +1,79 @@
...
+			if (var6.hasNext()) {
+				// We just set the enchantment to the next iterator value
+				// This will make the while false and it'll go back to vanilla functionality with a vanilla enchantment with a valid enchantment target
+				enchantment = var6.next();
+			// If the iterator doesn't have another value this is more tricky
+			// Simply ending the mixin without making the enchantment value have a valid target means that a null pointer will be thrown
+			} else {
+				// So, because we know the iterator is empty we know we can safely return the enchantment entry list without any unintended side effects
+				callback.setReturnValue(list);
+				callback.cancel();
```
i509VCB: `setReturnValue` implicitly calls `cancel` so you can drop that one

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/FabricEnchantment.java
```diff
@@ -0,0 +1,74 @@
...
+	 *
+	 * @see net.minecraft.item.Item#isEnchantable(ItemStack)
+	 * @see net.minecraft.screen.EnchantmentScreenHandler#onContentChanged(Inventory)
+	 * @see net.fabricmc.fabric.mixin.enchantment.EnchantmentHelperMixin
+	 */
+	@Override
+	public abstract boolean isAcceptableItem(ItemStack stack);
+
+	/**
+	 * A custom method that determines whether or not this enchantment's books
```
i509VCB: You could probably drop the `A custom method that` part of the javadoc

ghost: Changed in 4f13b2f.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantmentHelperMixin.java
ghost: Awesome, wasn't sure about that. Changed it in 4f13b2f.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantedBookItemMixin.java
ghost: Resolved in 4f13b2f.

shartte (approved):
LGTM

Prospector (commented):
Null value makes more sense to me anyway. The mixin code is heavily commented which is great for future maintenance. One thing I am curious about is how would someone make an enchantment that can, for example, be applied to a certain item only via enchanted book but not via enchanting table? Siphalor in their PR sorta explained this here: https://github.com/FabricMC/fabric/pull/203/files#r491412775

Idk if this is possible with your current api. A potential idea is to override that one method and make it final, and then add two new methods with better names for enchanting vs. applying enchants. Looks very close to good to merge, nice work!

ghost:
> One thing I am curious about is how would someone make an enchantment that can, for example, be applied to a certain item only via enchanted book but not via enchanting table? Siphalor in their PR sorta explained this here: https://github.com/FabricMC/fabric/pull/203/files#r491412775
> 
> Idk if this is possible with your current api. A potential idea is to override that one method and make it final, and then add two new methods with better names for enchanting vs. applying enchants. Looks very close to good to merge, nice work!

Yeah that sounds like that might be a good direction to go into. I wonder how vanilla does it, might be something to look into. If it does become two methods (one for can enchant in table and one for can enchant with book) I think it would also make sense to then have the can enchant in table method override the `Item.isEnchantable` method because currently an item that doesn't say that it's enchantable can't be enchanted in an enchantment table no matter what, but if the enchantment is saying it should specifically in a table then maybe there's reason to overwrite that.

ghost:
Apparently the way vanilla does it, is that some enchantments indicate that they can be added to an item, but those items don't indicate that they're enchantable, so the result is that the only way the enchantment can be added to the item is by using an anvil. I think this means we should just add one "isEnchantable" method that both overwrites the item's indication of enchantability, and states that the enchantment can be applied to the item.

ghost:
Actually it might be wiser to just let individual mod authors mix into the item they specifically want to target, because in the grand scheme of things it's about the simplest inject mixin possible and so it's really not essential to be in fabric

ghost:
This would actually probably be a good idea because it would cut out this extra registry iterator I had to make by instead having static compile time knowledge of the items that need to be enchantable or not.

ghost:
Okay I addressed the functionality Pros mentioned in https://github.com/FabricMC/fabric/pull/1102/commits/6aa093b2255a6fd201329c3bb23f70192fcb58bd. There was some discussion on how to handle this anvil situation. I chose to keep the actual code in the PR but disable the mixin. I also included an explanation at the head of the file that it's disabled because it doesn't fit with vanilla, but included for the sake of illustrating how to fix the problem if so desired.

Other than that this PR should be done I'm going to re-request reviews from the people who submitted feedback and then someone should request a review from muss and player.

NeusFear (approved):

ghost:
Alrighty, this should be done now! I can't conceive of anything else I'd want to add to it. There's now a callback event for applying an enchantment in the anvil, and enchanting in the table. This should hopefully justify the use of redirects in these sticky situations. Additionally, there's a R/W lock on the whole capture local cache kind of thing so that should quell any concerns over thread safety. That being said I am not a multi-threading professional and I've only done it in limited capacity in Rust and never in Java so conceptually I think it works, and it seems like it's functional in game, but I don't know how terse a solution it is.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/AnvilScreenHandlerMixin.java
```diff
@@ -0,0 +1,51 @@
...
+public class AnvilScreenHandlerMixin {
+	@Redirect(method = "updateResult", at = @At(value = "INVOKE", target = "Lnet/minecraft/enchantment/Enchantment;isAcceptableItem(Lnet/minecraft/item/ItemStack;)Z"))
+	public boolean updateResult(Enchantment enchantment, ItemStack itemStack) {
+		TriState eventCallback = EnchantmentEvents.ACCEPT_APPLICATION.invoker().shouldAccept(enchantment, itemStack);
+
+		if (eventCallback != TriState.DEFAULT) {
+			return eventCallback.get();
+		}
+
+		return enchantment.isAcceptableItem(itemStack);
```
Ueaj-Kerman, line 49: This could be replaced with a modifyarg and pass it an empty itemstack, but it doesn't really matter

Ueaj-Kerman (approved):

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/AnvilScreenHandlerMixin.java
i509VCB: :thinking:

ghost: Wouldn't that require another Inject capture locals? I went down that road but it didn't seem like ModifyArg allowed capturing locals.

ghost: Nvm I get what you mean, I'll look into it

ghost:
Ahhh I screwed things up oh no, okay I can fix this

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/AnvilScreenHandlerMixin.java
ghost: Actually, I don't think that's possible. I just went to actually try and implement what you were talking about, but there's no way to ensure that an item stack that's passed will always make the method true (Like there exists no item stack with which we could be sure that method will always return true) so it would completely get rid of the ability to have the callback determine the enchantability.

ghost:
Real talk I have no clue what's going on with the build test right now, it would be awesome if someone could take a look at that, because locally everything seems to work fine.

modmuss50:
You have a checkstyle violation somewhere, would be useful if the gradle log output showed it. There are IDE plugins that will aid with this.

ghost:
Fixed, could've sworn it was meeting the checkstyle earlier, really thought it was something else causing the error.

fabric-enchantment-api-v1/README.md
```diff
@@ -0,0 +1,40 @@
+# Fabric Enchantment API (v1)
```
i509VCB, line 1: Huh you wrote a readme. Neat.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,102 @@
...
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.api.util.TriState;
+
+/**
+ * Events to intercept the vanilla enchantment logic.
+ *
+ * @author Vaerian (vaeriann@gmail.com or @Vaerian on GitHub).
```
i509VCB, line 31: Author tag isn't required, we can just `git blame` in the worst case

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * <p>Please contact the author, Vaerian, at the email or GitHub profile listed above
+ * with any questions surrounding implementation choices, functionality, or updating
+ * newer versions of the game.</p>
+ */
+public class EnchantmentEvents {
```
i509VCB: Preferably a final class here

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,102 @@
...
+		return TriState.DEFAULT;
+	});
+
+	/**
+	 * Simple helper method to add an enchantment callback to both the enchantment and application
+	 * events.
+	 *
+	 * @param callback The enchantment callback to add to both enchantment events.
+	 */
+	public static void registerAll(AcceptEnchantment callback) {
```
i509VCB: I believe the standards for events here is that Accept and accept application would be two separate functional interfaces.

fabric-enchantment-api-v1/src/main/resources/fabric-enchantment-api-v1.mixins.json
```diff
@@ -0,0 +1,16 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.enchantment",
+  "compatibilityLevel": "JAVA_8",
+  "mixins": [
+    "EnchantmentHelperMixin",
+    "EnchantedBookItemMixin",
+    "AnvilScreenHandlerMixin",
+    // "DisabledAnvilScreenHandlerMixin"
```
i509VCB: Obviously this is disabled until we decide whether it is needed. Just noting here so we don't have an issue at release.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
ghost: I added it to all of the files just because I know there's a history of modules getting merged in and then the burden of updating them falls on someone who has no idea how it works. The idea was to put my contact information in there so that if someone needs information about how the module works, or needs it updated I can take a look at it so that the burden doesn't fall solely on someone else. It can be removed however.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
ghost: How would you propose handling a method such as this one if they were to become two separate interfaces? I was thinking maybe casting a BiFunction, or defining a parent functional interface to cast from.

ghost:
Everything has been changed to match the suggestions, the one thing I didn't change was the author tags in the javadocs for the reason I outlined above. I figure it can't hurt, if it's really a problem I'll just make another commit to remove them. This should be good to pull in now at the discretion of the modding gods above.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
modmuss50: I dont see a big issue with it being there

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,126 @@
...
+	});
+
+	/**
+	 * Simple helper method to add an enchantment callback to both the enchantment and application
+	 * events.
+	 *
+	 * @param callback The enchantment callback to add to both enchantment events.
+	 */
+	public static void registerAll(BiFunction<Enchantment, ItemStack, TriState> callback) {
+		ACCEPT_ENCHANTMENT.register((AcceptEnchantment) callback);
```
i509VCB, line 83: Can you cast like that? Really I would just use a lambda to call the bifunction

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/DisabledAnvilScreenHandlerMixin.java
```diff
@@ -0,0 +1,76 @@
...
+ * @author Vaerian (vaeriann@gmail.com or @Vaerian on GitHub).
+ *
+ * <p>Please contact the author, Vaerian, at the email or GitHub profile listed above
+ * with any questions surrounding implementation choices, functionality, or updating
+ * to newer versions of the game.</p>
+ */
+@Mixin(AnvilScreenHandler.class)
+public class DisabledAnvilScreenHandlerMixin {
+	@Inject(method = "updateResult", at = @At(value = "INVOKE", target = "Lnet/minecraft/enchantment/EnchantmentHelper;set(Ljava/util/Map;Lnet/minecraft/item/ItemStack;)V"), locals = LocalCapture.CAPTURE_FAILHARD)
+	public void updateResult(CallbackInfo callback, ItemStack itemStack, int i, int j, int k, ItemStack itemStack2, ItemStack itemStack3, Map map, int w) {
```
i509VCB, line 48: Would it be possible to name these parameters so we know what is what

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantmentHelperMixin.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+@Mixin(EnchantmentHelper.class)
+public class EnchantmentHelperMixin {
+	@Unique
+	private static Enchantment currentEnchantment;
+
+	// A simple read/write lock to prevent a data race by ensure that the
+	// value captured from the inject is not modified by the time it gets to the redirect mixin.
+	@Unique
+	private static Lock lock = new ReentrantLock();
```
i509VCB, line 60: Not sure if it would just be better to use a `ThreadLocal` here and avoid the weird while loops below

i509VCB (approved):

oliviathevampire:
you need to replace the current registerAll method in EnchantmentEvents with 
```java
public static void registerAll(BiFunction<Enchantment, ItemStack, TriState> callback) {
		ACCEPT_ENCHANTMENT.register(callback::apply);
		ACCEPT_APPLICATION.register(callback::apply);
	}

**Comment by maityyy** [comment]
I really need this API, when is it released? :sweat_smile: The requested changes seem simple.

**Comment by i509VCB** [comment]
I've told modmuss and player to look into this PR. Hopefully they get to reviewing it today or tomorrow.

File: fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java (hunk 10 above)
Comment (Player3324) [reply fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java:83]: It is definitely broken as-is, it'd have to do something like `ACCEPT_ENCHANTMENT.register(callback::apply);ACCEPT_APPLICATION.register(callback::apply);`

`registerAll` is not a good name, it stops making sense when we happen to add another event in this class. The BiFunction won't work with more args.

File: fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java [hunk 13]
```diff
@@ -0,0 +1,126 @@
...
+		ACCEPT_APPLICATION.register((AcceptApplication) callback);
+	}
+
+	public interface AcceptEnchantment {
```
Player3324, line 87: The callback interface names need to have the Callback suffix. It may also be a good idea to directly reference the vanilla context, e.g. something like SelectForEnchantmentTableEvent and SelectForAnvilEvent - there's better names for sure.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,126 @@
...
+		 * and force the item to accept the enchantment. Returning
+		 * {@link net.fabricmc.fabric.api.util.TriState#FALSE} will override the default functionality
+		 * and prevent the item from accepting the enchantment.</p>
+		 *
+		 * @param enchantment The enchantment looking to be applied.
+		 * @param stack The item stack looking to accept the given enchantment.
+		 * @return Whether or not the given enchantment should be applied to the given item stack
+		 * through the use of an enchantment table.
+		 */
+		TriState shouldAccept(Enchantment enchantment, ItemStack stack);
```
Player3324, line 104: It may be worth passing World + nullable BlockPos + nullable Player in these.

Player3324: The interface methods also shadow each other, preventing implementing both differently in the same class.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
ghost: Resolved in dev it'll get pushed with the next commit.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/DisabledAnvilScreenHandlerMixin.java
ghost: Resolved in dev it'll get pushed with the next commit.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantmentHelperMixin.java
ghost: The trade-off would be probably minimally more memory overhead in exchange for some minor performance gains in the case of two threads trying to access the enchantments (which should also typically never occur), but I did change it and we can see if we like that more.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
ghost: That can definitely be done for the anvil (I did do it for the anvil in dev, but it wasn't possible to get the block's position just the player's position through the player object). For the enchantment table that will be significantly more difficult and require actually mixing into the enchantment table and changing how that works (which we could totally do if we want to go down that path, but that would be totally different).

ghost:
Quick poll for those still interested in this PR. Player mentioned the idea of exposing a world, nullable block pos, and a nullable player to each of the enchantment callbacks. While that sounds useful it would require restructuring the mixins in this PR so that the enchantment table related mixins aren't into the EnchantmentHelper class, but instead into the enchantment table block entity itself.

The obvious issue this presents it is means that we would completely bypass the EnchantmentHelper class (which by nature of its implementation works outside of the context of the game state meaning you can query the enchantments for an item without actually having to have any access to a world, block, block position, etc.) and thus mods using the EnchantmentHelper class for logic would have unexpected results with mods that add custom enchantments. Really this comes to how invasive we want to be with this module. I could see it certainly being helpful to expose this additional functionality (and probably useful to many mods) but doing so would be more breaking and invasive than the current implementation. We'd effectively have to create a FabricEnchantmentHelper class and ask that all developers use that instead of the vanilla EnchantmentHelper class (otherwise their mod wouldn't play well with others).

For those still interested in this PR (mainly it seems that would be @sfPlayer1 @modmuss50 @i509VCB @oliviathevampire @Prospector @NeusFear @maityyy (sorry for the pings wasn't really sure how else to make sure this was seen and I'm not totally sure how to proceed)) we're gonna take a little poll so you can let me know what you think sounds best and I'll do that. 👍🏻 for the more invasive version but more versatile PR and 👎🏻 for the current version of the PR without that additional level of exposure.

maityyy:
I think that forcing Fabric mod developers to use an alternative to the `EnchantmentHelper` class is a very bad idea, if only because the Fabric API cannot be fully installed or used by players and mod devs. Also, it seems to me that this is contrary to the Fabric API concept. Fabric API are handy hooks for working with vanilla code, not independent Minecraft Modding API (like Bukkit API). It can also cause conflicts between mods.

i509VCB:
There may be a more appropriate name than `FabricEnchantmentHelper`?

@maityyy We do provide some extensions of vanilla things like a GameRuleVisitor which includes the double and enum rules. As long as the javadoc refers the mod dev to the vanilla method in the case of a reduced context it seems fine to me to include the extra context method.

maityyy:
I think the `registerAll` method should be renamed to something clearer? idk

maityyy:
So what will happen to this PR? I really need such simple events now...

i509VCB:
Pending on Vaerian: Vaerian may have finals so it might be a week or so.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/FabricEnchantment.java
```diff
@@ -0,0 +1,115 @@
...
+	 * says about the search tab.</p>
+	 *
+	 * @param group The item group querying whether or not to contain the
+	 * enchanted books for this enchantment.
+	 * @return A boolean value repressenting whether or not the given
+	 * item group should contain the enchanted books for this enchantment.
+	 *
+	 * @see net.fabricmc.fabric.mixin.enchantment.EnchantedBookItemMixin
+	 */
+	public abstract boolean isAcceptableItemGroup(ItemGroup group);
```
modmuss50, line 114: Minor nit pick, should this have a default? even if its always false?

modmuss50 (approved):
Clean and simple 👍

fabric-enchantment-api-v1/src/testmod/resources/assets/lang/en_us.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fabric-enchantment-api-v1-testmod.dirty": "Dirty"
```
haykam821, line 2: Shouldn't this be:

```suggestion
  "enchantment.fabric-enchantment-api-v1-testmod.dirty": "Dirty"
```

haykam821 (changes requested):

Draylar (approved):

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,126 @@
...
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.api.util.TriState;
+
+/**
+ * Events to intercept the vanilla enchantment logic.
+ *
+ * @author Vaerian (vaeriann@gmail.com or @Vaerian on GitHub).
+ *
+ * <p>Please contact the author, Vaerian, at the email or GitHub profile listed above
```
liach, line 33: Don't think author is visible in generated javadoc unless you turn `-author` flag on. I suggest moving the contacts out of the `@author` block tag.

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,126 @@
...
+
+	/**
+	 * Simple helper method to add an enchantment callback to both the enchantment and application
+	 * events.
+	 *
+	 * @param callback The enchantment callback to add to both enchantment events.
+	 */
+	public static void registerAll(BiFunction<Enchantment, ItemStack, TriState> callback) {
+		ACCEPT_ENCHANTMENT.register((AcceptEnchantment) callback);
+		ACCEPT_APPLICATION.register((AcceptApplication) callback);
```
liach, line 84:
```suggestion
		ACCEPT_ENCHANTMENT.register(callback::apply);
		ACCEPT_APPLICATION.register(callback::apply);
```

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/FabricEnchantment.java
```diff
@@ -0,0 +1,115 @@
...
+ * table can also be applied through the anvil just like vanilla.</p>
+ *
+ * @author Vaerian (vaeriann@gmail.com or @Vaerian on GitHub).
+ *
+ * <p>Please contact the author, Vaerian, at the email or GitHub profile listed above
+ * with any questions surrounding implementation choices, functionality, or updating
+ * to newer versions of the game.</p>
+ */
+public abstract class FabricEnchantment extends Enchantment {
+	protected FabricEnchantment(Rarity weight, EquipmentSlot...slotTypes) {
```
liach, line 45:
```suggestion
	protected FabricEnchantment(Rarity weight, EquipmentSlot... slotTypes) {
```

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/FabricEnchantment.java
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param stack The item stack querying the ability to accept this
+	 * enchantment.
+	 * @return A boolean value representing whether or not the given item
+	 * stack should be able to accept the enchantment from an enchantment
+	 * table.
+	 *
+	 * @see net.minecraft.item.Item#isEnchantable(ItemStack)
+	 * @see net.minecraft.screen.EnchantmentScreenHandler#onContentChanged(Inventory)
+	 * @see net.fabricmc.fabric.mixin.enchantment.EnchantmentHelperMixin
```
liach, line 93: Uh, don't link to mixins. They are good reference for impl detail but it hurts javadoc badly

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/FabricEnchantment.java
```diff
@@ -0,0 +1,115 @@
...
+	 * <p>It should be noted that all enchanted books, for all enchantments are
+	 * automatically added to the search tab regardless of what this method
+	 * says about the search tab.</p>
+	 *
+	 * @param group The item group querying whether or not to contain the
+	 * enchanted books for this enchantment.
+	 * @return A boolean value repressenting whether or not the given
+	 * item group should contain the enchanted books for this enchantment.
+	 *
+	 * @see net.fabricmc.fabric.mixin.enchantment.EnchantedBookItemMixin
```
liach, line 112: same mixin link. just mention in implnote I suggest

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/DisabledAnvilScreenHandlerMixin.java
```diff
@@ -0,0 +1,76 @@
...
+
+					if (oldRepairCost != 0) {
+						oldRepairCost = oldRepairCost / 2;
+					}
+				}
+
+				if (oldRepairCost == 0) {
+					tag.remove("RepairCost");
+				} else {
+					tag.putInt("RepairCost", oldRepairCost);
```
liach, line 68: Where are all these repair cost logic from?

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/mixin/enchantment/EnchantmentHelperMixin.java
liach, line 60: What if an exception is thrown between the lock.lock and lock.unlock calls? This feels inherently unsafe

fabric-enchantment-api-v1/src/main/java/net/fabricmc/fabric/api/enchantment/v1/EnchantmentEvents.java
Technici4n: I'm not fond of the `@author` annotation, and I would like it not to be used, when it comes to Fabric API.

liach: Yeah when something is in fabapi it's usually maintained by contributors alike than one specific person. That's a good point.

ghost:
Uhhh ok I'm back sorry

Binero:
What is this waiting for? Just the conflict?

Prospector:
There are quite a number of unresolved comments on it if you read the latest comments since the last commit

PR #1104, opened 2020-09-26
Add module dependencies to maven POM

kvverti:
This should allow developers to depend on a single module without having to depend on the transitive dependencies of that module as well.


ghost (approved):
Looks good, IMO should be marked as small change as this only really affects the main build.gradle.

modmuss50:
Changes look fine, I do want to test this in a dev env to ensure that nothing explodes.

Player3324:
I think we finally need to move towards making the mod json the authoritative source of build relevant data. The current approach works the opposite way, populating the json from gradle, which breaks source level compatibility as the json is incomplete until built.

The dep version handling needs a closer look either way, maven may not even be able to sufficiently express it such that some scenarios don't break...

kvverti:
I fail to see what that has to do with this. Changing the dependency management system is out of scope for this PR.

Player3324:
This PR is adding additional redundantly specified data, which isn't great longer term..

kvverti:
Additional data? Module dependencies are defined in the `moduleDependencies` call instead of in `dependencies`. If you're talking about duplication between the `build.gradle` and the `fabric.mod.json`, see "out of scope"

i509VCB:
@sfPlayer1 IMHO your concern there about extra redundant data is completely out of scope of fixing this issue as this adds no additional constants and instead changes registration process.

Fixing that stuff is for an entirely different pull request, probably best with the module relocation thing.

PR is fine to merge imo if modmuss believes the poms are good.

Player3324 (approved):

Player3324:
OK you are right. Since it's a build change I'll let modmuss merge this.

i509VCB (approved):

modmuss50 (approved):
Changes look fine, I will test this to make sure nothing weird happens.

PR #1107, opened 2020-10-02
Item explosion handlers

i509VCB:
Also adds Nullable annotations to part of the implementation.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
@@ -18,14 +18,11 @@ package net.fabricmc.fabric.api.item.v1;
 
 import net.minecraft.entity.ItemEntity;
 import net.minecraft.entity.damage.DamageSource;
-import net.minecraft.item.Items;
 
 /**
  * Allows custom logic to dictate whether an item may be destroyed in an explosion.
  *
  * <p>Item explosion handlers may be set with {@link FabricItemSettings#explosionHandler(ItemExplosionHandler)}.
- *
- * <p>Note that a {@link Items#NETHER_STAR} is always immune to explosions and cannot be set to be vulnerable to explosions.
  */
 @FunctionalInterface
 public interface ItemExplosionHandler {
@@ -37,5 +34,5 @@ public interface ItemExplosionHandler {
 	 * @param amount the amount of damage being applied to the item entity
 	 * @return whether the item should not be destroyed in the explosion.
 	 */
-	boolean shouldNotDestroy(ItemEntity entity, DamageSource source, float amount);
+	boolean shouldDestroy(ItemEntity entity, DamageSource source, float amount);
 }
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
@@ -25,7 +25,7 @@ import net.minecraft.item.Items;
  *
  * <p>Item explosion handlers may be set with {@link FabricItemSettings#explosionHandler(ItemExplosionHandler)}.
  *
- * Note that a {@link Items#NETHER_STAR} is always immune to explosions and cannot be set to be vulnerable to explosions.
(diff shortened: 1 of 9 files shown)


i509VCB:
aaaaaaaah why did build break lol

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
```diff
@@ -0,0 +1,41 @@
...
+public interface ItemExplosionHandler {
+	/**
+	 * Checks if an item should not be destroyed in an explosion.
+	 *
+	 * @param entity the item entity, may be used to get the item stack
+	 * @param source the damage source
+	 * @param amount the amount of damage being applied to the item entity
+	 * @return whether the item should not be destroyed in the explosion.
+	 */
+	boolean shouldNotDestroy(ItemEntity entity, DamageSource source, float amount);
```
modmuss50: Would shouldDestroy make more senese here? to me this feels inverted? what do you think?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemEntityMixin.java
```diff
@@ -0,0 +1,58 @@
...
+
+import net.fabricmc.fabric.api.item.v1.ItemExplosionHandler;
+import net.fabricmc.fabric.impl.item.ItemExtensions;
+
+@Mixin(ItemEntity.class)
+abstract class ItemEntityMixin {
+	@Shadow
+	public abstract ItemStack getStack();
+
+	@Redirect(method = "damage", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getItem()Lnet/minecraft/item/Item;"), slice = @Slice(to = @At(value = "FIELD", target = "Lnet/minecraft/item/Items;NETHER_STAR:Lnet/minecraft/item/Item;")))
```
modmuss50: Why the Redirect? Seems much easier to have an early return false inject?

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/ItemExplosionHandlerTest.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.item.v1.FabricItemSettings;
+
+public final class ItemExplosionHandlerTest implements ModInitializer {
+	private static final Item TEST = Registry.register(Registry.ITEM, new Identifier("fabric-item-api-v1-testmod"), new TestItem());
```
modmuss50: Register the item in onInitialize, doesnt make a huge amount of diffrence just a little nicer and more standarised IMO.

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/ItemExplosionHandlerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+	@Override
+	public void onInitialize() {
+	}
+
+	private static final class TestItem extends Item {
```
modmuss50, line 34: No need for this class?

modmuss50 (changes requested):

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/ItemExplosionHandlerTest.java
i509VCB: Cleanliness is the main reason for this class, otherwise you'll have some stupid long oneliner or die when you see me make it 1 line per method call

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemEntityMixin.java
i509VCB: Uses an inject now

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemExplosionHandler.java
i509VCB: Made that change

PR #1108, opened 2020-10-02
Cache the fuel time map in the fuel registry

Technici4n:
Currently, the fuel time map is rebuilt every time it is queried, which is taking 65% of the tick time on one of the public AOF3 servers.

This PR
1) Caches the map, rebuilding it only when necessary.
2) Changes the furnace implementation to use the `FuelRegistry` so that the vanilla furnaces benefit from the caching as well.

I tested this on a server with ezpas (EZ Pipes and Stuff), which does a lot of `isValid` calls every tick.
* Without the fix: https://spark.lucko.me/#I8zDOn3zKB
* With the fix using the `createFuelTimeMap` directly: https://spark.lucko.me/#YtC8VuWArQ
* With the fix using a map from fastutil: https://spark.lucko.me/#a8C6Tmj1yK


shartte (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FuelRegistryImpl.java
```diff
@@ -36,12 +38,26 @@
 	private static final Logger LOGGER = LogManager.getLogger();
 	private final Object2IntMap<ItemConvertible> itemCookTimes = new Object2IntLinkedOpenHashMap<>();
 	private final Object2IntMap<Tag<Item>> tagCookTimes = new Object2IntLinkedOpenHashMap<>();
+	private Object2IntMap<Item> fuelTimeMap;
+	private boolean fuelTimeMapNeedsUpdate = true;
 
-	public FuelRegistryImpl() { }
+	public FuelRegistryImpl() {
+		ServerLifecycleEvents.END_DATA_PACK_RELOAD.register((server, serverResourceManager, success) -> {
```
shartte, line 44: Have you tested if this event is emitted on a client when it receives the tags on game-join from the server?
@i509VCB Since you are the event-master, do you know this off the top of your head?

Technici4n: It doesn't, but 824377e addresses this issue!

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
```diff
@@ -34,4 +36,19 @@
...
 		FuelRegistryImpl.INSTANCE.apply(info.getReturnValue());
 	}
+
+	@Inject(at = @At("HEAD"), method = "canUseAsFuel", cancellable = true)
+	private static void canUseAsFuelHook(ItemStack stack, CallbackInfoReturnable<Boolean> info) {
+		info.setReturnValue(FuelRegistry.INSTANCE.get(stack.getItem()) != null);
+	}
+
+	@Inject(at = @At("HEAD"), method = "getFuelTime", cancellable = true)
+	protected void getFuelTimeHook(ItemStack fuel, CallbackInfoReturnable<Integer> info) {
```
shedaniel: Can't all of these turn into 1 single mixin at AbstractFurnaceBlockEntity#createFuelTimeMap returning the cached map?

Technici4n: It's a possibility. I'm not fond of it because it "leaks" an implementation detail (i.e. the cached map).

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinTagManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+
+import net.minecraft.tag.TagManager;
+
+import net.fabricmc.fabric.impl.content.registry.FuelRegistryImpl;
+
+@Mixin(targets = "net.minecraft.tag.TagManager$1")
+public abstract class MixinTagManager implements TagManager {
+	public void apply() {
```
i509VCB: This is effectively an overwrite, should be an inject

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
```diff
@@ -34,4 +36,19 @@
 	private static void fuelTimeMapHook(CallbackInfoReturnable<Map<Item, Integer>> info) {
 		FuelRegistryImpl.INSTANCE.apply(info.getReturnValue());
 	}
+
+	@Inject(at = @At("HEAD"), method = "canUseAsFuel", cancellable = true)
+	private static void canUseAsFuelHook(ItemStack stack, CallbackInfoReturnable<Boolean> info) {
```
i509VCB: Both new injects are soft overwrites, any other mod injecting here will never be handled.

i509VCB (changes requested):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinTagManager.java
Technici4n: I don't think it's possible to inject here because the `apply` method is a `default` from an interface. I tried injecting there first, but injection into default interface methods is not supported yet AFAIK.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
Technici4n: Would it be better to inject just before `createFuelTimeMap` ? That should work too.

Technici4n: But if I were to do that, any mod injecting at TAIL for example still wouldn't work I think. Is it really reasonable to expect mods to be injecting stuff in these two functions?

Technici4n: If you want I can lower the priority on the mixin itself... that works too, doesn't it?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
i509VCB: We would have to return an immutable map in this case which may change the semantics of anyone using the furnace. As such I'd probably just add getters in the fuel registry

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinTagManager.java
i509VCB: You are targeting an anonymous class? It should be possible to inject here I think. Of course I may be wrong

Technici4n: So I injected into `ClientNetworkPlayHandler` and used the `END_DATA_PACK_RELOAD` event. That removes the need for the overwrite.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
Technici4n: Resolved by d0ce38b.

shedaniel (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
```diff
@@ -34,4 +35,26 @@
...
 		FuelRegistryImpl.INSTANCE.apply(info.getReturnValue());
 	}
+
+	@Redirect(
+			method = "canUseAsFuel",
+			at = @At(
+				value = "INVOKE",
+				target = "Lnet/minecraft/block/entity/AbstractFurnaceBlockEntity;createFuelTimeMap()Ljava/util/Map;"
+			)
+	)
```
i509VCB: I personally dislike it but modmuss wants the mixin annotations to be one line

i509VCB (approved):
Codewise everything looks good

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/MixinAbstractFurnaceBlockEntity.java
Technici4n: Aight

i509VCB:
I will do some testing tomorrow

modmuss50 (approved):

oliviathevampire (approved):

YanisBft (approved):

LambdAurora (approved):

i509VCB:
@Technici4n can you fix the merge conflicts?

AMereBagatelle (approved):

Mysterious-Dev (approved):

Player3324 (approved):
Looks good after a few minor edits, check the commits for details please. This needs another test run before merging.

modmuss50 (approved):

modmuss50:
No new API so this can go right in.

PR #1110, opened 2020-10-05
Fix items with sword tool tag not instantly breaking bamboo

Virtuoel:
Bamboo is normally instamined by anything that's `instanceof SwordItem`. 
This change lets items in the `fabric:swords` tag do so as well.

(Not sure if these should be two separate mixins or one mixin targeting both classes since the injects are identical.)


fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/mixin/tool/attribute/BambooBlockMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.world.BlockView;
+
+import net.fabricmc.fabric.api.tool.attribute.v1.FabricToolTags;
+
+@Mixin(BambooBlock.class)
+public abstract class BambooBlockMixin {
+	@Inject(at = @At("HEAD"), method = "calcBlockBreakingDelta", cancellable = true)
+	private void onCalcBlockBreakingDelta(BlockState state, PlayerEntity player, BlockView world, BlockPos pos, CallbackInfoReturnable<Float> info) {
+		if (FabricToolTags.SWORDS.contains(player.getMainHandStack().getItem())) {
+			info.setReturnValue(1.0F);
```
Technici4n, line 43: I'd suggest using an `instanceof` `@Redirect` here instead of this `@Inject` that effectively overwrites the method.

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/mixin/tool/attribute/BambooSaplingBlockMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.world.BlockView;
+
+import net.fabricmc.fabric.api.tool.attribute.v1.FabricToolTags;
+
+@Mixin(BambooSaplingBlock.class)
+public abstract class BambooSaplingBlockMixin {
+	@Inject(at = @At("HEAD"), method = "calcBlockBreakingDelta", cancellable = true)
+	private void onCalcBlockBreakingDelta(BlockState state, PlayerEntity player, BlockView world, BlockPos pos, CallbackInfoReturnable<Float> info) {
+		if (FabricToolTags.SWORDS.contains(player.getMainHandStack().getItem())) {
+			info.setReturnValue(1.0F);
```
Technici4n: Same comment here.

Technici4n (changes requested):

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/mixin/tool/attribute/BambooBlockMixin.java
Virtuoel: If it was an unconditional cancel I'd agree that it's effectively an overwrite, but conditional cancels seem to be used all over the place in FabricAPI already?

Technici4n: As you wish, I'm just a random reviewer. It feels cleaner to me because all you really want to do is replace that `instanceof`, and that is exactly what `@Redirect` does.

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/mixin/tool/attribute/BambooBlockMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockView;
+
+import net.fabricmc.fabric.api.tool.attribute.v1.FabricToolTags;
+
+@Mixin(BambooBlock.class)
+public abstract class BambooBlockMixin {
+	@Inject(at = @At("HEAD"), method = "calcBlockBreakingDelta", cancellable = true)
+	private void onCalcBlockBreakingDelta(BlockState state, PlayerEntity player, BlockView world, BlockPos pos, CallbackInfoReturnable<Float> info) {
+		if (FabricToolTags.SWORDS.contains(player.getMainHandStack().getItem())) {
```
i509VCB, line 42: Please add a description of why the injects exist here just so future maintainers know the purpose for the mixin existing.

OroArmor:
> (Not sure if these should be two separate mixins or one mixin targeting both classes since the injects are identical.)

I think you could make them one mixin by changing the mixin annotation to
```@Mixin({BambooBlock.class, BambooSaplingBlock.class})```

Consolidating code is always better and makes it easier to change in the future. (Because people forget to check everything :) )

Virtuoel:
Mixins consolidated and Javadoc added.

Technici4n (approved):

shedaniel (approved):

i509VCB (approved):

YTG1234 (approved):

modmuss50 (approved):

oliviathevampire (approved):

Player3324 (approved):
LGTM, thanks!

PR #1112, opened 2020-10-06
Custom tool levels

shedaniel:
Close #1073 if applied first.

### Summary
Provide & recommend the use of `ToolLevel` instead of `int` in `breakByTool`. Allows in-between mining levels that are essentially decimal mining levels.

### Other changes
`MixinToolItem` has been removed, it is a broken mixin that was removed from the mixin json temporarily, and not needed after tool-attribute 1.1

### Non-Goals
It is not a goal to offer users a way to edit vanilla blocks' mining levels **in bulk** since those are hardcoded in their respective items.
However, for modders, mining levels for blocks individually (including vanilla blocks) can now be edited via `ToolManager#setBlockBreakLevel`.

### Material Level Voting
Voting can be done via `ToolLevel#by`, defining an `Identifier`, and creating a json file at `data/namespace/fabric_tool_levels/path.json`:
```json5
// Please don't be so dumb and copy this file directly, comments are not supported in real use.
{
  "replace": false, // Optional value
  "value": 1.5 // Decimal number defining the mining level. 1.5 is between 1 (stone) and 2 (iron)
}
```
The tool levels are voted by the data packs, and if there is more than 1 winner, the ones applied later shall prevail.
#### Vanilla tools
Mods can implement `FabricToolMaterial` instead of `ToolMaterial` to get the enhanced hook of `getToolLevel` returning `ToolLevel`.
Here's an example:
```java
private static final ToolMaterial TIN_MATERIAL = new FabricToolMaterial() {
	@Override
	public ToolLevel getToolLevel() {
		return ToolLevel.by(new Identifier("c", "tin"));
	}

	@Override
	public int getDurability() {
		return 500;
	}

	@Override
	public float getMiningSpeedMultiplier() {
		return 4;
	}

	@Override
	public float getAttackDamage() {
		return 1;
	}

	@Override
	public int getEnchantability() {
		return 30;
	}

	@Override
	public Ingredient getRepairIngredient() {
		return Ingredient.EMPTY;
	}
};
```
The `FabricToolMaterial` could be used as vanilla tool materials.
```java
new PickaxeItem(TIN_MATERIAL, 1, -2.8F, new Item.Settings())
```
#### Dynamic Tools
Here's an example:
```java
public class DynamicToolItem extends Item implements DynamicAttributeTool {
	private final Tag<Item> toolType;
	private final ToolLevel miningLevel;
	private final float toolSpeed;

	// The durability of the item can be applied via Settings#maxDamage
	private DynamicToolItem(Settings settings, Tag<Item> toolType, ToolLevel miningLevel, float toolSpeed) {
		super(settings);
		this.toolType = toolType;
		this.miningLevel = miningLevel;
		this.toolSpeed = toolSpeed;
	}

	@Override
	public float getToolMiningLevel(Tag<Item> tag, BlockState state, ItemStack stack, LivingEntity user) {
		if (tag.equals(toolType)) {
			return this.miningLevel.getLevel();
		}

		return ToolLevel.NONE.getLevel();
	}

	@Override
	public float getMiningSpeedMultiplier(Tag<Item> tag, BlockState state, ItemStack stack, LivingEntity user) {
		if (tag.equals(toolType)) {
			return toolSpeed;
		}

		return 1.0F;
	}
}
```


fabric-tool-attribute-api-v1/build.gradle
```diff
@@ -1,9 +1,10 @@
...
 
 dependencies {
-    compile project(path: ':fabric-api-base', configuration: 'dev')
-    compile project(path: ':fabric-tag-extensions-v0', configuration: 'dev')
-    testmodCompile project(path: ':fabric-object-builder-api-v1', configuration: 'dev')
-	testmodCompile project(path: ':fabric-events-lifecycle-v0', configuration: 'dev')
+	compile project(path: ':fabric-api-base', configuration: 'dev')
+	compile project(path: ':fabric-resource-loader-v0', configuration: 'dev')
+	compile project(path: ':fabric-tag-extensions-v0', configuration: 'dev')
+	testmodCompile project(path: ':fabric-object-builder-api-v1', configuration: 'dev')
```
i509VCB, line 5: Tool attributes depends on object builder and object builder depends on tool attributes.

Isn't this effectively a recursive dependency?

shedaniel: Well I am not sure, and I didn't actually add this line, all I did is reformat this file, also this PR bases on #1073 and this line was changed there as well

i509VCB:
Could you fix up the conflicts? I plan to review this tomorrow.

shedaniel:
Please do ping me in discord next time, I don't come across my prs usually

Prospector:
The one thing I would comment on here is the path of the tool level json files. Should they perhaps be in `data/namespace/fabric/tool_levels` to prevent any conflicts?

shedaniel:
I would do `data/namespace/fabric_tool_levels` instead tho, the slash would confuse people

fabric-tool-attribute-api-v1/build.gradle
```diff
@@ -8,5 +8,6 @@ dependencies {
 
 moduleDependencies(project, [
 		'fabric-api-base',
+		'fabric-resource-loader-v0'
```
OroArmor, line 11: Missing `,`

OroArmor (approved):

Technici4n:
Probably not gonna happen, I don't think it's in scope for fabric api. In any case, this would require a full reimplementation due to the various mining level changes.

PR #1115, opened 2020-10-10
Client-sided commands

Juuxel:
Extends fabric-command-api-v1 with support for fully client-side commands. Mirrored in a [FabLabs project](https://github.com/FabLabsMC/ClientCommands).

The commands are only executed on the client. If a valid client command is sent in chat, it won't reach the server and it will be executed. If the user sends something beginning with a command prefix but isn't a command, it will be sent as usual.

The API also adds a `/fcc help` client command that works just like `/help` if there are any other client commands present.


fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/ArgumentBuilders.java
```diff
@@ -0,0 +1,55 @@
...
+import com.mojang.brigadier.builder.RequiredArgumentBuilder;
+
+/**
+ * Provides helper methods for constructing argument builders
+ * for client-sided commands.
+ *
+ * <p>This is an alternative to the server-side helper methods in
+ * {@link net.minecraft.server.command.CommandManager}.
+ */
+public final class ArgumentBuilders {
```
i509VCB: The class name should specify it is for client arguments.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/ClientCommandRegistrationCallback.java
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Gets the command registration event with a custom command prefix.
+	 *
+	 * @param prefix the command prefix
+	 * @return the event object
+	 * @throws IllegalArgumentException if the prefix {@linkplain Character#isLetterOrDigit(char) is a letter or a digit},
+	 *                                  or if it {@linkplain Character#isWhitespace(char) is whitespace}.
+	 */
+	static Event<ClientCommandRegistrationCallback> event(char prefix) {
```
modmuss50: Not a big fan of allowing prefixes like this? Id keep this as close as possible to the normal `CommandRegistrationCallback`

fabric-command-api-v1/src/testmod/java/net/fabricmc/fabric/test/command/client/ClientCommandTest.java
```diff
@@ -0,0 +1,58 @@
...
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.command.v1.ArgumentBuilders;
+import net.fabricmc.fabric.api.client.command.v1.ClientCommandRegistrationCallback;
+
+public final class ClientCommandTest implements ClientModInitializer {
+	private static final DynamicCommandExceptionType IS_NULL = new DynamicCommandExceptionType(x -> new LiteralText("The " + x + " is null"));
+
+	@Override
+	public void onInitializeClient() {
+		ClientCommandRegistrationCallback.event().register(registerSimpleCommands("This is a client command"));
```
modmuss50: Might be nice to have some better tests, take a look at the normal command tests to see how they fail hard even if not ran. Has saved a bug once before. 👍

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/ClientCommandRegistrationCallback.java
Juuxel: Ah, alright. That lets me simplify the code a lot, then.

haykam821:
Will completions be derived from the client exclusively?

Juuxel:
@haykam821 For client commands, yes. (This doesn't touch server commands at all)

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/ArgumentBuilders.java
Juuxel: Renamed to `ClientArgumentBuilders`.

LambdAurora (approved):
Love it, looking forward to using it.

FlashyReese (approved):

modmuss50 (approved):
Looks good to me.

liach (approved):

i509VCB (approved):

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
```diff
@@ -0,0 +1,179 @@
...
+import net.minecraft.text.TranslatableText;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.fabric.api.client.command.v1.FabricClientCommandSource;
+
+@Environment(EnvType.CLIENT)
+public final class ClientCommandInternals {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final char PREFIX = '/';
```
OroArmor, line 59: I think there was already a discussion on this in the discord if these commands should have a different/customizable prefix. If not, it would be nice to have customizable prefixes and if the API is going through the trouble of adding client commands, we might as well have commands be able to have custom prefixes. (obviously default to `/`).

OroArmor (changes requested):

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
OroArmor: Ah, it seems that modmuss50 already gave some input on this, but I think that having two registration methods, one for a default prefix and one for a custom one is better. I feel like the main reason to have a different prefix is to think about how it has a different function than `/`, If we don't include this, it will exclude the modders that want to include these features, and force them to rewrite what we have already done. I feel that this is counterproductive to the purpose of the Fabric API where common needs are made accessible to all modders and do not require mixins.

Juuxel: Well, I had custom prefixes working quite nicely originally. I removed them because modmuss didn't want them in Fabric API (to keep the client commands closer to vanilla's server commands).

LambdAurora: Which seems a bit weird, because it doesn't seem to handle conflict very well.

Like if a mod register a command `a` and the server also has a command `a` what happens?

Custom prefixes could partially fix this issue, but at the same time some custom prefixes might be weird/annoying.

Client commands having the same prefix as everything else doesn't seem very user-friendly in the end.

Juuxel: > Like if a mod register a command `a` and the server also has a command `a` what happens?

I actually haven't tested this, but with the current impl the client command should override the server command. It might be more sensible to reverse that behaviour.

@modmuss50 Should we reconsider custom prefixes for compatibility with servers? I'm not sure whether there should use only `/`, only some client-specific character (say, `#`) or to have the full prefix support again.

LambdAurora: I don't know if it's possible, but for the conflict situation I would like to see the Bukkit behavior which is to add `namespace:` at the start of the command.

liach: `namespace:command` is fairly pointless when you can do `namespace command`

liach:
@Juuxel I wonder if it's valuable to provide a config option (loaded every time a world is loaded) to disable client commands entirely in case certain client commands interfere with custom server commands?

Juuxel:
@liach That's a nice idea, might add that. I'll talk about that with Player as well to prevent too much bikeshedding before writing the code :smile:

LambdAurora (approved):

Player3324 (approved):
Good, thanks! I tried this with Sampler, works as expected.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
Player3324: A custom prefix is out of scope for the initial PR at least, IMO the best solution is to eventually:
- give precedence to server side commands
- use `/fcc x` to always run the client command `/x`

This will resolve the ambiguity cleanly, as long as nothing else claims `/fcc`, which me might exclude from the server precedence.

Prospector: I feel like `/fcc` should just be a shortcut to a full `/fabricclientcommand` or whatever, much like `/tp` is a short alias for `/teleport`. It lets people find it if they don't know the acronym, and if they do then they can type it quickly.

haykam821:
I'm not a fan of `/fcc help` personally; it feels a bit too much out of scope to me.

Player3324:
The idea behind `/fcc help` was to close the discoverability gap. It is something that has to be provided by us and works towards the future plans at https://github.com/FabricMC/fabric/pull/1115#discussion_r575887531

liach:
> 
> 
> I'm not a fan of `/fcc help` personally; it feels a bit too much out of scope to me.

Agreed. Next time servers can prevent fabric clients from joining by requiring the `/fcc` command to verify :laughing: hence I suggest a config option to disable client commands if mods depend on it but packs don't want to enable it.

i509VCB:
> Next time servers can prevent fabric clients from joining by requiring the /fcc command to verify

Oh great security theater servers. Literally 3 seconds of using your brain shows that the client brand, registered channels or command precense to detect ("modified clients") is royally stupid.

liach:
> 
> 
> > Next time servers can prevent fabric clients from joining by requiring the /fcc command to verify
> 
> Oh great security theater servers. Literally 3 seconds of using your brain shows that the client brand, registered channels or command precense to detect ("modified clients") is royally stupid.

The point is not server hating fabric. The point is that since this mod adds `fcc`, it may cause command compatibility problems and had better an option to toggle in config (or in config, an option to disable client commands totally)

i509VCB:
We could add an alias then like a `fabric-command-api-v1:client` as a last fallback as anyone but intentionally would set that.

Player3324:
Could always add another level of workaround like `/fcc servercmd fcc really...` that works as `/fcc really...` would, which erases any conflict potential.

liach:
So no plan on allowing turning off client commands completely via config? Just curious.

i509VCB:
I'd atleast think add /fcc if there is no server command, there is then fallback to the always registered namespaced command.

Player3324:
We don't really have a proper config yet

liach:
Sure, this is a minor problem after all. Not worth blocking this giant feature for that tiny bikeshed

Juuxel:
> I'd atleast think add /fcc if there is no server command, there is then fallback to the always registered namespaced command.

@i509VCB That's hard to do with the current architecture of the API: the client registrations are global, but the server commands are local to each command tree packet received. This would basically need local dispatchers instead of a global one (and then we'd also have `ClientCommandRegistrationCallback` probably).

As for the longer alias, I think I'll add it unless @sfPlayer1 has any objections. (The `/fcc servercmd` and `/fcc <some client command>` things can come as later improvements.)

Juuxel:
This is probably the final change to this PR: I added `/fabric-command-api-v1:client` as an alias for `/fcc` (as suggested by @i509VCB).

RedstoneParadox (approved):

BeetMacol (approved):

FlashyReese (approved):

PR #1120, opened 2020-10-12
FabircItemSettings Attack Sound Events

biom4st3r:


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemMixin.java
```diff
@@ -61,4 +73,76 @@ public CustomDamageHandler fabric_getCustomDamageHandler() {
 	public void fabric_setCustomDamageHandler(CustomDamageHandler handler) {
 		this.customDamageHandler = handler;
 	}
+
+	@Override
+	public @Nullable SoundPlayer fabric_getCriticalHitSound() {
+		// TODO Auto-generated method stub
```
modmuss50: Remove all these todo's

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -32,6 +32,42 @@
...
+	public enum AttackSound {
+		STRONG,
+		WEAK,
+		CRITIAL,
+		KNOCKBACK,
+		NO_DAMAGE,
+		SWEEPING,
+	}
+
+	public FabricItemSettings soundEvent(AttackSound sound,SoundPlayer soundEvent) {
```
modmuss50: Missing a space, I dont think this will pass checkstyle

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/PlayerEntityMixin.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+    @Inject(
+        at = @At(
+            value="FIELD",
+            target="net/minecraft/sound/SoundEvents.ENTITY_PLAYER_ATTACK_NODAMAGE:Lnet/minecraft/sound/SoundEvent;",
+            opcode = Opcodes.GETSTATIC),
+        method = "attack",
+        cancellable = false)
+    public void onPlayNoDamageSound(CallbackInfo ci)
+    {
```
modmuss50: again, this wont pass checkstyle

modmuss50 (changes requested):
Please ensure this passes checkstyle:

Run: `gradlew check` (or even build)

Most IDE's should have a plugin to aid with this.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/PlayerEntityMixin.java
biom4st3r: Yeah i forgot about the checkstyle and i had the plugin off. I'm working on them now

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/PlayerEntityMixin.java
```diff
@@ -0,0 +1,117 @@
...
+	public float muteSoundEvent(float volume) {
+		if (muteSound) {
+			muteSound = false;
+			return 0.0F;
+		}
+
+		return volume;
+	}
+
+	@Inject(at = @At(value = "FIELD", target = "net/minecraft/sound/SoundEvents.ENTITY_PLAYER_ATTACK_KNOCKBACK:Lnet/minecraft/sound/SoundEvent;", opcode = Opcodes.GETSTATIC), method = "attack", cancellable = false)
```
i509VCB, line 52: Some javadoc here should explain that the sound being played will be muted if the item has it's own sound.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/PlayerEntityMixin.java
```diff
@@ -0,0 +1,117 @@
...
+	public void onPlayKnockbackSound(CallbackInfo ci) {
+		PlayerEntity pe = (PlayerEntity) (Object) this;
+		ItemExtensions item = ((ItemExtensions) pe.getMainHandStack().getItem());
```
i509VCB, line 55: Items aren't always used in the main hand, is there a way to get the currently active item and use that for dispatch?

i509VCB, line 55: You should hardwire the logic to not do anything if the active item stack is empty.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -32,6 +32,42 @@
  * {@code new FabricItemSettings()}.
  */
 public class FabricItemSettings extends Item.Settings {
+	public enum AttackSound {
```
i509VCB, line 35: I would call this `AttackSoundType`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -32,6 +32,42 @@
...
+		STRONG,
+		WEAK,
+		CRITIAL,
+		KNOCKBACK,
+		NO_DAMAGE,
+		SWEEPING,
+	}
+
+	public FabricItemSettings soundEvent(AttackSound sound, SoundPlayer soundEvent) {
```
i509VCB, line 44: Null check the parameters.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -32,6 +32,42 @@
...
+			FabricItemInternals.computeExtraData(this).strongHitSound(soundEvent);
+			break;
+		case SWEEPING:
+			FabricItemInternals.computeExtraData(this).sweepingHitSound(soundEvent);
+			break;
+		case WEAK:
+			FabricItemInternals.computeExtraData(this).weakHitSound(soundEvent);
+			break;
+		default:
+			break;
```
i509VCB, line 65: What should we do in the `default` case?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/SoundPlayer.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.world.World;
+
+@FunctionalInterface
+public interface SoundPlayer {
```
i509VCB, line 23: I do not like the name of this interface. To me it sounds like some sound related interface implemented onto the player.

To me it seems like a sort of preconfigured `SoundEffect` that you just pass the world and player listening to the sound into.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
i509VCB, line 44: Javadoc please.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
i509VCB, line 35: Javadoc please.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/SoundPlayer.java
```diff
@@ -0,0 +1,25 @@
...
+	void playSound(World world, PlayerEntity player);
```
i509VCB, line 24: Where is this called? Is this called on the logical client, logical server or both?

i509VCB (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
biom4st3r: I was considering just making it SoundType so i could add BURP and some others, because they are in the same position

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
biom4st3r: Is it bad form to just remove it?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/SoundPlayer.java
biom4st3r: It's called on both

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
i509VCB: Well removing the break case would essentially be a "do nothing". If the compiler requires it then it would be empty if we don't want to fail.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
biom4st3r: STRONG,WEAK,CRITICAL,KNOCKBACK,NO_DAMAGE,SWEEPING,SHOOT_BOW,BURP,SHIELD_BREAK,ARMOR_EQUIP,ITEM_BREAK,SHIELD_BLOCK
Those are the ones i could find

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/PlayerEntityMixin.java
biom4st3r: You can only only activate the attack related ones with main hand, and active item doesn't get set to the attacking weapon. some of the other Types activated by right click(use) i can probably use getActiveItem instead

biom4st3r: Should changing the default sounds(with an empty hand) be reserved to just resource packs? Genuine question, i'm not sure

i509VCB: I don't think you should be modifying air.

i509VCB: If that is the case then note in javadoc that the attack sounds only apply in main hand.

biom4st3r: Can do!

ghost:
I think in this situation we might be apt to ask if this PR is even necessary. To me it just seems like it's a really easy and simple mixin that's been repeated for every noise case. I don't know how many people are actually going to use this and so it might make for sense to just have the burden of implementing this fall on the developer of a mod instead of in fabric api (especially considering how easy the mixin seems).

biom4st3r:
I was under the impression that ModifyArgs are not compatible. Can you use multiple ModifyArgs on 1 spot?

PR #1122, opened 2020-10-15
Fix QuadEmitter#square and Indigo uv locking

### LINKED ISSUE #1024: QuadEmitter#sprite default implementation is broken
Here is my code:

https://github.com/Technici4n/fabric-rendering-issue/blob/master/src/main/java/net/fabricmc/example/FourSidedFurnaceModel.java

Result:
image

Expected result:
image

Technici4n:
Fixes #1024. Also, there was a mistake in the uv locking for the textures, which I fixed too.


grondag:
Looks like an appropriate fix.  Could you provide some before/after screencaps?

Technici4n:
Good idea indeed!

Before:
image

After:
image

- The wrong placement of the quads on the sides is fixed.
- The wrong UV for the top quad is fixed too.

PS: I've made sure that the new UVs match vanilla's!

modmuss50:
Is this a breaking change?

Technici4n:
Yes if people relied on the old behavior, but it's so broken I don't think anyone would have.

modmuss50 (approved):

PR #1124, opened 2020-10-17
Explore publishing releases using github actions

modmuss50:
This PR makes the required changes to use github actions to build and release fabric api to maven, curse and github. Currenly Fabric API is built by a jenkins that I have been running for a number of years. Builds are manually triggered after a version bump, this apporach has been kept.

There should be no user or modder facing impact to these changes. The only change they will notice is the build number will reset back to 1. This should not be an issue as it is only version metadata.

My longer term plan is move all builds over to github actions as its easier to manage, and has a number of benefits over self hosting a jenkins.

* Also updates gradle


haykam821:
> There should be no user or modder facing impact to these changes. The only change they will notice is the build number will reset back to 1.

Is there no workaround for this?

> This should not be an issue as it is only version metadata.

I think that resetting the numerical progression to zero might confuse Fabric API users, since they would typically pay more attention to that than the more complicated semver string.

modmuss50:
That build number is only there as a backup, it would be possible to remove it, and add a check in place to ensure we dont release the same version twice.

Failing that it we could just add 411 or what ever it would be at the time of merging, I do see your point. I believe right now the build numbers are a bit of a mess as they are per branch.

Player3324 (approved):
Nothing obviously wrong

modmuss50:
I have updated this to remove the build number and added a task to ensure that duplicate version cannot be released.

PR #678, opened 2020-06-11
Content registry v1

SollyW:
Fixes some issues with, and includes, #576, which the author no longer wants to maintain.

PR #576, opened 2020-04-15
fabric-content-registries-v1 - A refactor, now with Tilling, Stripping, and Flattening registries

Prospector:
The old fabric-content-registries-v0 module has now been deprecated in its entirety. In its place is fabric-content-registries-v1, with the old APIs rewritten (pretty much) from the ground up.

### Old features rewritten:
* FuelRegistry -> FuelItemRegistry
* CompostingChanceRegistry -> CompostableItemRegistry
* FlammableBlockRegistry -> FlammableBlockRegistry
* LootTableTypeRegistry -> Was deprecated & moved to another module ages ago

### New features:
* TillableBlockRegistry
* StrippableBlockRegistry
* FlattenableBlockRegistry

**Note:** The old .clear methods from Block/Item2ObjectMaps are now gone. I couldn't figure out a good use for them and they were a pain to implement across all registries. I spoke with asie about why they were added in the first place and he said possibly with tweaker mods in mind, but he wasn't sure. Either way, the functionality of restoring vanilla values seems like out of scope for Fabric API to me.

**This PR passes checkLicense and checkstyle & conforms to the new module naming scheme (#575)**

PR #1125, opened 2020-10-18
Content registry v1 Attempt 3

TheGlitch76:
This replaces #678, which replaces #576.

Commits had to be manually cherry-picked because of some missed commits on API not allowing an easy merge. The changed committer shouldn't be an issue since FAPI uses squash merges.

For clarity, the outstanding comments on the previous PR were:
- Trying to remove the `List<Runnable>`
- More documentation
- ~~Getting the prefix from the registry in FlammableBlockRegistry~~ (resolved)


fabric-content-registries-v1/src/main/java/net/fabricmc/fabric/api/content/registry/v1/util/ItemContentRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemConvertible;
+
+/**
+ * A registry that maps items to values.
+ */
+public interface ItemContentRegistry<V> extends ContentRegistry<Item, V> {
+	default V get(ItemConvertible key) {
+		return this.get(key.asItem());
```
i509VCB: Null check the `asItem` calls

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/util/Block2ObjectMap.java
```diff
@@ -19,6 +19,10 @@
 import net.minecraft.block.Block;
 import net.minecraft.tag.Tag;
 
+/**
+ * @deprecated replaced by {@link net.fabricmc.fabric.api.content.registry.v1.util.Taggable2ObjectMap}
```
i509VCB: Linked replacement class no longer exists. Javadoc should reflect that.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/util/Item2ObjectMap.java
```diff
@@ -20,6 +20,10 @@
 import net.minecraft.item.ItemConvertible;
 import net.minecraft.tag.Tag;
 
+/**
+ * @deprecated replaced by {@link net.fabricmc.fabric.api.content.registry.v1.util.Item2ObjectMap}
```
i509VCB: Linked replacement class no longer exists. Javadoc should reflect that.

fabric-content-registries-v1/src/main/java/net/fabricmc/fabric/api/content/registry/v1/FlammableBlockRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Registry of Blocks that can be set on fire.
+ */
+public interface FlammableBlockRegistry extends ContentRegistry<Block, FlammableBlockRegistry.Entry> {
+	/**
+	 * Create a new flammable block registry for a type of fire.
+	 * @param fireBlock the block that this registry belongs to
+	 * @param prefix the name of the block
+	 */
+	default FlammableBlockRegistry create(Block fireBlock, String prefix) {
```
i509VCB: Javadoc should document thrown exception if the block is not a fire block.

TheGlitch76: Wouldn't it be better to just only accept a `FireBlock`? It seems that Mojang is moving away from registering these sorts of things in code, so it's doubtful that the content registry for flammable blocks will ever need to expand.

fabric-content-registries-v1/src/main/java/net/fabricmc/fabric/api/content/registry/v1/util/ItemContentRegistry.java
TheGlitch76: as in, null check the result of asItem?

i509VCB: yes make sure `asItem` does not return null

fabric-content-registries-v1/src/main/java/net/fabricmc/fabric/api/content/registry/v1/FlammableBlockRegistry.java
i509VCB: Many mods expose their block instances as `Block` such as vanilla. So you would need a cast in some cases, but it would also reduce the chance of error.

TheGlitch76: I'm going to go for the documented throws approach

i509VCB (approved):

fabric-content-registries-v1/src/main/java/net/fabricmc/fabric/impl/content/registry/ContentRegistryImpl.java
```diff
@@ -0,0 +1,138 @@
...
+import net.fabricmc.fabric.api.resource.ResourceManagerHelper;
+import net.fabricmc.fabric.api.resource.ResourceReloadListenerKeys;
+import net.fabricmc.fabric.api.resource.SimpleSynchronousResourceReloadListener;
+
+public abstract class ContentRegistryImpl<K, V> implements ContentRegistry<K, V>, SimpleSynchronousResourceReloadListener {
+	private final Identifier reloadIdentifier;
+	private static final Collection<Identifier> RELOAD_DEPS = Collections.singletonList(ResourceReloadListenerKeys.TAGS);
+
+	private boolean tagsPresent = false;
+	private final List<Runnable> processor = new LinkedList<>(); // This is to preserve the order of how things are added and removed. This would be a lot simpler if we did not have to deal with Tags
```
liach, line 41: If the runnables are too ugly, maybe store them as pojos under a common superinterface so you might optimize them a bit than just calling runnables

i509VCB:
@TheGlitch76 conflicts need resolving and the optimization from Tech's pr should be ported over.

fabric-content-registries-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "depends": {
+    "fabricloader": ">=0.4.0",
+    "fabric-api-base": "*",
+    "fabric-resource-loader-v0": "*"
+  },
+  "description": "Adds registries for vanilla mechanics that are missing them.",
+  "mixins": [
+    "fabric-content-registries-v1.mixins.json"
+  ]
```
i509VCB, line 26: This module now needs to define a module lifecycle:
```json
  "custom": {
  }
```

i509VCB:
Conflicts also need to be addressed

Technici4n:
Are you still going to be working on this?

PR #1126, opened 2020-10-19
Improve Github Actions and gradle.properties

TheGlitch76:
On the actions:

- ~Add a general `check` task that runs on java 11 and checks licenses and checkstyle~
- ~The java-specific tasks now only compile java, remap, and compile test mods~
- Use Docker for building (This differs from Loom and uses the `openjdk` instead of `gradle` since we want to test with the wrapper only)
- Use a consistent version for uploading artifacts

On the gradle.properties
- ~Enable parallel building by default~
- Add Gimpansor's gradle options and reduce the needed RAM to 1gb. (There was no real impact from increasing the memory to 2gb)
- ~Disable the Gradle daemon by default--it sits around eating the Xmx worth of memory, and it's really not worth wasting 1-2gb of memory on most people's machines for this.~

It would be worth considering adding these options to fabric example mod by default, but that's a discussion for a different time and place.


.github/workflows/build.yml
```diff
@@ -7,22 +7,38 @@ jobs:
...
+      - uses: actions/checkout@v2
       - uses: gradle/wrapper-validation-action@v1
       - uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java-version }}
       - uses: eskatos/gradle-command-action@v1.3.2
         with:
           gradle-version: wrapper
-          arguments: check build publishToMavenLocal --stacktrace --parallel
+          arguments: remapAllJars compileTestmodJava --stacktrace --parallel
```
modmuss50: No, im not a fan of this. The idea with this workflow is to check that the full process completes, including publishing to a maven.

We also decided `--parallel` is pointless on GHA

.github/workflows/build.yml
```diff
@@ -7,22 +7,38 @@ jobs:
...
           dependencies-cache-enabled: false
           configuration-cache-enabled: false
       - uses: actions/upload-artifact@v2
         with:
           name: Artifacts
           path: build/libs/
       - uses: actions/upload-artifact@v2
+        # Technically this shouldn't cause issues, but the MC world runs on J8 and it's better safe than sorry.
+        if: ${{ matrix.java-version == 1.8 }}
```
modmuss50: No, I would use j14/15 here, thats what I am planning on using to do the release builds.

.github/workflows/build.yml
```diff
@@ -7,22 +7,38 @@ jobs:
...
           path: ./*/build/libs/
+  check:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - uses: eskatos/gradle-command-action@v1.3.2
```
modmuss50: I am moving away from this as it randomly fails, use the docker image.

gradle.properties
```diff
@@ -1 +1,4 @@
-org.gradle.jvmargs=-Xmx2560M
+org.gradle.jvmargs=-Xmx1G -XX:SoftRefLRUPolicyMSPerMB=10 -Dlog4j2.disable.jmx=true -Dlog4j.shutdownHookEnabled=false
+org.gradle.parallel=true
+# In most instances the leaking of the Xmx worth of memory isn't worth the slight reduction in startup time
+org.gradle.daemon=false
```
modmuss50: No, dont force this on people. Default should be off. And no point in disabling the deamon

gradle.properties
```diff
@@ -1 +1,4 @@
-org.gradle.jvmargs=-Xmx2560M
+org.gradle.jvmargs=-Xmx1G -XX:SoftRefLRUPolicyMSPerMB=10 -Dlog4j2.disable.jmx=true -Dlog4j.shutdownHookEnabled=false
```
modmuss50: If these do fix the issues with the AP processor it needs fixing in loom for everyone, not here.

.github/workflows/build.yml
```diff
@@ -7,22 +7,38 @@ jobs:
...
+        with:
+          gradle-version: wrapper
+          arguments: check --stacktrace --parallel
```
modmuss50: Im not quite sure on what the benefit is here of having it seprate?

modmuss50 (changes requested):
Some good changes such as only publishing from one build, Im not quite sure on the other chanages though.

gradle.properties
TheGlitch76: I don't see the point in waiting for a potential loom fix when this is affecting PR runs [right now](https://github.com/FabricMC/fabric/runs/1272351251).
In the event that a Loom fix is found, just remove this one line from the gradle.properties. There is large gain (reduced power required to build API, and less random PR run failures) for a tiny amount of technical debt.

And I did test this, and you can do it yourself--removing these parameters causes Gradle to crash when building on 1gb of ram. It does so with no (read: not consistent or large enough to be notable) slowdown vs 2 or even 3gb with these parameters.

.github/workflows/build.yml
TheGlitch76: Idea is to give more clarity on what is wrong with a PR. IMO it's nice to not have to open up logs in order to see if a PR run failed from a build failure or if someone forgot to run checkstyle, but it sounded in discord that you were against this.

.github/workflows/build.yml
```diff
@@ -4,25 +4,21 @@ jobs:
...
-      - uses: actions/setup-java@v1
-        with:
-          java-version: ${{ matrix.java-version }}
-      - uses: eskatos/gradle-command-action@v1.3.2
-        with:
-          gradle-version: wrapper
-          arguments: check build publishToMavenLocal --stacktrace --parallel
-          dependencies-cache-enabled: false
-          configuration-cache-enabled: false
+      - run: ./gradlew build publishToMavenLocal --stacktrace --parallel
```
TheGlitch76, line 14: note: `build` depends on `check`

.github/workflows/build.yml
TheGlitch76: I was mistaken and parallel does have a small benefit

gradle.properties
modmuss50: I have pushed these args into loom, please let me know how it goes.

TheGlitch76: It works fine!

.github/workflows/build.yml
```diff
@@ -4,25 +4,21 @@ jobs:
...
-      - uses: actions/setup-java@v1
-        with:
-          java-version: ${{ matrix.java-version }}
-      - uses: eskatos/gradle-command-action@v1.3.2
-        with:
-          gradle-version: wrapper
-          arguments: check build publishToMavenLocal --stacktrace --parallel
-          dependencies-cache-enabled: false
-          configuration-cache-enabled: false
+      - run: ./gradlew build publishToMavenLocal --stacktrace --parallel --no-daemon
```
modmuss50: `--no-daemon` makes no diffrence here, its used once and then the whole container is thrown away.

gradle.properties
```diff
@@ -1 +1 @@
-org.gradle.jvmargs=-Xmx2560M
+org.gradle.jvmargs=-Xmx1G
```
modmuss50: Any reason to limit to 1G? 2.5 should be a reasonable size?

modmuss50 (changes requested):

PR #1128, opened 2020-10-20
Add Turkish language string

egeesin:


i509VCB (approved):
Quick check on google translate seems to hide nothing malicious and translation looks correct. I don't speak Turkish of course; so any other reviewers who know Turkish are welcome to comment

modmuss50 (approved):

PR #1136, opened 2020-10-23
Fluid API v1

PepperCode1:
This module adds a new FabricFlowableFluid class and FabricFluidBlock class. The new fluid class lets the user register one fluid instead of both a still fluid and a flowing fluid the vanilla way. It also lets the user set the maximum level of the fluid via an abstract method. Both classes provide some optimizations over vanilla's counterpart classes and contain some new methods that can be overriden for more control.


fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	@Override
+	protected BlockState toBlockState(FluidState state) {
+		return getFluidBlock().getDefaultState().with(stateIndexProperty, state.get(stateIndexProperty));
+	}
+	
+	/**
+	 * If a fluid is still, its state index is zero.
+	 * If a fluid is flowing, its state index is non-zero and represents its level and height.
```
i509VCB: Could you elaborate further on what part of the state index is level and which part is height?

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,110 @@
...
+	
+	// Overrides duck method in mixin.
+	public boolean isFalling(FluidState state) {
+		return state.get(stateIndexProperty) == ((FlowableFluidExtensions) this).getMaxLevel();
+	}
+	
+	@Override
+	protected void appendProperties(StateManager.Builder<Fluid, FluidState> builder) {
+		stateIndexProperty = IntProperty.of("state_index", 0, getLevels());
+        builder.add(stateIndexProperty);
```
i509VCB: this line needs to be indented with tabs

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,110 @@
...
+	public Fluid getFlowing() {
+		return this;
+	}
+
+	@Override
+	public FluidState getStill(boolean falling) {
+		if (falling) {
+			return getStill().getDefaultState().with(stateIndexProperty, ((FlowableFluidExtensions) this).getMaxLevel());
+		} else {
+			return getStill().getDefaultState();
```
i509VCB: you can drop the redundant else here and just return at the end of the method the default state.

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.fluid.FlowableFluid;
+import net.minecraft.fluid.Fluid;
+import net.minecraft.fluid.FluidState;
+import net.minecraft.state.StateManager;
+import net.minecraft.state.property.IntProperty;
+
+public abstract class FabricFlowableFluid extends FlowableFluid {
+	private IntProperty stateIndexProperty;
```
i509VCB: This int property is never set?

Should also be final (if possible)

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluidBlock.java
```diff
@@ -0,0 +1,161 @@
...
+public abstract class FabricFlowableFluidBlock extends Block implements FluidDrainable {
+	public static final VoxelShape COLLISION_SHAPE = Block.createCuboidShape(0.0D, 0.0D, 0.0D, 16.0D, 8.0D, 16.0D);
+	
+	protected final FabricFlowableFluid fluid;
+	private final IntProperty stateIndexProperty;
+	
+	public FabricFlowableFluidBlock(Settings settings) {
+		super(settings);
+		this.fluid = getFluid();
+		stateIndexProperty = this.fluid.getStateIndexProperty();
```
i509VCB: The getStateIndexProperty should be wrapped with an `Objects.requireNonNull`

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluidBlock.java
```diff
@@ -0,0 +1,161 @@
...
+	}
+	
+	@Override
+	public Fluid tryDrainFluid(WorldAccess world, BlockPos pos, BlockState state) {
+		if (state.getFluidState().isStill()) {
+			world.setBlockState(pos, Blocks.AIR.getDefaultState(), 11);
+			return fluid;
+		} else {
+			return Fluids.EMPTY;
```
i509VCB: Redundant else block

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluidBlock.java
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	@Override
+	public void randomTick(BlockState state, ServerWorld world, BlockPos pos, Random random) {
+		state.getFluidState().onRandomTick(world, pos, random);
+	}
+
+	@Override
+	protected void appendProperties(StateManager.Builder<Block, BlockState> builder) {
+		builder.add(getFluid().getStateIndexProperty());
```
i509VCB: null check the state index property with an `Objects.requireNonNull`

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluidBlock.java
```diff
@@ -0,0 +1,161 @@
...
+
+public abstract class FabricFlowableFluidBlock extends Block implements FluidDrainable {
+	public static final VoxelShape COLLISION_SHAPE = Block.createCuboidShape(0.0D, 0.0D, 0.0D, 16.0D, 8.0D, 16.0D);
+	
+	protected final FabricFlowableFluid fluid;
+	private final IntProperty stateIndexProperty;
+	
+	public FabricFlowableFluidBlock(Settings settings) {
+		super(settings);
+		this.fluid = getFluid();
```
i509VCB: Null check the fluid

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+	}
+	
+	@Override
+	public boolean isFalling(FluidState state) {
+		return state.get(FlowableFluid.FALLING);
+	}
+	
+	@ModifyConstant(
+		method = {
+			"getUpdatedState(Lnet/minecraft/world/WorldView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/BlockState;)Lnet/minecraft/fluid/FluidState;"
```
i509VCB: No need for the brackets here around the method descriptor

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+		},
+		constant = {@Constant(intValue=8)})
```
i509VCB: No need for brackets around the `@Constant`

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+	public int onValue1(int value) {
```
i509VCB: Method name should be probably `modifyMaxFluidLevel`

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+		return maxLevel;
+	}
+	
+	@ModifyConstant(
+		method = {
+			"getHeight(Lnet/minecraft/fluid/FluidState;)F",
+			"getShape(Lnet/minecraft/fluid/FluidState;Lnet/minecraft/world/BlockView;Lnet/minecraft/util/math/BlockPos;)Lnet/minecraft/util/shape/VoxelShape;"
+		},
+		constant = {@Constant(floatValue=9.0f)})
```
i509VCB: No need for Brackets around `@Constant`

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+	public float onValue2(float value) {
```
i509VCB: Name this one as well

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/extension/FlowableFluidMixin.java
```diff
@@ -0,0 +1,95 @@
...
+		return maxLevel+1;
+	}
+	
+	@ModifyConstant(
+		method = {
+			"getVelocity(Lnet/minecraft/world/BlockView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/fluid/FluidState;)Lnet/minecraft/util/math/Vec3d;"
+		},
+		constant = {@Constant(floatValue=0.8888889f)})
```
i509VCB: I assume you get the point now. If it's one entry and the input object is an array, you can pass a single object in an annotation.

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
PepperCode1: It's not possible to make it final because it's set in the appendProperties method.

i509VCB: OK I see.

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluidBlock.java
PepperCode1: This value cannot be null unless mixins/asm are used. Should I still wrap it?

fabric-fluid-extensions-v1/src/main/java/net/fabricmc/fabric/api/fluid/extension/v1/FabricFlowableFluid.java
PepperCode1: In the newer commits I removed the "and height" part from the javadoc, since that is confusing. The height of the fluid is level/(maxLevel+1), as seen in FlowableFluid#getHeight.

sylv256:
# ~~i have *[some serious]* concerns~~

1. ~~where is fluid volume api~~
1. ~~does it use millibuckets or millitoffets?~~

nevermind

Prospector:
Imo, this module should just be called `fabric-fluid-api-v1` so other future stuff like adding more tags for fluid behavior (such as for swimming, drowning, etc.) can be included in a more general fluid module, much like we have for items (`fabric-item-api-v1`).

A fluid volume api would be `fabric-fluid-volume-api-v1`

PepperCode1:
I have renamed the module as per Prospector's suggestion.

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// Overrides duck method in mixin.
+	public boolean isFalling(FluidState state) {
+		return state.get(stateIndexProperty) == ((FlowableFluidExtensions) this).getMaxLevel();
+	}
+
+	@Override
+	protected void appendProperties(StateManager.Builder<Fluid, FluidState> builder) {
+		stateIndexProperty = IntProperty.of("state_index", 0, getLevels());
```
Virtuoel: This current approach would result in duplicated property instances when there are multiple instances of `FabricFlowableFluid` that have the same amount of levels as each other. The index property is also not accessible without having an instance of `FabricFlowableFluid` or `FabricFlowableFluidBlock` to call `getStateIndexProperty` on.  

Properties should be relatively easily accessible from a static context and also have no duplicated instances, with the same property instance being shared among all blocks or fluids that use it, similar to how the vanilla ones in `Properties` are all static with e.g. the 6-way direction property being one instance shared between all 6-way directional blocks.  


I suppose one approach could be a method that takes a level and returns the corresponding property from a map of levels to properties, or an array where the index is the levels, or some other data structure, with it first creating the property for the level if it doesn't exist yet.  

(Don't know the best approach or naming for something like that which would also fit with the rest of Fabric API. Maybe some sort of `FabricProperties.INSTANCE.getFluidIndexFromLevels` method for lazily creating/getting the corresponding property? Not sure.)  

Anyways, with properties, they should not have duplicate instances and should be easily accessible in a static context.

Virtuoel (changes requested):

PepperCode1:
I am unsure about this PR because my original goal with it was to essentially replace vanilla's fluid system (which IMO isn't very good), and this PR accomplishes this quite well. I am now realizing though, that forcing users to use a single fluid and the custom fluid block class takes away control, and that this PR might not wholly fit into Fabric API. In the Discord API channel, it was discussed that a unified fluid class does not fit into Fabric API and that the FluidBlock class should be mixed into instead of replaced, but I am really not sure how to improve the system only partially.

PR #1146, opened 2020-10-31
Sync and save builtin registries to fix MC-202036

modmuss50:
Currenly very Lightly tested, done in a rush.

Should fix https://bugs.mojang.com/browse/MC-202036

Testing and feedback would be much applicated.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/FabricRegistryInit.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/FabricRegistryInit.java
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.impl.registry.sync;
 
+import net.minecraft.util.registry.BuiltinRegistries;
 import net.minecraft.util.registry.Registry;
 
 import net.fabricmc.api.ModInitializer;
@@ -172,5 +173,11 @@ public class FabricRegistryInit implements ModInitializer {
 
 		// Doesnt seem to be serialised or synced.
 		RegistryAttributeHolder.get(Registry.LOOT_CONDITION_TYPE);
+
+		// Marked as modded to always fix https://bugs.mojang.com/browse/MC-202036
+		RegistryAttributeHolder.get(BuiltinRegistries.BIOME)
+				.addAttribute(RegistryAttribute.SYNCED)
+				.addAttribute(RegistryAttribute.PERSISTED)
+				.addAttribute(RegistryAttribute.MODDED);
 	}
 }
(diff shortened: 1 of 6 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/FabricRegistryInit.java
```diff
@@ -172,5 +173,11 @@ public void onInitialize() {
 
 		// Doesnt seem to be serialised or synced.
 		RegistryAttributeHolder.get(Registry.LOOT_CONDITION_TYPE);
+
+		// Marked as modded to always fix https://bugs.mojang.com/browse/MC-202036
+		RegistryAttributeHolder.get(BuiltinRegistries.BIOME)
+				.addAttribute(RegistryAttribute.SYNCED)
+				.addAttribute(RegistryAttribute.PERSISTED)
+				.addAttribute(RegistryAttribute.MODDED);
```
coderbot16, line 181: Could also note/comment here that always syncing the registry content won't mess up with vanilla clients since biome IDs are stable under vanilla, unlike block state IDs and similar.

coderbot16 (approved):
Looks good. The code probably should still be tested somehow, but otherwise it looks fine. I have a setup that can reliably reproduce the issue, so I'll test this updated module with that.

CorgiTaco (approved):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinLevelStorageSession.java
```diff
@@ -135,18 +137,24 @@ public void saveWorld(DynamicRegistryManager registryTracker, SaveProperties sav
...
 	}
 
 	// TODO: stop double save on client?
 	@Inject(method = "readLevelProperties", at = @At("HEAD"))
 	public void readWorldProperties(CallbackInfoReturnable<SaveProperties> callbackInfo) {
+		initialiseRegistryMap(RegistrySyncManager.STANDARD);
+		initialiseRegistryMap(RegistrySyncManager.BUILTIN);
+	}
+
+	private void initialiseRegistryMap(RegistrySyncManager syncManager) {
```
Chocohead, line 151: Should probably add a `@Unique` to this

Player3324 (approved):
Looks correct to me

ghost (approved):
Looks good, and doesn't seem all that controversial I would just personally recommend merging it because of it's importance. If there's a massive issue we encounter down the road (but again it looks good so I don't believe that's likely) we can always patch it.

modmuss50:
Im fairly happy with this, a few people have tested and confirmed the issue is fixed with this. I do want to test with mismatched fabric versions on the server + client along with a vanilla client/vanilla server to make sure that stuff isn’t broken.

When mojang fix the bug them selves (I believe they will be forced to once they add a biome?) a world upgrade path will needed to be added.

YTG1234 (approved):

modmuss50:
This fix is not correct. The dynamic registry needs sorting, not this one.

PR #1148, opened 2020-11-05
20w45a: Partially update Fabric API

i509VCB:
This updates modules where needed due to changes.

`fabric-lifecycle-events-v1` has had the (block)entity (un)load events turned off for the moment until we have mappings to help assist with updating.

Thanks to @LambdAurora resource-loader-v0 is also now updated in this PR.

Does not guarantee the game works, just a few modules that have been tested.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ def ENV = System.getenv()
 
 class Globals {
 	static def baseVersion = "0.25.1"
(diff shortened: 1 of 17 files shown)


Player3324 (approved):
LGTM, thanks

ghost (approved):

modmuss50 (approved):

PR #1152, opened 2020-11-06
Temp fix for feature renderer testmod, update mappings, reimplement client entity events

i509VCB:


Player3324 (approved):
LGTM, thanks. Did you check if the entities are always being reliably tracked now?

i509VCB:
Reminds me that I need to check an edge case.

PR #1163, opened 2020-11-13
Spin up a dedicated server on github actions + add run tasks for test mods.

modmuss50:
`runAutoTestServer` runs a dedicated server with all test mods, and then gracefully shuts it down. Mixin will also audit the remaining mixin targets.

See an example run here: https://github.com/modmuss50/fabric/runs/1397667784?check_suite_focus=true

### DIFF
--- a/fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseTestInit.java
+++ b/fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseTestInit.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


i509VCB (approved):
Nothing looks wrong to me

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseTestInit.java
```diff
@@ -0,0 +1,40 @@
...
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerTickEvents;
+
+public class FabricApiBaseTestInit implements ModInitializer {
+	private int ticks = 0;
+
+	@Override
+	public void onInitialize() {
+		if (Boolean.parseBoolean(System.getProperty("fabric.autoTest", "false"))) {
```
Player3324: There's a suitable niche API https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-

Player3324 (approved):

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseTestInit.java
Player3324: I'd however just use `System.getProperty("fabric.autoTest") != null`, ignoring the value (makes it work with just `-Dmy.property`)

PR #1168, opened 2020-11-17
Fix MC-202036 - Shifting biome IDs

modmuss50:
This is loosly bassed off the existing registrys sync stuff, but just the required bits, and trying to be minimal. I have tried to leave some comments explainng whats doing what, hopefully its fairly easy to follow.

This is missing the backing up of the data files, do we still need/want this? (Minecraft does appear to show a warning screen with an option to backup if changing this stuff?)

I look forward to your reviews and testing.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.fabricmc.fabric.impl.registry.sync;
 
 import java.io.IOException;
@@ -44,8 +60,14 @@ public class PersistentDynamicRegistryHandler {
 		CompoundTag registries = new CompoundTag();
 
 		// For now we only care about biomes, but lets keep our options open
+		CompoundTag biomeRegistryData = null;
+
+		if (existingTag != null) {
+			biomeRegistryData = existingTag.getCompound(Registry.BIOME_KEY.getValue().toString());
+		}
+
 		MutableRegistry<?> registry = dynamicRegistryManager.get(Registry.BIOME_KEY);
-		CompoundTag biomeIdMap = remapRegistry(Registry.BIOME_KEY.getValue(), registry, existingTag);
+		CompoundTag biomeIdMap = remapRegistry(Registry.BIOME_KEY.getValue(), registry, biomeRegistryData);
 		registries.put(Registry.BIOME_KEY.getValue().toString(), biomeIdMap);
 
 		CompoundTag outputTag = new CompoundTag();
@@ -59,14 +81,50 @@ public class PersistentDynamicRegistryHandler {
 	 * Then writes out the ids in the registry (remapped or a new world).
 	 * Keeps hold of the orphaned registry entries as to not overwrite them.
 	 */
-	@SuppressWarnings("unchecked")
-	private static CompoundTag remapRegistry(Identifier registryId, MutableRegistry registry, @Nullable CompoundTag existingTag) throws RemapException {
+	private static <T> CompoundTag remapRegistry(Identifier registryId, MutableRegistry<T> registry, @Nullable CompoundTag existingTag) throws RemapException {
 		if (!(registry instanceof RemappableRegistry)) {
 			throw new UnsupportedOperationException("Cannot remap un re-mappable registry");
 		}
 
-		// If we have some existing ids we remap the registry with those
-		if (existingTag != null) {
+		boolean isModded = registry.getIds().stream().anyMatch(id -> !id.getNamespace().equals("minecraft"));
+
+		// The current registry might not be modded, but we might have previous changed vanilla ids that we should try and remap
+		if (existingTag != null && !isModded) {
+			isModded = existingTag.getKeys().stream()
+					.map(existingTag::getString)
+					.map(Identifier::new)
+					.anyMatch(id -> !id.getNamespace().equals("minecraft"));
+		}
+
+		if (LOGGER.isDebugEnabled()) {
+			if (existingTag == null) {
+				LOGGER.debug("No existing data found, assuming new registry with {} entries", registry.getIds().size());
+			} else {
+				for (T entry : registry) {
+					//noinspection unchecked
+					Identifier id = registry.getId(entry);
+					int rawId = registry.getRawId(entry);
+
+					if (id == null) continue;
+
+					if (existingTag.getKeys().contains(id.toString())) {
+						int existingRawId = existingTag.getInt(id.toString());
+						if (rawId != existingRawId) {
+							LOGGER.debug("Remapping {} {} -> {}", id.toString(), rawId, existingRawId);
+						} else {
+							LOGGER.debug("Using existing id for {} {}", id.toString(), rawId);
+						}
+					} else {
+						LOGGER.debug("Found new registry entry {}", id.toString());
+					}
+				}
+			}
+
+		}
+
+		// If we have some existing ids and the registry contains modded/datapack entries we remap the registry with those
+		if (existingTag != null && isModded) {
+			LOGGER.debug("Remapping {} with {} entries", registryId, registry.getIds().size());
 			Object2IntMap<Identifier> idMap = new Object2IntOpenHashMap<>();
 
 			for (String key : existingTag.getKeys()) {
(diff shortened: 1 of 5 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinMain.java
```diff
@@ -0,0 +1,40 @@
...
+	private static void main(String[] args, CallbackInfo ci, OptionParser optionParser, OptionSpec optionSpec, OptionSpec optionSpec2, OptionSpec optionSpec3, OptionSpec optionSpec4, OptionSpec optionSpec5, OptionSpec optionSpec6, OptionSpec optionSpec7, OptionSpec optionSpec8, OptionSpec optionSpec9, OptionSpec optionSpec10, OptionSpec optionSpec11, OptionSpec optionSpec12, OptionSpec optionSpec13, OptionSpec optionSpec14, OptionSet optionSet, DynamicRegistryManager.Impl impl, Path path, ServerPropertiesLoader serverPropertiesLoader, Path path2, EulaReader eulaReader, File file, YggdrasilAuthenticationService yggdrasilAuthenticationService, MinecraftSessionService minecraftSessionService, GameProfileRepository gameProfileRepository, UserCache userCache, String string, LevelStorage levelStorage, LevelStorage.Session session, DataPackSettings dataPackSettings, boolean bl, ResourcePackManager resourcePackManager, DataPackSettings dataPackSettings2, CompletableFuture completableFuture, ServerResourceManager serverResourceManager2, RegistryOps registryOps) {
```
modmuss50: Just a few locals ;)

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Remaps a registry if existing data is passed in.
+	 * Then writes out the ids in the registry (remapped or a new world).
+	 * Keeps hold of the orphaned registry entries as to not overwrite them.
+	 */
+	@SuppressWarnings("unchecked")
+	private static CompoundTag remapRegistry(Identifier registryId, MutableRegistry registry, @Nullable CompoundTag existingTag) throws RemapException {
+		if (!(registry instanceof RemappableRegistry)) {
+			throw new UnsupportedOperationException("Cannot remap un re-mappable registry");
```
i509VCB: Maybe put the ID of the registry in the exception

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinMain.java
i509VCB: o p t i o n s p e c

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/client/MixinMinecraftClient.java
```diff
@@ -44,4 +56,17 @@ public void disconnectAfter(Screen screen_1, CallbackInfo info) {
...
 	}
+
+	@Inject(method = "createSaveProperties", at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/util/dynamic/RegistryOps;of(Lcom/mojang/serialization/DynamicOps;Lnet/minecraft/resource/ResourceManager;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;)Lnet/minecraft/util/dynamic/RegistryOps;"))
+	private static void createSaveProperties(LevelStorage.Session session, DynamicRegistryManager.Impl impl, ResourceManager resourceManager, DataPackSettings dataPackSettings, CallbackInfoReturnable<SaveProperties> cir) {
+		Path saveDir = ((AccessorLevelStorageSession) session).getDirectory();
+		PersistentDynamicRegistryHandler.remapDynamicRegistries(impl, saveDir);
+	}
+
+	// synthetic in method_29607
+	@Inject(method = "method_31125", at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/util/dynamic/RegistryOps;of(Lcom/mojang/serialization/DynamicOps;Lnet/minecraft/resource/ResourceManager;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;)Lnet/minecraft/util/dynamic/RegistryOps;"))
```
i509VCB, line 67: Maybe quote what player said about this injection here on discord?

coderbot16:
Looks good from a quick skim, only issue I have is that old biome registry content (from registry sync) will be lost when people upgrade. It would probably be a good idea to migrate that data over if possible?

modmuss50:
The only stuff we would have to migrate 1.16.1, it’s possible, just will take more work and I’d ideally need something to test with

modmuss50:
I have done some testing on both the client and server and as far as I can tell this is working as epxected.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinMain.java
modmuss50: It might not be a bad idea to look at using a redirect here, or a ModifyArg as this mixin is going to be super fragile to updates.

i509VCB: Yeah I had a similar thing in the server world load event. The local variable are insanity. If you do choose that please note with a comment about how the redirect is easier to maintain than a modify arg due to large amount of local variables.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		MutableRegistry<?> registry = dynamicRegistryManager.get(Registry.BIOME_KEY);
+		CompoundTag biomeIdMap = remapRegistry(Registry.BIOME_KEY.getValue(), registry, biomeRegistryData);
+		registries.put(Registry.BIOME_KEY.getValue().toString(), biomeIdMap);
+
+		CompoundTag outputTag = new CompoundTag();
+		outputTag.putInt("version", 1);
+		outputTag.put("registries", registries);
+		return registries;
```
Player3324, line 82: This completely ignores `outputTag`. I'd move the version wrapping to read/writeCompoundTag.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
```diff
@@ -0,0 +1,200 @@
...
+						}
+					} else {
+						LOGGER.debug("Found new registry entry {}", id.toString());
+					}
+				}
+			}
+		}
+
+		// If we have some existing ids and the registry contains modded/datapack entries we remap the registry with those
+		if (existingTag != null && isModded) {
```
Player3324: It may be desirable to fix the bug with vanilla datapacks too?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
```diff
@@ -0,0 +1,200 @@
...
+		// Save all ids as they appear in the remapped, or new registry to disk even if not modded.
+		for (T entry : registry) {
+			//noinspection unchecked
+			Identifier id = registry.getId(entry);
+
+			if (id == null) {
+				continue;
+			}
+
+			int rawId = registry.getRawId(entry);
```
Player3324, line 160: Do we want to validate rawId too? (skip < 0)

Player3324 (commented):
Looks good overall! I found an oversight/bug and two potential enhancements. Thanks for tackling this!

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
i509VCB: I think the registry is gonna give you -1 if the object is not keyed in registry

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
modmuss50: `isModded` in this case means any registry entry that contains an entry not in the minecraft namespace. The datapack I was testing with used its own namespace. I might be wrong but I would assume all custom biomes to not use the minecraft namespace?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
modmuss50: Good idea, might as well.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
modmuss50: Good idea. Yes.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/PersistentDynamicRegistryHandler.java
Player3324: Right, that's fine IMO. Maybe clarify the comment a little?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinMain.java
modmuss50: I have moved to 2 redirects.

Player3324 (approved):

PR #1040, opened 2020-08-22
Dimension API 1.16.2

shartte:
This makes the dimension API work again.

Vanilla no longer creates portals for dimensions other than the nether, so the entire Placement Handling is not necessarily needed for mods that just want to port to a specific point in their dimension.

I also introduced two different term here:
- ~~Untargeted Teleportation means teleporting to a dimension without specifying a destination point (nether portals, end portals, Vanilla `moveToWorld` API)~~
- Targeted Teleportation means teleporting to a dimension while specifying the destination point in the target dimension fully (this will apply to most mods porting players into their own dimensions using teleports or various other methods)

~~A modded dimension needs to opt-in to support the untargeted teleportation case, by specifying a handler for it. That handler would then carry out tasks such as:~~
- ~~Determining the actual destination, for example by deriving it from the entities current position~~
- ~~Placing a portal - if desired~~

~~To support special use cases such as performing an untargeted teleportation into a mod dimension _WITHOUT_ placing a portal, the API supports optionally passing attributes along with the teleportation request, which the handler can then inspect.~~

**edit**: I decided to rip out the untargeted teleportation for now since it seems like a niche use case and would require us to replicate Vanilla's coordinate mapping/scaling behavior to teleport to the overworld in an untargetted fashion.

Additionally, I introduced two new methods to allow mods to mark their dimension and dimension_type JSON files as stable so that they do not trigger the experimental warning screen anymore. This requires any used Chunk Generators to use a fully stable codec (the test mod's Codec was experimental, which cost me quite a bit of time).

The mixin into RegistryOps could potentially be replaced with dynamic registry events if those are merged, but is currently independent.

PR #1172, opened 2020-11-20
Update the Dimension API

Waterpicker:
Supersedes teleportation component of #1040


fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/FabricDimensions.java
```diff
@@ -37,74 +36,27 @@ private FabricDimensions() {
...
-	 * <p>After calling this method, {@code teleported} may be invalidated. Callers should use
-	 * the returned entity for any further manipulation.
-	 *
-	 * @param teleported   the entity to teleport
-	 * @param destination  the dimension the entity will be teleported to
-	 * @param customPlacer custom placement logic that will run before the default one,
-	 *                     or {@code null} to use the dimension's default behavior.
-	 * @param <E>          the type of the teleported entity
-	 * @return the teleported entity, or a clone of it
+	 * @param target      where the entity will be placed in the target world.
```
i509VCB, line 49: The code in `Entity#moveToWorld` will fail teleportation if the target is null.
I believe we should document that?

Reason for not requiring non-null here is a case such as another mod which may wish to intercept the teleport target via an event in the future.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/FabricDimensions.java
```diff
@@ -37,74 +36,27 @@ private FabricDimensions() {
...
+	 *                    As in Vanilla, the target's velocity is not applied to players.
+	 * @param <E>         the type of the teleported entity
+	 * @return Returns the teleported entity in the target dimension, which may be a new entity or <code>teleported</code>,
+	 * depending on the entity type.
 	 * @throws IllegalStateException if this method is called on a client entity
 	 * @apiNote this method must be called from the main server thread
 	 */
-	public static <E extends Entity> E teleport(E teleported, ServerWorld destination, /*Nullable*/ EntityPlacer customPlacer) {
+	public static <E extends Entity> E teleport(E teleported, ServerWorld destination, TeleportTarget target) {
```
i509VCB, line 59: Can you add a Nullable annotation here for the return value as an entity can fail to teleport?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
```diff
@@ -16,116 +16,41 @@
...
 
-import net.fabricmc.fabric.api.dimension.v1.EntityPlacer;
 import net.fabricmc.fabric.api.dimension.v1.FabricDimensions;
-import net.fabricmc.fabric.mixin.dimension.EntityHooks;
 
 public final class FabricDimensionInternals {
+	/**
+	 * The target passed to the last call to {@link FabricDimensions#teleport(Entity, ServerWorld, TeleportTarget)}.
+	 */
+	private static TeleportTarget currentTarget;
```
i509VCB, line 31: I believe this should be fine as the method which sets this field will fail unless the method is called on the server thread.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+@Mixin(value = Main.class, remap = false)
+public class ServerBugfixMixin {
+	private static LevelStorage.Session fabric_session;
+
+	private static DynamicRegistryManager.Impl fabric_dynamicRegistry;
+
+	private static RegistryOps<Tag> fabric_registryOps;
+
+	@ModifyVariable(at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/util/registry/DynamicRegistryManager;create()Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;"), method = "main", remap = false, allow = 1)
```
i509VCB, line 64: For others wondering, this capture via modify variable is required when discussing with Gimpansor who originally wrote the PR due to Fabric Loader's injection point for mod loading. This should be fine.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
```diff
@@ -0,0 +1,88 @@
...
+ * dimension testmod, and no world directory. If the dimension is available (i.e. in /execute in, or via
+ * the testmod's commands), then the bug is fixed and this Mixin can be removed.
+ */
+@Mixin(value = Main.class, remap = false)
+public class ServerBugfixMixin {
+	private static LevelStorage.Session fabric_session;
+
+	private static DynamicRegistryManager.Impl fabric_dynamicRegistry;
+
+	private static RegistryOps<Tag> fabric_registryOps;
```
i509VCB, line 62: Could you throw an `@Unique` annotation on those three static fields?

i509VCB (approved):

Player3324:
If I remember correctly we previously had layered teleport target handling:

- teleporter specifies the destination (what this PR's new API mandates always now)
- dimension specifies the destination (used to be the overload without target)
- some fabric default specifies it (probably just same pos+orientation)

There are quite a few cases where the teleporter doesn't know the specific dimension's properties, making in incapable of determining a good destination. Think of e.g. some mini-game like area with a very specific layout instead of an open world area. I wonder if anything can be feasibly done in this area and how it impacts the proposed API.

I'd be happy to go with the proposed limited approach for now though, the PR looks fine in this regard.

Player3324 (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
```diff
@@ -0,0 +1,92 @@
...
+import net.minecraft.util.dynamic.RegistryOps;
+import net.minecraft.util.registry.DynamicRegistryManager;
+import net.minecraft.world.gen.GeneratorOptions;
+import net.minecraft.world.level.LevelInfo;
+import net.minecraft.world.level.LevelProperties;
+import net.minecraft.world.level.storage.LevelStorage;
+
+/**
+ * This Mixin aims to solve a Minecraft Vanilla bug where datapacks are ignored during creation of the
+ * initial LevelProperties when a dedicated server creates a completely new level.
```
modmuss50, line 38: Just came across this the other day 👍

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * <p>This Mixin aims to solve this problem by saving and loading the level.dat file once, after
+ * a new set of level properties is created. This will apply the same logic as reloading the
+ * level.dat after a restart, now including all datapack dimensions.
+ *
+ * <p>See https://bugs.mojang.com/browse/MC-195468 for a related bug report.
+ *
+ * <p>In 1.17: Retest if this bug still occurs without this Mixin by launching a dedicated server with the
+ * dimension testmod, and no world directory. If the dimension is available (i.e. in /execute in, or via
+ * the testmod's commands), then the bug is fixed and this Mixin can be removed.
```
modmuss50, line 51: The testmod should fail hard on first tick if fails to find the modded dim, we now startup a fresh dedicated server when building on GHA. Should be trivial to add

modmuss50 (changes requested):
I would just add a little bit of self testing to the testmod as its trivial. 

Looks good otherwise 👍

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
i509VCB: Waterpicker, for the structure of this I would advise looking at command api for reference.

Waterpicker: I just pushed a potential automatic testmod test thingy.

fabric-dimensions-v1/build.gradle
```diff
@@ -9,5 +9,6 @@ dependencies {
 moduleDependencies(project, [
 		'fabric-api-base',
 		'fabric-networking-v0',
-		'fabric-registry-sync-v0'
+		'fabric-registry-sync-v0',
+		'fabric-lifecycle-events-v1'
```
i509VCB: Please place the lifecycle events dependencies under the `testmodCompile` configuration?

That is because only the testmod depends on the lifecycle events.

fabric-dimensions-v1/build.gradle
```diff
@@ -9,5 +9,6 @@ dependencies {
 moduleDependencies(project, [
 		'fabric-api-base',
 		'fabric-networking-v0',
```
i509VCB: I believe we can drop networking and registry sync as dependencies.

fabric-dimensions-v1/src/testmod/java/net/fabricmc/fabric/test/dimension/FabricDimensionTest.java
```diff
@@ -65,6 +70,37 @@ public void onInitialize() {
...
+			ServerWorld world = server.getWorld(WORLD_KEY);
+
+			if (world == null) {
+				throw new AssertionError("Test world doesn't exist.");
+			}
+
+			Entity entity = COW.create(overworld);
+
+			if(!entity.world.getRegistryKey().equals(World.OVERWORLD)) {
+				throw  new AssertionError("Entity starting world isn't the overworld");
```
i509VCB: double space between `throw` and `new`. Likely to fail checkstyle

modmuss50 (approved):

fabric-dimensions-v1/build.gradle
```diff
@@ -3,10 +3,10 @@ version = getSubprojectVersion(project, "1.0.0")
 
 dependencies {
 	testmodCompile project(path: ':fabric-command-api-v1', configuration: 'dev')
+	testmodCompile project(path: ':fabric-resource-loader-v0', configuration: 'dev')
+	testmodCompile project(path: 'fabric-lifecycle-events-v1', configurations: 'dev')
```
i509VCB:
```suggestion
	testmodCompile project(path: ':fabric-lifecycle-events-v1', configurations: 'dev')
```

i509VCB (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+@Mixin(value = Main.class, remap = false)
+public class ServerBugfixMixin {
```
LambdAurora, line 54: I'm not sure if this should really be in the dimensions API, this would rather fit in resource-loader as it doesn't affect only the worldgen but any datapack on world creation as the description states, so in theory this also affects tags, functions and more.

PR #1173, opened 2020-11-20
Add tag builder method for the new optional entries

williewillus:
Simple extension to the API to add optional entries to the tag builders. I hope this doesn't take 5 months to land like my last PR :tiny_potato:

### DIFF
--- a/fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
+++ b/fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
@@ -31,13 +31,13 @@ public interface FabricTagBuilder<T> {
 	void clearTagEntries();
 
 	/**
-	 * Add an optional object entry
+	 * Add an optional object entry.
 	 * @param id The ID of the object to add
 	 */
 	void addOptional(Identifier id);
 
 	/**
-	 * Add an optional tag entry
+	 * Add an optional tag entry.
 	 * @param id The ID of the tag to add
 	 */
 	void addOptionalTag(Identifier id);
--- a/fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
+++ b/fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
@@ -16,8 +16,6 @@
 
 package net.fabricmc.fabric.api.tag;
 
-import net.minecraft.util.Identifier;
-
 /**
  * Interface implemented by {@link net.minecraft.tag.Tag.Builder} instances when
  * Fabric API is present.
@@ -29,22 +27,4 @@ public interface FabricTagBuilder<T> {
 	 * Clear the contained entries and mark the tag as replaced.
 	 */
 	void clearTagEntries();
-
-	/**
-	 * Add an optional entry of type {@code <T>} to the tag.
-	 * The object identified by {@code id} is not required to be present on load,
-	 * which is useful for integration with other mods.
-	 * @param id The ID of the object to add
-	 * @see net.minecraft.tag.Tag.Builder#add(Identifier, String) for the non-optional version of this method.
-	 */
-	void addOptionalObject(Identifier id);
-
-	/**
-	 * Add an optional tag entry to the tag.
-	 * The tag identified by {@code id} is not required to be present on load,
-	 * which is useful for integration with other mods.
-	 * @param id The ID of the tag to add
-	 * @see net.minecraft.tag.Tag.Builder#addTag(Identifier, String) for the non-optional version of this method.
-	 */
-	void addOptionalTag(Identifier id);
 }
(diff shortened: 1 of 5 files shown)


fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagBuilder.java
```diff
@@ -57,4 +60,14 @@ public void clearTagEntries() {
...
 	}
+
+	@Override
+	public void addOptional(Identifier id) {
+		add(new Tag.OptionalObjectEntry(id), "fabric-tag-extensions");
+	}
+
+	@Override
+	public void addOptionalTag(Identifier id) {
+		add(new Tag.OptionalTagEntry(id), "fabric-tag-extensions");
```
williewillus: nb: this string is only for debug logging to show who added the entry, so I just chose the module name. Not worth adding a parameter for and burdening callers

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagBuilder.java
```diff
@@ -57,4 +60,14 @@ public void clearTagEntries() {
 		entries.clear();
 		fabric_clearCount++;
 	}
+
+	@Override
+	public void addOptional(Identifier id) {
+		add(new Tag.OptionalObjectEntry(id), "fabric-tag-extensions");
```
i509VCB: Source should match the fabric api module name completely.

```suggestion
		add(new Tag.OptionalObjectEntry(id), "fabric-tag-extensions-v0");
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagBuilder.java
i509VCB:
```suggestion
		add(new Tag.OptionalTagEntry(id), "fabric-tag-extensions-v0");
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagBuilder.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Optional;
 
 import com.google.gson.JsonObject;
+import net.minecraft.util.Identifier;
```
i509VCB: Could you verify the import order here?

Tip: `gradlew check` will run checkstyle et al

i509VCB (approved):
Generally looks good, three minor comments that should be a 30s fix.

williewillus:
comments addressed

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
```diff
@@ -27,4 +29,16 @@
 	 * Clear the contained entries and mark the tag as replaced.
 	 */
 	void clearTagEntries();
+
+	/**
+	 * Add an optional object entry.
+	 * @param id The ID of the object to add
+	 */
+	void addOptional(Identifier id);
```
Player3324: `addOptionalObject` maybe to reduce confusion? Would also be nice to describe these in more detail. The reader should get an idea what these methods are needed/useful for.

williewillus: They correspond directly to the non-optional methods on the base vanilla class, so I didn't really want to go into a whole explanation of the tag system.

Player3324: There is no need to explain the whole tag system, just the differences Tag<>Object and Optional<>"Normal". E.g. something along the lines of "Optional entries will not produce load errors if their referenced object are missing" or whatever is appropriate.

Please consider adding this documentation, we try to guide the curious viewer to a reasonable extent. This file is particularly unusual/confusing since it's a detached interface that needs casting to be useful, I hope we can at some point migrate it to a subclass of Tag.Builder. Right now the two addOptional methods are quite exposed compared to the vanilla equivalents if someone sees our class before the vanilla one.

modmuss50 (approved):

modmuss50:
Please correct me if im wrong, but I fail to see why we need an API for this, as far as I can tell everything exposed here is already exposed publicly in vanilla.

All of the `Tag.Builder.add` methods are public, along with the Builder its self.

The `OptionalObjectEntry` is also public.

williewillus:
@modmuss50 I don't see it:

image

All I can do is add a required item (or items), or a required tag.

williewillus:
The yarn name for the method is perhaps a misnomer. It doesn't return a tag builder, it returns an `ObjectBuilder` which wraps a private `Tag.Builder`

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/FabricTagBuilder.java
williewillus: done

modmuss50:
This Mixin is adding `addOptionalObject/Tag` to `Tag.Builder` where as far as I can tell no api is required to do achiveve this.

What I think you want is for those methods to target `AbstractTagProvider.ObjectBuilder`?

williewillus:
Doh, thanks

williewillus:
Changed to target proper class, I named the interface a bit differently than convention to distinguish it from the normal builder.

modmuss50 (approved):

PR #1176, opened 2020-11-21
Fix MC-149777 - java.util.ConcurrentModificationException when using Java 11 or above

modmuss50:
This fix has been well battle tested in: https://github.com/modmuss50/Voyager

I see no reason why it cannot go into fabric.


i509VCB (approved):

BoogieMonster1O1 (approved):

Player3324 (approved):
Looks good, pretty sure this is a complete fix in this case!

PR #1177, opened 2020-11-22
Reimplement block entity (un)load events

i509VCB:
This completely reimplements the lifecycle events module.

Pending testing of course to validate the mixins and reenabling the tests.


fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/impl/event/lifecycle/LoadedChunksCache.java
```diff
@@ -0,0 +1,38 @@
...
+package net.fabricmc.fabric.impl.event.lifecycle;
+
+import java.util.Set;
+
+import net.minecraft.world.chunk.WorldChunk;
+
+/**
+ * A simple marker interface which holds references to chunks which block entities may be loaded or unloaded from.
+ */
+public interface LoadedChunksCache {
```
modmuss50, line 26: In the future I can see this possibly being a useful api.

modmuss50 (approved):

PR #1178, opened 2020-11-23
Provide replacement for ModelAppender that allows loading non-blockstate paths

williewillus:
A common ask is the ability to load an arbitrary model json from the resource paths (for example, to use in a block entity or entity renderer). The existing `ModelAppender` system only accepts `ModelIdentifier`s, which identify a variant within a blockstate json (or a item model via the magic variant `inventory`), which means its use is limited to those cases.

This PR extends `ModelAppender` to support `Identifier` as well for loading a json path anywhere in the model tree. The existing vanilla `addModel` called by us only takes `ModelIdentifier`, but the restriction is arbitrary. The same logic can be reused verbatim for plain `Identifier`.

For those wiling to look, this is equivalent to Forge's `addSpecialModel` which I also added and has been used widely, so in the implementation/design space this is pretty well-tread ground already. It's used in Botania-forge and I need it for Botania-fabric :)


fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelAppender.java
```diff
@@ -20,8 +20,26 @@
...
+	/**
+	 * Provides an opportunity inform the game that you would like it to load and bake a model,
+	 * even if that model is not used by any blocks or items.
+	 * @param out Accepts paths to be loaded. Arguments that are {@link ModelIdentifier} will be
+	 *            loaded through the blockstate JSON system or, if the variant is {@code inventory}, the item model folder.
+	 *            Otherwise, the argument is directly loaded as a JSON.
+	 *            For example, <pre>new Identifier("mymod", "foo/bar")</pre> will request loading of the file
+	 *            <pre>/assets/mymod/models/foo/bar.json</pre>
+	 */
+	default void appendAllIdentifiers(ResourceManager manager, Consumer<Identifier> out) {
```
williewillus: this bothers me in that lambdas still default to the older/deprecated method. Should I just make a new interface for this and have the old interface extend it?

i509VCB: I believe that may be the best option. I am not going to call myself a master at using json models but within the api structure that would make the most sense since it would allow any old implementations which just move over to the new interface and change the generic to continue working.

i509VCB (approved):
API changes look good. Some class names may need a change or two but the general structure looks good.

Player3324:
This would be a good opportunity to rename this facility, ModelAppender isn't great at describing what this is/does and GeneralModelAppender is even worse.

I am thinking about ExtraModelProvider or similar: "Interface for mod code that submits additional models at the appropriate time. This will caused the desired models to be loaded and processed for ingame use in addition to the models directly linked to blocks or items by their name. Example use cases include [...]. The submitModels method will be invoked during resource pack reloads. @param ..."

appendAllIdentifiers -> submitModels

Please decouple the new interface from the old one, there is no good reason to try to adapt one to the other beyond simplifying the implementation insignificantly. It is more important to keep the API clean.

williewillus:
Renamed interface and separated it from the old one
The internals are still done by shimming the old interface to the new one on registration

Juuxel (approved):

Player3324 (approved):
Good now, thanks!

modmuss50 (approved):
Looks good, a simple test mod might be handy.

PR #1179, opened 2020-11-23
Assign a new ID to client only registry entries when syncing.

### LINKED ISSUE #1165: Particle type registry being completely overwritten on client
I am not sure if I'm at the right place and if this truly is a fabric issue, I am also making a lot of assumptions on how stuff works (especially registry synchronization), so please correct me if you notice something off.

### Original issues:
https://github.com/RandomMcSomethin/fallingleaves/issues/2#issuecomment-724280523
https://github.com/paulevsGitch/BetterEnd/issues/14

### Description/Recap:

Falling Leaves is a simple mod that adds custom particles that fall from leaves.
It is intended to work fully client side and it does work both in singleplayer and multiplayer - until BetterEnd is installed.

When BetterEnd is installed on the server, the falling leaves don't appear in multiplayer - singleplayer is fine.

What happens in multiplayer is that the leaves get added to the world, but the particles are never actually created because the custom particle type does not exist in the particle type registry anymore, despite being registered beforehand.

I tried changing the falling leaves' rawId to 126 to ensure it wouldn't be overwritten due to synchronization, but it appears the particle type registry is being completely _replaced_ and thus the lookup of the associated particle factory fails.

It turns out that commenting out these two lines:
https://github.com/paulevsGitch/BetterEnd/blob/70f5e3e02457cc0e5dafea8fede12d05c0f26064/src/main/java/ru/betterend/world/biome/BiomeMegalakeGrove.java#L19
https://github.com/paulevsGitch/BetterEnd/blob/c3a7a59d785bddc51523e2f842ce13ed2882619d/src/main/java/ru/betterend/world/biome/BiomeFoggyMushroomland.java#L21
'fixes' the issue.
Note that BetterEnd adds other biomes that use `.setParticles` (which ultimately sets `BiomeEffects.particleConfig`) but they use vanilla particle types and that has no effect on falling leaves.

### Interpretation/Guesswork:

My guess is that since BetterEnd adds biomes with custom particles , the server-side particle type registry is changed and then synchronized with the client.
Unexpectedly, this synchronization doesn't seem to merge the server registry into the client one (by overwriting or remapping ids) but it seems to completely replace it.
I think this means that this breaks any client-side mod adding particles.
Potentially, this could apply to some other registries as well(?)

### Potential fixes

In this case, Falling Leaves would probably be okay if it registered particle types after the server-client synchronization, though I have no idea if this is (currently) possible and it might cause other issues.

A 'proper' fix would probably be remapping particle type ids on the client when doing the server-client synchronization.

Thanks for your time reading this mess.

modmuss50:
This fixes #1165

The old behaviour was to silently remove the registry entry from the registry. This now assigns a new ID at the end of the registry for the entries.

Thanks to @Fourmisain for pointing me in the right direction with this issue in #1165

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdRegistry.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdRegistry.java
@@ -271,19 +271,26 @@ public abstract class MixinIdRegistry<T> extends Registry<T> implements Remappab
 			break;
 		}
 		case REMOTE: {
+			int maxId = -1;
+
 			for (Identifier id : getIds()) {
 				if (!remoteIndexedEntries.containsKey(id)) {
-					int newId = remoteIndexedEntries.values().stream()
-							.mapToInt(Integer::intValue)
-							.max()
-							.orElse(-1);
-
-					if (newId < 0) {
-						throw new RuntimeException("Failed to assign new id to client only registry entry");
+					if (maxId < 0) {
+						for (int value : remoteIndexedEntries.values()) {
+							if (value > maxId) {
+								maxId = value;
+							}
+						}
 					}
 
-					FABRIC_LOGGER.debug("An ID for {} was not sent by the server, assuming client only registry entry and assigning a new id: {}", id.toString(), newId + 1);
-					remoteIndexedEntries.put(id, newId + 1);
+					if (maxId < 0) {
+						throw new RemapException("Failed to assign new id to client only registry entry");
+					}
+
+					maxId++;
+
+					FABRIC_LOGGER.debug("An ID for {} was not sent by the server, assuming client only registry entry and assigning a new id ({}) in {}", id.toString(), maxId, getKey().getValue().toString());
+					remoteIndexedEntries.put(id, maxId);
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


modmuss50:
This might not be an ideal solution, It might not be a bad idea to have an API to mark specfic registry entries as client only, or use the existing registry attributes to mark the registry as client only compatible.

Player3324 (approved):
Looks correct to me

Fourmisain:
Just tested it, I can confirm it fixes the issue!

modmuss50:
Awesome, thanks a lot. I will prob get this out and release tomorrow if all goes well.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdRegistry.java
```diff
@@ -273,25 +271,22 @@ public void remap(String name, Object2IntMap<Identifier> remoteIndexedEntries, R
...
-					T object = get(id);
-					int rid = getRawId(object);
+					int newId = remoteIndexedEntries.values().stream()
+							.mapToInt(Integer::intValue)
+							.max()
+							.orElse(-1);
 
-					droppedIds.add(id);
+					if (newId < 0) {
+						throw new RuntimeException("Failed to assign new id to client only registry entry");
```
i509VCB: Could you include which registry entry failed to be reassigned?

i509VCB (approved):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinIdRegistry.java
i509VCB: Also should this throw a `RemapException` instead?

PR #487, opened 2020-01-12
Start on fabric-crafting-v1

modmuss50:
# fabric-crafting-v1
This PR covers 2 commonly asked for things:

## ItemCraftCallback 
ItemCraftCallback gets called when an item is crafted, this allows for the output stack to be manipulated.

## FabricRecipeRemainder
FabricRecipeRemainder is a stack based recipe remainder, this is most commonly used to allow items to be damaged when crafted.

Both of these apis have a nullable PlayerEntity, as these apis can be invoked via automated machines (Autocrafting table for example).

## Current issues with the PR
Shift clicking an item out of the crafting table fires the event, but does not maintain the modified NBT.

If there is anything else around this area that may be a good fit for this module let me know. 

Thanks in advance for the reviews.

PR #1180, opened 2020-11-24
Add an item setting for better recipe remainders

### LINKED ISSUE #50: Recipe remainders
1) There is no way to control the construction of ItemStacks for recipe remainders. This could be useful for recipes where items take damage during crafting.

2) There is no non-hacky way to make an item have itself as a crafting remainder.

A Function<ItemStack, ItemStack> as part of the Item.Builder might be a good idea.

BoogieMonster1O1:
Adds a conditional recipe remainder (that returns an ItemStack) to FabricItemSettings
also fixes an issue with duplicate keys in the fabric item api v1 testmod fabric.mod.json

gravel
sand

Fixes #50


i509VCB:
I guess I'll be the one to ask, where did the textures for the hammers come from?

BoogieMonster1O1:
> I guess I'll be the one to ask, where did the texture~~s~~ for the hammers come from?

from gimp
i made it for a mod im making
https://github.com/BoogieMonster1O1/ex/blob/main/ex-nihilo-superesse/src/main/resources/assets/ex-nihilo-superesse/textures/item/netherite_hammer.png

Shnupbups (approved):

i509VCB:
This kinda sounds like what would supersede #487 I believe?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.minecraft.inventory.Inventory;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.RecipeType;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@FunctionalInterface
+public interface RecipeRemainderProvider {
```
i509VCB, line 39: The javadoc on the interface should probably explain a common use case for this.

BoogieMonster1O1:
> This kinda sounds like what would supersede #487 I believe?

yes but only the recipe remainder part

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,41 @@
...
+	 * An {@link ItemStack} aware version of {@link Item#getRecipeRemainder()}.
+	 *
+	 * @param original The input item stack.
+	 * @param inventory The inventory that the stack is in.
+	 * @param type The recipe type being used.
+	 * @param world The world in which the inventory is in.
+	 * @param pos The position at which the inventory is.
+	 * @return
+	 */
+	ItemStack getRecipeRemainder(ItemStack original, Inventory inventory, RecipeType<?> type, World world, @Nullable BlockPos pos);
```
i509VCB: Is there a case where a player may be present like in #487

BoogieMonster1O1: there is no exposed PlayerEntity in AbstractFurnaceBlockEntity or RecipeManager

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Allows an item to conditionally specify the recipe remainder.
+ * The recipe remainder is an {@link ItemStack} instead of an {@link Item}.
+ * This can be used to allow your item to get damaged instead of
+ * getting removed when used in crafting.
+ *
+ * <p>Recipe remainder providers can be set with {@link FabricItemSettings#recipeRemainder(RecipeRemainderProvider)}</p>
+ */
+@FunctionalInterface
+public interface RecipeRemainderProvider {
```
Juuxel, line 39: Is this exposed somewhere for custom crafting devices like modded crafting tables or processing machines?

BoogieMonster1O1: yes, provided that the modded crafting table calls `RecipeManager#getRemainingStacks` to get recipe remainders

BoogieMonster1O1: should probably add a method to get the recipe remainder of an ItemStack

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+	 * vanilla recipe remainder is used.
+	 *
+	 * @param original The input item stack.
+	 * @param type The recipe type being used.
+	 * @param inventory The inventory that the stack is in.
+	 * @param world The world in which the inventory is in.
+	 * @param pos The position at which the inventory is.
+	 * @return the recipe remainder
+	 */
+	static ItemStack getRecipeRemainder(ItemStack original, RecipeType<?> type, Inventory inventory, World world, @Nullable BlockPos pos) {
```
modmuss50: Is there ever a case where it would be useful to have the RecipeType nullable?

BoogieMonster1O1: brewing, I guess

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,44 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+		return this.recipeRemainder((original, inventory, type, world, pos) -> {
+			if (!original.isDamageable()) {
+				return ItemStack.EMPTY;
+			}
+
+			ItemStack copy = original.copy();
+			copy.damage(by, ThreadLocalRandom.current(), null);
+
+			if (copy.getDamage() >= copy.getMaxDamage()) {
+				return ItemStack.EMPTY;
```
OroArmor, line 96: What if the item returns a broken value for it (this might be out of the scope for this pr, but its something that might be needed). ie if my hammer becomes a broken hammer or two sticks.

OroArmor (changes requested):
Looks good. I would change the name for the test hammer image, but that's a minor difference. The only other thing I would look at is what happens when the item breaks, and if someone wants to replace it with a different item(stack) as that allows for more customization for recipes. As well, a way to make it so that there is a simple method for returning the same item might be nice, as it doesn't require implementing a `RecipeRemainderProvider` for newer modders and for convenience.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,44 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
 
+	/**
+	 * Sets the stack-aware recipe remainder provider of the item.
+	 * @see RecipeRemainderProvider
+	 */
+	public FabricItemSettings recipeRemainder(RecipeRemainderProvider provider) {
+		FabricItemInternals.computeExtraData(this).recipeRemainderProvider(provider);
+		return this;
+	}
+
```
OroArmor, line 63:
```suggestion

/**
 * When crafted the used ItemStack returns itself, with no modification.
 */
public FabricItemSettings recipeReturnSelf() {
    return this.recipeRemainder((original, inventory, type, world, pos) -> original);
}
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
BoogieMonster1O1: should be up to the implementor to do that

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,55 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+			if (by == 0) {
+				return original;
+			}
+
+			if (!original.isDamageable()) {
+				return ItemStack.EMPTY;
+			}
+
+			ItemStack copy = original.copy();
+			copy.damage(by, ThreadLocalRandom.current(), null);
```
liach: Just make sure this doesn't cause game behavior changes.

Sollace: I assume that if no one calls this method, it won't take any affect, no?

Sollace:
```suggestion
			if (copy.damage(by, ThreadLocalRandom.current(), null)) {
						return ItemStack.EMPTY;
			}
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,55 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+
+			if (copy.getDamage() >= copy.getMaxDamage()) {
```
Sollace:
```suggestion
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,55 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+				return ItemStack.EMPTY;
```
Sollace, line 96:
```suggestion
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +55,55 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+			}
```
Sollace, line 97:
```suggestion
```

Mysterious-Dev (approved):

LambdAurora (approved):
Seems good.

Little nitpick: `@param` javadocs shouldn't start with capitalized letter and end with a comma like `@return` javadoc.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +53,54 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
+	}
+
+	/**
+	 * Sets the stack-aware recipe remainder to damage the item by a certain amount every time it is used in crafting.
+	 *
+	 * @param by the amount
+	 */
+	public FabricItemSettings damageIfUsedInCrafting(int by) {
+		return this.recipeRemainder((original, inventory, type, world, pos) -> {
+			if (by == 0) {
```
liach: If you can move this `by == 0` check outside of the lambda creation (i.e. create different lambdas when by is different values), we can have a performance boost for the `by == 0` special case lambdas.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +53,54 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
 		return this;
 	}
 
+	/**
+	 * Sets the stack-aware recipe remainder provider of the item.
+	 * @see RecipeRemainderProvider
```
liach:
```suggestion
```
No need, already present in parameter types imo

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+
+import net.fabricmc.fabric.impl.item.ItemExtensions;
+
+/**
+ * Allows an item to conditionally specify the recipe remainder.
+ * The recipe remainder is an {@link ItemStack} instead of an {@link Item}.
+ * This can be used to allow your item to get damaged instead of
+ * getting removed when used in crafting.
+ *
+ * <p>Recipe remainder providers can be set with {@link FabricItemSettings#recipeRemainder(RecipeRemainderProvider)}</p>
```
liach:
```suggestion
 * <p>Recipe remainder providers can be set with {@link FabricItemSettings#recipeRemainder(RecipeRemainderProvider)}.</p>
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+@FunctionalInterface
+public interface RecipeRemainderProvider {
+	/**
+	 * An {@link ItemStack} aware version of {@link Item#getRecipeRemainder()}.
+	 *
+	 * @param original The input item stack.
+	 * @param inventory The inventory that the stack is in.
+	 * @param type The recipe type being used.
+	 * @param world The world in which the inventory is in.
+	 * @param pos The position at which the inventory is.
```
liach:
```suggestion
	 * @param original The input item stack.
	 * @param inventory the inventory that the stack is in
	 * @param type the recipe type being used
	 * @param world the world in which the inventory is in
	 * @param pos the position at which the inventory is
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+	 * @return the recipe remainder
+	 */
+	ItemStack getRecipeRemainder(ItemStack original, Inventory inventory, @Nullable RecipeType<?> type, World world, @Nullable BlockPos pos);
+
+	/**
+	 * Returns the recipe remainder of an item stack.
+	 * If the item's recipe remainder provider is null, the
```
liach:
```suggestion
	 * If the item's recipe remainder provider is {@code null}, the
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+	 * vanilla recipe remainder is used.
+	 *
+	 * @param original The input item stack.
+	 * @param type The recipe type being used.
+	 * @param inventory The inventory that the stack is in.
+	 * @param world The world in which the inventory is in.
+	 * @param pos The position at which the inventory is.
```
liach:
```suggestion
	 * @param original the input item stack.
	 * @param type the recipe type being used
	 * @param inventory the inventory that the stack is in
	 * @param world the world in which the inventory is in
	 * @param pos the position at which the inventory is
```

liach (approved):
No critical problems as I can see.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/RecipeRemainderProvider.java
BoogieMonster1O1: Fixed

Prospector (approved):

heftymouse (approved):

Ueaj-Kerman (approved):

Technici4n:
Question: do we need to have custom behavior depending on whether the `RecipeRemainderProvider` is invoked for:
1. A furnace fuel. (A lot of mods will want to use this too);
2. A crafting table or 2x2 crafting inventory. (Some mods will want to use this too);
3. A brewing stand input. (Some mods might want to use this too?)?

I am pretty sure we do, and I believe the arguments could make that more explicit. Maybe we could have an enum for context? Or a `RemainderContext` interface if we want to be able to change it in the future?
```java
public interface RemainderContext {
    boolean isCrafting();
    boolean isBrewing();
    boolean isFuel();
    // throws if the context is not crafting
    Recipe<?> craftingRecipe();
}
```

Also, do we really need to have `World` and `BlockPos`? It is not because MC provides them that we should have them.

It's also not clear if the original `ItemStack` may be mutated or not, you should definitely add that to the javadoc.

(I thought about making this an event originally, but this can be added independently later, here we are just extending MC's recipe remainder system to make it more flexible and I think an event would be out of scope).

Technici4n:
Do we really need the non-`ItemStack` parameters? Imo the cases where you need access to the inventory or the player are better handled using a custom recipe type.

Sollace:
> ```java
>     boolean isCrafting();
>     boolean isBrewing();
>     boolean isFuel();
> ```

Why do we need these methods again that can't already be determined using `ctx.getRecipeType() == RecipeType.CRAFTING`?

Technici4n:
Last time I checked brewing and fuels have no recipe type...

But tbh I would really prefer an ItemStack extension to `getRecipeRemainder`: `getRecipeRemainder(ItemStack stack)`, ideally with the added context as well if the behavior needs to be different in these 3 (and possibly more in the future) cases.

Technici4n:
I can tolerate anything that needs to vary on a) the item stack and b) possibly the context, but if it needs to vary based on the actual recipe it should probably be using a custom recipe type... In general, do we even need this for recipes, or are custom recipe types enough?

Sollace:
> 
> 
> I can tolerate anything that needs to vary on a) the item stack and b) possibly the context, but if it needs to vary based on the actual recipe it should probably be using a custom recipe type... In general, do we even need this for recipes, or are custom recipe types enough?

I think custom recipe types are enough for most use cases.

If you're using this method, chances are it's because you want the behaviour applied to _all_ recipes of a specific kind (crafting/smoking/etc) involving your item. Otherwise you would just make a custom recipe type to use where needed.

NetUserGet (approved):
Good stuffs :>

ZestyBlaze:
Is it bad how i actually kind of agree with Technician that having a function in the Item class that extends ItemStack and returns the stack would probably be a lot easier and tidier then doing all that code in creating an item since that's a setting not a function we can override?

kwpugh:
Question:  Is this going to get incorporated into 1.18.x ?

Technici4n:
Superseded by #2464.

PR #1181, opened 2020-11-25
Add item setting for bundle occupancy

### LINKED ISSUE #1167: Add item setting for bundle item occupancy
This shouldn't be too hard as it would just be an inject at head of `BundleItem#getItemOccupancy(ItemStack)`

BoogieMonster1O1:
Adds an item setting for bundle occupancy to FabricItemSettings. 

Fixes #1167

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/BundleOccupancyProvider.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/BundleOccupancyProvider.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.ItemStack;
+
+/**
+ * Calculates the amount of space an item stack occupies inside of a {@link net.minecraft.item.BundleItem bundle}.
+ *
+ * <p>A bundle's maximum occupancy is 64 which corresponds to the maximum stack size of a normal item.
+ * If an item's stack size is not 64, then the occupancy of that specific item is scaled by {@code 64 / item.getMaxStackSize()}.
+ * If a bundle contains another bundle, then the occupancy of the bundle is 4 plus the current occupancy of the inner bundle.
+ *
+ * <p>Bundle occupancy providers can be set with {@link FabricItemSettings#bundleOccupancy(BundleOccupancyProvider)}.
+ *
+ * @deprecated Experimental feature, may be removed or changed without further notice: Snapshot feature.
+ */
+@ApiStatus.Experimental
+@Deprecated
+@FunctionalInterface
+public interface BundleOccupancyProvider {
+	/**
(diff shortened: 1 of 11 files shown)


i509VCB:
Well it looks like you beat me to opening the PR. Mine is here for reference on docs if you want some (https://github.com/i509VCB/fabric/tree/ft/bundle-item-weight)

There are a few things you need to probably suffice for, such as using a specific interface over a `ToIntFunction` and some javadoc explaining how the bundle weight is calculated to some extent

BoogieMonster1O1:
> Well it looks like you beat me to opening the PR. Mine is here for reference on docs if you want some (https://github.com/i509VCB/fabric/tree/ft/bundle-item-weight)
> 
> There are a few things you need to probably suffice for, such as using a specific interface over a `ToIntFunction` and some javadoc explaining how the bundle weight is calculated to some extent

Thanks! I've added your docs

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -16,11 +16,16 @@
 
 package net.fabricmc.fabric.api.item.v1;
 
+import java.util.function.ToIntFunction;
```
i509VCB: just pop that one import off since it is not used anymore

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
```diff
@@ -17,9 +17,14 @@
 package net.fabricmc.fabric.impl.item;
 
 import java.util.WeakHashMap;
+import java.util.function.ToIntFunction;
```
i509VCB: ditto

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ItemExtensions.java
```diff
@@ -16,12 +16,21 @@
 
 package net.fabricmc.fabric.impl.item;
 
+import java.util.function.ToIntFunction;
```
i509VCB: ditto

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BundleItemMixin.java
```diff
@@ -0,0 +1,41 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.item.BundleItem;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.impl.item.ItemExtensions;
+
+@Mixin(BundleItem.class)
+public class BundleItemMixin {
```
i509VCB: can this be abstract and package private?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/BundleOccupancyProvider.java
```diff
@@ -0,0 +1,45 @@
...
+@Deprecated
+@FunctionalInterface
+public interface BundleOccupancyProvider {
+	/**
+	 * Gets the occupancy size of the item stack.
+	 *
+	 * @param stack the item stack
+	 * @return the occupancy size of the item stack.
+	 */
+	int getBundleOccupancy(ItemStack stack);
```
i509VCB, line 44: I can't recall off head, does the bundle logic here assume the item stack is 1 and then later the count is multiplied onto the value? Could you check on that?

BoogieMonster1O1: the bundle occupancy returned here isn't multiplied with the stack size

i509VCB: Hmm must have remembered something different then

oliviathevampire (approved):

Mysterious-Dev (approved):

Mysterious-Dev:
So is there any change?

BoogieMonster1O1:
> So is there any change?

none *for now*

PR #1182, opened 2020-11-25
Add WorldRenderEvents

grondag:
`WorldRenderer.render` is complicated and extensive but many mods need to target it for rendering.  This PR makes injection into world rendering more convenient and promotes compatibility.

It also solves the problem of Canvas or a similar mod replacing all or large chunks of vanilla code and thereby breaking mixin targets. A mod that breaks the mixin targets simply needs to invoke the event handlers itself at the appropriate time.

These events are based on my experience of adding Canvas compatibility hooks for about a dozen mods via reflection hacks.  (Sadly necessary because these events don't exist yet.) Many common use cases should be covered.

These events are also a useful extension point for third-party renderers.  For example, when Canvas is active  the `VertexConsumerProvider` instance exposed in the event context can be cast to an extended interface with additional functionality.

That said, these events are self-contained and have no dependency on Indigo, Canvas or the renderer API more generally.

### DIFF
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -141,7 +141,7 @@ public interface WorldRenderContext {
 		 * will be {@code false} in later-stage event invocations. It (obviously)
 		 * also prevents the default outline render from running.
 		 *
-		 * <p>In {@code BLOCK_REDER} this prevents the default outline render
+		 * <p>In {@code BLOCK_RENDER} this prevents the default outline render
 		 * from running and {@link AfterBlockOutline#didCancelDefaultBlockOutline()}
 		 * will be {@code true} in later-stage event invocations.
 		 *
@@ -156,7 +156,7 @@ public interface WorldRenderContext {
 
 	/**
 	 * Used in {@code BLOCK_OUTLINE} to convey the parameters normally sent to
-	 * {@code Worldrenderer.drawBlockOutline}.
+	 * {@code WorldRenderer.drawBlockOutline}.
 	 */
 	public interface BlockOutline {
 		VertexConsumer vertexConsumer();
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -35,10 +35,14 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Matrix4f;
 import net.minecraft.util.profiler.Profiler;
 
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
 /**
  * Except as noted below, the properties exposed here match the parameters passed to
  * {@link WorldRenderer#render(MatrixStack, float, long, boolean, Camera, GameRenderer, LightmapTextureManager, Matrix4f)}.
  */
+@Environment(EnvType.CLIENT)
 public interface WorldRenderContext {
 	/**
 	 * The world renderer instance doing the rendering and invoking the event.
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -19,7 +19,6 @@ package net.fabricmc.fabric.api.client.rendering.v1;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.block.BlockState;
-import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.render.Camera;
 import net.minecraft.client.render.Frustum;
 import net.minecraft.client.render.GameRenderer;
@@ -30,7 +29,6 @@ import net.minecraft.client.render.WorldRenderer;
 import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.client.world.ClientWorld;
 import net.minecraft.entity.Entity;
-import net.minecraft.util.hit.HitResult;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Matrix4f;
 import net.minecraft.util.profiler.Profiler;
@@ -94,75 +92,34 @@ public interface WorldRenderContext {
 	 */
 	boolean advancedTranslucency();
 
-	public interface Drawable {
-		/**
-		 * The {@code VertexConsumerProvider} instance being used by the world renderer for most non-terrain renders.
-		 * Generally this will be better for most use cases because quads for the same layer can be buffered
-		 * incrementally and then drawn all at once by the world renderer.
-		 *
-		 * <p>IMPORTANT - all vertex coordinates sent to consumers should be relative to the camera to
-		 * be consistent with other quads emitted by the world renderer and other mods.  If this isn't
-		 * possible, caller should use a separate "immediate" instance.
-		 *
-		 * <p>This property is not exposed before {@link WorldRenderEvents#BEFORE_ENTITIES} or after
-		 * {@link WorldRenderEvents#BEFORE_DEBUG_RENDER} because the consumer buffers are not available before or
-		 * drawn after that in vanilla world rendering.  Renders that cannot draw in one of the supported events
-		 * must be drawn directly to the frame buffer, preferably in {@link WorldRenderEvents#LAST} to avoid being
-		 * overdrawn or cleared.
-		 */
-		VertexConsumerProvider consumers();
-	}
+	/**
+	 * The {@code VertexConsumerProvider} instance being used by the world renderer for most non-terrain renders.
+	 * Generally this will be better for most use cases because quads for the same layer can be buffered
+	 * incrementally and then drawn all at once by the world renderer.
+	 *
+	 * <p>IMPORTANT - all vertex coordinates sent to consumers should be relative to the camera to
+	 * be consistent with other quads emitted by the world renderer and other mods.  If this isn't
+	 * possible, caller should use a separate "immediate" instance.
+	 *
+	 * <p>This property is {@code null} before {@link WorldRenderEvents#BEFORE_ENTITIES} and after
+	 * {@link WorldRenderEvents#BEFORE_DEBUG_RENDER} because the consumer buffers are not available before or
+	 * drawn after that in vanilla world rendering.  Renders that cannot draw in one of the supported events
+	 * must be drawn directly to the frame buffer, preferably in {@link WorldRenderEvents#LAST} to avoid being
+	 * overdrawn or cleared.
+	 */
+	@Nullable VertexConsumerProvider consumers();
 
 	/**
-	 * Conveys WorldRenderer state to event handlers after invocation.
-	 * Includes view frustum, which is not available then.
+	 * View frustum, after it is initialized. Will be {@code null} during
+	 * {@link WorldRenderEvents#START}.
 	 */
-	public interface AfterFrustum {
-		Frustum frustum();
-	}
-
-	/**
-	 * Used in the block breaking callback to convey information on
-	 * ray trace hit and accept cancellation of default outline render.
-	 */
-	public interface PreBlockOutline {
-		/**
-		 * The game object currently under the crosshair target. Normally
-		 * equivalent to {@link MinecraftClient#crosshairTarget}.
-		 * Provided for convenience.
-		 */
-		@Nullable HitResult hitResult();
-
-		/**
-		 * Call to prevent the default block outline from rendering.
-		 * Use this for event handlers that fully replace the vanilla
-		 * block outline render.
-		 *
-		 * <p>The effect of this method depends on the event from which it is called:
-		 *
-		 * <p>In {@code BEFORE_BLOCK_RENDER} this prevents invocation of
-		 * the {@code BLOCK_RENDER} event and {@link AfterBlockOutline#didCancelDefaultBlockOutline()}
-		 * will be {@code false} in later-stage event invocations. It (obviously)
-		 * also prevents the default outline render from running.
-		 *
-		 * <p>In {@code BLOCK_RENDER} this prevents the default outline render
-		 * from running and {@link AfterBlockOutline#didCancelDefaultBlockOutline()}
-		 * will be {@code true} in later-stage event invocations.
-		 *
-		 * <p>In both cases, this method is idempotent - calling it from more
-		 * than one event subscriber has the same effect as calling it once.
-		 * All event handlers within the same event invocation will always
-		 * execute and the cancellation status cannot be inspected during the
-		 * block outline render events.
-		 */
-		void cancelDefaultBlockOutline();
-	}
+	@Nullable Frustum frustum();
 
 	/**
 	 * Used in {@code BLOCK_OUTLINE} to convey the parameters normally sent to
 	 * {@code WorldRenderer.drawBlockOutline}.
 	 */
-	public interface BlockOutline {
+	public interface BlockOutlineContext {
 		VertexConsumer vertexConsumer();
 
 		Entity entity();
@@ -177,31 +134,4 @@ public interface WorldRenderContext {
 
 		BlockState blockState();
 	}
-
-	/**
-	 * Used after the block outline events to test if {@link #cancelDefaultBlockOutline()}
-	 * was called during the {@code BLOCK_OUTLINE} event.
-	 */
-	public interface AfterBlockOutline {
-		/**
-		 * Tests if the default block outline render was cancelled in the {@code BLOCK_OUTLINE} event.
-		 *
-		 * <p>Will NOT be true if the outline was cancelled during {@code BEFORE_BLOCK_OUTLINE}.
-		 *
-		 * @return {@code true} if the default block outline render was cancelled in the {@code BLOCK_OUTLINE} event.
-		 */
-		boolean didCancelDefaultBlockOutline();
-	}
-
-	public interface FrustumContext extends WorldRenderContext, AfterFrustum { }
-
-	public interface MainContext extends FrustumContext, Drawable { }
-
-	public interface PreBlockOutlineContext extends MainContext, PreBlockOutline { }
-
-	public interface BlockOutlineContext extends PreBlockOutlineContext, BlockOutline { }
-
-	public interface PostBlockOutlineContext extends MainContext, AfterBlockOutline { }
-
-	public interface LateContext extends FrustumContext, AfterBlockOutline { }
 }
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.rendering.v1;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.render.Camera;
+import net.minecraft.client.render.Frustum;
+import net.minecraft.client.render.GameRenderer;
+import net.minecraft.client.render.LightmapTextureManager;
+import net.minecraft.client.render.VertexConsumer;
+import net.minecraft.client.render.VertexConsumerProvider;
+import net.minecraft.client.render.WorldRenderer;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.client.world.ClientWorld;
+import net.minecraft.entity.Entity;
+import net.minecraft.util.hit.HitResult;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Matrix4f;
+import net.minecraft.util.profiler.Profiler;
+
+/**
+ * Except as noted below, the properties exposed here match the parameters passed to
+ * {@link WorldRenderer#render(MatrixStack, float, long, boolean, Camera, GameRenderer, LightmapTextureManager, Matrix4f)}.
+ */
+public interface WorldRenderContext {
+	/**
+	 * The world renderer instance doing the rendering and invoking the event.
+	 *
+	 * @return WorldRenderer instance invoking the event
+	 */
+	WorldRenderer worldRenderer();
+
+	MatrixStack matrixStack();
+
+	float tickDelta();
+
+	long limitTime();
+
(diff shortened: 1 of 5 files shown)


fabric-rendering-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/MixinWorldRenderer.java
```diff
@@ -17,16 +17,152 @@
...
+				value = "INVOKE",
+				target = "Lnet/minecraft/client/render/WorldRenderer;renderChunkDebugInfo(Lnet/minecraft/client/render/Camera;)V"
+			)
+	)
+	private void onChunkDebugRender(CallbackInfo ci) {
+		WorldRenderEvents.LAST.invoker().onLast(context);
+	}
+
+	@Inject(method = "render", at = @At("RETURN"))
+	private void afternRender(CallbackInfo ci) {
```
ramidzkh: Typo

ramidzkh (approved):
I'm no rendering expert but the API and default implementation looks good from what I've seen

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+		 * block outline render.
+		 *
+		 * <p>The effect of this method depends on the event from which it is called:
+		 *
+		 * <p>In {@code BEFORE_BLOCK_RENDER} this prevents invocation of
+		 * the {@code BLOCK_RENDER} event and {@link AfterBlockOutline#didCancelDefaultBlockOutline()}
+		 * will be {@code false} in later-stage event invocations. It (obviously)
+		 * also prevents the default outline render from running.
+		 *
+		 * <p>In {@code BLOCK_REDER} this prevents the default outline render
```
BoogieMonster1O1: typo

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+		 * All event handlers within the same event invocation will always
+		 * execute and the cancellation status cannot be inspected during the
+		 * block outline render events.
+		 */
+		void cancelDefaultBlockOutline();
+	}
+
+	/**
+	 * Used in {@code BLOCK_OUTLINE} to convey the parameters normally sent to
+	 * {@code Worldrenderer.drawBlockOutline}.
```
pluiedev: Typo :)

pluiedev (commented):
looks good :)

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/MixinWorldRenderer.java
```diff
@@ -17,16 +17,152 @@
...
+import net.fabricmc.fabric.impl.client.rendering.WorldRenderContextImpl;
 
 @Mixin(WorldRenderer.class)
 public abstract class MixinWorldRenderer {
+	@Shadow private BufferBuilderStorage bufferBuilders;
+	@Shadow private ClientWorld world;
+	@Shadow private ShaderEffect transparencyShader;
+	@Shadow private MinecraftClient client;
+	private final WorldRenderContextImpl context = new WorldRenderContextImpl();
+	private boolean didRenderParticles;
```
modmuss50: Add @Unique on added fields.

modmuss50 (approved):
LGTM, just one minor thing with the mixin I noticed.

👍

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
```diff
@@ -0,0 +1,268 @@
...
+	 * with mod-specific outline renders that cannot be well-handled by a
+	 * general-purpose outline renderer.
+	 *
+	 * <p>Default block outline rendering does not have to happen here.  If a custom
+	 * default outline effect benefits from being drawn at a later stage it can
+	 * be cancelled here and then drawn in {@link #AFTER_TRANSLUCENT} or {@link #LAST}
+	 * if desired. To facilitate this pattern, those events can check
+	 * {@link WorldRenderContext.AfterBlockOutline#didCancelDefaultBlockOutline()}.
+	 */
+	public static final Event<BlockOutline> BLOCK_OUTLINE = EventFactory.createArrayBacked(BlockOutline.class, callbacks -> context -> {
```
i509VCB: AfterBlockOutline?

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+import net.minecraft.util.hit.HitResult;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Matrix4f;
+import net.minecraft.util.profiler.Profiler;
+
+/**
+ * Except as noted below, the properties exposed here match the parameters passed to
+ * {@link WorldRenderer#render(MatrixStack, float, long, boolean, Camera, GameRenderer, LightmapTextureManager, Matrix4f)}.
+ */
+public interface WorldRenderContext {
```
i509VCB, line 44: Could all the classes be marked as client only? `@Environment(EnvType.CLIENT)`

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+	/**
+	 * The world renderer instance doing the rendering and invoking the event.
+	 *
+	 * @return WorldRenderer instance invoking the event
+	 */
+	WorldRenderer worldRenderer();
```
i509VCB, line 50: I think the naming should be more explicit being like `getWorldRenderer`.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+	 */
+	public interface AfterBlockOutline {
+		/**
+		 * Tests if the default block outline render was cancelled in the {@code BLOCK_OUTLINE} event.
+		 *
+		 * <p>Will NOT be true if the outline was cancelled during {@code BEFORE_BLOCK_OUTLINE}.
+		 *
+		 * @return {@code true} if the default block outline render was cancelled in the {@code BLOCK_OUTLINE} event.
+		 */
+		boolean didCancelDefaultBlockOutline();
```
i509VCB: `shouldCancelDefaultBlockOutline`

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	public interface FrustumContext extends WorldRenderContext, AfterFrustum { }
```
i509VCB: I feel the `Context` on the end of these nested interfaces may be redundant. At the same time `WorldRenderContext.Frustrum` could be a bit weird, so this needs a bit of an evaluation.

i509VCB: Considering the `FrustrumContext` is only used twice, it may be neater to just specify the method across both interface impls. May want to consider this as to reduce the amount of interfaces sub implementing eachother.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,203 @@
...
+		 * from running and {@link AfterBlockOutline#didCancelDefaultBlockOutline()}
+		 * will be {@code true} in later-stage event invocations.
+		 *
+		 * <p>In both cases, this method is idempotent - calling it from more
+		 * than one event subscriber has the same effect as calling it once.
+		 * All event handlers within the same event invocation will always
+		 * execute and the cancellation status cannot be inspected during the
+		 * block outline render events.
+		 */
+		void cancelDefaultBlockOutline();
```
i509VCB: Player raised a concern here about this context in use mixing up the cancellation + notification handling.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: That's not necessarily what it means.  It means the block outline was cancelled during the BLOCK_OUTLINE event.  The event consumer can do whatever they want with that information, which *might* mean not rendering the block outline event if they  are replacing the default outline renderer, but that could have been handled during the event depending on how it is renderers.  Consumers of this may want it to decide if they should render some kind of additive information, etc.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: I'll need a little more elaboration to understand what the problem is here.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: Not following what "both" refers to.  FrustumContext is the ~~parent~~ super type for every context other than the starting one.

image

grondag: Not a hill I'm prepared to die on, but absent some specific direction leaving as is.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: Ugh. None of these are settable properties.  I'll do it if you make me but I don't see a need for it here.

Even `cancelDefaultBlockOutline` is not a property - it's an action you can take but it can't be inspected in the same way a property could.  Later on there is a read-only property to inspect if the action did occur during a specific event, but to me that's not the same thing.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
grondag: Changed to referr to the context instead of the interface.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/MixinWorldRenderer.java
grondag: Done

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: Done

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: To elaborate a bit more, if `cancelDefaultBlockOutline` is called during the BLOCK_RENDER event then this PR ensures `drawBlockOutline` also exits before the vanilla rendering code runs.  If a mod has done mixins downstream of that to change outline rendering, then they don't need to do anything to cancel their render.  But at least one mod cancels the vanilla outline render in all cases and instead renders it later during world rendering, so they need to know at that later time if it was cancelled.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
Player3324: We want to avoid mixing notification and cancellation in the same event since that'd incorrectly notify some subscribers. If the 2nd subscriber happened to cancel, the 1st one would never know that the notified action won't actually happen with inline cancellation.

So e.g. instead of
```
foreach (subscriber) {
  if (!subscriber.onBeforeX()) return;
}

<vanilla logic>

foreach (subscriber) {
  subscriber.onAfterX();
}
```
it should do
```
boolean cancelled = false;

foreach (subscriber) {
  if (!subscriber.shouldDoX()) {
    cancelled = true;
    break;
  }
}

if (!cancelled) {
  foreach (subscriber) {
    subscriber.onBeforeX();
  }

  <vanilla logic>

  foreach (subscriber) {
    subscriber.onAfterX();
  }
} else {
  foreach (subscriber) {
    subscriber.onXCancelled();
  }
}
```
for an example with the full set of notifications (a real case would only provide those that are reasonable)

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
Player3324: IMO some nullable getters wouldn't hurt, e.g. adding nullable `frustum()` to `WorldRenderContext` allows dropping both `AfterFrustum` and `FrustumContext`. Nullable `consumers` in `WorldRenderContext` further removes `Drawable` and `MainContext`. `AfterBlockOutline` can IMO also be in `WorldRenderContext` without much potential for confusion. `BlockOutline` could be passed to the appropriate event as a separate parameter (even if it has the same backing object).

With all of the above changes we'd have a single context type everywhere, `WorldRenderContext`, at the cost of a few nullable methods that only yield a proper result in some documented scenarios.

`Drawable` (actually `VertexConsumerProvider` then) may be nicer to pass in the event signature similar to what i proposed for `BlockOutline` above.

grondag: I'm fine with using `@Nullable` instead of separate ~~instances~~ interfaces if that is preferable.

I don't like varying the method signature because I think what may be needed by consumers of these events is still likely to change over time and having everything in the context avoids the need for breaking changes and/or hanging on to deprecated event signatures.  The context is also a more robust extension point for Canvas or others that want to go there.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: That helps, thank you.  I'll need to think about how it applies here.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
Player3324: I am not talking about varying the signature everywhere, but passing the big differentiators (BlockOutline and VertexConsumerProvider) directly where it is appropriate, next to the regular context that captures all the generic info.

grondag: VertexConsumerProvider is specifically one I'd like to keep out of the method signature to better support extension use cases where it (along with the context itself) may be sub-typed. I'd prefer to make it nullable.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
grondag: I've considered this some more.  The main problem seems to be the API causes confusion.

First, the behavior is not like this:

> We want to avoid mixing notification and cancellation in the same event since that'd incorrectly notify some subscribers. If the 2nd subscriber happened to cancel, the 1st one would never know that the notified action won't actually happen with inline cancellation.

The block render events cannot cancel an event for any peer subscriber.  Ever.  This is clearly stated in the docs:

> ... [cancelDefaultBlockOutline()] is idempotent - calling it from more than one event subscriber has the same effect as calling it once. **All event handlers within the same event invocation will always execute** and the cancellation status cannot be inspected during the block outline render events.

The "cannot be inspected" is specifically meant to prohibit subscribers from trying to compete with or depend on each other.  The order of calling is not deterministic and they don't directly affect each other, so that would only cause problems.

What it cancels depends on usage. 
* In `BEFORE_BLOCK_RENDER` it prevents the block outline render entirely. `BLOCK_RENDER` will not be called.
* In `BLOCK_RENDER` it prevents the *vanilla* block outline render from running but has no effect on other subscribers.

Let's cover each in more detail.

**`BEFORE_BLOCK_RENDER`**
The primary use here is when the player is looking at something like a fluid or entity that wants to mimic block behavior and display an outline. A fluid or entity would never register as a hit for block outline rendering, so this provides a way to detect those special hits and render if appropriate.  

If the block outline render isn't cancelled here, then depending on circumstance _two_ outlines will be rendered.   The desired behavior is to pretend there was never a block hit in the first place, and so the entire block outline handling is cancelled and `BLOCK_RENDER` never fires, which is exactly what should normally happen if the player isn't targeting a block.

The `BEFORE_BLOCK_RENDER` is also potentially useful for rendering in-world information that doesn't necessarily care about whether or not block outline rendering occurs.  Subscribers with this use case can rely on being called every frame even if the block outline render is cancelled.

What happens if an entity with a special bounding box is standing in a fluid that also has a special bounding box?   Both boxes will render. Any other answer would make this much more elaborate than it needs to be for the vast majority of cases. 

**`BLOCK_RENDER`**
This is called when a conventional block outline render would normally happen - when the player is targeting a nearby block and we aren't pretending otherwise because some mod is faking a hit.

This is where mods that want to replace block outline rendering for specific blocks should subscribe.  Every subscriber gets called, and every mod that wants to can render.  Exotic Blocks, for example, has many non-cubic shapes that would not render well with a conventional outline renderer so it can use this event instead of a mixin to do that.

Again, the subscribers here cannot cancel each other - they can only cancel the default block outline render that would normally come after.

This event is *not* a good solution for mods that want to replace the default outline renderer for *all* blocks.  First, such a mod should probably not render an outline if a subscriber to `BLOCK_RENDER` has cancelled it.  Presumably a more specific outline renderer will do a better job and should be allowed to do it.

I assumed a general-purpose outline mod would not use the `BLOCK_RENDER` event and instead mix in to `drawShapeOutline`, which will not be called if `BLOCK_RENDER` is cancelled. That's a nice, clean target and generally only one mod is going to be able to replace it fully.

However, at least one such mod (Custom Selection Box) doesn't want to render at that point because it is going to employ translucency that wouldn't render correctly until later in the frame.  It is going to `NOOP` the `drawShapeOutline` and render the outline in a later WorldRenderEvent. 

However, such a mod needs a way to know if `drawShapeOutline` was ever called.  It *could* set some state there and then reference it a few microseconds later, but for convenience I added `didCancelDefaultBlockOutline()` to the context so it can check directly.  But that's a false economy because said mod still needs to mix in to `drawShapeOutline` to prevent it from doing anything.

And the larger problem is that the overall scheme around block outline rendering is very confusing.

I think it may be best to simplify the two block outline events and remove `didCancelDefaultBlockOutline()`.  General-purpose block outline replacements can fend for themselves using the usual and customary methods. More robust support for a general-purpose outline replacement would require another event, at least one more mixin, or perhaps some kind of `BlockOutlineRendererPlugIn` API, all of which seems well outside the scope of this PR, and probably outside the scope of Fabric API generally.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
```diff
@@ -213,7 +237,7 @@ private WorldRenderEvents() { }
 	 * down transient state in event handlers or as a hook that precedes hand/held item
 	 * and GUI rendering.
 	 */
-	public static final Event<End> END = EventFactory.createArrayBacked(End.class, callbacks -> context -> {
+	public static final Event<End> END = EventFactory.createArrayBacked(End.class, context -> { }, callbacks -> context -> {
```
i509VCB, line 252: Is there a reason you are specifying the empty invoker impl for the events?

OroArmor (commented):
Looks really good and well documented. It might be useful to provide a list of the calls in order and how canceling can affect the others to help create a better visual picture (in the javadoc for the main API class).

Leo40Git (approved):

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
grondag: Per the Fabric docs, it makes a small performance difference and these events are called every frame and there are several of them.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
```diff
@@ -0,0 +1,317 @@
...
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.fabric.api.client.rendering.v1.WorldRenderContext.BlockOutlineContext;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Mods should use these events to introduce custom rendering during {@link WorldRenderer#render(net.minecraft.client.util.math.MatrixStack, float, long, boolean, net.minecraft.client.render.Camera, net.minecraft.client.render.GameRenderer, net.minecraft.client.render.LightmapTextureManager, net.minecraft.util.math.Matrix4f)}
+ * without adding complicated and conflict-prone injections there.  Using these events also enables 3rd-party renderers
+ * that make large-scale changes to rendering maintain compatibility by calling any broken even invokers directly.
```
Juuxel:
```suggestion
 * that make large-scale changes to rendering maintain compatibility by calling any broken event invokers directly.
```

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -0,0 +1,137 @@
...
+	 * View frustum, after it is initialized. Will be {@code null} during
+	 * {@link WorldRenderEvents#START}.
+	 */
+	@Nullable Frustum frustum();
+
+	/**
+	 * Used in {@code BLOCK_OUTLINE} to convey the parameters normally sent to
+	 * {@code WorldRenderer.drawBlockOutline}.
+	 */
+	public interface BlockOutlineContext {
```
Juuxel, line 123: I don't think the `@Environment` propagates to nested types, so this needs it as well.

Juuxel (changes requested):

OroArmor (approved):

Juuxel (changes requested):
Contents look good, but I think the callback interfaces should have `@Environment(CLIENT)` as well, just to be safe (sorry for not catching this last time).

Juuxel (approved):

dblsaiko (approved):

NeusFear (approved):

Player3324 (approved):
Good, thanks! It'd be nice if the remaining WorldRenderContext methods could be documented before merging.

LambdAurora (approved):

oliviathevampire (approved):

zml2008 (approved):
I've been toying with the previous iteration of this API (`frex-events`) in one of my mods and it's worked well for me so far. Glad to see it coming to fabric API proper :)

i509VCB (approved):

PR #1183, opened 2020-11-25
Allow specifying the render layer in FabricBlockSettings

BoogieMonster1O1:
Adds three methods to allow giving the block a different render layer

The testmod does not crash on a dedicated server
server480

### DIFF
--- a/fabric-object-builder-api-v1/build.gradle
+++ b/fabric-object-builder-api-v1/build.gradle
@@ -7,7 +7,8 @@ dependencies {
 
 moduleDependencies(project, [
 		'fabric-api-base',
-		'fabric-tool-attribute-api-v1'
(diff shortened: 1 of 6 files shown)


modmuss50 (changes requested):
Im not a fan of this, we (and in places vanilla) generally try to keep a good split between the client and server.

frqnny:
I also agree, any client-only things should be done in the client mod initializer. Has this been tested in a dedicated server?

BoogieMonster1O1:
> I also agree, any client-only things should be done in the client mod initializer. Has this been tested in a dedicated server?

i've tested it on a dedicated server. look at the image i posted

BoogieMonster1O1:
probably not gonna get merged considering that people don't want it, so im just gonna close this one

PR #1187, opened 2020-11-28
Item stack clicking interaction events

BoogieMonster1O1:
This adds two interaction events that fire before `ItemStack#onClicked` and `ItemStack#onStackClicked`. Both are cancellable.

here's the testmod (a miniature fluid api 😄) - https://i.imgur.com/2APkSj3.mp4


oliviathevampire (approved):

oliviathevampire:
Seems like you have some unused imports which makes the checks fail

Prospector:
The API looks fine to me, my only suggestion would be to maybe simplify the test mod? It seems a bit complicated and is nearly half of the PR. It's also not entirely clear to me from your video how it is *supposed* to be working. Perhaps some more obvious and simplified tests would be better, and more helpful to future maintainers of this code.

maityyy:
Any updates?

kanpov:
> Any updates?

another abandoned fabric pr in the bag

Issue #1130, opened 2020-10-20
ItemPickup Event


TheCoolestPaul:
I think an OnItemPickup(ItemStack itemStack) would be a nice addition.

ghost:
This definitely seems like a good fit for where fabric api can go in the future, I would agree. In the meantime, however, if you are looking for a way to do this right now it should be as simple as mixing into `ItemEntity.onPlayerCollision` or `LivingEntity.sendPickup` depending upon what your purpose is. If you'd like more help with getting a quick and dirty item pickup event working, join the [fabric discord](https://discord.gg/v6v4pMv) and ask for help in any of the `#mod-dev-1`, `#mod-dev-2`, or `#mod-dev-3` channels.

ghost:
As a little history lesson for anyone trying to turn this into a PR. Addressing this feature was previously by Draylar in #261 and #449, but neither made it into production (however discussion on the implementation in those two cases should probably be considered for a future attempt as well). In terms of events standardization, it has been discussed that events should have an "allow", "before", "after", and "cancelled" callback (see #1121), which I personally think would fit well in this case. Additionally, this could probably be bundled with an item drop event. In terms of organization this would likely go in `fabric-events-interaction-v0`.

TheCoolestPaul:
I did a DIRTY implementation, but it works great for what I needed.
I have it [here](https://github.com/TheCoolestPaul/Pauls-Tweaks-Fabric/blob/totem-of-removal/src/main/java/net/thirdshift/paulstweaks/mixin/PlayerInventoryMixin.java)

boholder:
As far as in 1.19.3, the logic of `ItemEntity.onPlayerCollision` is:

```java
 public void onPlayerCollision(PlayerEntity player) {
        if (!this.world.isClient) { // this always = 0 Air
            ItemStack itemStack = this.getStack();
            Item item = itemStack.getItem(); // Real item
            int i = itemStack.getCount();
            if (this.pickupDelay == 0 && (this.owner == null || this.owner.equals(player.getUuid())) && player.getInventory().insertStack(itemStack)) {
                player.sendPickup(this, i);
                if (itemStack.isEmpty()) {
                    this.discard();
                    itemStack.setCount(i);
                }

                player.increaseStat(Stats.PICKED_UP.getOrCreateStat(item), i); // The item will be directly sent into player's inventory
                player.triggerItemPickedUpByEntityCriteria(this);
            }

        }
    }
```

[...]

PR #1192, opened 2020-11-30
Add ItemPickup Event

Mysterious-Dev:
This PR adds an event which is called when the player picks up an item from the ground.

It relates with this issue : #1130

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/ItemPickupEvent.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/ItemPickupEvent.java
@@ -1,0 +1,29 @@
+package net.fabricmc.fabric.api.event.player;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ActionResult;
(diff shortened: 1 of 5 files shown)


YTG1234:
is this event common enough that it needs to go in FAPI?

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/ItemPickupEvent.java
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.Hand;
+import net.minecraft.util.TypedActionResult;
+import net.minecraft.world.World;
+
+public interface ItemPickupEvent {
+	Event<ItemPickupEvent> EVENT = EventFactory.createArrayBacked(ItemPickupEvent.class,
+			listeners -> (player, itemStack) -> {
```
i509VCB, line 15: I believe the question that held up past attempts was whether entities other than players would handle this

TheGlitch76: If there isn't a consistent, easy way to detect this then there's too much complexity involved, since what are you going to do with modded entities?

Daomephsta: @TheGlitch76 
Modded entities can fire this event themselves, if its modified to not be player-specific.

PR #1196, opened 2020-12-02
Add warning when trying to run Fabric API as an executable

IMS212:
This PR shows this when the jar is run as an executable. This would fix a lot of the confusion around the API, and the current code does not seem to have problems with anything else in the API.
fabricapi


modmuss50:
I don’t believe we are allowed to release this onto curseforge. I’m 99% sure that executable jars are against the TOS. Please correct me if I’m wrong.

IMS212:
I have asked around, and it seems it is in the clear. BetterFPS does this as well and is allowed on curseforge.

OroArmor (approved):

OverlordsIII (approved):

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.launch;
+
+import javax.swing.JOptionPane;
+
+public final class APIWarning {
+	public static void main(String[] args) {
+		String st = "Fabric API is not meant to be run.\nPlease place this file in your mods folder.";
```
liach: Preferably this should be handled by a resource bundle

build.gradle
```diff
@@ -137,11 +137,21 @@ allprojects {
...
 		classifier = "dev"
 	}
 
 	afterEvaluate {
 		remapJar {
+			manifest {
+				attributes 'Implementation-Title': 'FabricAPI',
+						'Implementation-Version': project.version,
+						'Main-Class': 'net.fabricmc.fabric.impl.launch.APIWarning'
+			}
```
liach: You should move this block into `tasks.withType(Jar)` so this implementation info and main class is available for any jar produced

IMS212: I think I've fixed both these, you mind reviewing again?

IMS212:
Wait, what? I didn't remove the request for API developers

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Locale;
+import java.util.ResourceBundle;
+
+import javax.swing.JOptionPane;
+
+public final class APIWarning {
+	public static void main(String[] args) {
+		Locale defaultLocale = Locale.getDefault();
+		ResourceBundle WarningAPI = ResourceBundle.getBundle("lang/WarningAPI", defaultLocale);
+		JOptionPane.showMessageDialog(null, WarningAPI.getString("api.warning"));
```
i509VCB: This will fail if fabric api is run in a headless environment. Just check `GraphicsEnvironment` to validate that otherwise print the message in sys out.

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
```diff
@@ -16,11 +16,15 @@
...
+import java.util.ResourceBundle;
+
 import javax.swing.JOptionPane;
 
 public final class APIWarning {
 	public static void main(String[] args) {
-		String st = "Fabric API is not meant to be run.\nPlease place this file in your mods folder.";
-		JOptionPane.showMessageDialog(null, st);
+		Locale defaultLocale = Locale.getDefault();
+		ResourceBundle WarningAPI = ResourceBundle.getBundle("lang/WarningAPI", defaultLocale);
```
liach: I hope we can have a bundle setup like https://github.com/FabricMC/fabric-installer/blob/e76d587e22ae9767c0f97fc782c991e461881b1c/src/main/java/net/fabricmc/installer/util/Utils.java#L36-L50 so that we don't need to use iso encoding and ugly unicode escapes for the translation properties

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
AlexIIL: Would system error be better? Since it's already an error, and this way it will be highlighted red (in consoles/terminals that treat error differently to out).

IMS212:
It's been updated with both ResourceBundles and headless support, you mind re-reviewing?

OroArmor:
I would also add something to link to the fabric installer, so that people can install the loader if they don't have it yet

IMS212:
Done.

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.PropertyResourceBundle;
+import java.util.ResourceBundle;
+
+import javax.swing.JOptionPane;
+import javax.swing.UIManager;
+import javax.swing.UnsupportedLookAndFeelException;
+
+public final class APIWarning {
+	public static void main(String[] args) {
+		final ResourceBundle WarningAPI = ResourceBundle.getBundle("lang/WarningAPI", Locale.getDefault(), new ResourceBundle.Control() {
```
liach: we prefer name the local var in lower camel like `warningApi`

IMS212: fixed

liach (approved):

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
IMS212: this has been fixed with Alex's PR

YTG1234 (approved):

nnym (approved):

src/main/java/net/fabricmc/fabric/impl/launch/APIWarning.java
```diff
@@ -0,0 +1,67 @@
...
+					if (stream != null) {
+						try (InputStreamReader reader = new InputStreamReader(stream, StandardCharsets.UTF_8)) {
+							return new PropertyResourceBundle(reader);
+						}
+					}
+				} catch (IOException e) {
+					e.printStackTrace();
+				}
+
+					return super.newBundle(baseName, locale, format, loader, reload);
```
williambl: Indentation

liach:
```suggestion
				return super.newBundle(baseName, locale, format, loader, reload);
```

IMS212: Should be good? Unless you mean a different area

Juuxel (approved):

williambl (approved):

teddyxlandlee:
Buddy, this made me laugh :laughing: 

Is this really necessary? I believe 95%+ of players know this should be dragged into `mods` directory. Other mod APIs, including Forge, Quilt & LiteLoader haven't shown up a warning like this.

If players don't know Fabric API is a **mod** instead of an executable library, you can add a tip in the installer, noticing them download Fabric API in CurseForge/Modrinth and put it into `mods` directory. Besides, can't those players who download Fabric API in CurseForge/Modrinth realize such a thing from such a website should be treated as a mod? I think it's not our duty to add a dumb Swing app & make Fabric API fatter and fatter.

LambdAurora:
> @teddyxlandlee 

Accessibility features make you laugh? sad

While, yes, a large user base should know this should be put into the `mods` directory, there's also the question of *new* users.
In one of my most recent user support I've encountered an user who seemed to be a computer novice, which made the user support quite more intense. I don't think being hostile to new users and features facilitating support is a good approach.

This comment honestly comes off as a rather elitist take and hostile towards new users. I would rather not have the same issue as Linux here in Fabric or its alternatives.

apple502j:
This feature looks good, but I believe it should belong in Loom so that everyone can use it (with optout ofc). Also the branch is 1.16, which is way outdated.

teddyxlandlee:
> This feature looks good, but I believe it should belong in Loom so that everyone can use it (with optout ofc). Also the branch is 1.16, which is way outdated.

Great idea!

However, I think the class should be shrunk as small as possible. Using proguard and package-less class name can shrink it in a specific way.

Also, add a boolean toggle in `remapJar`. If it is set to true, the already-shrunk class binary will be dumped into the jar along with `Main-Class` manifest (if not set).

To make it short:  
1. The class may be mod-insensitive, which means Fabric API uses the same class & the same warning text as other mods.  
2. No resource bundles, if possible. Texts are stored inside the class directly. Also, it's a bad idea to let the class trying to understand `fabric.mod.json`.  
3. Loom copies the already-configured class binary into your remapped jar. This means you don't need to compile it again, because the binary has been stored in Loom.

IMS212:
This was made as my first PR ever. I am now a developer for many mods, and can likely recreate this much better, so as such, I'll close this.

IMS212:
> This feature looks good, but I believe it should belong in Loom so that everyone can use it (with optout ofc). Also the branch is 1.16, which is way outdated.

How would it be included in Loom? Would it just embed itself into the jar?

teddyxlandlee:
> How would it be included in Loom? Would it just embed itself into the jar?

To store the class file in Loom, and dumps it into `remapJar` along with `Main-Class` manifest attribute.

PR #1197, opened 2020-12-02
Add registry for creating entity model layers and texture data

i509VCB:
This adds a new experimental logic to allow registration of entity model layers and the corresponding textured model data provided by the model layers

### DIFF
--- a/fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerHelper.java
+++ b/fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerHelper.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.rendereregistry.v1;
+
+import java.util.Objects;
+
+import net.minecraft.client.model.TexturedModelData;
+import net.minecraft.client.render.entity.model.EntityModelLayer;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.fabric.impl.client.renderer.registry.EntityModelLayerImpl;
+import net.fabricmc.fabric.mixin.client.renderer.registry.EntityModelLayersAccessor;
+
+/**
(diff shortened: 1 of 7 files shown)


YanisBft (approved):

oliviathevampire (approved):

modmuss50 (approved):

oliviathevampire:
progress on this?

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityRendererRegistry.java
```diff
@@ -16,6 +16,8 @@
 
 package net.fabricmc.fabric.api.client.rendereregistry.v1;
 
+import net.minecraft.client.render.entity.EntityRenderDispatcher;
+import net.minecraft.client.render.entity.EntityRenderer;
```
liach, line 20: Extra imports.

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerHelper.java
```diff
@@ -0,0 +1,62 @@
...
+@Deprecated
+@Environment(EnvType.CLIENT)
+public final class EntityModelLayerHelper {
+	/**
+	 * Registers an entity model layer and registers a provider for a {@linkplain TexturedModelData}.
+	 *
+	 * @param modelLayer the entity model layer
+	 * @param provider the provider for the textured model data
+	 */
+	public static void registerModelLayer(EntityModelLayer modelLayer, TexturedModelDataProvider provider) {
```
liach: Why not just take a `Supplier<TexturedModelData>`?

liach (changes requested):

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerHelper.java
i509VCB: I believe we are avoiding generics there to be more explicit about the type of functional interface being passed there. I'd say it's similar to event callback interfaces in newer prs.

Also we have been working on a guidelines document for dev (to become CONTRIBUTING) - however it's a new document on hackmd so you can't really see it unless I DM it to you are you pop on discord.

liach: Hmm, if you don't mind using github team chat...
I agree with not using supplier. Then you need to add docs :sunglasses: for the functional interface.

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityRendererRegistry.java
i509VCB: I think this was the result of resolving some imports so javadoc resolved.

`gradlew check` doesn't complain about this

liach (approved):

Player3324 (approved):

ghost (approved):
Whilst the API is good, I'm not sure if it's even worth adding as it all is being moved to json.

liach:
This is just a temporary bridge and may be removed safely since this is deprecated + beta on addition

PR #1201, opened 2020-12-04
Optimization in fabric api redirect for block entity unload

i509VCB:
This replaces the simple `removeAll` in vanilla and replaces it with a set allocation and use of removeIf + set lookup for reducing overhead in removing from the list of a world's block entities that fabric api redirects for it's block entity events.

We may want to consider only allocating the list of the world's block entity list is only so big? Apparently this issue occurs with 20k block entities on AOF3 NA.

This is result of the following spark profile: https://spark.lucko.me/#SwMPsVKNM1

@Technici4n came up with the general idea for the optimization, it needs to be combed through with the profiler.

Specifically for @sfPlayer1: you should make sure this fix isn't insane or if a better solution exists.

I think a good test for this is to fill a several chunks with a load of block entities and unload and load em frequently.

This is not needed in 1.17, but since block entities are stored by chunk in that version, things may be faster.

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
@@ -85,7 +85,7 @@ public abstract class WorldMixin {
 			return blockEntityList.removeAll(removals);
 		}
 
-		// Allocating a set when dealing with huge amounts of loaded block entities and then using removeIf is more performant than List#removeAll
+		// List -> List.removeAll is slower than reallocating the List as a Set and then doing a Set -> List.removalAll when dealing with large amounts of loaded block entities.
 		final Set<BlockEntity> removalSet = new ReferenceOpenHashSet<>(removals);
 		return blockEntityList.removeAll(removalSet);
 	}
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
@@ -81,13 +81,8 @@ public abstract class WorldMixin {
 
 		// Mimic vanilla logic - with some perf boosts
 		// Gate perf boost behind a minimum amount of removals
-		if (removals.size() <= 4) {
-			return blockEntityList.removeAll(removals);
-		}
-
 		// List -> List.removeAll is slower than reallocating the List as a Set and then doing a Set -> List.removalAll when dealing with large amounts of loaded block entities.
-		final Set<BlockEntity> removalSet = new ReferenceOpenHashSet<>(removals);
-		return blockEntityList.removeAll(removalSet);
+		return blockEntityList.removeAll(removals.size() <= 4 ? removals : new ReferenceOpenHashSet<>(removals));
 	}
 
 	@Inject(at = @At("RETURN"), method = "tickBlockEntities")
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
@@ -19,7 +19,9 @@ package net.fabricmc.fabric.mixin.event.lifecycle;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 
+import it.unimi.dsi.fastutil.objects.ReferenceArraySet;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
@@ -77,8 +79,13 @@ public abstract class WorldMixin {
 			}
 		}
 
-		// Mimic vanilla logic
-		return blockEntityList.removeAll(removals);
+		// Mimic vanilla logic - with some perf boosts
+		// Allocating a set when dealing with huge amounts of loaded block entities and then using removeIf is more performant than List#removeAll
+		final Set<BlockEntity> removalSet = new ReferenceArraySet<>(removals);
+		return blockEntityList.removeIf(removalSet::contains);
+
+		// Original code
+		// return blockEntityList.removeAll(removals);
 	}
 
 	@Inject(at = @At("RETURN"), method = "tickBlockEntities")
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
@@ -21,7 +21,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
-import it.unimi.dsi.fastutil.objects.ReferenceArraySet;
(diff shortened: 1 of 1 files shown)


Technici4n (approved):
Thank you! I will try to benchmark that when I find the time.

Player3324 (changes requested):
Confirmed insane! :100:

1. ReferenceArraySet is terrible. As its JavaDoc clearly states, it is a pretend set based on an array that is being used almost exactly like the original ArrayList (scan through the entire thing with each contains). Only a proper hash set like ReferenceOpenHashSet will yield the desired complexity benefit..

2. ArraySet.removeAll is just fine and better than removeIf(capturing lambda), it internally calls removalSet.contains as well with batch removal (no superfluous element moving). ArrayList.removeIf uses a two step approach to make sure the list isn't being modified (thus changing element indices) before testing every element. That's obviously some overhead beyond the functional operator indirection.

3. I'd gate the optimization a bit, if removals is very small, there is no need to construct the extra set. I'd use it for `removals.size() > 4`.

maityyy:
This is already optimized by Lithium.
https://github.com/jellysquid3/lithium-fabric/pull/102

i509VCB:
What I see is lithium covers the actual entity backing list while this optimization reduces the overhead in accessing entries from the removals list to then remove from the backing list.

Not sure on the implications here but this should be fine I think and not cause any adverse issues with lithium.

Player3324 (approved):

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
```diff
@@ -80,12 +80,14 @@ protected boolean onPurgeRemovedBlockEntities(List<BlockEntity> blockEntityList,
...
-		final Set<BlockEntity> removalSet = new ReferenceArraySet<>(removals);
-		return blockEntityList.removeIf(removalSet::contains);
+		// Gate perf boost behind a minimum amount of removals
+		if (removals.size() <= 4) {
+			return blockEntityList.removeAll(removals);
+		}
 
-		// Original code
-		// return blockEntityList.removeAll(removals);
+		// Allocating a set when dealing with huge amounts of loaded block entities and then using removeIf is more performant than List#removeAll
```
Player3324: The comment is slightly out of date

maityyy:
> What I see is lithium covers the actual entity backing list while this optimization reduces the overhead in accessing entries from the removals list to then remove from the backing list.
> 
> Not sure on the implications here but this should be fine I think and not cause any adverse issues with lithium.

Well yes, you are right. There is no point in not doing this... /shrug

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
```diff
@@ -77,8 +79,15 @@ protected boolean onPurgeRemovedBlockEntities(List<BlockEntity> blockEntityList,
 			}
 		}
 
-		// Mimic vanilla logic
-		return blockEntityList.removeAll(removals);
+		// Mimic vanilla logic - with some perf boosts
+		// Gate perf boost behind a minimum amount of removals
+		if (removals.size() <= 4) {
```
maityyy: Why exactly 4? In vanilla survival, I guess that number won't appear at all. A useless check in essence.

i509VCB: The number itself could be tweaked, allocating a new set is more costly than removing from List -> List with certain amounts of elements.

Technici4n: We are talking about bases with maybe 1000 BEs in some chunks, so yes definitely needed.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
```diff
@@ -85,7 +85,7 @@ protected boolean onPurgeRemovedBlockEntities(List<BlockEntity> blockEntityList,
 			return blockEntityList.removeAll(removals);
 		}
 
-		// Allocating a set when dealing with huge amounts of loaded block entities and then using removeIf is more performant than List#removeAll
+		// List -> List.removeAll is slower than reallocating the List as a Set and then doing a Set -> List.removalAll when dealing with large amounts of loaded block entities.
```
liach, line 83: You should really say that it's because `List#contains` is significantly slower than `Set#contains`, which `removeAll` calls.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
liach: I suggest moving this to a static final field, like the fence used for converting a linked list to a red-black tree in hash map

liach (changes requested):

liach:
Actually I prefer the whole patch to be like this instead:
```java
// Make sure removing.contains() call would be efficient enough
final Collection<BlockEntity> removing = removals.size() <= 4 ? removals : new ReferenceOpenHashSet<>(removals);
return blockEntityList.removeAll(removing);
```
Or you can shrink it into one line either. This imo is cleaner than your current approach.

modmuss50 (approved):
Looks fine, just needs testing of course.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
```diff
@@ -77,8 +79,15 @@ protected boolean onPurgeRemovedBlockEntities(List<BlockEntity> blockEntityList,
...
-		return blockEntityList.removeAll(removals);
+		// Mimic vanilla logic - with some perf boosts
+		// Gate perf boost behind a minimum amount of removals
+		if (removals.size() <= 4) {
+			return blockEntityList.removeAll(removals);
+		}
+
+		// List -> List.removeAll is slower than reallocating the List as a Set and then doing a Set -> List.removalAll when dealing with large amounts of loaded block entities.
+		final Set<BlockEntity> removalSet = new ReferenceOpenHashSet<>(removals);
+		return blockEntityList.removeAll(removalSet);
```
liach:
```suggestion
		// List -> List.removeAll is slower than reallocating the List as a Set and then doing a Set -> List.removalAll when dealing with large amounts of loaded block entities.
		return blockEntityList.removeAll(removals.size() <= 4 ? removals : new ReferenceOpenHashSet<>(removals));
```

liach (approved):

liach:
Hmm, there's a checkstyle error that needs fixing locally. Is it line length for that long comment?

modmuss50:
Im not sure, would be useful if it said in the build log what was wrong :D

Technici4n:
`Unused import - java.util.Set. [UnusedImports]` - you just need to look a bit more carefully. ;)

i509VCB:
The checkstyle error is in lifecycle events.

What on earth

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java
```diff
@@ -19,7 +19,9 @@
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
```
liach:
```suggestion
```

liach: Ignore this

PR #1206, opened 2020-12-07
Expose BucketItem#fluid

Ueaj-Kerman:
I don't know why this is private, pretty odd, but me and others end up making accessors for this field. 
So I thought it might be nice to just add it to fapi to make it easier for everyone


shedaniel:
AW isn’t transitive is it

Ueaj-Kerman:
is it not, I thought it was

vini2003:
in the name of god, in the name of the mighty lord, i approv

Ueaj-Kerman:
nvm I don't think it's transitive

PR #1087, opened 2020-09-17
New module to help with the hardcoded features of Minecraft

OroArmor:
Minecraft has a lot of hardcoded features that make adding unique items and blocks often very difficult, with many mixins. The goal of this module is to create a common set of mixins to help with this hardcoding so that mod developers dont need to track down 30 references of a item/block and that these mixins dont conflict (one of the main reasons i believe this should be part of fabric-api).

PR #1210, opened 2020-12-10
Add a Bow API into the Item API

OroArmor:
This PR is part of many that split up the proposed extensibility API into their own modules. This one focuses on the usage of bows in Minecraft. The main goal is to make sure that no mods will have any conflicts with the mixins for their custom bows


florensie:
Link to make it easier for everyone to navigate from/to old PR: #1087

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/api/bow/item/v1/SimpleBowItem.java
```diff
@@ -0,0 +1,35 @@
...
+
+package net.fabricmc.fabric.api.bow.item.v1;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * This is the default implementation for FabricBow, allowing for the easy creation of new bows with no new modded functionality.
```
BoogieMonster1O1: think you mean `BowInterface`

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/mixin/bow/EntityUseBowMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.entity.mob.AbstractSkeletonEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.RangedWeaponItem;
+
+// {} is to show that this mixin can be used for many entities
+@Mixin({AbstractSkeletonEntity.class})
```
BoogieMonster1O1: correct me if I'm wrong but don't illusioners also use bows?

OroArmor: It's in `BowAttackGoal`, which I think I chose not to modify (not sure why). There are a couple of methods in there that I need to modify, but it's not too major.

The skeleton entities have their own special canUseRangedWeapon method that the illusioner does not. I think this comes from the fact that some entities can hold different items/have it break. Because the illusioner is not finished mojang left some mojank.

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/api/bow/item/v1/SimpleBowItem.java
OroArmor: Ah yeah, thats left over from when I had old names

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/mixin/bow/EntityUseBowMixin.java
BoogieMonster1O1: `BowAttackGoal` can also be used by other modded entities since it's not just a plain inner class.

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/api/bow/item/v1/BowInterface.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement for all custom bows in fabric. <br>
+ * Note: This is meant to be used on a BowItem class, the functionality will not work otherwise.
+ *
+ * @see SimpleBowItem
+ */
+public interface BowInterface {
```
i509VCB: On second thought, maybe `BowLike`?

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/mixin/bow/EntityUseBowMixin.java
OroArmor: Alright i fixed this up. I also noticed another thing from bow attack goal to fix

ghost:
Shouldn't this be contained within the items API? The bow is an item correct? Why make an entire module for it. Hey, let's make a swords API next.

OroArmor:
Yeah, this was discussed in the discord, and I haven't had time to update the code yet. However, a review on the content of the pr would be more helpful as changes to how it works is more important than where it is.

OroArmor:
I have now merged the Bow API into the Item API, and I am looking for reviews

gradle.properties
```diff
@@ -1 +1 @@
-org.gradle.jvmargs=-Xmx2560M
+org.gradle.jvmargs=-Xmx5120M
```
florensie: I don't think you meant to commit this. And this change could be problematic for people with only 8GB of ram.

BoogieMonster1O1: Agreed. I only have four gigs to spare, and I’m pretty sure a chunk of people do as well

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/api/bow/item/v1/BowInterface.java
florensie: Not really a fan of having the word interface in the name of an interface, it's almost the same as calling it `IBow`. `BowLike` or even `BowFunctionality` could work.

gradle.properties
OroArmor: I added this because on one of my previous prs, it was failing due to a lack of memory.

fabric-bow-api-v1/src/main/java/net/fabricmc/fabric/api/bow/item/v1/BowInterface.java
OroArmor: What about `BowExtension(s)` because it is focused on extending the bow functionality.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * In this method you can modify the behavior of arrows shot from your custom bow. Applies all of the vanilla arrow modifiers first.
+	 *
+	 * @param bowStack                   The ItemStack for the Bow Item
+	 * @param arrowStack                 The ItemStack for the arrows
+	 * @param user                       The user of the bow
+	 * @param remainingUseTicks          The ticks remaining on the bow usage
+	 * @param persistentProjectileEntity The arrow entity to be spawned
+	 */
+	void modifyShotProjectile(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, int remainingUseTicks, PersistentProjectileEntity persistentProjectileEntity);
```
BoogieMonster1O1: Won't `remainingUseTicks` always be 0 after shooting?

BoogieMonster1O1: Probably rename this to `onProjectileShot`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Gets the pull progress of the bow between 0 and 1.
```
BoogieMonster1O1:
```suggestion
	 * Returns the pull progress of the bow between 0 and 1.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EntityUseBowMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import net.minecraft.item.BowItem;
+import net.minecraft.item.RangedWeaponItem;
+
+// {} is to show that this mixin can be used for many entities
+@Mixin({AbstractSkeletonEntity.class})
+public class EntityUseBowMixin {
+	// Allows Entities that use bows to shoot custom bows.
+	@Inject(method = "canUseRangedWeapon", at = @At("HEAD"), cancellable = true)
+	public void canUseRangedWeapon(RangedWeaponItem weapon, CallbackInfoReturnable<Boolean> cir) {
+		if (weapon instanceof BowItem) {
```
BoogieMonster1O1: this can be simplified to `cir.setReturnValue(weapon instanceof BowItem)` since there's no other logic in the target method

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
OroArmor: ah i was trying to get pull progress

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * In this method you can modify the behavior of arrows shot from your custom bow. Applies all of the vanilla arrow modifiers first.
+	 *
+	 * @param bowStack                   The ItemStack for the Bow Item
+	 * @param arrowStack                 The ItemStack for the arrows
+	 * @param user                       The user of the bow
+	 * @param pullProgress          	 The pull progress of the bow from 0.0 to 1.0
+	 * @param persistentProjectileEntity The arrow entity to be spawned
+	 */
+	void onProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity persistentProjectileEntity);
```
BoogieMonster1O1: Still, won’t the pull progress be 1 when the projectile is shot? Isn’t the projectile supposed to be shot only when the pull progress is 1?

OroArmor: no, you know how you can partially charge a bow? thats what this value is

BoogieMonster1O1: ah I see

BoogieMonster1O1 (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractSkeletonEntityMixin.java
```diff
@@ -0,0 +1,37 @@
...
+import net.minecraft.entity.mob.AbstractSkeletonEntity;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(AbstractSkeletonEntity.class)
+public class AbstractSkeletonEntityMixin {
+	//Allows Skeletons to shoot custom bows
```
florensie: Add a space after `//`?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BowItemMixin.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.item.ArrowItem;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(BowItem.class)
+public class BowItemMixin {
+	//Allows custom bows to modify the projectile shot by bows
```
florensie: Same thing

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BowItemMixin.java
```diff
@@ -0,0 +1,54 @@
...
+	@Inject(method = "onStoppedUsing(Lnet/minecraft/item/ItemStack;Lnet/minecraft/world/World;Lnet/minecraft/entity/LivingEntity;I)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;spawnEntity(Lnet/minecraft/entity/Entity;)Z"), locals = LocalCapture.CAPTURE_FAILHARD)
+	public void onStoppedUsing(ItemStack bowStack, World world, LivingEntity user, int remainingUseTicks, CallbackInfo info, PlayerEntity playerEntity, boolean bl, ItemStack arrowStack, int i, float pullProgress, boolean bl2, ArrowItem arrowItem, PersistentProjectileEntity persistentProjectileEntity) {
+		if (this instanceof BowExtensions) {
+			((BowExtensions) this).onProjectileShot(bowStack, arrowStack, user, pullProgress, persistentProjectileEntity);
+		}
+	}
+
+	@Redirect(method = "onStoppedUsing", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/BowItem;getPullProgress(I)F"))
+	private float redirectPullProgress(int useTicks) {
```
florensie: It's not immediately clear from this code why a modder can't simply override `getPullProgress` on their own implementation, unless you look at it's source (note: it's static). Things that are not immediately apparent like this should have an explaining comment.

This then raises my next question: why use multiple redirects instead of a cancellable inject in `getPullProgress`? Again, if there's a good reason for that then make sure it's explained with a comment.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EntityUseBowMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.entity.mob.AbstractSkeletonEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.RangedWeaponItem;
+
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+// {} is to show that this mixin can be used for many entities
```
florensie: I'm confused, this doesn't actually do anything does it? I would leave out the redundant braces and write a comment like this:
```
This mixin should be applied to any vanilla MobEntity that can use a bow
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ProjectileUtilMixin.java
```diff
@@ -0,0 +1,47 @@
...
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(ProjectileUtil.class)
+public class ProjectileUtilMixin {
+	// Because the uses of this method are hardcoded, checking each hand for the Fabric interfaces of the items is needed.
+	// Note: this does not cancel for the vanilla items unless they are holding a custom implementation of the items
+	@Inject(method = "getHandPossiblyHolding", at = @At(value = "HEAD"), cancellable = true)
+	private static void getHandPossiblyHolding(LivingEntity entity, Item item, CallbackInfoReturnable<Hand> cir) {
+		if (item == Items.BOW) {
+			boolean inMainHand = entity.getStackInHand(Hand.MAIN_HAND).getItem() instanceof BowExtensions;
```
florensie: How is this for a nifty alternative?
```java
for (Hand hand : Hand.values()) {
	if (entity.getStackInHand(hand).getItem() instanceof BowExtensions) {
		cir.setReturnValue(hand);
		return
	}
}
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/HeldItemRendererMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(HeldItemRenderer.class)
+public class HeldItemRendererMixin {
+	@Redirect(method = "renderItem(FLnet/minecraft/client/util/math/MatrixStack;Lnet/minecraft/client/render/VertexConsumerProvider$Immediate;Lnet/minecraft/client/network/ClientPlayerEntity;I)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getItem()Lnet/minecraft/item/Item;", ordinal = 0))
+	private Item renderItem(ItemStack heldItem) {
```
florensie, line 34: Explain what this method actually achieves as a result.

florensie (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BowItemMixin.java
OroArmor: `getPullProgress` only has one parameter for how many ticks the bow has been held. You don't have access to the item stack and therefore cannot determine the bow item that is being pulled

florensie: Oh yeah that makes sense.

florensie (approved):
Personally I like to make all my mixin classes abstract but there's no style guideline for this afaik so you decide.

Looks good to me!

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,51 @@
...
+package net.fabricmc.fabric.api.item.v1.bow;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement for all custom bows in fabric. <br>
+ * Note: This is meant to be used on a BowItem class, the functionality will not work otherwise.
```
liach:
```suggestion
 *
 * @implNote The implementations should be a subclass of {@link BowItem}. The functionality will not work otherwise.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BowAttackGoalMixin.java
```diff
@@ -0,0 +1,56 @@
...
+
+@Mixin(BowAttackGoal.class)
+public class BowAttackGoalMixin {
+	@Shadow
+	@Final
+	private HostileEntity actor;
+
+	@Inject(method = "isHoldingBow()Z", at = @At("HEAD"), cancellable = true)
+	private void isHoldingCustomBow(CallbackInfoReturnable<Boolean> callbackInfo) {
+		boolean holdingCustomBow = actor.isHolding((item) -> item instanceof BowExtensions);
```
liach:
```suggestion
		boolean holdingCustomBow = actor.isHolding(BowExtensions.class::isInstance);
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ProjectileUtilMixin.java
```diff
@@ -0,0 +1,47 @@
...
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(ProjectileUtil.class)
+public class ProjectileUtilMixin {
+	// Because the uses of this method are hardcoded, checking each hand for the Fabric interfaces of the items is needed.
+	// Note: this does not cancel for the vanilla items unless they are holding a custom implementation of the items
+	@Inject(method = "getHandPossiblyHolding", at = @At(value = "HEAD"), cancellable = true)
+	private static void getHandPossiblyHolding(LivingEntity entity, Item item, CallbackInfoReturnable<Hand> cir) {
+		for (Hand hand : Hand.values()) {
+			if (item == Items.BOW) { // Make sure we only check for bows when searching for bows and allow for other items like crossbows in future
```
liach, line 41: Should make the if check outside of the for loop

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EntityUseBowMixin.java
```diff
@@ -0,0 +1,37 @@
...
+import net.minecraft.item.RangedWeaponItem;
+
+import net.fabricmc.fabric.api.item.v1.bow.BowExtensions;
+
+@Mixin(AbstractSkeletonEntity.class)
+public class EntityUseBowMixin {
+	// Allows Entities that use bows to shoot custom bows.
+	@Inject(method = "canUseRangedWeapon", at = @At("HEAD"), cancellable = true)
+	public void canUseRangedWeapon(RangedWeaponItem weapon, CallbackInfoReturnable<Boolean> cir) {
+		cir.setReturnValue(weapon instanceof BowItem || weapon instanceof BowExtensions);
```
liach:
```suggestion
		cir.setReturnValue(weapon instanceof BowItem);
```
per your contract `BowExtensions` implementations always implement `BowItem`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	void onProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity persistentProjectileEntity);
+
+	/**
+	 * Returns the pull progress of the bow between 0 and 1.
+	 *
+	 * @param useTicks The number of ticks the bow has been pulled.
+	 * @return The progress of the pull from 0.0f to 1.0f.
+	 */
+	default float getCustomPullProgress(int useTicks) {
```
liach: Hmm, can you offer item stack access in this method? [`onStoppedUsing`](https://github.com/FabricMC/fabric/pull/1210/files#diff-87cd612012ddfcde82094648318e7dc39dc9eeef80ba2c80c52583fb8ece0e32R46) passes the item stack afaik.

fabric-item-api-v1/build.gradle
```diff
@@ -1,6 +1,10 @@
 archivesBaseName = "fabric-item-api-v1"
 version = getSubprojectVersion(project, "1.2.0")
```
liach, line 2: Bump version

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ProjectileUtilMixin.java
OroArmor: yes, but when I create my crossbow API, i will modify this method and i want to show how it can be used for a multitude of items.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
OroArmor: Ah. I should make it clearer that this *can* work on non-bow items.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/BowExtensions.java
OroArmor: It doesnt seem to, but it might be something useful, like with a quickcharge enchantment.

OroArmor: hm looking into the minecraft code, id have to make some pretty nasty mixins to pass in the bow stack. but who cares about implementation details if it works 😎.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EntityUseBowMixin.java
OroArmor: As mentioned above, people can implement this on custom items, but it is recommended on bows.

florensie (changes requested):
None of those hacky mixins are necessary to add the stack to your pull progress. A redirect can capture the outer method's arguments.

florensie (approved):

OroArmor:
Alright so a serious thing that i thought about is an event that is run when an arrow is fired.
I just made modifications that will make it easy to add an event, but i don't really know where to put it. `ServerEntityCombatEvents` seems like the most logical place for it. i am also not sure how to 100% make the event because if two mods modify the projectiles to new entities, some of the features may get lost. That may also be outside the scope of fabric API and would need those two mods to resolve the conflict. The main reason i though of this was that someone was trying to make an enchantment that makes arrows summon lightening, and being able to do it this way seems the best.

OroArmor:
After some discussion in the discord. A couple people (especially @florensie thanks) suggested creating a new module `fabric-combat-api-v1`. This module would include both the events and API for expanding the combat system of Minecraft. Other features like tridents, crossbows, and shields would have apis to easily create new variants, and events like shot projectiles, thrown tridents, on hit, and more would have hooks provided by this module. ~why does it always come back to this~

OroArmor:
Alright take 3 (4?). This is a new module for combat-related events and items. Bows, Crossbows, Shields, and Tridents are all extremely hardcoded into Minecraft, and working on expanding this is a key part of this module. Another large part will be for combat events. Shot bow and crossbow projectiles, thrown tridents, thrown potions, shield block, projectile hit, and more events will be added to add more hooks into minecrafts combat system. This pr will not cover all of these, and smaller parts should be added at a time. Currently the pr only covers bows, as that is small. Expect more PRs to come as the previous is merged in

LambdAurora (approved):
In my opinion I would rename SimpleBowItem and BowExtensions to FabricBowItem and FabricBowExtensions to avoid confusion especially with new modders.

overall good :)

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,48 @@
...
+package net.fabricmc.fabric.api.combat.v1;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public interface ShotProjectileEvents {
```
i509VCB: Make this a final class with private constructor.

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/bow/BowExtensions.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.combat.v1.ShotProjectileEvents;
+
+/**
+ * An interface to implement for all custom bows in fabric. <br>
+ *
+ * @implNote This is meant to be used on a BowItem class. Unless similar functionality is implemented on your custom item, most functionality will not work.
```
i509VCB: I've been told the jdk doclet does not support this tag, make it a new line in javadoc instead.

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/AbstractSkeletonEntityMixin.java
```diff
@@ -0,0 +1,45 @@
...
+@Mixin(AbstractSkeletonEntity.class)
+public abstract class AbstractSkeletonEntityMixin extends MobEntity implements RangedAttackMob {
+	protected AbstractSkeletonEntityMixin(EntityType<? extends MobEntity> entityType, World world) {
+		super(entityType, world);
+	}
+
+	// Allows Skeletons to shoot custom bows
+	@Redirect(method = "updateAttackType", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getItem()Lnet/minecraft/item/Item;"))
+	private Item updateAttackType(ItemStack itemStack) {
+		return itemStack.getItem() instanceof BowExtensions ? Items.BOW : itemStack.getItem();
```
i509VCB: Make sure you put a comment saying "return bow to evaluate to true".

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/BowAttackMixin.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.entity.mob.MobEntity;
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.entity.projectile.ProjectileUtil;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.combat.v1.ShotProjectileEvents;
+
+@Mixin({AbstractSkeletonEntity.class, IllusionerEntity.class})
```
i509VCB: Comment that this may need to be updated when new mobs that can weild bows are added.

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/BowItemMixin.java
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraft.item.ArrowItem;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.combat.v1.ShotProjectileEvents;
+import net.fabricmc.fabric.api.combat.v1.bow.BowExtensions;
+
+@Mixin(BowItem.class)
+public class BowItemMixin {
```
i509VCB: Abstract

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/EntityUseBowMixin.java
```diff
@@ -0,0 +1,37 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.entity.mob.AbstractSkeletonEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.RangedWeaponItem;
+
+import net.fabricmc.fabric.api.combat.v1.bow.BowExtensions;
+
+@Mixin(AbstractSkeletonEntity.class)
+public class EntityUseBowMixin {
```
i509VCB: Abstract

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/ProjectileUtilMixin.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.projectile.ProjectileUtil;
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
+import net.minecraft.util.Hand;
+
+import net.fabricmc.fabric.api.combat.v1.bow.BowExtensions;
+
+@Mixin(ProjectileUtil.class)
+public class ProjectileUtilMixin {
```
i509VCB: Abstract

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/client/HeldItemRendererMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.minecraft.client.render.item.HeldItemRenderer;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+
+import net.fabricmc.fabric.api.combat.v1.bow.BowExtensions;
+
+@Mixin(HeldItemRenderer.class)
+public class HeldItemRendererMixin {
```
i509VCB: Abstract

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/client/SkeletonEntityModelMixin.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.minecraft.client.render.entity.model.SkeletonEntityModel;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+
+import net.fabricmc.fabric.api.combat.v1.bow.BowExtensions;
+
+@Mixin(SkeletonEntityModel.class)
+public class SkeletonEntityModelMixin {
```
i509VCB: Abstract

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,48 @@
...
+	Event<ProjectileFromBow> BOW_SHOT_PROJECTILE = EventFactory.createArrayBacked(ProjectileFromBow.class, callbacks -> (ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity persistentProjectileEntity) -> {
```
i509VCB: There may be a usage issue here, so the way it looks in rest of codebase, the custom bow logic is implemented via this event directly on the bow item. However this can cause an issue where a custom bow launches a rock and then a listener replaces said rock with a poisoned arrow.

Essentially this has an issue where the last mod to be registered wins.

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/BowAttackMixin.java
```diff
@@ -0,0 +1,53 @@
...
+public abstract class BowAttackMixin extends MobEntity implements RangedAttackMob {
+	protected BowAttackMixin(EntityType<? extends MobEntity> entityType, World world) {
+		super(entityType, world);
+	}
+
+	@Redirect(method = "attack", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;spawnEntity(Lnet/minecraft/entity/Entity;)Z"))
+	public boolean modifyShotProjectile(World world, Entity persistentProjectileEntity, LivingEntity target, float pullProgress) {
```
i509VCB: This will not replace the parameter, you need to use ModifyVariable in order to replace the old entity with a new one.

i509VCB (changes requested):

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
OroArmor: Yeah, im not sure how to get around this issue

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/BowAttackMixin.java
OroArmor: doesnt the event system chain together the events then return the final result? here i am just redirecting the method call and then replacing the old entity with the events and then calling the spawn method with my new entity

i509VCB: This is a mixin issue:

Here is an example of what happens with your mixin:

```java
void attack(params) {
    projectile = createProjectile();
    if (redirect$zzz000$modifyShotProjectile(projectile)) {
        ...
    }

    ...
}
```

The when you pass the parameter into the added redirect, you do not mutate the input parameter, if you change the parameter within your own mixin.

One way to consider this is every method is implicitly like this:
```java
public boolean modifyShotProjectile(final World world, final Entity persistentProjectileEntity, final LivingEntity target, final float pullProgress) {
```

You can set the value without it being final, but it acts more like a variable declared in your method which will drop out of scope when the mixin's method is exited.

OroArmor: so if i just made a different local variable would that fix it?

i509VCB: I probably did not eloquently describe this, but the simple fix is to change the projectile entity after this redirect by querying a sort of "temporary" field.

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * This event modifies the projectile entity. Returning a new entity will not change the entity that is spawned in.
+	 */
+	public static final Event<ModifyProjectileFromBow> BOW_MODIFY_SHOT_PROJECTILE = EventFactory.createArrayBacked(ModifyProjectileFromBow.class, callbacks -> (bowStack, arrowStack, user, pullProgress, persistentProjectileEntity) -> {
+		for (ModifyProjectileFromBow callback : callbacks) {
+			if (callback instanceof FabricBowItem) {
+				if (callback == bowStack.getItem()) {
+					callback.modifyProjectileShot(bowStack, arrowStack, user, pullProgress, persistentProjectileEntity);
+				}
```
liach: Just fyi on your logic of skipping fire when the callback is a fabric bow and the handled bow is not the callback's bow. If there is exactly one callback, your invoker generation logic will be skipped. See https://github.com/FabricMC/fabric/blob/4681c750b8a508e37a08c1fe94498f21fef7e42b/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java#L46-L47
Imo player shouldn't assume he can just delegate to single callback than bypass invoker logic

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/ProjectileUtilMixin.java
```diff
@@ -0,0 +1,47 @@
...
+import net.fabricmc.fabric.api.combat.v1.bow.FabricBowExtensions;
+
+@Mixin(ProjectileUtil.class)
+public abstract class ProjectileUtilMixin {
+	// Because the uses of this method are hardcoded, checking each hand for the Fabric interfaces of the items is needed.
+	// Note: this does not cancel for the vanilla items unless they are holding a custom implementation of the items
+	@Inject(method = "getHandPossiblyHolding", at = @At(value = "HEAD"), cancellable = true)
+	private static void getHandPossiblyHolding(LivingEntity entity, Item item, CallbackInfoReturnable<Hand> cir) {
+		for (Hand hand : Hand.values()) {
+			if (item == Items.BOW) { // Make sure we only check for bows when searching for bows and allow for other items like crossbows in future
```
liach: can we move the `item == Items.BOW` check outside of the for loop? Each time the for loop is called a new array is generated by an Hand.values() call

OroArmor: The crossbow also uses this method, and i plan on adding functionality in the same mixin later on

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
OroArmor: i think that only happens when the invoker is registered and then future callbacks are then in future ones... i could be wrong and need to test this

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/ProjectileUtilMixin.java
liach: Shame that mojang didn't make some VALUES array for hand enum

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/api/combat/v1/ShotProjectileEvents.java
liach: You can just write an event where the invoker generator is like
```java
callbacks -> {
  LOGGER.info("Called invoker factory with callback with length {}", callbacks.length);
  return (params) -> {
    for (CallbackName each : callbacks) {
       each.call(params);
    }
  }
}
```
and see the log outputs.

OroArmor: Alright, i just realized that im already checking in the fabric bow class, i don't need to duplicate the check logic

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/ProjectileUtilMixin.java
OroArmor: Alright, i made a quick cached constant to not generate it each time the method is called

fabric-combat-api-v1/src/main/java/net/fabricmc/fabric/mixin/combat/EntityUseBowMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.minecraft.entity.mob.AbstractSkeletonEntity;
+import net.minecraft.item.BowItem;
+import net.minecraft.item.RangedWeaponItem;
+
+import net.fabricmc.fabric.api.combat.v1.bow.FabricBowExtensions;
+
+@Mixin(AbstractSkeletonEntity.class)
+public abstract class EntityUseBowMixin {
+	// Allows Entities that use bows to shoot custom bows.
```
Prospector:
```suggestion
	// Allows Entities that use bows to use custom bows.
```

You just used the word use, but then shoot? You don't really shoot a bow, you use a bow to shoot

OroArmor: works for me

Platymemo (approved):
Add a tag? 👉👈

liach:
> 
> 
> Add a tag? 👉👈

Imo checking by tag isn't significantly faster and may be error prone if the item in tag happens to not be a fabric bow.

OroArmor:
Tbh I think it's kinda in the middle. It can be used to make items render like a bow and for people to check tags easily, but it wouldn't provide any helpful methods/use the default ones from bow. That would allow people to only use the interface if they want to add custom functionality.

LambdAurora (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,83 @@
...
+		 * In this method you can replace the arrow shot from your custom bow. Applies all of the vanilla arrow modifiers first.
+		 *
+		 * @param bowStack                   the ItemStack for the Bow Item
+		 * @param arrowStack                 the ItemStack for the arrows
+		 * @param user                       the user of the bow
+		 * @param pullProgress               the pull progress of the bow from 0.0 to 1.0
+		 * @param persistentProjectileEntity the arrow entity to be spawned
+		 * @return the arrow entity, either new or {@code null} to signify no changes occurred
+		 */
+		PersistentProjectileEntity replaceProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, @NotNull PersistentProjectileEntity persistentProjectileEntity);
```
Juuxel: `NotNull` is not really needed on the parameter since it's assumed. (Also, consider a shorter parameter name like `projectile`)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,83 @@
...
+		/**
+		 * In this method you can modify the behavior of arrows shot from your custom bow. Applies all of the vanilla arrow modifiers first.
+		 *
+		 * @param bowStack                   the ItemStack for the Bow Item
+		 * @param arrowStack                 the ItemStack for the arrows
+		 * @param user                       the user of the bow
+		 * @param pullProgress               the pull progress of the bow from 0.0 to 1.0
+		 * @param persistentProjectileEntity the arrow entity to be spawned
+		 */
+		void modifyProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, @NotNull PersistentProjectileEntity persistentProjectileEntity);
```
Juuxel: Same here

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ProjectileUtilMixin.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.entity.projectile.ProjectileUtil;
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
+import net.minecraft.util.Hand;
+
+import net.fabricmc.fabric.api.item.v1.bow.FabricBowExtensions;
+
+@Mixin(ProjectileUtil.class)
+public abstract class ProjectileUtilMixin {
+	private static final Hand[] HANDS = {Hand.MAIN_HAND, Hand.OFF_HAND}; // Cache the hands to not create the hands array each time the loop is run
```
Juuxel:
```suggestion
	private static final Hand[] HANDS = Hand.values(); // Cache the hands to not create the hands array each time the loop is run
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/FabricBowExtensions.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.fabricmc.fabric.api.item.v1.ShotProjectileEvents;
+
+/**
+ * An interface to implement for all custom bows in fabric. <br>
+ * This is meant to be used on a {@link BowItem} class. Unless similar functionality is implemented on your custom item, most functionality will not work.
+ *
+ * @see FabricBowItem
+ */
+public interface FabricBowExtensions extends ShotProjectileEvents.ModifyProjectileFromBow {
```
Juuxel: Extending the callback interface might be a bit confusing for users since it's not actually registered as a callback unless it's done manually.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/FabricBowItem.java
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	@Override
+	public final void modifyProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, @NotNull PersistentProjectileEntity persistentProjectileEntity) {
+		if (bowStack.getItem() == this) {
+			onProjectileShot(bowStack, arrowStack, user, pullProgress, persistentProjectileEntity);
+		}
+	}
+
+	public void onProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity persistentProjectileEntity) {
```
Juuxel: This also needs some note about the required manual registration.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/bow/BowItemMixin.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.item.v1.ShotProjectileEvents;
+import net.fabricmc.fabric.api.item.v1.bow.FabricBowExtensions;
+
+@Mixin(BowItem.class)
+public abstract class BowItemMixin {
+	@Unique
+	private PersistentProjectileEntity shotProjectile;
```
Juuxel, line 42: Resetting this at the end of `onStoppedUsing` might be good to allow GC to do its job (and prevent other possible side effects).

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
Juuxel: There should be a `@Nullable` on this method as well.

Juuxel (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ProjectileUtilMixin.java
shedaniel: We don't need this then, just inline it.
EDIT: Nevermind, my bad.

liach: Yeah the fact that java allocates a new array on each `values()` call is meh

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/FabricBowExtensions.java
```diff
@@ -0,0 +1,42 @@
...
+
+package net.fabricmc.fabric.api.item.v1.bow;
+
+import net.minecraft.item.BowItem;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.item.v1.ShotProjectileEvents;
+
+/**
+ * An interface to implement for all custom bows in fabric. <br>
```
Juuxel:
```suggestion
 * An interface to implement for all custom bows in Fabric. <br>
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/bow/FabricBowItem.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	@Override
+	public final void modifyProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity projectile) {
+		if (bowStack.getItem() == this) {
+			onProjectileShot(bowStack, arrowStack, user, pullProgress, projectile);
+		}
+	}
+
+	public void onProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity persistentProjectileEntity) {
```
Juuxel:
```suggestion
	public void onProjectileShot(ItemStack bowStack, ItemStack arrowStack, LivingEntity user, float pullProgress, PersistentProjectileEntity projectile) {
```

Juuxel (changes requested):

liach:
why did you suddenly close these prs?

Technici4n:
Same as i5 yesterday. Some people have are not happy with how the Fabric project is ran, and have decided to contribute to a fork instead called Quilt, soon to be made public, but they are planning structural changes first.

liach:
hmm, but doesn't i5 literally have write access? that's dumb

Technici4n:
It's not because he has write access that he can do whatever he wants without breaking the trust of the community and the project leaders though.

liach:
:cry: Since I am out of the discord drama, I guess this only involves the api part without a plan for a separate toolchain like a separate mapping/gradle plugin/mod loader? Hope fabric doesn't die because of this

i509VCB:
I will not be using my write access FYI, that can be revoked when project administration finds time.

liach:
Sure. Sounds like you are just quitting other projects within fabric altogether? Good luck man

Earthcomputer:
Thanks liach, I'm sure our fork will do just fine. Whether fabric dies or not, you are of course welcome to help.

Technici4n:
Their plan is an entirely separate toolchain with compatibility with the official one, but we will see how that turns out.

Short term, this means less yarn and API contributions due to some people leaving, but that should not have a big impact on other contributors and most of the users, and there are things we can do to encourage more people to contribute, working through the backlog of open PRs would be a start... 😄

liach:
> entirely separate toolchain

Hmm, so you are going to write a new matcher as well, beyond everything else?

Per my understanding, fabric's existence is solely due to Matcher, which allows fabric to update across Minecraft versions efficiently way before the release of obfuscation maps. In contrast, MCP updated with the obfuscation maps offered by mojang in secret before manually.

Otherwise, the loader, mappings, intermediary, gradle plugin are all much more simple. Many comparable infrastructures exist. It is mainly the matcher that gave life to fabric for snapshots.

On a side note, is fabric really that inefficient in terms of, say, mappings? I don't find yarn too slow yet...

Earthcomputer:
Only some of what Technici4n said is correct.

Technici4n:
Uhhh they are forking at least loader and api, that's what I mean by "entirely separate" I guess. 😆

Anyway, we will see what comes out of this, but no worries there are still people interested in Fabric (the official one, that is).

liach:
Agreed. Imo this would be comparable to another implementation of jvm that is unrelated to hotspot; or at least like the "java" api in android. What comes out is the final manifest of this situation, and I am not disappointed enough to stop contributing yet :tm:.

Earthcomputer:
It's not really comparable to that. The most important aspect of other JVMs is that they are conformant to some spec. While it is true that we plan to be compatible with Fabric in the short-to-medium term, we plan to go above and beyond what Fabric is.

The reason this isn't going into Fabric itself cannot be ignored. It's because of Player's indirect suppression of minority groups, controversal handling of the discord since he was given owner, including banning and muting anyone who questions him. And then there is the pedophilia incident. Some may be able to turn a blind eye, but I personally can't.

We all remain committed to making this fork not about the issues of the past, but to think forward and keep implementing features for fun. Hopefully it will be technical reasons in the end which encourage people to switch.

liach:
> he was given owner

Did modmuss decide to retreat from the stage, or is this just so that fabric discord can get verified?

About player's personality... I've lived with him on a Minecraft 1.10 server 5 years ago. He was quite amicable! But you know, things change; I haven't looked at forge in a while, and industrialcraft 2 is still on 1.12. So can't say much on that.

Earthcomputer:
At the risk of Player coming here and saying this is all fake news...

I didn't know him 5 years ago, but my impression of him is that he is not the best fit for moderating a discord. He has made many controversial decisions without the consent and without consulting the moderators, and refuses to delegate any responsibility to anyone else.

Turning a blind eye to hate is not a good moderation policy, now it has grown into a much bigger issue about Player's abuse and centralization of power.

Technici4n:
I do not agree with some of the statements made by Earthcomputer regarding Player, ultimately they are an opinion and should be framed as such. You could use discord to get more context, but I am not sure this is a productive use of your time... If you want to continue the discussion, please do it elsewhere.

> Did modmuss decide to retreat from the stage

modmuss is taking a break atm.

> We all remain committed to making this fork not about the issues of the past, but to think forward and keep implementing features for fun. Hopefully it will be technical reasons in the end which encourage people to switch.

Good, I am always on the lookout for technical innovations.

Player3324:
FYI none of Earthcomputer's allegations are true, people were sanctioned for violating the rules repeatedly and knowingly. Anything not in line with our full tolerance demands has been and will always be moderated appropriately. We most definitely don't exclude anyone that follows the rules.

This is the same drama wave from the same crowd that has been besieging the Fabric Discord from its early days. They had plenty of opportunity to present any concrete cases that allegedly harmed minority groups and weren't handled by moderation. Not a single recent case was presented, only two old ones where moderation acted slowly but without proper notification. One of those old cases (centered around an alleged "trans<->pedo comparison") keeps being misrepresented in both what it really entails, its severity and omission of our sanction despite having a perfect written record. Attempts of eliding potential coverage holes are being ignored.

At the core of the drama wave is harassment against influential staff, plenty of defamation and a good amount of lies - directly, by omission or just misrepresentation in general. People who don't follow their line have been attacked or approached privately.

So overall we have a very vocal group that not only harms the Discord server and thus the project, but also demands draconian policies and recognition for some problem that hasn't been shown to exist. Of course we (mostly me) shut the drama wave down eventually.

Nobody can expect action or even empathy for behaving this way. Not having anything to point at for months speaks for itself.



Further misuse of the Github project spaces will be curbed.

Issue #575, opened 2020-04-14
[RFC] New module naming scheme


Prospector:
## Module Naming
We currently plan on moving forward with a new naming scheme for Fabric API modules which will finally bring consistency to the module names that up until this point have been mostly a free-for-all. The current naming scheme of `fabric-xxx` is confusing for 2 main reasons: 
- FabricMC has other projects such as `fabric-loader` and `fabric-loom` which are not Fabric API modules, yet follow identical name format to API modules
- A module with the name `fabric-biome-api` is much clearer of a name and more representative of what it does than `fabric-biomes`. As Player put it:
    > fabric-biomes adds 5 new lil' tater biomes

Modules that do not contain API (API in data form is still considered API, such as tags) will not have an `-api` suffix, such as `fabric-resource-loader-v1` and `fabric-indigo-renderer`

Event-based and registry-based modules will use the `-events` and `-registries` suffixes, respectively. These *are* API, however the `-api` suffix is redundant because events and registries are inherently API.

Modules names will always be written in plain, grammatical, U.S. English. In English, adjectives come before nouns so that is why modules that are currently named such as `fabric-renderer-indigo` or `fabric-events-lifecycle-v0` would be named `fabric-indigo-renderer` and `fabric-lifecycle-events` respectively under the new naming scheme.

Every name consisting of multiple words will have a hypen (`-`) between the words. This includes class names such as `BlockEntity` being represented as `block-entity`.

Module names are intended to be easy to type and remember. Grouping based on type of module (for example, event-based modules) can be part of a wiki article, instead of with a prefix in the naming scheme.

Modules should use `-api` instead of other words that essentially mean the same thing, unless they are especially significant (such as `-events` or `-registries`). For example, in `fabric-tag-extensions-v0`, the `-extensions` is fundamentally the same as `-api`, so it would be replaced with `fabric-tag-api-v1`

## Package Structure
Package structure will not change as much, and if anything just unify to the existing standards. Packages must follow the following format (using `fabric-biome-api-v1` as an example):
- API Packages use the root `net.fabricmc.fabric.api.biome.v1`
- Implementation Packages use the root `net.fabricmc.fabric.impl.biome`
- Mixin Packages use the root `net.fabricmc.fabric.mixin.biome`

**Note that all package names are singular and follow standard java conventions for package names**

Multi-word names will be split or combined depending on what makes sense from a hierarchical standpoint:
[...]

PR #1231, opened 2020-12-19
Added "Fabric tags"

gudenau:
Currently this has tags for:
 - Bookshelves
 - Totems of Undying

I figured these would be useful as it would allow mods to add objects that have tags to not use mixins and help with compatibility between mods. It would also allow for other mods such as custom enchanting table to easily support custom bookshelves.

I named the tag JSONs `bookshelf.json` and `totem_of_undying.json` because it is closer to the original names in the code, even though they could also be plural. (As well as `totems_of_undying` and `totem_of_undyings` booth seem kinda of awkward.)

Hopefully as time goes on there can be more things added to this module.

Reasons for mixins:
 - EnchantingTableBlockMixin: This is used to make the enchantment particles use the tag instead of a `Blocks` reference.
 - EnchantmentScreenHandlerMixin: This is used to make the bookshelf check use the tag instead of a `Blocks` reference.
 - ClientPlayNetworkHandlerMixin: This is used to get the totem of undying for the activation effect based on the tags instead of a `Items` reference. This fixes it always using the totem instead of the correct item.
 - LivingEntityMixin: This is used to change the totem check to use the tag instead of a `Items` reference. It also patches the statistics to use the correct item instead of always using the Totem.


haykam821:
Tags should be plural, i.e. `fabric:bookshelves` and `fabric:totems_of_undying`.

i509VCB:
I'm not sure this should be part of `tag-extensions-v0`. Maybe a different module?

haykam821:
This is a different module

fabric-tags-v0/build.gradle
```diff
@@ -0,0 +1,7 @@
+archivesBaseName = "fabric-tag-extensions-v0"
```
haykam821:
```suggestion
archivesBaseName = "fabric-tags-v0"
```

fabric-tags-v0/src/main/java/net/fabricmc/fabric/api/tags/FabricBlockTags.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.fabric.api.tags;
+
+import net.fabricmc.fabric.api.tag.TagRegistry;
+import net.minecraft.block.Block;
+import net.minecraft.tag.Tag;
+import net.minecraft.util.Identifier;
+
+public final class FabricBlockTags{
+	public static final Tag<Block> BOOKSHELF = TagRegistry.block(new Identifier("fabric-tags-v0", "bookshelf"));
```
haykam821:
```suggestion
	public static final Tag<Block> BOOKSHELVES = TagRegistry.block(new Identifier("fabric", "bookshelves"));
```

fabric-tags-v0/src/main/java/net/fabricmc/fabric/api/tags/FabricItemTags.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.fabric.api.tags;
+
+import net.fabricmc.fabric.api.tag.TagRegistry;
+import net.minecraft.item.Item;
+import net.minecraft.tag.Tag;
+import net.minecraft.util.Identifier;
+
+public final class FabricItemTags{
+	public static final Tag<Item> TOTEM_OF_UNDYING = TagRegistry.item(new Identifier("fabric-tags-v0", "totem_of_undying"));
```
haykam821:
```suggestion
	public static final Tag<Item> TOTEMS_OF_UNDYING = TagRegistry.item(new Identifier("fabric", "totems_of_undying"));
```

haykam821 (changes requested):

i509VCB:
Welp mistake then in my head.

Okay then:
- [ ] module should be v1
- [ ] singular module name (no tags)
- [ ] module name should be more descriptive (tags is vague)

gudenau:
Maybe "vanilla-tag-additions" or similar?

haykam821:
Would `fabric-content-tag` work? It might have too much overlap with tool tags though...

gudenau:
I'm bad with names so it doesn't really matter much to me. This is kind of a hard thing to name.

Also why v1, there are a bunch of v0 modules?

fabric-tags-v0/src/main/java/net/fabricmc/fabric/api/tags/FabricBlockTags.java
Virtuoel: Considering previous debates about tag identifier namespaces, I'd say having tags use the module's mod id instead of `fabric` might be better than just `fabric`. (also less worry about migration if/when the `fabric` mod id becomes `fabric-api` or similar)

i509VCB:
v0 refers to legacy modules. Per the RFC all new modules are v1: https://github.com/FabricMC/fabric/issues/575

i509VCB:
I know this isn't documented well, this will be mentioned in contributing docs in near future.

gudenau:
That's kinda silly but whatever.

gudenau:
That should have resolved everything except for the name issue.

fabric-tags-v1/src/main/java/net/fabricmc/fabric/api/tags/v1/FabricBlockTags.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * The place where the new block tags added by Fabric live.
+ */
+public final class FabricBlockTags{
+	/***
+	 * Allows for the addition of more types of Bookshelves via the tag system.
+	 *
+	 * <p>This allows the enchantment table particles and enchantment levels to work with custom blocks.
+	 */
+	public static final Tag<Block> BOOKSHELVES = TagRegistry.block(new Identifier("fabric-tags-v1", "bookshelves"));
```
LambdAurora, line 18:
```suggestion
	public static final Tag<Block> BOOKSHELVES = TagRegistry.block(new Identifier("fabric", "bookshelves"));
```

fabric-tags-v1/src/main/java/net/fabricmc/fabric/api/tags/v1/FabricItemTags.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * The place where the new item tags added by Fabric live.
+ */
+public final class FabricItemTags{
+	/***
+	 * Allows for the addition of more types of Elytra via the tag system.
+	 *
+	 * <p>The textures for the Elytra feature renderer are located at <code>assets/IDENTIFIER_NAMESPACE/textures/entity/elytra/IDENTIFIER_PATH.png</code>
+	 */
+	public static final Tag<Item> ELYTRA = TagRegistry.item(new Identifier("fabric-tags-v1", "elytra"));
```
LambdAurora, line 18:
```suggestion
	public static final Tag<Item> ELYTRA = TagRegistry.item(new Identifier("fabric", "elytra"));
```

fabric-tags-v1/src/main/java/net/fabricmc/fabric/api/tags/v1/FabricItemTags.java
```diff
@@ -0,0 +1,25 @@
...
+
+	/***
+	 * Allows for the addition of more types of Totems of Undying via the tag system.
+	 *
+	 * <p>The effect for the activation of the totem will automatically use the item that triggered it.
+	 */
+	public static final Tag<Item> TOTEMS_OF_UNDYING = TagRegistry.item(new Identifier("fabric-tags-v1", "totems_of_undying"));
```
LambdAurora, line 25:
```suggestion
	public static final Tag<Item> TOTEMS_OF_UNDYING = TagRegistry.item(new Identifier("fabric", "totems_of_undying"));
```

fabric-tags-v1/src/main/java/net/fabricmc/fabric/mixin/tags/extension/elytra/ElytraFeatureRendererMixin.java
```diff
@@ -0,0 +1,73 @@
...
+	@Shadow @Final private static Identifier SKIN;
+	private static Map<Item, Identifier> fabric_tags$TEXTURES;
+	private ItemStack fabric_tags$stack;
+	
+	@Inject(
+		method = "<clinit>",
+		at = @At("TAIL")
+	)
+	private static void staticInit(CallbackInfo ci){
+		fabric_tags$TEXTURES = new HashMap<>();
```
LambdAurora: would use a fastutil map

LambdAurora (changes requested):
Conceptually good. Though I would like to see some changes especially namespace wise.

Also checkstyle is really not happy.

fabric-tags-v1/src/main/java/net/fabricmc/fabric/mixin/tags/extension/elytra/ElytraFeatureRendererMixin.java
i509VCB: IdentityHashMap since item probably

gudenau:
Changes in this one:
- Checkstyle is less mad
- Using fastutils now
- Renamed packages to try and make Checkstyle happy but it failed

gudenau:
Fixed a couple minor whitespace mistakes that Checkstyle didn't catch.

fabric-tags-v0/src/main/java/net/fabricmc/fabric/api/tags/FabricBlockTags.java
Juuxel: Fabric tags do use `fabric` as the namespace already (tool tags like `#fabric:axes`).

Prospector:
It seems to me like the totem of undying one would fit into `fabric-item-api` nicely, however the bookshelves one brings up a new issue. Should an item tag for bookshelves also be added? Would it be confusing to use `fabric:bookshelves` for the block tag, but `c:bookshelves` for the item tag? And if an item tag was to be added as well, would the item tag for bookshelves go into a `fabric-block-api` with the block tag, or would they be split into different modules?

shedaniel:
Fabric tags are functional, I don’t see any reason to add the item tag for bookshelves, that serves no functions.

fabric-tags-v1/src/main/java/net/fabricmc/fabric/api/tags/v1/FabricItemTags.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * The place where the new item tags added by Fabric live.
+ */
+public final class FabricItemTags {
+	/**
+	 * Allows for the addition of more types of Elytra via the tag system.
+	 *
+	 * <p>The textures for the Elytra feature renderer are located at <code>assets/IDENTIFIER_NAMESPACE/textures/entity/elytra/IDENTIFIER_PATH.png</code>
+	 */
+	public static final Tag<Item> ELYTRA = TagRegistry.item(new Identifier("fabric-tags-v1", "elytra"));
```
OroArmor, line 18: As mentioned in discord, the unbreaking enchantment works differently on armor from all other tools, and there being no way to check this may cause some confusion from developers when they get reports of elytra damaging too quickly. Other than that, a pr to make elytras official is great because of all the redirects.

OroArmor (changes requested):
Splitting this into the item and block apis is better in my opinion to make it easier to find these features.

gudenau:
Honestly I'm fine if this gets split up and closed, as long as the features end up in the API in one form or another.

This was just a quick draft after all.

I do think we need to have the bookshelf feature in some form because you need to redirect it. So you can't have multiple mods that do it. I have a mod that I would like to make that would require this as well.

gudenau:
I'm going to close this since it seems to have been determined to be an incorrect fit and appears to have died.

Issue #1166, opened 2020-11-16
The Fluid API discussion issue


i509VCB:
This is an issue for discussing the fluid transfer api. Welcome here is discussion on implementation, api, documentation and use cases for this api. This is also loosely related to an item transfer api and the framework of an energy api (fabric api will not provide one explicitly). Discussion here should be on topic. If discussion is disorderly or misconduct is found this issue will be closed and locked to collaborators.

**What will Fabric API not do:**
Redefine cauldrons to be 4 bottles.
Game breaking changes
Design an api which completely prohibits alternative implementations from existing.

mounderfod:
We should use the next six months of snapshots to experiment with all our options, and see what works best in production.

yuesha-yc:
I will give a start on this. I have read a lot of discussions on the fluid volume debate, this debate will be really not necessary if Vanilla sets up a 1 Bucket = 4 Bottle / 8 Bottle / 10 Bottle standard... but anyway.

I hope we can discuss this proposal. 

# Volume API

## 1000MB

I personally think we should just stick with the old way Forge did: 1 Bucket = 1000 Liter.

**First and above all**, A 1000 Liter (milibucket) = 1 Bucket system is easy for normal players to memorize and understand since it uses a standard used in normal life (1000L = 1m^3). It is not counter-intuitive for a normal player! 

**Secondly**, I am not saying we should just clone what Forge did but I would say a similar fashion fluid volume api with essential functions that don't compel with the above listed three things Fabric API won't do is a good choice. Also, I believe it would ease people porting mods from older version of Minecraft or Forge. They don't have to worry about refactoring their fluid system entirely and can save a lot of time. Yes you can say porting from older version to either modern (1.14+) Forge or Fabric is nearly a rewrite, but it could save some time at least. It could also invite people who originally worked in Forge ecosystem to join Fabric. 

**An Issue I want to address**
There is one issue with this, and a long-discussed one, which is the 3 bottles = 1 cauldron equivalence set up by Vanilla that makes one Bottle = 333.3... Liter. I suggest Fabric API to not provide a default implementation of any kind for the bottle system. This can enable others who like to use the bottle system to create their own API. 

[...]

PR #1072, opened 2020-09-11
Fabric API Provider

grondag:
Replaces #1067 

This module allows API instances to be associated with game objects without specifying how the association is implemented. This is useful when the same API could be implemented more than once or implemented in different ways.

For example, a fluid API might be implemented for a block in the following ways:
- Using BlockState and some kind of wrapper (A vanilla cauldron implementation might do this)
- Directly in a BlockEntity for a simple tank
- As a more complicated data structure for a multi-block tank
- As one of several instances within a machine block containing multiple fluid buffers

 This variation is managed by a mapping function given when an API provider is registered. The provider interface
is meant to be extended as needed to handle a given use case.  (Sidedness, instance ID, permissions, etc)

Once registered, API providers can be retrieved without any knowledge of the API or provider implementations, and then used according to the contract of the provider to obtain an API instance.

Note that this model allows API providers to be associated with game objects that don't natively support them. This will require a wrapper or data attachment facility (or mixins) to be effective in most cases, all of which are outside the scope of this module. However, the decoupling of retrieval from implementation affords maximum flexibility for implementors.

This module could easily extended to cover additional game objects and probably should be if accepted.

This module could be implemented with slightly more efficiency if a data attachment facility were available, but that is left for another day. 

Here is a simple (and silly) example of usage:
```java
// The API we want to expose
interface Boop {
	void boop();
}

// Our absent provider
static final Boop NO_BOOP_FOR_YOU = () -> {};

// Our provider interface - in this case we don't rely on block entity
// Illustrates how providers can add the access parameters they need
@FunctionalInterface
interface BoopProvider extends ApiProvider<BoopProvider, Boop> {
	Boop getBoop(Direction side, PlayerEntity player);

	@Override
	default Boop getApi() {
		return NO_BOOP_FOR_YOU;
	}
}

// Our singleton provider access mapping function - real examples would usually get/create providers from game state
static final BoopProvider HAVE_A_BOOP_MAYBE = (side, player) -> (side == Direction.UP && player.isSneaking()) ? () -> System.out.println("Boop! Side = " + side.toString()) : NO_BOOP_FOR_YOU;

// Registration of our access along with our absent provider 
public static final BlockApiProviderAccess<BoopProvider, Boop> BOOP_ACCESS = BlockApiProviderAccess.registerAcess(new Identifier("boop:boop"), Boop.class, (side, player) -> NO_BOOP_FOR_YOU);

// Registration of our provider for a specific block
static {
[...]

PR #1234, opened 2020-12-21
Fabric API Lookup

### LINKED ISSUE #1199: CCA&LBA-like APIs should be considered in Fabric API
Currently, people use LBA and CCA to accomplish most functions that Capability in Forge provides. 

I suggest providing these types of API in Fabric API. 

I understand people worry about Fabric being to heavy, but I think this is essential for Fabric to grow better with better interop between mods, and allows larger mods with more content to be created more easily. **Many mod developers that are now considering upgrading to modern Minecraft Versions are hesitating to use Fabric because the lack of a Capability-like API in Fabric.** 

I would be glad to help if this is agreed upon most people.

Technici4n:
# Fabric API Lookup API v1
## Introduction
This module allows Api instances to be associated with game objects without specifying how the association is implemented. This is useful when the same Api could be implemented more than once or implemented in different ways.

Many thanks to @Grondag for providing the original concept (#1072).
Thanks also go to @i509VCB, @Pyrofab, @sfPlayer1 and the others who were involved with the design of this module.

This is the foundation upon which can be built for example a fluid transfer api (#1166). Closes #1199.

## Building blocks
This PR was changed a lot, please have a look at the README, the package info, and the javadoc for `BlockApiLookup` and `ApiLookupMap` for up-to-date documentation.

## More usage examples
FastTransferLib (https://github.com/Technici4n/FastTransferLib) is an experiment to build an item, fluid and energy transfer api on top of this module. (Which was until recently called `fabric-provider-api-v1`.)

## Missing things?
~~I could add an overload of `BlockApiLookup#find` with nullable `BlockState` and `BlockEntity` parameters, so that the caller can directly provide them if they are available for some reason.~~ Added in later commits.

There is no module to retrieve apis from items or entities yet because there were unsolved issues with those. The community can use the provided building blocks to experiment with their own implementations of `ItemStackApiLookup` and `EntityApiLookup` until the way forward becomes clear, but let's please not delay the `BlockApiLookup` because of that.


CoolMineman (approved):
Has been nice to use and works well

falseresync (approved):
Can find a use case for it already :+1:

liach (approved):
Fantastic. This API is concise and useful.

oliviathevampire (approved):

i509VCB (approved):
This is like obama putting the medal on himself meme but this API is good for merge to me.

Juuxel (approved):

LambdAurora (approved):

oliviathevampire (approved):

marcus8448 (approved):

pluiedev (approved):

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiLookupImpl.java
```diff
@@ -0,0 +1,160 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.lookup.v1.ApiProviderMap;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup;
+import net.fabricmc.fabric.mixin.lookup.BlockEntityTypeAccessor;
+
+public final class BlockApiLookupImpl<T, C> implements BlockApiLookup<T, C> {
+	private static final Logger LOGGER = LogManager.getLogger();
```
PepperCode1: Would adding a String name to this logger make it clearer for the user where the error came from?

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
```diff
@@ -0,0 +1,76 @@
...
+import net.fabricmc.fabric.impl.lookup.block.BlockApiLookupImpl;
+
+/**
+ * A {@link BlockApiLookup} bound to a {@link ServerWorld} and a position, providing much faster Api access.
+ * Refer to {@link BlockApiLookup} for example code.
+ *
+ * <p>{@link BlockApiLookup#get(World, BlockPos, BlockState, BlockEntity, Object) BlockApiLookup#get()} looks up
+ * the block state, the block entity, and the Api provider registered for the target block.
+ *
+ * <br>This object caches the block entity at the target position, and the last used Api provider, removing those queries.
```
PepperCode1: This is the only occurrence of `<br>`. Is it necessary for formatting or can it be replaced with a `<p>` for consistency?

PepperCode1 (approved):

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiLookupImpl.java
Juuxel: IMO `BlockApiLookupImpl` links the log messages to the module pretty well with `Lookup` in the name, but using the module ID could also be good (I've done that in the screen handler API).

YTG1234 (approved):

YTG1234 (approved):

hYdos (approved):
I need this

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
Technici4n: Thanks!

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiLookupImpl.java
Technici4n: Let's go for `fabric-api-lookup-api-v1`.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/ApiLookupMap.java
```diff
@@ -0,0 +1,55 @@
...
+import net.fabricmc.fabric.impl.lookup.ApiLookupMapImpl;
+
+/**
+ * Access to generic Api lookup instances. This is is meant to be used by implementors of Api lookup registries.
+ *
+ * <p>Note: To store a lookup class with type parameters {@code <T, C>}, it is recommended to store it as {@code <?, ?>} and perform an unchecked cast on queries internally.
+ *
+ * @param <L> The type of the lookup.
+ */
+public interface ApiLookupMap<L> extends Iterable<L> {
```
modmuss50: Unless im missing something I dont see these 2 interfaces needing to be exposed as API? They seem to be related to impl details?

modmuss50 (commented):
Test mod is excellent 👍 API looks solid to me, I have yet to review the impl details.

Just one minor query so far.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiCacheImpl.java
```diff
@@ -0,0 +1,121 @@
...
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerBlockEntityEvents;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiCache;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup;
+
+public final class BlockApiCacheImpl<T, C> implements BlockApiCache<T, C> {
+	private final BlockApiLookupImpl<T, C> lookup;
+	private final ServerWorld world;
+	private final BlockPos pos;
+	/**
+	 * We always cache the block entity, even if it's null. We rely on BE load and unload events to invalidate the cache when necessary.
```
modmuss50, line 35: What happens if you try to cache something that isnt a BE? registred via `registerForBlocks`.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/ApiLookupMap.java
Technici4n: They are indeed implementation details, but they are provided to accelerate implementations of other types of api lookups.

See for example the third-party https://github.com/Technici4n/FastTransferLib/blob/transactions/src/main/java/dev/technici4n/fasttransferlib/impl/item/ItemKeyApiLookupRegistryImpl.java and https://github.com/Technici4n/FastTransferLib/blob/transactions/src/main/java/dev/technici4n/fasttransferlib/impl/item/ItemKeyApiLookupImpl.java for an example of how it can be used by other mods.

The plan is to add item and entity api lookups as well in later PRs, but this could still be useful in other contexts, basically any time you want to convert a game object to some type `T`. For example, someone might want to expose a `WorldApiLookup`.

modmuss50: Ok that’s cool. Just don’t want to expose something for the sake of it, but this is fine.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiCacheImpl.java
Technici4n: If there is no block entity at the target position, the following should happen, matching the behavior of `BlockApiLookup`:
- If the block state is not provided to the cache it will be queried in the world.
- If the provider for the block state is a `WrappedBlockEntityProvider`, which is not supposed to happen, it should probably log an error and not pass a null BE. (I should change that). Otherwise, the provider will receive `(world, pos, state, context)`.
- The fallback providers will receive a `null` block entity instance.

Not sure if that's the explanation you expected. If needed, I can probably explain better on Discord, just ping me.

modmuss50: I really need to boot it up and play with it. Easier to find out that way :)

shedaniel (approved):

Ueaj-Kerman (approved):

modmuss50 (approved):
Had a play around with it, and cannot see anything wrong with the IMPL.

fabric-api-lookup-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "depends": {
+    "fabricloader": ">=0.9.2",
+    "fabric-api-base": "*",
+    "fabric-lifecycle-events-v1": "*"
+  },
+  "description": "A universal way to expose and query Apis",
+  "mixins": [
+    "fabric-api-lookup-api-v1.mixins.json"
+  ]
```
i509VCB: This module now needs to define a module lifecycle:
```json
  "custom": {
  }
```

LemmaEOF:
Is this superceded by #1278, and if so, should this one be closed?

fabric-api-lookup-api-v1/README.md
```diff
@@ -0,0 +1,28 @@
+#  Fabric Api Lookup API (v1)
```
LemmaEOF: copying my non-merge-blocking bikeshed from 1278:

I'm so sorry but I honestly think I need to commit bikeshed and say argue that "Api Lookup Api" is a terrible name that, while being very literal, is clunky to use and ends up hurting ease of understanding of the code due to the redundancy of the term "api". Calling it a component API isn't quite accurate either, since entity-component systems are inherently about managing storage as well as lookup. I think the best term to use here would be "property", as it's looking or traits of an object without caring about how they're stored.

LemmaEOF (approved):

Technici4n:
This PR is the prerequisite for #1278, which includes the commits of this PR. #1278 will be much easier to review when this is merged.

fabric-api-lookup-api-v1/README.md
Technici4n: I think the only redundancy of the term "api" is in the name of this module, which is sometimes referred to as "API API" anyway (an API to expose APIs). In code, it only appears as `ApiLookup` or `ApiCache`.

As for the exact name, it may be a bit literal, but I think it conveys very well that this API is just meant to look up API instances from objects, I think it's more general than property.

oliviathevampire (approved):

fabric-api-lookup-api-v1/README.md
liach: how about "interface lookup api" or "hook lookup api"

Technici4n: I don't think those are really better. In theory, an Api doesn't necessarily have to be an interface. Also, it's mostly bikeshedding at this point... :-P

Azzyypaaras:
Legend has it player will some day look at this

williambl (approved):

liach:
Make sure you carry over changes suggested in #1278 back here

Technici4n:
Nothing yet as far as I know.

liach:
I recall ItemKey was at least changed.

Technici4n:
The item module is not part of this PR, it will be for another one ;)

valoeghese:
i dunno feels unneccesary

Ueaj-Kerman:
Do you want flapi or not smh

valoeghese:
Ok on second inspection it’s more useful than I thought but like why not just use mixin 😎

liach:
> 
> 
> Ok on second inspection it’s more useful than I thought but like why not just use mixin 😎

lol you didn't read the pr at all, this api is nothing about mixin

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/ApiLookupMapImpl.java
```diff
@@ -0,0 +1,73 @@
...
+	public synchronized Iterator<L> iterator() {
+		return lookups.values().stream().map(storedLookup -> storedLookup.lookup).collect(Collectors.toList()).iterator();
+	}
+
+	private static class StoredLookup<L> {
+		private final L lookup;
+		private final Class<?> apiClass;
+		private final Class<?> contextClass;
+
+		private StoredLookup(L lookup, Class<?> apiClass, Class<?> contextClass) {
```
liach:
```suggestion
	private static final class StoredLookup<L> {
		final L lookup;
		final Class<?> apiClass;
		final Class<?> contextClass;

		StoredLookup(L lookup, Class<?> apiClass, Class<?> contextClass) {
```
This ain't java 11, no nest host yet

Technici4n: No idea what the difference is. I don't think `final` is important for an impl class?

liach: 1. `final` on classes can make jvm better optimize pojos
2. Removing `private` from inner class members (not class itself, as the class modifiers are effectively package and public) allows the compiler to not generate bridge methods for each place where the private field is accessed or the private constructor is called. fyi the accessor in compiled java looks like `access$$0` something, while the constructors have some useless arguments inserted by compiler to avoid descriptor conflicts (all constructors are named `<init>`, so descriptors are their only difference). 
   - In Java 11, the accessor problem is gone as jvm would recognize nest mates for private access and compiler no longer has to generate bridges. But remember, we are on Java 8

Technici4n: I'm pretty sure the JVM already recognizes effectively `final` classes, but this change can't hurt anyway, and sure it's a good idea to remove synthetic methods as well. :)

valoeghese:
> > Ok on second inspection it’s more useful than I thought but like why not just use mixin 😎
> 
> lol you didn't read the pr at all, this api is nothing about mixin

It's about attaching data to stuff right?

i509VCB:
> It's about attaching data to stuff right?

Yes and no.

The API exposes atleast the way to obtain the data, but attachment is an implementation detail to the mod which wants to provide said data.

Player3324:
Some suggestons:

1. move ApiLookupMap and ApiProviderMap to a subpackage `custom`, similar to `block`, to get these niche classes out of view.
2. replace BlockApiLookupRegistry with a static `BlockApiLookup.get` to avoid too many levels/kinds of "register" and reduce the class count
3. rename existing `BlockApiLookup.get` to `BlockApiLookup.find` to disambiguate with the above static BlockApiLookup instance getter and highlight that a non-null result is really not to be expected
4. Javadoc needs some work. Info such as "Register a {\@link BlockApiProvider} for some blocks." is too shallow. it needs to directly tell the high level purpose. In the example that'd be potentially exposing the API for the given blocks.
5. Maybe we can use A for API instead of T and generally talk about "the API" instead of "the queried object"?
6. API lifetime handling is a weak point atm. Holders with individual instances (BE, entity, chunk, world etc) can usually track their own lifecycle properly, which would allow to bypass all the logic in BlockApiCacheImpl if there was some link and a way to opt into strong management. BlockApiCacheImpl having only a single instance per world+pos would help there.

Technici4n:
1/2/3/4/5. Absolutely, I will apply those this week.
6. Not sure what we can do about the caching. It's already very fast and extremely simple to use, and more control is too error-prone imo. If absolutely necessary, the caching API is so opaque atm that we can add more specialized methods later, but please let's not do that until queries are proven to be a bottleneck in real-world benchmarks, which we won't have until weeks after version 1.0.0 of this module is merged anyway.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
```diff
@@ -40,25 +40,30 @@
...
 public interface BlockApiCache<A, C> {
 	/**
-	 * Retrieve an Api from a block in the world, using the world and the position passed at creation time.
+	 * Attempt to retrieve an API from a block in the world, using the world and the position passed at creation time.
 	 *
-	 * <p>Note: If the block state is known, it is more efficient to use {@link BlockApiCache#get(BlockState, Object)}.
+	 * <p>Note: If the block state is known, it is more efficient to use {@link BlockApiCache#find(BlockState, Object)}.
+	 *
+	 * @param context Additional context for the query, defined by type parameter C.
+	 * @return The retrieved API, or {@code null} if no API was found.
```
liach, line 50:
```suggestion
	 * @param context additional context for the query
	 * @return the retrieved API, or {@code null} if no API was found
```
Imo you should move the description of type `C` to the `@param` tags for class doc

I think we are more inclined to use sentence fragments for the param and return tags than full sentences. This just serves as an example.

Technici4n:
Ok, I have modified the PR quite heavily, and I think it's ready for another (final I hope) round of review. Notable changes:
* `ApiLookupMap` and `ApiProviderMap` have been moved to a new `custom` subpackage and have received a very complete usage example.
* More generally, javadoc should be much clearer now.
* Functions related to API queries are now called `find`, and `BlockApiLookup#get` is used instead of `BlockApiLookupRegistry#getLookup`.
* `BlockApiLookup` now has a `registerSelf` method that can be used in cases where the block entity directly implements the API. The testmod has been updated to test this too.
* `//CHECKSTYLE.OFF: <InspectionName>` and `//CHECKSTYLE.ON: <InspectionName>` can now be used to toggle specific checkstyle inspections because they were causing me trouble with the class javadoc of `ApiLookupMap`.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
Technici4n: I find `context additional context for the query` much harder to parse than `context Additional context for the query, defined by type parameter C.` as it's harder to directly tell where the parameter name ends and where the description starts.

Technici4n: Also, the whole PR is consistent with itself and it's an entirely new module - nobody complained the first time and I really don't feel like going through all of the very extensive javadoc one more time to make it less readable in the end.

liach: You should just document what `C` is in class's javadoc. You have the C as parameter type, so people already know it's C

liach:
Also why is the generic api lookup/provider map moved to a `custom` package? It feels generic enough for general purposes

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
Technici4n: In this particular case, the parameters are not documented in `BlockApiCache` because they are the same as those of `BlockApiLookup`, and the cache already refers to it. I added this comment in `BlockApiCache` to be consistent with `BlockApiLookup#find` where it's imo necessary anyway due to `find` being one of the API entrypoints, and people might miss the `@param <C>` at the end of the huge javadoc.

Technici4n:
Player suggested moving it to a `custom` subpackage to reduce its visibility so that people browsing through the API would look at the `block` package first.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
```diff
@@ -0,0 +1,129 @@
...
+ * <pre>{@code
+ * public interface ItemStackApiLookup<A, C> {
+ *     static <A, C> ItemStackApiLookup<A, C> get(Identifier lookupId, Class<A> apiClass, Class<C> contextClass) {
+ *         return ItemStackApiLookupImpl.get(lookupId, apiClass, contextClass);
+ *     }
+ *     // Find an API instance.
+ *     ＠Nullable
+ *     A find(ItemStack stack, C context);
+ *     // Expose the API for some item.
+ *     void register(ItemStackApiProvider<A, C> provider, Item item);
```
i509VCB, line 44: the reason checkstyle complained here is you need to use the html entity codes for `<` and `>`

Technici4n: Then it fails to render correctly as the `{@code}` tag already escapes them, I tried that first...

i509VCB: It won't render in IDE but the doclet will render it

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
liach: > might miss the `@param <C>` at the end of the huge javadoc

All parameters are generated alike as they are block tags whose contents are appended to end of member output. Per your argument, you shouldn't add param tags for method parameters because they might be missed.

Imo it's still best to document them in class. If you document the params, javadoc will generate the link for param `C` so when you click the clickable `C` in method parameter types, you will be redirected to the definition of the type parameter at the top of the class

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
Technici4n: `gradlew javadoc` was unable to render them correctly when I tried on my laptop - just disable the inspection and be done with it, there are more important things in life and javadoc escaping is weird anyway. :P

modmuss50:
Oops sorry @liach i didn’t mean to add you as a reviewer again. Feel free to though :)

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
Technici4n: IntelliJ does not seem to link the class directly, so I kept that comment in the javadoc of the `find` methods, but I also added the params to `BlockApiCache` for extra clarity and for web browser javadoc users. Hope that works for you.

liach (approved):
My previous questions are addressed already. Just a side note that when we add apiNote etc. command line tags, we should probably move `<p>Note: ` to those tags.

Technici4n:
I moved away from using `@apiNote` because `gradlew javadoc` tool didn't seem to recognize it.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
Technici4n: Fun fact: the `＠` symbol in the javadoc is not a regular `@` 😆 (https://stackoverflow.com/a/46332643/13567109).

liach:
> 
> 
> I moved away from using `@apiNote` because `gradlew javadoc` tool didn't seem to recognize it.

This is what you need https://github.com/FabricMC/yarn/blob/0025fbf845a27efa50d44ba1e2bac7c028eea57d/build.gradle#L722-L726

Technici4n:
> 
> 
> > I moved away from using `@apiNote` because `gradlew javadoc` tool didn't seem to recognize it.
> 
> This is what you need https://github.com/FabricMC/yarn/blob/0025fbf845a27efa50d44ba1e2bac7c028eea57d/build.gradle#L722-L726

Ok that's interesting, will be good to have eventually.

grondag:
I've scanned through this briefly and it seems consistent with the original intent and well-considered in its implementation.  Nice job.

Would it support client-side lookups?  From what I read it appeared it may be server-side only.

Technici4n:
Thanks! Client-side lookups are possible, it's just that the `BlockApiCache` will not be usable with them because it takes a `ServerWorld`.
The regular `BlockApiLookup` takes a `World` and should work fine.

Player3324 (approved):
Good now as far as I can tell. Thanks!

Mysterious-Dev (approved):

marcus8448 (approved):

PR #1237, opened 2020-12-22
Fix example implementation of using FAPI modules for Groovy

maityyy:
The previous implementation does not quite correctly describe what it does

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ modImplementation "net.fabricmc.fabric-api:fabric-api:FABRIC_API_VERSION"
 modImplementation("net.fabricmc.fabric-api:fabric-api:FABRIC_API_VERSION")
 ```
 
-Alternatively, modules from Fabric API can be specified individually as shown below:
+Alternatively, modules from Fabric API can be specified individually as shown below (including module jar to your mod jar):
 
 ### Groovy DSL
 
@@ -69,7 +69,7 @@ setOf(
     "fabric-networking-v0"
 ).forEach {
     // Add each module as a dependency
-    modImplementation(fabricApi.module(it, FABRIC_API_VERSION))
+    include(modImplementation(fabricApi.module(it, FABRIC_API_VERSION)))
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ setOf(
     "fabric-networking-v0"
 ).forEach {
     // Add each module as a dependency
-    include(modImplementation(fabricApi.module(it, FABRIC_API_VERSION)))
+    modImplementation(fabricApi.module(it, FABRIC_API_VERSION))
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ Set<String> apiModules = [
 
 // Add each module as a dependency
 apiModules.forEach {
-    modImplementation(fabricApi.module(it, FABRIC_API_VERSION))
+    include(modImplementation(fabricApi.module(it, FABRIC_API_VERSION)))
```
i509VCB, line 57: I believe we should mention the fact this includes the module inside the built jar also with that change.

Otherwise yes this is fine

i509VCB (approved):

i509VCB:
I'll amend the kotlin dsl section when this is merged accordingly

README.md
maityyy: I initially thought that it should be so. But the original code didn't work for me at all, so I added "include" and it worked. Perhaps it was my mistake. Ok, I will describe your request a little later ;)

maityyy: Will my change work for Kotlin (add "include")? I haven't worked with Kotlin, so I don't know. It's just that these two codes should produce the same result.

i509VCB: Not exactly. Kotlin's type checking would not exactly like that.

In kotlin `modImplementation("blah")` and any other gradle configuration returns a `Dependecy?` which is nullable. This won't work if you chain the `include` like you do in groovy. You'd either need to make a custom configuration and add that configuration to both `modImplementation` and `include` or use the `?.let { include(it) }` syntax after the last bracket on `modImplementation(...)`

maityyy: ~~Oh, I'll fix mistake I made in the new commit now.~~

maityyy:
> I'll amend the kotlin dsl section when this is merged accordingly

Didn't see your message, hah. Okey.

Juuxel:
@i509VCB I'm pretty sure you can chain dependencies (with `include(modImplementation(...))`), as I do that already in Adorn: https://github.com/Juuxel/Adorn/blob/42cb0e3c1efbf435103e10b169ab5d0e14946ecb/build.gradle.kts#L66

i509VCB:
It doesn't really work for me juuz. I was going to suggest possibly just creating a new configuration in example but that does add complexity to the process.

Example:
```kotlin
val modApiAndInclude by configurations.register("modApiAndInclude")

dependencies {
    ...
    modApiAndInclude("blah:blah:1.0.0")
    ...
    add(sourceSets.main.get().getTaskName("mod", JavaPlugin.API_CONFIGURATION_NAME), modApiAndInclude)
    add(Configurations.INCLUDE, modApiAndInclude)
}
```

Atleast in my case the only thing that works there is the `?.let { include(it) }` syntax

Juuxel:
Huh. Maybe different overloads have different nullabilities? Considering it's Gradle, I wouldn't be that surprised...

modmuss50 (approved):

PR #1238, opened 2020-12-23
Allow passing UV coordinates as Vec2f

yyny:
Rotating UV coordinates using vector math becomes extremely annoying when every `sprite` call requires the individual `u` and `v` coordinates to be extracted again.

This change will allow passing UV coordinates in a container that will always be available.

### DIFF
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/api/renderer/v1/mesh/MutableQuadView.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/api/renderer/v1/mesh/MutableQuadView.java
@@ -22,6 +22,7 @@ import net.minecraft.client.render.model.BakedQuad;
 import net.minecraft.client.texture.Sprite;
 import net.minecraft.client.util.math.Vector3f;
 import net.minecraft.util.math.Direction;
+import net.minecraft.util.math.Vec2f;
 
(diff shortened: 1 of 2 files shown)


liach (approved):

Player3324 (approved):
Why not I guess, I hope nobody will create Vec2fs specifically to call this though.

modmuss50 (approved):

yyny:
Great concern, we can add a javadoc comment if this is an issue, but I doubt many people will stumble upon this method unless they already have a vector they want to pass.

oliviathevampire (approved):

Juuxel (approved):

i509VCB (approved):
Please do add the javadoc to that method about allocation.

grondag:
As Player already noted,  I wouldn't allocate a `Vec2f` just for this but if you already have one, it's handy enough.

I suggest removing the `spriteIndex` parameter (use 0) because Indigo does not support it, FRAPI defines no behavior for it, and I've removed support for it in FREX/Canvas as well.

It was convenient to use when it was supported, but the lack of consistency was creating unnecessary complications.  I intend to submit a PR to deprecate and replace the methods that accept `spriteIndex` and then hopefully remove them in the 1.17 cycle.

grondag (approved):

PR #1240, opened 2020-12-23
Fix a number of issues with reload listeners

modmuss50:
- Fix resource reload listeners not being sorted.
- Fix missing/broken reload listener keys.
- Add a debug log line for missing reload keys
- Add some basic tests.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceReloadListenerKeys.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceReloadListenerKeys.java
@@ -30,9 +30,6 @@ public final class ResourceReloadListenerKeys {
 	public static final Identifier MODELS = new Identifier("minecraft:models");
 	public static final Identifier LANGUAGES = new Identifier("minecraft:languages");
 	public static final Identifier TEXTURES = new Identifier("minecraft:textures");
-	public static final Identifier ITEM_RENDERER = new Identifier("minecraft:item_renderer");
-	public static final Identifier BLOCK_RENDERER = new Identifier("minecraft:block_renderer");
-	public static final Identifier WORLD_RENDERER = new Identifier("minecraft:world_renderer");
 
 	// server
 	public static final Identifier TAGS = new Identifier("minecraft:tags");
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceReloadListenerKeys.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceReloadListenerKeys.java
@@ -30,6 +30,9 @@ public final class ResourceReloadListenerKeys {
 	public static final Identifier MODELS = new Identifier("minecraft:models");
 	public static final Identifier LANGUAGES = new Identifier("minecraft:languages");
 	public static final Identifier TEXTURES = new Identifier("minecraft:textures");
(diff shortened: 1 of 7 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceReloadListenerKeys.java
```diff
@@ -30,6 +30,9 @@
 	public static final Identifier MODELS = new Identifier("minecraft:models");
 	public static final Identifier LANGUAGES = new Identifier("minecraft:languages");
 	public static final Identifier TEXTURES = new Identifier("minecraft:textures");
+	public static final Identifier ITEM_RENDERER = new Identifier("minecraft:item_renderer");
+	public static final Identifier BLOCK_RENDERER = new Identifier("minecraft:block_renderer");
+	public static final Identifier WORLD_RENDERER = new Identifier("minecraft:world_renderer");
```
i509VCB: Should be `Identifier("minecraft", "blah")`

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/client/KeyedResourceReloadListenerClientMixin.java
```diff
@@ -36,11 +38,13 @@
 
 @Mixin({
 		/* public */
-		SoundLoader.class, BakedModelManager.class, LanguageManager.class, TextureManager.class,
+		SoundManager.class, BakedModelManager.class, LanguageManager.class, TextureManager.class,
 		/* private */
 		WorldRenderer.class, BlockRenderManager.class, ItemRenderer.class
```
i509VCB, line 41: You added keys for some of these?

fabric-resource-loader-v0/src/testmod/java/net/fabricmc/fabric/test/resource/loader/ResourceReloadListenerTestMod.java
```diff
@@ -0,0 +1,118 @@
...
+	private static boolean serverResources = false;
+
+	@Override
+	public void onInitialize() {
+		setupClientReloadListeners();
+		setupServerReloadListeners();
+
+		ServerTickEvents.START_WORLD_TICK.register(world -> {
+			if (!clientResources) {
+				throw new RuntimeException("Client reload listener was not called.");
```
i509VCB: Throw `AssertionError` instead of runtime

i509VCB (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/client/KeyedResourceReloadListenerClientMixin.java
modmuss50: All of these either had keys, or do now.

i509VCB: Well my assumption was the private comment meant there was no key, so the mixins should be moved up?

modmuss50: Oh, I see what you mean. IDK why they were private before tbh. I might undo that for now, as this is supposed to be a bug fix PR.

PR #629, opened 2020-05-21
Loot table API v2

Juuxel:
- The main change is that v2 uses new 1.16 Yarn names for all methods, and the previously renamed `LootTable` instead of `LootSupplier` as well.
- All uses of reflection were removed.
- v2 supports the entire API of v1 with the exception of `LootJsonParser` and `LootEntryTypeRegistry` which have become useless in Minecraft updates. (Now you can just use `LootGsons` and `Registry.LOOT_POOL_ENTRY_TYPE` from vanilla.)
- Added a few new methods in `FabricLootPoolBuilder` that work with collections of children.
- Changes to v1:
  - v1 was retrofitted to always fall back to v2 when possible.
  - v1's `LootEntryTypeRegistry` was changed to directly register to the vanilla registry.

PR #160, opened 2019-04-27
Loot table manipulation API

asiekierka:
This follows on the work laid out in #77 with some changes:

* Adapted to new directory format
* The mixins now use Accessors
* LootUtilities -> LootJsonParser with some naming clean-up
* LootEntryTypeRegistry is now a part of api.loot.v1 (with deprecation)
* LootTableLoadingCallback now receives the LootManager instance, allowing to get other loot tables than one's own.
* LootTableLoadingCallback now applies modifications at the end of all processing, which means LootManager can be read from safely.
* LootTableLoadingCallback received a setter, allowing you to override the loot table instance being iterated on completely. I am not sure if that is a good idea, but I think that's the only way we can faciliate removals at this time?

PR #1241, opened 2020-12-24
Loot table API v2, iteration 2

Juuxel:
**Edit (13 May 2022):** Now merged with 1.18.2 branch and official TAW module. Also includes an internal API for resource tracking ported from #1983.

**Edit (31 Jan 2022):** This PR has now been updated to 1.18.1, uses interface injection and also creates the TAW module. The multipurpose `LootTableLoadingCallback` was replaced with two events with clear purposes (replacing or modifying).

**Edit (27 Nov 2021):** After 18 months of working this API, I made another refactor and started using transitive AWs instead of static accessor classes. They'll be merged into the TAW module once that comes along, so this is a draft for now.

## What does the loot table API do?
The main use case of this API is adding items/loot pools to loot tables (example: modded seeds to vanilla grass) or modifying them in some other way. There are also extended and more convenient versions of the vanilla loot table/pool builders, like in v1.

This PR is also a replacement for #629 with a different approach. This one is also backwards compatible (I had some trouble with that using the same design).

## Changes from V1

This API uses modern Yarn names and avoids "magic interfaces" implemented on vanilla classes. It also drops some unnecessary functionality that is fully public in modern vanilla.

### Transitive access wideners instead of magic interfaces and static access classes

The magic interfaces `FabricLootSupplier` and `FabricLootPool` are replaced with **transitive access wideners**. I also dropped some functionality that is now available in vanilla (`LootJsonParser` and `LootEntryTypeRegistry`).

- The returned lists from V1 magic interfaces are now immutable. The mutability was not documented at all, and this is safer.
- The access wideners also expose the bonus rolls of a loot pool.
- The implementation is much simpler now for V1 as well:
  - `LootEntryTypeRegistry` no longer uses mixins but calls `Registry.register` directly.
  - `LootJsonParser` no longer uses reflection.

### Injected builder interfaces

In the new V2 module, there are `FabricLootPoolBuilder` and `FabricLootTableBuilder` to replace the V1 builders.

- They are now interfaces that are injected to `LootPool.Builder` and `LootTable.Builder`, respectively.
- Minor renames: `of(LootTable/Pool)` -> `copyOf`
- The loot pool builder now has methods for working with collections as parameters.
- All `copyFrom` methods are gone. They had weird and arbitrarily chosen boolean flag parameters, and their few use cases can be replicated using v2 API methods.

### Changes to events

`LootTableLoadingCallback` has been replaced with `LootTableEvents.REPLACE` and `LootTableEvents.MODIFY` which clearly separate the two different purposes. `REPLACE` runs before `MODIFY`.

- The callback loop is now stopped when a loot table is replaced. This isn't really a breaking change to v1 as that behaviour was not defined at all.
- You can access a loot table's source in the events. This can be used to only apply modifications to built-in resources and not user-overridden ones, for example.
  - `VANILLA` and `MOD` for built-in content
  - `DATA_PACK` for external data packs
  - `REPLACED` for loot tables created in `LootTableEvents.REPLACE` (naturally only exists for loot tables processed in `LootTableEvents.MODIFY`)
  - This is implemented via a new internal API in resource loader for resource source tracking. The `FabricResource` interface, which might be promoted to public API in the future, exposes the `ResourcePackSource` of the `ResourcePack` instance that created the `Resource`.


fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/event/LootTableLoadingCallback.java
```diff
@@ -45,5 +45,6 @@
 			}
 	);
 
+	// TODO: What should I do about FabricLootSupplierBuilder here?
```
i509VCB: New event sadly :(

fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/FabricLootTable.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.loot.LootPool;
+import net.minecraft.loot.LootTable;
+import net.minecraft.loot.context.LootContextType;
+import net.minecraft.loot.function.LootFunction;
+
+/**
+ * Accessor methods for loot table properties.
+ * This interface is automatically implemented on all {@link LootTable} instances.
+ */
+public interface FabricLootTable {
```
i509VCB: Ideally this interface should need to be exposed in API, a helper class with static `getBlah` methods would suffice here and be more resilient to api changes.

fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/FabricLootTableBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * An extended version of {@link LootTable.Builder}.
+ */
+public class FabricLootTableBuilder extends LootTable.Builder {
+	private final LootSupplierBuilderHooks extended = (LootSupplierBuilderHooks) this;
+
+	protected FabricLootTableBuilder() {
+	}
+
+	private FabricLootTableBuilder(LootTable supplier) {
```
i509VCB:
```suggestion
	private FabricLootTableBuilder(LootTable lootTable) {
```

fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/event/LootTableLoadingCallback.java
Juuxel: Yeah, I don't think there's another way. I'll make the new event be in the v1 root package and register this event's invoker as a callback for it.

fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/FabricLootTable.java
Juuxel: I think you mean this interface *shouldn't* be exposed in the API? For that, I can make a `LootTables` class (prefixed with Fabric if needed) with a bunch of getters, and maybe do the same to loot pools too.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPoolBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+
+		return this;
+	}
+
+	/**
+	 * Creates an empty loot pool builder.
+	 *
+	 * @return the created builder
+	 */
+	public static FabricLootPoolBuilder builder() {
```
liach:
```java
FabricLootPoolBuilder.builder()
```
looks a bit weird in usage.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootTableBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+
+		return this;
+	}
+
+	/**
+	 * Creates an empty builder.
+	 *
+	 * @return the created builder
+	 */
+	public static FabricLootTableBuilder builder() {
```
liach: Same, `FabricLootTableBuilder.builder()` sounds a bit weird

liach (approved):

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPoolBuilder.java
i509VCB: I think in this case since the builder is being made it would become `FabricLootPoolBuilder.create()`

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPools.java
```diff
@@ -0,0 +1,58 @@
...
+		Objects.requireNonNull(pool, "pool cannot be null");
+		return ((LootPoolAccessor) pool).getRolls();
+	}
+
+	public static List<LootPoolEntry> getEntries(LootPool pool) {
+		Objects.requireNonNull(pool, "pool cannot be null");
+		return ImmutableList.copyOf(((LootPoolAccessor) pool).getEntries());
+	}
+
+	public static List<LootCondition> getConditions(LootPool pool) {
```
i509VCB: javadoc needs to explain the resulting lists are immutable

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableLoadingCallback.java
```diff
@@ -0,0 +1,65 @@
...
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * A callback that is called when loot tables are loaded.
+ * Use {@link #EVENT} to register instances.
+ */
+@FunctionalInterface
+public interface LootTableLoadingCallback {
+	Event<LootTableLoadingCallback> EVENT = EventFactory.createArrayBacked(
```
i509VCB: is there a possibility for future loot table events?

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableLoadingCallback.java
```diff
@@ -0,0 +1,65 @@
...
+	);
+
+	/**
+	 * Called when a loot table is loading.
+	 *
+	 * @param resourceManager the server resource manager
+	 * @param lootManager     the loot manager
+	 * @param id              the loot table ID
+	 * @param tableBuilder    a builder of the loot table being loaded
+	 * @param setter          a loot table setter for completely replacing the loaded loot table
```
i509VCB: does setting the replacement cancel future notification or does it silently change return result. I imagine this could be an area of contention a mod or two may fight over.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPoolBuilder.java
Juuxel: I'll change these to `create`.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPools.java
Juuxel: Oops, forgot to document these ones. I'll add that explanation as well.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableLoadingCallback.java
Juuxel: There's nothing that comes to my mind at least, but any future ones can be PR'd later. (I think a top-level callback containing its own `EVENT` field is fine for now)

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableLoadingCallback.java
Juuxel: It silently changes the result, which is problematic IMO (and also according to @asiekierka who added it in #160). We could figure out a better solution now.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPoolBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+import net.minecraft.loot.entry.LootPoolEntry;
+import net.minecraft.loot.function.LootFunction;
+
+import net.fabricmc.fabric.mixin.loot.LootPoolBuilderAccessor;
+
+/**
+ * An extended version of {@link LootPool.Builder}.
+ */
+public class FabricLootPoolBuilder extends LootPool.Builder {
+	private final LootPoolBuilderAccessor access = (LootPoolBuilderAccessor) this;
```
i509VCB: I'd probably add an internal comment here about implementors needing to add any new methods to this and the copyOf method.

fabric-loot-table-api-v2/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,26 @@
...
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.4.0",
+    "fabric-api-base": "*"
+  },
+  "description": "Hooks for manipulating loot tables.",
+  "mixins": [
+    "fabric-loot-table-api-v2.mixins.json"
+  ]
```
i509VCB: This module now needs to define a module lifecycle:
```json
  "custom": {
  }
```

liach:
Hmm, just curious, how are priorities assigned?

<details>
<summary>About why I don't use discordapp anymore</summary>
Discord is problematic. I don't know if my account is even there because it may get nuked when one of the guilds I'm in gets nuked. It causes extraneous distraught and has very few meaningful discussion compared to like github discussion, irc (no bouncer still so not using as of now), or mailing lists.
</details>

i509VCB:
I'll detail priorities in a bit, currently it is somewhat along the lines of urgency

maityyy:
Is there some way to replace pool entries? It seems that the current API does not have this flexibility and it came to me to use AccessWidener to get access to many private fields...

maityyy:
`ItemEntry` has `item` field and it is private. It is rather inconvenient to replace one pool entry with another, because you have to add your own accessors/duck interfaces/access wideners... Perhaps this PR wants to solve this problem.

maityyy:
Why is `LootTableLoadingCallback` not in the event subpackage? If we want to add another one, it will be a mess. Other modules always add events to their package, so I think it is bad to move away from this concept.

Juuxel:
@maityyy I don't really think it would be a mess. 1) There are only four other classes in that package. 2) Other new modules are also moving away from event subpackages. Some of them use a `*Events` class instead, but I decided to leave the structure the same from v1.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableLoadingCallback.java
Juuxel: I updated this to cancel future callbacks now. It's different behaviour than in v1's original implementation, but it's also completely undefined in v1. It doesn't make sense to continue the callback loop anyway.

RedstoneParadox (approved):

valoeghese (approved):

liach:
Should this target 1.17?

Juuxel:
I'm not sure what the general policy here is. It should be really easy to port, though, if needed.

Technici4n:
> Should this target 1.17?

I don't think it should target 1.17, we would probably want to backport it anyway.

PS: Yes I will review this soon:tm:.

modmuss50:
Makes little diffrence now, the main reason why we suggested targgeting the stable release is it saves PRs having to be updated each week. It takes the same amount of effrot to forward port or backport.

liach:
> Makes little diffrence now, the main reason why we suggested targgeting the stable release is it saves PRs having to be updated each week. It takes the same amount of effrot to forward port or backport.

oh, notice that porting may be affected by modern java features

Technici4n (approved):
Looks fantastic.

maityyy:
Why is this a draft?

Juuxel:
I need to rework it again to have interface injection according to @Technici4n - and interface injection is months away from stabilising at the very least. *(Edit: this aged like milk)*

It also needs the transitive access widener module to replace `FabricLootTables` and `FabricLootPools`, which I don't really want to create in a PR that will be stuck in the "waiting-forever-for-toolchain-features" limbo. The merge conflicts would be too much if another PR created the module.

modmuss50:
> and interface injection is months away from stabilising at the very least.

It should be mostly ready now. Just need to update loom to 0.11.

fabric-loot-table-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/FabricLootPoolBuilder.java
```diff
@@ -0,0 +1,114 @@
...
+@ApiStatus.NonExtendable
+public interface FabricLootPoolBuilder {
+	/**
+	 * Adds an entry to this builder.
+	 *
+	 * @param entry the added loot entry
+	 * @return this builder
+	 */
+	default LootPool.Builder with(LootPoolEntry entry) {
+		throw new UnsupportedOperationException("Unimplemented");
```
Technici4n: Could possibly use a more descriptive error message to avoid scaring users unfamiliar with this pattern? `"Implemented via mixin"` or so seems nicer.

Technici4n (approved):
Looks very good.

deirn (approved):

Technici4n:
Now that the TAW module exists, this might be ready for a final (sigh) update?

Player3324 (approved):
LGTM, thanks! (excl module vs pkg name as discussed)

fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.resource.ResourceManager;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events for manipulating loot tables.
+ */
+public final class LootTableEvents {
```
Technici4n, line 32: Would be nice to clarify that replaced loot tables can be modified in the `MODIFY` event, maybe.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourceMixin.java
```diff
@@ -14,20 +14,16 @@
...
 
-@Mixin(LootPoolEntryTypes.class)
-public interface LootPoolEntryTypesAccessor {
-	@Invoker("register")
-	static LootPoolEntryType register(String id, JsonSerializer<? extends LootPoolEntry> serializer) {
-		throw new UnsupportedOperationException("Mixin dummy");
-	}
+import net.fabricmc.fabric.impl.resource.loader.FabricResource;
+
+// Even though the interface is injected, we must extend it here,
```
Technici4n: It's not injected anymore, is it?

modmuss50 (approved):
Looks awesome 👍 Thanks a lot for this. I will put it into last call.

PR #1242, opened 2020-12-24
Mark module lifecycle inside buildscripts [Part 1 of buildscript overhaul]

i509VCB:
This does not do anything at the moment but it will be used in the future to allow creating distributions of Fabric API which do not bundle deprecated modules.

This also updates licenser since I had to apply it in buildSrc project.

**Whats the deal with a whole extension for this?**

The plan is to move a lot of the cruft in the main buildscript and centralize some module information such as the big mess for `moduleDependencies` into this `module` closure.

The entire buildscript for a module could look like this when this whole thing is complete:
```groovy
module {
	name("fabric-registry-sync-v0") // This is archivesBaseName
	version("0.8.0")
	stable() // The only part included in the PR so far
	accessWidener() // Resolves the access widener at `src/main/resources/fabric-registry-sync-v0.accesswidener`

	dependency("fabric-api-base") // This would replace `moduleDependencies`
	dependency("fabric-networking-api-v1")
	testDependency("fabric-lifecycle-events-v1") // This would simplify the mess of dependencies in the `testmodCompile` configuration
}
```

The above module configuration could also be used to generate a `fabric.mod.json` at compile time for publish or testing in IDE as to reduce the amount of inconsistency between the `fabric.mod.json` between every module.


buildSrc/src/main/java/net/fabricmc/fabric/impl/gradle/ApiModuleExtension.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.gradle;
+
+public class ApiModuleExtension {
+	private ModuleType moduleType;
```
liach: This really sounds like life cycle more

buildSrc/src/main/java/net/fabricmc/fabric/impl/gradle/ApiModuleExtension.java
```diff
@@ -0,0 +1,37 @@
...
+
+	public void deprecated() {
+		this.moduleType = ModuleType.DEPRECATED;
+	}
+
+	public void experimental() {
+		this.moduleType = ModuleType.EXPERIMENTAL;
+	}
+
+	public void stable() {
```
liach: I personally prefer a setter. You can still call like
```
lifecycle = ModuleLifecycle.DEPRECATED
```

liach:
```gradle
module {
	dependency("fabric-api-base") // This would replace `moduleDependencies`
	dependency("fabric-networking-api-v1")
	testDependency("fabric-lifecycle-events-v1") // This would simplify the mess of dependencies in the `testmodCompile` configuration
}
```
Personally I disagree with such usage to declare dependencies.
I would prefer these `dependency` be one single `Dependency dependency(String notation)` method defined in `ApiModuleExtension`, which creates a resolvable dependency object that you can use in **any** gradle configuration.

i509VCB:
I could have the dependency methods return a gradle dependency at convince. The two methods will add the specific other module to the `(testmod)Compile` configuration. Return type could be a gradle dependency, yes.

Also note that section of the pr is conceptual stuff for a later pr.

buildSrc/src/main/java/net/fabricmc/fabric/impl/gradle/ApiModuleExtension.java
i509VCB: There are so few lifecycle types to where I think the method is more concise.

liach: Reasonable enough.

liach (approved):

i509VCB:
This is a huge PR and will be hell to maintain - I am going to approach it in multiple parts in near future.

PR #1243, opened 2020-12-25
Lifecycle events: Add event when client has changed worlds

i509VCB:
Somewhat requested by @LambdAurora, the event added in this pull request is called when the client's world is changed. This may be used for reference cleanup when the client has changed worlds in case a mod tracks the current client world.

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
@@ -53,8 +53,8 @@ public final class ClientLifecycleEvents {
 	});
 
 	/**
-	 * Called when the current player in the Minecraft game is moved to a different world.
-	 * This occurs before the client's {@link MinecraftClient#world current world} is disposed and a new world is created.
+	 * Called when the current player in the Minecraft game has moved to a different world.
+	 * This occurs when a new world has been created, but before the client's {@link MinecraftClient#world current world} is disposed.
 	 */
 	public static final Event<ChangeWorld> CHANGE_WORLD = EventFactory.createArrayBacked(ChangeWorld.class, callbacks -> (client, world, destination) -> {
 		for (ChangeWorld callback : callbacks) {
@@ -78,7 +78,7 @@ public final class ClientLifecycleEvents {
 	@FunctionalInterface
 	public interface ChangeWorld {
 		/**
-		 * Called when the current player in the Minecraft game is moved to a different world.
+		 * Called when the current player in the Minecraft game has moved to a different world.
 		 * A mod may use this event for reference cleanup on the current world.
 		 *
 		 * @param client the Minecraft client
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
@@ -17,6 +17,7 @@
 package net.fabricmc.fabric.api.client.event.lifecycle.v1;
 
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.world.ClientWorld;
 
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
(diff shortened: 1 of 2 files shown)


LambdAurora (approved):

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
```diff
@@ -51,13 +52,39 @@ private ClientLifecycleEvents() {
...
 	@FunctionalInterface
 	public interface ClientStopping {
 		void onClientStopping(MinecraftClient client);
 	}
+
+	@Environment(EnvType.CLIENT)
+	@FunctionalInterface
+	public interface ChangeWorld {
+		/**
+		 * Called when the current player in the Minecraft game is moved to a different world.
```
liach:
```suggestion
		 * Called when the current player in the Minecraft game has moved to a different world.
```
indicate the move is already done

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
```diff
@@ -51,13 +52,39 @@ private ClientLifecycleEvents() {
 		}
 	});
 
+	/**
+	 * Called when the current player in the Minecraft game is moved to a different world.
```
liach:
```suggestion
	 * Called when the current player in the Minecraft game has moved to a different world.
```

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
```diff
@@ -51,13 +52,39 @@ private ClientLifecycleEvents() {
...
+	 * This occurs before the client's {@link MinecraftClient#world current world} is disposed and a new world is created.
```
liach:
```suggestion
	 * This occurs when a new world has been created, but before the client's {@link MinecraftClient#world current world} is disposed.
```

liach (commented):
Just a few clarifications in documents.

liach (approved):

Player3324:
wouldn't this application want a clientside world unload instead? going trough the main menu doesn't quite meet what i'd understand as changing the world, but it's at least as relevant for cleanup.

LambdAurora:
> wouldn't this application want a clientside world unload instead? going trough the main menu doesn't quite meet what i'd understand as changing the world, but it's at least as relevant for cleanup.

@sfPlayer1 only having a world unload event is insufficient as there's things that needs to be cleanup when changing dimensions as well as when leaving entirely a world.

Player3324:
Considering we have https://github.com/FabricMC/fabric/blob/1.16/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java the module choice seems wrong at least.

Mote specifically, lifecycle = load+unload and entity-events = anything else

i509VCB:
I understand what you say there player - but this is not really involved with entities as it handles the entire world/client player instance for the client being canned and remade.

OverlordsIII (approved):

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
```diff
@@ -51,13 +52,39 @@ private ClientLifecycleEvents() {
...
+	 * Called when the current player in the Minecraft game has moved to a different world.
+	 * This occurs when a new world has been created, but before the client's {@link MinecraftClient#world current world} is disposed.
+	 */
+	public static final Event<ChangeWorld> CHANGE_WORLD = EventFactory.createArrayBacked(ChangeWorld.class, callbacks -> (client, world, destination) -> {
+		for (ChangeWorld callback : callbacks) {
+			callback.onChangeWorld(client, world, destination);
+		}
+	});
+
+	@Environment(EnvType.CLIENT)
```
TheGlitch76, line 65: If ClientLifecycleEvents is already being stripped isn't this unnecessary?

TheGlitch76 (approved):
seems simple enough

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
i509VCB: At bytecode level nested/inner classes can still be touched outside of client

liach: For the JVM inner classes are just treated as other classes in the same package. So when outer classes are removed, inner classes, like other classes in the same package, are not touched.

PR #1245, opened 2020-12-29
Update networking API to 20w46a+, simplify mixin impl

ChloeDawn:
**What broke?**
Mojang introduced a new API class, that I am donning "Observer" which wraps a tracking server player, providing a getter for it, and a simple packet dispatching method alongside. The API was still mixing in expecting a collection of server players to expose for `PlayerLookup#tracking(Entity)` which has since been replaced which this new observer type. This generic type mismatch results in a class cast exception at runtime when this API method is called. This pull request migrates the mixins and PlayerLookup implementation to support this change and fix the crash.

**What changed?**
There has been a migration to the new API class internally, and a refactor of entity tracker access. Previously the networking API had a mixin into TACS that shadowed the field and added a non-null lookup. The shadow has been replaced with an accessor mixin and the null checking migrated to the PlayerLookup call site.


LemmaEOF (approved):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PlayerLookup.java
```diff
@@ -111,9 +113,15 @@
...
+			EntityTrackerAccessor tracker = ((ThreadedAnvilChunkStorageAccessor) storage).getEntityTrackers().get(entity.getEntityId());
 
 			// return an immutable collection to guard against accidental removals.
-			return Collections.unmodifiableCollection(((ThreadedAnvilChunkStorageTrackingExtensions) storage).fabric_getTrackingPlayers(entity));
+			if (tracker != null) {
+				return Collections.unmodifiableCollection(tracker.getPlayersTracking()
+						.stream().map(class_5629::method_32311).collect(Collectors.toSet()));
+			}
+
+			return Collections.emptySet();
```
i509VCB, line 124: Should this be returning an empty set or fail since the entity is not being tracked at all by the server world.

ChloeDawn: This is following the semantics of the late `ThreadedAnvilChunkStorageMixinfabric_getTrackingPlayers`

i509VCB: Hmm I see.

i509VCB (approved):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PlayerLookup.java
```diff
@@ -111,9 +113,15 @@
 
 		if (manager instanceof ServerChunkManager) {
 			ThreadedAnvilChunkStorage storage = ((ServerChunkManager) manager).threadedAnvilChunkStorage;
+			EntityTrackerAccessor tracker = ((ThreadedAnvilChunkStorageAccessor) storage).getEntityTrackers().get(entity.getEntityId());
 
 			// return an immutable collection to guard against accidental removals.
-			return Collections.unmodifiableCollection(((ThreadedAnvilChunkStorageTrackingExtensions) storage).fabric_getTrackingPlayers(entity));
+			if (tracker != null) {
+				return Collections.unmodifiableCollection(tracker.getPlayersTracking()
+						.stream().map(class_5629::method_32311).collect(Collectors.toSet()));
```
Player3324, line 121: pre-sized output collection with manual for-each please, the Stream is needless overhead+debugging hazard

LambdAurora (approved):

ghost (approved):
Seems good, I can finally use it in my 1.17 mods without worrying about bugs.

modmuss50 (approved):
Looks good once player's suggestion has been done.

modmuss50:
Going to merge this as its quite crticial and only an impl change. We can look at de-streamifying this stuff at a later date.

PR #1249, opened 2021-01-01
Auto update meta about new fapi versions

dexman545:
Companion to FabricMC/fabric-meta#10

When a release is built, adds the API version to the list of MC versions based on the MC version it was built against.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ plugins {
 	id "net.minecrell.licenser" version "0.4.1"
 	id "org.ajoberstar.grgit" version "3.1.1"
 	id "com.matthewprenger.cursegradle" version "1.4.0"
+	id "org.ajoberstar.git-publish" version "3.0.0"
 }
 
 def ENV = System.getenv()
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -389,6 +390,64 @@ task checkVersion {
...
+	repoUri = versionMapRepo
+	branch = "main"
+	commitMessage = "Automated addition of new api version from FabricMC repo"
+
+	preserve {
+		include "README.md"
+		include "LICENSE"
+		include "apiVersions.properties"
+		include ".gitignore"
+	}
```
liach, line 399: I personally suggest moving these stuff to github actions than using another gradle plugin which is useless in most of the use cases.

dexman545: Can actions read and write a properties file in a sane way?
I can replace the plugin and use grgit or an `Exec` task if the adding of a new plugin is undesired.

liach: Imo a exec would be good enough. It would avoid problems like upgrading gradle and having this plugin break because it doesn't support new gradle, and people who just want to develop api would not need to grab this gradle plugin.

dexman545:
I've made it use grgit now (which was already in use by the script) ~didn't know how to specify user by git command line without fear of screwing things up~.

If the modules included an MC deps with version ranges, I could make it a bit better with generating MC versions, but would mean bringing in a semver parser to the script. Just having built-against MC version would be a big step up.

liach:
@modmuss50 @sfPlayer1 Is this concept acceptable? If the overall direction is wrong, halting this than letting this go nowhere would be better.

modmuss50:
This not how I would go about doing this, and something we decided to put on hold.

PR #1252, opened 2021-01-01
Optimization idea for registry

maityyy:
I think it's worth using larger expected sizes of these maps for other mods.

However, I have not tested if the performance has increased. Perhaps someone has thoughts on this matter and perhaps my idea is meaningless — let me know, I'm interested.


i509VCB:
I would probably collapse this into `MixinSimpleRegistry` or else we are gonna have three mixins in same module targetting same class this new one, `MixinSimpleRegistry` and `MixinIdRegistry`

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -0,0 +1,69 @@
...
+// Use larger expected sizes for other mods
+@Mixin(SimpleRegistry.class)
+public class SimpleRegistryMixin<T> {
+
+	@Redirect(
+			method = "<init>",
+			at = @At(
+					value = "INVOKE",
+					target = "Lcom/google/common/collect/HashBiMap;create()Lcom/google/common/collect/HashBiMap;",
+					opcode = 0
```
i509VCB: Isnt there a field constant in `Opcodes` for this magic value?

i509VCB (changes requested):
Will need to wait for modmuss or player to comment on this pull request about scope.

My recommendation to avoid 2 redirects is to inject at TAIL and reinitialize the maps at larger size while copying the contents of old map you are replacing in case someone does some funky thing with adding to registries during init (in which case you should not be using mixin in registry to register stuff).

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
TheGlitch76: I assume this is a typo and they meant `ordinal`, because INVOKE doesn't accept opcode targets.

modmuss50:
Really needs to be backed up by numbers as it sounds like a guess ATM. If its only a small improvement id suggest that this change would be out of scope and not worth us maintaining, however if its a significant improvement then we can think again.

A lot of the registries never have close to 2048 entries, so it seems a bit of a waste on those? Again would need backuping up by numbers.

modmuss50:
Had a very quick look at this, and in my TR dev env (includes a handful of other larger conent mods) 54ms is spent in SimpleRegistry.add. There might well be improvement to be made here, but it doesnt seem a big enough issue for fabric to tackle IMO.

liach:
Maybe changing the growth factor for hash map than the initial size helps more.

maityyy:
> Maybe changing the growth factor for hash map than the initial size helps more.

Maybe you are right! But I don't have time to check it out...

modmuss50:
I dont think we really need this? Any objection to closing it or is there more invesigation work required here?

Issue #1118, opened 2020-10-11
[RFC] Reorganize location of modules in repository structure


i509VCB:
At the time of writing, Fabric API has 39+ modules which definitely clutter the root folder of the repository. Among these 39 modules, 11 are client only, 7 are deprecated, 1 is experimental and 1 has not been updated yet. There is currently no real way to discern between these types of modules without opening each of them up.

## The proposal

The modules shall be organized into 5 main categories:
 - `client`
 - `common`
 - `server`
 - `deprecated`
 - `experimental`

Each category will have it's own folder in the root of the repository and modules contained in the category would be resolved via project dir relocation.

The `client` and `common` categories will contain a majority of the modules, where the `client` category contains modules which are only loaded when running a Minecraft client and `common` for modules designed for use on a Minecraft client and dedicated server.

The `server` category is for modules which only function on a dedicated server. As of 1.13 all server classes are present on the client.

The `deprecated` category is where deprecated modules will be located. The deprecation of a module means the module should be in the deprecated category, regardless of whether the module was previously in `common`, `client` or `server`.

The `experimental` category is where experimental modules are located per the RFC in #499. A module should only be present in the `experimental` category if the entire module is experimental. If a module in `client`, `server` or `common` has some parts marked as experimental, the module will stay in `client`, `server` or `common`.

XakepSDK:
What about server only modules? E.g. modules that won't prevent vanilla client to work with a modded server

i509VCB:
Modules that target the logical server would be part of `common`. This is because the "server" exists on both a client (as an integrated server) and dedicated server.

If a module only works on the dedicated server, it would be part of `common` also since Mojang no longer has any dedicated server only classes.

If this does change in the future, we will revisit a specific dedicated server category.

ghost:
Kind of unrelated, but also slightly related I think it would be good if in each module readmes there was information relating to the status of the module (stuff relating to what this RFC addresses). Like what if there were some badges under the title of each module that mentioned the version that it is updated to, the status of the module, and which sides it applies to.

As an example

# Fabric Lifecycle Events (v1)
![](https://img.shields.io/badge/version-1.16.3-blueviolet?style=flat) ![](https://img.shields.io/badge/status-active-brightgreen?style=flat) ![](https://img.shields.io/badge/side-common-yellow?style=flat)

This way we could display more meta information about each module for mod developers. This is just a list of some ideas for the labels.

[...]

PR #1253, opened 2021-01-02
Add custom value denoting module lifecycles. 

i509VCB:
This adds a custom value to all Fabric API modules:

This custom value is located at `custom.fabric-api:module-lifecycle`. This is a string custom value and has three supported values: `stable`, `experimental` and `deprecated`. These denotations are per the suggestions in #1118. This also adds a task to verify each submodule has this custom value in it's `fabric.mod.json` and it contains the proper value.

This new custom value may be used by other mods to have a look into the module lifecycle of a fabric api module - possibly modmenu.

~~My hand has sadly been forced by the fact JsonSlurper is not available in main buildscript no matter what I try so I have been forced to open up buildSrc.~~

This adds a `validateModules` task that is called as part of the `check` task so module validation will also be handled in github actions.

This pull request also updates the gradle wrapper to 6.7.1


buildSrc/src/main/groovy/net/fabricmc/fabric/impl/gradle/ValidateModuleTask.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+import groovy.json.JsonSlurper
+import org.gradle.api.DefaultTask
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.TaskAction
+
+class ValidateModuleTask extends DefaultTask {
+    @TaskAction
+    void validate() {
+        project.subprojects { subProject ->
```
zml2008: Crossing project boundaries seems to be discouraged by Gradle -- what you probably want to do is have `ValidateModuleTask` be registered on every subproject, and only act on that project. Then, through Gradle's workings running `./gradlew validateModule` would run the task in every subproject it's present in.

i509VCB: No more interproject mingling now.

zml2008 (approved):
Looks good to me, at least from a gradle-implementation perspective. (though, do you want to have a license header in your `buildSrc` project?)

modmuss50 (approved):

Player3324 (approved):

gradle/module-validation.gradle
```diff
@@ -0,0 +1,55 @@
...
+	ValidateModuleTask() {
+		group = "verification"
+
+		// Hook up validation to check task
+		project.tasks.check.dependsOn(this)
+	}
+
+	@TaskAction
+	void validate() {
+		def json = new JsonSlurper().parse(project.file("src/main/resources/fabric.mod.json") as File) as Map<String, Map<String, String>>;
```
liach, line 35: Should grab the mod json from main sourceset's resources dir than using this hardcoded path.

gradle/module-validation.gradle
```diff
@@ -0,0 +1,55 @@
...
+
+		if (json.custom == null) {
+			throw new GradleException("Module ${project} does not have a custom value containing module lifecycle!")
+		}
+
+		if (json.custom.get("fabric-api:module-lifecycle") == null) {
```
liach: I wonder how costly calling this is, but I personally prefer storing to local var to avoid extra map lookups

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-all.zip
```
liach, line 3: just bump to 6.8 i guess

settings.gradle
```diff
@@ -1,6 +1,7 @@
 pluginManagement {
 	repositories {
 		jcenter()
+		mavenCentral()
```
liach: for what? This is already covered by jcenter

gradle/module-validation.gradle
i509VCB: I believe JsonSlurper deserializes the file into the respective `Map`, `List`, `Number`, `Boolean`, `null`s for the things in json

settings.gradle
i509VCB: Probably leftover from some debugging I did earlier

gradle/wrapper/gradle-wrapper.properties
modmuss50: I’d prefer to leave it be, I doubt there will be any issues but it’s easier to track down if updating this stuff is done on its own

PR #1258, opened 2021-01-05
Added ItemTooltipComponentCallback

gudenau:
Works similar to `ItemTooltipCallback` but for the new 1.17 stuff that bundles use.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipComponentCallback.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipComponentCallback.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.item.v1;
+
+import net.minecraft.client.gui.tooltip.TooltipComponent;
+import net.minecraft.client.item.TooltipContext;
+import net.minecraft.client.item.TooltipData;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+@Environment(EnvType.CLIENT)
+public interface ItemTooltipComponentCallback {
+	/**
+	 * Fired when the game is creating a {@link TooltipContext}.
+	 */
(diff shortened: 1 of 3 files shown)


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipComponentCallback.java
```diff
@@ -0,0 +1,54 @@
...
+public interface ItemTooltipComponentCallback {
+	/**
+	 * Fired when the game is creating a {@link TooltipContext}.
+	 */
+	Event<ItemTooltipComponentCallback> EVENT = EventFactory.createArrayBacked(ItemTooltipComponentCallback.class, callbacks -> (tooltipData) -> {
+		for (ItemTooltipComponentCallback callback : callbacks) {
+			TooltipComponent component = callback.createTooltipComponent(tooltipData);
+
+			if (component != null) {
+				return component;
```
OroArmor, line 39: I see one issue of this, if mod A is loaded then mod B, you get one result vs B then A, especially if they both create a component for the same data.

OroArmor (commented):
With my main issue, I'm not sure how you would implement it, but i think it could cause a lot of unessicary conflicts.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipComponentCallback.java
gudenau: You can't really have "merged" data. This would mostly be an issue if two mods added Shulker previews or something. I don't think it will be a huge issue in production.

Would be easy enough to mark it and warn once per set of component conflicts, but again it should be rare.

OroArmor: I dont know if it would be easy to warn, but I do agree it will have minimal impact.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/TooltipComponentMixin.java
```diff
@@ -0,0 +1,47 @@
...
+
+import net.fabricmc.fabric.api.client.item.v1.ItemTooltipComponentCallback;
+
+@Mixin(TooltipComponent.class)
+public interface TooltipComponentMixin {
+	/**
+	 * @author gudenau
+	 * @reason Mixins can't inject into interfaces
+	 */
+	@Overwrite
```
emilyploszaj, line 35: This overwrite is nasty, you can get the same effect by injecting into where `TooltipComponent.of` is called

gudenau:
Closed because of the premature nature of 1.17.

emilyploszaj (changes requested):
I have pretty big problems with this API. Like previously stated, it doesn't have an option for merged or compound tooltip components. The override also makes it extremely difficult for modders to add custom tooltip components to just their new items, making them go through a callback instead of the built in method ([here](https://github.com/Queerbric/Inspecio/blob/main/src/main/java/io/github/queerbric/inspecio/mixin/ScreenMixin.java)'s the lambda injection used in Inspecio).

My suggestion is add a new `ConvertibleTooltipData` type that is handled in the lambda injection (instead of overwrite) and have this file have a single method for converting itself, the `TooltipData`, into a `TooltipComponent`, and have this be used for the generic conversion (see [this](https://github.com/Queerbric/Inspecio/blob/main/src/main/java/io/github/queerbric/inspecio/tooltip/ConvertibleTooltipData.java) example from Inspecio). I also suggest that tooltips should inherently be mergeable through this callback (potentially even allowing tooltip components to specify a sortnum) and be merged into a `CompoundTooltipComponent` and/or `CompoundTooltipData` (another example from Inspecio is [here](https://github.com/Queerbric/Inspecio/blob/main/src/main/java/io/github/queerbric/inspecio/tooltip/CompoundTooltipComponent.java))

PR #1263, opened 2021-01-10
Add provides entry and change modid

AMereBagatelle:
This PR at the time of writing does two things:
- Changes the mod id to `fabric-api`
- Adds a `provides` entry with `fabric` as a provided mod


src/main/resources/fabric.mod.json
```diff
@@ -1,6 +1,9 @@
 {
   "schemaVersion": 1,
-  "id": "fabric",
+  "id": "fabric-api",
+  "provides": [
```
i509VCB, line 4: You need to bump loader dep to use this

i509VCB (changes requested):

src/main/resources/fabric.mod.json
AMereBagatelle: Done

modmuss50:
You actaully need to bump the loader version in the build.gradle. Please also test to make sure it works as you clearly havent done.

AMereBagatelle:
The game now actually launches :P
Also, I tested with a mod that depends on the `fabric` modid and it works (I can tell by the warning thrown in the log)

florensie (approved):

LambdAurora (approved):

altrisi (approved):

williambl (approved):

modmuss50 (approved):
I forgot about this, seems like everyone else did as well. Pending testing this is good to go as far as I can see.

Martmists-GH (approved):

Juuxel (approved):

Martmists-GH:
@AMereBagatelle I'd recommend resolving conflicts

AMereBagatelle:
Conflicts resolved

Player3324:
I've been playing with the idea to abuse the mod id change to also change module auto-loading behavior. This could work as follows:
- change the modules' load conditions to if-recommended (except for a few that we always want to load, they'd use if-possible or always)
- change the fapi's mod id to fabric-api
- add a jij with id `fabric` and load condition if-needed that recommends all fapi modules
- make loom auto-generate module deps

-> migrating to fabric-api migrates to opt-in module loading, everything keeps working as expected

Whether we want to do this is TBD, but ultimately relevant for whether the change in this PR should be done or not.

ImperatorStorm (approved):

liach:
I'm out of the loop and hope modmuss can decide on this.

altrisi:
> I've been playing with the idea to abuse the mod id change to also change module auto-loading behavior. This could work as follows:
> 
> * change the modules' load conditions to if-recommended (except for a few that we always want to load, they'd use if-possible or always)
> 
> * change the fapi's mod id to fabric-api
> 
> * add a jij with id `fabric` and load condition if-needed that recommends all fapi modules
> 
> * make loom auto-generate module deps
> 
> 
> -> migrating to fabric-api migrates to opt-in module loading, everything keeps working as expected
> 
> Whether we want to do this is TBD, but ultimately relevant for whether the change in this PR should be done or not.

This sounds really interesting, however, how does it affect the id change?

If the dependencies are going to be generated for specific modules (or, at least, there would be dependencies for separate modules and not a single one for the `fabric-api` id, if I understood correctly), new mods that opt in to that system wouldn't depend directly on `fabric-api` (but on the modules), so I can't see the problem with changing it earlier.

Unless you mean that the mod would depend on `fabric-api` and then loader (or some library/loader plugin by the api) would determine the actual dependencies via those generated dependency metadata. That would solve the problem of people not knowing where to find fabric api if the error screen tells them to get a module, something the other method doesn't provide.

Player3324:
You are correct, the fabric-api dep was more meant to serve as user guidance. I however have plans to cover this in metadata format extensions instead, so we may be able to get away with a simple rename after all.

thecatcore (approved):

modmuss50:
Superseded by [2446](https://github.com/FabricMC/fabric/pull/2446) that targets 1.19.

Thanks everyone for the reviews and feedback on this 👍

PR #1264, opened 2021-01-11
Interaction Events v1 (WIP)

### LINKED ISSUE #1247: Refactor of interaction events
This issue exists to lay out the requests people may have for a refactored interaction events module. This would culminate in `fabric-interaction-events-v1`

The current interaction events module does work but it has issues such as combining notification and cancellation into the same callback.

The event instances in `PlayerBlockBreakEvents` are the general structure I would use for the refactored module. However splitting `BEFORE` into an `ALLOW` and `BEFORE` phase for the event.

The new module would also make it more specific which logical side the events are dispatched on: So `PlayerBlockBreakEvents` would become `ServerPlayerBlockBreakEvents`.

The new module would also include events for placing blocks on the server and client and porting over the server block break events to the client.

Another thing that will be considered is the existing block/entity attack/use callbacks. These callbacks do work but they do not separate cancellation and notification at all. These events would receive an implementation in a similar structure to block break events with a slight change to allow setting an action to occur. I believe these would require 4 events:

 - `Allow` (Cancellable - Prevents the action completely and implicitly returns `ActionResult.FAIL`)
Contractually you should not modify the world here. If an action is not allowed to occur no further events are called.

 - `Intercept` [Name is not final] (Allows a mod to capture full control the interaction and cancel an interaction event while specifying the action result to return). This may be used to for example use a wrench on a chest to rotate it but cancel opening the gui.

 - `Before` (Happens before the vanilla interaction method is called)

 - `After` (Happens after the vanilla interaction method is called)

Other existing events:

Item pick events could see enhancement, suggestions will be taken there.

Timeline: I plan to start on this and have a pull request in the next weeks.

i509VCB:
This is not complete yet, but available for viewing so people can comment on the new module as it comes together.

So far I have implemented the block break/place events for the logical server and other parts will be implemented over the next few days.

There are also other things being considered such as an Interact/Attack air event and other things discussed in #1247

This will resolve #1247


fabric-interaction-events-v1/src/main/java/net/fabricmc/fabric/impl/interaction/InteractionRaycasting.java
```diff
@@ -0,0 +1,109 @@
...
+ * Utilities related to raycasting in preparation for calling interaction events.
+ */
+public final class InteractionRaycasting {
+	/**
+	 * Gets the reach distance of a player.
+	 *
+	 * @param player the player
+	 * @return the reach distance
+	 */
+	public static double getReachDistance(ServerPlayerEntity player) {
```
maityyy, line 37: Why `ServerPlayerEntity` and not just `PlayerEntity`?

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientPickBlockCallback.java
```diff
@@ -20,14 +20,16 @@
...
 import net.minecraft.util.hit.HitResult;
 
+import net.fabricmc.fabric.api.client.interaction.v1.event.player.ClientPickItemEvents;
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 /**
  * This event handler has been deprecated due to not hooking nicely
- * into the game. Please use the alternatives.
+ * into the game and is no longer implemented. Please use the alternatives provided in {@link ClientPickItemEvents}.
```
TheGlitch76: If this is no longer implemented, it needs to be removed, full stop.

Silent failing bugs are *impossible* to track down, please do not introduce one into fabric api.

OroArmor:
I think some of the combat events should be added to its own module: `fabric-combat-api-v1`. I have a start here: https://github.com/FabricMC/fabric/pull/1210

fabric-interaction-events-v1/src/main/java/net/fabricmc/fabric/impl/interaction/InteractionRaycasting.java
i509VCB: I don't believe that will be needed on the logical client, but time will tell there

L2750558108:
Why dont the BlockPlaceEvent merge? I really need it. Or is there any ways to replace it?

PR #1266, opened 2021-01-13
Create EntityBrainHelpers

falseresync:


falseresync:
Forgot license headers on mixins

maityyy:
It looks cool. Will remove a lot of mixin-accessors from mods. Why not?

Technici4n (approved):

liach (changes requested):
Please bump the version of the object builder api. Otherwise looks great

i509VCB:
@liach we have actually been discouraging bumping versions in PRs recently - modmuss bumps them at merge time.

liach (approved):
Then I see no reason not to merge this

i509VCB (commented):
I would null check the params in the factory methods with `Objects.requireNonNull`. Otherwise this looks good.

comp500:
As I mentioned on discord earlier, would something to create instances of Activity also be in scope?

falseresync:
@i509VCB okay, I’ll add null check to the SensorType
@comp500 I’ll add activities, they are in scope of this

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/EntityBrainHelper.java
```diff
@@ -0,0 +1,104 @@
...
+		return Registry.register(Registry.MEMORY_MODULE_TYPE, id, MemoryModuleTypeAccessor.init(Optional.empty()));
+	}
+
+	/**
+	 * Creates and registers a {@link MemoryModuleType}.
+	 *
+	 * @param id the id of this {@link MemoryModuleType}.
+	 * @param codec {@link Codec} used to serialize and deserialize actual memory modules.
+	 * @return a new {@link MemoryModuleType}.
+	 * @implNote codec can be null, though then the shortcut method should be used instead.
```
comp500, line 96: Current implementation doesn't allow for a null Codec (`Optional.of` does an `Objects.requireNonNull` call) - either change this note to suggest you use the other overload if you don't want to supply a codec, or change to `Optional.ofNullable` and add an `@Nullable`.
Technici4n suggested doing the latter and changing the overload to call this one.

comp500 (approved):

Technici4n (approved):
LGTM

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/EntityBrainHelper.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**
+	 * Creates and registers a {@link SensorType}.
+	 *
+	 * @param id the id of this {@link SensorType}.
+	 * @param sensorFactory a supplier to provide a {@link Sensor}.
+	 * @return a new {@link SensorType}.
+	 */
+	public static <U extends Sensor<?>> SensorType<U> registerSensorType(Identifier id, Supplier<U> sensorFactory) {
```
RedstoneParadox, line 72: `U` should be annotated as `@NotNull` (the Jetbrains one) since the `Suplier` can't return `null`.

Technici4n: `NotNull` is implicit by default, otherwise you end up spamming `@NotNull` everywhere. Also, this would introduce 3 kinds of types (not null, nullable, and ????), which is very bad.

Issue #1271, opened 2021-01-18
Codegen for events


yyny:
Writing events is currently very verbose, requiring lots of boilerplate code.

Additionally, many events simply take the form

```java
for (Callback callback : callbacks) {
    callback.onEvent(...);
}
```

See also [this comment in `EventFactory`](https://github.com/FabricMC/fabric/blob/0d4d7e34589be55835349b26cacc9fdfe05170b6/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java#L73)

maityyy:
I agree. It would be very helpful.

Daomephsta:
We could achieve this now with an annotation processor. It might look something like this; where `@GenerateSimpleEvent` is applicable to types, source-only, and repeatable.

Foo.java
```java
@GenerateSimpleEvent(name = "foo.bar.Baz.QUX")
// Requirement of GenerateSimpleEvent so the processor doesn't have to enforce the functional interface contract
@FunctionalInterface
public interface Foo
{
    public void doFoo();
}
```

Baz.java (generated file)
```java
package foo.bar;

public static final Event<Foo> QUX = EventFactory.createArrayBacked(Foo.class, callbacks -> () -> 
{
    for (Foo callback : callbacks)
    {
        callback.doFoo();
    }
})
```

ASM generated classes are also an option, but first we'd need to add a hook in Loader that allows mods (such as FAPI) to generate classes.

Ueaj-Kerman:
generating classes doesn't strictly need an api

i509VCB:
I believe we would end up doing the following in some form:

Store all event instances in an array on the event class when using a generated event. Use the classloader and define a new class for the invoker type which simply just pushes each instance on the stack and pops it off by running the callback method.

The generated code would be something like below but optimized more than the equivalent compiled code would be:

```java
class Foo$InvokerImpl implements Foo {
    Foo$InvokerImpl(Foo foo1, Foo foo2, Foo foo3, Foo ...) { ... set instance fields ... }

    void doFoo(Bar bar) {
        this.foo1.doFoo(bar);
        this.foo2.doFoo(bar);
        this.foo3.doFoo(bar);
        ...
    }
}
```

That invoker impl would be set as the invoker on the event - we will only generate the invoker using asm in two cases:

**First invocation of the event via a sort of factory invoker which generates the invoker class and then calls it**
Reason here is to avoid generating the invoker 30 times during mod init.
[...]

Issue #129, opened 2019-03-21
Events for listening to keybinds


moonheart08:
At the moment, there is no sane way to have an event fire upon pressing a key without using mixins to modify the input code.

I propose the addition of an event that allows a function to be called whenever a key is pressed or released, to help keep the input code clean of external mixins.

yyny:
I see two potential events here:

 1. Event fired whenever a registered `KeyBinding` changes state (e.g. `"key.jump"`), the key set in the options menu
 2. Event fired whenever a key is pressed or released (e.g. `"key.keyboard.space"`), the actual key code
 
The first is more obviously useful to most mod developers (I can see "magical dances" or key loggers for screen recording being a thing for example).
The second could be abused to hardcode key-bindings, but can still be potentially useful for debugging, logging, scripting, or special items.

yyny:
Alternatively, we could implement more generic "window events" that include key presses.

Linguardium:
still hoping for this. Would even be nice to have a keybinding helper method doing effectively a non-clearing "wasPressed" to call handlers and then decrements the pressed count if the event consumes. seems the PR that was supposed to do this has been waiting for review for over a year now

PR #1272, opened 2021-01-18
User Input Events v1

yyny:
I need something like this for 1.16.4


i509VCB (changes requested):
This is going to need some... cleanup.

I will comment around that things that probably need to be cleaned up

fabric-input-events-v0/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,27 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-input-events-v0",
```
i509VCB: First order of business - all new modules should be `v1`

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilTypeMixin.java
```diff
@@ -0,0 +1,14 @@
...
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import net.minecraft.client.util.InputUtil;
+import net.minecraft.client.util.InputUtil.Key;
+
+@Mixin(InputUtil.Type.class)
+public interface InputUtilTypeMixin {
+	@Accessor("map")
+	public Int2ObjectMap<Key> fabric_getMap();
```
i509VCB:
```
	public Int2ObjectMap<Key> fabric_getMap();
```

No need for `fabric_` prefix when dealing with accessor mixins. Only case you should use it really is it there is a conflict and such conflict

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/api/client/FabricMouse.java
```diff
@@ -0,0 +1,45 @@
...
+import net.fabricmc.fabric.impl.client.FabricMouseImpl;
+
+public interface FabricMouse {
+	public static FabricMouse INSTANCE = FabricMouseImpl.INSTANCE;
+
+	public double impl_getX();
+	public double impl_getY();
+	public int impl_getPressedButtons();
+	public boolean impl_isButtonPressed(int button);
+	public int impl_getMods();
```
i509VCB: These all can be moved into the `impl` package and accessed via util methods there

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/api/client/FabricMouse.java
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client;
+
+import net.fabricmc.fabric.impl.client.FabricMouseImpl;
+
+public interface FabricMouse {
```
i509VCB: If we only expose static methods for the API, this should become a final class + private constructor.

I would also rename the class to something more akin to the purpose? FabricMouse just seems a bit weird sounding imo

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/api/client/FabricKeyboard.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client;
+
+import net.fabricmc.fabric.impl.client.FabricKeyboardImpl;
+import net.minecraft.client.util.InputUtil.Key;
+
+public interface FabricKeyboard {
+	public static FabricKeyboard INSTANCE = FabricKeyboardImpl.INSTANCE;
```
i509VCB: See comments on `FabricMouse` and also apply them here

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/impl/client/FabricMouseImpl.java
```diff
@@ -0,0 +1,89 @@
...
+		if (GLFW.glfwGetMouseButton(handle, GLFW.GLFW_MOUSE_BUTTON_4) == GLFW.GLFW_PRESS)
+			this.buttons |= (1 << GLFW.GLFW_MOUSE_BUTTON_4);
+		if (GLFW.glfwGetMouseButton(handle, GLFW.GLFW_MOUSE_BUTTON_5) == GLFW.GLFW_PRESS)
+			this.buttons |= (1 << GLFW.GLFW_MOUSE_BUTTON_5);
+		if (GLFW.glfwGetMouseButton(handle, GLFW.GLFW_MOUSE_BUTTON_6) == GLFW.GLFW_PRESS)
+			this.buttons |= (1 << GLFW.GLFW_MOUSE_BUTTON_6);
+		if (GLFW.glfwGetMouseButton(handle, GLFW.GLFW_MOUSE_BUTTON_7) == GLFW.GLFW_PRESS)
+			this.buttons |= (1 << GLFW.GLFW_MOUSE_BUTTON_7);
+		if (GLFW.glfwGetMouseButton(handle, GLFW.GLFW_MOUSE_BUTTON_8) == GLFW.GLFW_PRESS)
+			this.buttons |= (1 << GLFW.GLFW_MOUSE_BUTTON_8);
```
i509VCB: I think you could probably whip up a util method for this in the impl

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
```diff
@@ -0,0 +1,154 @@
...
+import net.fabricmc.fabric.api.client.event.input.MouseScrollEvent;
+import net.fabricmc.fabric.impl.client.FabricKeyboardImpl;
+import net.minecraft.client.options.KeyBinding;
+import net.minecraft.client.util.InputUtil;
+
+@Mixin(InputUtil.class)
+public class InputUtilMixin {
+	@ModifyVariable(method="setKeyboardCallbacks(JLorg/lwjgl/glfw/GLFWKeyCallbackI;Lorg/lwjgl/glfw/GLFWCharModsCallbackI;)V", at=@At("HEAD"), index=2)
+	private static GLFWKeyCallbackI fabric_changeKeyCb(GLFWKeyCallbackI keyCb) {
+		return (long window, int code, int scancode, int action, int mods) -> {
```
i509VCB: types in lambdas can be implicitly determined by javac and therefore aren't needed.

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilTypeMixin.java
yyny: > Only case you should use it really is it there is a conflict and such conflict

Doesn't every mixin expose inject their symbols into the target class? I'm don't think it's needed here, but what about the `@ModifyVariable`s?

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
i509VCB: This seems to replace the entire key callback.

I would actually inject into the synthetic method that the lambda is in vanilla so we don't need to duplicate vanilla code as you only seem to inject in like 3 spots in the entire method

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
```diff
@@ -0,0 +1,154 @@
...
+					ClientInputEvents.KEYBIND_REPEATED.invoker().onKeybind(keybind);
+					break;
+				}
+			}
+			keyCb.invoke(window, code, scancode, action, mods);
+		};
+	}
+	@ModifyVariable(method="setKeyboardCallbacks(JLorg/lwjgl/glfw/GLFWKeyCallbackI;Lorg/lwjgl/glfw/GLFWCharModsCallbackI;)V", at=@At("HEAD"), index=3)
+	private static GLFWCharModsCallbackI fabric_changeCharModsCb(GLFWCharModsCallbackI charModsCb) {
+		return (long window, int codepoint, int mods) -> {
```
i509VCB: Apply similar rationale of above comments to all callback replacements

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/KeyBindingMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.minecraft.client.options.KeyBinding;
+import net.minecraft.client.util.InputUtil;
+import net.minecraft.client.util.InputUtil.Key;
+
+@Mixin(KeyBinding.class)
+public interface KeyBindingMixin {
+	@Accessor("keyToBindings")
+	public static Map<InputUtil.Key, KeyBinding> fabric_getKeyToBindings() {
+		throw new AssertionError();
```
i509VCB:
```suggestion
		throw new AssertionError("Untransformed accessor!");
```

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/api/client/FabricMouse.java
yyny: >I would also rename the class to something more akin to the purpose?

I originally named it `Mouse` but it caused a conflict with the Minecraft class when I tried to use it.

i509VCB:
> Doesn't every mixin expose inject their symbols into the target class? I'm don't think it's needed here, but what about the @ModifyVariables?

At runtime mixin partially scrambles the names of injectors/redirects/modifywhatever. - so `onBlockBreak` -> `inject$zzz000$onBlockBreak`.

Mixin does not scramble if the thing being implemented is a duck interface

i509VCB:
Also some advice: You can run the `gradlew check` task and you will get a list of style issues printed. `gradlew licenseFormat` will apply all license headers for you.

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
yyny: I wasn't sure how that would work with other mods trying to do the same thing. This way we can be sure events get emitted before anyone else can screw things up. Also, the callback is still called, of course.

yyny:
> You can run the `gradlew check` task and you will get a list of style issues printed. `gradlew licenseFormat` will apply all license headers for you.

Thanks. Put it in a `CONTRIBUTING.md`!

i509VCB:
We do have most of a contribution doc but I need to publish it here

fabric-input-events-v0/src/main/java/net/fabricmc/fabric/api/client/FabricMouse.java
yyny: > final class + private constructor

This seems so obvious in hindsight. I guess in my head I intended for there to be multiple implementations.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/input/ClientInputEvents.java
```diff
@@ -0,0 +1,109 @@
...
+		listener -> mouse -> listener.accept(mouse),
+		listener -> mouse -> listener.onMouseScrolled(mouse));
+	public static final Event<FileDrop> FILE_DROPPED
+		= EventFactory.createSimpleArrayBacked(String[].class, FileDrop.class,
+		listener -> paths -> listener.accept(paths),
+		listener -> paths -> listener.onFilesDropped(paths));
+
+	@FunctionalInterface
+	public interface KeyState {
+		void onKey(KeyEvent key);
```
i509VCB: Okay so the big changes start here. Typically we would just pass the parameters here, such as key code, modifiers, etc

This avoids the need to needlessly allocate an object here.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -75,6 +77,52 @@ public static void invalidate() {
...
+	 *
+	 * @param eventT         The event class type.
+	 * @param listenerT      The listener class type.
+	 * @param make           Factory function for turning consuemrs into listeners
+	 * @param execute        Function for executing listeners
+	 * @param <T>            The event type.
+	 * @param <F>            The listener type.
+	 * @return The Event instance.
+	 */
+	public static <T, F> Event<F> createSimpleArrayBacked(Class<T> eventT, Class<F> listenerT,
```
i509VCB: I don't fully think we need these - we can avoid the alloc of consumers or whatever and the extra invokedynamic for a few lines of code. Yes it's ugly but a longer term solution would generate these with asm.

fabric-input-events-v1/build.gradle
```diff
@@ -0,0 +1,6 @@
+archivesBaseName = "fabric-input-events-v0"
+version = getSubprojectVersion(project, "0.1.0")
```
i509VCB: 1.0.0 for new modules

fabric-input-events-v1/build.gradle
```diff
@@ -0,0 +1,6 @@
+archivesBaseName = "fabric-input-events-v0"
```
i509VCB: Make sure to change the archivesBaseName to v1

i509VCB:
Not going to be able to comment much more for a few hours - hopefully stuff so far will bring the pr up to step

yyny:
I will clean up some more later tonight

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/input/ClientInputEvents.java
yyny: I might start an issue about this because I think this is the wrong way to go about things.

In this case, specifically, it is very useful to obtain additional information not originally provided by the event, such as getting the mouse position or buttons being held during a scroll event or `KeyEvent::getKey` and `KeyEvent::isPressed`.

But, I think it is good to limit the listeners to one input even in the general case because then events would map much more cleanly to monads/the observer pattern (without wrapper overhead).

Adhering to the observer pattern would also make the design of the future event generation codegen much easier if any would even be needed at all (Observers compose really well).

> This avoids the need to needlessly allocate an object here.

You can cache objects and use getter interfaces to prevent mutability. One function call of extra overhead but it has no allocation cost and makes the entire events API much more extensible (can add new inputs at any time). I don't think the overhead is even an issue for 99% of use cases (Can just call the function once and store the result in a local variable, everyone does that for `Minecraft.getInstance` already anways), and in the rare case we do need to call thousands of listeners per second we still have the option of providing "unsafe" events that provide the mutable event object as input (Under the contract that no listeners modify the input, of course). It's probably even possible to create a class with final members and then use ASM to modify the cached event object.

All that together would actually be faster because it doesn't have the same hidden cost of allocation stack frames and copying parameters over and over again.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
yyny: I originally created a `Consumer<T>`-like for my own convenience but because of the lax requirement on event inputs that created lots of boilerplate anyway. See the comment above. Also see #1271, since there wasn't a tracked issue for this yet. I will revert this for now.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/input/ClientInputEvents.java
yyny: Actually, on the topic of mutability, aren't we already implicitly requiring listeners to not modify their inputs anyway?

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
```diff
@@ -0,0 +1,90 @@
...
+	public void update() {
+		MinecraftClient client = MinecraftClient.getInstance();
+
+		if (client == null) {
+			return;
+		}
+
+		Window window = client.getWindow();
+		long handle = window.getHandle();
+		GLFW.glfwGetCursorPos(handle, current_x, current_y);
```
comp500: I suggest using FloatBuffers with the MemoryStack API instead of arrays https://github.com/LWJGL/lwjgl3-wiki/wiki/1.3.-Memory-FAQ#1-stack-allocation

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
```diff
@@ -0,0 +1,45 @@
...
+public class FabricKeyboardImpl {
+	public static FabricKeyboardImpl INSTANCE = new FabricKeyboardImpl();
+
+	private int modKeys = 0;
+
+	private FabricKeyboardImpl() {
+	}
+
+	public boolean isKeyPressed(int keycode, int scancode) {
+		return InputUtil.isKeyPressed(keycode, scancode);
```
LambdAurora: Why having methods for that if `InputUtil` already provides them?

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
```diff
@@ -0,0 +1,97 @@
...
+	private double x = 0.0;
+	private double y = 0.0;
+	private int buttons = 0;
+	private int modKeys = 0;
+
+	private FabricMouseImpl() {
+	}
+
+	public double getX() {
+		return this.x;
```
LambdAurora: Coordinates of the mouse, how to say that...

Why does the update method calls a GLFW method and not `Mouse#getX` which can be get using a `MinecraftClient` instance?

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
```diff
@@ -0,0 +1,173 @@
...
+import net.fabricmc.fabric.api.event.client.input.MouseMoveEvent;
+import net.fabricmc.fabric.api.event.client.input.MouseScrollEvent;
+import net.fabricmc.fabric.impl.client.input.FabricKeyboardImpl;
+import net.fabricmc.fabric.impl.client.input.FabricMouseImpl;
+
+@Mixin(InputUtil.class)
+public class InputUtilMixin {
+	@ModifyVariable(method = "setKeyboardCallbacks(JLorg/lwjgl/glfw/GLFWKeyCallbackI;Lorg/lwjgl/glfw/GLFWCharModsCallbackI;)V", at = @At("HEAD"), index = 2)
+	private static GLFWKeyCallbackI changeKeyCb(GLFWKeyCallbackI keyCb) {
+		return (window, code, scancode, action, mods) -> {
```
LambdAurora: Please inject into synthetic/lambda methods of `Keyboard` instead.

Same for the mouse ones and `Mouse`.

LambdAurora (changes requested):
I honestly don't really understand the point of these events as usually all input related queries are done in tick methods or in `Screen`.

And there's nothing documented which does not help.

yyny:
> I honestly don't really understand the point of these events as usually all input related queries are done in tick methods or in Screen

Because I don't need to do the usual querying or screen handling, I need clientside events. And I don't need keybinds, I need raw key presses. See also #129 and Forge `InputEvent.KeyInputEvent`. I am not alone in this.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
yyny: Just for putting everything in one place, `isKeyPressed(Key)` is generally more useful. `FabricKeyboard` and `FabricMouse` are for the bits of information that Minecraft doesn't provide.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
yyny: See the comment on FabricKeyboardImpl above. Also, I believe Mouse#getX is only updated once per tick, not per event.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
yyny: Fair enough

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
LambdAurora: Except this will break a feature of one of my mods very badly and I won't be able, in any way (or I have to mixin in FAPI entirely to change that), to fix that.

It literally destroys the possibility of having a virtual mouse (very useful for """splitscreen""" and controller support) and will lock everything to the system mouse, which is arguably bad in this case.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.client.input;
+
+import net.minecraft.client.util.InputUtil;
+import net.minecraft.client.util.InputUtil.Key;
+
+public class FabricKeyboardImpl {
+	public static FabricKeyboardImpl INSTANCE = new FabricKeyboardImpl();
```
LemmaEOF: Is this supposed to be accessed by end users? If so, make an interface that defines these methods, then have an instance that points to this.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/input/client/InputUtilMixin.java
LemmaEOF: Yeah, this is way more destructive than Fabric wants to be.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.input;
+
+import net.minecraft.client.util.InputUtil.Key;
+
+public abstract class GenericKeyEvent {
```
LemmaEOF: Instead of allocating a new object, just pass these objects directly.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseMoveEvent.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.input;
+
+import net.fabricmc.fabric.api.client.input.v1.FabricMouse;
+
+public class MouseMoveEvent extends GenericMouseEvent {
```
LemmaEOF: This is gonna allocate a whole new object every single time the mouse moves. Not acceptable for literally four doubles and two ints. Pass them directly to the lambda.

LemmaEOF (changes requested):
I don't think this is gonna fly as it is right now. The creation of holder objects instead of just passing the raw values is gonna create a lot of allocation bloat, and that's not gonna fly. It makes sense for Forge's annotation-based event handler, but not in a post-lambda world.

modmuss50 (commented):
I have a feeling this is out of scope for fabric, for listening for key presses we have a keybinding API, events like this cover a tiny amount of cases were a keybind wont work.

As for the mouse events, most people will be using their own screen, all of this data is already passed into screen via the vanilla methods.

What is your use case for this? I dont see many mods needing to use these, its a huge amount to maintain and support for very little gain. For your mods small targetted mixins might be better IMO.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
yyny: No, this is just the singleton implementation for `FabricKeyboard`. I guess it could be a static class instead, though.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
yyny: That's true right now, but [there are several ways around that](https://github.com/FabricMC/fabric/pull/1272#discussion_r559790013).

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseMoveEvent.java
yyny: See `GenericMouseEvent`, there is a lot more information that is relevant for each event. Also, see the comment above.

yyny:
> What is your use case for this

It's a port for a scripting API. Scripts listen for keyboard events to implement minigames or special items without having to write java.

yyny:
> its a huge amount to maintain and support for very little gain

I mean its ~1000loc including comments so I don't know about that first part

> I have a feeling this is out of scope for fabric, for listening for key presses we have a keybinding API, events like this cover a tiny amount of cases were a keybind wont work.

[This](https://github.com/FabricMC/fabric/pull/1272#discussion_r560905226) seems like at least one reason why having a single unifying event helps portability.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
yyny: Interesting concern. I don't know how your mod feature is implemented, but it sounds like something that should be supported by fabric.

LambdAurora: The way it's done is by ignoring the GLFW callback and calling the mouse position update manually of `Mouse`:
https://github.com/LambdAurora/LambdaControls/blob/1.16/fabric/src/main/java/me/lambdaurora/lambdacontrols/client/controller/InputManager.java#L73

Not calling the GLFW method when using a virtual mouse is obviously done to be able to unlink entirely the physical mouse from the game, which is needed for "splitscreen" with one instance using entirely the controller with virtual mouse.

yyny:
> Please inject into synthetic/lambda methods of Keyboard instead.

I am working on this, btw, but sponge is causing my gradle builds to crash.

LemmaEOF:
> This seems like at least one reason why having a single unifying event helps portability.

you're literally citing someone saying that your PR will break things and make them less portable...

yyny:
> you're literally citing someone saying that your PR will break things and make them less portable...

Yes, in it's current state.

yyny:
I think that the fact that the naive implementation I wrote here already has portability issues shows that there is demand for a more complete unifying event. If not in fabric, then in a library or another project like `architectury`.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacks.java
```diff
@@ -0,0 +1,136 @@
...
+import net.fabricmc.fabric.api.event.client.input.CharEvent;
+import net.fabricmc.fabric.api.event.client.input.ClientInputEvents;
+import net.fabricmc.fabric.api.event.client.input.KeyEvent;
+import net.fabricmc.fabric.api.event.client.input.KeybindEvent;
+import net.fabricmc.fabric.api.event.client.input.MouseButtonEvent;
+import net.fabricmc.fabric.api.event.client.input.MouseMoveEvent;
+import net.fabricmc.fabric.api.event.client.input.MouseScrollEvent;
+import net.fabricmc.fabric.mixin.event.input.client.KeyBindingMixin;
+
+public final class InputCallbacks {
```
i509VCB: I'd almost call this `InputEventsImpl` so people don't try to import it.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
i509VCB: I don't think a singleton would be needed here since the client is also backed by a singleton. Static accessors via MinecraftClient.getInstance should be fine.

i509VCB (changes requested):

yyny:
> I have a feeling this is out of scope for fabric

I'd like to clarify for just a bit, I didn't originally submit this pull request because I think fabric absolutely needs it, I'm submitting it because _I_ need it, because issue #129 from 2019 hasn't neither been closed nor even been commented on yet, and because I just wanted to get this part of my mod done instead of waiting another year. I already wrote most of the original code before I even considered submitting it. From my POV, this is the right place to put this code, because:

 1. I am porting a large legacy mod that I am intending to be compatible and portable for a long time to come
 2. Solving portability issues that have nothing to do with my code directly shouldn't be a concern for me, but for fabric itself.
 3. Solving portability issues together in a single repo seems much more efficient than on an individual mod-by-mod basis, especially for fundamental systems like user input (Case point: Within days of me submitting the pull request someone immediately reported an issue).
 4. Solving portability issues seems to be the entire reason why fabric-api exists in the first place.  
     > Fabric API is the library for essential hooks and interoperability mechanisms for Fabric mods. Examples include: Adding events, hooks, and APIs to improve interoperability between mods.
 5. Forge has had these events too without issues
 6. There was interest in events like this all the way in 2019, so clearly there is a demand for it in Fabric too.
 7. The events themselves are rather trivial, it's the implementation that is the only issue here, I'd rather argue about that than about its existence in fabric because, at least from my POV, it seems obvious.

Preferably, we would want to support @LambdAurora's use-case as well, but once again, maintenance for that has to be done collaboratively. If it wasn't for this pull request to bring other people in on it, I wouldn't have even _known_ my mod has compatibility issues.

Also:

> for listening for key presses we have a keybinding API

I'm pretty sure the API does not actually handle that

LambdAurora:
>  >   for listening for key presses we have a keybinding API
> 
> I'm pretty sure the API does not actually handle that

It kind of does, the API allows you to register keybindings and the "normal" way of listening for key presses with those is calling `KeyBinding#wasPressed` in a tick method.

Though if a KeyBinding press event was done it would need to take into consideration mods that allows keybindings to be pressed using multiple keys, which becomes quite complicated very quickly.

yyny:
> It kind of does, the API allows you to register keybindings and the "normal" way of listening for key presses with those is calling `KeyBinding#wasPressed` in a tick method.
> 
> Though if a KeyBinding press event was done it would need to take into consideration mods that allows keybindings to be pressed using multiple keys, which becomes quite complicated very quickly.

So then the API only handles registering keybindings then, once again what I need is an event. It would be inefficient to check key bindings on ticks where no new input key was pressed. Also, I need to check _all_ keybindings, not just specific ones (Well, technically, for right now I only need the Vanilla keybindings but I can see exposing modded keybinds at some point aswell). If other mods want to do things like pretend a key is being pressed or allow for multi-key keybinds, then that just seems like one more reason for me to add input events to fabric-api.

You do make a good point that perhaps these events should only be emitted once per tick, although I see no real harm with the current implementation regarding that.
Also, I'll have to test if the current implementation works for custom keybindings as you mentioned.

yyny:
Apparently, Forge only sends `InputEvent`s when there isn't a screen being opened. I think the event system implemented here is more versatile, but it's at least something to consider.

If all Events were Observables, then we could simply (and efficiently) do `InputEvents.KEY_PRESSED_FORGE = InputEvents.KEY_PRESSED.filter(key -> MinecraftClient.getInstance().currentScreen == null);`.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/input/v1/FabricMouse.java
```diff
@@ -16,29 +16,47 @@
...
+	 * @param button the GLFW.GLFW_MOUSE_BUTTON to check
+	 */
 	public static boolean isButtonPressed(int button) {
 		return FabricMouseImpl.isButtonPressed(button);
 	}
 
+	/**
+	 * Get the GLFW.GLFW_MOD modifier keys currently being pressed.
+	 */
 	public static int getModKeys() {
```
i509VCB, line 68: This seems duplicated from keyboard stuff?

yyny: Already resolved in 
29ab4bc?

It's added to `FabricMouse` as well because that's the way GLFW does it. Seems useful when you want to implement e.g. CTRL + LMB

yyny: OS events typically duplicate modifier keys in mouse events as well.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
yyny: Should be fixed by d7e6b2f

Arguably there should be a way to inject and modify input events as they come in, but that can come later. Once again, something more of a concern for a potential future `Observable<EventType>` implementation than the `Event<ListenerType>` implementation we have right now.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/CharEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.input;
+
+public class CharEvent {
```
i509VCB: From the big guy himself (modmuss)

These classes need to be inlined into the event interfaces. I don't see any real reason to expose any interfaces here as there is very little extra stuff you could be exposing here.

Simply inline the methods and all into the parameters of each callback interface.

i509VCB (changes requested):

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/CharEvent.java
yyny: For this event, definitely, but for `(double x, double y, double dx, double dy, int buttons, int modKeys, double scrollX, double scrollY) -> { /* ... */ }` and `(int code, int scancode, int action, int modKeys, Key key, Keybind keybind) -> { /* ... */ }`? My intent is to move away from multi-parameter event listeners entirely since they have several drawbacks and no direct benefits but we'll save that for another pull request. I will try to remove the object allocations at the very least.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/ClientInputEvents.java
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	public static final Event<FileDrop> FILE_DROPPED = EventFactory.createArrayBacked(FileDrop.class, listeners -> paths -> {
+		for (FileDrop listener : listeners) {
+			listener.onFilesDropped(paths);
+		}
+	});
+
+	@FunctionalInterface
+	public interface KeyState {
+		void onKey(KeyEvent key);
```
i509VCB: To clarify, all the event objects just not char

yyny: So we are doing 6 parameter listeners? Doesn't that seem extremely inconvenient?

yyny:
> I don't see any real reason to expose any interfaces here as there is very little extra stuff you could be exposing here

This is always subjective, most events that get passed a `MinecraftClient` don't strictly need that either since they can always call `MinecraftClient.getInstance()`.

In the case of `KeyEvent`, I chose to add `Key key;` in addition to the `int code, int scancode, int action, int modKeys` from the raw GLFW event because it already has to be looked up anyway for generating a `KeybindEvent` and because it seems useful enough to have. Same situation with `KeybindEvent.key` and `MouseButtonEvent.key`. All the others, like `GenericKeyEvent#isPressed` and `MouseScrollEvent.cursorX` are there purely for convenience, and because they are cheap to come by.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/input/v1/FabricKeyboard.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.input.v1;
+
+import net.minecraft.client.util.InputUtil.Key;
+
+import net.fabricmc.fabric.impl.client.input.FabricKeyboardImpl;
+
+public final class FabricKeyboard {
```
liach, line 35: `@Environment`

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/client/input/v1/FabricMouse.java
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.input.v1;
+
+import net.fabricmc.fabric.impl.client.input.FabricKeyboardImpl;
+import net.fabricmc.fabric.impl.client.input.FabricMouseImpl;
+
+public final class FabricMouse {
```
liach, line 31: `@Environment`

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/ClientInputEvents.java
liach: We are doing 6 parameters. We don't need to create extra objects that way.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
liach: What's the purpose of this class at all? You can just make it an interface or a final pojo than this weird abstract class which is abstract for no advantage

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
```diff
@@ -0,0 +1,27 @@
...
+	public abstract int getCode();
+	public abstract int getScancode();
+	public abstract int getModKeys();
+	public abstract boolean isPressed();
+	public abstract Key getKey();
```
liach: Hmm, can you get other codes from this key? If you can I suggest change those into static methods so only people who need will convert to those codes

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericMouseEvent.java
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.input;
+
+public abstract class GenericMouseEvent {
```
liach: Same, just make it a pojo and move some impl details here than leaving such an abstract weird thing.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseButtonEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+	public MouseButtonEvent(int button, int action, int modKeys, Key key) {
+		super(FabricMouse.getX(), FabricMouse.getY(), 0.0, 0.0, FabricMouse.getPressedButtons(), FabricMouse.getModKeys());
+		this.button = button;
+		this.action = action;
+		this.key = key;
+		this.modKeys = modKeys;
+	}
+
+	public Key getKey() {
```
liach: No need for this if you've gone all the way public final fields

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseMoveEvent.java
```diff
@@ -0,0 +1,34 @@
...
+	public final double x;
+	public final double y;
+	public final double dx;
+	public final double dy;
+
+	public MouseMoveEvent(double x, double y, double dx, double dy) {
+		super(x, y, dx, dy, FabricMouse.getPressedButtons(), FabricMouse.getModKeys());
```
liach: Why do you grab stuff from static methods if people can just grab them from the methods on demand? I see in impl that they are just field values

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricKeyboardImpl.java
liach: These input util calls imo can just be moved to api. The method bodies of api is effectively impl.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+		Window window = client.getWindow();
+		long handle = window.getHandle();
+
+		try (MemoryStack stack = MemoryStack.stackPush()) {
+			DoubleBuffer current_x = stack.callocDouble(1);
+			DoubleBuffer current_y = stack.callocDouble(1);
+			GLFW.glfwGetCursorPos(handle, current_x, current_y);
+			FabricMouseImpl.x = current_x.get();
+			FabricMouseImpl.y = current_y.get();
```
liach: How much does this hurt, I wonder...

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
```diff
@@ -0,0 +1,216 @@
...
+
+public final class InputCallbacksImpl {
+	private static final KeyEvent sharedKeyEvent = new KeyEvent(-1, -1, GLFW.GLFW_RELEASE, 0);
+	private static final KeybindEvent sharedKeybindEvent = new KeybindEvent(-1, -1, GLFW.GLFW_RELEASE, 0, null, null);
+	private static final MouseMoveEvent sharedMouseMoveEvent = new MouseMoveEvent(0.0, 0.0, 0.0, 0.0);
+	private static final MouseButtonEvent sharedMouseButtonEvent = new MouseButtonEvent(0, GLFW.GLFW_RELEASE, 0, null);
+	private static final MouseScrollEvent sharedMouseScrollEvent = new MouseScrollEvent(0.0, 0.0);
+
+	private static KeyEvent createKeyEvent(int code, int scancode, int action, int modKeys) {
+		KeyEventAccessor accessor = (KeyEventAccessor) (Object) sharedKeyEvent;
```
liach: This can't be worse. You create something people believe are pojo; they are not. You just fuck up people's assumptions this way, especially that you don't document any of these :angry:

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseButtonEvent.java
i509VCB: public final fields is code smell imo - I think yyny should just commit to getters

liach: This debate is currently pointless because these objects are really all mutable, but mutation is done through mixin which is really dumb

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
yyny: That was the original idea, to only calculate values on demand, but we didn't want interfaces, and we didn't want record-likes, so we'll have to pre-calculate the raw underlying values and pass the information directly (and let the users figure it out what to do with them). I've optimized as much as possible for the common case where nobody is listening to input events and added raw events for those people who don't care about the calculated values, so the situation regarding this should be a bit better now at least.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/GenericKeyEvent.java
yyny: Most of these had to be precalculated anyway, these getters were mostly to help people who DON'T need some values, for example, I can imagine most people would never need to check the scancode.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseButtonEvent.java
yyny: Well IMHO non-composable, unoptimizable events and lambda functions that copy all their arguments seems really dumb too but I'll stick to convention even if it's dumb. Also if we don't want allocations, we don't want mutable events, _and_ we don't want getter interfaces, then this is the only option left anyway.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/api/event/client/input/MouseMoveEvent.java
yyny: Outdated, but should have been read directly from the Impl, yes.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
yyny: I've removed it, don't need it if we cache the mouse position anyway.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
yyny: I've already talked about this, if we aren't going to allow anything other than multi-arg lambda's, then we'll just have to use those.

yyny:
> These input util calls imo can just be moved to api. The method bodies of api is effectively impl.

aight

yyny:
I'm not entirely happy with `FabricKeyboard` and `FabricMouse` since they duplicate vanilla functionality, but I don't see a way to give the users the modifier keys any other way. I know from experience that modKeys don't always directly correlate to their corresponding key, for example some Linux window managers remap them, and some mice have special buttons for them, but I think people with setups like that know what they are getting into, so we might still just force users to do e.g. `Keyboard.isPressed(GLFW.GLFW_KEY_CTRL)` anyway. If users _really_ need the modifier keys, we can keep them available in the raw input events (`ClientInputEvents.KEY` and `ClientInputEvents.MOUSE_BUTTON`). Would be code duplication if multiple API users need them, but oh well.

Same goes for the `modKey` and `scancode` parameters, we could consider removing those from the non-raw input events, Vanilla never uses them for anything, and I suspect most mods don't either. This would help reduce the number of arguments in lambda functions. Note that I do need at least the `scancode` for my particular use case, but this is the dilemma we get ourselves into when we only allow multi-argument lambda listeners. Either everyone has to pay the price of unused parameters, or someone has to pay the price of duplicated calculations (Or we can add 20 different events for every possible permutation of input arguments to fake composability, but yeah, let's not. I'd still argue record-like events (Immutable object sor getter interfaces) are the better solution here).

In it's current state, `net.minecraft.client.Mouse.getX()` and `net.minecraft.client.Mouse.getY()` are left outdated while we are in event listeners, in the same way it is outdated while inside a `Screen` handler. We can fix this ourselves, but that would break compatibility with vanilla (Although I suspect this is a bug in Minecraft). We could also inject after Vanilla has updated them, but that would limit the usability of these events, since then screen handlers will already have been called. Finally, we could deprecate `getX` and `getY`, although I don't know how harmful/useful they really are compared to `FabricMouse`. For now I will document the difference without making any additional changes.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
```diff
@@ -0,0 +1,82 @@
...
+			return;
+		}
+
+		Window window = client.getWindow();
+		long handle = window.getHandle();
+
+		try (MemoryStack stack = MemoryStack.stackPush()) {
+			DoubleBuffer current_x = stack.callocDouble(1);
+			DoubleBuffer current_y = stack.callocDouble(1);
+			GLFW.glfwGetCursorPos(handle, current_x, current_y);
```
LambdAurora: This has to be removed and `Mouse#getX`, `Mouse#getY` should be used instead, they are cached and don't need extra JNI calls or object allocation.

LambdAurora (changes requested):
I also noticed another **big** flaw.

Most of the events **never** check if the window they're triggered for is the same as the MC's window, which destroys the possibility for mod authors to create their own windows (even if really niche, this has to be considered).

In the current state this PR will still cause too many issues.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
```diff
@@ -0,0 +1,151 @@
...
+							|| ClientInputEvents.KEY_RELEASED.hasListeners()
+							|| ClientInputEvents.KEY_REPEATED.hasListeners()
+							|| ClientInputEvents.KEYBIND_PRESSED.hasListeners()
+							|| ClientInputEvents.KEYBIND_RELEASED.hasListeners()
+							|| ClientInputEvents.KEYBIND_REPEATED.hasListeners();
+
+		if (hasListeners) {
+			Map<InputUtil.Key, KeyBinding> keyToBindings = KeyBindingAccessor.getKeyToBindings();
+			Key key = InputUtil.fromKeyCode(code, scancode);
+			KeyBinding binding = keyToBindings.get(key);
```
liach: You should move this to some sort of cached supplier, which you can then expose to your event callbacks, instead of adding a `hasListeners` in the events.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
```diff
@@ -43,6 +43,17 @@ public final T invoker() {
...
 	}
 
+	/**
+	 * Some events can be optimized away if nobody is listening. If an Event
+	 * implementation can quickly check for this condition, it may override this
+	 * method to return false in that case.
+	 *
+	 * @return True if the event might have any listeners.
+	 */
+	public boolean hasListeners() {
```
liach: I don't know how this works out with the concurrency-compatible design of fab events (that's why invokers are a field because player wants you to update it real time than through some delay). But such logic usually have problems in a concurrent setting, where you are supposed to just try do an action and see its results than ask if you can do and then do an action (during the ask and do something else can meddle)

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
liach:
```java
Lazy<KeyBinding> lazyBinding = new Lazy<>(() -> {
    Map<InputUtil.Key, KeyBinding> keyToBindings = KeyBindingAccessor.getKeyToBindings();
    return keyToBindings.get(InputUtil.fromKeyCode(code, scancode));
});
```
Hmm, I now wonder if you should rather pass in like a functional interface with a method `KeyBinding getKey(int scancode, int keycode)` for performance reasons. It's debatable, but not of priority.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
i509VCB: If I recall player does do COW for adding new listeners in the events.

liach: Yes, I believe he does that though generating new invokers through the function supplied to `createArrayBacked`, which is an array to created invoker function. In an invocation the old invoker can be removed from the event instance but its invocation will continue, which is effectively copy-on-write.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
```diff
@@ -0,0 +1,151 @@
...
+		FabricKeyboardImpl.updateModKeys(modKeys);
+
+		ClientInputEvents.KEY.invoker().onKeyChanged(code, scancode, action, modKeys);
+
+		boolean hasListeners = ClientInputEvents.KEY_PRESSED.hasListeners()
+							|| ClientInputEvents.KEY_RELEASED.hasListeners()
+							|| ClientInputEvents.KEY_REPEATED.hasListeners()
+							|| ClientInputEvents.KEYBIND_PRESSED.hasListeners()
+							|| ClientInputEvents.KEYBIND_RELEASED.hasListeners()
+							|| ClientInputEvents.KEYBIND_REPEATED.hasListeners();
```
yyny: I would usually use spaces here so it would look the same even without `.editorconfig`. I'm assuming this is fine considering `.editorconfig` has `tab_width = 4`.

Juuxel: The general style used through Fabric API's codebase is to always use two tabs for continuation indent in places like this, not five. (Using tabs for visual alignment is unreliable as seen here.)

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/FabricMouseImpl.java
yyny: I forgot to remove it, it's already unused.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
yyny: This seems to add a lot of complexity and inconvenience, and an extra object allocation every key-press which I was told wasn't allowed, for very little benefit. If we are going to allocate a functional interface anyway, I would much rather go back to the original `KeyEvent` implementation, considering all the benefits I have outlined for that approach earlier.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
yyny: I don't see this as an issue right now because `ArrayBackedEvent` already has a race condition anyway, the `hasListeners` only makes the race window slightly bigger, might need to be documented but not sure about refactoring. Notice the small window between [updating the handlers](https://github.com/FabricMC/fabric/blob/ed0c348cc4cda9cff8abc5953b370fa872cba943/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java#L64-L71) and [changing the invoker](https://github.com/FabricMC/fabric/blob/ed0c348cc4cda9cff8abc5953b370fa872cba943/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java#L49). If an `Event` implementation _really_ wants to be as concurrent as possible it doesn't even have to implement this function anyway. This approach can only ever extend the window if an invoker is added in between the call to `Event#hasListeners()` and the call to `Event#invoker()`. As with the `dummyInvoker` optimization in `ArrayBackedEvent` or the COW, there really is no way to do this optimization without creating a race condition. Do we want to optimize away ~100 bytecode instructions for every event invocation (If no-one is listening), or do we want event listeners to start being able to receive events ~100 bytecodes earlier? It doesn't seem like a big deal to me to not invoke listeners for the sake of optimization for events that were emitted before the listener was even added.

Another option I considered was to instead expose a `T Event#noop()` that the `invoker` can be compared against (Would return `this.dummyHandler` in `ArrayBackedEvent`), this doesn't change the race condition in any way, but it might make it more obvious that the `invoker()` is slightly out of date by the time it gets invoked. I preferred the boolean approach purely for performance reasons because it doesn't force the event implementation to cache a `noop` handler if it wants to have this optimization.

yyny:
> I also noticed another **big** flaw.
> 
> Most of the events **never** check if the window they're triggered for is the same as the MC's window, which destroys the possibility for mod authors to create their own windows (even if really niche, this has to be considered).

This doesn't seem like a big flaw, if the mod doesn't take care to mix into the input event handlers, then the Vanilla code would be executed as well, since Minecraft doesn't check this condition either. And I've mentioned this before, but if we have an input event, it would be more obvious for mod developers that want to do crazy things like that to figure out how they should go about implementing such a feature without breaking any mods.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/impl/client/input/InputCallbacksImpl.java
yyny: > Using tabs for visual alignment is unreliable as seen here

Yep, I configured my editors to always use spaces for visual alignment for exactly this reason, but the style checker didn't allow it. I just regex replaced them and didn't notice this until now.

LemmaEOF:
None of the changes made have fixed the fundamental problem: this way of doing things is antithetical to the Fabric ethos. This PR will *still* break dozens of mods and add an unacceptable amount of runtime bloat. I'm exercising my authority as a member of the triage team to close this PR. Put this in your own library; it's literally free.

fabric-input-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/input/client/KeyboardMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.Keyboard;
+
+import net.fabricmc.fabric.impl.client.input.InputCallbacksImpl;
+
+@Mixin(Keyboard.class)
+public class KeyboardMixin {
+	@Inject(method = "method_22678(JIIII)V", at = @At("HEAD"))
```
i509VCB, line 32: Could you please leave a comment saying which method owns this synthetic method in the signature for all of these across keyboard and mouse?

i509VCB:
Reopening this since the implementation has changed and actually uses synthetic method injects which is more mod compatible and does not use the extra event object anymore as seen in e11c4b8 and beyond, just some areas still need to be ironed out such as some possibly redundant methods being dropped and documentation being used to refer to the builtin methods instead.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
```diff
@@ -43,6 +43,17 @@ public final T invoker() {
 		return invoker;
 	}
 
+	/**
+	 * Some events can be optimized away if nobody is listening. If an Event
+	 * implementation can quickly check for this condition, it may override this
+	 * method to return false in that case.
```
i509VCB: I don't agree with this `hasListeners` method being necessary atm. What will occur is that if there is no invoker, the event will generate a sort of "dummy invoker" based on the lambda in the EventFactory.createArrayBacked(..., Function<..., ...>) thing.

If you are truly concerned about performance, you can specify an empty invoker manually via the other `createArrayBacked` method which uses your specified invoker to handle the event. Do note that the little overhead you get from this is not going to be very negligible unless it's something like events in rendering which specify the empty invoker for performance reasons.

liach:
@yyny 
1. Thanks i509 for reopening. After another glance this is really fantastic
2. For your `hasListeners`, I am offering a different approach:


Now you can write this for your events: (please check the code, I wrote it by hand on github and this may have compile errors)
```java
	public static final Event<KeyState> KEY_REPEATED = EventFactory.createArrayBacked(KeyState.class, listeners -> (code, scancode, modKeys, nullKey) -> {
		// no listeners? no object creation! but still need to check keybind version
		if (listeners.length == 0) {
			// change the event name to the corresponding keybind version
			KEYBIND_REPEATED.invoker().onKeybind(code, scancode, modKeys, null, null); // let them compute both if they need to
			return;
		}
		
		Key key = InputUtil.fromKeyCode(code, scancode);
		
		for (KeyState listener : listeners) {
			listener.onKey(code, scancode, modKeys, key); // listeners are guaranteed to receive non-null keys
		}
		// same, change event name
		KEYBIND_REPEATED.invoker().onKeybind(code, scancode, modKeys, key, null); // let them compute keybind on need
	});

	// Keybind version
	public static final Event<KeybindState> KEYBIND_REPEATED = EventFactory.createArrayBacked(KeybindState.class, listeners -> (code, scancode, modKeys, nullableKey, nullKeybind) -> {
		// no extra object creation
		if (listeners.length == 0) {
			return;
		}
		
		// So when we pass key and key bindings to invoker() we pass null; we compute it here
		Key key = nullableKey == null ? InputUtil.fromKeyCode(code, scancode) : nullableKey; // compute if absent
		KeyBinding binding = KeyBindingAccessor.getKeyToBindings().get(key); // compute it here
		
		for (KeybindState listener : listeners) {
			listener.onKeybind(code, scancode, modKeys, key, binding); // listeners are guaranteed to receive non-null keys and bindings
		}
	});
```
And replace this huge chunk of code at https://github.com/FabricMC/fabric/pull/1272/files#diff-2ee30e5d1b4f872827ccff6490183560cffe52bc0b983ffc0005b0b5c08c5698R39-R76 with:
```java
	// So when we pass key objects to invoker() we pass null; we compute it on demand
	switch (action) {
	case GLFW.GLFW_PRESS:
		ClientInputEvents.KEY_PRESSED.invoker().onKey(code, scancode, modKeys, null);
		break;
	case GLFW.GLFW_RELEASE:
		ClientInputEvents.KEY_RELEASED.invoker().onKey(code, scancode, modKeys, null);
		break;
	case GLFW.GLFW_REPEAT:
		ClientInputEvents.KEY_REPEATED.invoker().onKey(code, scancode, modKeys, null);
		break;
	}
```

Thus, `hasListeners` is rendered useless and can be safely removed.

Wait a sec, I am not that pleased. Please let me open ide and give you a better version

liach:
Just fyi I made https://github.com/yyny/fabric/pull/1 which includes my suggestions. Now I think my suggestions are still quite cursed.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
RedstoneParadox: I'd personally say that it's out of scope for this PR.

yyny: I'm not very concerned about performance, but since having no listeners on this event will be very common, might as well optimize for it.

liach:
@Technici4n Should this target 1.18 or even 1.19 now, or will we merge into 1.16 first and then port?

Technici4n:
I'd say target 1.19 as it will take a long time to properly review (quite low priority).

Phoupraw:
Mark and top.

PR #1278, opened 2021-01-23
The Council Of Bikeshed (Transfer API v1)

### LINKED ISSUE #1166: The Fluid API discussion issue

Technici4n:
**Disclaimer**: _Yes, this is a fluid api PR. Please keep the discussion constructive, or I will make sure that this issue is locked to contributors._

This PR is very big, and I expect many questions, so please open a code review thread to discuss a specific part of it.

This PR is also a draft. Please review the API as much as possible, as I think it is close to its final state. I am opening this PR already because the implementation details are taking a while, but there is no need to delay the review of the API for that... So please, I need your comments on the API already while I polish the implementation details and the javadoc. If you want to comment on that too, go ahead, I appreciate any feedback.

I am using `[WIP]` blocks in this message to indicate what still needs to be done, the rest can be considered ready for review.

# Fabric Transfer API v1
## Introduction
Building on #1234, this pull requests introduces `fabric-transfer-api-v1`: a new module for generic resource transfer. That module contains generic resource transfer classes and helpers, and a few helpers to make working with fluids and items easier. Generic means that the API can be used to transfer fluids, items, energy, gasses, enchanted potions, Thaumcraft essentia, or any other quantifiable game resource that can be imagined. (Note: the diff contains #1234 as well, I will rebase when that is merged.)

This pull requests also introduces an `item` package to `fabric-api-lookup-api-v1` which contains ApiLookups for item-provided Apis, and `ItemKey` which is the equivalent of `Fluid` for item transfer.

Many thanks to @Grondag who wrote Fluidity, of which this PR is very much inspired.

Closes #1166.

### Problem scope
The problems this API aims to solve is:
* Item transfer between blocks. Example use case: AE2, item pipes, machines, ...
* Fluid transfer between blocks. Example use case: AE2, fluid pipes, machines, ...
* Fluid transfer with items. Example use case: emptying and filling buckets or modded containers, from the player inventory or from a machine inventory.

More specifically, when we talk about transfer, we mean:
* Insertion of resources.
* Extraction of resources.
* Reading of available resources (think ME Storage Bus).

### The core transfer API: `Storage` and `StorageView`
[`Storage<T>`](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java) is a container for resources of type `T`. It allows inserting, extracting, and reading the contents. `getVersion` is optional, and is really meant for AE2-like mods to skip checking inventories that haven't change since the last scan.

[`StorageView<T>`](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java) represents some stored resource, and optionally allows extracting it as well.

Note that there is an asymmetry between insertion and extraction.
That is because when we are inserting we know what we want to insert,
but in some cases we don't know what exactly to extract and we have to read the inventory's storage.
That's why it's also possible to extract directly from a `StorageView`.

[WIP] Base implementations (combined, insert only, extract only).

### Transactions
Note that multiple functions have a transaction parameter. A transaction is a global operation where participants guarantee atomicity: either the whole operation succeeds, or it is completely cancelled. In that case, we say that it is aborted. Nested transactions are also supported.

See [the `Transaction` class](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/Transaction.java) for more details. Important detail: participants are automatically enrolled in all the transactions of the transaction stack.

### Fluid API
The fluid api is simply sided access to `Storage<Fluid>` instances.
Unsided access may be considered for a later PR.

[WIP] Cauldron compatibility is NYI, and will be added soon for regular 1.16 water cauldrons. For 1.17, I have a plan to allow registration of modded cauldrons.
[WIP] I want to provide a `SimpleFluidStorage` implementation, and will get to it eventually.

### Item API
The item api is sided access to `Storage<ItemKey>` instances. `ItemKey`s are an item, and an immutable associated NBT tag. See [the class](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java) for more details. It is planned that no mods will be using `Inventory` and `SidedInventory` directly after this API is merged. Thanks to a fallback provider in the `BlockApiLookup`, vanilla containers and modded containers relying on the vanilla interfaces are still supported.

`Inventory`, `SidedInventory` and `PlayerInventory` wrappers are also available: [InventoryWrappers](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryWrappers.java). Note the existence of an `offerOrDrop` function with transaction support.

Again, unsided access may be considered for a later PR.

[WIP] I will provide a base implementation for "simple" item storages.
[WIP] I will provide a `Slot` implementation to allow interacting with `ItemKey`-based storages. This is quite tricky as all the vanilla `ScreenHandler` methods mutate the returned item stacks directly.

### Item-provided APIs
This PR adds `ItemApiLookup` and `ItemApiLookupRegistry` to the `fabric-api-lookup-api-v1` module (see #1234).

This PR then defines [`ContainerItemContext`](https://github.com/FabLabsMC/fabric/blob/transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java) to allow items to interact with the underlying container.

It uses these two building blocks to define the fluid api for container items. Item-containing items (think shulker boxes or backpacks) are out of scope for this PR.

[WIP] I will explain in detail how it works.
[WIP] Vanilla compat is not done yet, for now only water and lava bucket emptying is supported.

#### Unsided APIs
Unsided APIs are out of scope because they are not essential for mod compatibility, and the requirements are not that clear to me. The community can use the provided transfer APIs and the `fabric-api-lookup-api-v1` module to experiment with that.

#### Change notification
I also believe that storage change notification is way too complicated to be added to the API. It requires listener management for inventories, and it introduces the need to identify the slot where the change happened, which is non-trivial for complicated inventories such as ME Chests. It would also be optional (we can't support it for `Inventory` wrappers anyway), and so it would still require special casing by the api consumer. I hope that `Storage#getVersion` will be good enough for small and mid size storages. Very large storages probably require a specialized thrid party API anyway, for which this PR can probably provide a lot of helpful building blocks.

[WIP] An extensive test mod will be added, for now I am porting my own mod Modern Industrialization to this API.
[WIP] A README will be added too, and a lot of javadoc, including `package-info.java`.


CoolMineman (changes requested):
Fractional items just so you can make fractional fluids? Just enforce a sane fixed unit so we don't get stuck with a bunch of mods using 1/1000 and not working with bottles as well as small amounts of fluid not transferring between mods as their denominators are incompatible.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/Participant.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.transaction;
+
+/**
+ * An object that can take part atomically in a {@link Transaction}.
+ * @see Transaction
+ */
+public interface Participant<State> {
```
liach: Probably give this type parameter a nicer name like `S`

Ueaj-Kerman:
with a fixed denominator, you also don't need StorageFunction, which IMO just adds unneeded complexity. 

- it isn't type-safe since there's no difference between an extracting storage function and inserting one's type. 
- isEmpty is the wrong name for an inserting function, it should be isFull
- extra object allocation every time the method is called (lambda, no one is going to cache it, I guarantee you)
- there is no way of knowing whether or not a container just can't be extracted from, or is just empty
- it's just a glorified wrapper for a method

Ueaj-Kerman (changes requested):
I'd rather have fractions than per-container denominators

liach (commented):
Also your assumption that the numbers are fractions imo is quite unnecessary. You should just use integers all the way. If there is fractions I think they can create custom subtypes of transactions

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorage.java
```diff
@@ -0,0 +1,57 @@
...
+package net.fabricmc.fabric.api.transfer.v1.base;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+import net.fabricmc.fabric.api.transfer.v1.storage.StorageFunction;
+
+public class CombinedStorage<T, S extends Storage<T>> implements Storage<T> {
```
i509VCB: What would occur if we tried to combine two different storage types using some unsafe casts?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorage.java
```diff
@@ -0,0 +1,57 @@
...
+		return insertionFunction;
+	}
+
+	@Override
+	public StorageFunction<T> extractionFunction() {
+		return extractionFunction;
+	}
+
+	@Override
+	public boolean forEach(Visitor<T> visitor) {
```
i509VCB: You don't forEach a visitor, this should be `accept`

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
```diff
@@ -0,0 +1,52 @@
...
+
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookupRegistry;
+import net.fabricmc.fabric.api.lookup.v1.item.ItemApiLookup;
+import net.fabricmc.fabric.api.lookup.v1.item.ItemApiLookupRegistry;
+import net.fabricmc.fabric.api.transfer.v1.context.ContainerItemContext;
+import net.fabricmc.fabric.api.transfer.v1.fluid.base.FluidContainingItems;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+
+public class FluidApi {
```
i509VCB: final

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/FluidContainingItems.java
```diff
@@ -0,0 +1,103 @@
...
+	public static ItemApiLookup.ItemApiProvider<Storage<Fluid>, ContainerItemContext> getFullItemProvider(
+			Item emptyVariant, Fluid fluid, long numerator, long denominator) {
+		ItemPreconditions.notEmpty(emptyVariant);
+		FluidPreconditions.notEmpty(fluid);
+		Preconditions.checkArgument(numerator > 0);
+		Preconditions.checkArgument(denominator > 0);
+		return (itemKey, ctx) -> new SimpleFluidContainingItem(ctx, itemKey, emptyVariant, fluid, numerator, denominator);
+	}
+
+	private static class SimpleFluidContainingItem implements Storage<Fluid>, FixedDenominatorStorageView<Fluid> {
```
i509VCB: To the impl

i509VCB (commented):
This is not going to be easy to review until the prerequisite pr is done

Ueaj-Kerman:
This PR also doesn't use the keyed transaction model, which IMO is the superior option. It is far more intuitive and the API handles most of the logic for you. This code snippet should make it obvious:
<table>
<tr>
<td> Direct </td> <td> Keyed </td>
</tr>
<tr>
<td>

```java
public class FluidVolume implements Participant {
	protected Fluid fluid;
	protected int quantity;

	@Override
	public int insert(Fluid type, int amount) {
		Fluid fluid = this.fluid;
		if (fluid == Fluids.EMPTY || fluid == type) {
			if (fluid != type) {
				this.fluid = type;
			}

			this.quantity = this.quantity) + amount;
			return amount;
		}
		return 0;
	}

	@Override
	public int extract(Fluid type, int amount) {
		if (this.fluid == type) {
			int oldLevel = this.quantity;
			int toExtract = Math.min(oldLevel, amount);
			int newLevel = oldLevel - toExtract;
			if (newLevel == 0) {
				this.fluid = Fluids.EMPTY;
			}
           		this.quantity = newLevel;
			return toExtract;
		}

		return 0;
	}
}
```

</td>
<td>

```java
public class FluidVolume implements Participant {
	protected final Key.Object<Fluid> fluid;
	protected final Key.Int quantity;

	@Override
	public int insert(Transaction transaction, Fluid type, int amount) {
		Fluid fluid = this.fluid.get(transaction);
		if (fluid == Fluids.EMPTY || fluid == type) {
			if (fluid != type) {
				this.fluid.set(transaction, type);
			}

			this.quantity.set(transaction, this.quantity.get(transaction) + amount);
			return amount;
		}
		return 0;
	}

	@Override
	public int extract(Transaction transaction, Fluid type, int amount) {
		if (this.fluid.get(transaction) == type) {
			int oldLevel = this.quantity.get(transaction);
			int toExtract = Math.min(oldLevel, amount);
			int newLevel = oldLevel - toExtract;
			if (newLevel == 0) {
				this.fluid.set(transaction, Fluids.EMPTY);
			}
            		this.quantity.set(transaction, newLevel);
			return toExtract;
		}

		return 0;
	}
}
```

</td>
</tr>
</table>

Technici4n:
**Note: this message is outdated, but kept for historical purposes. Please scroll down after reading it. Thanks!**

_Should Fabric API enforce a fluid denominator for all mods to use?_
**Why enforce that?**
All fluid transfer would be perfect, i.e. any transfer from tank from mod A to empty large tank from mod B through pipes from mod C would be guaranteed to transfer all the fluid.
In particular, no small fluid amount gets stuck in the A->C and C->B conversions.

**Pros of letting mod authors pick a denominator while the API handles conversions**
* The API is future-proof: if vanilla adds cauldrons with 7 levels, we can support it with no problem. If another denominator proves to be preferable in the future, we can support it.
  Picking a fixed denominator would force us to support it forever.
* Any mod is free to define any denominator who suits them best, and we don't need to debate which denominator is better.
* Containers which cannot accept all amounts of fluid (for example cauldrons can only accept thirds) are much easier to implement.
* I don't think it's Fabric's role to pick a denominator for the mods, especially when we consider the previous points, and given that everyone has a different opinion on that.

**Cons**
* It is possible for some small amounts of fluid to be stuck here and there, although **in theory** mod authors can agree on a denominator to prevent this.

**What about fractions?**
This PR doesn't include `Fraction`s because of the object allocation involved, the amount of code involved (we would probably need `Fraction` and `MutableFraction` at the very least) and the fact that any fraction operation can overflow, which makes them very hard to work with in practice in an environment where a large number of mods interact.
With unsigned `long`s however, the only overflow that can happen is when using `+`. The provided `FixedDenominatorStorage{View,Function}`
Mods who want to support any denominator can find it through `Storage#forEach` and `StorageView#denominator`, and be extra careful with overflow.

**Why not separate the item api into an integer-only api?**
I think that having to add a `1` parameter here and there in the item api is better than copy/pasting all the helper functions.

Imo the benefits of the flexible denominator approach outweigh the drawbacks. If you disagree, but don't have new evidence to bring to the table, please react 👎 and move on. This will probably be for the admins to decide anyway.

Ueaj-Kerman:
> The API is future-proof: if vanilla adds cauldrons with 7 levels, we can support it with no problem. If another denominator proves to be preferable in the future, we can support it. Picking a fixed denominator would force us to support it forever.

This is already possible with API API, and infact it has exactly the same effect as having per-mod denominators. If mojang changes the denominator in the future, API API could handle it, and we'd just have 2 incompatible denominators (until the old module is deprecated and removed) I much prefer 2 temporarily incompatible denominators, over every mod choosing their own.

> Any mod is free to define any denominator who suits them best, and we don't need to debate which denominator is better.

mods design themselves around the API, the only requirement is that the API must be compatible with vanilla. If your fluid must be in 970ths, you can put it in 1000ths, and rescale your mod a little. Which is what happens over in voldeland

> Containers which cannot accept all amounts of fluid (for example cauldrons can only accept thirds) are much easier to implement.

dunno what's so complicated about Math.floorDiv

> I don't think it's Fabric's role to pick a denominator for the mods, especially when we consider the previous points, and given that everyone has a different opinion on that.

Either, we pick one now, or the community will have to fight over it, with the users paying the cost (with fluids stuck in pipes, etc.)

> It is possible for some small amounts of fluid to be stuck here and there, although in theory mod authors can agree on a denominator to prevent this.

If we can't decide on a denominator here after 2 years of talk, what makes you think all modders will come together with one denominator? 81000, 1000, 648, 1620, 3000 already exist.


and, in the worst possible scenario, where a custom denominator is **absolutely necessary**, API API is already capable of handling it. So there really is no point

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorage.java
Technici4n: Different `S`s you mean? Different `T`s probably shouldn't happen in normal API usage scenarios.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorage.java
Technici4n: I am not sure about this. This is really to replace a `for(StorageView<T> view : /* ... */)` with something giving more control to the `Storage` implementation. It's not yet in the javadoc, but I think that `forEach` should not have to be reentrant, and it should be allowed to `throw` an exception if it's invoked by the visitor.

CoolMineman:
> Any mod is free to define any denominator who suits them best

this is a con as now users will be stuck with different fluid systems that will not work well together
> It is possible for some small amounts of fluid to be stuck here and there

this will be common as mods that treat bottles as 1/3 and mods that use 1/1000 will be common
> Containers which cannot accept all amounts of fluid (for example cauldrons can only accept thirds) are much easier to implement.

Implementing cauldrons with the old FTL (which used fixed 1/81000) was trivial
> The API is future-proof: if vanilla adds cauldrons with 7 levels, we can support it with no problem. If another denominator proves to be preferable in the future, we can support it.

You will wind up with mods using 1/1000 and not working with existing cauldrons. Additionally, if you think there is a real chance of MC adding cauldrons with a denominator of 7 just make the fixed denom support that.
> In theory mod authors can agree on a denominator to prevent this.

Lets just do this now rather than later and reduce complexity in the API while increasing compat.
> I don't think it's Fabric's role to pick a denominator for the mods

A goal of Fabric API is to increase compat between mods and this would fit within said goal.

Technici4n:
Ok, thanks a lot people for your input on this. After some consideration, I have decided to move this PR to 81000 for fluids. The description will be updated shortly.

I had indeed completely forgotten that API API can be used for seamless conversion between fluid systems if that is required, which solves my main concern of such a choice, i.e. future-proofness.
Combined with the fact that `fabric-transfer-api-v1` is kind of a "transfer API factory", anyone is free to define their own system with a different denominator and register the fallbacks if they really want to.

Additionally, we don't have to worry about denominators nor GCD conversions in this already complex API, which allowed me to remove fixed-denominator and integer base implementations.

Finally, I don't think fluids getting stuck in pipes is that big of an issue because the main pipes right now are Modern Industrialization's, for which I was always planning to use 81000 anyway. However, if some people start using `144 mb` and others `111 1/9 mb` for ingots, we will be in big trouble. That's one more reason to enforce 81000 as the standard.

I am now investigating if `StorageFunction` is still useful, or if it should be removed.

fabric-api-lookup-api-v1/README.md
```diff
@@ -0,0 +1,28 @@
+#  Fabric Api Lookup API (v1)
```
LemmaEOF, line 1: I'm so sorry but I honestly think I need to commit bikeshed and say argue that "Api Lookup Api" is a terrible name that, while being very literal, is clunky to use and ends up hurting ease of understanding of the code due to the redundancy of the term "api". Calling it a component API isn't quite accurate either, since entity-component systems are inherently about managing storage as well as lookup. I think the best term to use here would be "property", as it's looking or traits of an object without caring about how they're stored.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
```diff
@@ -0,0 +1,105 @@
...
+import net.minecraft.item.ItemConvertible;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.PacketByteBuf;
+
+import net.fabricmc.fabric.impl.lookup.item.ItemKeyImpl;
+
+/**
+ * The immutable combination of an item and additional NBT data. Compare using {@link ItemKey#equals}.
```
LemmaEOF, line 34: Note in javadocs that it differs from an item stack in that it doesn't have count? Also possibly state that it's for filtering

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageFunction.java
```diff
@@ -0,0 +1,47 @@
...
+package net.fabricmc.fabric.api.transfer.v1.storage;
+
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+import net.fabricmc.fabric.impl.transfer.TransferApiImpl;
+
+/**
+ * A function that can be used to operate on a resource, for example to insert or extract, depending on the context.
+ */
+public interface StorageFunction<T> {
+	long apply(T resource, long maxAmount, Transaction tx);
```
LemmaEOF: I was a lot less happy when this when it had a numerator and denominator because nothing in minecraft modding reasonably needs fraction storage aside from fluids, but making it a single long should work perfectly fine.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+import net.fabricmc.fabric.impl.transfer.TransferApiImpl;
+
+/**
+ * An object that can store resources.
+ * <p><ul>
+ *     <li>See {@link CombinedStorage} for a wrapper around multiple {@code Storage}s.</li>
+ * </ul></p>
+ * @param <T> The type of the stored resources.
+ */
+public interface Storage<T> {
```
LemmaEOF, line 35: This doesn't seem to have any way to get individual entries in a Storage for purposes of testing. That's because this is only meant as an exposure for inter-object interactions yeah?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.storage;
+
+public interface StorageView<T> {
```
LemmaEOF, line 27: Probably needs javadocs

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/Transaction.java
```diff
@@ -0,0 +1,94 @@
...
+ * As such, it is better to defer irreversible success until {@link Participant#onFinalCommit} is called,
+ * which will only be called on success for the root transaction.
+ *
+ * <p>Only one outermost transaction can be open at any given time.
+ * Attempts to open a new outer transaction from the thread of the active transaction will throw a RuntimeException.
+ * Attempts to open a new outer transaction from another thread will block until the active transaction is closed.
+ * The server thread is scheduled before the other threads when possible.
+ * Still, try to keep the transaction work on the other threads minimal.
+ */
+public interface Transaction extends AutoCloseable {
```
LemmaEOF, line 49: :iapprove:

LemmaEOF (approved):
This is looking really good! I've got a couple of things, but nothing I'd really consider merge-blocking. My biggest issue is with the name "api lookup api" - see the comment on it below

fabric-api-lookup-api-v1/README.md
Technici4n: You may want to comment on #1234 about that, as it's nearing merge. This PR includes the commits of #1234 as well, which is why it appears in the diff... :/

LemmaEOF: ahhhh, okay, I'll send it there

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
Technici4n: Definitely needs more javadoc, I agree. This is both for filtering and direct storage in the item transfer api.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageFunction.java
Technici4n: As long as we can agree that 81000 is good enough, we do indeed not need fractions. I have not yet decided if `StorageFunction` should be kept, or replaced by `insert` and `extract` functions in `Storage`.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
Technici4n: Added, hope that makes it clearer.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
Technici4n: I'm not sure exactly what you mean by testing in this context.

CoolMineman (changes requested):
🚀

CoolMineman (approved):

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemApiLookupRegistry.java
```diff
@@ -0,0 +1,35 @@
...
+
+package net.fabricmc.fabric.api.lookup.v1.item;
+
+import java.util.Objects;
+
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.lookup.item.ItemApiLookupRegistryImpl;
+
+public class ItemApiLookupRegistry {
```
i509VCB: Final

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public interface ItemKey {
```
i509VCB, line 39: Is this an interface only the api should be overriding? I'd probably put a `NonOverridable` annotation here if other mods are not meant to implement this.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyCache.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.item.Item;
+import net.minecraft.nbt.CompoundTag;
+
+import net.fabricmc.fabric.api.lookup.v1.item.ItemKey;
+
+class ItemKeyCache {
+	static ItemKey get(Item item, @Nullable CompoundTag tag) {
```
i509VCB, line 27: Hmm how do you plan to cache this?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorageFunction.java
```diff
@@ -0,0 +1,55 @@
...
+
+package net.fabricmc.fabric.api.transfer.v1.base;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import net.fabricmc.fabric.api.transfer.v1.storage.StorageFunction;
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+
+// TODO: check >= 0
```
i509VCB: The backing list could change? So caching there wouldn't be bullet proof

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/FilteredStorageFunction.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.base;
+
+import java.util.function.Predicate;
+
+import net.fabricmc.fabric.api.transfer.v1.storage.StorageFunction;
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+
+public final class FilteredStorageFunction<T> implements StorageFunction<T> {
```
i509VCB: I don't think this should be final as a base implementation?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,63 @@
...
+	 * @throws RuntimeException If there aren't enough items to replace, that is if {@link ContainerItemContext#getCount this.getCount()} < count.
+	 */
+	// TODO: consider using an enum instead of a boolean? what about TransactionResult?
+	boolean transform(long count, ItemKey into, Transaction transaction);
+
+	static ContainerItemContext ofPlayerHand(PlayerEntity player, Hand hand) {
+		return PlayerEntityContainerItemContext.ofHand(player, hand);
+	}
+
+	static ContainerItemContext ofPlayerCursor(PlayerEntity player) {
```
i509VCB, line 56: I'd elaborate on this method as cursor could be quite vague.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidPreconditions.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.fluid;
+
+import com.google.common.base.Preconditions;
+
+import net.minecraft.fluid.Fluid;
+import net.minecraft.fluid.Fluids;
+
+public class FluidPreconditions {
```
i509VCB, line 27: Final and private ctor

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidPreconditions.java
```diff
@@ -0,0 +1,35 @@
...
+	public static void notEmpty(Fluid fluid) {
```
i509VCB, line 28: Add the `throws` declaration on the method.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryWrappers.java
```diff
@@ -0,0 +1,51 @@
...
+import net.minecraft.entity.player.PlayerInventory;
+import net.minecraft.inventory.Inventory;
+import net.minecraft.util.math.Direction;
+
+import net.fabricmc.fabric.api.lookup.v1.item.ItemKey;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+import net.fabricmc.fabric.impl.transfer.item.InventoryWrappersImpl;
+
+public class InventoryWrappers {
+	// List<Storage<ItemKey>> has 7 values.
```
i509VCB: Final

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryWrappers.java
```diff
@@ -0,0 +1,51 @@
...
+	private static final WeakHashMap<Inventory, List<Storage<ItemKey>>> WRAPPERS = new WeakHashMap<>();
+
+	public static Storage<ItemKey> of(Inventory inventory, @Nullable Direction direction) {
+		Objects.requireNonNull(inventory, "Null inventory is not supported.");
+		List<Storage<ItemKey>> storages = WRAPPERS.computeIfAbsent(inventory, InventoryWrappersImpl::ofInventory);
+
+		return direction != null ? storages.get(direction.ordinal()) : storages.get(6);
+	}
+
+	public static PlayerInventoryWrapper ofPlayerInventory(PlayerInventory playerInventory) {
```
i509VCB, line 67: Might be useful to have an overload that has a `PlayerEntity` parameter.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemApi.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.math.Direction;
+
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookupRegistry;
+import net.fabricmc.fabric.api.lookup.v1.item.ItemKey;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+
+public final class ItemApi {
+	public static final BlockApiLookup<Storage<ItemKey>, Direction> SIDED =
```
i509VCB, line 33: The field should also specify the access target? The field mentioning that it's a block lookup.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemPreconditions.java
```diff
@@ -0,0 +1,43 @@
...
+package net.fabricmc.fabric.api.transfer.v1.item;
+
+import com.google.common.base.Preconditions;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
+
+import net.fabricmc.fabric.api.lookup.v1.item.ItemKey;
+
+public class ItemPreconditions {
```
i509VCB: Final, suppress default ctor

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/PlayerInventoryWrapper.java
```diff
@@ -0,0 +1,30 @@
...
+
+import net.fabricmc.fabric.api.lookup.v1.item.ItemKey;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+import net.fabricmc.fabric.api.transfer.v1.storage.StorageFunction;
+
+/**
+ * A wrapper around a PlayerInventory.
+ */
+public interface PlayerInventoryWrapper extends Storage<ItemKey> {
+	StorageFunction<ItemKey> offerOrDropFunction();
```
i509VCB: This should probably get javadoc, unsure if yarn has javadoc on this.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Movement.java
```diff
@@ -0,0 +1,71 @@
...
+package net.fabricmc.fabric.api.transfer.v1.storage;
+
+import java.util.function.Predicate;
+
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+
+/**
+ * Helper functions to move resources between two {@link Storage}s.
+ */
+public class Movement {
```
i509VCB: Final

i509VCB: And javadoc for each method

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -0,0 +1,83 @@
...
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookupRegistry;
+import net.fabricmc.fabric.api.lookup.v1.item.ItemApiLookupRegistry;
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+import net.fabricmc.fabric.impl.transfer.TransferApiImpl;
+
+/**
+ * An object that can store resources.
+ * @param <T> The type of the stored resources.
+ */
+public interface Storage<T> {
```
i509VCB, line 35: I know one intended method of using the lookup api is allowing people to implement these interfaces onto their block entities. One issue I see is that you can't implement multiple types of storages because of type erasure. Would it make sense to have a specific storage interface for each resource so this implementation strategy is still possible.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageFunction.java
```diff
@@ -0,0 +1,47 @@
...
+	@SuppressWarnings("unchecked")
+	static <T> StorageFunction<T> empty() {
+		return TransferApiImpl.EMPTY;
+	}
+
+	/**
+	 * Return an identity storage function, i.e. one that always returns the passed amount.
+	 */
+	@SuppressWarnings("unchecked")
+	static <T> StorageFunction<T> identity() {
```
i509VCB: Could you detail an example where identity could be used

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.transaction;
+
+public enum TransactionResult {
+	ABORTED,
+	COMMITTED;
+
+	public boolean wasAborted() {
```
i509VCB, line 23: was -> is

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventoryWrappersImpl.java
```diff
@@ -0,0 +1,458 @@
...
+		public void onClose(ItemStack state, TransactionResult result) {
+			if (result.wasAborted()) {
+				inventory.setStack(slot, state);
+			}
+		}
+
+		@Override
+		public void onFinalCommit() {
+			// TODO: is this necessary for player inventories? in what circumstances?
+			/*if (inventory instanceof PlayerInventory) {
```
i509VCB: Do any of the set stack methods in player inventory do any packet logic? That would probably determine this answer.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionImpl.java
```diff
@@ -0,0 +1,173 @@
...
+		STACK.add(new TransactionImpl());
+	}
+
+	@SuppressWarnings("rawtypes")
+	private final IdentityHashMap<Participant, Object> stateStorage = new IdentityHashMap<>();
+	private boolean isOpen = false;
+
+	public static void setServerThread(Thread serverThread) {
+		if (innerLock.isLocked()) {
+			throw new AssertionError("Something is terribly wrong.");
```
i509VCB: This is not very clear?

Azzyypaaras:
WAIT HOLY SHIT, IT IS FINALLY HAPPENING? By fuck I mean, this was coming eventually but it just feels... surreal for us to be this close... If this gets through it could mark a new era of fabric modding, we are on _verge_ of greatness. I have been harsh on my views of the fabric community and the nature of the fapi before but this... it brings me hope, there is something just... simply... there is something that makes me genuinely happy about the fact that, in spite of our differences, of how long we've been in strife over this, in spite of all the issues that plague the fabric community at times... we've still come _this_ far.
Cheers people, a new, woven dawn approaches!

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionImpl.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		TransactionImpl.serverThread = serverThread;
+	}
+
+	// validate that Transaction instance methods are valid to call.
+	private void validateCurrent() {
+		if (!innerLock.isHeldByCurrentThread()) {
+			throw new IllegalStateException("Transaction operations are not allowed on this thread.");
```
i509VCB: Maybe print the name of the thread you should be dispatching on?

i509VCB (changes requested):

Technici4n:
I have decided to remove `StorageFunction`s entirely. They are a bit surprising to use (why store a `private final` function and return it in a method) and they are not strictly necessary anymore now that we are using integers only. `isEmpty` is replaced by `supportsInsertion` and `supportsExtraction` in `Storage`. I will soon add `InsertOnlyStorage` and `ExtractOnlyStorage` interfaces to provide default implementations for extraction and insertion respectively.

PS: I know the description desperately needs an update, I will do it eventually.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/CombinedStorageFunction.java
Technici4n: Yeah I figured. Outdated, and the new `CombinedStorage` simply assumes that it's non-trivial. Implementations are free to override it.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/Transaction.java
Technici4n: Resolving because this PR thread is too big already, and this is just the beginning. 😅

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/base/FilteredStorageFunction.java
Technici4n: Removed due to `StorageFunction` removal.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
Ueaj-Kerman: it's a result, `was` makes sense

i509VCB: Past tense isn't really going to make sense. The transaction is canceled.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: This whole class really needs a solid explanation, both in javadoc and in the upcoming README.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidPreconditions.java
Technici4n: We probably shouldn't. `Preconditions` doesn't use it, and I just found this reference: https://stackoverflow.com/questions/824217/should-methods-that-throw-runtimeexception-indicate-it-in-method-signature.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemApi.java
Technici4n: I'm not sure I understand what you mean. Javadoc will be added for the field in the future though.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
liach: Imo we should just make it a concrete pojo like what's suitable for a modern java record. I personally don't see point of making a separate interface, given method bodies of classes in the api are considered implementation details.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/PlayerInventoryWrapper.java
Technici4n: In a recent commit, I have added a TODO for this: we need to figure exactly which parts and functions of the PlayerInventory need to be wrapped. Resolving because this is outdated.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyCache.java
liach: As long as we can ensure the tag is not modified elsewhere (much like handling java arrays) it won't be too hard to cache by hashcode. Or if we don't want to cache tag we can just cache if the tag is null, which would make the logic much easier.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
liach: Imo we should just move it to a boolean. If this enum will have more entries this defeats the point of using an enum vs a pseudoenum that allows extension

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageFunction.java
Technici4n: I used it for a creative tank, but this was removed anyway.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
Technici4n: This is indeed a drawback of this approach. I do however believe that having to create inner classes in such cases is better than duplicating the entire api, which would also compromise its genericity, in the sense that it can transfer any game resource, and not just fluids or items.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
Technici4n: When it is used, the transaction is already closed. Therefore I think it makes more sense to use a past tense.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventoryWrappersImpl.java
Technici4n: I think `offerOrDrop` does it. But frankly I am not satisfied of `PlayerInventoryWrapper` and friends, that still needs to be discussed.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
liach: Past tense for transaction doesn't mean past tense for the result. The result ***is*** always there since it ***has been*** or ***was*** derived.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyCache.java
Technici4n: I plan on storing tag-less keys directly in a field in the `Item`, and using a `LoadingCache` for keys with a tag (à la https://github.com/grondag/fluidity/blob/1.16/src/main/java/grondag/fluidity/impl/article/ArticleCache.java).

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
```diff
@@ -31,7 +32,10 @@
 
 /**
  * The immutable combination of an item and additional NBT data. Compare using {@link ItemKey#equals}.
+ *
+ * <p>Do not implement.
  */
+@ApiStatus.NonExtendable
 public interface ItemKey {
 	ItemKey EMPTY = ItemKeyImpl.of(Items.AIR, null);
```
liach, line 40: Any reason of keeping a separate ItemKeyImpl? Imo you can just have this as a pojo in api than writing a separate impl

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionImpl.java
Technici4n: I don't think that information is available without subclassing `ReentrantLock`: there is a `getOwner` method, but it's protected.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
i509VCB: I'd agree here with a final class pojo for this.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
Technici4n: I don't think should be a boolean. It prevents errors (e.g. wrongly naming the boolean `successful` instead of `failed` is very hard to troubleshoot) by making the API so much more explicit, and it's really a trivial class. I am against result `boolean`s in general for those reasons.

liach: We usually use true for success and false for failure/voided stuff. So much that

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
Ueaj-Kerman: I think it should be an interface, to allow people to make copy-on-write ItemKeys and whatnot

liach: Item keys are supposed to be immutable and thus thread safe, what the heck are you suggesting @devan-kerman
also if this is an interface, it is hard to implement equals across interfaces

Ueaj-Kerman: What? the ItemKey itself can be immutable, the underlying nbt tag is copied when the ItemKey is initialized, which isn't always necessary, people should be allowed to make their own ItemKey implementations, because they can optimize, like that case. Another one would be if some tool's ItemKey only cares about the durability, instead of storing a CompoundTag, it can store an int (if it doesn't have additional data of course)

liach: Then how does the equals logic between different item keys work? It doesn't, and the item key interface has no point of existence. just make it a concrete class

Ueaj-Kerman: ??? what, just compare the compound tag and item in the equals method. Our special ItemKey would just test if the other ItemKey had some specific data in it by grabbing its tag. Or, lazily initializing its own tag and asking the other itemkey to compare them

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/transaction/TransactionResult.java
Technici4n: This is based on my experience, when I was implementing the API I didn't remember if the `boolean` was for success or failure. A simple `result.wasAborted()` is so much more expressive. We are talking about a 30 LOC helper enum... I really think booleans should only be used for conditions, and not as flags...

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
Technici4n: I don't think this should be a pojo. The `ItemKey`s will be cached to improve performance as much as possible, and the hash code is cached too. All those impl details should not be exposed to the API user. We can probably move all the default implementations to the impl though, as they clutter the API class a lot for no reason.

As for custom implementations of an item key... Item keys can be passed to other threads or stored directly in any container without further notice, and introducing any way to mutate them is the worst possible thing we could do besides using `ItemStack` directly...

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventoryWrappersImpl.java
Technici4n: I don't think it's necessary in the end, the need stems from a flawed Mojang implementation as far as I can tell.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
liach: > just compare the compound tag and item in the equals method

@devan-kerman you realize this just defeats your whole point of tool not having to keep a full stack for performance purposes?
In the end the nbt isn't too big a hurdle. Your "save a tag" is pretty worthless compared to the problems and risks it introduces. if tags are that expensive why does mojang add a nbt to every tool item?

liach: > I don't think this should be a pojo

Being a pojo doesn't prevent it from being cached. You will just hide the constructor to be package private and users can only call a static method to obtain instances.

Also since there is mutability concern, I suggest making ItemKey a public final class to ensure safety.

Technici4n:
I have updated the description of the PR with the changes that happened today.

We really need #1234 merged ASAP now that this PR needs reviews.
Speaking of reviews, although it is far from mergeable, I am pretty happy with how this PR was received, and how fast it's converging to an excellent transfer API imo. By all means let's keep discussing and improving it!

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemKey.java
Technici4n: I am not sure I understand the value of making this a pojo? I fail to see a direct benefit, and I would very much prefer the ugly implementation details to be kept outside of public API classes.

Ueaj-Kerman: > you realize this just defeats your whole point of tool not having to keep a full stack for performance purposes?

no? copying 2 tags != copying 1 tag, and u could also short cut in some cases too.

> if tags are that expensive why does mojang add a nbt to every tool item?

they also made TACS, item performance can actually be a pretty big problem

Technici4n: I don't think Mojang's decisions are made with modded item transfer in mind...

Anyway, I don't think there is a viable alternative to the `ItemKey` concept as it currently stands. You can't reasonably have immutable item + NBT, and allow mutating the NBT at the same time. Do however note that a lot of modded containers or pipes never touch the NBT directly. In some cases this may cause a tiny loss of performance, but I expect chest performance to be much better with `ItemKey`s because the stacks don't have to be copied... Splitting an expensive NBT stack of 2 items into 2 chests does not involve any allocation with item keys.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyImpl.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Override
+	public CompoundTag toTag() {
+		CompoundTag result = new CompoundTag();
+		result.putString("item", Registry.ITEM.getId(item).toString());
+
+		if (tag != null) {
+			result.put("tag", tag.copy());
+		}
```
OroArmor, line 69: why not 
```suggestion		
		result.put("tag", copyTag());		
```
?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.fluid;
+
+public final class FluidConstants {
+	public static final long BUCKET = 81000;
+	public static final long BOTTLE = 27000;
+	public static final long INGOT = 9000;
+	public static final long NUGGET = 1000;
```
OroArmor, line 23: I think it would be better to define these off of each other, that way if we ever change the amount of one, it does it for all
```suggestion
	public static final long BUCKET = 81000;
	public static final long BOTTLE = BUCKET / 3;
	public static final long INGOT = BUCKET / 9;
	public static final long NUGGET = INGOT / 9;
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * An object that can store resources.
+ *
+ * <p><b>Important note:</b> Unless otherwise specified, all transfer functions take a non-empty resource
+ * and a non-negative maximum amount as parameters.
+ * Implementations are encouraged to throw an exception if these preconditions are violated.
+ *
+ * <p>For transfer functions, the returned amount must be non-negative, and smaller than the passed maximum amount.
+ * Consumers of these functions are encourage to throw an exception if these postconditions are violated.
+ * @param <T> The type of the stored resources.
```
OroArmor, line 33: I would give some examples of what the stored resources can be. should I store `Item`s or `ItemStack`s?

Ueaj-Kerman: `ItemKey`

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
ramidzkh: ~~*Also make 81000 not a compile constant (`long BUCKET = Objects.requireNonNull(81000);`) so I can break it with a mod*~~

Technici4n: These constants must never be changed in the future, for obvious reasons... If they need to be updated, we will need to deprecate the older classes, and create new ones instead.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
Technici4n: It should be obvious when you see how it's defined in ItemApi, although I agree that this needs more example code.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
ramidzkh: Defining them like how @OroArmor suggested is easier to understand imo. And if required, why not change these values within snapshot cycles instead of deprecating the class? Since they're all compile time constants, deleting the class wouldn't break old mods anyway

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyImpl.java
Technici4n: That's better considering a stack with no tag and a stack with null tag are different.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
Technici4n: No, there is nothing worse than silently changing values, be it during snapshot cycles or not. You can't possibly know that your code still compiles but is now broken after a FAPI update.

I prefer using the direct values, as they are more explicit. If the API user can't understand why an ingot is 9000 droplets after I add javadoc I don't think we can expect them to understand the rest of the API anyway.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/item/ItemKeyImpl.java
Technici4n: Wait no that's precisely why I didn't do it.

Technici4n: We don't need to waste storage space on an empty tag.

OroArmor: Yeah I just realized it

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
```diff
@@ -0,0 +1,82 @@
...
+	public static void registerEmptyItem(Item emptyItem, Item fullItem, Fluid fluid, long amount) {
+		ItemPreconditions.notEmpty(emptyItem);
+		ItemPreconditions.notEmpty(fullItem);
+		FluidPreconditions.notEmpty(fluid);
+		Preconditions.checkArgument(amount > 0);
+
+		EmptyItemsRegistry.registerEmptyItem(emptyItem, fullItem, fluid, amount);
+	}
+
+	// TODO: potion handling
```
williambl, line 115: Does this imply it's in-scope to treat Potions similarly to Fluids, like LBA does?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
```diff
@@ -0,0 +1,82 @@
...
+
+		ITEM.register((key, ctx) -> new SimpleFluidContainingItem(ctx, key, emptyItem, fluid, amount), fullItem);
+	}
+
+	/**
+	 * Register an item that is empty, and may be filled with some fluid entirely.
+	 */
+	// TODO: document params and conflicts
+	// TODO: pick parameter order, probably the same for both methods?
+	public static void registerEmptyItem(Item emptyItem, Item fullItem, Fluid fluid, long amount) {
```
williambl: Would it be a good idea to have `registerItems(...) { registerFullItem(...); registerEmptyItem(...); }`?

They would seemingly be called together most of the time, so this might be a little more convenient for modders.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Movement.java
```diff
@@ -0,0 +1,63 @@
...
+	public static <T> long move(Storage<T> from, Storage<T> to, Predicate<T> filter, long maxAmount) {
+		try (Transaction moveTransaction = Transaction.openOuter()) {
+			long result = move(from, to, filter, maxAmount, moveTransaction);
+			moveTransaction.commit();
+			return result;
+		}
+	}
+
+	public static <T> long move(Storage<T> from, Storage<T> to, Predicate<T> filter, long maxAmount, Transaction transaction) {
+		long[] totalMoved = new long[] { 0 };
```
williambl, line 36: Just out of interest, is there a name for the pattern of using a 1-length array as if it were a pointer or reference like this? I've never seen it before and I think it's quite clever.

williambl (commented):
Feels like an elegant API. I'd be happy to use this. (Not going to mark this as an approval just because the PR's not complete yet).

I used to be in favour of dynamic-denominator fractions (mostly just because LBA uses it and LBA's really nice), but honestly I think 81000-denom works fine for pretty much all purposes. (As long as nobody exposes it to the player of course)

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
Technici4n: Not exactly. The plan is to allow any mod to define a mapping `Potion` -> `(Fluid, long)` so that the `Storage<Fluid>` for a full potion can correctly empty potions with a registered fluid mapping. In particular, this would only be used by the API for water bottles which are actually `Potions.Water`, but other mods may use it to register a `Fluid` per `Potion` if they want to.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
Technici4n: Yes this is absolutely planned, it's just that the API for this is very much WIP at the moment. I would also like to make sure that people don't swap emptyItem and fullItem by mistake if that's possible somehow.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Movement.java
Ueaj-Kerman: `out parameters at home` :tiny_potato:

Technici4n: No idea. I thought it was very common.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
williambl: Ah, I see. Sounds useful.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidApi.java
williambl: Makes sense. Maybe making both methods be `(Fluid fluid, Item emptyItem, Item fullItem, long amount)`? Not having an `Item` at the start would make it clear that the first param is the same for both.

Technici4n: I have updated this with a new `registerEmptyAndFullItems`. The name is a bit verbose, but I think `registerItems` may be a bit unclear.

Now that there are mapping functions as well, I think the target item at the end makes the most sense.

williambl: Looks good. (I'm not sure how to mark this conversation as 'resolved')

Technici4n: image
I have this "resolve" button, don't you have it? 🤔

Juuxel: IIRC only the PR author (and maybe repo collaborators) can see it.

Technici4n: :githubthonk:

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public interface Storage<T> {
+	/**
+	 * Return whether this storage supports insertion, meaning that {@link #insert} will not always return 0.
```
Prospector, line 37: Could this be clearly defined as to when you should return false? Should it depend on state (does this support insertion right now?) or not (does this *ever* support insertion?)? For example, if a tank is at its capacity and cannot currently accept any insertions, should this be false? What is this intended to be used for? Would a pipe choose not to connect to a Storage that returns false here?

Technici4n: It never depends on the state, that certainly needs better javadoc I agree.

Technici4n:
Replaced by #1352, #1356 and subsequent PRs.

PR #1280, opened 2021-01-24
Upload to modrith from gradle

modmuss50:
Also adds a preRelase global var that will be sent for snapshots, in the past I was just changing curseforge to beta.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ plugins {
 	id "net.minecrell.licenser" version "0.4.1"
 	id "org.ajoberstar.grgit" version "3.1.1"
 	id "com.matthewprenger.cursegradle" version "1.4.0"
+	id "com.modrinth.minotaur" version "1.1.0"
 }
 
 def ENV = System.getenv()
@@ -22,6 +23,7 @@ class Globals {
 	static def mcVersion = "1.16.5"
 	static def yarnVersion = "+build.1"
 	static def loaderVersion = "0.10.5+build.213"
+	static def preRelease = false
 }
 
 version = Globals.baseVersion + "+" + (ENV.GITHUB_RUN_NUMBER ? "" : "local-") + getBranch()
@@ -341,7 +343,7 @@ curseforge {
 	project {
 		id = "306612"
 		changelog = "A changelog can be found at https://github.com/FabricMC/fabric/commits"
-		releaseType = "release"
+		releaseType = Globals.preRelease ? "beta" : "release"
 		addGameVersion "1.16.5"
 		addGameVersion "Fabric"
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -375,12 +377,30 @@ task github(dependsOn: remapMavenJar) {
...
 		releaseBuilder.body("A changelog can be found at https://github.com/FabricMC/fabric/commits")
 		releaseBuilder.commitish(getBranch())
+		releaseBuilder.prerelease(Globals.preRelease)
 
 		def ghRelease = releaseBuilder.create()
 		ghRelease.uploadAsset(file("${project.buildDir}/libs/${archivesBaseName}-${version}.jar"), "application/java-archive");
 	}
 }
 
+task modrinth(type: com.modrinth.minotaur.TaskModrinthUpload, dependsOn: remapMavenJar) {
```
i509VCB, line 387: I would hook this up to the publish task in all honesty. Something like `tasks.publish.dependsOn(modrinth)` and in general for curse and github releases

build.gradle
```diff
@@ -375,12 +377,30 @@ task github(dependsOn: remapMavenJar) {
...
+	onlyIf {
+		ENV.MODRINTH_TOKEN
+	}
+
+	token = ENV.MODRINTH_TOKEN
```
i509VCB, line 392: This will be fine for now, I do need to go make minotaur use properties instead of public fields - it shouldn't break if I believe when I do that.

liach: What's wrong with this using env?

i509VCB: There isn't an issue with the token in the ENV - but rather it's a nasty bit of the gradle plugin and how it exposes it's api to consumers. Not a blocking issue but I plan to neaten up the plugin upstream.

Not related to FAPI here - but Minotaur stuff inside of the plugin.

Typically you will want to use a getter + setter with an `@Input` annotation in gradle, minotaur uses public mutable fields atm which is a bit of code smell. In groovy the `field = value` is essentially adapted into a `get/set` like how kotlin handles var types.

liach: if you update the plugin dependency gradle will recompile the buildscript, so this is a non-issue on fabric's end

i509VCB: Yes - and I believe I can do the upstream changes fine in the future without the buildscript breaking

liach (approved):
Why not approve this then

i509VCB:
I am mainly waiting for a response on the publish task depending on modrinth publish before I set an approval

OroArmor (approved):

YTG1234 (approved):

liach:
> 
> 
> I am mainly waiting for a response on the publish task depending on modrinth publish before I set an approval

I personally disagree on making publish depend on curseforge or modrinth. Those are not maven publication targets.

PR #1282, opened 2021-01-24
Document custom values

i509VCB:
This documents the custom values Fabric API provides or consumes.

### DIFF
--- a/docs/CUSTOM-VALUES.md
+++ b/docs/CUSTOM-VALUES.md
@@ -35,26 +35,4 @@ All Fabric API modules
 
 Custom values consumed by Fabric API are specified by other mods.
 These may be used to enable/disable features in fabric api.
-
-## Fabric Renderer Indigo
-
-### `fabric-renderer-api-v1:contains_renderer`
-
-If a mod contains this custom value, the default indigo renderer will be disabled.
-This is intended for mods which implement the rendering api.
-
-**Type:** Any
-
-**Presence:**
-A mod which implements the rendering api.
-
-### `fabric-renderer-indigo:force_compatibility`
-
-If a mod contains this value AND no mod has set `fabric-renderer-api-v1:contains_renderer`, the Indigo renderer implementation will use a compatability mode.
-
-This will disable vanilla block tesselation and ensures vertex format compatibility.
-
-**Type:** Any
-
-**Presence:**
-A mod which requires vanilla vertex format compatability.
+These custom values are documented in each module which consumes these types of custom values.
--- a/docs/CUSTOM-VALUES.md
+++ b/docs/CUSTOM-VALUES.md
@@ -11,7 +11,8 @@ Custom values provided by Fabric API are for use by other mods.
 ## `fabric-api:module-lifecycle`
 
 Specifies the lifecycle of a fabric api module.
-This may be used by other mods to determine 
+This may be used by other mods to determine the current lifecycle of a fabric api module, such as if a module is experimental
+and may be subject to unannounced changes.
 
 **Type:** Json String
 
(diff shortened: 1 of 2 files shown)


Technici4n (approved):

liach:
Should this be moved to individual apis?

i509VCB:
The custom values for indigo, possibly, but I feel that location for the document is more accessible since very few modules consume custom values but all modules provide the module lifecycle.

liach:
The module lifecycle definitely stays here. However module specifics I suggest moving to each module.
In addition, I think this file should be merged with files describing entrypoints offered by each module.

i509VCB:
> describing entrypoints offered by each module

I assume you mean the readme like in the lifecycle events v1 module?

liach:
More specifically, I mean the custom values consumed by the fabric api mods should be placed together with the entrypoints consumed by fabric api mods.

For example, if a `fabric-blah-api-v1` mod consumes both `fabric-blah-api-v1:callback_entry` entrypoint and `fabric-blah-api-v1:custom_logic` values from mod jsons, these had better be documented together because they are both from `fabric.mod.json` of the mods using blah api.

i509VCB:
In your example, would the consumed custom value entries become part of the module readme under a `custom values` category or would that be a secondary markdown file within the module.

Similar question to entrypoints - a category in markdown or it's own file in the module.

liach:
Possibly a separate usage file. The readme imo should be more player oriented. Something like that famous tool attribute api from some youtuber cheating scandal definitely need a player-oriented readme and a separate modder oriented usage.

i509VCB:
I don't see the purpose behind a user readme within each module folder - typically users would be reading the readme at the root of the repository. That readme mentions the user stuff before any mod developer things.

liach:
Then feel free. imo the main thing distinguishing "category in markdown" versus "own file" is the size of the markdown. TLDR!

docs/CUSTOM-VALUES.md
```diff
@@ -0,0 +1,59 @@
...
+The use of custom values ranges from enabling/disabling features to extra metadata about fabric api modules.
+
+# Provided by Fabric API
+
+Custom values provided by Fabric API are for use by other mods.
+
+## `fabric-api:module-lifecycle`
+
+Specifies the lifecycle of a fabric api module.
+This may be used by other mods to determine
```
Daomephsta: Determine what? Seems to be some text missing here.

i509VCB: Okay now I have determined the answer

i509VCB:
Okay @liach I have moved the indigo custom values into indigo (and wrote a very rudimentary readme for it)

liach (approved):

docs/CUSTOM-VALUES.md
```diff
@@ -0,0 +1,38 @@
+# Custom values
+
+Fabric API modules may have certain custom values set or check for the presence of specific custom values.
+
+The use of custom values ranges from enabling/disabling features to extra metadata about fabric api modules.
+
+# Provided by Fabric API
```
Juuxel: Shouldn't this technically be a `##` since `#` is for `# Custom values`? (just a small nitpick)

Juuxel (approved):

LambdAurora (approved):
(and I'm backing up @Juuxel's suggestion)

Mysterious-Dev (approved):

docs/CUSTOM-VALUES.md
haykam821: In that case, the following header should be a level deeper as well.

modmuss50 (approved):

PR #1285, opened 2021-01-25
Also init screen events in the constructor

modmuss50:
Fixes crashes with screens that do not call super.init such as REI here: https://github.com/shedaniel/RoughlyEnoughItems/blob/39ed7b648c853e19fc448f297aa9716ccfa7d60a/RoughlyEnoughItems-runtime/src/main/java/me/shedaniel/rei/gui/DelegateScreen.java#L261

The events may not fire, or behave correctly in this case, but its better than crashing.

### DIFF
--- a/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
+++ b/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
@@ -30,6 +30,7 @@ import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gui.Element;
 import net.minecraft.client.gui.screen.Screen;
 import net.minecraft.client.gui.widget.AbstractButtonWidget;
+import net.minecraft.text.Text;
 
 import net.fabricmc.fabric.api.client.screen.v1.ScreenEvents;
 import net.fabricmc.fabric.impl.client.screen.ScreenExtensions;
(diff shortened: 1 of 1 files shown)


LambdAurora (approved):

fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
```diff
@@ -95,9 +96,26 @@
...
 	private void beforeInitScreen(MinecraftClient client, int width, int height, CallbackInfo ci) {
 		// All elements are repopulated on the screen, so we need to reinitialize all events
+		initEvents();
+
+		ScreenEvents.BEFORE_INIT.invoker().beforeInit(client, (Screen) (Object) this, width, height);
+	}
+
+	@Inject(method = "init(Lnet/minecraft/client/MinecraftClient;II)V", at = @At("TAIL"))
+	private void afterInitScreen(MinecraftClient client, int width, int height, CallbackInfo ci) {
+		ScreenEvents.AFTER_INIT.invoker().afterInit(client, (Screen) (Object) this, width, height);
```
i509VCB, line 114: This will never be called if an override of that method never calls super.

i509VCB (changes requested):

modmuss50:
The real fix is on the mods side of things, having unprodicable behaviour isnt great so after some throught its prob not a good idea to go with this. We might need better logging around screens that do not initialise correctly. Hopefuly only a very small subset of mods are affected. If its a greater number then we might need to think again.

modmuss50:
Closing, Hard crash seems to be the real soltuion here. Needs another PR to improve the crash dump tho

PR #1291, opened 2021-01-29
Remove all inter-module dependencies on fabric-networking-v0

i509VCB:
This migrates all uses of v0 networking to v1 api, and changes the readme to not mention networking-v0 in the example for dependencies.

This effectively makes v0 removable if it breaks in the future beyond repair. Though I expect v0 to last quite a bit longer.

This also fixes an issue where the new screen handler api does not release the packet byte buf after opening the screen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Set<String> apiModules = [
     "fabric-api-base",
     "fabric-command-api-v1",
     "fabric-lifecycle-events-v1",
-    "fabric-networking-v0"
+    "fabric-networking-api-v1"
 ]
 
 // Add each module as a dependency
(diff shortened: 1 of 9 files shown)


OroArmor (approved):
lgtm

liach (approved):
Hmm, I recall more modules used v0 before. Maybe the leftover of some incomplete migration

LambdAurora (approved):

Juuxel (approved):

YTG1234 (approved):
I don't see any problems here.

modmuss50 (approved):
Looks fine, can be merged in after some testing. Should be qutie safe as it was using v1 under the hood anyway 👍

PR #1293, opened 2021-01-30
Fix #1220 (Fix issue where removing a dimension removes nether and end)

### LINKED ISSUE #1220: Deleting a custom dimension in a datapack deletes Nether and End
This is a huge problem in modpacks like AOF3 as I can't touch mods like Astromine even tho they're causing gameplay issues as the removal of the mod will also.. Cliffhanger... Remove the Nether and End.

Link to issue:
https://bugs.mojang.com/browse/MC-197860

i509VCB:
This fixes a bug in vanilla code where removing a dimension will also remove every other dimension but the overworld even if the other dimensions are present.

Tracking Mojang issue: https://bugs.mojang.com/browse/MC-197860

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/CompoundTagMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/CompoundTagMixin.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.dimension;
+
+import java.util.Iterator;
+import java.util.Map;
+
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.nbt.Tag;
+
+import net.fabricmc.fabric.impl.dimension.CompoundTagExtensions;
+
+@Mixin(CompoundTag.class)
+abstract class CompoundTagMixin implements CompoundTagExtensions {
+	@Shadow
+	@Final
+	private Map<String, Tag> tags;
+
+	@Override
+	public Iterator<Map.Entry<String, Tag>> fabric_entries() {
+		return this.tags.entrySet().iterator();
+	}
+}
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
@@ -83,8 +83,7 @@ abstract class LevelStorageMixin {
 			return new RuntimeException("Failed to get the dimension registry");
 		}).getFirst();
 
-		CompoundTag worldGenSettingsTag = properties.getCompound("WorldGenSettings");
-		CompoundTag dimensions = worldGenSettingsTag.getCompound("dimensions");
+		CompoundTag dimensions = properties.getCompound("WorldGenSettings").getCompound("dimensions");
 
 		for (String key : dimensions.getKeys()) {
 			Identifier identifier;
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.dimension;
+
+import com.mojang.datafixers.DataFixer;
+import com.mojang.datafixers.util.Pair;
+import com.mojang.serialization.DataResult;
+import com.mojang.serialization.Dynamic;
+import com.mojang.serialization.Lifecycle;
+import org.jetbrains.annotations.Nullable;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.InvalidIdentifierException;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryLookupCodec;
+import net.minecraft.world.dimension.DimensionType;
+import net.minecraft.world.gen.GeneratorOptions;
+import net.minecraft.world.level.storage.LevelStorage;
+
+/*
+ * This is a bug fix.
+ * This fixes an issue where removing a mod or datapack that adds a dimension will cause every dimension but the overworld to also be disposed of.
+ *
+ * Testing:
+ * A good way to test this is to create a world with the dimension test mod enabled and then without to simulate a dimension removal.
+ *
+ * Tracking issue: https://bugs.mojang.com/browse/MC-197860
+ */
+@Mixin(LevelStorage.class)
+abstract class LevelStorageMixin {
+	/*
+	 * Why?
+	 * When the custom dimension is removed, DimensionGeneratorSettings.CODEC fails to fully parse the Dynamic<T>,
+	 * so the DataResult<DimensionGeneratorSettings> only return the Either#Right part of the result,
+	 * which only includes the overworld.
+	 *
+	 * This causes all dimensions to cease to exist.
+	 * To fix this issue, we will intercept the dynamic being parsed, and remove any dimension entries that are NOT in the registry.
+	 */
+
+	/**
+	 * @param dynamic the dynamic containing the properties
+	 * @param dataFixer the data fixer
+	 * @param version the data version of the properties being read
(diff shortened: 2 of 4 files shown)


fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,109 @@
...
+		// Fix the dynamic in place if the value is nbt
+		if (dynamic.getValue() instanceof CompoundTag) {
+			properties = (CompoundTag) dynamic.getValue();
+		} else {
+			// Not supported atm
+			throw new IllegalStateException("Dynamic is not backed by a Compound Tag! This is not supported at the moment.");
+		}
+
+		// Get the dimension registry
+		DataResult<Pair<Registry<DimensionType>, T>> result = RegistryLookupCodec.of(Registry.DIMENSION_TYPE_KEY).codec().decode(propertiesDynamic);
```
liach, line 86: Is the `codec` returning an element codec? Just a side note if this should be updated in yarn.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Testing:
+ * A good way to test this is to create a world with the dimension test mod enabled and then without to simulate a dimension removal.
+ *
+ * Tracking issue: https://bugs.mojang.com/browse/MC-197860
+ */
+@Mixin(LevelStorage.class)
+abstract class LevelStorageMixin {
+	/*
+	 * Why?
```
liach:
```suggestion
```

liach (commented):
Is this code a copypasta from vanilla? It needs some improvement in quality

i509VCB:
Most of the code is original, the only thing I had to steal from vanilla was getting the dimension registry

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
i509VCB: This is pretty much absuing a codec to access the registry from the dynamic registry manager in registry ops from what I see.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,108 @@
...
+			} catch (InvalidIdentifierException e) {
+				throw new RuntimeException("Failed to fix generator properties because of invalid dimension identifier", e);
+			}
+
+			// Lookup the dimension
+			@Nullable
+			DimensionType dimension = dimensionRegistry.getOrEmpty(identifier).orElse(null);
+
+			// Dimension is not in registry, remove it from the compound tag so DFU does not freak out
+			if (dimension == null) {
```
liach: Why not just
```suggestion
			// Dimension is not in registry, remove it from the compound tag so DFU does not freak out
			if (!dimensionRegistry.containsId(identifier)) {
```

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
liach: Lol no, this is [a mojang name](https://github.com/Mojang/DataFixerUpper/blob/e843dac1eb96af7b86b1e11165afeca7caffb7e6/src/main/java/com/mojang/serialization/MapCodec.java#L105)! I find confidence in yarn again

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,108 @@
...
+				dimensions.remove(key);
```
liach: Wait a second, will this cause CME? You should use explicit iterator and call `iterator.remove()`

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		// Get the dimension registry
+		DataResult<Pair<Registry<DimensionType>, T>> result = RegistryLookupCodec.of(Registry.DIMENSION_TYPE_KEY).codec().decode(propertiesDynamic);
+		Registry<DimensionType> dimensionRegistry = result.result().orElseThrow(() -> {
+			return new RuntimeException("Failed to get the dimension registry");
+		}).getFirst();
+
+		CompoundTag worldGenSettingsTag = properties.getCompound("WorldGenSettings");
+		CompoundTag dimensions = worldGenSettingsTag.getCompound("dimensions");
```
liach:
```suggestion
		CompoundTag dimensions = properties.getCompound("WorldGenSettings").getCompound("dimensions");
```
Useless local var

liach (changes requested):
I highly doubt if your map key removal code works

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
liach: See https://maven.fabricmc.net/docs/yarn-21w03a+build.23/net/minecraft/util/registry/Registry.html#containsId(net.minecraft.util.Identifier)

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
i509VCB: Not sure if getting the keys from a compound tag returns the key set from the backing map or a copy?

liach: :cold_sweat:
```java
    public Set<String> getKeys() {
        return this.tags.keySet();
    }
```
ConcurrentModificationException noises. Also removing within iterator is way more efficient than copying the iterable and remove on original iterable

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+package net.fabricmc.fabric.mixin.dimension;
+
+import com.mojang.datafixers.DataFixer;
+import com.mojang.datafixers.util.Pair;
+import com.mojang.serialization.DataResult;
+import com.mojang.serialization.Dynamic;
+import com.mojang.serialization.Lifecycle;
+import org.jetbrains.annotations.Nullable;
```
liach:
```suggestion
```
Just in case you didn't do it locally. It killed the build

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
i509VCB: There is no way to get the iterator without mixing into compound tag. This seems to be the only way I see but it is a CME nightmare lmao, so the mixin it is.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/CompoundTagMixin.java
```diff
@@ -0,0 +1,41 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.nbt.Tag;
+
+import net.fabricmc.fabric.impl.dimension.CompoundTagExtensions;
+
+@Mixin(CompoundTag.class)
+abstract class CompoundTagMixin implements CompoundTagExtensions {
```
Chocohead, line 32: Why is this a duck rather than a straight `@Accessor` mixin?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,113 @@
...
+
+			try {
+				identifier = new Identifier(entry.getKey());
+			} catch (InvalidIdentifierException e) {
+				throw new RuntimeException("Failed to fix generator properties because of invalid dimension identifier", e);
+			}
+
+			// Dimension is not in registry, remove it from the compound tag so DFU does not freak out
+			if (!dimensionRegistry.containsId(identifier)) {
+				entryIterator.remove();
```
modmuss50, line 109: This should be logged out.

liach (changes requested):
Now looking at how you make mixins just to remove tags... I feel cursed

I strongly suggest you move keys pending removal to a collection and after your first loop, loop the collection for removal of those keys

50ap5ud5 (commented):
Hello, thanks for making this! 

This is an essential PR I'm waiting on for my dimension mods. 

Have you considered taking a took at how the dimensions are deserialised by Mojang's DataFixerUpper library? From some tests, it appears the UnboundedMapCodec they use in SimpleRegistry, which is used to register the dimensions, appears to the the root cause of the problem by being too aggressive with any parsing failures.

I think your approach is a cool, but I think it could be worth looking at the backend systems being used for json deserialise/serialising.

Also, how do you plan to address worlds which were affected by the bug before this fix is implemented? I.e. Those worlds whose Nether and End dimensions have already been "deleted"?

Looking forward to this getting merged!

RedstoneParadox (approved):

RedstoneParadox:
> Hello, thanks for making this!
> 
> This is an essential PR I'm waiting on for my dimension mods.
> 
> Have you considered taking a took at how the dimensions are deserialised by Mojang's DataFixerUpper library? From some tests, it appears the UnboundedMapCodec they use in SimpleRegistry, which is used to register the dimensions, appears to the the root cause of the problem by being too aggressive with any parsing failures.
> 
> I think your approach is a cool, but I think it could be worth looking at the backend systems being used for json deserialise/serialising.
> 
> Also, how do you plan to address worlds which were affected by the bug before this fix is implemented? I.e. Those worlds whose Nether and End dimensions have already been "deleted"?
> 
> Looking forward to this getting merged!

 I believe that was something that was discussed and dismissed.
 
 Edit: Don't know why this was made a review.

liach:
This pr is still quite problematic imo. It starts with a good intention but starts going way off in implementations, and is still pending an update.

RedstoneParadox:
> This pr is still quite problematic imo. It starts with a good intention but starts going way off in implementations, and is still pending an update.

Are you going to be more specific than that? It's hard to see what you find problematic when it's so small.

i509VCB:
> This pr is still quite problematic imo. It starts with a good intention but starts going way off in implementations, and is still pending an update.

I've just been a bit busy recently. If you want to fix the impl yourself feel tree to pr to my own fork. I'll try to attend to using the accessors today.

modmuss50:
Is this still a big issue? I havent heard of it much recently if so, is it worth another pair of eyes looking at it?

i509VCB:
This issue is still a big one, however I have not really had the time to switch over to the branch and replace that duck interface with an accessor.

sleepy-evelyn:
> Is this still a big issue? I havent heard of it much recently if so, is it worth another pair of eyes looking at it?

Recently had this issue appear about 2 weeks ago on one of our fabric test servers running on 1.16.5. For reference: https://github.com/DaFuqs/StarrySky/issues/18
  
Guess it's an easy enough fix if you know what your doing and keep backups of your worlds but if you don't it's pretty scary when u can't travel between worlds

50ap5ud5:
Hi, are there any plans to fix this issue in the future? I'm curious because this is still a major issue for all dimension mods/datapacks.

PR #1295, opened 2021-01-31
Create a simple renderer api testmod

i509VCB:
This adds a simple testmod to the renderer api to allow some level of testing if the implementation works.

I also used this pull request to figure out where renderer api documentation is lacking and will open an issue to address the issues I have seen in documentation.

### DIFF
--- a/fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/WorldRenderExtensions.java
+++ b/fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/WorldRenderExtensions.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.test.renderer;
+
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+/**
+ * Extension interface for a world to notify the world that a block needs to be re-rendered.
+ */
+public interface WorldRenderExtensions {
+	static void scheduleBlockRerender(World world, BlockPos pos) {
+		((WorldRenderExtensions) world).scheduleBlockRerender(pos);
+	}
+
+	void scheduleBlockRerender(BlockPos pos);
+}
--- a/fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/mixin/ClientWorldMixin.java
+++ b/fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/mixin/ClientWorldMixin.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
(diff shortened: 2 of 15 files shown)


fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/mixin/ClientWorldMixin.java
```diff
@@ -0,0 +1,38 @@
...
+@Mixin(ClientWorld.class)
+abstract class ClientWorldMixin extends WorldMixin {
+	@Shadow
+	@Final
+	private WorldRenderer worldRenderer;
+
+	@Override
+	public void scheduleBlockRerender(BlockPos pos) {
+		// Update the block at the position to trigger chunk re-render.
+		this.worldRenderer.updateBlock(null, pos, null, null, 0);
```
liach, line 36: Make sure the nulls are safe to pass

fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/WorldRenderExtensions.java
```diff
@@ -0,0 +1,31 @@
...
+
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+/**
+ * Extension interface for a world to notify the world that a block needs to be re-rendered.
+ */
+public interface WorldRenderExtensions {
+	static void scheduleBlockRerender(World world, BlockPos pos) {
+		((WorldRenderExtensions) world).scheduleBlockRerender(pos);
```
liach, line 27: Why can't you just make only client worlds implement this interface if you know the world passed in is always going to be a client world (as it is held by a client block entity)?

liach (commented):
Is the block rerender hook worthy to add to the renderer api itself?

i509VCB:
> Is the block rerender hook worthy to add to the renderer api itself?

I would say yes, but not sure if it would go in the renderer api as it is more Minecraft specific than anything. Maybe `fabric-rendering-v1`?

fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/WorldRenderExtensions.java
i509VCB: I typically avoid casting to client world in possibly common codepath. This will make the server world's logic no-op here.

fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/mixin/ClientWorldMixin.java
i509VCB: Nulls are safe here, this just forces a redraw.

fabric-renderer-api-v1/src/testmod/java/net/fabricmc/fabric/test/renderer/simple/client/RendererClientTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.blockrenderlayer.v1.BlockRenderLayerMap;
+import net.fabricmc.fabric.api.client.model.ModelLoadingRegistry;
+import net.fabricmc.fabric.test.renderer.simple.RendererTest;
+
+public final class RendererClientTest implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		ModelLoadingRegistry.INSTANCE.registerResourceProvider(_manager -> new FrameModelResourceProvider());
```
liach: Maybe just name the lambda param `manager`, `_manager` is weird
```suggestion
		ModelLoadingRegistry.INSTANCE.registerResourceProvider(manager -> new FrameModelResourceProvider());
```

liach (approved):
Looks good. My question on moving the world inject thing to api was a digression and doesn't belong to this pr.

modmuss50 (approved):
Seems fine, its a test mod so doesnt need to be perfect. This would have been helpful the other day.

modmuss50:
I dont see a reason not to merge this.

PR #1301, opened 2021-02-02
Fix #1273 (Make MultipartBakedModel and WeightedBackedModel implement FabricBakedModel properly)

### LINKED ISSUE #1273: Renderer API's vanilla adapter doesn't work with all multipart/weighted models
From `BakedModel` implementing `FabricBakedModel`, all vanilla models will return `true` for  `FabricBakedModel#isVanillaAdapter`. For most models this is fine, but for a multipart or weighted model where one of the components would otherwise return `false` this can cause problems from the component completely not supporting being rendered the vanilla way:
<details>
  <summary>Example crash</summary>

In this particular example, this is from trying to render [Ignis Fatuus](https://github.com/CoolMineman/ignis-fatuus)'s [carved pumpkin](https://github.com/CoolMineman/ignis-fatuus/blob/master/src/main/resources/assets/ignis-fatuus/blockstates/carved_pumpkin_block.json), made up of both vanilla and non-vanilla components
```
java.lang.NullPointerException: Tesselating block model
	at java.util.ArrayList.addAll(Unknown Source)
	at net.minecraft.class_1095.method_4707(class_1095.java:67)
	at net.minecraft.class_778.renderModelSmooth(class_778.java:143)
	at net.minecraft.class_778.renderModel(class_778.java:101)
	at net.minecraft.class_776.renderModel(class_776.java:83)
	at net.minecraft.class_776.method_3355(class_776.java:74)
	at net.minecraft.class_846$class_851$class_4578.redirect$zbc000$hookChunkBuildTesselate(class_846.java:715)
	at net.minecraft.class_846$class_851$class_4578.method_22785(class_846.java:891)
	at net.minecraft.class_846$class_851$class_4578.method_22783(class_846.java:718)
	at net.minecraft.class_846.method_22757(class_846.java:182)
	at java.util.concurrent.CompletableFuture.uniComposeStage(Unknown Source)
	at java.util.concurrent.CompletableFuture.thenCompose(Unknown Source)
	at net.minecraft.class_846.method_22763(class_846.java:180)
	at net.minecraft.class_846.method_23086(class_846.java:254)
	at net.minecraft.class_3846.method_16907(class_3846.java:94)
	at net.minecraft.class_3846.method_16900(class_3846.java:137)
	at net.minecraft.class_3846.run(class_3846.java:105)
	at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(Unknown Source)
	at java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(Unknown Source)
	at java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
	at java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)


A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Thread: Render thread
Stacktrace:
	at java.util.ArrayList.addAll(Unknown Source)
	at net.minecraft.class_1095.method_4707(class_1095.java:67)
	at net.minecraft.class_778.renderModelSmooth(class_778.java:143)

-- Block model being tesselated --
Details:
	Block: Block{ignis-fatuus:carved_pumpkin_block}[facing=north,lid=false,torch=false]
	Block location: World: (-201,67,-220), Chunk: (at 7,4,4 in -13,-14; contains blocks -208,0,-224 to -193,255,-209), Region: (-1,-1; contains chunks -32,-32 to -1,-1, blocks -512,0,-512 to -1,255,-1)
	Using AO: true
	Model class: net.minecraft.class_1095
	Is Fabric model: true
	Is adapted vanilla model: true
Stacktrace:
	at net.minecraft.class_778.renderModel(class_778.java:101)

-- Block being tesselated --
Details:
	Block: Block{ignis-fatuus:carved_pumpkin_block}[facing=north,lid=false,torch=false]
	Block location: World: (-201,67,-220), Chunk: (at 7,4,4 in -13,-14; contains blocks -208,0,-224 to -193,255,-209), Region: (-1,-1; contains chunks -32,-32 to -1,-1, blocks -512,0,-512 to -1,255,-1)
	Block render type: MODEL
Stacktrace:
	at net.minecraft.class_776.renderModel(class_776.java:83)
	at net.minecraft.class_776.method_3355(class_776.java:74)
```
</details>

This does seem to have been noticed and handled within JMX ([here](https://github.com/grondag/json-model-extensions/blob/16e056194f387bc6ed5f6efc7f566cbdd21b06e8/src/main/java/grondag/jmx/mixin/MixinMultipartBakedModel.java#L58-L65) and [here](https://github.com/grondag/json-model-extensions/blob/16e056194f387bc6ed5f6efc7f566cbdd21b06e8/src/main/java/grondag/jmx/mixin/MixinWeightedBakedModel.java#L57-L69)), but it might be nice if the rendering API itself would do it to avoid the otherwise quite unclear crashes from this edge case.

CoolMineman:
Makes `MultipartBakedModel`s and `WeightedBakedModel`s use frapi.

### DIFF
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
@@ -55,7 +55,7 @@ public class MixinMultipartBakedModel implements FabricBakedModel {
 
 	@Override
 	public void emitBlockQuads(BlockRenderView blockView, BlockState state, BlockPos pos, Supplier<Random> randomSupplier, RenderContext context) {
-		BitSet bitSet = (BitSet) this.stateCache.get(state);
+		BitSet bitSet = this.stateCache.get(state);
 
 		if (bitSet == null) {
 			bitSet = new BitSet();
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.renderer.client;
+
+import java.util.BitSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.client.render.model.BakedModel;
+import net.minecraft.client.render.model.MultipartBakedModel;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockRenderView;
+
+import net.fabricmc.fabric.api.renderer.v1.model.FabricBakedModel;
+import net.fabricmc.fabric.api.renderer.v1.render.RenderContext;
+
+@Mixin(MultipartBakedModel.class)
+public class MixinMultipartBakedModel implements FabricBakedModel {
+	@Shadow
+	@Final
+	private List<Pair<Predicate<BlockState>, BakedModel>> components;
+
+	@Shadow
+	@Final
+	private Map<BlockState, BitSet> stateCache;
+
+	@Override
+	public boolean isVanillaAdapter() {
+		return false;
+	}
+
+	@Override
+	public void emitBlockQuads(BlockRenderView blockView, BlockState state, BlockPos pos, Supplier<Random> randomSupplier, RenderContext context) {
+		BitSet bitSet = (BitSet) this.stateCache.get(state);
+
+		if (bitSet == null) {
+			bitSet = new BitSet();
+
+			for (int i = 0; i < this.components.size(); i++) {
+				Pair<Predicate<BlockState>, BakedModel> pair = components.get(i);
+
+				if (pair.getLeft().test(state)) {
+					((FabricBakedModel) pair.getRight()).emitBlockQuads(blockView, state, pos, randomSupplier, context);
+					bitSet.set(i);
+				}
+			}
+
+			stateCache.put(state, bitSet);
+		} else {
+			for (int i = 0; i < this.components.size(); i++) {
+				if (bitSet.get(i)) ((FabricBakedModel) components.get(i).getRight()).emitBlockQuads(blockView, state, pos, randomSupplier, context);
+			}
+		}
+	}
+
+	@Override
+	public void emitItemQuads(ItemStack stack, Supplier<Random> randomSupplier, RenderContext context) {
+		// Vanilla doesn't render MultipartBakedModel's as the BlockState in getQuads
+		// will be null
+		// Also it doesn't use MultipartBakedModel's for items anyways
+	}
+}
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
@@ -57,7 +57,7 @@ public class MixinMultipartBakedModel implements FabricBakedModel {
 
 	@Override
 	public boolean isVanillaAdapter() {
-		return false;
+		return isVanilla;
 	}
 
 	@Inject(at = @At("RETURN"), method = "<init>")
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
@@ -27,6 +27,10 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import net.minecraft.block.BlockState;
 import net.minecraft.client.render.model.BakedModel;
@@ -48,9 +52,22 @@ public class MixinMultipartBakedModel implements FabricBakedModel {
 	@Final
 	private Map<BlockState, BitSet> stateCache;
 
+	@Unique
+	boolean isVanilla = true;
+
 	@Override
 	public boolean isVanillaAdapter() {
 		return false;
+	}
+
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void onInit(List<Pair<Predicate<BlockState>, BakedModel>> components, CallbackInfo cb) {
+		for (Pair<Predicate<BlockState>, BakedModel> component : components) {
+			if (!((FabricBakedModel) component.getRight()).isVanillaAdapter()) {
+				isVanilla = false;
+				break;
+			}
+		}
 	}
 
 	@Override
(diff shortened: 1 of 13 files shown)


fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinWeightedBakedModel.java
```diff
@@ -0,0 +1,60 @@
...
+import net.fabricmc.fabric.api.renderer.v1.render.RenderContext;
+
+@Mixin(WeightedBakedModel.class)
+public class MixinWeightedBakedModel implements FabricBakedModel {
+	@Shadow
+	@Final
+	private int totalWeight;
+	@Shadow
+	@Final
+	private List models; // WeightedBakedModel.Entry is not visable
```
liach:
```suggestion
	private List models; // WeightedBakedModel.Entry is not visible
```
preferably add a `@SuppressWarnings("rawtypes")` as well

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
```diff
@@ -0,0 +1,86 @@
...
+	private Map<BlockState, BitSet> stateCache;
+
+	@Override
+	public boolean isVanillaAdapter() {
+		return false;
+	}
+
+	@Override
+	public void emitBlockQuads(BlockRenderView blockView, BlockState state, BlockPos pos, Supplier<Random> randomSupplier, RenderContext context) {
+		BitSet bitSet = (BitSet) this.stateCache.get(state);
```
liach: Why do you need a cast here? Also I prefer you build the bitset here than later. However, both are minor.

liach (approved):

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
```diff
@@ -0,0 +1,86 @@
...
+	@Final
+	private List<Pair<Predicate<BlockState>, BakedModel>> components;
+
+	@Shadow
+	@Final
+	private Map<BlockState, BitSet> stateCache;
+
+	@Override
+	public boolean isVanillaAdapter() {
+		return false;
```
Technici4n: I think it would be better to return true if all the components are vanilla adapters. This can certainly be cached.

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinWeightedBakedModel.java
```diff
@@ -0,0 +1,61 @@
...
+	@Final
+	private int totalWeight;
+	@SuppressWarnings("rawtypes")
+	@Shadow
+	@Final
+	private List models; // WeightedBakedModel.Entry is not visible
+
+	@Override
+	public boolean isVanillaAdapter() {
+		return false;
```
Technici4n: Same comment.

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
```diff
@@ -0,0 +1,86 @@
...
+				if (bitSet.get(i)) ((FabricBakedModel) components.get(i).getRight()).emitBlockQuads(blockView, state, pos, randomSupplier, context);
+			}
+		}
+	}
+
+	@Override
+	public void emitItemQuads(ItemStack stack, Supplier<Random> randomSupplier, RenderContext context) {
+		// Vanilla doesn't render MultipartBakedModel's as the BlockState in getQuads
+		// will be null
+		// Also it doesn't use MultipartBakedModel's for items anyways
```
Technici4n:
```suggestion
		// Vanilla doesn't use MultipartBakedModel for items.
```

Technici4n (changes requested):
As discussed on Discord, I recommend making sure the models are not ran through the frapi pipeline if it's not necessary, by checking `isVanillaAdapter` on the components of the models. Except for that, the fix looks appropriate to me.

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
```diff
@@ -48,11 +52,24 @@
...
 	@Override
 	public boolean isVanillaAdapter() {
 		return false;
 	}
 
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void onInit(List<Pair<Predicate<BlockState>, BakedModel>> components, CallbackInfo cb) {
+		for (Pair<Predicate<BlockState>, BakedModel> component : components) {
+			if (!((FabricBakedModel) component.getRight()).isVanillaAdapter()) {
+				isVanilla = false;
```
liach, line 67: Hmm, is `isVanilla` called before `emitBlockQuads`? I think if they are called closely you can merge the initialization logic in two places

CoolMineman: > Allows the renderer to optimize or route vanilla models through the unmodified vanilla pipeline if desired.

It can be called earlier to avoid calling emitBlockQuads at all.

Technici4n (approved):
LGTM, thanks!

LemmaEOF (approved):

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinWeightedBakedModel.java
```diff
@@ -0,0 +1,78 @@
...
+
+@Mixin(WeightedBakedModel.class)
+public class MixinWeightedBakedModel implements FabricBakedModel {
+	@Shadow
+	@Final
+	private int totalWeight;
+	@SuppressWarnings("rawtypes")
+	@Shadow
+	@Final
+	private List models; // WeightedBakedModel.Entry is not visible
```
i509VCB: I would personally make the generic type in the list the accessor

CoolMineman: Is that legal and a good idea?

i509VCB: Per type erasure, yes. Means you don't have to cast in that case.

CoolMineman: Attempted to do this
```
        @Shadow
        @Final
	private int totalWeight;
	@SuppressWarnings("rawtypes")
	@Shadow
	@Final
	private List<WeightedBakedModelEntryAccessor> models;
	@Unique
	boolean isVanilla = true;

	@SuppressWarnings("rawtypes")
	@Inject(at = @At("RETURN"), method = "<init>")
	private void onInit(List models, CallbackInfo cb) {
		for (int i = 0; i < models.size(); i++) {
			if (!((FabricBakedModel) models.get(i).getModel()).isVanillaAdapter()) {
				isVanilla = false;
				break;
			}
		}
	}
```
causes the compile error:
```
~/Programming/fabric/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinWeightedBakedModel.java:57: error: cannot find symbol
                        if (!((FabricBakedModel) models.get(i).getModel()).isVanillaAdapter()) {
                                                              ^
  symbol:   method getModel()
  location: class Object
```
additionally for the WeightedPicker.getAt calls I need to cast to `List` anyways

i509VCB: Welp best not to make more of a mess then.

liach: Yo, you need to change
```java
private void onInit(List models, CallbackInfo cb) {
```
to
```java
private void onInit(List<WeightedBakedModelEntryAccessor> models, CallbackInfo cb) {
```
this will work cuz mixin doesn't check about generics when injecting

liach: @coolmineman needa update this field shadow to be parameterized too (use the accessor as type argument)

Technici4n (approved):

LambdAurora (approved):

Technici4n (changes requested):
This needs a testmod so that reviewers can test this correctly before and after, without having to write it themselves. I'm sorry to say this so late, but I have come to the realization that PRs take a long time to be reviewed due to how much validation work needs to be done by the maintainers. I believe setting a higher bar will be better in the long run.

PepperCode1 (approved):

ghost (approved):

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
```diff
@@ -0,0 +1,101 @@
...
+	boolean isVanilla = true;
+
+	@Override
+	public boolean isVanillaAdapter() {
+		return isVanilla;
+	}
+
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void onInit(List<Pair<Predicate<BlockState>, BakedModel>> components, CallbackInfo cb) {
+		for (Pair<Predicate<BlockState>, BakedModel> component : components) {
```
Juuxel, line 65: Ah yes, Mojang has two pair classes of their own (DFU and Minecraft), yet they also use Apache's :thonkjang:

Juuxel (approved):

fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/mixin/renderer/client/MixinMultipartBakedModel.java
liach: They probably will fix this in a massive refactor at some point if they want

Technici4n (approved):
Rebased and expanded the renderer API testmod to test this. Works as expected!

PR #1305, opened 2021-02-03
Make event registration thread safe

Player3324:
This will ensure that concurrent registrations are safe and ensure safe publication of the updated invoker. My chosen approach should be the most appropriate until Java 9, which could tweak the invoker access to use release/acquire instead of the full volatile.

Issue originally spotted by @Technici4n

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java
@@ -18,6 +18,8 @@ package net.fabricmc.fabric.impl.base.event;
 
 import java.lang.reflect.Array;
 import java.util.Arrays;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 
 import net.fabricmc.fabric.api.event.Event;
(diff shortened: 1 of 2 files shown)


Technici4n (approved):

i509VCB (approved):

maityyy:
Lock is too expensive on Windows OS, can't we use a synchronized block?

fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java
```diff
@@ -50,21 +53,27 @@ void update() {
...
 
-		if (handlers == null) {
-			//noinspection unchecked
-			handlers = (T[]) Array.newInstance(type, 1);
-			handlers[0] = listener;
-		} else {
-			handlers = Arrays.copyOf(handlers, handlers.length + 1);
-			handlers[handlers.length - 1] = listener;
-		}
+		lock.lock();
```
liach, line 63: We can just use `synchronized` as we don't really need fancy stuff from the lock.

From https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html:
> ... with the same basic behavior and semantics as the implicit monitor lock accessed using synchronized methods and statements

liach (approved):

fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/ArrayBackedEvent.java
Technici4n: Player wanted to avoid associating a monitor with the object in order not to influence the lock-free path.

CoolMineman (approved):

modmuss50 (approved):

PR #1308, opened 2021-02-05
Fix material builder

PepperCode1:
The `notSolid` method in `FabricMaterialBuilder` returns `Material.Builder` instead of `FabricMaterialBuilder`. Because this is a builder pattern, returning the super type prevents any subsequent method calls from using `FabricMaterialBuilder` methods.


Technici4n (approved):
Is this a breaking change? I would say no because I don't think the return type is part of the method signature, can someone else confirm?

LambdAurora (approved):

Juuxel:
The return type *is* a part of the signature, but the compiler generates a bridge method with the wider return type as well.

Juuxel (approved):

liach (approved):

modmuss50 (approved):
Might be best to double check this is none breaking, im fairly sure its not.

PR #1317, opened 2021-02-14
Compatibility with Gradle 7

### LINKED ISSUE #1316: Modules use Gradle's compile configuration
## The Problem

Gradle's `compile` dependency configuration has long been deprecated, but all of the modules in Fabric API still use it.

Quoting the Gradle documentation, `compile` is
> Superseded by `implementation`.

## Where?

- `build.gradle`:
    - Line 69 (`compile it`)
    - Line 126 (`modCompile "net.fabricmc:fabric-loader:${Globals.loaderVersion}"`)
    - Line 236 (`testmodCompile sourceSets.main.output`)
    - Lines 330 and 333 (`compile project(...)`)
- Inside modules' `build.gradle`s, there is sometimes a `testmodCompile`.

## Why Did I Open This Issue?
`compile` should generally **never** be used, and `implementation` should be used instead. If we need dependencies to be on the compile classpath of consumers, use `api`.

YTG1234:
I used Gradle's CLI `--warning-mode all` option to show deprecation warnings then fixed them.

When importing with Gradle, it still shows the infamous warning:

> ```
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
> Use '--warning-mode all' to show the individual deprecation warnings.
> See https://docs.gradle.org/6.8.2/userguide/command_line_interface.html#sec:command_line_warnings
> ```

But I suspect that's got something to do with Loom itself, which will hopefully be fixed in 0.6.

fixes #1316

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,7 @@ def moduleDependencies(project, List<String> depNames) {
 	def deps = depNames.iterator().collect { project.dependencies.project(path: ":$it", configuration: 'dev') }
 	project.dependencies {
 		deps.each {
-			compile it
+			implementation it
 		}
 	}
 	project.publishing {
@@ -123,7 +123,7 @@ allprojects {
 	dependencies {
 		minecraft "com.mojang:minecraft:$Globals.mcVersion"
 		mappings "net.fabricmc:yarn:${Globals.mcVersion}${Globals.yarnVersion}:v2"
-		modCompile "net.fabricmc:fabric-loader:${Globals.loaderVersion}"
+		modImplementation "net.fabricmc:fabric-loader:${Globals.loaderVersion}"
 	}
 
 	configurations {
@@ -144,8 +144,8 @@ allprojects {
 
 	afterEvaluate {
 		remapJar {
-			input = file("${project.buildDir}/libs/$archivesBaseName-${version}-dev.jar")
-			archiveName = "${archivesBaseName}-${version}.jar"
+			input = file("${project.buildDir}/libs/$archivesBaseName-${archiveVersion}-dev.jar")
+			archiveFileName = "${archivesBaseName}-${archiveVersion}.jar"
 		}
 
 		artifacts {
@@ -233,7 +233,7 @@ task runAutoTestServer(type: RunServerTask) {
 
 subprojects {
 	dependencies {
-		testmodCompile sourceSets.main.output
+		testmodImplementation sourceSets.main.output
 	}
 
 	task remapMavenJar(type: Copy, dependsOn: remapJar) {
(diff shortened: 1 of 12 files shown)


YTG1234:
[[Actions](https://github.com/FabricMC/fabric/pull/1317/checks?check_run_id=1897114928)] This error is concerning, I don't know how to fix it.

YTG1234:
RE: 73f158f142b951fe111989f9ae146c3c98148563

`version` has been deprecated in favour of `archiveVersion` but `archiveVersion` doesn't work... ?

YTG1234:
Concern, I need help

liach:
This is why I'n reluctant when you asked in #1316. You need to know more about basic gradle, like you need to use archiveVersion.get() to get an actual string

YTG1234:
> This is why I'n reluctant when you asked in #1316. You need to know more about basic gradle, like you need to use archiveVersion.get() to get an actual string

except the Gradle docs didn't mention that... I read the docs

YTG1234:
well maybe I missed it

EDIT: Yes, found it, the type is `Property<String>` and not `String`

liach:
> > This is why I'n reluctant when you asked in #1316. You need to know more about basic gradle, like you need to use archiveVersion.get() to get an actual string
> 
> except the Gradle docs didn't mention that... I read the docs

you did? the doc would say it is a `Property<String>` and you need to call get to get its actual value. there is no syntax sugar here that implies get calls.

in general, i am just concerned when someone does something because they see a warning than they know what they are actually doing. if you figure out i'm glad to support this

YTG1234:
Usually Gradle doesn't hate me this much, today is just a bad day for Gradle 😄

liach:
yes, some syntax sugars in gradle are quite cursed and can cause significant difficulties in debugging cause of problems. like you will realize the problem if the property doesn't use = for set call sugars

YTG1234:
Is `prop = thing` syntax sugar for `prop.set(thing)`? If so, I'm more concerned. Doesn't Groovy have sane properties?

liach:
also for gradle using property: just like file collection, they allow lazy computation on demand, just like java functional code

liach:
> Is `prop = thing` syntax sugar for `prop.set(thing)`? If so, I'm more concerned. Doesn't Groovy have sane properties?

i think this is gradle specific, not groovy. i personally use explicit set calls in my builds

YTG1234:
> > Is `prop = thing` syntax sugar for `prop.set(thing)`? If so, I'm more concerned. Doesn't Groovy have sane properties?
> 
> i think this is gradle specific, not groovy. i personally use explicit set calls in my builds

I usually use the Kotlin DSL so I *have* to use `set`. Anyway, I never would have thought that GitHub PRs can get off topic 🤔

liach (approved):

modmuss50:
This is something I had planned to tackle all at once with the next loom update long with some other changes I have planned for the buildscript side of things. No point pulling this in now before gradle 7 is even out, and without the loom update.

i509VCB:
If I recall the compile configuration is transitive, does this mean we can use implementation or do we need to use api configuration.

modmuss50:
implementation is runtime transitive, api is compile transitive. API should be used where the API of a module depends on another module (Does this even happen?)

Juuxel:
fabric-api-base should be `api` in many cases as `Event` is exposed in APIs.

modmuss50:
These chanages are far from what is required. There are still some issues in loom that will need fixing. I have started a larger scope PR here for buildscript changes: https://github.com/FabricMC/fabric/pull/1338/

PR #1319, opened 2021-02-14
Fix a WorldRenderEvents.BLOCK_OUTLINE bug

Technici4n:
When `getBuffer` is called on `VertexConsumerProvider.Immediate` (the one used for outline rendering) with a different layer than the active one, it draws the old render layer and then changes its internal state to accept vertices for the format of the new render layer. In particular, this means that the `RenderLayer.getLines()` `VertexConsumer` obtained by vanilla before the `BLOCK_OUTLINE` event but used after the handlers are run may become invalid if the event handlers change the current render layer of the immediate mode VCP.

To reproduce, just put `worldRenderContext.consumers().getBuffer(RenderLayer.getTranslucent())` in an event handler and the game will crash when rendering the default block outline as it tries to emit lines to the immediate mode VCP still configured to emit translucent quads.

This PR:
* Adds a call to `getBuffer(RenderLayer.getLines())` after the invocation of the event handlers to make sure the default block outline emits to the immediate mode VCP in the correct state.
* Changes the implementation of `BlockOutlineContext#vertexConsumer()` to make sure that the VCP is in the correct state after each call, and deprecates the method as there is no reason to keep it.

Full discussion on fabricord: https://discord.com/channels/507304429255393322/566276937035546624/810582855603847169.

### DIFF
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -120,7 +120,11 @@ public interface WorldRenderContext {
 	 * {@code WorldRenderer.drawBlockOutline}.
 	 */
 	@Environment(EnvType.CLIENT)
-	public interface BlockOutlineContext {
+	interface BlockOutlineContext {
+		/**
+		 * @deprecated Use {@link #consumers()} directly.
+		 */
+		@Deprecated
 		VertexConsumer vertexConsumer();
 
 		Entity entity();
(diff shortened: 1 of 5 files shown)


liach (approved):
I know this immediate swaps buffers, but didn't expect that to cause a bug

LambdAurora (approved):
Just a note: imo, the deprecated method in this PR should be directly removed in the next 1.17 snapshots after this PR is merged so it's not just plain forgotten afterwards.
(If really we can wait 1.18 to remove it or module major version, but then make sure to not forget it)

Platymemo (approved):

Technici4n (commented):
I will be adding a little testmod + some screenshots to make reviewing easier.

Technici4n:
Testmod:
image

Uncomment the added line 117 in `MixinWorldRenderer` to see the original crash that lead to this PR. The changes to `vertexConsumer()` are there to prevent a similar crash from happening if mods decide to use it.

modmuss50 (approved):
LGTM. Someone who knows rendering a bit better might be able to do a better review than me . Thanks for adding a testmod 👍

Player3324 (approved):

PR #1320, opened 2021-02-15
Add clear to MutableQuadView

alex5nader:
Added a clear method to MutableQuadView and inheritors.

Indigo currently combines the MQV and QE impls anyway, and the clear method actually already exists. This PR provides a safer way to access it.

### DIFF
--- a/fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/MutableQuadViewImpl.java
+++ b/fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/MutableQuadViewImpl.java
@@ -53,7 +53,7 @@ public abstract class MutableQuadViewImpl extends QuadViewImpl implements QuadEm
 		clear();
 	}
 
-    @Override
+	@Override
 	public MutableQuadViewImpl clear() {
 		System.arraycopy(EMPTY, 0, data, baseIndex, EncodingFormat.TOTAL_STRIDE);
 		isGeometryInvalid = true;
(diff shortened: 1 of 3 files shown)


Technici4n:
This is a breaking change, so I'm afraid we can't allow it. There are workarounds to clear the quad at the moment.

alex5nader:
Is this breaking because it changes the signature of clear?

Technici4n:
No, it's because third-party renderers such as Canvas or Indium will have to implement this additional method.

alex5nader:
Does that mean FRAPI never going to allow any changes that introduce interface methods? For what it's worth, both Canvas and Indium both already have this method present, like Indigo (all that's needed is to change the return type).

Technici4n:
I'm not sure, I'll wait for others to comment.

fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/MutableQuadViewImpl.java
```diff
@@ -53,7 +53,8 @@ public final void begin(int[] data, int baseIndex) {
 		clear();
 	}
 
-	public void clear() {
```
i509VCB, line 56: Breaking change

i509VCB: Wait this is impl lmao

Technici4n: This is an impl class though, so we don't care?

i509VCB:
More specifically what's the use case for this?

alex5nader:
I'm adapting my mod's QuadTransform code to store a Mesh directly. The quad transform needs to returns false (culling the quad), so I was hoping that I could have a similar functionality in the QuadEmitter I'm using. Basically, return false => clear, return true => emit.

Technici4n:
After a discussion on Discord, we can't allow this in as it's a breaking change.
A new `QuadEmitter` can be queried and used to cancel the previous the quad if it was not emitted.

Link: https://discord.com/channels/507304429255393322/566276937035546624/810789836348063754.

PR #1321, opened 2021-02-16
Add ability for minecarts to specify their detector rail comparator value

williewillus:


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,18 @@
+package net.fabricmc.fabric.api.object.builder.v1.entity;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+public interface MinecartComparatorLogic<T extends AbstractMinecartEntity> {
+	/**
+	 * Compute the comparator output of a detector rail when a minecart is resting
```
haykam821, line 30: It would probably be good to document how often this is called.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,21 @@
...
+public class FabricMinecartComparatorLogicRegistry {
+	private static final Map<EntityType<?>, MinecartComparatorLogic<?>> logics = new HashMap<>();
+
+	/* @Nullable */
+	@SuppressWarnings("unchecked") // just to appease typechecker, we checked in the entity type builder that the type of the logic exactly matches the entity type
+	public static MinecartComparatorLogic<AbstractMinecartEntity> getCustomComparatorLogic(EntityType<?> type) {
+		return (MinecartComparatorLogic<AbstractMinecartEntity>) logics.get(type);
+	}
+
+	static void registerCustomLogic(EntityType<?> type, MinecartComparatorLogic<?> logic) {
```
haykam821: This could simply be named `register` like [`FabricDefaultAttributeRegistry`](https://github.com/FabricMC/fabric/blob/9354966b6c119b52108449b5c5f5622c89c5c53e/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDefaultAttributeRegistry.java).

```suggestion
	static void register(EntityType<?> type, MinecartComparatorLogic<?> logic) {
```

haykam821 (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
```diff
@@ -269,6 +281,30 @@ protected FabricEntityTypeBuilder(SpawnGroup spawnGroup, EntityType.EntityFactor
 		return type;
 	}
 
+	public static class Minecart<T extends AbstractMinecartEntity> extends FabricEntityTypeBuilder<T> {
+		private MinecartComparatorLogic<T> customComparatorLogic = null;
+
+		protected Minecart(SpawnGroup spawnGroup, EntityType.EntityFactory<T> factory) {
```
liach: I personally am against this. With this, you need to override every superclass method that returns self to return a self with a correct type (i.e. type `Minecart` than `FabricEntityTypeBuilder`)

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import net.minecraft.entity.EntityType;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+
+public class FabricMinecartComparatorLogicRegistry {
+	private static final Map<EntityType<?>, MinecartComparatorLogic<?>> logics = new HashMap<>();
+
+	/* @Nullable */
```
liach: Don't we already have jetbrains nullable?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,21 @@
...
+	@SuppressWarnings("unchecked") // just to appease typechecker, we checked in the entity type builder that the type of the logic exactly matches the entity type
```
liach:
```suggestion
	@SuppressWarnings("unchecked")
```
This documentation imo isn't too useful as people know this is a trick for compiler

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/DetectorRailBlockMixin.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.object.builder.v1.entity.FabricMinecartComparatorLogicRegistry;
+
+@Mixin(DetectorRailBlock.class)
+public abstract class DetectorRailBlockMixin {
+	@Shadow protected abstract <T extends AbstractMinecartEntity> List<T> getCarts(World world, BlockPos pos, Class<T> entityClass, /* @Nullable */ Predicate<Entity> entityPredicate);
+
+	@Inject(at = @At("HEAD"), method = "getComparatorOutput", cancellable = true)
```
liach, line 42: Had mixin some tools to identify where operand stacks are empty, I bet people would be way more eager to inject within method bodies! :laughing: I really hope we can inject before the ldc for `CommandBlockMinecartEntity.class`

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
LambdAurora: We have Jetbrains nullable.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
williewillus: So what do you propose instead? There are already two overrides in the same file ~~that don't do this~~ (they do), so I'm not seeing how this is constructive unless you provide a solution.

Should I just do the same?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/DetectorRailBlockMixin.java
williewillus: That is way more brittle for no benefit.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
williewillus: Sure.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
Technici4n: What's the point, can't people call `customComparatorLogic` before calling the other methods?

emilyploszaj (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
liach: I suggest something in the lines of https://github.com/FabricMC/fabric/blob/1.16/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricDefaultAttributeRegistry.java
i.e. people can register through entity type builder, yet they are allowed to associate comparator output to entity type freely

Technici4n: I believe liach is right. It's simpler to directly expose `MinecartComparatorLogicRegistry` (we can drop the `Fabric` prefix imo). (Nevermind there is already `FabricDefaultAttributeRegistry` so we can keep the prefix 😄)

i509VCB: The reason I added the Living and mob subclass builders is that it puts all the code in one spot, vs registering and then forgetting to call a method later for entity attributes and spawn logic.

i509VCB: The other facility exists fyi like with default attribute registries is to override those on existing entities.

Technici4n: > and then forgetting to call a method later for entity attributes and spawn logic

That's a very weak point imo. Basing testing would show that it clearly doesn't work, and nothing stops you from forgetting to call the specialized builder method anyway.

Technici4n: Especially with something as "unusual" as minecarts, maybe it makes more sense for entity attributes and spawn logic which are quite important to a lot of entities.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,34 @@
...
+	 * on top of it. Called from {@link net.minecraft.block.DetectorRailBlock#getComparatorOutput}.
+	 * @param minecart The minecart on the rail
+	 * @param state BlockState of the rail
+	 * @param pos Position of the rail
+	 * @return A redstone power value >= 0 to use, else a value < 0 to run vanilla's logic
```
liach:
```suggestion
	 *
	 * @param minecart the minecart on the rail
	 * @param state the block state of the rail
	 * @param pos the position of the rail
	 * @return a redstone power value {@literal >= 0} to use, else a value {@literal < 0} to run vanilla's logic
```
stylistic issues, mainly the left and right angle brackets will break javadoc gen

liach (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
williewillus: I'm fine with splitting it from the entity type, but what module should I put it in if I do that?

liach: Still object builders. Notice how FabricDefaultAttributeRegistry sits right next to the entity type builder in object builders.

Technici4n:
Hi @williewillus, are you planning to update this PR?

williewillus:
Not right now. Someone else can feel free to pick it up.

williewillus:
Hey all, reviving this. Addressed all comments, please review. Thanks!

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.entity.EntityType;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+
+public class FabricMinecartComparatorLogicRegistry {
```
williewillus: not sure if this needs `Fabric` in front.

Technici4n: Imo it's not that useful.

Technici4n: I'd suggest inlining the registry's `get` and `set` into the custom comparator logic interface. This will make the overall PR a little bit longer, but the API surface shorter.

Technici4n (commented):
Nice to have you back!

Would it be possible to add a testmod to help test this? Perhaps just changing the behavior of one of the existing minecarts? (Not sure if the API allows it)

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/DetectorRailBlockMixin.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.object.builder.v1.entity.FabricMinecartComparatorLogicRegistry;
+
+@Mixin(DetectorRailBlock.class)
+public abstract class DetectorRailBlockMixin {
+	@Shadow protected abstract <T extends AbstractMinecartEntity> List<T> getCarts(World world, BlockPos pos, Class<T> entityClass, /* @Nullable */ Predicate<Entity> entityPredicate);
```
liach: Fyi we have since enabled jetbrain annotations, so we can import `org.jetbrains.annotations.Nullable` and use it instead

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,34 @@
...
+public interface MinecartComparatorLogic<T extends AbstractMinecartEntity> {
+	/**
+	 * Compute the comparator output of a detector rail when a minecart is resting
+	 * on top of it. Called from {@link net.minecraft.block.DetectorRailBlock#getComparatorOutput}.
+	 * @param minecart The minecart on the rail
+	 * @param state BlockState of the rail
+	 * @param pos Position of the rail
+	 * @return A redstone power value {@literal >=} 0 to use, else a value {@literal <} 0 to run vanilla's logic
+	 */
+	int getComparatorValue(T minecart, BlockState state, World world, BlockPos pos);
```
liach: We might not need the world parameter as it's just `minecart.world`, isn't it?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
liach: Don't think it's that much better when you need another class in impl package to hold the map field. Imo the true concerns should be whether we support force overriding (like as is) or throw if a logic is already present.

liach (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricMinecartComparatorLogicRegistry.java
williewillus: Yeah not sure about inlining.

As for force overriding, I can match what other parts of this module do and just log a message when something is overridden. I don't think throwing is necessary.

williewillus:
Updated:

1. Remove needless World param from interface
2. Log when a registration is overridden, like what FabricDefaultAttributeRegistry does
3. use Jetbrains Nullable where noted
4. Try all custom logic-capable carts before moving on to vanilla.

Technici4n (approved):
Looks good. Sorry for the delay, I had forgotten about this...

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * Compute the comparator output of a detector rail when a minecart is resting
+	 * on top of it. Called from {@link net.minecraft.block.DetectorRailBlock#getComparatorOutput}.
+	 * @param minecart The minecart on the rail
+	 * @param state BlockState of the rail
+	 * @param pos Position of the rail
+	 * @return A redstone power value {@literal >=} 0 to use, else a value {@literal <} 0 to try the next minecart with
+	 * 	a registered logic. If no logic chooses to provide a value, vanilla's logic is invoked.
+	 */
+	int getComparatorValue(T minecart, BlockState state, BlockPos pos);
```
Juuxel, line 38: Doesn't really matter, but `getComparatorOutput` would match the yarn naming here, esp. since "comparator value" sounds a bit weird. Either way is fine.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.entity.EntityType;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+import net.minecraft.util.registry.Registry;
+
+public class MinecartComparatorLogicRegistry {
```
Juuxel:
```suggestion
/**
 * A registry for {@linkplain MinecartComparatorLogic custom minecart compator logic}.
 */
public final class MinecartComparatorLogicRegistry {
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final Map<EntityType<?>, MinecartComparatorLogic<?>> logics = new HashMap<>();
+
+	@Nullable
```
Juuxel, line 43:
```suggestion
	/**
	 * Gets the registered custom comparator logic for the specified minecart entity type.
	 *
	 * @param type the entity type
	 * @return the comparator logic, or {@code null} if not registered
	 */
	@Nullable
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.object.builder.v1.entity;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.vehicle.AbstractMinecartEntity;
+import net.minecraft.util.math.BlockPos;
+
+public interface MinecartComparatorLogic<T extends AbstractMinecartEntity> {
```
Juuxel, line 28:
```suggestion
/**
 * Provides custom comparator output for minecarts resting on detector rails.
 * @param <T> the handled minecart type
 */
@FunctionalInterface
public interface MinecartComparatorLogic<T extends AbstractMinecartEntity> {
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogic.java
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * Compute the comparator output of a detector rail when a minecart is resting
+	 * on top of it. Called from {@link net.minecraft.block.DetectorRailBlock#getComparatorOutput}.
+	 * @param minecart The minecart on the rail
+	 * @param state BlockState of the rail
```
Juuxel:
```suggestion
	 * @param state Block state of the rail
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+	@SuppressWarnings("unchecked")
+	public static MinecartComparatorLogic<AbstractMinecartEntity> getCustomComparatorLogic(EntityType<?> type) {
+		return (MinecartComparatorLogic<AbstractMinecartEntity>) logics.get(type);
+	}
+
+	public static <T extends AbstractMinecartEntity> void register(EntityType<T> type, MinecartComparatorLogic<? super T> logic) {
```
Juuxel, line 58:
```suggestion
	/**
	 * Registers a comparator logic for a minecart entity type.
	 *
	 * <p>Registering a second value for an entity type will replace the old logic.
	 *
	 * @param <T>   the handled minecart type
	 * @param type  the minecart entity type
	 * @param logic the logic to register
	 */
	public static <T extends AbstractMinecartEntity> void register(EntityType<T> type, MinecartComparatorLogic<? super T> logic) {
```

Juuxel (changes requested):
Suggested missing JD and some other minor comments.

Juuxel:
Also... object builders feels like a weird place for this API - this has nothing to do with "builders for objects vanilla has locked down". Especially this sort of custom logic stuff is usually in other modules like item API for items, although there's no catch-all equivalent for entities.

williewillus:
Thanks for the comments. It can always be moved in the future if more things come up. I think Fabric API has followed that process for a few modules now, where things organically "bud off" into their own modules when the need arises. There's other things in object builders that are kind of questionable too :P

Player3324 (approved):
LGTM with Juuxel's proposed changes

Juuxel (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+	 *
+	 * <p>Registering a second value for an entity type will replace the old logic.
+	 *
+	 * @param <T>   the handled minecart type
+	 * @param type  the minecart entity type
+	 * @param logic the logic to register
+	 */
+	public static <T extends AbstractMinecartEntity> void register(EntityType<T> type, MinecartComparatorLogic<? super T> logic) {
+		if (logics.put(type, logic) != null) {
+			LOGGER.debug("Overriding existing minecart comparator logic for entity type {}", Registry.ENTITY_TYPE.getId(type));
```
modmuss50: This should be a warning at the very least. In other places similar to this there is a hard crash to prevent unpredictable behaviour

modmuss50 (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
williewillus: The place I referenced this from didn't have a crash, but I can make this a warning.

modmuss50 (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/MinecartComparatorLogicRegistry.java
modmuss50: Great thanks 👍

ramidzkh (approved):

PR #1322, opened 2021-02-16
Adds SlotWithClickAction

emilyploszaj:
Adds `SlotWithClickAction`, an interface slots can implement to have a click action à la `Item#onClicked`. This is called before `Item#onClicked` and `Item#onStackClicked`. I've run into a handful of situations where this behavior would be useful, here's a couple potential use cases:
* An item filter, where clicking a stack onto the filter slot adds it as a filter, instead of consuming the stack.
* A sorted storage screen, where clicking on any slot with a stack in hand will deposit it into the system, instead of swapping with the clicked slot, and clicking on it with an empty hand will pull at most 64 items out of the stacks that internally hold more than 64.
* An inventory that consumes added items and spits out new resources.
* Other cases where you want slots that don't actual represent an `ItemStack`, but a more abstract portion of a screen.


fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
```diff
@@ -0,0 +1,38 @@
...
+
+import net.minecraft.entity.player.PlayerInventory;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ClickType;
+
+/**
+ * An interface for slots that consume a click action.
+ * Slot click actions are called before item click actions, and returning true will neglect to have them called whatsoever.
+ */
+public interface SlotWithClickAction {
```
liach, line 29: I prefer to name this `ClickAwareSlot`

jaskarth: I think `SlotWithClickAction` is better as it represents the slot having an action when it's clicked.

l-Luna: ClickAwareSlot is shorter, but less descriptive. All slots are clickable, but only these slots have a special click action.

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package net.fabricmc.fabric.api.screenhandler.v1;
+
+import net.minecraft.entity.player.PlayerInventory;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ClickType;
+
+/**
+ * An interface for slots that consume a click action.
```
Juuxel: `Slot` should be linked here (especially since 1.17 adds that entity command item slot as well)

emilyploszaj: Good call

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
TelepathicGrunt: `SlotWithClickAction` sounds better and easier to understand. Shorter names aren't always the best when you lose important details

liach: Then `ClickCallbackSlot` imo is better. All slot clicks have click action objects assigned, dont' they?

In fact in this case you can also name it `SlotClickCallback`. Either way, these two names proposed are better than `SlotWithClickAction` as every slot can be clicked.

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
liach: well, the fact that screen handler's slot and commands' item slots are distinct systems internally (though often times closely related) already can be confusing. So linking is a good idea.

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
emilyploszaj: This isn't really what I think of when I think of a callback, but also, slots by themselves *do not* have a click action, screen handlers mutate them manually, the same way items didn't have any until 1.17. I think calling what happens when a click is handled differently that standard a `ClickAction` makes more sense by far than a `ClickCallback`.

l-Luna: every slot can be clicked, not every slot has a click action. SlotClickCallback implies you're passing a seperate callback object elsewhere, and ClickCallbackSlot is just more confusing and longer.

jaskarth (approved):

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
liach: But this is a callback. You are using the slot object effectively as a callback object and how this method is called is like a callback. `XxWithYy` are generally confusing names for our naming style, just like `CallbackInfoReturnable` is a bad name because it is an info than a returnable.

In addition, vanilla has a click action for text. This use of click action will just be like what juuxel said, increase confusion.

l-Luna: at this point you're being contrary for the sake of being contrary. all other reviews agree with the existing name and find it to be the best suggested name.

Technici4n: `ClickActionSlot` and be done with it. ;)

liach: > at this point you're being contrary for the sake of being contrary.

Lol what, this is just that typical hypocrisy where you say you allow freedom of speech but enforce people to think alike. So much for arguments! When you cannot debate, you eliminate the opposition!

I still hold that names with `With`, like `ScreenWithHandler`, are confusing, just like speaking in passive voice than the active voice. My emphasis is that it's `ClickAware` and it serves as a `ClickCallback` so it comes first in name. Yours just don't align with the suffix naming style.

williewillus (approved):

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
Technici4n: I don't like `With`, but I don't like `Aware` either. We can just remove `With` and go for `ClickActionSlot` imo.

emilyploszaj: It's a slot with a click action, it's not a click action slot, I do agree that `With` is generally not great, but when something is defined as being one thing with another thing, that naming scheme is at the very least applicable, and in my opinion ideal.

Vazkii: if anyone asks me where to put their bikes I'll link them to this thread

Technici4n: `ClickActionSlot` -> `Slot`, but also has a `ClickAction`. We may want to emphasize somehow that this click action happens before `Item#onClick`.

williewillus: > Lol what, this is just that typical hypocrisy where you say you allow freedom of speech but enforce people to think alike. So much for arguments! When you cannot debate, you eliminate the opposition!

100% agreement ~~never~~ rarely happens for any debate. Being in the minority does not mean your "freedom of speech" is being suppressed, it simply means most people don't believe your point. Don't try to bring that crap into a technical discussion, it's obviously in bad faith.

This just reads to me as "if I don't get it my way or people don't agree with me, then you're suppressing my right to debate", which is patently false.

Honestly, Fabric API has a huge reputation for bikeshedding, to the point where I have spoken to people that are afraid to even contribute because minor, inconsequential pieces of their PR will be nitpicked and bikeshedded to death. And from my brief experience contributing and interacting with some reviewers, that's true. For a project that so many people moved to due to a certain project next door's supposedly unhealthy contribution culture, I think we can do better.

emilyploszaj: > ClickActionSlot -> Slot, but also has a ClickAction. We may want to emphasize somehow that this click action happens before Item#onClick.

Documentation for both the interface and the method specifies that this is called beforehand, I think it would be cluttersome to put in the method name somehow.

liach: > 
> 
> I don't like `With`, but I don't like `Aware` either. We can just remove `With` and go for `ClickActionSlot` imo.

This is a good one. :+1:

eutro (approved):

LambdAurora (approved):

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
liach: > 
> 
> > Lol what, this is just that typical hypocrisy where you say you allow freedom of speech but enforce people to think alike. So much for arguments! When you cannot debate, you eliminate the opposition!
> 
> 100% agreement ~never~ rarely happens for any debate. Being in the minority does not mean your "freedom of speech" is being suppressed, it simply means most people don't believe your point. Don't try to bring that crap into a technical discussion, it's obviously in bad faith.
> 
> This just reads to me as "if I don't get it my way or people don't agree with me, then you're suppressing my right to debate", which is patently false.

Another TLDR, huh? Also in your quote above, you missed this:
` you're being contrary for the sake of being contrary.`
What is the purpose of this sentence? It clearly asks me to stop arguing.

Anyways, I am bikeshedding in my domain, i.e. suggesting intuitive and efficient names for things. That's why I send pull requests to yarn. In addition, I am not requesting changes to this pr for this name; the rest of the pr is great.

hYdos (approved):

cplir-c (approved):

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
emilyploszaj: Most people seem to prefer `SlotWithClickAction` to any alternatives suggested, `ClickActionSlot` isn't very clear what exactly it is either. Is it a slot with a click action? Or an event for clicking an action slot? A slot that *is* a click action? Sometimes with brevity you lose clarity, and I think this is a situation where that is the case.

haykam821: I prefer `SlotClickListener` or `SlotClickHandler`.

snakefangox (approved):

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
Technici4n: Let's go for `SlotWithClickAction`, it should be clear enough. `Listener` or `Handler` is too vague because it doesn't convey that the slot itself is handling the click action.

i509VCB: Sometimes you need to use longer names to clearly specify what the class does. Hence why we have `LivingEntityFeatureRendererRegistrationCallback`.

I think the name is fine. For making it more obvious what to do with the interface, I would add some documentation along the lines of "To allow for enhanced click action handling, implement this interface onto a Slot".

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
```diff
@@ -0,0 +1,38 @@
...
+	/**
+	 * Called when this slot is clicked, before calling any item methods.
+	 * Used to determine whether the click is consumed by the slot.
+	 *
+	 * @param heldStack the stack the in the player's cursor
+	 * @param type the click type, either left or right click
+	 * @param inventory the player's inventory, the held stack can be safely mutated
+	 * @return whether to consume the click event or not, returning false will have the event processed by items, and if left unconsumed will be processed by the screen handler
+	 */
+	boolean onClicked(ItemStack heldStack, ClickType type, PlayerInventory inventory);
```
i509VCB, line 39: Do users need to take into account special logic on the logical server/logical client or should users just treat this like normal.

emilyploszaj: This is called on both client and server, mutations to state should be consistent across client and server, as is the case with any other slot method or `Item#onClicked` or `Item#onStackClicked`, but sided actions can take place (such as spawning entities using server worlds) as it will run on both sides.

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/SlotWithClickAction.java
emilyploszaj: > For making it more obvious what to do with the interface, I would add some documentation along the lines of "To allow for enhanced click action handling, implement this interface onto a Slot".

is `An interface for {@link net.minecraft.screen.slot.Slot}s that consume a click action.` not ideal? Would it be better to use the phrase "enhanced click action handling" or something similar to this effect?

i509VCB: The current comment is sufficient, but elaborating on what you do with the interface explicitly (implement the interface) is much more clear to a new user.

emilyploszaj: I can't close this chain because it has been about more than just docs, but I've added a bit of elaboration to that tune.

i509VCB (approved):
This looks good, thanks for the PR.

LemmaEOF (approved):
LGTM!

Platymemo (approved):

RedstoneParadox (approved):

fabric-screen-handler-api-v1/src/main/resources/fabric-screen-handler-api-v1.mixins.json
```diff
@@ -3,7 +3,8 @@
   "package": "net.fabricmc.fabric.mixin.screenhandler",
   "compatibilityLevel": "JAVA_8",
   "mixins": [
-    "ServerPlayerEntityMixin"
+    "ServerPlayerEntityMixin",
+    "ItemStackMixin"
```
RedstoneParadox: I believe standard convention is to have these in alphabetical order.

Technici4n (changes requested):
This needs a testmod so that reviewers can test this correctly before and after, without having to write it themselves. I'm sorry to say this so late, but I have come to the realization that PRs take a long time to be reviewed due to how much validation work needs to be done by the maintainers. I believe setting a higher bar will be better in the long run.

Editing the existing screen handler test mod and adding a comment in `ScreenHandlerTest` to explain the expected behavior should be a good start. Making the `BagScreenHandler` prevent interaction with bundles would be one way to do it, or even writing a `FilterScreenHandler` with a filter item would work.

(if you need help running fabric API you can ask on the `#api` channel on discord)

Juuxel (approved):

Technici4n (approved):
Thanks for the test mod, I tested this a bit in a singleplayer world and it seems to be working fine.

LambdAurora (approved):

Technici4n:
Also, I had to add the following lines to the `build.gradle` to get the test mod to run with only its dependencies on the classpath, I think you should add them even if it's not directly related to the rest of the PR. :)
```groovy
	testmodCompile(project(path: ':fabric-tag-extensions-v0', configuration: 'dev'))
	testmodCompile(project(path: ':fabric-tool-attribute-api-v1', configuration: 'dev'))
```

modmuss50 (approved):
Seems easy enough to me, Is this better than an event? It does kinda limit it to your own slots where as an event would allow you to incercept any slot click?

emilyploszaj:
I'd argue the situations in which you want to capture a *lot* of different slots is probably one where you're working with specific items, where implementing click actions on those items with the vanilla system (which does allow for distinguishing slots) is going to cover that use case. A mixin for applying this interface to existing slots is also relatively painless and I'd say quite a niche use case. I can not think of a use case for intercepting all slot clicks for all items where you shouldn't just be using mixin to make those drastic changes to behavior.

modmuss50:
Ok, I agree. Sounds reasonable 👍

Player3324:
I think the feature is good to have. There are some open questions:

- The above reasoning behind not making this an event appears to come down to "use mixin for foreign classes", however the philosophy in Fabric-API is that Mixin use shouldn't be mandatory for sufficiently common use cases. Targeting another mod's screen needs to be considered as well as the need to have a concrete class to target in the first place. From my POV this strongly suggests that mods would end up re-implement this API very similarly but in event fashion for those applications. 
Any interface solution needs to have clear advantages over an event and the foreign slot class case has to be sufficiently exotic. Usually the main reason against events is the performance impact from the global inquiry (asking everyone about everything) - which doesn't really apply for manually triggered actions.
There are also composition/chainability concerns, for events there can obviously be many subscribers and with allowX+onX we further isolate otherwise conflicting use cases.
Events and Fabric-API's features in general are perfectly fine for implementing drastic behavior changes, Mixin is only another tool to go beyond what's normally possible. Trying to limit feature impact is an unwinnable effort, there are many other ways to screw the gui up. Event registrations aren't particularly opaque or debugging hostile IMO.
- Whether the boolean return means pass or cancel is not obvious from the method name (vs e.g. the allowX events) and may not be consistent with other similar APIs.
- The method name may be able to convey the order better. Something like handleClick should already be an improvement, we also use present tense for onX type notifications by convention.
- ItemStack.onStackClicked doesn't exist in 1.16 while the API mirrors its signature closely - may thus still be in flux in MC's side. This concern is easily resolved by researching the 1.16 port. 
- Providing the player over the player inv makes more sense imo and seems closer to what the implementer may need directly

emilyploszaj:
To respond to your second through fifth questions in order,
* I agree with this, which is why it's documented, the reasoning for returning a boolean here is to adequately mirror the `ItemStack` equivalent methods.
* The name here is similarly mirroring the `ItemStack` equivalent methods (particularly with the past tense) so a renaming of this event in tandem with the yarn mappings for the `ItemStack` methods may be in order.
* In regards to 1.16, my intention was to not backport this feature, as it exists as a very simple piggyback off of the `ItemStack` equivalent methods. However, you are right in saying this method is still in flux. It was stable for a couple months but the last snapshot changed the signature again (primarily due to changes to the `ScreenHandler` system) which leads into the next point,
* This will be changed to a player, as the signature of the `ItemStack` methods were changed to be passed a player with the last snapshot.

In regards to your first question,
* One of the more significant concerns for me about implementing this as an event is that it makes the most common use case harder to implement and more opaque to understand, for the small benefit of covering more esoteric use cases, I imagine most of these situations would be intrusively acting upon the game or other mods already anyway, by nature of the scope, so cutting a mixin out in this case would hardly serve much benefit. It's reasonably trivial to implement slot click behavior yourself with an injection into `ItemStack` or handling the clicks with indices manually in your screen handler, but the large downside is the behavior is significantly more opaque. The greatest boon of implementing this handling on a slot itself is the fact that it is stored with the slot, where it makes structural sense to be, aiding in code readability and ease of implementation. Switching to an event system would non-intuitively make this entire system significantly less useful, in my opinion.
* I don't think layerability makes much sense in the context of a slot click action. I'd agree that an event makes more sense for click actions for item stacks, as behavior for arbitrary items is vague, but as it is slots are designed for specific purposes and their click actions will generally reflect this.

Player3324:
Wrt mirroring ItemStack methods and anything in the realm: IMO it is not Fabric-API's job to somehow expose vanilla code as-is, but to allow its users to solve problems in the most appropriate way.

In that sense the existence of the ItemStack method is merely a nice way to make our implementation easier, but it shouldn't guide the feature itself too much. The ideal API is abstract enough to outlive large vanilla refactors without having to compromise too much on the supported features or implementation complexity.


I'm aware that handling the event is quite annoying right now. We don't have a nice approach to work with temporary context objects like Slot since event subscriptions are meant to be static, much like actually using `static { }` to wire them up. Likewise it isn't ideal how we have to register every event individually despite the interface declaration already more or less stating what is expected to happen by the mod. I think there is some room for a mechanism to solve this in general, I'll have to do some experimentation...

Technici4n:
@emilyalexandra This PR can be updated now that #1369 was merged.

Technici4n:
Let me know if you are going to continue work on this PR.

Note that this didn't turn out to be required for AE2 or MI because we already override the slot click logic for our special fake slots, but I guess there could be some other uses.

PR #1327, opened 2021-02-22
Loot table Visitors [WIP]

RedstoneParadox:
This PR intends to provide a way to directly modify certain aspects vanilla loot tables without having to write complex boilerplate or replace the whole thing (See #1287). This is done using the visitor pattern to avoid coupling the algorithm to the structure of the loot table in the event it changes.

This PR is nowhere near done and is only being opened so people are aware that it's being worked on and can comment on it.

### DIFF
--- a/fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/visitor/LootTableVisitor.java
+++ b/fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/visitor/LootTableVisitor.java
@@ -1,9 +1,29 @@
 package net.fabricmc.fabric.api.loot.v1.visitor;
 
-public abstract class LootTableVisitor {
-	abstract void visitTable(LootNode.LootTableNode table);
+public interface LootTableVisitor {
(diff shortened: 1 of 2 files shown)


fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/visitor/LootTableVisitor.java
```diff
@@ -0,0 +1,65 @@
+package net.fabricmc.fabric.api.loot.v1.visitor;
+
+public interface LootTableVisitor {
```
Juuxel, line 3: I think this needs some `default`s 😄

RedstoneParadox: Yes, definitely, once I add methods to nodes that allow getting their children, which will come once I add the dozens of different nodes required for all vanilla functions, loot conditions, and whatever else I deem necessary.

RedstoneParadox:
If anyone wants to take over this, be my guest. I no longer have an interest in continuing it.

PR #1338, opened 2021-02-27
Buildscript updates

modmuss50:
* Upgrading to loom 0.6 and the new run config system
* Fix some of the gradle 7 deprecation warnings, there are more left in loom. Once complelte I want to fail the build on deprecations.
* Start looking at singing the jars - currently just pgp. jar singing is TODO
* Build reproducable jars
* Build testmod jars - a fat jar depends on some loom work


modmuss50:
This PR is currenly locked behind some loom issues.

build.gradle
```diff
@@ -112,18 +114,25 @@ allprojects {
...
-		classpath sourceSets.testmod.runtimeClasspath
-	}
-
-	task runTestmodServer(type: RunServerTask) {
-		classpath sourceSets.testmod.runtimeClasspath
+	loom {
+		runs {
+			testmodClient {
+				client()
+				ideConfigGenerated false
```
i509VCB: Why not generate an ide run config?

modmuss50: Becuase you would have over 100 of them, should just do that for the root project thinking about it.

modmuss50: Done.

liach:
what's the lowest gradle supported after this update? imo feel free to bump

modmuss50:
Whatever the wrapper is set to is the only "supported" version. I dont see why you would want to try and use it on anything else. This does include most of the required changes for 7 tho.

liach:
Can we use a gradle exec to call [jdk's jarsigner](https://docs.oracle.com/en/java/javase/15/docs/specs/man/jarsigner.html) to sign jars? A side note: If we do so, it should in an extra `doLast` for a jar task since this tool seems to change jar files in-place

modmuss50:
No, the jars are signed on a remote server, my plugin does not change the jars in place.

liach:
After looking at your [remote sign](https://github.com/modmuss50/RemoteSign/blob/master/src/main/java/me/modmuss50/remotesign/RemoteSignExtension.java) I have a few comments:
1. The 3 string fields should probably be made task-specific. Can keep it in extension so task can query from extension in case the string properties aren't set.
2. Consider merging https://github.com/modmuss50/RemoteSign/blob/e6913049148a71cb9c0c66d2c069625d4f57271d/src/main/java/me/modmuss50/remotesign/RemoteSignJarTask.java#L35-L57 and https://github.com/modmuss50/RemoteSign/blob/e6913049148a71cb9c0c66d2c069625d4f57271d/src/main/java/me/modmuss50/remotesign/RemoteSignExtension.java#L92-L114 to one common utility method (maybe merge the requirements for file, requestUrl, pgpKey to that new method as well)

modmuss50:
Uhh, I broke this fork.

PR #1339, opened 2021-02-27
Server only registry entries.

modmuss50:
This PR adds an API to allow registring an entry that will not be synced to the client. 

Is this something we want to support? Are there better solutions to this?

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryUtils.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryUtils.java
@@ -1,0 +1,27 @@
+package net.fabricmc.fabric.api.event.registry;
+
+import net.fabricmc.fabric.impl.registry.sync.RegistrySyncManager;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
+
+public class FabricRegistryUtils {
+	/**
+	 * Register an object with a given registry that will not be synced to any clients.
+	 * If this registry entry is present on the client it will be given a new id that will not match the server!
+	 *
+	 * @param registry The target registry
+	 * @param id the {@link Identifier} to be used to register the entry
+	 * @param entry the registry entry to be added to the target registry
+	 * @param <T> The type of the registry/entry
+	 * @return Returns the input entry
+	 */
+	public static <T> T registerServerOnlyRegistryEntry(Registry<T> registry, Identifier id, T entry) {
+		if (!RegistryAttributeHolder.get(registry).hasAttribute(RegistryAttribute.SYNCED)) {
+			throw new UnsupportedOperationException("Cannot register a server only entry to a none synced registry.");
+		}
+
+		T result = Registry.register(registry, id, entry);
+		RegistrySyncManager.addServerOnlyRegistryId(registry, id);
+		return result;
+	}
+}
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
@@ -20,6 +20,9 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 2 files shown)


Technici4n:
Imo it will just cause issues with anyone relying on ID sync. Why use synced registries at all if the entries are going to be server-only?

modmuss50:
The usecase I was given was server side only entities.

YTG1234:
Another usecase is for something like [PolyMc](https://github.com/TheEpicBlock/PolyMc). The way it currently does things is just [disabling registry sync as a whole](https://github.com/TheEpicBlock/PolyMc/blob/master/src/main/java/io/github/theepicblock/polymc/mixins/FabricRegistrySyncDisabler.java#L36).

Technici4n:
> The usecase I was given was server side only entities.

Ok that seems to make sense, but I hope nobody tries to sync entity types by id then...

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -304,4 +317,14 @@ public static void unmap() throws RemapException {
 	public static void bootstrapRegistries() {
 		postBootstrap = true;
 	}
+
+	public static void addServerOnlyRegistryId(Registry<?> registry, Identifier identifier) {
+		Set<Identifier> identifiers = SERVER_ONLY_REGISTRY_IDS.computeIfAbsent(registry, (r) -> new HashSet<>());
+
+		if (identifiers.contains(identifier)) {
+			throw new IllegalStateException(identifier + " has already been been marked as server only in " + registry);
```
liach, line 325: No need to be this harsh. Imo a log warning is enough as if an id is added twice there won't be any behavioral changes

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -53,6 +56,7 @@
 	static final Identifier ID = new Identifier("fabric", "registry/sync");
 	private static final Logger LOGGER = LogManager.getLogger("FabricRegistrySync");
 	private static final boolean DEBUG_WRITE_REGISTRY_DATA = System.getProperty("fabric.registry.debug.writeContentsAsCsv", "false").equalsIgnoreCase("true");
+	private static final Map<Registry<?>, Set<Identifier>> SERVER_ONLY_REGISTRY_IDS = new HashMap<>();
```
liach, line 59: Not a pro on registry stuff I am, but does this only work with non-dynamic registries?

liach (commented):
Two other problems:
1. Make sure vanilla registry works well with intermittent integer ids
2. License header

liach:
Another thing I'd like to see is a client-specific (represented by a predicate on Server Play Network Handler) rule for syncing registry entries. That is what polymc actually need, not one generalized rule like what we see here.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
i509VCB: Dynamic registries do not need this as the client will always have enough information to recreate the server's registry

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryUtils.java
```diff
@@ -0,0 +1,27 @@
...
+	 * If this registry entry is present on the client it will be given a new id that will not match the server!
+	 *
+	 * @param registry The target registry
+	 * @param id the {@link Identifier} to be used to register the entry
+	 * @param entry the registry entry to be added to the target registry
+	 * @param <T> The type of the registry/entry
+	 * @return Returns the input entry
+	 */
+	public static <T> T registerServerOnlyRegistryEntry(Registry<T> registry, Identifier id, T entry) {
+		if (!RegistryAttributeHolder.get(registry).hasAttribute(RegistryAttribute.SYNCED)) {
```
i509VCB, line 19: This may cause an issue if the registry is untouched before you try to add a server only entry. As adding any new entries will mark the registry as synced.

liach: Is fabric adding SYNCED flags this way? weird. Also worth discussing is whether the mark for server only and the registration should be bound together or separate.

i509VCB: Not nearly that simple, but yes. Other registries are marked as synced by default (which well requires an entry per registry) since some registries refer to entries in packets by raw id

liach: Hmm, Registry's codec functionality can store elements as int ids if it uses compression (which afaik minecraft doesn't yet use).

modmuss50:
Im going to close this as I dont think its something I want to support, atleast not now. The usecase I provided still required plently of hacks to make it work.

haykam821:
[Would this be a valid use-case?](https://github.com/NucleoidMC/plasmid/blob/91efd10afe0caf9ea2ce848e6e87abb257c616f3/src/main/java/xyz/nucleoid/plasmid/mixin/fake/RegistrySyncManagerMixin.java)

PR #1343, opened 2021-03-04
Add FabricItemSettings.fuel() as an alternative for FuelRegistry

YanisBft:
This is just more handy to be able to register an item as fuel directly in its builder

### DIFF
--- a/fabric-item-api-v1/build.gradle
+++ b/fabric-item-api-v1/build.gradle
@@ -2,5 +2,8 @@ archivesBaseName = "fabric-item-api-v1"
 version = getSubprojectVersion(project, "1.2.1")
 
 moduleDependencies(project, [
-		'fabric-api-base'
+		'fabric-api-base',
+		'fabric-content-registries-v0',
+		'fabric-lifecycle-events-v1',
+		'fabric-resource-loader-v0'
 ])
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
@@ -61,7 +61,7 @@ public class FabricItemSettings extends Item.Settings {
 	 * @see net.fabricmc.fabric.api.registry.FuelRegistry
 	 */
 	public FabricItemSettings fuel(int cookTime) {
-		FabricItemInternals.setFuelCookTime(cookTime);
+		FabricItemInternals.computeExtraData(this).fuel(cookTime);
 		return this;
 	}
 
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
@@ -53,6 +53,18 @@ public class FabricItemSettings extends Item.Settings {
 		return this;
 	}
 
+	/**
+	 * Sets this item as a fuel, with a provided {@code cookTime}.
+	 *
+	 * @param cookTime the number of ticks before this item will be consumed in a furnace
+	 * @return this builder
+	 * @see net.fabricmc.fabric.api.registry.FuelRegistry
+	 */
+	public FabricItemSettings fuel(int cookTime) {
+		FabricItemInternals.setFuelCookTime(cookTime);
+		return this;
+	}
+
 	// Overrides of vanilla methods
 
 	@Override
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
@@ -54,14 +54,14 @@ public class FabricItemSettings extends Item.Settings {
 	}
 
 	/**
-	 * Sets this item as a fuel, with a provided {@code cookTime}.
+	 * Sets this item as a fuel, with a provided {@code burnTime}.
 	 *
-	 * @param cookTime the number of ticks before this item will be consumed in a furnace
+	 * @param burnTime the number of ticks before this item will be consumed in a furnace fuel slot
 	 * @return this builder
 	 * @see net.fabricmc.fabric.api.registry.FuelRegistry
 	 */
-	public FabricItemSettings fuel(int cookTime) {
-		FabricItemInternals.computeExtraData(this).fuel(cookTime);
+	public FabricItemSettings fuel(int burnTime) {
+		FabricItemInternals.computeExtraData(this).fuel(burnTime);
 		return this;
 	}
 
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
@@ -56,7 +56,6 @@ public class FabricItemSettings extends Item.Settings {
 	/**
 	 * Sets this item as a fuel, with a provided {@code burnTime}.
 	 *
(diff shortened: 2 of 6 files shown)


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
```diff
@@ -18,13 +18,15 @@
...
 
+import net.fabricmc.fabric.api.registry.FuelRegistry;
 import net.minecraft.item.Item;
 
 import net.fabricmc.fabric.api.item.v1.CustomDamageHandler;
 import net.fabricmc.fabric.api.item.v1.EquipmentSlotProvider;
 
 public final class FabricItemInternals {
 	private static final WeakHashMap<Item.Settings, ExtraData> extraData = new WeakHashMap<>();
+	private static int fuelCookTime = -1;
```
RedstoneParadox: This should be stored in an an `ExtraData` instance so that it doesn't get applied to all items created after it. See `FabricItemSettings#equipmentSlot()`

RedstoneParadox (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
YanisBft: Done!

Technici4n (approved):
Looks easy enough, although I haven't tested it yet. Will probably do so this weekend.

fabric-item-api-v1/build.gradle
```diff
@@ -2,5 +2,8 @@ archivesBaseName = "fabric-item-api-v1"
 version = getSubprojectVersion(project, "1.2.1")
 
 moduleDependencies(project, [
-		'fabric-api-base'
+		'fabric-api-base',
+		'fabric-content-registries-v0',
+		'fabric-lifecycle-events-v1',
+		'fabric-resource-loader-v0'
```
Juuxel, line 8: Why does this need resource loader and lifecycle events?

fabric-item-api-v1/build.gradle
```diff
@@ -2,5 +2,8 @@ archivesBaseName = "fabric-item-api-v1"
 version = getSubprojectVersion(project, "1.2.1")
 
 moduleDependencies(project, [
-		'fabric-api-base'
+		'fabric-api-base',
+		'fabric-content-registries-v0',
```
Juuxel, line 6: This dependency should be added in the fabric.mod.json (and the others if they're needed).

fabric-item-api-v1/build.gradle
Technici4n: Content registries depends on it.

YanisBft: Yeah they are not applied recursively

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -53,6 +53,18 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
...
 	}
 
+	/**
+	 * Sets this item as a fuel, with a provided {@code cookTime}.
+	 *
+	 * @param cookTime the number of ticks before this item will be consumed in a furnace
+	 * @return this builder
+	 * @see net.fabricmc.fabric.api.registry.FuelRegistry
+	 */
+	public FabricItemSettings fuel(int cookTime) {
```
Juuxel: Hmm, the content registry module calls it "fuel burn time" which might be better (it's not only used for cooking after all).

Technici4n: Oops yeah `cook time` is for furnace inputs, not furnace fuels.

YanisBft: The module calls it "cook time" actually https://github.com/FabricMC/fabric/blob/f9b3753bc0fc312e5dd20036df77d4d31f01480b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FuelRegistryImpl.java#L67

Technici4n: It should probably be changed in the module then imo, yarn calls it `burnTime` if you look into `AbstractFurnaceBlockEntity`. Out of scope for this PR obviously, but feel free to open another one.

fabric-item-api-v1/build.gradle
liach: I think they are recursive for compilation but not for runtime. So if you don't declare, it crashes your test mod (that's how it was when I last tested)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
Juuxel: https://github.com/FabricMC/fabric/blob/f9b3753bc0fc312e5dd20036df77d4d31f01480b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FuelRegistry.java#L23

The *API* of that module calls it "fuel burn time", though. Not sure why it's different in the impl :eyes:

YanisBft: Renamed it

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -54,14 +54,14 @@ public FabricItemSettings customDamage(CustomDamageHandler handler) {
 	}
 
 	/**
-	 * Sets this item as a fuel, with a provided {@code cookTime}.
+	 * Sets this item as a fuel, with a provided {@code burnTime}.
 	 *
-	 * @param cookTime the number of ticks before this item will be consumed in a furnace
+	 * @param burnTime the number of ticks before this item will be consumed in a furnace fuel slot
```
Technici4n: A bit misleading - the item is consumed directly if it's the only one in the slot. Imo you can just omit this `@param` as it's obvious.

Technici4n (approved):

Mysterious-Dev (approved):

liach (approved):
Looks fantastic. Assuming mojang isn't going to add fuel registry to data packs soon:tm:, this would make stuff convenient as usual, just like how we have entity utility registrations in entity type builder.

Juuxel (approved):

BoogieMonster1O1 (approved):

LambdAurora (commented):
From the Discord, Prospector said:

> the fact that it adds 3 module dependencies to item-api is kinda unfortunate
> the point of modularization is kinda lost if loads of modules just depend on one another

which I agree on

I would prefer a `then(Consumer<Item> handler)` instead, which is called at the end of the item constructor, it could be used to still store Fuel stuff, or more in the settings without hard-depending on the module.
same for `FabricBlockSettings` but that's a bit unrelated to this PR.

OroArmor:
This should be reconsidered with the new custom item settings API that Haven King is working on. It might not need a huge change, but a rework of the backend would align the two

frqnny:
This and any future systems of item settings should also have a way to add to the composter as well, which is also a list if I recall correctly.

Mysterious-Dev:
This PR is dead ?

Technici4n:
I think so. If we add this feature, we will likely have to support 1000 similar ones. I think it would be better to find a more general way of doing this?

YanisBft:
I'm closing this PR then, and will just hope Mojang do it themselves (even though I still think this is handy for modders, more than having a separate instruction)

PR #1351, opened 2021-03-08
Add `ItemApiLookup`

Technici4n:
This PR extends #1234 to Item-provided APIs. The only (rather straightforward) API addition is [`ItemApiLookup`](https://github.com/FabLabsMC/fabric/blob/item-api-lookup/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemApiLookup.java), there you can also find example code.

I decided to only have `Item` as the mandatory first parameter of queries, and let other information such as NBT or count be provided by the generic context `C`. I am pretty sure most APIs will be able to find an expressive context that suits them better.

There is no `ItemApiCache` because there are no expensive block state and block entity queries to cache.


Technici4n:
Let me squash this somehow, this includes all the commits from 1234.

PR #1352, opened 2021-03-08
Add ItemApiLookup

Technici4n:
This PR extends #1234 to Item-provided APIs. The only (rather straightforward) API addition is [`ItemApiLookup`](https://github.com/FabLabsMC/fabric/blob/item-api-lookup/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemApiLookup.java), there you can also find example code.

~~I decided to only have `Item` as the mandatory first parameter of queries, and let other information such as NBT or count be provided by the generic context `C`. I am pretty sure most APIs will be able to find an expressive context that suits them better.~~ The queries now have an `ItemStack` parameter.

There is no `ItemApiCache` because there are no expensive block state and block entity queries to cache.

### DIFF
--- a/fabric-api-lookup-api-v1/README.md
+++ b/fabric-api-lookup-api-v1/README.md
@@ -4,7 +4,7 @@ This is useful when the same API could be implemented more than once or implemen
 See also the [package-info.java file](src/main/java/net/fabricmc/fabric/api/lookup/v1/package-info.java).
 
 * What we call an API is any object that can be offered or queried, possibly by different mods, to be used in an agreed-upon manner.
-* This module allows flexible retrieving of such APIs from blocks in the world, represented by the generic type `A`.
+* This module allows flexible retrieving of such APIs, represented by the generic type `A`, from blocks in the world or from items.
 * It also provides building blocks for defining custom ways of retrieving APIs from other game objects.
 
 # Retrieving APIs from blocks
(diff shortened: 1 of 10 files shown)


liach:
just wondering, how often are item instances used without stacks?

Technici4n:
That's a very good question. Originally I planned to use `ItemStack`, then `ItemKey`, then back to `Item`. I am still not totally sold on `Item`.

The problem with `ItemStack` is that mutation is hard to control. For a fluid API, sometimes mutating the stack directly may be acceptable, and sometimes it's not. It's acceptable when you are filling a tank item in the player's hand by editing its NBT. It is not acceptable when you need to replace a full bucket by an empty bucket (can't change the item of a stack). So `ItemStack` is basically unsuitable for the item/fluid transfer API.

However, `ItemStack` is very nice for something like TR energy which can safely assume that all energy holders can store their data in their NBT if they want to, and with the current state either the query has to take `(Item, ItemStack)` or something like `(Item, NbtAndCount)`, in both cases it's not very satisfying... (the energy stored in the NBT is per-item and has to be divided by the stack count).

I guess the limitation of `ItemStack` can be worked around by passing a copy of the original stack to the API query. I expect that people will be using helpers that directly get the API such as `FluidTransfer.getFromPlayerHand(Hand)` or `FluidTransfer.getFromPlayerCursor()` anyway, so the copy can be done transparently for 99% of the users and the javadoc will take care of the remaining 1%.

All in all I am considering moving back to `ItemStack`. I would appreciate any input.

Technici4n:
I think `ItemStack`s are a better choice, and I've updated the PR accordingly.

liach:
imo it will be safe if we add contract that the api provider should not mutate the input item stack.

Technici4n:
It depends on the API, the energy API will probably mutate the stack directly. I will add a note in the javadoc.

EDIT: Done.

liach:
Ah, I meant that we need to make sure the provider doesn't change the input stack when it returns the API.

I didn't want to add contract on the returned API.

Technici4n:
That's not a bad idea. Should we enforce it?

Technici4n:
I would say yes if it's practical to do so. Comparing `Item` and the tag hashcode before/after should do.

liach:
> 
> 
> I would say yes if it's practical to do so. Comparing `Item` and the tag hashcode before/after should do.

Not enforcing by checks in code. Imo you should add this check to assert, or it will hurt performance a bit. I raised this point because I thought that's what made you initially choose Item than ItemStack for the api lookup target.

Technici4n:
Might as well leave it out if it's an assert, nobody ever enables them anyway. :-P
The main problem with `ItemStack` is that it does not fit APIs where mutation is restricted (think `canSetStack(Items.BUCKET)`), but in those cases we can just pass a copy.

PR is updated with the additional remark BTW.

liach (approved):

RedstoneParadox:
What about situations in which you don't care about the stack size or NBT?

Technici4n:
You can always pass `new ItemStack(item)`. :-P

Juuxel (approved):

GabrielOlvH (approved):

maximumpower55:
I really hope this can be merged soon
I cant update my mods to 1.17 until this gets merged

Technici4n:
Just rebased, and bumped version to `1.2.0` as #1486 was merged with version `1.1.0`.

Player3324 (approved):
LGTM, thanks!

modmuss50 (approved):

Issue #842, opened 2020-06-25
Attribute#mergeAttributes doesn't respect modifier UUIDs


bluebear94:
[Attribute#mergeAttributes](https://github.com/FabricMC/fabric/blob/686e6e57ad66a76aa674f622b0492dda402ea0e3/fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/AttributeManager.java#L40) completely ignores the UUIDs of the modifiers in the input multimaps, causing problems for code that depends on particular UUIDs (such as `Item.ATTACK_DAMAGE_MODIFIER_ID`). I'm not sure what the intended behavior for this method is.

LemmaEOF:
I'll take a look when I can.

bluebear94:
Actually, since the purpose of `mergeAttributes` is to prevent duplicate entries in the tooltip, couldn't the current calls to that method be replaced with one that merges only the multimaps (using `ImmutableMultimap.Builder#putAll`), while handling duplicate entries when the tooltip is displayed?

bluebear94:
I don't mean to be rude, but is there any progress on this issue?

bluebear94:
It's been 3 weeks since this issue was opened; any response would be nice.

LemmaEOF:
#855 might be helping with it? I've taken a step back from modding for the time being due to personal circumstance.

bluebear94:
#855 concerns mining levels. It isn't relevant to attribute modifiers at all from what I can get from the code.

LemmaEOF:
They're both part of the same module.

shedaniel:
IMO they should become two different modules then.

Also yes #855 will not affect this.

LemmaEOF:
I mean, `AttributeManager` solely exists to help deal with Fabric Tool Attribute API, so it doesn't make sense to spin it out. We're not NPM.

shedaniel:
No, I mean that the mining level stuff is completely unrelated to tool attributes

LemmaEOF:
uhhh, no? the mining level stuff was literally one of the primary points of tool attributes

shedaniel:
The attributes mean the actual entity attributes, like attack speed and stuff. The mining speed and is effective is not really an attribute in my opinion.

shedaniel:
Anyway, I completely went off-topic here, sorry about that. As stated, #855 won't fix this issue and this issue should be looked into.

i509VCB:
[...]

PR #1355, opened 2021-03-09
Update Dynamic Attribute Tool modifier application to fix #1353

### LINKED ISSUE #1353: DynamicAttributeTool#getDynamicModifiers is bugged
It has come to my attention that the `getDynamicModifiers` method in `DynamicAttributeTool` is bugged. This is due to a UUID not being specified in `AttributeManager#mergeAttributes`. A random UUID is chosen when one is not specified, so the game fails to remove the attribute when needed. This results in the attribute being stacked[?] every time you select a `DynamicAttributeTool` in your hotbar.

---

**Context / Problem**:

`Item` has a method called `getAttributeModifiers`. This method returns the modifiers that should be added to the player/entity when the item is equipped and is called in `LivingEntity#method_30129`. Fabric provides `DynamicAttributeTool`, which has a `getDynamicModifiers` method. This method, unlike `getAttributeModifiers`, gives you context on `ItemStack` and the entity holding the item.

When the player switches an item around, the attribute modifiers for the previous stack and the new stack are calculated. The old modifiers are removed, and the new modifiers are added. An attribute is properly removed if the UUID of the attribute that is being removed matches an existing UUID (`Map<UUID, Modifier>`). The important thing to note is that **if the modifiers are the same, but the UUIDs are different, the old modifier will not be removed.**

When `getDynamicModifiers` is called, the results are merged with the result of `getAttributeModifiers`. This is done to avoid having multiple attributes of the same type in the tooltip (such as +4 Damage and +2 Damage on 2 separate lines). The problem is that the entity attribute modifiers created in the merge method **do not specify a UUID**, which means the method will not return the same results when called multiple times in a row. When the "remove old modifiers," logic runs, it fails, because the UUIDs are different.

---

**Reproducing**:
```java
public class TestItem extends Item implements DynamicAttributeTool {

    public static final UUID TEST_UUID = UUID.fromString("CB3F55D3-645C-4F38-A497-9C13A33DB5CF");

    public TestItem(Settings settings) {
        super(settings);
    }

    @Override
    public Multimap<EntityAttribute, EntityAttributeModifier> getDynamicModifiers(EquipmentSlot slot, ItemStack stack, @Nullable LivingEntity user) {
        if(slot.equals(EquipmentSlot.MAINHAND)) {
            ImmutableMultimap.Builder<EntityAttribute, EntityAttributeModifier> builder = ImmutableMultimap.builder();
            builder.put(EntityAttributes.GENERIC_MOVEMENT_SPEED, new EntityAttributeModifier(TEST_UUID, "TEST", -1.0, EntityAttributeModifier.Operation.ADDITION));
            return builder.build();
        } else {
            return EMPTY;
        }
    }
}
```

[Resulting - streamable](https://streamable.com/nolbtn)
*Note how the attribute persists despite me switching off the item.*

---

I will be looking into writing a fix for this soon, but I wanted to post an issue first for additional context when that PR arrives.

Draylar:
**Full Context:** read #1353
**tl;dr:** `DynamicAttributeTool#getDynamicModifiers` does not remove attributes properly. This fix makes it arguably cleaner than before while patching up bugs while still providing entity context.

[Before - streamable](https://streamable.com/nolbtn)
[🌟After ✨ - streamable](https://streamable.com/vanr6g)

---

This PR removes the current attribute-merging functionality of the API (which seems to boil down to supporting Entity context) and consolidates all logic for dynamic attributes into `Item#getDynamicAttributes`. The main issue is still entity context, so we inject into all the locations `getDynamicAttributes` is called at to set context on the stack before calling the method.

The only potential edge case I can think of with this API is someone calling `Item#getAttributeModifiers` and expecting to get entity-context-dependent dynamic attributes in return, but the caller is also free to go straight to `getDynamicModifiers` and pass in the entity themselves.

Most of the decisions here are the result of eliminating everything else, so if you have any particular questions about other approaches (or what I did), feel free to ask. 

---

Note: [almost nobody uses this particular functionality](https://github.com/search?q=getDynamicModifiers&type=code), and there are no breaking changes, so I think we can get this through faster, considering how the current version is not usable.


fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/api/tool/attribute/v1/ItemStackContext.java
```diff
@@ -0,0 +1,9 @@
+package net.fabricmc.fabric.api.tool.attribute.v1;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.entity.LivingEntity;
+
+public interface ItemStackContext {
+	void fabricToolAttributes_setContext(@Nullable LivingEntity contextEntity);
```
i509VCB: was this meant to be in api package?

Draylar: No, I just don't know where to put it lol

Technici4n: Put it in `impl` if it's just a duck interface.

Draylar:
For anyone interested in previewing isolated mechanics of this PR, or using the PR before it has been approved, please visit [Dynamic Attributes](https://github.com/Draylar/dynamic-attributes). This library extracts the core functionality from this PR for you to test, and works alongside current Fabric API (and yes, FAPI's Apache 2.0 license is properly attributed where needed). 

The example/test item of the library [assigns attack damage based on the user's Y-position](https://github.com/Draylar/dynamic-attributes/blob/master/src/main/java/draylar/da/DynamicAttributes.java). You can see it in-action [in this streamable clip](https://streamable.com/nslsrn). The main thing to point out here is that:
1. the attack damage is properly applied in the tooltip & when hitting the mob
2. the attack damage does _not_ persist when I switch to my hand, which is the main bug with the current API

frqnny (changes requested):
Shouldn't the test item classes be static classes in the test mod initializer so people don't have to look at 4 different classes to see the examples? It would make readability better. It's annoying- for me at least, to have to search through many different classes to make sure I did it right isn't fun at all.

Technici4n:
It's fine I think, the change is non-trivial so it should be looked at with an IDE anyway.

Technici4n:
I had a first look and this is looking pretty solid, thanks! I have to go now, but I will properly test this PR soon (hopefully tomorrow), and then it should be good to go.

Does removing (like you did) `AttributeManager#mergeAttributes` have any side-effect besides the tooltip? If it's the only side-effect then it should be fine to remove it.

Draylar:
> Does removing (like you did) `AttributeManager#mergeAttributes` have any side-effect besides the tooltip? If it's the only side-effect then it should be fine to remove it.

Temporary attributes, which is the type we are dealing with here, are reloaded every time you relog (or... something close to that), so there should be no chance for anything like "old attributes persisting on players when the system changes" happening, if that is what you mean. I think the merge functionality was mostly for entity context anyways, so removing it should not cause any other problems. I have tested the isolated library extensively in other projects, and it has worked fine so far.

Technici4n (approved):
Tested this (single player and dedicated server) and it's working. I also checked the impl and it seems fine. And the little bit of added javadoc can't hurt!

Technici4n:
Closes #842 as well.

Player3324 (approved):
LGTM, thanks!

PR #1358, opened 2021-03-10
Fix sendError discrepency with client commands

jpenilla:
This fixes a discrepancy between how error messages are formatted in client commands vs server commands.

### DIFF
--- a/fabric-command-api-v1/src/main/java/net/fabricmc/fabric/mixin/command/client/ClientCommandSourceMixin.java
+++ b/fabric-command-api-v1/src/main/java/net/fabricmc/fabric/mixin/command/client/ClientCommandSourceMixin.java
@@ -25,6 +25,7 @@ import net.minecraft.client.network.ClientCommandSource;
 import net.minecraft.client.network.ClientPlayerEntity;
 import net.minecraft.client.world.ClientWorld;
 import net.minecraft.network.MessageType;
+import net.minecraft.text.LiteralText;
 import net.minecraft.text.Text;
 import net.minecraft.util.Formatting;
 import net.minecraft.util.Util;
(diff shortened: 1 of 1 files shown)


i509VCB:
Could you attach a screenshot of before and after?

Technici4n:
What is the difference?
Edit: i5 was first.

jpenilla:
Yeah, give me a minute, will attach screenshots and test code

jpenilla:
```java
new LiteralText("Invalid Command Syntax. Correct command syntax is: ")
        .append(
                new LiteralText(String.format("/%s", e.getCorrectSyntax()))
                        .styled(style -> style.withColor(Formatting.GRAY))
        )
```
Before: image
After: image

i509VCB (approved):

LambdAurora:
I wonder if `shallowCopy` would be better.

It looks nicer than a `new LiteralText("").append()`.

Technici4n:
What is the problem? Does `copy()` not perform a deep copy?

LambdAurora:
> What is the problem? Does `copy()` not perform a deep copy?

It doesn't, it's even in [the javadoc](https://maven.fabricmc.net/docs/yarn-21w10a+build.1/net/minecraft/text/Text.html#copy()).

Technici4n:
Ohhhh nice javadoc link. 😄
I'm all for `.shallowCopy()` then.

jpenilla:
Using `shallowCopy()` would probably produce a visually similar result in the provided example, but would still be a discrepancy from `ServerCommandSource`. Instead of creating an empty component with a style and adding the message as a child, we would be modifying the style of the message, which with an example like 
```java
new LiteralText("Blue error")
        .styled(style -> style.withColor(Formatting.BLUE))
```
would produce a visually inconsistent result from `ServerCommandSource`.

Technici4n (approved):
Let's match vanilla behavior.

LambdAurora (approved):

liach (approved):
Looks good, based on the assumption that the top level literal text may be modified the rest of the message won't be modified.

modmuss50 (approved):

modmuss50:
Tiny bug fix, I dont see a reason to do a release just for it. Will pull it in with the next batch of changes. 👍

Juuxel (approved):

PR #1359, opened 2021-03-11
Add Support for Custom Item Settings

Haven-King:
Currently, if a mod wants to pass custom settings for items, the best way is to subclass either `Item.Settings` or `FabricItemSettings` and add the behavior there yourself. An example of this is in [The Aether](https://github.com/devs-immortal/The-Aether/blob/master/src/main/java/com/aether/items/AetherItemSettings.java), where custom item settings are used for a custom rarity and enchantment glint color. If another mod, however, adds an item setting for their items via the same method, there's no way for one item to share both properties.

This pull request solves that by allowing mods to attach arbitrary data to item settings. As part of this change, both the `EquipmentSlotProvider` and `CustomDamageHandler` have been migrated to use this system. Their public-facing API's remain unchanged.

This pull request also fixes a small bug in the testmod, wherein one of the test entrypoints would fail to launch.


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSettingType.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class CustomItemSettingType<T> {
+	private final Supplier<@NotNull T> defaultValue;
+
+	private CustomItemSettingType(Supplier<@NotNull T> defaultValue) {
+		this.defaultValue = defaultValue;
+	}
+
+	// There are a lot of these for convenience sake.
+	public static <T> CustomItemSettingType<T> of(Supplier<T> defaultValue) {
```
liach: Imo the supplier version should have a different name from the direct getting version, and create another 
`public static <T> CustomItemSettingType<T> of(T defaultValue) {`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSettingType.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * A type of setting that can be passed to item constructors.
+ * This feature can be used by mods to add non-standard settings
+ * to items in a way that is compatible with other mods that add
+ * settings to items.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public final class CustomItemSettingType<T> {
```
liach: Maybe just `CustomItemSetting`

liach:
Pretty good. I assume these settings are all used by other vanilla items in modified bytecode, right? If those are exclusively used by mods mods should just add extra constructor parameters.

Haven-King:
> Pretty good. I assume these settings are all used by other vanilla items in modified bytecode, right? If those are exclusively used by mods mods should just add extra constructor parameters.

Adding constructor parameters isn't an elegant solution when you want to be able to add settings to a wide range of item types, potentially from different mods, or if you want to ensure a large group of items have the same settings.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
```diff
@@ -16,42 +16,35 @@
...
+import net.fabricmc.fabric.api.item.v1.CustomItemSettingType;
 import net.fabricmc.fabric.api.item.v1.CustomDamageHandler;
 import net.fabricmc.fabric.api.item.v1.EquipmentSlotProvider;
 
 public final class FabricItemInternals {
-	private static final WeakHashMap<Item.Settings, ExtraData> extraData = new WeakHashMap<>();
+	public static final CustomItemSettingType<EquipmentSlotProvider> EQUIPMENT_SLOT_PROVIDER = CustomItemSettingType.of(() -> null);
+	public static final CustomItemSettingType<CustomDamageHandler> CUSTOM_DAMAGE_HANDLER = CustomItemSettingType.of(() -> null);
+
+	private static final Map<Item.Settings, Map<CustomItemSettingType<?>, Object>> customSettings = new WeakHashMap<>();
```
liach: Why not just add a field to `Item.Settings` though mixin?

Haven-King: That would be my first approach as well. This is how it was done [prior to this PR](https://github.com/FabricMC/fabric/blob/616c01224d0b1009f7a635311f87ce2f3bb50de0/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java) (with `ExtraData` instead of `Map<CustomItemSettingType<?>, Object>`, but still) so I chose to stick with that.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSettingType.java
Haven-King: The reason I use a supplier for non-primitive types is to ensure safety. Primitive types are passed by value, so it's safe to offer the same one to every item. Items might mutate the resulting value though, which would cause problems if we passed the same `ArrayList` for example to every item.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
liach: Imo you should follow what block settings has https://github.com/FabricMC/fabric/blob/1e6c71f9bcb3fe386d9fa8030d061b37b9afe31e/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/AbstractBlockSettingsMixin.java#L30

Also would be good to add this to block settings

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * A type of setting that can be passed to item constructors.
+ * This feature can be used by mods to add non-standard settings
+ * to items in a way that is compatible with other mods that add
+ * settings to items.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public final class CustomItemSetting<T> {
+	private final Supplier<@NotNull T> defaultValue;
```
liach:
```suggestion
	private final Supplier<T> defaultValue;
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,99 @@
...
+
+	private CustomItemSetting(Supplier<@NotNull T> defaultValue) {
```
liach:
```suggestion
	private CustomItemSetting(Supplier<T> defaultValue) {
```

liach (commented):
https://github.com/FabricMC/fabric/pull/1359/files#diff-9e98586cb53b15bc81915c0e3bf96ce908536638e4e075730d83a9bf53153404R29-R30

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSettingType.java
liach: I mean you can have a default value like a collection empty list. This isn't really a problem, as there are way more ways for bad code to break things, especially for minecraft where there's no strict scrutiny to prevent some other parts of code from intentionally breaking stuff.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
Haven-King: I've changed it to follow this. And I agree that it would be good to add something similar to block settings, but as that's a different module, it would be a different PR.

liach (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSettingType.java
Haven-King: That's absolutely true. I still think it's worth providing that safety where we can.

LambdAurora (approved):

Technici4n:
So I thought of another way of doing this, based on one of Lambda's idea.

Let's say we add `.then(Consumer<Item>... itemConsumers)` to `FabricItemSettings` that registers consumers to be invoked at the end of the item constructor, then we could achieve exactly the same thing, and we could replace `ExtraData` with `List<Consumer>` basically.

For example
```java
public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvider) {
    FabricItemInternals.computeExtraData(this).equipmentSlot(equipmentSlotProvider);
}
```
becomes
```java
public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvider) {
    return then(item -> ((ItemExtensions) item).fabric_setEquipmentSlotProvider(equipmentSlotProvider);
}
```

By separating the `Consumer<Item>` from the item settings, this can then be trivially expanded upon by other mods, for example:
```java
public static Consumer<Item> fuel(int burnTicks) {
    return item -> FuelRegistry.add(item, burnTicks);
}

// used like this:
.then(fuel(42))
```

The only downside is that it's not possible to get the current setting as it's stored within the lambda, but the current PR doesn't provide that option either.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+
+	public static CustomItemSetting<Double> of(double defaultValue) {
+		return new CustomItemSetting<>(() -> defaultValue);
+	}
+
+	public static CustomItemSetting<Boolean> of(boolean defaultValue) {
+		return new CustomItemSetting<>(() -> defaultValue);
+	}
+
+	public static CustomItemSetting<Byte> of(byte defaultValue) {
```
i509VCB: I don't think byte, short and char are needed here.

LambdAurora:
@Technici4n imo, the consumers being executed at the end of the constructor and not after item registration is a bit painful as it would prevent calling anything that relies on the item to be in the register.
Other issue is fake items stuff, it's possible to instantiate an item without registering it, not recommended but the consumers being executed at constructor end would very likely cause issues in that case.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * A type of setting that can be passed to item constructors.
+ * This feature can be used by mods to add non-standard settings
+ * to items in a way that is compatible with other mods that add
+ * settings to items.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public final class CustomItemSetting<T> {
```
i509VCB: So this is compared by identity?

Technici4n:
If you register a `then()` for your own fake `Item` then that's not my problem. ^^

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemMixin.java
```diff
@@ -24,41 +28,28 @@
...
 import net.fabricmc.fabric.impl.item.FabricItemInternals;
 import net.fabricmc.fabric.impl.item.ItemExtensions;
+import net.fabricmc.fabric.api.item.v1.CustomItemSetting;
 
 @Mixin(Item.class)
-abstract class ItemMixin implements ItemExtensions {
-	@Unique
-	private EquipmentSlotProvider equipmentSlotProvider;
-
+class ItemMixin implements ItemExtensions {
```
i509VCB: Where abstract

Haven-King:
I agree that `.then` would be a good feature, but I don't see how it relates directly to this PR. I suppose it could replace the intermediary step of copying the map in the constructor, but that has its own issues. The most important issue I see is that when `then` is fired you expect the item to be fully initialized, which might not be the case if custom item settings are being copied via the same method.

LambdAurora:
> If you register a `then()` for your own fake `Item` then that's not my problem. ^^

Can't wait to have to strip every consumers from the re-usable settings for the fake item :|

Plus it's mutable and there's no way to copy an item settings properly so really not happy about that.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+	public static CustomItemSetting<Character> of(char defaultValue) {
+		return new CustomItemSetting<>(() -> defaultValue);
+	}
+
+	public static CustomItemSetting<String> of(String defaultValue) {
+		return new CustomItemSetting<>(() -> defaultValue);
+	}
+
+	@ApiStatus.Internal
+	public T getDefaultValue() {
```
i509VCB: I wouldn't mark this as Internal. If you need to hide impl details mark an interface as non extendable.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ItemExtensions.java
```diff
@@ -16,12 +16,13 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import net.fabricmc.fabric.api.item.v1.CustomItemSetting;
 
 public interface ItemExtensions {
-	/* @Nullable */ EquipmentSlotProvider fabric_getEquipmentSlotProvider();
-	void fabric_setEquipmentSlotProvider(EquipmentSlotProvider equipmentSlotProvider);
-	/* @Nullable */ CustomDamageHandler fabric_getCustomDamageHandler();
-	void fabric_setCustomDamageHandler(CustomDamageHandler handler);
+	@NotNull Map<CustomItemSetting<?>, Object> fabric_getCustomItemSettings();
```
i509VCB: NotNull is redundant

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -49,10 +56,27 @@ public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvi
...
+	 * @param value the object containing the setting itself
+	 * @return this builder
+	 */
+	public <T> FabricItemSettings customSetting(CustomItemSetting<T> setting, T value) {
+		this.customSettings.put(setting, value);
 		return this;
 	}
 
+	@ApiStatus.Internal
+	public Map<CustomItemSetting<?>, ?> getCustomSettings() {
```
i509VCB: This method is not very useful for actual api consumers.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class CustomItemSetting<T> {
+	private final Supplier<T> defaultValue;
+
+	private CustomItemSetting(Supplier<T> defaultValue) {
+		this.defaultValue = defaultValue;
+	}
+
+	// There are a lot of these for convenience sake.
+	public static <T> CustomItemSetting<T> of(Supplier<T> defaultValue) {
```
i509VCB: of implies that the implementation will give you the most applicable previous instance if present. This should probably be `create`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/**
+	 * Returns the current value of this setting for the given {@link Item}.
+	 * Should only be called after or within item construction.
+	 *
+	 * @param item the item
+	 * @return the current setting if present, the default setting if not
+	 */
+	public T getValue(Item item) {
```
i509VCB: This method should be mentioned at the top level class javadoc so people providing such a system should be told how to get the value?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,98 @@
...
+ * to items in a way that is compatible with other mods that add
+ * settings to items.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public final class CustomItemSetting<T> {
+	private final Supplier<T> defaultValue;
+
+	private CustomItemSetting(Supplier<T> defaultValue) {
+		this.defaultValue = defaultValue;
```
i509VCB: `Objects.requireNonNull(...)`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
i509VCB: Objects.requireNonNull

i509VCB (commented):
With regards to new item settings added by fabric api, those should still recieve dedicated setting methods.

haykam821:
For fake items, there can just be a `FabricItemSettings.done(Item)` to call when necessary.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
liach: Why? This isn't really beneficial and we aren't facing some malicious users like the jdk is

i509VCB: Fail fast with an error you will be told about explicitly

LambdAurora:
@haykam821 it still doesn't solve the issue of stuff being called in then relying on the item being in the registry.
If you don't guarantee that the item is in the registry for `then`, then there is no benefits to add `then`.

i509VCB:
.then has the issue of delaying the exposure of errors. If there is an error you should get it as you try to register an item setting.

Technici4n:
So, where this PR has value (which I overlooked the first time) is that it allows retrieving the setting from the item in a standard way.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Technici4n: Yup, agree with i5 here.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Technici4n: I am not sure those are needed at all, all you save is `() ->`.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Technici4n: Yup, interface would be my suggestion as well. The constructor could then be put under `impl` too.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
Technici4n: I believe that was the reasoning for external storage of the extra data?

Technici4n (commented):
Ok after seeing the testmod I must say I like this addition more than I thought. 😄

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: Yep. Mods that want to allow their setting to be be used should expose it publicly.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -39,7 +39,7 @@
 	 * @return this builder
 	 */
 	public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvider) {
-		FabricItemInternals.computeExtraData(this).equipmentSlot(equipmentSlotProvider);
+		((CustomItemSettingImpl<EquipmentSlotProvider>) CustomItemSettingImpl.EQUIPMENT_SLOT_PROVIDER).set(this, equipmentSlotProvider);
 		return this;
```
Technici4n:
```suggestion
		return customSetting(CustomItemSettingImpl.EQUIPMENT_SLOT_PROVIDER, equipmentSlotProvider);
```
I believe this would work too?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -49,7 +49,19 @@ public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvi
 	 * @see CustomDamageHandler
 	 */
 	public FabricItemSettings customDamage(CustomDamageHandler handler) {
-		FabricItemInternals.computeExtraData(this).customDamage(handler);
+		((CustomItemSettingImpl<CustomDamageHandler>) CustomItemSettingImpl.CUSTOM_DAMAGE_HANDLER).set(this, handler);
+		return this;
```
Technici4n: Same thing.

Technici4n (commented):
New impl is looking pretty nice overall, will take a closer look at some point!

I would consider removing all the `create()` overloads as imo they are not very useful and clutter the API surface a lot. Is `CustomItemSetting.create(42)` really such a big improvement over `CustomItemSetting.create(() -> 42);` that it warrants "polluting" the API?

OroArmor (approved):
really good simplification and expansion of the item settings API. 10/10 should be merged.

Implementing the things Technici4n mentioned are the only changes I recommend

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	T getValue(Item item);
+
+	/**
+	 * Creates a new CustomItemSetting with the given default value.
+	 *
+	 * @param defaultValue the value all items that do not explicitly set this setting will have.
+	 * @return a new CustomItemSetting
+	 */
+	static <T> CustomItemSetting<T> create(Supplier<T> defaultValue) {
```
Technici4n, line 73: I wonder if there would be a use for an `Identifier` and a `Class<T>` parameter that would allow retrieving the unique instance associated with a given id, just like BlockApiLookup#get but for item settings.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,57 @@
...
+
+import net.fabricmc.fabric.impl.item.CustomItemSettingImpl;
+
+/**
+ * A type of setting that can be passed to item constructors.
+ * This feature can be used by mods to add non-standard settings
+ * to items in a way that is compatible with other mods that add
+ * settings to items.
+ *
+ * <p>Values of this setting can be retrieved from an item using {@link CustomItemSetting#getValue(Item)}</p>
```
Technici4n: `</p>` can be omitted.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param <T> the type of the setting to be attached
+ */
+@ApiStatus.NonExtendable
```
Technici4n: I would mention in the javadoc something along the lines of `Do not implement. Use {@link #create} to retrieve instances.`.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Technici4n: (It is possibly overkill.)

Haven-King: This doesn't seem very likely to me. I would want some concrete examples before implementing that kind of behavior.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: It *can*, but should it?

Technici4n: Yes :P

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Technici4n: Yeah, it's probably overkill.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: But... Why?

Technici4n: I would suggest adding, in the javadoc:
- Something along the lines of "Values are directly stored inside this `CustomItemSetting`, per instance. Users that wish to expose a custom setting may do so by exposing the `CustomItemSetting` instance, for example in a `public static final` field."
- Maybe a little example wrapped in `<pre>{@code ... }</pre>` to follow that up and make usage a bit more obvious, a `CustomItemSetting<String>` like the testmod would do I think.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -30,7 +30,26 @@
  * to items in a way that is compatible with other mods that add
  * settings to items.
  *
- * <p>Values of this setting can be retrieved from an item using {@link CustomItemSetting#getValue(Item)}</p>
+ * <p>Values of this setting can be retrieved from an item using {@link CustomItemSetting#getValue(Item)}
+ *
+ * <p>Users that wish to expose a custom setting for use in other mods should do so by exposing
+ * the CustomItemSetting instance.
```
Technici4n:
```suggestion
 * <p>Values of this setting can be retrieved from an item using {@link CustomItemSetting#getValue(Item)},
 * and users that wish to expose it for use in other mods should do so by exposing the CustomItemSetting instance.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -30,7 +30,26 @@
...
+ * @Override
+ * public void onInitializeClient() {
+ *     ItemTooltipCallback.EVENT.register((stack, context, lines) -> {
+ *         String tooltip = CUSTOM_TOOLTIP.getValue(stack.getItem());
+ *
+ *         if (tooltip != null) {
+ *             lines.add(new LiteralText(tooltip));
+ *         }
+ *     }
+ * }}</pre>
```
Technici4n, line 50: Thanks for adding this. Might want to explain a little bit what's going on here with a few comments, and show how it is used with the item settings, i.e. `// Create the setting instance`, `// Register a setting for an item` and finally `// Use the custom setting`. And possibly prefix with `<h3>Usage Example</h3>`.

Then we can also link to this from `FabricItemSettings.custom()`: `See {@link CustomItemSetting} for a full usage example.`

Yes, writing javadoc is pain, but it is the key to good API use by the community. Don't compromise on quality here because people are likely to copy/paste even if they shouldn't. 😄

Mysterious-Dev (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>Do not implement. Use {@link CustomItemSetting#create(Supplier)} to retrieve an instance.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+@ApiStatus.NonExtendable
```
Juuxel: Why is this an interface if it shouldn't be implemented by anything else than the FAPI implementation? It think it would make more sense to have it as a normal class.

Technici4n: See the history. If it's a class, implementation details leak. I think interfaces should always be the default tbh.

liach: Nope. In jdk's libraries, they always hide implementation details by using package-private methods. Consequently they have some package access, like JavaLangAccess etc. to access those package private methods. Imo using interface is more dangerous because people can intentionally create unsuitable implementations and break stuff even though you have that non extendable annotation there.

Technici4n: I would say it is more readable to only have the relevant methods in the public API, and keep the rest under the `impl` package. I don't think we want `FabricItemApiAccess` :P

Technici4n: I can tolerate another choice, this is not very important. It's just how I think about APIs.

liach: Another trick is like:
In the impl package utility class, you hold a `Setter` functional interface instance that has `set(CustomItemSettings type, Item.Settings settings, T value)`. Then in the static block of `CustomItemSettings`, you set that setter to a method reference to the package-private `set` function in `CustomItemSettings`.

Since the `set` in `CustomItemSettings` is not exposed, when you want to set, you will call the impl class's Setter instead. Imo this is a better way.

Technici4n: Nice one. 😄

liach: Actually, this is exactly how java's sharedsecrets/jla access etc. works. in that impl class, you might need an illegal state check in the setter's getter method to ensure that the setter is there (i.e. `CustomItemSettings` is loaded) when the static setter field is used. I think that should be safe as when you call that setter, most likely you already have a `CustomItemSettings`, which has already static initialized and the setter should be there.

If you try to classload `CustomItemSettings` in the impl class, I guess it might cause classloading lock. Not really sure on that part though.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/CustomItemSettingImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+	@Override
+	public T getValue(Item item) {
+		Objects.requireNonNull(item);
+
+		return this.customItemSettings.computeIfAbsent(item, i -> this.defaultValue.get());
+	}
+
+	public void set(Item.Settings settings, T value) {
+		this.customSettings.put(settings, value);
```
liach, line 60: Should require the settings to be not null as null key is supported by weak hashmap. Also `@Override` annotation

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/CustomItemSettingImpl.java
```diff
@@ -0,0 +1,71 @@
...
+		CUSTOM_SETTINGS.computeIfAbsent(settings, s -> new HashSet<>()).add(this);
+	}
+
+	public void build(Item.Settings settings, Item item) {
+		this.customItemSettings.put(item, this.customSettings.getOrDefault(settings, this.defaultValue.get()));
```
liach, line 67: Same null check and override annotation

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/CustomItemSettingImpl.java
liach: I mean `Objects.requireNonNull` call.

LambdAurora:
Uh... The usual rule of thumb in Fabric API is that parameters shouldn't be null unless `@Nullable` is specified, we don't really use `@NotNull` annotations here.

liach:
Also I meant to add `Objects.requireNonNull` calls than `@NotNull` annotations.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Signals to this CustomItemSetting to associate its value in the {@link Item.Settings} instance with an Item.
+	 *
+	 * <p>Should really only be called by the implementation. Is exposed for people who need highly specialized settings
+	 *
+	 * @param settings the item settings to draw a value from
+	 * @param item the item to have its value set
+	 */
+	void build(Item.Settings settings, Item item);
```
Player3324: There is no obvious reason to have this method, the documentation at least doesn't sufficiently describe it. The behavior description is unclear as well. The process is better described as association, there's no builder (applies to build naming in the impl too, onBuild is better as onConstruct).

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Creates a new CustomItemSetting with the given default value.
+	 *
+	 * @param defaultValue the value all items that do not explicitly set this setting will have.
+	 * @return a new CustomItemSetting
+	 */
+	static <T> CustomItemSetting<T> create(Supplier<T> defaultValue) {
```
Player3324: An override that takes T directly would be useful, it is what most uses need.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -49,7 +48,18 @@ public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvi
...
+
+	/**
+	 * Sets a custom setting of the item.
+
+	 * @param setting the unique type for this setting
+	 * @param value the object containing the setting itself
+	 * @return this builder
+	 */
+	public <T> FabricItemSettings custom(CustomItemSetting<T> setting, T value) {
+		((CustomItemSettingImpl<T>) setting).set(this, value);
```
Player3324, line 66: Is there anything wrong with invoking set on the interface, i.e. without the cast?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/CustomItemSettingImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+	@Override
+	public void build(Item.Settings settings, Item item) {
+		Objects.requireNonNull(settings);
+
+		this.customItemSettings.put(item, this.customSettings.getOrDefault(settings, this.defaultValue.get()));
+	}
+
+	public static void onBuild(Item.Settings settings, Item item) {
+		for (CustomItemSettingImpl<?> setting : CUSTOM_SETTINGS.getOrDefault(settings, Collections.emptyList())) {
```
Player3324, line 73: this should mention that Item.Settings is reusable, thus onBuild potentially happening for more than one item

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public interface CustomItemSetting<T> {
+	/**
+	 * Returns the current value of this setting for the given {@link Item}.
+	 * Should only be called after or within item construction.
+	 *
+	 * @param item the item
+	 * @return the current setting if present, the default setting if not
+	 */
+	T getValue(Item item);
```
Player3324: Just `get` is fine here IMO.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,95 @@
...
+ * <p>You should probably not implement this interface, unless you have a very highly specialized use case.
+ *
+ * <p>Use {@link CustomItemSetting#create(Supplier)} to retrieve an instance of Fabric API's default implementation.
+ *
+ * @param <T> the type of the setting to be attached
+ */
+public interface CustomItemSetting<T> {
+	/**
+	 * Returns the current value of this setting for the given {@link Item}.
+	 * Should only be called after or within item construction.
```
Player3324: Not sure how one would possibly call this before the constructor.. Only real option is injecting before your mixin, which isn't really worth mentioning.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
```diff
@@ -0,0 +1,95 @@
...
+ *         // Gets the setting from the specified item.
+ *         String tooltip = CUSTOM_TOOLTIP.getValue(stack.getItem());
+ *
+ *         if (tooltip != null) {
+ *             lines.add(new LiteralText(tooltip));
+ *         }
+ *     }
+ * }}</pre>
+ *
+ * <p>You should probably not implement this interface, unless you have a very highly specialized use case.
```
Player3324: I wouldn't suggest mods implementing it at all.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
```diff
@@ -39,8 +39,7 @@
 	 * @return this builder
 	 */
 	public FabricItemSettings equipmentSlot(EquipmentSlotProvider equipmentSlotProvider) {
-		FabricItemInternals.computeExtraData(this).equipmentSlot(equipmentSlotProvider);
-		return this;
+		return custom(CustomItemSettingImpl.EQUIPMENT_SLOT_PROVIDER, equipmentSlotProvider);
```
ghost: Can you change this to `return this.custom(...` please otherwise change L51 to `return custom(...`

Has been resolved...

Player3324 (commented):
The is an overlap with the api-lookup api that needs further invesigation.

This PR handles both a) transporting the data from Item.Settings.custom(data) to the constructor and b) holding and retrieving said data externally with any item instance. The latter is what api-api does as well, although significantly less conveniently.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: I had [direct overrides for primitive types](https://github.com/Haven-King/fabric/blob/08a1d61c3ee3248b229cc78952ead5b98ce93d53/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java) before, but people said it created too much clutter. I would be happy to add them back. An override that takes T directly for non-primitive types could be an issue due to mutability.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: Having this method allows me to drop [this cast](https://github.com/FabricMC/fabric/pull/1359#discussion_r602466780), I simply missed that one spot. I will clarify the documentation and name.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemSettings.java
Haven-King: This is leftover from before the interface included the `set` method. I must have missed it.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: Exposing this method means that we can call it [here](https://github.com/Haven-King/fabric/blob/a8457846b6aa98a47fe5528d2f48626943be93da/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/CustomItemSettingImpl.java#L74) on any instance of `CustomItemSetting` without a cast to `CustomItemSettingImpl`. This is useful because mods implementing `CustomItemSetting` themselves despite warnings won't break everything.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Player3324: Mutability is not a concern that applies here, we can't reasonably cater to every eventuality and the supplier would yield the same mutable object with the exact same fallout.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Player3324: We can actually reject custom instances in FabricItemSettings, anything else shouldn't break.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Haven-King: The supplier will not yield the same mutable object.
```java
// Safe, each item will have its own list.
CustomItemSetting<List<Integer>> setting = CustomItemSetting.create(() -> new ArrayList<>());
```
```java
// Not safe, each item will share the one list.
// Very easy mistake to make, easily preventable.
CustomItemSetting<List<Integer>> setting = CustomItemSetting.create(new ArrayList<>());
```
One would have to go through a fair bit of effort to get the unsafe behavior with the supplier.
```java
List<Integer> list = new ArrayList<>();
CustomItemSetting<List<Integer>> setting = CustomItemSetting.create(() -> list);
```

Player3324:
I forgot to mention there that adding such a mechanism would imply doing the same for blocks. I don't really want to clutter the interface with a generic arg for block vs item, so some code duplication is in order. Not sure how much the impl can be shared.

Haven-King:
I do plan on adding a similar mechanism for block settings. I want to get this PR merged as a precedent for how I will handle blocks.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemSetting.java
Player3324: Your example is quite contrived, I think the object typically comes from elsewhere like the game, config, field - not constructed in place as a mutable container. The Item.Settings aren't meant to store evolving data, so one shouldn't store an empty ArrayList in it, but one of the many unmodifiable List species.

I am very much not a fan of making the common case annoying in favor of some limited safety benefits in niche cases.

PseudoDistant:
@sfPlayer1 was something like this ever finished?

I could pick it up if needbe.

liach:
Can probably be replaced by #2172

PseudoDistant:
Ah, maybe.

Technici4n:
@liach No, this PR is for immutable data that's stored in the `Item` and set at creation time. #2172 is for mutable data that is attached to many game objects (worlds, chunks, entities, block entities) and can be updated / persisted.

PR #1365, opened 2021-03-12
Cleanup Events

Technici4n:
This PR cleans up the `ArrayBackedEvent` implementation and explores the use of a new `createArrayBacked` function that uses the `invokerFactory` for any number of listeners (the current functions don't call the `invokerFactory` when there is only 1 listener).

The main motivation for this PR is that something like #1322 can best be implemented as an event + an `instanceof` check for easy self-implementation on the relevant argument:
```java
public interface SlotClick {
    // Return false to stop vanilla Item#onClicked processing.
    boolean allowClick(Slot slot, ItemStack heldStack, ClickType type, PlayerInventory inventory);
}

Event<SlotClick> SLOT_CLICK = EventFactory.createArrayBacked(SlotClick.class, listeners -> (slot, heldStack, type, inventory) -> {
    boolean vanillaClick = true;
    if (slot instanceof SlotClick) {
        vanillaClick = ((SlotClick) slot).allowClick(slot, heldStack, type, inventory) && vanillaClick;
    }
    for (SlotClick listener : listeners) {
        vanillaClick = listener.allowClick(slot, heldStack, type, inventory) && vanillaClick;
    }
    return vanillaClick;
})
```

Sadly this won't work currently, as this will not be invoked when there is exactly one registered listener.
This PR explores how this could be improved, by introducing a new function similar to `createArrayBacked` that gives the factory full freedom to manage its listeners. Note how the old behavior is trivial to implement in terms of the new function.

I am suggesting a change in name so that the weird `createArrayBacked` that calls the factory in all cases except when there is 1 listener is not used by mistake.
The other overload can stay for performance-sensitive cases: either you specify an empty invoker AND singleton listeners are directly returned, or the invoker factory is called in every case, but I have deprecated the old name for consistency over the long run.

The new function name is obviously not suitable but I couldn't come up with a better name while writing this, so I decided to make the intent clear. I would like something quite different from the old name. Suggestions are welcome.

I have also massively cleaned up the `ArrayBackedEvent` implementation because frankly it was horrible. Registration performance is irrelevant anyway (as long as it's reasonable).

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
@@ -47,18 +47,70 @@ public final class EventFactory {
 
 	/**
 	 * Create an "array-backed" Event instance.
+	 * The factory will be used for any number of listeners.
+	 * Consider using {@link #createUnbad the more optimized overload} if you need a slight performance increase
+	 * when there are 0 or 1 listeners.
 	 *
 	 * @param type           The listener class type.
 	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
 	 * @param <T>            The listener type.
 	 * @return The Event instance.
 	 */
+	public static <T> Event<T> createUnbad(Class<? super T> type, Function<T[], T> invokerFactory) {
+		return EventFactoryImpl.createUnbad(type, invokerFactory);
+	}
+
+	/**
+	 * Create an "array-backed" Event instance, with a custom empty invoker.
+	 * The empty invoker will be used when there are no listeners,
+	 * and when there is only one listener, it will be used directly.
+	 * The factory will only be used when there are at least two listeners.
+	 * Consider using {@link #createUnbad} if this optimization is unsuitable or unneeded.
+	 *
+	 * @param type           The listener class type.
+	 * @param emptyInvoker   The custom empty invoker.
+	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
+	 * @param <T>            The listener type.
+	 * @return The Event instance.
+	 */
+	public static <T> Event<T> createUnbad(Class<? super T> type, T emptyInvoker, Function<T[], T> invokerFactory) {
+		return createUnbad(type, invokers -> {
+			if (invokers.length == 0) {
+				return emptyInvoker;
+			} else if (invokers.length == 1) {
+				return invokers[0];
+			} else {
+				return invokerFactory.apply(invokers);
+			}
+		});
+	}
+
+	/**
+	 * Create an "array-backed" Event instance.
+	 * The factory will be used when there are zero listeners, or at least two.
+	 * If there is only one listener, that one will be used as the invoker, and the factory will not be used!
+	 *
+	 * @param type           The listener class type.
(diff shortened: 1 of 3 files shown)


Technici4n:
For anyone wondering: the performance and the behavior of the old functions should be exactly the same. The indirection in `createArrayBacked` is only for invoker creation, so it's out of the hot path!

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -47,18 +47,70 @@ public static void invalidate() {
...
+	 * @param type           The listener class type.
+	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
+	 * @param <T>            The listener type.
+	 * @return The Event instance.
+	 * @deprecated Use {@link #createArrayBackedUsingTheFactoryEveryTime(Class, Function) createUnbad}.
+	 */
+	@Deprecated
 	public static <T> Event<T> createArrayBacked(Class<? super T> type, Function<T[], T> invokerFactory) {
-		return EventFactoryImpl.createArrayBacked(type, invokerFactory);
+		return createArrayBackedUsingTheFactoryEveryTime(type, invokers -> {
```
liach, line 101: No, you can just pass in the old invoker factory. The old invoker factory doesn't break when the input array is empty or has length 1. So you can just replace the old `createArrayBacked` with `createArrayBackedUsingTheFactoryEveryTime` without concerns

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -47,18 +47,70 @@ public static void invalidate() {
...
+	 * The factory will only be used when there are at least two listeners.
+	 * Consider using {@link #createArrayBackedUsingTheFactoryEveryTime} if this optimization is unsuitable or unneeded.
+	 *
+	 * @param type           The listener class type.
+	 * @param emptyInvoker   The custom empty invoker.
+	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
+	 * @param <T>            The listener type.
+	 * @return The Event instance.
+	 */
+	public static <T> Event<T> createArrayBackedCustomEmptyInvoker(Class<? super T> type, T emptyInvoker, Function<T[], T> invokerFactory) {
```
liach, line 76: Why? Why can't the invoker factory handle the case when the input array is empty?

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: I am hesitant to introduce such a subtle behavior change. What if people relied on the old behavior? (I don't know why, but still)

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: Apparently it's slightly faster, which could be useful for events that are not used but are in a hot path. Arguably this is useless as soon as there is more than 1 listener.

Technici4n:
I don't know why I felt the need to over-engineer this. Opening a new PR.

Issue #1357, opened 2021-03-10
The modularization of the Fabric API has gone too far


OroArmor:
The title says it all. We currently have 42 modules in the Fabric API. Often times these modules have somewhat of an overlaping scope. I propose a massive refactor to reduce the number of modules, remove the deprecated modules, and propose higher scrutiny for new modules. Looking through the modules, they often only have a handful of classes, and should be merged into other modules.
My proposed modules:
`module`: existing modules to include
`item` : `item`, `itemgroup`, `tool-attribute`, `mining-levels`, the item section of `object-builder` (model predicate providers)
`block`: `block-render-layer`, block section of `object-builder`, `container`
`entity`: `entity-events`, villager parts of `object-builder`
`world`: `biome`, `dimension`, `structure`, `gamerule`, (`command`?), `loot-tables`, advancements section of `object-builder`
`lifecycle`: `lifecycle`, (`command`?)
`rendering`: all rendering apis, `particles`, `textures`, `models`
`client`: `keybinds`
`screen`: `screen`, `screen-handler`
`registry`: `registry-sync`

Modules that stay:
`resource-loader`, `networking`, `tag-extensions`, `crash-reports`, and `base`

This list may not be complete, but i think reducing the number of modules will be an important part of the fabric API moving forward

LambdAurora:
I agree with most.

But I entirely disagree with the rendering part, renderer-api should stay separate from the other rendering stuff.

OroArmor:
I have little experience with the rendering apis, but there are 7 modules with the word `rendering` in them, there has to be a way to combine some of them

LambdAurora:
> I have little experience with the rendering apis, but there are 7 modules with the word `rendering` in them, there has to be a way to combine some of them

the renderer API is fundamentally different from the others.

i509VCB:
I think some modules need to be combined. mini modules unless actually justified aren't that efficient with space.

Also less modules -> quicker builds

liach:
Another criteria for grouping modules imo would be allowing them to break in updates. I hope this can be considered when we try to merge modules, i.e. don't merge frequently changed modules with relatively stable modules. Biomes can stay as it blows up often given mojang's frequent updates.

PR #460, opened 2019-12-11
fabric-mining-levels-v0 -> fabric-tool-attribute-api-v1 (1.15 edition)

LemmaEOF:
Reimplementation of #352 on the 1.15 branch, now without merge conflicts!

Moves the ToolManager and tool tags declared in fabric-mining-levels-v0 to a new fabric-tool-attributes-v1, along with adding new interfaces to be implemented on tools so that mining speed, mining level, and entity attribute modifiers can be decided by the itemstack.

PR #359, opened 2019-09-07
fabric-rendering-v1 module & added HudRenderCallback

shedaniel:
This will aid mods adding text/stuff on-screen (either maybe those tiny hud mods where you can see fps without f3)

also bumped module version

PR #1366, opened 2021-03-12
Demodularization of the Fabric API

OroArmor:
This picks up from #1357 and how the fabric API is bloated. This targets the 1.17 snapshots as any changes to 1.16 would be too breaking. This is a work in progress and any suggestions will be considered.

Edit: I know this is super early, but the sooner we start, the easier it will be to meet the target when 1.17 is implemented


OroArmor:
This first commit removes the deprecated modules from the fabric API, as they have no need to be included any more

Technici4n:
I believe you should hold on until this is discussed with and approved by the maintainers, possibly on an issue or on Discord, or it will be a waste of effort. Not gonna close this, but I would encourage you to.

OroArmor:
That is precisely why I am doing it. To start conversations and get the ball rolling. Yes this can be done when 1.17 rolls out, but if we build it up over time, we can get a better result and one that has community input

liach:
Not really. If we demodularize, we can still have like at least 3 modules:
1. base module, something maybe useful for both sides but doesn't require presence on the other side
1. server module, in which all modifications are safe for server-only uses (i.e. usable by plugin-like mods), such as command api
2. client module, in which all modifications are like clientside-only (for client-only mods), such as client command api
3. fat module, depend on the other two, adds mechanisms that need both sides' supports

we need at least these 4 modules imo.

maityyy:
I like this idea because it will help in the development of server mods. Either modules need to be grouped, or combined. In addition, there are no clear rules on how to divide modules and where to add any functionality.

maityyy:
liach offered a good idea of ​​how many modules should be and how to split them. 

The base module definitely needs to be separated from the rest to make it possible to develop your own API.

Technici4n:
I am against de-modularization. The benefit of modules is clear: if a module breaks at some point during an update it's much easier to disable it entirely and have someone PR the update for the entire module, than disable a few classes here and there and have to guess what needs porting and what not.

The only drawback I see is for contributors to API: modules make compilation slower, but if you run with IntelliJ it's fine, and I am pretty sure there are still huge improvements to be made when it comes to gradle performance.

For API users, I still fail to understand what's so hard with `modImplementation "net.fabricmc.fabric-api:fabric-api:${project.fabric_version}"`?

TL;DR _Why?_

OroArmor:
I think a little bit was lost from the issue. I do not support the complete integration of the entire api. I do believe a part of fabrics strength is the modularity (Mostly @liach). I think that we just have too many modules that have overlapping scope, and some of this needs to be fixed.

OroArmor:
This pr is really a staging ground for how we think to best move forward with the api. That's why the first thing I did was remove the deprecated modules, as there is strong consensus on removing them for 1.17 or 1.18 by the latest. Next my goal is to break the object builder api into specific modules that more closely match their goal

Technici4n:
> I think that we just have too many modules that have overlapping scope

As long as the implementations and features are relatively independent it makes sense to keep them separate so that they can be updated separately.

For example, yes `blockrenderlayer-v1` could be merged into `rendering-v1`, but then if mojang decides to change block render layers it breaks rendering v1. Your issue completely misses the point imo. There is no downside to keeping them separate besides slightly increased compilation times.

OroArmor:
Let's sya the block render layer is broken one snapshot. If we comment out the module, that breaks any mods that update to the new version as that module doesn't exist.

On the other hand, we add `@Deprecated` and comment out the insides of the methods, making them do nothing. This allows mods to soft fail, as things don't work 100% but it doesn't cause a crash.

We could do the same with our current setup, but why would we need it. As well the current object builder api sorta goes against your logic, which seems it would be in favor of increasing the number of modules

Technici4n:
> This allows mods to soft fail, as things don't work 100% but it doesn't cause a crash.

Yes that's worse than any other option.

> As well the current object builder api sorta goes against your logic, which seems it would be in favor of increasing the number of modules.

Not sure about that one, in theory it could be split up. In practice I don't care and don't want to spend time refactoring APIs just for the sake of it.

liach:
Afai see, some of our modules already exhibit a tendency toward growing large. An example would be the object builder. If we do have mutual intermodule dependencies, we can consider merging those modules in those cases.

modmuss50:
Im not against removing some of the deprecated modules. some of the ones being removed here have been deprecated quite recetally so IMO should stay around for another MC version.

The other side of my argugment is that these modules should only be removed when they cause a mataintance burden, right now they aren't really causing too much trouble being here so why remove them for the sake of it?

I also want to look at making some of the unstable modules stable, most notablily the biome modifications.

frqnny (commented):
Personally, I find fabric-networking-v0 to still be used by some mods who haven't updated. But If it is going to be removed in 1.17, that should be done with an official warning to modders about every module that is going to be removed.

OroArmor:
1.17 introduces a bunch of breaking changes to networking anyway. If we remove it, that's less mantinence and makes sure people don't run 1.16 in 1.17

modmuss50:
There were no modder/api facing changes to the networking in 1.17.

V0 networking is staying in 1.17 as its still being used by a larger number of mods and V1 is still quite new.

liach:
Hmm, I thought v0 is totally superseded by v1 as all features of v0 are already on v1 (networking)?

frqnny:
> Hmm, I thought v0 is totally superseded by v1 as all features of v0 are already on v1 (networking)?

Yeah but many mods still use the old api because the old module continues to work as of right now, and taking it away would break lots of mods as modmuss said. Maybe if any big networking changes come?... probably not soon but still needs anple time

Ueaj-Kerman:
imo we should deprecate/remove APIs when there isn't much else to update, deleting all the APIs the same time a big update comes out makes it a pain to update.

liach:
> 
> 
> deleting all the APIs the same time a big update comes out makes it a pain to update.

I disagree. Imo if there is a throughout change, we just don't fear making it more throughout. But since we are already in 1.17 snapshots, we should probably push module removal back when mojang introduces some major mechanism change during snapshots (like how things went in 1.13/1.14 snapshots where there were major breaks). Say we deprecate biome api. mojang update biomes and breaks the old biome api, then it's a good time for us to remove the old api (already deprecated)

LambdAurora:
I have gone to do a little dip in old PRs and see when and for what versions some of the v1 replacement modules were merged.

- `command-api-v1` got merged in 1.15 replacing `commands-v0` module, had 2 versions for migrating.
  -> Should be removed in 1.17
- `screen-handler-api-v1` got merged in 1.16 replacing `containers-v0`
  -> A bit recent. Container naming is extremely out-of-date. I don't recall seeing many mods using this one for screen handler anymore
  -> Removal up to debate (imo, could be removed in 1.17)
- `lifecycle-events-v1` got merged in 1.15 and 1.16 replacing `events-lifecycle-v0`, happened in July.
  -> Removal up to debate, should be removed at least for 1.18
- `key-binding-api-v1` got merged in 1.15 (June) replacing `keybindings-v0`.
  -> Modders had the whole 1.16 version lifecycle to update.
  -> Potentially could be removed in 1.17
  -> Up to debate
- `networking-api-v1` got merged in 1.16 very recently replacing `networking-v0`
  -> As much as I would like to remove all deprecate modules, it's safe to say that this one should stay for the 1.17 lifecycle and be removed in 1.18
- `object-builder-api-v1` got merged in 1.16 snapshots (April), replacing `object-builders-v0` and was backported up to 1.14.
  -> Should be removed in 1.17
- `rendering-v1` got merged in 1.15 (December 2019 #359), replacing `rendering-v0`
  -> There was plenty of time to migrate, should be removed in 1.17.
- `tool-attribute-api-v1` got merged in 1.15 (April #460), replacing `mining-levels-v0`
  -> Should be removed in 1.17
  
I believe removal of `interaction-events` is an error as it doesn't seem to have been fully replaced by another module.

liach:
So here most of the apis we have in @LambdAurora's list are largely untouched on minecraft's side. Imo we can add an exception that if an api is touched by minecraft heavily (say like commands in 1.13), we can discard the old deprecated version immediately.

Technici4n:
> Imo we can add an exception that if an api is touched by minecraft heavily (say like commands in 1.13), we can discard the old deprecated version immediately.

Always has been, see biomes. ;)

Pyrofab:
> Imo if there is a throughout change, we just don't fear making it more throughout

I have seen this mentality in Forge development, and I can say it's not enjoyable as a modder. I do not mind fixing a few things every now and then, but having to fix a huge pile of highlighted mess is extremely disheartening.

That said, in this instance the replacements are already available on current versions, so I would not have an issue removing any of the modules mentioned by lambda in 1.17.

Technici4n:
Having to rewrite 80% of your mod is no fun... (Hello 1.13+ Forge)

In this particular case, we are talking about deprecated APIs. Modders can update to the new APIs at any time, and they will have had over 1 year to do so.

i509VCB:
Any APIs we wish to mark for removal in a few versions should also get the stronger `@MarkedForRemoval` annotation.

Moving the subproject directories into a `deprecated` folder would probably be a good start for the modules that have replacements and are not ready for removal.

liach:
Imo moving around the directories isn't worth it, as it just adds to maintenance cost with no real gain. But an anno for removal is good.

Technici4n:
Yeah we can add the `ScheduledForRemoval` annotation. Someone can PR that!

OroArmor:
Discussion should continue in #1357

PR #1367, opened 2021-03-13
Add a Crossbow API to the Item API

OroArmor:
Very similar to #1210, in fact, some classes are the same, just with the other pr's logic removed (I will update the other or once one is merged). Luckily crossbows are less :mojank: and the logic for firing them is not attached to the users, but the item itself, reducing the complexity of the event system.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import org.jetbrains.annotations.NotNull;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.projectile.PersistentProjectileEntity;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ShotProjectileEvents {
+	private ShotProjectileEvents() {
+	}
+
+	/**
+	 * This event modifies the projectile entity shot from a crossbow.
+	 */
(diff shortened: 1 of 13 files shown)


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		return persistentProjectileEntity;
+	});
+
+	public interface ReplaceProjectileFromCrossbow {
+		/**
+		 * In this method you can replace the arrow shot from your custom crossbow. Applies all of the vanilla arrow modifiers first.
+		 *
+		 * @param crossbowStack              The ItemStack for the Crossbow Item
```
LambdAurora: For all javadoc: no capitalization on the first character in param and return javadoc!

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ShotProjectileEvents.java
```diff
@@ -0,0 +1,81 @@
...
+		 * @param projectileStack            The ItemStack for the projectile currently being shot
+		 * @param user                       The user of the crossbow
+		 * @param persistentProjectileEntity The arrow entity to be spawned
+		 * @return The new projectile entity. Return null if you do not change the entity.
```
LambdAurora:
```suggestion
		 * @return the new projectile entity. Return {@code null} if you do not change the entity.
```

For all literals using `{@code}` would be appreciated. It changes the font in a rendered javadoc which makes it easier to distinct from normal text.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/crossbow/CrossbowItemMixin.java
```diff
@@ -0,0 +1,52 @@
...
+		persistentProjectileEntity = ShotProjectileEvents.CROSSBOW_REPLACE_SHOT_PROJECTILE.invoker().replaceProjectileShot(crossbow, projectileStack, entity, persistentProjectileEntity);
+		ShotProjectileEvents.CROSSBOW_MODIFY_SHOT_PROJECTILE.invoker().modifyProjectileShot(crossbow, projectileStack, entity, persistentProjectileEntity);
+		cir.setReturnValue(persistentProjectileEntity);
+	}
+
+	//Redirecting this method in order to get the item stack and shooting entity
+	@Redirect(method = "use", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/CrossbowItem;shootAll(Lnet/minecraft/world/World;Lnet/minecraft/entity/LivingEntity;Lnet/minecraft/util/Hand;Lnet/minecraft/item/ItemStack;FF)V"))
+	private void shootAll(World world, LivingEntity entity, Hand hand, ItemStack stack, float speed, float divergence) {
+		float _speed = stack.getItem() instanceof FabricCrossbowExtensions ? ((FabricCrossbowExtensions) stack.getItem()).getProjectileSpeed(stack, entity) : speed;
```
LambdAurora: That underscore is a bit cursed imo. `newSpeed` is a bit more verbose but respects Java naming conventions.

LambdAurora (changes requested):
Conceptually good.

My complaints are mostly about the javadocs.

LambdAurora (approved):

PR #1369, opened 2021-03-13
Change event behavior for one listener, introduce `AutoInvokingEvent`, update `ArrayBackedEvent` implementation

### LINKED ISSUE #683: ArrayBackedEvent will throw ClassCastException when the backing type is primitive.
The two calls of `Array.newInstance` will return primitive array types if `this.type` represents a primitive type. The unchecked casts to `T[]` (erased to `Object[]`) will subsequently fail.

This issue is most likely academic, since the intended types for `T` will be functional interface types, but the issue exists in theory.

I suggest checking whether the backing type is primitive in the constructor (via `Class#isPrimitive`) and throwing `IllegalArgumentException` if that is the case.

### LINKED ISSUE #1311: ArrayBackedEvent assuming invoker generation when there is single callback
In the ArrayBackedEvent, the logic assumes that when there is a single callback, it can be used straight as an invoker.

However, this is not the case, as frequently invoker factory has different logic beyond simple dispatches.

An example: https://github.com/FabricMC/fabric/pull/1210/files#diff-11d2ce35b7dd909366a206d94a5b08a5f4d7fb4a37e8949c5e0ad2a741fed199R38-R45

Similarly, some invokers might have some arguments null when invoked, only lazily computed when listeners are present.

Technici4n:
This PR changes the behavior of `EventFactory.createArrayBacked` so that the factory is invoked even when there is one listener.

The main motivation for this PR is that something like #1322 can best be implemented as an event + an `instanceof` check for easy self-implementation on the relevant argument:
```java
public interface SlotClick {
    // Return false to stop vanilla Item#onClicked processing.
    boolean allowClick(Slot slot, ItemStack heldStack, ClickType type, PlayerInventory inventory);
}

@AutoInvokingEvent // This PR adds this annotation as well!
Event<SlotClick> SLOT_CLICK = EventFactory.createArrayBacked(SlotClick.class, listeners -> (slot, heldStack, type, inventory) -> {
    boolean vanillaClick = true;
    if (slot instanceof SlotClick) {
        vanillaClick = ((SlotClick) slot).allowClick(slot, heldStack, type, inventory) && vanillaClick;
    }
    for (SlotClick listener : listeners) {
        vanillaClick = listener.allowClick(slot, heldStack, type, inventory) && vanillaClick;
    }
    return vanillaClick;
})
```

Sadly this won't work currently, as the invoker factory will not be invoked when there is exactly one registered listener. I don't think this behavior change should break anything.

This PR also introduces [`AutoInvokingEvent`](https://github.com/Technici4n/fabric/blob/event-cleanup/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java) to document such events.

I have also simplified the `ArrayBackedEvent` implementation by moving the empty listener/singleton listener to the body of the custom empty invoker overload, and by removing the special case where `listeners` could be `null` by allocating an empty array in the constructor.

Edit: Closes #1311.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
@@ -1,6 +1,10 @@
 package net.fabricmc.fabric.api.event;
 
-import java.lang.annotation.*;
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
  * Indicates that this {@link Event} is auto-invoking:
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
@@ -1,0 +1,35 @@
+package net.fabricmc.fabric.api.event;
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that this {@link Event} is auto-invoking:
+ * it calls the event callback implemented by a context parameter type automatically and without registration.
+ *
+ * <p>This means that this event can be listened to in two ways:
+ * <ul>
+ *     <li>If the consumer is the context parameter and it implements the callback, it will be automatically invoked, don't register manually.
+ *     <li>Otherwise, there is no invocation and the listener needs manual registration as usual.
+ * </ul>
+ *
+ * <p>Do note that there may be more than one context parameter.
+ *
+ * <p>A typical use case is feature augmentation, for example to expose raw clicks to slots.
+ * The event callback has a slot parameter - the context parameter - and the event itself is carrying this annotation.
+ * All the slot needs to receive slot clicks is to implement {@code SlotClickCallback} on itself.
+ * It shouldn't do any explicit event registration like {@code SLOT_CLICK_EVENT.register(this::onSlotClick)},
+ * otherwise it will see extraneous callback invocations.
+ *
+ * <p>In general, an auto-invoking event bridges the gap between the flexibility of an event with global reach,
+ * and the convenience of implementing an interface that gets detected automatically.
+ *
+ * <p>This is a documentation-only annotation, the event factory has to implement the functionality explicitly by checking the parameter type and invoking it.
+ * On top of adding this annotation, the event field or method should document which parameters are context parameters,
+ * and under which circumstances they are invoked.
+ */
+// TODO: explore enforcing that auto-invoked listeners don't register themselves.
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.METHOD })
+public @interface AutoInvokingEvent {
+}
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.fabricmc.fabric.api.event;
 
 import java.lang.annotation.Documented;
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
@@ -47,70 +47,25 @@ public final class EventFactory {
 
 	/**
 	 * Create an "array-backed" Event instance.
-	 * The factory will be used for any number of listeners.
-	 * Consider using {@link #createArrayBackedCustomEmptyInvoker(Class, Object, Function)}  the more optimized overload} if you need a slight performance increase
-	 * when there are 0 or 1 listeners.
+	 * The factory will be used to create the invoker for any number of listeners.
+	 *
+	 * Consider using {@linkplain #createArrayBacked(Class, Object, Function) a custom empty invoker}
+	 * if performance of this event is critical.
 	 *
 	 * @param type           The listener class type.
 	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
 	 * @param <T>            The listener type.
 	 * @return The Event instance.
 	 */
-	public static <T> Event<T> createArrayBackedUsingTheFactoryEveryTime(Class<? super T> type, Function<T[], T> invokerFactory) {
-		return EventFactoryImpl.createArrayBackedUsingTheFactoryEveryTime(type, invokerFactory);
-	}
-
-	/**
-	 * Create an "array-backed" Event instance, with a custom empty invoker.
-	 * The empty invoker will be used when there are no listeners,
-	 * and when there is only one listener, it will be used directly.
-	 * The factory will only be used when there are at least two listeners.
-	 * Consider using {@link #createArrayBackedUsingTheFactoryEveryTime} if this optimization is unsuitable or unneeded.
-	 *
-	 * @param type           The listener class type.
-	 * @param emptyInvoker   The custom empty invoker.
-	 * @param invokerFactory The invoker factory, combining multiple listeners into one instance.
-	 * @param <T>            The listener type.
-	 * @return The Event instance.
-	 */
-	public static <T> Event<T> createArrayBackedCustomEmptyInvoker(Class<? super T> type, T emptyInvoker, Function<T[], T> invokerFactory) {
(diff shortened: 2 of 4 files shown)


Technici4n:
@liach This should be much simpler and better than my attempt from yesterday.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -70,9 +77,16 @@ public static void invalidate() {
 	 * @param <T>            The listener type.
 	 * @return The Event instance.
 	 */
-	// TODO: Deprecate this once we have working codegen
 	public static <T> Event<T> createArrayBacked(Class<T> type, T emptyInvoker, Function<T[], T> invokerFactory) {
-		return EventFactoryImpl.createArrayBacked(type, emptyInvoker, invokerFactory);
+		return createArrayBacked(type, invokers -> {
+			if (invokers.length == 0) {
+				return emptyInvoker;
+			} else if (invokers.length == 1) {
```
liach: You should remove this branch.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -70,9 +77,16 @@ public static void invalidate() {
 	 * @param <T>            The listener type.
 	 * @return The Event instance.
 	 */
-	// TODO: Deprecate this once we have working codegen
 	public static <T> Event<T> createArrayBacked(Class<T> type, T emptyInvoker, Function<T[], T> invokerFactory) {
```
liach, line 84: Need to deprecate this one since now empty invoker is generated by the factory.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -47,6 +47,10 @@ public static void invalidate() {
 
 	/**
 	 * Create an "array-backed" Event instance.
+	 * The factory will be used to create the invoker for any number of listeners.
```
liach: This one is redundant imo. I don't see why people would think differently about how the invoker factory works. If there should be documentation, it should be on the old unintuitive version than this version.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -47,6 +47,10 @@ public static void invalidate() {
...
+	 *
+	 * <p>Consider using {@linkplain #createArrayBacked(Class, Object, Function) a custom empty invoker}
+	 * if performance of this event is critical.
```
liach, line 53: After your change, the empty invoker version seems to be less efficient.

Technici4n: It is more efficient because the factory is invoked only in the event `update()` function. So in this case the resulting `emptyInvoker` is returned directly and stored in the event, and then returned directly when `.invoke()` is called.

Note that standard use passes `listeners -> (event signature) -> { ... }`, whereas here this is only `listeners -> /* some logic */`.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: It is a bit faster as it bypasses the usual loop in the factory.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
liach: Hmm, might need to specify. Like if the function passed in does not have an optimiztion for `listeners.length == 0` or `listeners.length == 1`, then this version is better. Otherwise, it isn't.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
liach: I mean, the empty invoker and the factory can be customized to do extra work beyond calling the invokers directly. This branch destroys the possibility of customizing in the case when `listeners.length == 1`.

liach: Also the lambda param should be better named `listeners`, as invoker is a single object that dispatch events in the fabric base api definition.

Technici4n: That's true, but this will be a bit faster for factories that only deal with listener composition, so most of them, and certainly all the existing ones. Factories that want to add custom behavior can use the other overload.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
liach: Looking at the usage, please mark this as resolved as this isn't a problem.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: > Also the lambda param should be better named listeners, as invoker is a single object that dispatch events in the fabric base api definition.

Yup, was distracted apparently. 😄

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: Good idea, will update the javadoc in a bit.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
liach: > 
> 
> That's true, but this will be a bit faster for factories that only deal with listener composition, so most of them, and certainly all the existing ones. Factories that want to add custom behavior can use the other overload.

Imo you should document that for invoker customization purposes, this shouldn't be used if it aims to do more than just calling invokers.

Say the event in your pr description is seen by others. Seeing that the empty invoker version is more "optimized", they incorrectly optimize the example event `SLOT_CLICK` in the pr description, only to see it go :boom: because a custom invoker ain't created when there is only one listener.

liach: Or should we change the empty invoker variant to have an alternative name? Like `createDirectArrayBacked` (bad name here) to indicate this aims to directly dispatch events?

Technici4n: It is painful to come up with a sufficiently distinct name for such a tiny behavior difference, and we would need to essentially duplicate the function and deprecate the old one just for the rename. I'm not fond of it.

Technici4n:
@liach I believe all your comments have been addressed so far.

liach (approved):
Great addition. I believe #1272 can also benefit from this as that pr aims to only create objects for keys when there is need (as the events are fired every tick)

Technici4n:
Updated to fix #683.

If anyone wants to test this, just copy/paste the following class next to `FabricApiBaseTestInit` and call `init()`: https://gist.github.com/Technici4n/0c174c0660a1e965fa01a2f449c511f0. (It will crash due to the `int.class` at the end).

Would it be worth it to document the test a bit and put it in the module with this PR?

liach:
Hmm, IMO throwing that class cast is not that different from throwing an illegal argument.

Technici4n:
Yes you are right. This PR moves the cast to the constructor so it will crash anyway. Before that it would only crash later.

Pyrofab:
I believe an exception with a proper message is preferable, otherwise modders may not understand why primitive arrays cause a failed cast.

liach:
> 
> 
> I believe an exception with a proper message is preferable, otherwise modders may not understand why primitive arrays cause a failed cast.

This would only happen if you have an `Event<Integer> event = EventFactory.createArrayBacked(int.class, invokers -> i -> {/* stuff */})`.

Meanwhile, `Event<Integer> event = EventFactory.createArrayBacked(Integer.class, invokers -> i -> {/* stuff */})`

This thing really can only happen due to extremely bad practice. That is really a non-issue.

Player3324 (approved):
LGTM, thanks!

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * <p>In general, an auto-invoking event bridges the gap between the flexibility of an event with global reach,
+ * and the convenience of implementing an interface that gets detected automatically.
+ *
+ * <p>This is a documentation-only annotation, the event factory has to implement the functionality explicitly by checking the parameter type and invoking it.
+ * On top of adding this annotation, the event field or method should document which parameters are context parameters,
+ * and under which circumstances they are invoked.
+ */
+// TODO: explore enforcing that auto-invoked listeners don't register themselves.
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
```
liach, line 52: Since this is just a decorative annotation, I don't see any point giving it a runtime retention.

Technici4n: We thought it could possibly be used at runtime to perform some checks, so in doubt we decided to give it runtime retention. Now that you mention it I'm not 100% sure... Does this introduce a lot of bloat? I don't think a few annotated events will be a problem.

modmuss50 (approved):

modmuss50:
Still fails to build due to license violations

Technici4n:
Damn, I forgot to push that yesterday. ;)

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/AutoInvokingEvent.java
liach: It won't. After all it's an **annotation**, and keeping extra annotations won't kill people. Looks good.

actuallyasmartname (approved):
LGTM

PR #1374, opened 2021-03-14
Add an Elytra API to the Item API

OroArmor:
Follows in the footsteps of #1210 and #1367 for extensions of vanilla items. Only adds one API class for an elytra interface.

The reasoning for adding the player to the can use logic is so that elytras can be based on player capabilities, allowing for things like the test mod, where an elytra is creative only.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
@@ -21,7 +21,8 @@ import net.minecraft.item.ElytraItem;
 import net.minecraft.item.ItemStack;
 
 /**
- * An interface to implement on all custom elytras.
+ * An interface that allows implementing items to enable elytra flight for players, when equipped in the chest equipment slot.
+ While being used, the item will be damaged <b>every 20 ticks</b> through {@link ItemStack#damage(int,LivingEntity,Consumer)}.
  */
 public interface FabricElytraExtensions {
 	/**
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1.elytra;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ElytraItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement on all custom elytras.
+ */
+public interface FabricElytraExtensions {
+	/**
+	 * Returns if the elytra is usable based on its stack and user.
+	 *
+	 * @param stack The stack for the elytra item
+	 * @param user  The user of the elytra
+	 * @return True if the elytra is usable
+	 */
+	default boolean isUsable(ItemStack stack, LivingEntity user) {
+		return ElytraItem.isUsable(stack);
+	}
+}
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
@@ -25,7 +25,8 @@ import net.minecraft.item.ItemStack;
  */
 public interface FabricElytraExtensions {
 	/**
-	 * Returns if the elytra is usable based on its stack and user.
+	 * Returns {@code true} if this item allows flight based on its stack and user, and {@code false} otherwise.
+	 * This function will be called when a player attempts to start elytra flight, and will also be called every tick during elytra flight to make sure it is still allowed.
 	 *
 	 * @param stack the stack for the elytra item
 	 * @param user  the user of the elytra
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
@@ -29,7 +29,7 @@ public interface FabricElytraExtensions {
 	 *
 	 * @param stack The stack for the elytra item
 	 * @param user  The user of the elytra
-	 * @return True if the elytra is usable
+	 * @return {@code true} if the elytra is usable
 	 */
 	default boolean isUsable(ItemStack stack, LivingEntity user) {
 		return ElytraItem.isUsable(stack);
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
@@ -27,8 +27,8 @@ public interface FabricElytraExtensions {
 	/**
 	 * Returns if the elytra is usable based on its stack and user.
 	 *
-	 * @param stack The stack for the elytra item
-	 * @param user  The user of the elytra
+	 * @param stack the stack for the elytra item
+	 * @param user  the user of the elytra
 	 * @return {@code true} if the elytra is usable
 	 */
 	default boolean isUsable(ItemStack stack, LivingEntity user) {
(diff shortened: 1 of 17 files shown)


LambdAurora (approved):
I like this.

But how do you plan on addressing elytra rendering?

OroArmor:
That already has an api that can do that for you, i don't think i new to make the same thing again (correct me i could be wrong)
https://github.com/OroArmor/Netherite-Plus-Mod/blob/53a673986c20eaaf4199bb93cafa4b20d992e324/common/src/main/java/com/oroarmor/netherite_plus/client/render/NetheriteElytraFeatureRenderer.java and https://github.com/OroArmor/Netherite-Plus-Mod/blob/53a673986c20eaaf4199bb93cafa4b20d992e324/fabric/src/main/java/com/oroarmor/netherite_plus/client/NetheritePlusClientModFabric.java#L77-L81 show how it can easily be done

LambdAurora:
Nice then. Didn't know about `LivingEntityFeatureRendererRegistrationCallback` ahah.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement on all custom elytras.
+ */
+public interface FabricElytraExtensions {
+	/**
+	 * Returns if the elytra is usable based on its stack and user.
+	 *
+	 * @param stack The stack for the elytra item
```
LambdAurora: I actually haven't seen it when I originally reviewed it, but I noticed it when reviewing #1367.

No capitalization on the first character of a `@param` or `@return` javadoc. (this is for the whole PR)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+	 * @param user  The user of the elytra
+	 * @return True if the elytra is usable
```
LambdAurora:
```suggestion
	 * @return {@code true} if the elytra is usable
```

Using `{@code}` on literals would be appreciated.

LambdAurora (changes requested):

LambdAurora (approved):
Very cool :)

Azzyypaaras:
Lil' pogchamp moment

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+package net.fabricmc.fabric.api.item.v1.elytra;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ElytraItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement on all custom elytras.
+ */
+public interface FabricElytraExtensions {
```
hYdos: maybe an better name here?

OroArmor: This follows what i was told to do with the bow in #1210

Platymemo (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
hYdos: ah ok

liach (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
hYdos: I like fornite guys

Technici4n:
Ok, I need this! How is elytra damage handled though? Imo you need to expose a `damage` function in `FabricElytraExtensions`.

OroArmor:
not quite sure @Technici4n. I was thinking it would either be done the same way as the elytra currently, or the damage handler API could be merged and then it can be done via that

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
Technici4n: Couldn't this be called `FabricElytraItem`?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+	/**
+	 * Returns if the elytra is usable based on its stack and user.
```
Technici4n:
```suggestion
	 * Returns {@code true} if this item allows flight based on its stack and user, and {@code false} otherwise.
	 * This function will be called when a player attempts to start elytra flight, and will also be called every tick during elytra flight to make sure it is still allowed.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/elytra/FabricElytraExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package net.fabricmc.fabric.api.item.v1.elytra;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ElytraItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface to implement on all custom elytras.
```
Technici4n:
```suggestion
 * An interface that allows implementing items to enable elytra flight for players, when equipped in the chest equipment slot.
 While being used, the item will be damaged <b>every 20 ticks</b> through {@link ItemStack#damage(int,LivingEntity,Consumer)}.
```

Technici4n (commented):
A few javadoc/naming issues. Make sure the implications of inheriting the elytra item interface are clear for anyone reading only its javadoc.

OroArmor:
I cant replay to the comment on the name, but that follows what I was told in #1210 for the bow, so I'm keeping it consistent across my prs `Fabric<Tool>Extensions`

Technici4n:
I can't find where you were told to use `Fabric<Tool>Extensions`. I find that naming obscure at best, and it is already used for mixin ducks anyway.
1. An extension extends something, and this PR doesn't extend Elytra functionality, it is just exposing it for non-`Items.ELYTRA` items.
2. `FabricElytraExtension` doesn't make it clear that this is supposed to be implemented on items.

PS: The PR is excellent besides that naming issue.

Technici4n:
Another thing, others please tell me if it makes sense or not...
I would prefer `allowElytraFlight` over `isUsable`, I know it doesn't match the yarn name but that yarn name is very old and can be changed later to match this PR. `isUsable` is quite obscure and could easily conflict with other methods on non-trivial items.

haykam821:
Maybe `allowsFallFlying` would work?

Technici4n:
+1 for `allowFallFlying`.

haykam821:
Oh, since that is a method, I would prefer `shouldAllowFallFlying` instead.

Technici4n:
Too long imo, that `should` prefix is unnecessary. :P

Should the interface be renamed to `(Fabric)FallFlyingItem`? 🤔

haykam821:
If it's changed to `FabricFallFlyingItem`, then the entire package and other related classes should be renamed as well.

I think it's best to keep with Yarn conventions here; `allowFallFlying` would only be used for a builder method, and `allowsFallFlying` would only be used for a field. The best choice for a normal method would be `shouldAllowFallFlying`.

liach:
Hmm, I thought `allows` etc (verbs with s) are used as boolean method names in yarn :thinking:

Technici4n (approved):
It's working perfectly, thanks! I would recommend moving `FabricFallFlyingItem` into the top-level item API package though.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/fallflying/FabricFallFlyingItem.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ElytraItem;
+import net.minecraft.item.ItemStack;
+
+/**
+ * An interface that allows implementing items to enable elytra flight for players, when equipped in the chest equipment slot.
+ * While being used, the item will be damaged <b>every 20 ticks</b> through {@link ItemStack#damage(int, LivingEntity, Consumer)}.
+ */
+public interface FabricFallFlyingItem {
```
modmuss50: This needs another name, FabricElytraItem seems a lot better to me. The yarn naming is a little odd, that needs fixing there where it can be changed easily.

modmuss50 (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/fallflying/FabricFallFlyingItem.java
haykam821: 'Fall flying' is both the Yarn name and the NBT name. I don't think it can be changed very easily, unless we want to overrule the NBT name and change both Yarn and Fabric API to use something else like 'gliding'.

Technici4n: `FabricElytraItem` and `allowElytraFlight` are also what I'd suggest - sorry for suggesting another name in the first place. :wink:

OroArmor: alright ill change the name

OroArmor: do we want `allowElytraFlight` or `shouldAllowElytraFlight`

OroArmor: went with `shouldAllowElytraFlight`

Technici4n: Perfect, this is good for me!

PR #1387, opened 2021-03-21
Delayed Tick Events

OroArmor:
could be expanded to include all tick events for the client and the server, but receiving feedback on the start is more important

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
@@ -25,7 +25,7 @@ import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
 
 /**
- * Allows the scheduling of events after a specified number of client ticks
+ * Allows the scheduling of events after a specified number of client ticks.
  */
 @Environment(EnvType.CLIENT)
 public class ClientTaskScheduler {
@@ -37,7 +37,7 @@ public class ClientTaskScheduler {
 	private static final List<Pair<Integer, List<ClientTickEvents.StartTick>>> executables = new ArrayList<>();
 
 	/**
-	 * Adds a new event to be run after the specified {@code delayTicks}
+	 * Adds a new event to be run after the specified {@code delayTicks}.
 	 * @param executable the event to be run
 	 * @param delayTicks the number of ticks to wait to run the event
 	 */
@@ -57,7 +57,7 @@ public class ClientTaskScheduler {
 	}
 
 	/**
-	 * Adds a new event to be run the next tick
+	 * Adds a new event to be run the next tick.
 	 * @param executable the event to be run
 	 */
 	public static void execute(ClientTickEvents.StartTick executable) {
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.event.lifecycle.v1;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import net.minecraft.client.MinecraftClient;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+/**
+ * Allows the scheduling of events after a specified number of client ticks
+ */
+@Environment(EnvType.CLIENT)
+public class ClientTaskScheduler {
+	static {
+		// Don't add this the the event handler until the class is loaded
+		ClientTickEvents.START_CLIENT_TICK.register(ClientTaskScheduler::runExecutables);
+	}
+
+	private static final List<Pair<Integer, List<ClientTickEvents.StartTick>>> executables = new ArrayList<>();
+
+	/**
+	 * Adds a new event to be run after the specified {@code delayTicks}
+	 * @param executable the event to be run
+	 * @param delayTicks the number of ticks to wait to run the event
+	 */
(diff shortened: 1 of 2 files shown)


amusingimpala75:
Looking at the code, there is a lot of similarity between the two sides, and there are duplicate Pair classes. Maybe try to simplify/combine the duplicated code?

OroArmor:
it definitely can be simplified, im just more concerned about the algorithm and make sure its not super expensive

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
```diff
@@ -0,0 +1,110 @@
...
+	 * Adds a new event to be run the next tick.
+	 * @param executable the event to be run
+	 */
+	public static void execute(ClientTickEvents.StartTick executable) {
+		execute(executable, 1);
+	}
+
+	private static void runExecutables(MinecraftClient client) {
+		for (Pair<Integer, List<ClientTickEvents.StartTick>> pair : executables) {
+			int ticksLeft = pair.getLeft() - 1;
```
dexman545, line 69: Decrementing all the items every tick seems inefficient - why not store a tickTime (tickDelta + currentTickTime at registration), then compare them at execution? Would also allow for MC's Pair to be used rather than the custom one.

maityyy:
I think you forgot to change the version of the module

i509VCB:
Ignore the version for now, it will be bumped a merge time.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
```diff
@@ -0,0 +1,110 @@
...
+ * Allows the scheduling of events after a specified number of client ticks.
+ */
+@Environment(EnvType.CLIENT)
+public class ClientTaskScheduler {
+	static {
+		// Don't add this the the event handler until the class is loaded
+		ClientTickEvents.START_CLIENT_TICK.register(ClientTaskScheduler::runExecutables);
+	}
+
+	private static final List<Pair<Integer, List<ClientTickEvents.StartTick>>> executables = new ArrayList<>();
```
liach, line 37: Yeah, you should use int to object hash map (either java, fastutil int map, or guava multimap) instead. 

Also I would add a field in this class indicating the current tick number and increment that field than decrementing all the integer keys of the list. Using that field is okay and it doesn't hurt from integer overflow either.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * Adds a new event to be run after the specified {@code delayTicks}.
+	 * @param executable the event to be run
+	 * @param delayTicks the number of ticks to wait to run the event
+	 */
+	public static void execute(ClientTickEvents.StartTick executable, int delayTicks) {
+		// See if there is already a pair with the same tick value
```
liach, line 45: Just use map computeIfAbsent once you track the current tick in a field.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
```diff
@@ -0,0 +1,110 @@
...
+		List<ClientTickEvents.StartTick> newList = new ArrayList<>();
+		newList.add(executable);
+		executables.add(new Pair<>(delayTicks, newList));
+	}
+
+	/**
+	 * Adds a new event to be run the next tick.
+	 * @param executable the event to be run
+	 */
+	public static void execute(ClientTickEvents.StartTick executable) {
```
liach, line 63: This is more like `schedule` imo. Execute feels instantaneous.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTaskScheduler.java
```diff
@@ -0,0 +1,110 @@
...
+	private static void runExecutables(MinecraftClient client) {
+		for (Pair<Integer, List<ClientTickEvents.StartTick>> pair : executables) {
+			int ticksLeft = pair.getLeft() - 1;
+
+			if (ticksLeft <= 0) {
+				for (ClientTickEvents.StartTick executable : pair.getRight()) {
+					executable.onStartTick(client);
+				}
+
+				pair.getRight().removeIf(startTick -> true);
```
liach, line 76: Dunno how you learned collections. You should call `clear` than call `removeIf` with a predicate that always return `true`

liach (changes requested):
The concept is good but the implementation is a bit weak

enbrain:
I think the best way to implement the idea is to use [`PriorityQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/PriorityQueue.html). This data structure allows you to create a task queue sorted by run tick (currentTick + delayTick at registration) and simplify your code to something like:

```java
PriorityQueue<ScheduledTask> tasks = ...;

void runTasks() {
  while (!tasks.isEmpty() && tasks.peek().getRunTick() == currentTick) {
    tasks.poll().run();
  }
}

class ScheduledTask {
  ...
}
```

Technici4n:
I am really wondering what the use case of this API would be, and why we need it in Fabric API?

OroArmor:
Someone had a thing where they wanted to run some code after a tick, so I thought of this. However, this current implementation doesn't have any way to save events after the game closes, so it is rendered useless pretty much. The client side could stay, but the server should not

Technici4n:
If you need to run stuff after a tick you can just have a `List<Runnable>` where you enqueue stuff for the next tick and a `BEFORE_SERVER_TICK` event handler...

liach:
> I think the best way to implement the idea is to use [`PriorityQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/PriorityQueue.html). This data structure allows you to create a task queue sorted by run tick (currentTick + delayTick at registration) and simplify your code to something like:
> 
> ```java
> PriorityQueue<ScheduledTask> tasks = ...;
> 
> void runTasks() {
>   while (!tasks.isEmpty() && tasks.peek().getRunTick() == currentTick) {
>     tasks.poll().run();
>   }
> }
> 
> class ScheduledTask {
>   ...
> }
> ```

a problem with priority queue is that it's not efficient for element removal. also wrapping each ticker with an int isn't particularly efficient either.

enbrain:
> a problem with priority queue is that it's not efficient for element removal. also wrapping each ticker with an int isn't particularly efficient either.

Do registration and removal occur so often? If so, your argument is valid. Now I wonder what the real use cases of this API are.

That aside, It's always good to perform benchmark tests if we really care efficiency.

Technici4n:
> Someone had a thing where they wanted to run some code after a tick, so I thought of this. However, this current implementation doesn't have any way to save events after the game closes, so it is rendered useless pretty much. The client side could stay, but the server should not

Yeah I definitely don't think there is a huge demand for this. I'd rather have other mods implement it and we can look into providing this feature if a lot of mods request it. I'm not going to close it but I don't think we should have it for the time being.

Issue #1373, opened 2021-03-13
Virtual Resource Pack API


LambdAurora:
tl;dr writing a virtual resource pack API that allows mod to write and read virtual resources and register the pack.

Useful for a lot of stuff like LambdaBetterGrass' texture generation, and other similar stuff.

Would likely only expose low-level stuff.

note: this issue primarily exist so it can be referred to when anyone talk about this subject, plus to assign it to myself.

liach:
Imo the resource loader would also change to generate one single pack (presumably modifying the builtin minecraft pack) for all mod-required resources, data, etc. In the process of building that single pack, we will invoke other resource/data suppliers registered by mods.

But one problem in my logic is that some resources need merging, and merging is inconsistent. For tags it may merge, for loot tables or recipes, those are total replacements. Need to find a consistent way for those, or if we can tweak the reloadable resoruce manager to accomplish it.

LambdAurora:
One of the other goal that isn't noted anywhere is to remove the Fabric Mods resource pack and let mod resource packs be in the Default resource pack.

Resource pack merging logic already exists in Fabric API anyway.

liach:
@LambdAurora I don't think your group resource pack works correctly. Say you group multiple packs that add to the same tag without replacing. Only the highest priority tag file will actually be loaded, and this is definitely wrong.

LambdAurora:
It doesn't?

Please read the PR that introduced them.

It keeps the behavior of `getResources` intact by special casing group resource packs.

If this was broken the PR would never have been merged in the first place!

liach:
Hmm, I plan to hack resource pack profile's factory instead...

Here is a few problems I spotted in your original pr. https://github.com/FabricMC/fabric/pull/1186#pullrequestreview-611757180 I personally ain't super fond of it (not as good as your recipe api before you added all those fancy stuff but better than the 1500-line updated recipe api)

PR #1371, opened 2021-03-13
Recipe API v1

LambdAurora:
*Being able to register recipes directly through code is amazing.*

So here it is! Recipe API v1!
It's basically a very simple API to register recipes through code.

## Why should I use this?

This allows you to register recipes without ever having to touch JSON, the way it is done avoid serializing recipes to JSON to end up deserializing them.  
It allows you to register recipes conditionally (in the case of mod compatibility) and do data-generation.

## What does it replace?

In Fabric API? Nothing.
In ARRP? The recipe stuff it has. Though this API is much more simpler as it only accepts `Recipe<?>` objects, and there is no helper for the Vanilla recipe types in this API, so there is still an interest into using ARRP but I would recommend ARRP to change how they handle recipes internally.

## But won't it break datapacks?

tl;dr no

The way this works is it will only truly register a recipe if nothing provides it already.

## What could be future plans for this module?

In the future this module could be extended to have more helpers to build Vanilla recipes.
Like an IngredientHelper to build a DefaultList of ingredients using a String pattern with keys and ingredient values.
But this won't be in this PR to keep it simple and lightweight (so easier to review)

## Testmod?

The testmod tests the static recipe registering and the dynamic registering, the static one is just a Stick -> Diamond shapeless recipe, the dynamic one is [Iron,Gold,Coal,Charcoal] -> randomized item at reload 4x4 shaped recipe.

PR #1389, opened 2021-03-22
Add StructurePoolAddCallback

frqnny:
Hello! Today this PR adds a simple StructurePoolAddCallback and a FabricStructurePool.

A few notes:

- The callback is called during data pack reloads (I think?)
- FabricStructurePool currently contains methods to add to a structure pool, however more methods can be added upon reviews/requests
- This is my first fabric pr, be gentle!


fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,56 @@
...
+
+	public void addStructurePoolElement(StructurePoolElement element) {
+		addStructurePoolElement(element, 1);
+	}
+
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		//adds to elementCounts list
+		List<Pair<StructurePoolElement, Integer>> list = new ArrayList<>(((StructurePoolAccessor) pool).getElementCounts());
+		list.add(Pair.of(element, weight));
+		((StructurePoolAccessor) pool).setElementCounts(list);
```
Juuxel: Creating a new list each time seems to be a bit wasteful IMO, maybe check first if it's already an `ArrayList`

frqnny:
don't review yet i need to eat and then fix checkstyle

fabric-structure-api-v1/build.gradle
```diff
@@ -1,2 +1,10 @@
 archivesBaseName = "fabric-structure-api-v1"
 version = getSubprojectVersion(project, "1.1.4")
```
frqnny, line 2: Should I update the subproject version and to what?

modmuss50: No, it generally gets handled after it’s been merged.

fabric-structure-api-v1/build.gradle
```diff
@@ -1,2 +1,10 @@
...
+
+minecraft {
```
Juuxel:
```suggestion
loom {
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,48 @@
...
+ * <p>Example usage:
+ * <pre>{@code
+ * StructurePoolAddCallback.EVENT.register(structurePool -> {
+ * 	if (structurePool.getStructurePool().getId().getPath().equals("village/common/butcher_animals")) {
+ * 		structurePool.addStructurePoolElement(StructurePoolElement.method_30425("village/common/animals/pigs_1").apply(StructurePool.Projection.RIGID), 2);
+ *    }
+ * });}
+ * </pre></p>
+ */
+
```
Juuxel:
```suggestion
```
Unnecessary empty line here

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	Event<StructurePoolAddCallback> EVENT = EventFactory.createArrayBacked(StructurePoolAddCallback.class,
+			listeners -> initialPool -> {
+				for (StructurePoolAddCallback listener : listeners) {
+					listener.add(initialPool);
+				}
+			}
+	);
+
+	void add(FabricStructurePool initialPool);
```
Juuxel: This doesn't really *add* a pool. I'd call this `onAdd` or something similar.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/MixinDynamicRegistryManager.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.util.registry.SimpleRegistry;
+
+import net.fabricmc.fabric.api.structure.v1.FabricStructurePool;
+import net.fabricmc.fabric.api.structure.v1.StructurePoolAddCallback;
+
+@Mixin(DynamicRegistryManager.class)
+public abstract class MixinDynamicRegistryManager {
+	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At(value = "TAIL"), cancellable = true, locals = LocalCapture.CAPTURE_FAILSOFT)
+	private static <E> void load(RegistryOps<?> ops, DynamicRegistryManager.Impl manager, DynamicRegistryManager.Info<E> info, CallbackInfo ci, RegistryKey<? extends Registry<E>> registryKey, SimpleRegistry<E> simpleRegistry) {
+		if (registryKey.getValue().toString().contains("template_pool")) {
```
Juuxel: This check is a bit concerning IMO. I think there should be a check for the registry first (whether it's the structure pool registry), instead of checking this first.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/MixinDynamicRegistryManager.java
```diff
@@ -0,0 +1,49 @@
...
+			for (Map.Entry<RegistryKey<Object>, Object> e : ((SimpleRegistryAccessor) simpleRegistry).getKeyToEntry().entrySet()) {
```
Juuxel: Since you're only using the values of the registry, you can just iterate the registry directly here instead of the map (I think registries are `Iterable` on their own).

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/MixinDynamicRegistryManager.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKey;
+import net.minecraft.util.registry.SimpleRegistry;
+
+import net.fabricmc.fabric.api.structure.v1.FabricStructurePool;
+import net.fabricmc.fabric.api.structure.v1.StructurePoolAddCallback;
+
+@Mixin(DynamicRegistryManager.class)
+public abstract class MixinDynamicRegistryManager {
+	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At(value = "TAIL"), cancellable = true, locals = LocalCapture.CAPTURE_FAILSOFT)
```
Juuxel:
```suggestion
	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
```
- `CAPTURE_FAILHARD` is better so that it doesn't launch the game with a broken API if this fails.
- No need for `cancellable = true` if you don't cancel

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,58 @@
...
+
+	public FabricStructurePool(StructurePool underlying) {
+		this.pool = underlying;
+	}
+
+	public void addStructurePoolElement(StructurePoolElement element) {
+		addStructurePoolElement(element, 1);
+	}
+
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
```
Juuxel, line 55: This needs documentation, especially about the `weight` parameter.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,58 @@
...
+		list.add(Pair.of(element, weight));
+		((StructurePoolAccessor) pool).setElementCounts(list);
+
+		//adds to elements list
+		for (int i = 0; i < weight; i++) {
+			((StructurePoolAccessor) pool).getElements().add(element);
+		}
+	}
+
+	public StructurePool getStructurePool() {
```
Juuxel: Odd naming, you don't really get a structure pool from a (Fabric) structure pool.

Juuxel (changes requested):

liach:
Imo the resource pack prebuilding api in #1373 should probably take care of this. The current setup will allow mods to modify user structure pools and cause unintended side effects, which #1371 the recipe modification api won't (and it's already roasted badly without that shortcoming)

frqnny:
I don't know enough about those PRs to properly persuade you to keep this, but the reasons for this callback to exist is to allow modifying the structure pools to allow for adding to vanilla structure pools as well as other mod's structure pools without having to add this code to your mod yourself. How do you think this could have unintended side effects?

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
frqnny: right... so then i'll just add to that list instead.

liach:
does this inject to the builtin default of the dynamic registries or after it's loaded from a resource manager? this isn't very clear from the code. if after it's loaded, then you will modify user pools as well and users cannot disallow your modifications unless you expose explicit hooks.

liach:
also the proposed resource pack building is far away, so no worries and i am not killing this pr. i just point out that this hurting user configuration might need to be taken care of.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
frqnny: should it be getUnderlying and make the field name "underlying"?

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/MixinDynamicRegistryManager.java
frqnny: got it, replaced it by checking with RegistryKey#isOf, will upload commit very soon once i finish all the other stuff you want

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
liach: You can probably make `FabricStructurePool` an interface that's injected by mixin onto regular structure pools, like https://github.com/FabricMC/fabric/blob/351679a7decdd3044d778e74001de67463bee205/fabric-loot-tables-v1/src/main/java/net/fabricmc/fabric/api/loot/v1/FabricLootPool.java#L31

frqnny:
How would I be able to address turning off those modifications with the current PR? 

Also, Juuz, I am working on getting a less concerning Mixin.

liach:
You should add in the javadoc of the event that mods may be editing on the structure pool from user configured resource pack than what's from the builtin minecraft resources. In fact, we cannot do much beyond that. If we really want it safe, we do need that resource prebuilding, which is, for now, very far away and is not an immediate alternative available.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package net.fabricmc.fabric.api.structure.v1;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * A callback for newly added structure pools.
+ * Word of warning: Mods may be editing on the structure pool from user configured resource pack
```
liach:
```suggestion
 *
 * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured resource pack
```
making the warning a bit more obvious by using bold and starting on a new paragraph (otherwise it is just a new sentence without line break)

liach (approved):
Looks adequate by this point

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+package net.fabricmc.fabric.api.structure.v1;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * A callback for newly added structure pools.
+ *
+ * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured resource pack
+ * than what's from the builtin minecraft resources.
```
Juuxel:
```suggestion
 * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured data packs
 * instead of the builtin Minecraft or mod resources.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>Example usage:
+ * <pre>{@code
+ * StructurePoolAddCallback.EVENT.register(structurePool -> {
+ * 	if (structurePool.getStructurePool().getId().getPath().equals("village/common/butcher_animals")) {
```
Juuxel:
```suggestion
 * 	if (structurePool.getUnderlying().getId().getPath().equals("village/common/butcher_animals")) {
```

Another minor thing: I'd also prefer checking the whole ID (it's possible that someone has a modded structure pool named `their_namespace:village/common/butcher_animals`. (Also applies to the test mod)

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class FabricStructurePool {
+	private final StructurePool underlying;
+
+	public FabricStructurePool(StructurePool underlying) {
+		this.underlying = underlying;
+	}
+
+	/**
+	 * Adds a new {@code StructurePoolElement} to the {@code StructurePool}.
```
Juuxel:
```suggestion
	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,68 @@
...
+	 * See the alternative {@linkplain #addStructurePoolElement(StructurePoolElement, int)}  for details.
```
Juuxel:
```suggestion
	 * See the alternative {@link #addStructurePoolElement(StructurePoolElement, int)} for details.
```
Better as a non-plaintext link

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.mojang.datafixers.util.Pair;
+
+import net.minecraft.structure.pool.StructurePool;
+import net.minecraft.structure.pool.StructurePoolElement;
+
+import net.fabricmc.fabric.mixin.structure.StructurePoolAccessor;
+
+/**
+ * Represents a modifiable structure pool that would have several helper methods for modders.
```
Juuxel:
```suggestion
 * Represents a modifiable structure pool that has several helper methods for modders.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,68 @@
...
+	 * See the alternative {@linkplain #addStructurePoolElement(StructurePoolElement, int)}  for details.
+	 *
+	 * @param element    The element you want to add.
+	 */
+	public void addStructurePoolElement(StructurePoolElement element) {
+		addStructurePoolElement(element, 1);
+	}
+
+	/**
+	 * Adds a new {@code StructurePoolElement} to the {@code StructurePool}.
```
Juuxel:
```suggestion
	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,68 @@
...
+
+		//adds to elements list
+		for (int i = 0; i < weight; i++) {
+			((StructurePoolAccessor) underlying).getElements().add(element);
+		}
+	}
+
+	/**
+	 * Allows you to get the {@code StructurePool} itself.
+	 * @return The underlying {@code StructurePool}.
```
Juuxel:
```suggestion
	 * Allows you to get the {@link StructurePool} itself.
	 * @return The underlying {@link StructurePool}.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+ * 		structurePool.addStructurePoolElement(StructurePoolElement.method_30425("village/common/animals/pigs_1").apply(StructurePool.Projection.RIGID), 2);
+ *    }
+ * });}
+ * </pre></p>
+ */
+public interface StructurePoolAddCallback {
+	/**
+	 * Called when Structure Pools are reloaded at datapack reload time.
```
Juuxel:
```suggestion
	 * Called when structure pools are reloaded at data pack reload time.
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+ * than what's from the builtin minecraft resources.
+ *
+ * <p>Example usage:
+ * <pre>{@code
+ * StructurePoolAddCallback.EVENT.register(structurePool -> {
+ * 	if (structurePool.getStructurePool().getId().getPath().equals("village/common/butcher_animals")) {
+ * 		structurePool.addStructurePoolElement(StructurePoolElement.method_30425("village/common/animals/pigs_1").apply(StructurePool.Projection.RIGID), 2);
+ *    }
+ * });}
+ * </pre></p>
```
Juuxel:
```suggestion
 * </pre>
```

Juuxel (changes requested):
Mostly javadoc notes

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
liach: Also this example should probably include some pseudocode that only registers this callback if a config entry is enabled (as otherwise users cannot turn off pool modifications from this mod on user pools easily)

frqnny:
I’m thinking about doing a method in FabricStructurePool called “hasIdOf”, a helper method to check if the id is the same as the string you provide.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * A callback for newly added structure pools.
+ *
+ * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured data packs
+ * instead of the builtin Minecraft or mod resources.
+ *
+ * <p>Example usage:
+ * <pre>{@code
+ * StructurePoolAddCallback.EVENT.register(structurePool -> {
+ * 	if (structurePool.getUnderlying().getId().getPath().equals("village/common/butcher_animals")) {
```
Juuxel:
```suggestion
 * 	if (structurePool.getUnderlying().getId().toString().equals("minecraft:village/common/butcher_animals")) {
```
To match the test mod

Juuxel (changes requested):

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
liach:
```java
private static final TARGET_POOL = new Identifier("minecraft", "village/common/butcher_animals")
...
if (structurePool.getUnderlying().getId().equals(TARGET_POOL)) {
```
shouldn't direct identifier comparison be better than string comparison?

Juuxel: Well yes, but this is just a javadoc example. That would be better for the test mod though.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+ * 		structurePool.addStructurePoolElement(StructurePoolElement.method_30425("village/common/animals/pigs_1").apply(StructurePool.Projection.RIGID), 2);
+ *    }
+ * });}
+ * </pre>
+ */
+public interface StructurePoolAddCallback {
+	/**
+	 * Called when structure pools are reloaded at data pack reload time.
+	 */
+	Event<StructurePoolAddCallback> EVENT = EventFactory.createArrayBacked(StructurePoolAddCallback.class,
```
BoogieMonster1O1, line 41: imo this should be replaced with an `event(Identifier)` method that returns a corresponding event factory for the structure pool id. 
It is convenient for modders as there's no extra `if` statement, and it is also quite uncommon to be adding the same pool element to two different structure pools.
`ScreenEvents` and the registry events do this

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/MixinDynamicRegistryManager.java
```diff
@@ -0,0 +1,45 @@
...
+
+import net.fabricmc.fabric.api.structure.v1.FabricStructurePool;
+import net.fabricmc.fabric.api.structure.v1.StructurePoolAddCallback;
+
+@Mixin(DynamicRegistryManager.class)
+public abstract class MixinDynamicRegistryManager {
+	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager$Impl;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private static <E> void load(RegistryOps<?> ops, DynamicRegistryManager.Impl manager, DynamicRegistryManager.Info<E> info, CallbackInfo ci, RegistryKey<? extends Registry<E>> registryKey) {
+		if (registryKey.equals(Registry.TEMPLATE_POOL_WORLDGEN)) {
+			for (StructurePool pool : BuiltinRegistries.STRUCTURE_POOL) {
```
BoogieMonster1O1: wouldn't this only work for builtin structure pools and not json ones?

frqnny: I think you are right about this, how would I fix that?

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
frqnny: I mean, you probably will have to add your element to multiple structure pools with villages that have many structure pools for each biome they spawn in. It is also convenient for keeping your if checks in one event rather than having several events.

frqnny:
Addressing the newest commits, I had to revert back to creating a new array list and setting the list on the structure pool because the elementCount list is normally an immutable list in minecraft.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,75 @@
...
+
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 *
+	 * @param element The element you want to add.
+	 * @param weight  Minecraft handles weight by adding it that amount of times into the StructurePool#elements.
+	 */
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		//adds to elementCounts list; minecraft makes these immutable lists so we replace them with an array list
+		List<Pair<StructurePoolElement, Integer>> list = new ArrayList<>(((StructurePoolAccessor) underlying).getElementCounts());
```
LambdAurora: I probably would do an instanceof on that list to know if it's actually useful to replace it with a mutable copy to avoid copying when already mutable.

LambdAurora (changes requested):

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
frqnny: would I check for an ArrayList or is there a way to get any mutable list

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,79 @@
...
+
+	public FabricStructurePool(StructurePool underlying) {
+		this.underlying = underlying;
+	}
+
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 * See the alternative {@link #addStructurePoolElement(StructurePoolElement, int)} for details.
+	 *
+	 * @param element The element you want to add.
```
LambdAurora:
```suggestion
	 * @param element the element to add
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public void addStructurePoolElement(StructurePoolElement element) {
+		addStructurePoolElement(element, 1);
+	}
+
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 *
+	 * @param element The element you want to add.
```
LambdAurora:
```suggestion
	 * @param element the element to add
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,79 @@
...
+	 * @param weight  Minecraft handles weight by adding it that amount of times into the StructurePool#elements.
```
LambdAurora:
```suggestion
	 * @param weight  Minecraft handles weight by adding it that amount of times into the {@link StructurePool#elements}
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,79 @@
...
+		//adds to elements list
+		for (int i = 0; i < weight; i++) {
+			((StructurePoolAccessor) underlying).getElements().add(element);
+		}
+	}
+
+	/**
+	 * Allows you to get the {@link StructurePool} itself.
+	 *
+	 * @return The underlying {@link StructurePool}.
```
LambdAurora:
```suggestion
	 * @return the underlying {@link StructurePool}
```

LambdAurora (changes requested):
javadocs `@param`, `@return` and some other like `@throw`, etc should not have a capitalized character as first character nor end with punctuation.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
Juuxel: ArrayList would probably be the best, since the only way to check for mutable lists is modifying and seeing whether it throws UnsupportedOperationException.

frqnny:
Gonna close this and see if I can make something better.

PR #1390, opened 2021-03-23
Allow getting BakedModels using Identifiers

PepperCode1:
If a model is loaded using an `ExtraModelProvider`, there is no way to retrieve it from the `BakedModelManager` after it is baked because it does not have a corresponding `ModelIdentifier`. This pull request adds a `ModelHelper` that allows retrieving a model via `Identifier` instead of `ModelIdentifier`. This works by exposing the full functionality of `BakedModelManager`'s internal model map, which stores `Identifier`s as keys, even though vanilla's retrieval method requires a `ModelIdentifier`. The same `Identifier` used to load the model can be used in `ModelHelper#getModel` to get the baked model.

### DIFF
--- a/fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelHelper.java
+++ b/fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelHelper.java
@@ -31,9 +31,9 @@ public final class ModelHelper {
 	 * Models loaded using {@link ExtraModelProvider} do not have a corresponding ModelIdentifier, so that method cannot be used to retrieve them.
 	 * The Identifier that was used to load them can be used in this method to retrieve them.
 	 *
-	 * @param manager The manager that holds models.
-	 * @param id The id of the model.
-	 * @return The model.
+	 * @param manager the manager that holds models
+	 * @param id the id of the model
+	 * @return the model
 	 */
 	@Nullable
 	public static BakedModel getModel(BakedModelManager manager, Identifier id) {
--- a/fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelHelper.java
+++ b/fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelHelper.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.model;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.render.model.BakedModel;
+import net.minecraft.client.render.model.BakedModelManager;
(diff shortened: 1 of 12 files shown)


fabric-models-v0/src/main/java/net/fabricmc/fabric/api/client/model/ModelHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+public final class ModelHelper {
+	/**
+	 * An alternative to {@link BakedModelManager#getModel(ModelIdentifier)} that accepts an {@link Identifier} instead.
+	 * Models loaded using {@link ExtraModelProvider} do not have a corresponding ModelIdentifier, so that method cannot be used to retrieve them.
+	 * The Identifier that was used to load them can be used in this method to retrieve them.
+	 *
+	 * @param manager The manager that holds models.
+	 * @param id The id of the model.
+	 * @return The model.
```
liach:
```suggestion
	 * @param manager the manager that holds models
	 * @param id the id of the model
	 * @return the model
```

liach (approved):

Technici4n (commented):
Looking good, will test in a bit!

What are the guarantees of this wrt. `bake` ordering? Is it guaranteed that model dependencies are already baked and can be safely retrieved using `getModel`, and would it be worth mentioning this in the javadoc?

Other small thing: `fabric-renderer-v1-api` already has a `ModelHelper`, it's probably not a big deal though, and `import static` can be used in case of conflicts. Thoughts?

PepperCode1:
The BakedModel map is retrieved all at once from the ModelLoader in `BakedModelManager#apply`, so this method (as well as the vanilla method) should only be used after the `BakedModelManager` has completed reloading.

I feel like there could be cases where both `ModelHelper`s might need to be used in the same class. Would a better name be `BakedModelHelper` or `ModelRetrievalHelper`, or maybe something else?

liach:
Imo finding a different name for helper class isn't the good way. #1366 should be the way to go, i.e. merging modules that are unnecessarily split.

PepperCode1:
I'm fine with merging this class, but does that mean this PR will have to wait until the module PR is merged?

OroArmor:
go ahead and work on this pr. if #1366 gets merged (which it will probably not) this can be changed, or if this is merged before, #1366 will make the changes to reduce work for you

LambdAurora (approved):

Technici4n:
> Imo finding a different name for helper class isn't the good way. #1366 should be the way to go, i.e. merging modules that are unnecessarily split.

No. This class is about baked model loading, the other one is about baked model rendering, they should be separate. `ModelLoadingHelper` maybe?

LambdAurora:
> > Imo finding a different name for helper class isn't the good way. #1366 should be the way to go, i.e. merging modules that are unnecessarily split.
> 
> No. This class is about baked model loading, the other one is about baked model rendering, they should be separate. `ModelLoadingHelper` maybe?

Doesn't look very loading to me, the only method present in it is called after all loading is done.

Technici4n:
> The BakedModel map is retrieved all at once from the ModelLoader in `BakedModelManager#apply`, so this method (as well as the vanilla method) should only be used after the `BakedModelManager` has completed reloading.

This would be worth adding to the javadoc imo. Trying to get another baked model during `bake()` is very tempting. :-P

> I feel like there could be cases where both `ModelHelper`s might need to be used in the same class.

Yeah.

Technici4n:
`BakedModelManagerExtensions`?

LambdAurora:
> `BakedModelManagerExtensions`?

`BakedModelManagerHelper` maybe? Would be consistent with resource loader with its `ResourceManagerHelper` but considering it's a v0 module I don't really know.

liach:
Maybe like `Models` or `BakedModels`

Technici4n:
`BakedModelManagerHelper` is good I think.

Technici4n:
image
Coolest test mod ever!!!! 🤠

Technici4n (approved):

Player3324 (approved):
LGTM, thanks!

PR #1391, opened 2021-03-24
Expose screen handler creation

PepperCode1:
This pull request adds two new methods to the `ScreenHandlerRegistry` that allow users to create `ScreenHandler`s without registering them. There is no simple alternative for this because vanilla's factory is private and `ExtendedScreenHandlerType` is in the impl package, meaning only the API itself should use it. Creating a `ScreenHandler` without registering it could be useful when, for example, making builders that create objects but those objects are registered later in a different class.

### DIFF
--- a/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/ScreenHandlerRegistry.java
+++ b/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/screenhandler/v1/ScreenHandlerRegistry.java
@@ -72,18 +72,42 @@ public final class ScreenHandlerRegistry {
 	}
 
 	/**
+	 * Creates a new {@code ScreenHandlerType} that creates client-sided screen handlers using the factory.
+	 *
+	 * @param factory the client-sided screen handler factory
+	 * @param <T>     the screen handler type
+	 * @return the created type object
+	 */
+	public static <T extends ScreenHandler> ScreenHandlerType<T> createSimple(SimpleClientHandlerFactory<T> factory) {
+		// Wrap our factory in vanilla's factory; it will not be public for users.
+		// noinspection Convert2MethodRef - Must be a lambda or else dedicated server will crash
+		return new ScreenHandlerType<>((syncId, inventory) -> factory.create(syncId, inventory));
+	}
+
+	/**
+	 * Creates a new {@code ScreenHandlerType} that creates client-sided screen handlers with additional
+	 * networked opening data.
+	 *
+	 * <p>These screen handlers must be opened with a {@link ExtendedScreenHandlerFactory}.
+	 *
+	 * @param factory the client-sided screen handler factory
+	 * @param <T>     the screen handler type
+	 * @return the created type object
+	 */
+	public static <T extends ScreenHandler> ScreenHandlerType<T> createExtended(ExtendedClientHandlerFactory<T> factory) {
+		return new ExtendedScreenHandlerType<>(factory);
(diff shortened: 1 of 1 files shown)


liach:
Hmm, I don't understand your use case. Why would you need the handler type to be outside of the registry?

> making builders that create objects but those objects are registered later in a different class.

is there anything that prevents you from registering and creating at the same time? If you split them, it would be more error-prone as handlers are useless (I recall) unless they are registered

liach:
Note: If you have some existing code that can be migrated to use this hook, feel free to share your code and we can see if this is the best approach for your problem.

Juuxel:
Note that the registration-only API in the screen handler API module was a request by @sfPlayer1 in Fabricord's #api channel, and I originally had this type of API:
> [16:20] Player: how does it end up in the actual registry btw?
[16:22] Juuz: Just a normal Registry.register. That's also in the new example in javadocs that you'll see when I copy the changes from fablabs
[16:23] Player: wouldn't it make sense for the api to do that internally?
[16:23] Player: like what else would you do with the type if not register it?
[16:35] Player: yes pretty sure we want to take the id in .simple/.extended and register it directly
[16:37] Player: which is pretty much what vanilla does in ScreenHandlerType with its register method likewise

Juuxel (approved):

liach:
Then I'd hear from player. I don't see the usage of standalone handler type creation as the standalone ones aren't shared on sides, which defeats the whole purpose of screen handler.

Juuxel:
With standalone creation you can handle *when* you register the stuff more easily. If you have your screen handler types in static final fields/object `val`s in Kotlin, the registration will happen when the class is loaded (which might happen too early). If the creation method can be separated, the modder can instead have a separate `init` method registering the content.

liach:
> 
> 
> With standalone creation you can handle _when_ you register the stuff more easily. If you have your screen handler types in static final fields/object `val`s in Kotlin, the registration will happen when the class is loaded (which might happen too early). If the creation method can be separated, the modder can instead have a separate `init` method registering the content.

then the solution should be registering to a cache before the registry is ready, like what we have for ber registrations

PepperCode1:
The code that uses this (I mixin-ed to Fabric API to add this functionality) isn't public yet, but what I said with the builders is basically what it does. There are builders for blocks, items, entities, fluids, screen handlers, and a few others. Objects are created through an abstract method, and then a callback (interface) does more processing on the object, and only registers it at the very end. Making a special case just for screen handlers in a setup like this is quite difficult, takes control away from the user, and makes the code base slightly messier.

With all objects that can be registered into vanilla's `Registry` class, `ScreenHandler` is the only one that must be registered at creation time, leading to an inconsistency. Technically, `BlockEntityType`s and `EntityType`s are also useless unless registered, but Fabric API doesn't automatically register them when the corresponding builder has its `build` method called.

I also feel like Player's argument of

> [16:37] Player: which is pretty much what vanilla does in ScreenHandlerType with its register method likewise

is invalid because most if not all of vanilla's initialization classes (`Blocks`, `Items`, etc.) have this method, but once again the restriction of automatically registering objects isn't imposed.

Technici4n:
I'm not sold... This is adding a cartesian product of functions for a very niche use case. Imo you can work around this "limitation" in your registration API.

liach (approved):

Technici4n:
@PepperCode1 we discussed on Discord that this is ok a while ago, with some name changes. And you wanted to do something similar for another registry. Any update?

Technici4n:
Let me know if you are going to work on this and I'll reopen.

PR #1392, opened 2021-03-24
Allow more flexibility when registering BERs

PepperCode1:
This pull request allows the `BlockEntityRendererRegistry` to accept `BlockEntityRenderer`s that render for a super type of the `BlockEntity` that the current `BlockEntityType` refers to. This makes sense, because one renderer can work for multiple types of `BlockEntity`s (that have the same super type) or an abstract renderer that has one type of `BlockEntity` that it renders for can change its behavior based on implementations of abstract methods. The workaround for this currently is to cast both arguments to raw types, which is not ideal.

### DIFF
--- a/fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/BlockEntityRendererRegistry.java
+++ b/fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/BlockEntityRendererRegistry.java
@@ -40,5 +40,5 @@ public interface BlockEntityRendererRegistry {
 	 *                            class is already loaded
 	 * @param <E> the {@link BlockEntity}
 	 */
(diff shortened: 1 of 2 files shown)


Technici4n (approved):
Looks good! I have a use case for this in my mod. Will test this in a bit and then after a few reviews it can be merged ASAP hopefully.

liach (approved):

BoogieMonster1O1 (approved):

frqnny (approved):

Juuxel (approved):

LambdAurora (approved):

Technici4n (approved):
Yup, it's working as expected. :)

ghost (approved):
Looks good, also tested it in production, my *existing* mod still works fine.

Player3324 (approved):
Good, thanks!

modmuss50 (approved):

PR #1393, opened 2021-03-26
Add fluid overlay sprite API

PepperCode1:
This PR adds the ability of defining custom fluid overlay sprite behavior to the `FluidRenderHandler`. This feature is needed because if a custom fluid is in the water tag, it will use the water overlay texture no matter what, which makes it look incorrect when behind glass or other similar transparent blocks. This change should not affect any existing code since the extra entry in the sprite array is optional. The javadoc has been updated to explain what an overlay sprite does, how it is used, and how to define custom behavior.

From the javadoc:

> The "overlay" sprite is rendered instead of the "flowing" sprite when behind
TransparentBlock or LeavesBlock. It is also only rendered on the
outside of the fluid block, whereas the other two sprites are rendered on both
sides. When the "overlay" sprite is used, only the top left quarter is actually
used for rendering.  This section is also reflected horizontally before rendering.

> return An array of size two or three. The first entry contains the "still" sprite,
while the second entry contains the "flowing" sprite. If the array is of size two,
the default behavior for the "overlay" sprite is used. If the array is of size three,
the third entry contains the "overlay" sprite. If it is null, no "overlay" sprite is
used.

### DIFF
--- a/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandler.java
+++ b/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandler.java
@@ -40,8 +40,7 @@ public interface FluidRenderHandler {
 	 * {@link TransparentBlock} or {@link LeavesBlock}. It is also only rendered on the
 	 * outside of the fluid block, whereas the other two sprites are rendered on both
 	 * sides. When the "overlay" sprite is used, only the top left quarter is actually
-	 * used for rendering. The section used is also reflected horizontally before
-	 * rendering.
+	 * used for rendering. This section is also reflected horizontally before rendering.
 	 *
 	 * @param view  The world view pertaining to the fluid. May be null!
 	 * @param pos   The position of the fluid in the world. May be null!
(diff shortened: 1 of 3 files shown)


grondag:
Letting fluids be swimmable without being in the water tag seems like a more worthwhile problem to solve.  Fluid overlay is not a great effect.

Azzyypaaras:
This tbh, a proper way to dictate entity swimming interactions rather than just tricking the game into thinking a fluid is water is a must. It could also allow for more complicated interactions like only allowing swimming based on predicates and affecting swimming speed.

liach:
imo you should make a new default method than change an existing method.

Technici4n:
Is this still needed given grondag's comment?

PepperCode1:
> Is this still needed given grondag's comment?

I would say yes, because the water tag does a lot more than just being swimmable, and separating all properties of the tag out is a lot of work with many mixins (I think Prospector was working on one). Plus, I'm not aware of a swimmable API that can be PRed or an already open PR, and this PR is already mostly done.

Technici4n:
Ok, why not indeed. Would you mind providing a few screenshots so that it's easier to visualize what this PR allows?

PepperCode1:
Some examples of this PR in action:

Default water:
water_default

Water with the overlay disabled:
water_no_overlay

Water with an overlay sprite of obsidian:
water_obsidian_overlay

Default lava:
lava_default

Lava with the same overlay sprite water uses by default (this same behavior happens when a custom fluid is added to the water tag):
lava_water_overlay

Lava with an overlay sprite of magma:
lava_magma_overlay

The overlay being only a quarter of the actual sprite is intended, because that's how the `FluidRenderer` displays it. This is explained in the javadoc that this PR adds. Also, I only used vanilla fluids here, but it does work with custom fluids as well.

liach (approved):

LambdAurora (approved):
Could always be useful one day or another.

Technici4n:
Superseded by #1687.

PR #1394, opened 2021-03-26
Add ServerPlayerEvents.ALLOW_DEATH event

williambl:
### PR Contents

This PR adds a new event, `ServerPlayerEvents.ALLOW_DEATH`. This is called when a player has just taken fatal damage. Mods can return `false` from this to cancel the death.

### Rationale

Mods may want to cancel a player's death in certain circumstances - [Gunpowder](https://github.com/Gunpowder-MC/Gunpowder/blob/c62a696eda68ca5648d78ed10779bf363efcb506/gunpowder-base/src/main/java/io/github/gunpowder/mixin/base/LivingEntityMixin_Base.java#L43) and [Haema](https://github.com/williambl/haema/blob/c7026d9ee18e9414b2a1e7eba8f033a68a2a44a8/src/main/java/com/williambl/haema/mixin/LivingEntityMixin.java#L46) are two mods which do.

Unfortunately, this requires a redirect mixin - however, this same redirect being needed for multiple mods (thus making them incompatible) suggests that it is appropriate for Fabric API.

CC: @martmists-gh

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -62,14 +62,14 @@ public final class ServerPlayerEvents {
 	 *     apply</li>
 	 * </ul>
 	 */
-	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
-		for (CancelDeath callback : callbacks) {
-			if (callback.cancelDeath(player, damageSource, damageAmount)) {
-				return true;
+	public static final Event<AllowDeath> ALLOW_DEATH = EventFactory.createArrayBacked(AllowDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
+		for (AllowDeath callback : callbacks) {
+			if (!callback.allowDeath(player, damageSource, damageAmount)) {
+				return false;
 			}
 		}
 
-		return false;
+		return true;
 	});
 
 	@FunctionalInterface
@@ -97,16 +97,16 @@ public final class ServerPlayerEvents {
 	}
 
 	@FunctionalInterface
-	public interface CancelDeath {
+	public interface AllowDeath {
 		/**
 		 * Called when a player takes fatal damage (before totems of undying can take effect).
 		 *
 		 * @param player the player
 		 * @param damageSource the fatal damage damageSource
 		 * @param damageAmount the damageAmount of damage that has killed the player
-		 * @return true if the death should be cancelled, false otherwise.
+		 * @return true if the death should go ahead, false otherwise.
 		 */
-		boolean cancelDeath(ServerPlayerEntity player, DamageSource damageSource, float damageAmount);
+		boolean allowDeath(ServerPlayerEntity player, DamageSource damageSource, float damageAmount);
 	}
 
 	private ServerPlayerEvents() {
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -68,6 +68,7 @@ public final class ServerPlayerEvents {
 				return true;
 			}
 		}
+
 		return false;
 	});
 
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -53,14 +53,14 @@ public final class ServerPlayerEvents {
 	 * <p>Mods can cancel this to keep the player alive.
 	 *
 	 * <p>Vanilla checks for player health <= 0 each frame (with {@link LivingEntity#isDead()}), and kills if true -
-	 * so the player will die next frame if this is cancelled. It's assumed that the listener will do something to
-	 * prevent this, for example:
+	 * so the player will still die next tick if this event is cancelled. It's assumed that the listener will do
+	 * something to prevent this, for example:
 	 *
-	 * <li>
-	 *     <ul>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</ul>
-	 *     <ul>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
-	 *     apply</ul>
-	 * </li>
+	 * <ul>
+	 *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
+	 *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+	 *     apply</li>
+	 * </ul>
 	 */
 	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
 		for (CancelDeath callback : callbacks) {
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -16,7 +16,6 @@
 
 package net.fabricmc.fabric.api.entity.event.v1;
 
-import net.fabricmc.fabric.api.util.TriState;
 import net.minecraft.entity.damage.DamageSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 
@@ -54,12 +53,13 @@ public final class ServerPlayerEvents {
 	 */
 	public static final Event<BeforeDeath> BEFORE_DEATH = EventFactory.createArrayBacked(BeforeDeath.class, callbacks -> (oldPlayer, damageSource, amount) -> {
 		boolean result = true;
+
 		for (BeforeDeath callback : callbacks) {
 			result &= callback.beforeDeath(oldPlayer, damageSource, amount);
 		}
+
 		return result;
 	});
-
 
 	@FunctionalInterface
 	public interface CopyFrom {
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -62,9 +62,9 @@ public final class ServerPlayerEvents {
 	 *     apply</ul>
 	 * </li>
 	 */
-	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (oldPlayer, damageSource, damageAmount) -> {
+	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
 		for (CancelDeath callback : callbacks) {
-			if (callback.cancelDeath(oldPlayer, damageSource, damageAmount)) {
+			if (callback.cancelDeath(player, damageSource, damageAmount)) {
 				return true;
 			}
 		}
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.api.entity.event.v1;
 
+import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.damage.DamageSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 
@@ -50,15 +51,24 @@ public final class ServerPlayerEvents {
 	 * An event that is called when a player takes fatal damage.
 	 *
 	 * <p>Mods can cancel this to keep the player alive.
+	 *
+	 * <p>Vanilla checks for player health <= 0 each frame (with {@link LivingEntity#isDead()}), and kills if true -
+	 * so the player will die next frame if this is cancelled. It's assumed that the listener will do something to
+	 * prevent this, for example:
+	 *
+	 * <li>
+	 *     <ul>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</ul>
+	 *     <ul>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+	 *     apply</ul>
+	 * </li>
 	 */
-	public static final Event<BeforeDeath> BEFORE_DEATH = EventFactory.createArrayBacked(BeforeDeath.class, callbacks -> (oldPlayer, damageSource, amount) -> {
-		boolean result = true;
-
-		for (BeforeDeath callback : callbacks) {
-			result &= callback.beforeDeath(oldPlayer, damageSource, amount);
+	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (oldPlayer, damageSource, damageAmount) -> {
+		for (CancelDeath callback : callbacks) {
+			if (callback.cancelDeath(oldPlayer, damageSource, damageAmount)) {
+				return true;
+			}
 		}
-
-		return result;
+		return false;
 	});
 
 	@FunctionalInterface
@@ -86,16 +96,16 @@ public final class ServerPlayerEvents {
 	}
 
 	@FunctionalInterface
-	public interface BeforeDeath {
+	public interface CancelDeath {
 		/**
-		 * Called when a player takes fatal damage.
+		 * Called when a player takes fatal damage (before totems of undying can take effect).
 		 *
 		 * @param player the player
-		 * @param source the fatal damage source
-		 * @param amount the amount of damage that has killed the player
-		 * @return whether or not the player should die
+		 * @param damageSource the fatal damage damageSource
+		 * @param damageAmount the damageAmount of damage that has killed the player
+		 * @return true if the death should be cancelled, false otherwise.
 		 */
-		boolean beforeDeath(ServerPlayerEntity player, DamageSource source, float amount);
+		boolean cancelDeath(ServerPlayerEntity player, DamageSource damageSource, float damageAmount);
 	}
 
 	private ServerPlayerEvents() {
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -52,7 +52,7 @@ public final class ServerPlayerEvents {
 	 *
 	 * <p>Mods can cancel this to keep the player alive.
 	 *
-	 * <p>Vanilla checks for player health <= 0 each frame (with {@link LivingEntity#isDead()}), and kills if true -
+	 * <p>Vanilla checks for player health {@code <= 0} each tick (with {@link LivingEntity#isDead()}), and kills if true -
 	 * so the player will die next frame if this is cancelled. It's assumed that the listener will do something to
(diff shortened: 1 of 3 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -45,6 +46,21 @@
...
+	/**
+	 * An event that is called when a player takes fatal damage.
+	 *
+	 * <p>Mods can cancel this to keep the player alive.
+	 */
+	public static final Event<BeforeDeath> BEFORE_DEATH = EventFactory.createArrayBacked(BeforeDeath.class, callbacks -> (oldPlayer, damageSource, amount) -> {
+		boolean result = true;
+
+		for (BeforeDeath callback : callbacks) {
+			result &= callback.beforeDeath(oldPlayer, damageSource, amount);
```
liach: hmm, does the callback gets called if the result is already false? and should it be called in that case?

williambl: They do, I was thinking about doing it either way but decided to do it this way (i.e. not short-circuiting) just to avoid surprising behaviour ('I registered this callback, why does it not always run?' etc.). If the consensus is that it's better short-circuiting, then I can change it to be done that way.

liach: It indeed does not short circuit, which is imo good enough. :+1:

liach:
I have a question: In your example, you give the player 3 health as the player holds apple on death. What if the player wears a helmet and a mod gives the helmet wearer like 5 health? We need a way to address such conflicts.

williambl:
> I have a question: In your example, you give the player 3 health as the player holds apple on death. What if the player wears a helmet and a mod gives the helmet wearer like 5 health? We need a way to address such conflicts.

I don't quite understand the issue - surely two mods could do that for basically any event, with the last-running event listener 'winning'?

The 'set to 3 health' thing is only in the example so that you don't die next tick due to health being below zero - there are other ways of preventing next-tick-death, such as giving a status effect, doing player.heal(), or by mixing into and modifying the logic used by isDead - I'd imagine many mods wouldn't be setting health directly.

liach (approved):
Sure, other mods can just inject in mods' handler methods if they want to make sure they register their listener after

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -69,6 +85,19 @@
...
+	public interface BeforeDeath {
+		/**
+		 * Called when a player takes fatal damage.
+		 *
+		 * @param player the player
+		 * @param source the fatal damage source
+		 * @param amount the amount of damage that has killed the player
+		 * @return whether or not the player should die
+		 */
+		boolean beforeDeath(ServerPlayerEntity player, DamageSource source, float amount);
```
Juuxel: `damage` would be a nicer and more descriptive name for `amount`

Juuxel (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
williambl: It seems a bit weird to have `source` but also `damage` - maybe `damageSource` and `damageAmount`?

Martmists-GH:
Gunpowder already offers events for this incompatibility, those might be a good point of reference.

Technici4n:
If the point of this event is to cancel player deaths, I think it should:
* be called `CANCEL_DEATH`,
* allow returning `true` to cancel death,
* short-circuit subsequent handling if `true` was returned so that death is not cancelled multiple times.

The javadoc should mention this, and also mention that this event is called before totems of undying have a chance to be used (and returning `true` will prevent totems of undying from being used).
I would also prefer `return true if ..., false otherwise` over `return whether ...`.

What happens if a handler cancels the death but doesn't change the player health? Does this leave the player in some kind of "undead" state, and should we guard against it?

williambl:
> If the point of this event is to cancel player deaths, I think it should:
> 
>     * be called `CANCEL_DEATH`,
> 
>     * allow returning `true` to cancel death,
> 
>     * short-circuit subsequent handling if `true` was returned so that death is not cancelled multiple times.
> 
> 
> The javadoc should mention this, and also mention that this event is called before totems of undying have a chance to be used (and returning `true` will prevent totems of undying from being used).
> I would also prefer `return true if ..., false otherwise` over `return whether ...`.

Sure, all makes sense to me.

> What happens if a handler cancels the death but doesn't change the player health? Does this leave the player in some kind of "undead" state, and should we guard against it?

Vanilla checks for health<=0 each frame (through isDead), and kills if true; so the player will die next frame. It's assumed that the listener will do something to prevent this, for example:
 - a server-side minigame mod teleporting the player into some dedicated 'respawn room' and setting their health to 20.0 again
 - a mod that redoes death mechanics switching the player over to the mod's own play-mode, where death doesn't apply
 - in the case of Haema, we bypass for vampires the vanilla code that checks for death on tick, so that vampires can survive on 0 health (as long as the fatal damage isn't done by certain damage sources, hence the need for this event)

Technici4n:
> > What happens if a handler cancels the death but doesn't change the player health? Does this leave the player in some kind of "undead" state, and should we guard against it?
> 
> Vanilla checks for health<=0 each frame (through isDead), and kills if true; so the player will die next frame. It's assumed that the listener will do something to prevent this, for example:

Ok, thanks! That is also worth adding to the javadoc imo, so someone willing to do something specific on death doesn't have to spend 3 hours reverse-engineering Minecraft code. :smile:

The content is good, I am just making sure that someone who is not very familiar with Minecraft's death handling can get a head start by reading the javadoc of the event. :)

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
Technici4n: Yeah why not have `damageSource` and `damageAmount`.

LambdAurora:
I think that if we have a `CANCEL` event we should have 2 other events, one which notifies that the death got cancelled and another one which notifies the death.
This is a thing commonly requested for events PR I believe.

Technici4n:
> I think that if we have a `CANCEL` event we should have 2 other events, one which notifies that the death got cancelled

How can the handler tell if it did cancel the death or not if there are multiple handlers that could have cancelled the death? Imo needs to be a single event that does both.
> and another one which notifies the death.

Possibly, but this is out of scope for this simple compatibility PR. (Imho)

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
williambl: Done

williambl:
Made a few changes in response to feedback - see latest commit message

Technici4n (approved):
Looks good, thanks! I will test this in a bit.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
```diff
@@ -17,24 +17,41 @@
...
 	private void onEntityKilledOther(DamageSource source, CallbackInfo ci, Entity attacker) {
 		// FIXME: Cannot use shadowed fields from supermixins - needs a fix so people can use fabric api in a dev environment even though this is fine in this repo and prod.
 		//  A temporary fix is to just cast the mixin to LivingEntity and access the world field with a few ugly casts.
 		ServerEntityCombatEvents.AFTER_KILLED_OTHER_ENTITY.invoker().afterKilledOtherEntity((ServerWorld) ((LivingEntity) (Object) this).world, attacker, (LivingEntity) (Object) this);
 	}
+
+	@Redirect(method = "damage", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/LivingEntity;isDead()Z", ordinal = 1))
+	boolean beforePlayerKilled(LivingEntity livingEntity, DamageSource source, float amount) {
+		if (livingEntity instanceof PlayerEntity) {
+			return isDead() && !ServerPlayerEvents.CANCEL_DEATH.invoker().cancelDeath((ServerPlayerEntity) livingEntity, source, amount);
```
ghost: Wouldn't it be safer change the instanceof check to `ServerPlayerEntity` here since that's what you cast `livingEntity` to.

```suggestion
		if (livingEntity instanceof ServerPlayerEntity) {
			return isDead() && !ServerPlayerEvents.CANCEL_DEATH.invoker().cancelDeath((ServerPlayerEntity) livingEntity, source, amount);
```

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -45,6 +47,30 @@
...
+	 *
+	 * <li>
+	 *     <ul>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</ul>
+	 *     <ul>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't
+	 *     apply</ul>
+	 * </li>
+	 */
+	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (oldPlayer, damageSource, damageAmount) -> {
+		for (CancelDeath callback : callbacks) {
+			if (callback.cancelDeath(oldPlayer, damageSource, damageAmount)) {
```
ghost: Imo I'd change `oldPlayer` to just `player` as that's what is used in the event interface method parameters.

```suggestion
	public static final Event<CancelDeath> CANCEL_DEATH = EventFactory.createArrayBacked(CancelDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
		for (CancelDeath callback : callbacks) {
			if (callback.cancelDeath(player, damageSource, damageAmount)) {
```

ghost (changes requested):
Looks good.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
williambl: Whoops, didn't catch that. Thanks.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
williambl: That's what I get for copy-pasting the copyplayerevent :sweat_smile:

ghost:
@williambl FYI, PR now fails checkstyle due to extra import and missing blank line, see https://github.com/FabricMC/fabric/pull/1394/checks?check_run_id=2231683199#step:5:1859
or run the checkstyle task I guess.

williambl:
Yeah, I'll fix it in the morning, I don't want to deal with checkstyle at 10pm :tiny_potato:

williambl:
Checkstyle is now passing :)

ghost (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -45,6 +47,31 @@
 		}
 	});
 
+	/**
+	 * An event that is called when a player takes fatal damage.
+	 *
+	 * <p>Mods can cancel this to keep the player alive.
+	 *
+	 * <p>Vanilla checks for player health <= 0 each frame (with {@link LivingEntity#isDead()}), and kills if true -
```
liach:
```suggestion
	 * <p>Vanilla checks for player health {@code <= 0} each tick (with {@link LivingEntity#isDead()}), and kills if true -
```
This shouldn't be frame. This is server player event, so it should be a tick. If it's a frame as in fps on client, move this to client events.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -45,6 +47,31 @@
...
+	 * so the player will die next frame if this is cancelled. It's assumed that the listener will do something to
```
liach:
```suggestion
	 * so the player will die next tick if this is not cancelled. It's assumed that the listener will do something to
```

liach (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
williambl: Whoops, thanks.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
williambl: Should be 'so the player will still die next tick if this event is cancelled' - I'll commit it when I'm at my computer.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -45,6 +47,31 @@
 		}
 	});
 
+	/**
+	 * An event that is called when a player takes fatal damage.
+	 *
+	 * <p>Mods can cancel this to keep the player alive.
+	 *
+	 * <p>Vanilla checks for player health {@code <= 0} each frame (with {@link LivingEntity#isDead()}), and kills if true -
```
liach:
```suggestion
	 * <p>Vanilla checks for player health {@code <= 0} each tick (with {@link LivingEntity#isDead()}), and kills if true -
```
Somehow....

williambl: merging skill: 0

liach (approved):

williambl:
A thought: is this redirect really needed? An inject into `tryUseTotem` could be used instead:
image

williambl:
Although, injecting into `tryUseTotem` means that the damage amount cannot be accessed. Doing this would work:
```java
LivingEntityMixin {
    private int damageAmountTmp;
}
```

```java
LivingEntity#damage(source, amount) {
[...]

if (this.isDead()) {
:INJECT_1
    damageAmountTmp = amount;
:END_INJECT_1
    if (!this.tryUseTotem(source)) {

[...]
}
```

```java
LivingEntity#tryUseTotem(source) {
:HEAD_INJECT
cir.setReturnValue(ServerPlayerEvents.CANCEL_DEATH.invoker().cancelDeath((ServerPlayerEntity) livingEntity, source, damageAmountTmp));
:END_HEAD_INJECT
}
```

liach:
Hmm, can you capture locals for redirects in mixin? Haven't used mixin in a while so not sure. If yes we probably can pass the damage if it's part of the local variables

williambl:
Possibly, but this way of doing it uses no redirects.

liach:
Imo we shouldn't design our api because the limits of mixin (such as that on redirects). If we lose too much valuable information if we make an alternative injection, then I'd prefer to still use a redirect.

Also, you can check out the current use case of mods and see how important is the damage value (we already got the damage source). If it's not important you can just drop the damage value info altogether.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -69,6 +96,19 @@
 		void afterRespawn(ServerPlayerEntity oldPlayer, ServerPlayerEntity newPlayer, boolean alive);
 	}
 
+	@FunctionalInterface
+	public interface CancelDeath {
```
modmuss50: Is this the best name? It can be used for more than just cancel'ing death.

I would have 2 events here, an AllowDeath and a BeforeDeath - this would allow mods to cancel death and for others to lisen to it knowing that the player is actually going to die. See the ScreenMouseEvents for a similar example.

modmuss50 (changes requested):

Technici4n:
I think having access to the damage is useful, e.g. if you want to add an armor that only allows kills with a low damage. I think the current iteration of the PR is fine.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
Technici4n: Imo this is out of scope, this PR is just a simple compat hook to allow multiple mods to cancel deaths in non-conflicting ways. I don't see why we need to debate additional events instead of just merging this simple addition.

modmuss50 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
```diff
@@ -17,24 +17,40 @@
...
+	public abstract boolean isDead();
+
 	@Inject(method = "onDeath", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/Entity;onKilledOther(Lnet/minecraft/server/world/ServerWorld;Lnet/minecraft/entity/LivingEntity;)V", shift = At.Shift.AFTER), locals = LocalCapture.CAPTURE_FAILEXCEPTION)
 	private void onEntityKilledOther(DamageSource source, CallbackInfo ci, Entity attacker) {
 		// FIXME: Cannot use shadowed fields from supermixins - needs a fix so people can use fabric api in a dev environment even though this is fine in this repo and prod.
 		//  A temporary fix is to just cast the mixin to LivingEntity and access the world field with a few ugly casts.
 		ServerEntityCombatEvents.AFTER_KILLED_OTHER_ENTITY.invoker().afterKilledOtherEntity((ServerWorld) ((LivingEntity) (Object) this).world, attacker, (LivingEntity) (Object) this);
 	}
+
+	@Redirect(method = "damage", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/LivingEntity;isDead()Z", ordinal = 1))
```
i509VCB, line 48: I believe there is a better injection for this? I will try to look into it since damage is a weird spot people forget to not call `super.damage` in.

modmuss50: Quite possibly, I imagine this is quite safe as its just the player.

williambl: > I believe there is a better injection for this? I will try to look into it since damage is a weird spot people forget to not call `super.damage` in.

I expressed this idea in above comments: https://github.com/FabricMC/fabric/pull/1394#issuecomment-814902555 https://github.com/FabricMC/fabric/pull/1394#issuecomment-814908478

Technici4n:
image
Working perfectly fine on dedicated servers (server is the test mod, and the client is vanilla), and I believe the scope, documentation and injection point are alright. Marking this as `merge-me-please`.

Player3324 (changes requested):
Thanks for the contribution, the changes are generally fine but Tech's naming comment unfortunately missed the conventions we already established in the screen API and a few other places.

Please rename the event field to ALLOW_DEATH, the interface name to AllowDeath and the interface method to allowDeath to be consistent with other APIs. This inverts the sementic meaning of the return value so documentation and the check in the invocation loop need to be adjusted.

Technici4n:
Hello @williambl, just wondering if you are going to be fixing this PR, or if someone else should do it?

williambl:
Sure. I'm not actually planning on using it myself anymore (I prefer injecting into canUseTotem (especially since I don't need damage amount)), but I suppose it'll be useful for others who want to do this without needing a mixin.

williambl:
Done.

Technici4n (approved):
Thanks, the update looks correct!

Player3324 (approved):
Good now, thanks!

Draylar (approved):

immibis:
Suggest rebasing these 11 commits down to 1

i509VCB:
Squash at merge time should be fine.

PR #1395, opened 2021-03-26
Conditional recipes - Revived from #1075

OroArmor:
Pr #1075 has fallen a bit to the wayside. This PR picks up where it left off. 

The changes I have made so far were switching the storage of `ResourceCondition`s to a registry and exposing the default fabric conditions publicly so that they can be used outside of the API.


fabric-conditional-resource-api-v1/src/main/java/net/fabricmc/fabric/api/conditionalresource/v1/ResourceConditions.java
```diff
@@ -0,0 +1,165 @@
...
+	 */
+	public static final Registry<ResourceCondition> RESOURCE_CONDITION_REGISTRY = FabricRegistryBuilder.createSimple(ResourceCondition.class, new Identifier("fabric-conditional-resource-api-v1", "conditions")).attribute(RegistryAttribute.SYNCED).buildAndRegister();
+
+	/**
+	 * An impossible {@link ResourceCondition}.
+	 */
+	public static final ResourceCondition IMPOSSIBLE = (fabricMetaId, element) -> false;
+
+	/**
+	 * An always true {@link ResourceCondition}.
```
frqnny, line 57: Are we sure we need an impossible condition and an always condition?

LambdAurora (changes requested):
I still maintain that conditional stuff should be handled by resource packs (cf built-in resource packs) with possibly a way to hide them from the user if needed to avoid too much pollution.

`WrappedResourcePack` breaks resource loader `ModResourcePack`.

Wrapping any resource pack coming by is terrible because of group resource packs.

And with the current implementation I am very worried about the slow down this will bring to resource loading with the indexation.

Technici4n:
> I still maintain that conditional stuff should be handled by resource packs (cf built-in resource packs) with possibly a way to hide them from the user if needed to avoid too much pollution.

I believe there is value in having conditional resources in general, this is easier to manage for modders and it can work nicely with datagen, but if this proves too hard or slow we can just resort to per-recipe and per-advancement conditions like Forge has.

liach:
Imo group resource pack is a bridgegap measure. We need #1373 to prebuild an actual pack from mod configurations (i.e. shipping existing jsons, generate new jsons based on the possible wood type-woodware types, and applying resource conditions), and that imo would solve most of our problems.

liach:
> `WrappedResourcePack` breaks resource loader `ModResourcePack`.
> 
> [Wrapping any resource pack coming by is terrible because of group resource packs.](https://github.com/FabricMC/fabric/pull/1395#pullrequestreview-622708415)

Unexpectedly, wrapping resource pack is exactly how vanilla handles format 3/4 resource packs on the client. This just proves the incompatibility of group resource pack imo. I think resource pack grouping should be instead done by a custom implementation of resource pack profile that can return a list of resource packs to resource pack managers.

LambdAurora:
> > `WrappedResourcePack` breaks resource loader `ModResourcePack`.
> > [Wrapping any resource pack coming by is terrible because of group resource packs.](https://github.com/FabricMC/fabric/pull/1395#pullrequestreview-622708415)
> 
> Unexpectedly, wrapping resource pack is exactly how vanilla handles format 3/4 resource packs on the client. This just proves the incompatibility of group resource pack imo. I think resource pack grouping should be instead done by a custom implementation of resource pack profile that can return a list of resource packs to resource pack managers.

What??

That's a bad idea, doing that would be regression!  
It would re-introduce the whole fucked up priority stuff that mod resource packs suffered before group resource packs were added.

OroArmor:
It might be better if this is included with the new resource loader that @LambdAurora is going to make, as integrating the systems is a must have, and trying to merge them after the fact will just cause issues

PR #1402, opened 2021-03-31
Fluid Flow API

OroArmor:
Adds an event when a fluid flows. Yes i know its in the object builder API, I'm waiting for #1136 to be merged, or if we merge these into one larger fluid api

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
@@ -30,11 +30,14 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Direction;
 import net.minecraft.world.World;
 
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
 public final class FluidFlowEvents {
 	private FluidFlowEvents() {
 	}
 
-	private static final Map<Block, Map<Block, List<Pair<Direction[], FluidFlowInteractionEvent>>>> EVENT_MAP = new HashMap<>();
+	private static final Map<Block, Map<Block, List<Pair<Direction, Event<FluidFlowInteractionEvent>>>>> EVENT_MAP = new HashMap<>();
 
 	/**
 	 * Registers a new event on a fluid flow. The same two blocks can register a different event for different directions, but the same directions will run the event that was registered first.
@@ -45,9 +48,30 @@ public final class FluidFlowEvents {
 	 * @param interactionEvent      The event to run when the conditions are met.
 	 */
 	public static void register(Block flowingBlock, Block interactionBlock, Direction[] interactionDirections, FluidFlowInteractionEvent interactionEvent) {
-		Map<Block, List<Pair<Direction[], FluidFlowInteractionEvent>>> flowBlockEvents = EVENT_MAP.getOrDefault(flowingBlock, new HashMap<>());
-		List<Pair<Direction[], FluidFlowInteractionEvent>> interactionEvents = flowBlockEvents.getOrDefault(interactionBlock, new ArrayList<>());
-		interactionEvents.add(new Pair<>(interactionDirections, interactionEvent));
+		Map<Block, List<Pair<Direction, Event<FluidFlowInteractionEvent>>>> flowBlockEvents = EVENT_MAP.getOrDefault(flowingBlock, new HashMap<>());
+		List<Pair<Direction, Event<FluidFlowInteractionEvent>>> interactionEvents = flowBlockEvents.getOrDefault(interactionBlock, new ArrayList<>());
+
+		if (interactionEvents.isEmpty()) {
+			for (Direction direction : Direction.values()) {
+				interactionEvents.add(new Pair<>(direction, EventFactory.createArrayBacked(FluidFlowInteractionEvent.class, fluidFlowInteractionEvents -> (flowingBlockState, interactingBlockState, flowPos, world) -> {
+					for (FluidFlowInteractionEvent event : fluidFlowInteractionEvents) {
+						if (!event.onFlow(flowingBlockState, interactingBlockState, flowPos, world)) {
+							return false;
+						}
+					}
+
+					return true;
+				})));
+			}
+		}
+
+		for (Pair<Direction, Event<FluidFlowInteractionEvent>> pair : interactionEvents) {
+			for (Direction direction : interactionDirections) {
+				if (pair.getLeft() == direction) {
+					pair.getRight().register(interactionEvent);
+				}
+			}
+		}
 
 		if (!flowBlockEvents.containsKey(interactionBlock)) {
 			flowBlockEvents.put(interactionBlock, interactionEvents);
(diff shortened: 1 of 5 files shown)


PepperCode1:
Does this API work for specifying interactions by fluid tag instead of block? Also, does it work for fluid blocks that do not extend `FluidBlock`?

OroArmor:
No, it doesnt work by tag, as each fluid tag should correspond to a block.
Also no, as it hooks directly into the fluid flow method of fluid block. I don't see why someone would make a fluid that doesn't extend fluid block

OroArmor:
OHH i see what you mean by tag, but with the current implementation, i don't think its easily addible

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
```diff
@@ -0,0 +1,101 @@
...
+						}
+					}
+				}
+			}
+		}
+
+		return null;
+	}
+
+	public interface FluidFlowInteractionEvent {
```
Juuxel, line 111: Should be named `FluidFlowInteractionCallback` to match other callbacks.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/FluidBlockMixin.java
```diff
@@ -0,0 +1,54 @@
...
+@Mixin(FluidBlock.class)
+public class FluidBlockMixin extends Block {
+	public FluidBlockMixin(Settings settings) {
+		super(settings);
+	}
+
+	@Inject(method = "receiveNeighborFluids", at = @At("HEAD"), cancellable = true)
+	private void receiveNeighborFluids(World world, BlockPos pos, BlockState state, CallbackInfoReturnable<Boolean> cir) {
+		for (Direction direction : Direction.values()) {
+			FluidFlowEvents.FluidFlowInteractionEvent event = FluidFlowEvents.getEvent(this, world.getBlockState(pos.offset(direction)).getBlock(), direction);
```
Juuxel: Might be a nice micro-optimisation and code clarification to cache `pos.offset(direction)` in a variable (and the block state too).

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * Gets the event from the following blocks and direction.
+	 *
+	 * @param flowingBlock         The fluid block that flowed.
+	 * @param interactionBlock     The block it interacts with.
+	 * @param interactionDirection The interaction direction
+	 * @return An event if the conditions are met, otherwise {@code null}
+	 */
+	public static @Nullable FluidFlowInteractionEvent getEvent(Block flowingBlock, Block interactionBlock, Direction interactionDirection) {
```
Juuxel: This is a bit concerning, IMO this should allow `Event`-like chaining of callbacks.

fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/FluidFlowEventTests.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.sound.SoundEvents;
+import net.minecraft.util.math.Direction;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.fluid.FluidFlowEvents;
+
+public class FluidFlowEventTests implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		FluidFlowEvents.register(Blocks.WATER, Blocks.BLUE_ICE, new Direction[]{Direction.DOWN}, (blockState, blockState2, blockPos, world) -> {
```
Juuxel: This should use the parameter names from the functional interface.

Juuxel (changes requested):
Some general comments about this PR:

- ~~The module for this API is wrong, it definitely does not belong in object builders.~~ nevermind, didn't read the PR desc carefully enough
- Using a `Map<Pair<Block, Block>, Event<...>>` could be a better solution: you would get event chaining for free as well as a more usual API. (I'd also change `register` to `Event<...> flowInteraction(Block, Block)` to match normal events)

OroArmor:
The reason i didnt use the current event system is that those are array-based, and this cant easily be array based unless i expose all the data, which would require each event to check if there conditions are met. this is simpler in the sense that you specify the interacting fluid and block to make your callback for

Juuxel:
You can have a map/collection of events matching to each condition, though. That doesn't require any checks in the callbacks.

OroArmor:
Ooooohh yeah i could

OroArmor:
> Using a Map<Pair<Block, Block>, Event<...>> could be a better solution:

I cant use a pair for the block/block matching as the Minecraft pair doesn't have an equals method

PepperCode1:
Personally, I think this API should be a callback that is not attached to anything else so that the user has maximum control and so that if a mod needs multiple interactions they can be put into the same callback for more efficiency.

Juuxel:
@OroArmor You *can* use a `Pair` from some other library or just a custom data class for this as it's an impl detail. I'd still prefer having a public API method returning an `Event` for a block+block+direction combination (or a general event like @PepperCode1 suggested), as that matches the style done everywhere else in FAPI.

OroArmor:
I dont see it hitting multiple callbacks very often, but in the case it does, its a nice to have

OroArmor:
Also i should make an impl class for this as there are way to many impl details in the API part

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
```diff
@@ -0,0 +1,123 @@
...
+	 * Registers a new event on a fluid flow. The same two blocks can register a different event for different directions, but the same directions will run the event that was registered first.
+	 *
+	 * @param flowingBlock          The fluid block that flowed.
+	 * @param interactionBlock      The block in one of the {@code interactionDirections}.
+	 * @param interactionDirections The direction to search for {@code interactionBlock}.
+	 * @param interactionEvent      The event to run when the conditions are met.
+	 */
+	public static void register(Block flowingBlock, Block interactionBlock, Direction[] interactionDirections, FluidFlowInteractionEvent interactionEvent) {
+		Map<Block, List<Pair<Direction, Event<FluidFlowInteractionEvent>>>> flowBlockEvents = EVENT_MAP.getOrDefault(flowingBlock, new HashMap<>());
+		List<Pair<Direction, Event<FluidFlowInteractionEvent>>> interactionEvents = flowBlockEvents.getOrDefault(interactionBlock, new ArrayList<>());
```
liach, line 52: You need to learn to use `computeIfAbsent`

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
```diff
@@ -0,0 +1,123 @@
...
+				}
+			}
+		}
+
+		if (!flowBlockEvents.containsKey(interactionBlock)) {
+			flowBlockEvents.put(interactionBlock, interactionEvents);
+		}
+
+		if (!EVENT_MAP.containsKey(flowingBlock)) {
+			EVENT_MAP.put(flowingBlock, flowBlockEvents);
```
liach, line 81: Had you used computeIfAbsent, you won't need these in the end

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/event/fluid/FluidFlowEvents.java
```diff
@@ -0,0 +1,123 @@
...
+	/**
+	 * Gets the event from the following blocks and direction.
+	 *
+	 * @param flowingBlock         The fluid block that flowed.
+	 * @param interactionBlock     The block it interacts with.
+	 * @param interactionDirection The interaction direction
+	 * @return An event if the conditions are met, otherwise {@code null}
+	 */
+	public static @Nullable Event<FluidFlowInteractionEvent> getEvent(Block flowingBlock, Block interactionBlock, Direction interactionDirection) {
+		if (EVENT_MAP.containsKey(flowingBlock)) {
```
liach, line 94: just check if the get result is null. extra contains check offers no real benefits

PR #1404, opened 2021-04-02
Make the client command registration match server commands with per world registration

OroArmor:
https://discord.com/channels/507304429255393322/507982478276034570/827246117867356250 for context.

Its really hard to make commands for server side or client side only. This pr switches to an event system that is very similar to the command API, but instead of passing a boolean for a dedicated server, it passes a boolean for a remote server (lan or dedicated). The dispatcher is reset each time the client joins a new world, but this is unavoidable with how brigadier is set up

### DIFF
--- a/fabric-command-api-v1/build.gradle
+++ b/fabric-command-api-v1/build.gradle
@@ -6,5 +6,6 @@ dependencies {
 }
 
 moduleDependencies(project, [
-		'fabric-api-base'
+		'fabric-api-base',
+		'fabric-networking-api-v1'
 ])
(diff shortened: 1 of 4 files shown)


Juuxel (changes requested):
- This is a breaking change to a stable API, and I don't see why it's needed. It would break existing mods.
- As far as I know, the remote server thing can be checked with a `requires` on the command node.
- The dispatcher-only design without events was (as usual for my PRs, it seems) requested by @sfPlayer1. I really don't see why this PR would be necessary.

OroArmor:
Ah i didnt think of the `.requires` method. That would work.

PR #1407, opened 2021-04-02
Update loom and gradle

modmuss50:
- Upgrading to loom 0.7 and the new run config system
- Update to gradle 7
- Start looking at singing the jars - currently just pgp. jar singing is TODO
- Build reproducable jars
- Build testmod jars - a fat jar depends on some loom work
- Incease default mem to 4G
- Enable parallel by default

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 	id "eclipse"
 	id "idea"
 	id "maven-publish"
-	id "fabric-loom" version "0.6.local" apply false
+	id "fabric-loom" version "0.7.local" apply false
 	id "org.cadixdev.licenser" version "0.5.0"
 	id "org.ajoberstar.grgit" version "4.1.0"
 	id "com.matthewprenger.cursegradle" version "1.4.0"
@@ -33,6 +33,7 @@ logger.lifecycle("Building Fabric: " + version)
 
 import org.apache.commons.codec.digest.DigestUtils
 import net.fabricmc.loom.task.RemapJarTask
+import net.fabricmc.loom.task.RemapSourcesJarTask
 
 def getSubprojectVersion(project, version) {
 	if (grgit == null) {
@@ -150,13 +151,33 @@ allprojects {
 	}
 
 	jar {
-		archiveClassifier.set "dev"
+		archiveClassifier = "dev"
 	}
+
+	task testmodJar(type: Jar) {
+		from sourceSets.testmod.output
+		archiveClassifier = "testmod-dev"
+	}
+
+	task remapTestmodJar(type: RemapJarTask, dependsOn: testmodJar) {
+		input=  testmodJar.getArchiveFile()
+		archiveClassifier = "testmod"
+		classpath sourceSets.testmod.compileClasspath
+		addNestedDependencies = false
+	}
+
+	task testmodSourceJar(type: Jar, dependsOn: classes) {
+		classifier = 'testmodSources'
+		from sourceSets.testmod.allSource
+	}
+
+	// Gets setup fully in loom
+	task remapTestmodSourcesJar(type: RemapSourcesJarTask)
 
 	afterEvaluate {
 		remapJar {
 			input = file("${project.buildDir}/libs/$archivesBaseName-${archiveVersion.get()}-dev.jar")
-			archiveFileName.set "${archivesBaseName}-${archiveVersion.get()}.jar"
+			archiveFileName = "${archivesBaseName}-${archiveVersion.get()}.jar"
 		}
 
 		artifacts {
@@ -175,6 +196,17 @@ allprojects {
 			header rootProject.file("HEADER")
 			include "**/*.java"
 		}
+
+		// Setup batched remap for testmod jars. Not really a supported api in loom so copy this at your own risk.
+		net.fabricmc.loom.configuration.RemapConfiguration.setupRemap(
+				project,
+				"testmodJar",
+				"testmodSourceJar",
+				"remapTestmodJar",
+				"remapTestmodSourcesJar",
+				"remapAllTestmodJars",
+				"remapAllTestmodSources"
+		)
 	}
 
 	task sourcesJar(type: Jar, dependsOn: classes) {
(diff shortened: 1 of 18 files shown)


liach (approved):

LambdAurora:
I thought loom 0.7 wasn't ready and gradle 7 is not even entirely released yet?

modmuss50:
Most of the work that went into 0.7 was around getting a project like fabric API working on gradle 7 and creating the jars for the testmods.

This will work on gradle 6 but by the time this is ready to be merged I expect that gradle 7 will be out, release of loom 0.7 depends on this being tested as well.

modmuss50:
This has now been updated to gradle 7.0 release

build.gradle
```diff
@@ -6,14 +6,16 @@ buildscript {
 
 plugins {
 	id "java"
+	id "java-library"
```
Juuxel, line 9:
```suggestion
	id "java-library"
```
`java-library` applies `java` (and I think Loom does too)

build.gradle
```diff
@@ -6,14 +6,16 @@ buildscript {
...
 	id "eclipse"
 	id "idea"
```
Juuxel, line 11: Loom also applies these two so they're not really needed.

build.gradle
maityyy: Loom is already using `java`, yes.

i509VCB: Java library is where you'll get the api configuration

modmuss50:
This is currenly broken with

```
Build file '/Users/mark/Documents/fabric-loom/fabric/build.gradle' line: 202

A problem occurred configuring project ':fabric-api-base'.
> Cannot change dependencies of dependency configuration ':api' after it has been included in dependency resolution.
```

The whole thing (down into loom as well) is a tottal mess of afterEvaulates expected to run in a certain order, im not sure how this even really worked before.

modmuss50:
After fixing that, another similar issue:

```
A problem occurred configuring root project 'fabric-api'.
> Could not resolve all dependencies for configuration ':compileClasspath'.
   > A problem occurred configuring project ':fabric-api-base'.
      > Failed to notify project evaluation listener.
         > Task with name 'remapAllSources' not found in root project 'fabric-api'.
         > Task with name 'remapAllTestmodSources' not found in root project 'fabric-api'.
```

modmuss50:
Im not really too sure how this ever worked before, and what to do about fixing it. A propper fix to make this work properly is months of loom work.

Its kinda blocking me being happy with 0.7 loom to call it stable

Technici4n:
Doesn't the problem lie within the fabric api buildscript? I'm not sure because I don't know how Gradle or Loom work.

liach:
Hmm, when was `remapAllSources` task registered? Does its registration depend on a lot of afterEvaluate code?

modmuss50:
Yes, the whole thing (this buildscript and loom) depends on a lot of fragile afterEvaluate code. Its made even harder by trying to enable parrell builds by default.

Toffikk:
I managed to build fabric-api with loom 0.7 and gradle 7.0 , but without signing the jars 
[build.gradle.txt](https://github.com/FabricMC/fabric/files/6370461/build.gradle.txt)

windows works too

modmuss50:
What is here is tempermental, I can always reproduct it on my laptop (8 core MacOS) after deleting the .fabric folder. Subsquent runs will work if the .fabric is there.

build.gradle
```diff
@@ -139,36 +151,62 @@ allprojects {
 	}
 
 	jar {
-		classifier = "dev"
+		archiveClassifier = "dev"
+	}
+
+	artifacts {
+		dev file: file("${project.buildDir}/libs/$archivesBaseName-${version}-dev.jar"), type: "jar", builtBy: jar
```
Toffikk, line 158: Use
${project.version} instead of ${version}

Toffikk (changes requested):

build.gradle
```diff
@@ -139,36 +151,62 @@ allprojects {
...
+
 	afterEvaluate {
 		remapJar {
-			input = file("${project.buildDir}/libs/$archivesBaseName-${version}-dev.jar")
-			archiveName = "${archivesBaseName}-${version}.jar"
-		}
-
-		artifacts {
-			dev file: file("${project.buildDir}/libs/$archivesBaseName-${version}-dev.jar"), type: "jar", builtBy: jar
+			input = file("${project.buildDir}/libs/$archivesBaseName-${archiveVersion.get()}-dev.jar")
```
Toffikk, line 183: Don’t use
${archiveVersion.get()} , use ${project.version}

Toffikk (changes requested):

build.gradle
```diff
@@ -139,36 +151,62 @@ allprojects {
...
+			archiveFileName = "${archivesBaseName}-${archiveVersion.get()}.jar"
```
Toffikk, line 184: The same as above

Toffikk (changes requested):

build.gradle
```diff
@@ -263,16 +312,29 @@ subprojects {
...
+	}
 }
 
-task remapMavenJar(type: net.fabricmc.loom.task.RemapJarTask, dependsOn: jar) {
+task remapMavenJar(type: RemapJarTask, dependsOn: jar) {
 	afterEvaluate {
-		input = file("${project.buildDir}/libs/${archivesBaseName}-${version}-dev.jar")
-		archiveName = "${archivesBaseName}-${version}-maven.jar"
+		input = file("${project.buildDir}/libs/${archivesBaseName}-${archiveVersion.get()}-dev.jar")
+		archiveFileName = "${archivesBaseName}-${archiveVersion.get()}-maven.jar"
```
Toffikk, line 328: The same as above for all cases

Toffikk (changes requested):

build.gradle
liach: I don't see anything in the scope that provides a `version`, and `project` is a default scope offered for `build.gradle` script.

Juuxel: Note that you can just have `$project.buildDir` and `$version` without the curly braces. (like `$archivesBaseName`)

Toffikk:
Maybe clone this PR Now?

Toffikk:
*close

Toffikk:
Because it’s outdated

modmuss50:
Yes, im going to close this. This PR does have some extra features such as building test mod jars but that can come later.

I based my changes in the snapshot from this PR 👍

PR #1421, opened 2021-04-16
Custom Elytras

Technici4n:
Reopening #1374 already. This is not final, I have to investigate if this addition will also work for non-player living entities, and document that behavior.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/AllowElytraFlight.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/AllowElytraFlight.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.entity.event.v1;
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ElytraItem;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.Util;
+import net.minecraft.world.event.GameEvent;
+
+import net.fabricmc.fabric.api.event.Event;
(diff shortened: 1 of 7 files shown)


Technici4n:
Updated with a remark, I checked this on a dedicated server and there was no issue, and I checked that it redirected every necessary call in the past. Once you agree, it can be marked as `merge-me-please` @liach @Juuxel.

liach (approved):

liach:
also quick question, how has the api's content mechanism changed?

OroArmor:
@Technici4n , after recent events, i would prefer if this pull request did not have me attached to it. thank you

Technici4n:
> @Technici4n , after recent events, i would prefer if this pull request did not have me attached to it. thank you

@OroArmor Want me to squash and remove you from the authors? Would that be OK?

OroArmor:
Yeah. That works

Technici4n:
First time using interactive rebase, I believe this is correct (the diff is empty).

Technici4n:
Needs an update to broadcast an `ALLOW_ELYTRA_FLIGHT` event, and use #1369 to allow easy implementation on items. There are lots of mods that could want to expose elytra flight behavior in other ways, such as Trinkets or a mod adding custom potions, etc...

Technici4n:
Ok, this is the result of another investigation for all uses of `Items.ELYTRA`:
* `ModelPredicateProviderRegistry#register`, specific to the elytra item.
* `ClientPlayerEntity#tickMovement`, responsible for starting the elytra flight. Need a hook there.
* `CapeFeatureRenderer` to stop capes from rendering when an elytra is worn. No need for an elytra hook there imo, we can have a generic `ALLOW_CAPE_RENDER` later and let modders use mixin for now.
* `ElytraFeatureRenderer` to render the elytra. This is specific to the elytra item, other modders can either hook there or register their own renderer.
* `EndTabAdvancementGenerator` to generate the advancement. No hook needed.
* `LivingEntity#onEquipStack` for the equip sound - I think we already have a hook for this. In all cases, no hook needed for elytra flight.
* `LivingEntity#initAi` for the actual elytra motion. This handles both the flight continuation check and the damage of the item. Need a hook there.
* `MobEntity#getPreferredEquipmentSlot` - no need for a hook.
* `PlayerEntity#checkFallFlying` - used by `ClientPlayerEntity#tickMovement` to start elytra flight, we can group both into one event dispatch.
* `EndCityGenerator.Piece` - no need for a hook.

TL;DR:
* We need an `ALLOW_ELYTRA_FLIGHT_START` **client-only** event to start the flight. **`ClientPlayerEntity` only.**
* We need a `CONTINUE_ELYTRA_FLIGHT` event on both the client and the server to continue the flight, and handle the damage of the item (or mana or whatever) **server-side only**. We will have to more or less overwrite the vanilla `initAi` function.  **Any `LivingEntity`.**

Technici4n:
And I just checked that LivingEntities can be fallflying, but won't enable it for themselves automatically. It can be activated with commands though (or by another mod), will be worth adding to the javadoc.

https://www.reddit.com/r/minecraftsuggestions/comments/6crat3/mobs_that_fall_while_wearing_an_elytra/

Technici4n:
Going to open a new one actually.

PR #1430, opened 2021-04-26
Fix trade offers by writing to the trade map directly

### LINKED ISSUE #1377: Trade helper bugs
The way fabric api works on trades, the trade reloading logic won't work.

The trade reloading logic is a bit too mod load order dependent. Take this example:

Mod registers via helper
Another mod registers directly to offer map
A third mod registers but via helper
2nd mod's trades drop. 

This means depreciating the existing api regarding reloading and depreciating the registration consumer.

This should just become a [level, profession] = trade.

I originally intended to add that api to allow for data reloadable trades but it wont really work.

### LINKED ISSUE #1427: TradeHelpers will duplicate trade entries like crazy
Not only was TradeHelpers[ nuking other mod's trades a thing](https://github.com/FabricMC/fabric/issues/1377), it seems the API is also broken as well. Using the latest Fabric API, I noticed that villagers would show multiple entries for the same trade even though I only added it once.
image

In this case with a custom villager, calling `TradeOfferHelper.registerVillagerOffers` multiple times to add trades to each trade level will duplicate all the other trade levels each time. Specifically, the issue is caused by [refreshVillagerOffers](https://github.com/FabricMC/fabric/blob/12865e786ce2102d344304a679b70084be84d166/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java#L80-L102) being called every time `TradeOfferHelper.registerVillagerOffers` is called by a mod. 

Here is a more detailed look at what is going on using a debugger. I am calling `TradeOfferHelper.registerVillagerOffers` 5 times in order to add trades to each trade level. Show here is the entries in `TradeOfferInternals.VILLAGER_TRADE_FACTORIES` at the start of refreshVillagerOffers.

After the 1st TradeOfferHelper.registerVillagerOffers call:
image

After the 2nd TradeOfferHelper.registerVillagerOffers call:
image

After the 3rd TradeOfferHelper.registerVillagerOffers call:
image

Notice how on the second call, level 1 trades were duplicated. And on the third call, level 1 and 2 trades were duplicated again. And it keeps growing every time it is called making the trade map bloated like crazy. Someone suggested it is because it doesn't do a deep copy of the original vanilla maps it is suppose to replace. Instead it is feeding itself over and over.
https://github.com/FabricMC/fabric/blob/12865e786ce2102d344304a679b70084be84d166/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/TradeOffersMixin.java#L42

However, a better solution for everyone is to trash the whole system and have the API just add/remove trades directly from vanilla's trade maps instead with none of these caching and reloadable system. It would both solve this issue and prevent the API from nuking other mod's trades when those mods do not use the API.

Technici4n:
This will just write to the underlying map directly. It might be too simple, did I miss something obvious?

Closes #1377.
Closes #1427.


haykam821:
Does the issue occur with wandering traders as well?

Technici4n:
> Does the issue occur with wandering traders as well?

The duplication one doesn't, because there is only one layer of map, and that was properly copied by the old implementation. The duplication issue for villagers arises from inner maps not being properly copied.

Still, this new implementation is much simpler and shares logic, whereas the old one was 75% copy/pasted.

And it will also stop nuking existing trade offers registered by a 3rd party, i.e. this will also fix the "ordering" issue with wandering traders.

frqnny:
Has this been tested against a mod that FAPI broke?

liach:
Imo instead of keeping track of original trades, we should keep track of the custom trades added through api endpoint. Then, for refreshing, we remove the custom trades added by api endpoint and call all adders to create custom trades again for us to track.

~~After a bit more thinking, now I think that we should patch the trade getting calls in the trade factories (than modifying those original arrays). This can make tracking api endpoint's added trades more easily, and can prevent edgy cases when there are identical trade factories from the api endpoint and in the original trades.~~

Nevermind, the fields are public maps, and my crossed out approach won't work.

Technici4n:
> Has this been tested against a mod that FAPI broke?

I tested this with the testmod by reproducing the issue and then fixing it.

Technici4n:
> Imo instead of keeping track of original trades, we should keep track of the custom trades added through api endpoint. Then, for refreshing, we remove the custom trades added by api endpoint and call all adders to create custom trades again for us to track.

What refreshing? It never worked properly, and we can't just invoke the registered `Consumers` repeatedly now because the old implementation didn't do that.
Trade offers are hardcoded in vanilla anyway, if mods want to implement correct refreshing, etc... they can do it themselves imo.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
```diff
@@ -29,94 +27,29 @@
...
-		final Int2ObjectMap<TradeOffers.Factory[]> trades = new Int2ObjectOpenHashMap<>(TradeOfferInternals.DEFAULT_WANDERING_TRADER_OFFERS);
-
-		for (int level : TradeOfferInternals.WANDERING_TRADER_FACTORIES.keySet()) {
-			// Get all registered offers and add them to current entries
-			final TradeOffers.Factory[] factories = TradeOfferInternals.WANDERING_TRADER_FACTORIES.get(level);
-			trades.put(level, ArrayUtils.addAll(factories, trades.computeIfAbsent(level, key -> new TradeOffers.Factory[0])));
-		}
+	// Shared code to register offers for both villagers and wandering traders.
+	// synchronized guards against concurrent modifications - Vanilla does not mutate the underlying arrays (as of 1.16),
+	// so reads will be fine without locking.
```
Technici4n: Not sure this is useful... In theory all registrations should happen during the init phase.

liach (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/trade/TradeOfferHelper.java
```diff
@@ -57,11 +57,9 @@ public static void registerWanderingTraderOffers(int level, Consumer<List<TradeO
 	}
 
 	/**
-	 * Refreshes the trade list by resetting the trade lists to vanilla state, and then registering all trade offers again.
-	 *
-	 * <p>This method is geared for use by mods which for example provide data driven villager trades.
+	 * @deprecated This never did anything useful.
 	 */
+	@Deprecated
 	public static void refreshOffers() {
```
modmuss50, line 63: Not a fan of having methods that seeming do nothing. At the very least log a warning.

As for actaully making this I will need to find time to figure out what it was supposed to do

modmuss50 (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/trade/TradeOfferHelper.java
Technici4n: Currently it's (expected to) remove from and then re-add all non-vanilla trade offers to the map.

Technici4n: But that's entirely useless if the registered `Consumer<List<TradeOffers.Factory>>` are not invoked again, and the old impl doesn't do that.

Technici4n: So either we make `refreshOffers` a no-op and let authors of data-driven trade mods remove their old factories from the map themselves, or we introduce a breaking change by potentially calling the consumers multiple times.

Technici4n: I would favor the former, with a warning in the log.

Technici4n: After discussion with i5 on Discord, the latter might be better. His plan was that `refreshOffers` would remove all the trade offers registered by this API and then re-invoke all the registered consumers.

Technically this would be a subtle behavior change that could be considered breaking, but given how buggy this API is I think it will be fine.

I will update the PR shortly.

Technici4n: Ok, do we really care about `refreshOffers`? It's annoying to implement, and we can't really catch some edge cases where other mods could duplicate entries and/or reorder the underlying arrays. I'd just deprecate it tbh, if mods want to replace trade entries they can just remove (it's a public map so it doesn't even require mixin, and we can add support later if necessary) and re-add trades.

Technici4n:
Updated with a warning when `TradeOfferHelper#refreshOffers` is called, for example from the test mod initializer:
```
[00:27:43] [main/WARN] (fabric-object-builder-api-v1) TradeOfferHelper#refreshOffers does not do anything, yet it was called! Stack trace:
 java.lang.Throwable: null
	at net.fabricmc.fabric.impl.object.builder.TradeOfferInternals.printRefreshOffersWarning(TradeOfferInternals.java:61) ~[fabric-api.fabric-object-builder-api-v1.main/:?]
	at net.fabricmc.fabric.api.object.builder.v1.trade.TradeOfferHelper.refreshOffers(TradeOfferHelper.java:64) ~[fabric-api.fabric-object-builder-api-v1.main/:?]
	at net.fabricmc.fabric.test.object.builder.VillagerTypeTest2.onInitialize(VillagerTypeTest2.java:51) ~[fabric-api.fabric-object-builder-api-v1.testmod/:?]
	at net.fabricmc.loader.entrypoint.minecraft.hooks.EntrypointUtils.invoke0(EntrypointUtils.java:50) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.fabricmc.loader.entrypoint.minecraft.hooks.EntrypointUtils.invoke(EntrypointUtils.java:36) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.fabricmc.loader.entrypoint.minecraft.hooks.EntrypointClient.start(EntrypointClient.java:32) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.minecraft.client.MinecraftClient.<init>(MinecraftClient.java:437) [minecraft-1.16.5-projectmapped-net.fabricmc.yarn-1.16.5+build.1-v2.jar:?]
	at net.minecraft.client.main.Main.main(Main.java:177) [minecraft-1.16.5-projectmapped-net.fabricmc.yarn-1.16.5+build.1-v2.jar:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_192]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_192]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_192]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_192]
	at net.fabricmc.loader.game.MinecraftGameProvider.launch(MinecraftGameProvider.java:224) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.fabricmc.loader.launch.knot.Knot.init(Knot.java:140) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.fabricmc.loader.launch.knot.KnotClient.main(KnotClient.java:26) [fabric-loader-0.10.5+build.213@net.fabricmc.yarn.1_16_5.1.16.5+build.1-v2.jar:?]
	at net.fabricmc.devlaunchinjector.Main.main(Main.java:86) [dev-launch-injector-0.2.1+build.8.jar:?]
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/trade/TradeOfferHelper.java
liach: > His plan was that `refreshOffers` would remove all the trade offers registered by this API and then re-invoke all the registered consumers.

Isn't this exactly my original suggestion :thinking: Imo you should just leave this no-op and log liek what you currently have.

In addition, for resource reloading, I personally think that mod can implement reload-aware trade offer factories than relying on this refresh to update things.

Technici4n: Updated with the log, please let me know if I should un-resolve this conversation.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
Technici4n: Note to myself: the `computeIfAbsent` call on line 39 needs to be `synchronized` as well.

Player3324 (approved):
LGTM, thanks!

Issue #1439, opened 2021-05-02
Memory leak in Networking API


Ampflower:
The networking API retains copies of the server when the integrated server is shut down.
This can lead to a substantial amount of memory left allocated as chunks are never properly unloaded.
Both the ServerLoginNetworkAddon and ServerPlayNetworkAddon contains a reference to MinecraftServer, and are never removed when the server shuts down.

Below are images of the heap dump showing where instances of MinecraftServer is retained unnecessarily. The dump is a bit noisy due to presence of other mods, but this can be fully reproduced with only the API base and networking API present.  
![image](https://user-images.githubusercontent.com/8528256/116829963-ff633f00-ab5b-11eb-8da9-fc466f9e52e2.png)
![image](https://user-images.githubusercontent.com/8528256/116829968-0a1dd400-ab5c-11eb-865a-554427c368d0.png)

## Steps to Reproduce
1. Load a world. Let it load then note the memory floor when a GC hits. (i.e. 57%)
2. Unload the world, then load it again. Let it load then note the memory floor when a GC hits.

## Expected Results
The memory to be at the same floor (57% in this example) no matter how many times you reload the same world.

## Actual Results
The memory jumps up a significant amount (roughly 3-5% for 4 GBs allocated with 32 chunk render distance). This can eventually lead to an unresolveable out of memory error condition when worlds continue to be reloaded.

Player3324:
Thanks for the report!

SP test with regular menu actions yields that

- the login -> play transition doesn't clean GlobalReceiverRegistry.trackedAddons up
- the client play disconnect event/processing doesn't fire, so no GlobalReceiverRegistry cleanup either
- only server play worked properly

Other actions may be incompletely covered as well, e.g. disconnects during the login stage or disconnects from the other side.

Player3324:
Should be fixed via https://github.com/FabricMC/fabric/pull/1440

PR #1440, opened 2021-05-03
Fix network api accumulating connection handlers and not firing some disconnect events

Player3324:
The existing networking-api implementation doesn't maintain GlobalReceiverRegistry.trackedAddons properly, accumulating stale Network(Handler)Addon instances. It also doesn't fire the disconnect event reliably, in particular the clientside disconnect event doesn't happen when leaving a SP game normally through the menu.

I fixed the issue by clearing LoginNetworkAddon instances when client/server transition from LOGIN to PLAY and by hooking the Netty-driven ClientConnection.channelInactive to capture all disconnects eventually. I left the regular disconnect hooks in to avoid potentially introducing extra delays in case the vanilla code doesn't close the connection immediately besides other potential regressions.

This should fix https://github.com/FabricMC/fabric/issues/1439


Ampflower:
Can confirm that this does fix #1439. Reloaded the world several times and saw that the minimum memory after GC was relatively the same each time. Also took a quick look at the heap dump via VisualVM and saw only one instance of server, which was the one that's currently live.

modmuss50 (approved):
LGTM.

This seems worthy of backporting to 1.14 and 1.15

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractNetworkAddon.java
```diff
@@ -44,6 +45,7 @@
 	// Sync map should be fine as there is little read write competition
 	// All access to this map is guarded by the lock
 	private final Map<Identifier, H> handlers = new HashMap<>();
+	private AtomicBoolean disconnected = new AtomicBoolean(); // blocks redundant disconnect notifications
```
Technici4n: `final` maybe?

Technici4n (approved):
LGTM, tested this on AOF3 and it seems to fix the issue in singleplayer, and multiplayer still works.

PR #1448, opened 2021-05-09
Change button highlight from blue tint to white outlines.

Jummit:
This is more in line with the new textures.
Preview:
highlight


ghost:
IMO for these buttons it's probably better to keep it as a blue tint as that is what vanilla still does for ingame gui buttons.
Or just remove the tint and they'll behave the same as creative inventory tab buttons.
In any case it's up to project maintainers / wider community to decide.
image

Technici4n:
In doubt, I don't think changing it is worthwhile, but you can put that in a resource pack if you want :)

Jummit:
Seems like the consensus is to keep it as-is, so closing. If Mojang decides to make buttons like these have a white outline this can be re-opened.

PR #1459, opened 2021-05-18
Deprecation: add ApiStatus Experimental to experimental api and forRemoval to dying api

liach:
Because java 16 made it a possibility.


haykam821:
Can the `@Deprecated` annotations be dropped?

liach:
> 
> 
> Can the `@Deprecated` annotations be dropped?

For experimental apis? I recall both modmuss and player opposed that. @modmuss50 @sfPlayer1 have your opinions changed or do you still think deprecated is needed?

Technici4n:
No they should be kept for better IDE support and warnings.

liach:
> No they should be kept for better IDE support and warnings.

not really, as those api don't have more stable variety. we don't want to encourage users to use mixins on theirvown to avoid deprecations imo

liach:
after all changing deprecated on ecp api isn't in the scope of this. we can look at that later

LambdAurora:
I still would like to not see FAPI to use `@Deprecated` when it's like the only API that does it even if experimental, I can't count anymore the number of modders who are confused about the biome modifications API because it's `deprecated` and considered for removal instead of experimental.

Technici4n:
https://softwareengineering.stackexchange.com/a/170029

That + `@ApiStatus.Experimental` should prevent any confusion.

LambdAurora:
> https://softwareengineering.stackexchange.com/a/170029
> 
> That + `@ApiStatus.Experimental` should prevent any confusion.

Still would cause confusion to those who don't read the direct source and just see a strikethroughed name :P

Technici4n:
If they don't read the source nor the decompiled version nor the documentation then how do they even use the API? ^^

modmuss50:
No, I think we should keep deprecated for experimental apis, it follows the same path that java uses. Why would they have added forRemoval then? Having the JB annotation on it makes it clear.

I do want to try and make some of the experimental apis stable.

The biome API is the big one that I know about however there is 1 major issue with it that needs resolving first (datapack biomes cannot be modified).

I am not aware of any issues or pending changes with the dimension API so I would say that can be made stable.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/FabricDimensions.java
```diff
@@ -30,6 +31,7 @@
  *
  * @deprecated Experimental feature, may be removed or changed without further notice due to potential changes to Dimensions in subsequent versions.
  */
+@ApiStatus.Experimental
 @Deprecated
```
modmuss50: I suggest we make this stable.

Technici4n: Removing dimensions still destroys worlds though. Not sure if that can be considered stable, or not? Idk

modmuss50: Thats a vanilla bug that has nothing to do with the API that is provided here (teleporting).

liach: Done.

liach:
Indeed, people can always use `@SuppressWarnings("deprecation")` to bypass non-removal deprecations.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/FabricDimensions.java
```diff
@@ -17,6 +17,7 @@
 package net.fabricmc.fabric.api.dimension.v1;
 
 import com.google.common.base.Preconditions;
+import org.jetbrains.annotations.ApiStatus;
```
modmuss50: This is breaking checkstyle

modmuss50 (approved):
LGTM, just fix checkstyle

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/FabricDimensions.java
liach: Done. I hope this can get merged soon (mainly because of that loom bump; without loom bump I cannot develop fabric or work on other prs)

Player3324:
Not sure if I like the forRemoval=true additions. The elements will likely get stripped through module version bumping, leaving the legacy module intact. In that case we'd mark the entire legacy module as deprecated for removal, but we technically don't actually remove the deprecated elements.

The outcome is similar, but it may allow us to reserve the forRemoval deprecation for cases that can't or won't be bridged?

liach:
> it may allow us to reserve the forRemoval deprecation for cases that can't or won't be bridged?

Afai see, in the JDK, they use `forRemoval` for both cases: in the removal of security manager in progress in java 17, https://github.com/openjdk/jdk/commit/576161d15423f58281e384174d28c9f9be7941a1 we can see that both parts with and without replacement (see https://openjdk.java.net/jeps/411 for details) are annotated with `forRemoval=true`.

The distinguishing of the two cases is done through the javadocs, namely the `@deprecated` tag. We should rely on Javadocs for this purpose as well, and leave `forRemoval` for binary compatibility (as in both cases, binary/runtime compatibility will break if they stay on the code, no matter there is a new alternative or not)

Player3324:
I am not talking about logical replacements in a different place (alternative APIs). The deprecated elements will stay around with their full signature in the legacy module (until that gets removed completely). The JDK doesn't use such a trick, it purges the elements and may sometimes provide an alternative to do the same. Likewise it may not be a bad idea to use forRemoval=true if and only if we plan on removing something directly and incompatibly, with a major version bump when that happens.

liach:
First, good point! If a deprecation means a project needs to change one of its dependencies to another one, `forRemoval` indeed isn't well attested.

However, a concern is that many developers treat all fabric apis as a whole dependency (look at the example mod). In this case, the deprecation of one api becomes much like the deprecation of a few packages, which is a bit concerning.

My proposal for this is to introduce dev and run-time variants of the overall fat api: we might consider removing the deprecated apis from compile-time dependencies but keep them in runtime to facilitate migration, if we do not add `forRemoval=true` to those outdated api modules.

Since we don't have this distribution scheme of making gradle ignore deprecated apis for dependencies, I personally still think `forRemoval=true` is a convenient bridgegap measure that we will drop once we get proper dependencies for mods using the whole api.

Player3324:
We can fairly easily make gradle ignore the legacy modules by providing a full and a non-legacy api build. The main issue with the extra legacy content is the clutter, harming discoverability and feature like auto-import selection, both if which get solved by using the non-legacy Fabric API dep.

The whole aspect is a rather different concern though, as far as the PR is concerned I am just saying that forRemoval=true is probably the wrong signal. I at least don't see anything particularly favorable in it, warnings etc don't change all that much.

Player3324:
Shouldn't https://github.com/FabricMC/fabric/pull/1459/files#diff-c414d777beffffb3ab5d7c0d5066cf4eae581531db12231e6fbf8c830ef356ceR37 etc be reverted as well? Unless there's a technical issue they'd likely be fixed through v2

liach:
> 
> 
> Shouldn't https://github.com/FabricMC/fabric/pull/1459/files#diff-c414d777beffffb3ab5d7c0d5066cf4eae581531db12231e6fbf8c830ef356ceR37 etc be reverted as well? Unless there's a technical issue they'd likely be fixed through v2

No? They should be just removed in a next major version like from 1.x to 2.x. They may be less meaningful for snapshots but definitely needed for long-term versions like major minecraft releases, where the modules may update more often than minecraft itself. We don't really make new modules that quickly yet!

But once we have decided a module will go die before it gets another major, the existing `forRemoval = true` will be removed. But it doesn't always happen.

Player3324:
For me the default is to always do this through module replacement, breaking changes including major version bumps are much harder to implement without poor outcomes. That means that it's unlikely that the deprecated elements would be removed in the strict sense, only if we have no other choice basically. Also note that I don't subscribe to using MC releases as justification to break our API, it just ends up piling even more work onto modders when they already have to deal with vanilla induced breakage.

liach:
Hmm, I don't quite agree. Some changes to a module doesn't always require a new module dependency imo. @modmuss50 your opinions on this?

liach:
@sfPlayer1 So what is the conclusion now? No deprecated API should ever get `forRemoval=true` just because either you will only allow writing a new API than killing existing bad apis slowly over time?

Player3324:
"No .. should ever" is a bit too strong, there are some exceptions:
- can't be bridged to a v2 (v_n+1) module
- too much effort to bridge for the impact
- severely broken anyway (think crash on use)
- harms performance or usability
- MC forces the change

So really not common practice, IMO it's much easier to copy the module in v2 and do some rewiring most of the time.

liach:
Thanks for this outline! Will do a quick review of the current forRemoval ones and update this pr.

kanpov:
> 
> 
> No, I think we should keep deprecated for experimental apis, it follows the same path that java uses. Why would they have added forRemoval then? Having the JB annotation on it makes it clear.
> 
> I do want to try and make some of the experimental apis stable.
> 
> The biome API is the big one that I know about however there is 1 major issue with it that needs resolving first (datapack biomes cannot be modified).
> 
> I am not aware of any issues or pending changes with the dimension API so I would say that can be made stable.

Stable Biome Modifications API would be a miracle.

liach:
withdrawing because in fact, most our apis aren't really deprecating for removal; they just wait for the api to be phased out

PR #1460, opened 2021-05-19
WIP 21w20a

modmuss50:
WIP 21w20a port, fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java will fail to apply.

Runs without the screen api, I ran out of time for this evening so leaving this here if anyone wants to pick up


Player3324:
I fixed and pushed the remainder

Issue #1269, opened 2021-01-15
Biome Modification API does not apply changes to datapack biomes.


TelepathicGrunt:
My mod Repurposed Structures utilizes the Biome Modification API for all of its structures. However, when quickly testing edge cases for my configs to make sure my structures are applied to the right biome, I used a datapack and found my changes never gets applied. This affects any biome overwritten by the datapack's json biome and also affects any brand new biome made by the datapack.

To reproduce, download Repurposed Structures and this datapack and create a new world with the datapack on.
https://www.curseforge.com/minecraft/mc-mods/repurposed-structures-fabric
[data.zip](https://github.com/FabricMC/fabric/files/5821883/data.zip)

The Overworld, Nether (these two dim have all their biomes overridden by datapack), and Custom Dimension (this dim has the custom brand new biome called new_plains) is what we are going to look at. If the Biome Modification API worked as intended, the following result for the locate command should be the following:

 **OVERWORLD**
 - /locate stronghold : **failed**
 - /locate repurposed_structures:stronghold_stonebrick : **found**
 - /locate repurposed_structures:stronghold_nether : **failed**
 
 **NETHER**
 - /locate stronghold : **failed**
 - /locate repurposed_structures:stronghold_stonebrick : **failed**
 - /locate repurposed_structures:stronghold_nether : **found**
 
 **CUSTOM**
 - /locate stronghold : **failed**
 - /locate repurposed_structures:stronghold_stonebrick : **found**
 - /locate repurposed_structures:stronghold_nether : **failed**

However, in reality, we find that the API does not do any changes and the result of the locate command is actually this with the datapack on.

 **OVERWORLD**
 - /locate stronghold : **found**
 - /locate repurposed_structures:stronghold_stonebrick : **failed**
 - /locate repurposed_structures:stronghold_nether : **failed**
 
 **NETHER**
 - /locate stronghold : **failed**
 - /locate repurposed_structures:stronghold_stonebrick : **failed**
 - /locate repurposed_structures:stronghold_nether : **failed**
 
 **CUSTOM**
 - /locate stronghold : **found**
 - /locate repurposed_structures:stronghold_stonebrick : **failed**
 - /locate repurposed_structures:stronghold_nether : **failed**

Biomes done by code seem to be fine. Next to test is whether other mods that make their own biomes by JSON within themselves also suffer from this issue with my mod using the API.

TelepathicGrunt:
[...]

PR #1463, opened 2021-05-22
Fixed Biome Modification API not working for json biome

TelepathicGrunt:
Issue report: https://github.com/FabricMC/fabric/issues/1269

It was what I thought. The set of "modified biome registrykeys" was failing and causing Biome Modification API to not apply changes to JSON biomes. In fact, the whole set is completely unnecessary too as each time the RegistryOps is fired, the biome instances inside are unique each time and have no changes from the other times RegistryOps fired. Thus, yeeting the set fixed the issue during my testing.

To double check this PR, use this datapack
[desert_test.zip](https://github.com/FabricMC/fabric/files/6527047/desert_test.zip)

I added a desert well once per chunk in the test mod to the vanilla desert biome. The very center top of the well shows what chunk spawned it when you show chunk boundaries with F3+G. 

Run the test without the datapack by creating a world. Do /locate desert and teleport to it to see the 1 well per chunk. Then exit that world and re-enter it and go to a new desert. See how it is still 1 well per chunk. Repeat these two tests but with the datapack on each time and the wells now still generate once per chunk.


fabric-biome-api-v1/src/testmod/java/net/fabricmc/fabric/test/biome/FabricBiomeTest.java
```diff
@@ -30,9 +30,7 @@
 import net.minecraft.world.biome.GenerationSettings;
 import net.minecraft.world.biome.SpawnSettings;
 import net.minecraft.world.gen.GenerationStep;
-import net.minecraft.world.gen.feature.ConfiguredFeatures;
-import net.minecraft.world.gen.feature.ConfiguredStructureFeatures;
-import net.minecraft.world.gen.feature.DefaultBiomeFeatures;
+import net.minecraft.world.gen.feature.*;
```
liach: mind use individual imports? and possbly add an actual data pack biome to the test

modmuss50:
Awesome 👍 I think I assumed the modified biomes was for the biome instance and not the reg key...

Idk if a better soultion would be to add a boolean to the biome object to note if its been modified? However it seems that check might not have even been needed in the first place?

TelepathicGrunt:
New problem, an edge case was missed. ~~When there's no json biomes, every biome gets remade as a new instance every time the RegistryOps fires.~~ (edit: nope, non-json biomes are always kept) When there is Json biomes, the json biome is remade every time the RegistryOps fires.

However, when there are json biomes present, the non-json biome that are not overridden gets their instances passed down through all 2-3 RegistryOps firing and only then. Thus, duplicating the biome modifications to those biomes.

The solution was keeping track of the biomes. But not with a set. Instead, it needs to be a list of biome instances so we know for sure the biome was processed once and only once. 

Now there is some good new, even the non-json biome instances are recreated when exiting and re-entering a world so we need to hook somewhere to know when we exit a world or finished the 2 or 3 RegistryOps firing in order to clear the modified biome list. Any ideas?

TelepathicGrunt:
Ok now it should be good. Just swapping the set of registrykeys to biome instances was enough to solve this and it was already being cleared out when re-entering worlds so there is no memory leak from the new biomes every time a person exits and re-enters worlds without quitting mc. 

Do note, while the test mod has the biome json in its data folder, it seems test mods data folders aren't loaded? But I included the datapack itself so you can just drag and drop it into the datapack screen upon world creation to do the test. The json overridden desert is flatter and has pink skies to show it is the json biome.

modmuss50 (approved):
Looks good to me 👍 

The build is failing due to checkstyle failiures.

Technici4n (approved):
Looks good, I will let someone with more biome/registry experience confirm that this is the correct fix.

Once this is merged, a subsequent PR can be made to stabilize this API I believe.

LambdAurora (approved):
Perfect!

Though I would recommend to not include the custom biome json in the mod resource pack directly has there's no way to disable mod data packs, I would recommend to use a built-in datapack instead in the test module to allow testing without the pack.

Juuxel (approved):

TelepathicGrunt:
> Perfect!
> 
> Though I would recommend to not include the custom biome json in the mod resource pack directly has there's no way to disable mod data packs, I would recommend to use a built-in datapack instead in the test module to allow testing without the pack.

What do you mean by a built-in datapack? My understanding is either the datapack json is in the mod's data folder which then is read as its own built-in datapack to the game. Or you manually drag the datapack into the datapack screen.

Juuxel:
You can use fabric-resource-loader-v0's `ResourceManagerHelper.registerBuiltinResourcePack` to add builtin but disabled-by-default mod data packs.

TelepathicGrunt:
I can't get the test mod to run if I attempt to use ResourceManagerHelper.registerBuiltinResourcePack as if it cannot ever resolve the imports even when the dependency is added. So instead, I just completely removed the test mod's data folder entirely. The zipped datapack itself is in the resource folder anyway and can just be drag and dropped into a world for the test. I am not gonna let this important PR be hanged up over whatever weird behavior testmods are doing with dependencies and internal data folders. If someone else wants to take up the pain of getting the testmod's data folder to work the way others want it to work, they can amend this PR. Otherwise, drag and dropping in the datapack itself is fine.

liach:
@modmuss50 mind approve running github actions for this?

liach (approved):

Player3324 (approved):
This Biome Set won't leak, right?

modmuss50:
I don’t think so as it will go out of scope with the dynamic registry.

I think a better solution to guarantee that will be to add a Boolean into Biome it’s self, but I don’t believe this is an issue.

liach:
can't we always use a resource reloader to clear the set after each data pack reload?

shartte:
```
        		// Vanilla will sometimes call RegistryOps.of twice for the same dynamic registry manager,    
        		// which usually will not result in a reload of objects, since it reuses existing objects
```
(if I remember correctly) refers to some edge cases in vanilla when you create a world, then re-create the world, it would call it twice for the same registry manager. It only applies to singleplayer worlds though. 
Regardless, you should modify the comment though, since now it doesn't match the actual code.

PR #1469, opened 2021-05-28
Energy v1

maximumpower55:
A Nice Fabric Alternative to Forge Energy


Technici4n:
We are not looking for an energy API at the moment. I would recommend using [Tech Reborn Energy](https://github.com/TechReborn/Energy), or [Fast Transfer Lib](https://github.com/Technici4n/FastTransferLib) (disclaimer: I am the author).

maximumpower55:
Ok I will check them out

Technici4n:
More precisely: there is no compelling reason to include an Energy API, as you can pick your library of choice and JiJ it - both are compatible with each other. However, if you want to contribute other additions, there is obviously no issue with that, just make sure to discuss a bit what you plan to do beforehand so we can point you to prior art. ;)

An energy API in particular is a recurring topic, and we are certain we don't want it in the API right now - that's why I closed the PR so fast.

Issue #427, opened 2019-11-17
(1.15) fabric-renderer-registries package name has a typo


Juuxel:
The package is called `rendereregistry` (missing an R between "rendere" and "registry").

modmuss50:
Not a huge amount we can do if it touches api classes tbh

Technici4n:
Will be closed when #1383 is implemented.

ethsanders:
Should be closed

PR #1470, opened 2021-06-01
Deprecate and move fabric-rendering-registries-v1 into fabric-rendering-v1

### LINKED ISSUE #1383: Deprecate `renderer-registries-v1` and copy the APIs into `rendering-v1`
It is confusing that those 3 registries are in another module, they should all be in `rendering-v1`.
This refactor is approved by @modmuss50. If someone wants to reduce the number of modules in Fabric API, this is your chance. 😄

altrisi:
Resolves #1383.

Of course keeps all public API methods and classes functional, though just they call the new API instead.

Used a Java 9 feature (private methods in interfaces) in order to register a handler to invoke the old event from the new one.


fabric-renderer-registries-v1/src/main/resources/fabric-renderer-registries-v1.mixins.json
```diff
@@ -3,11 +3,6 @@
   "package": "net.fabricmc.fabric.mixin.client.renderer.registry",
```
Technici4n: You can drop this file entirely (and remove it from the mod .json as well)

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/BlockEntityRendererRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+	 * Register a BlockEntityRenderer for a BlockEntityType. Can be called clientside before the world is rendered.
+	 *
+	 * @param blockEntityType the {@link BlockEntityType} to register a renderer for
+	 * @param blockEntityRendererFactory a {@link class_5614} that creates a {@link BlockEntityRenderer}, called
+	 *                            when {@link BlockEntityRenderDispatcher} is initialized or immediately if the dispatcher
+	 *                            class is already loaded
+	 * @param <E> the {@link BlockEntity}
+	 */
+	<E extends BlockEntity> void register(BlockEntityType<E> blockEntityType, BlockEntityRendererFactory<? super E> blockEntityRendererFactory);
+}
```
Technici4n: I'd say this is a good opportunity to move to a static method instead of an `INSTANCE` field, with the implementation in a static method in `BlockEntityRendererRegistryImpl`.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/EntityRendererRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+	 * Register a BlockEntityRenderer for a BlockEntityType. Can be called clientside before the world is rendered.
+	 *
+	 * @param entityType the {@link EntityType} to register a renderer for
+	 * @param entityRendererFactory a {@link EntityRendererFactory} that creates a {@link EntityRenderer}, called
+	 *                            when {@link EntityRenderDispatcher} is initialized or immediately if the dispatcher
+	 *                            class is already loaded
+	 * @param <E> the {@link Entity}
+	 */
+	<E extends Entity> void register(EntityType<? extends E> entityType, EntityRendererFactory<E> entityRendererFactory);
+}
```
Technici4n: Same comment here regarding moving away from `INSTANCE` fields in favor of static methods.

Technici4n:
Thanks for the PR! I noticed this is the chance to move away from the old `INSTANCE` pattern, if you don't mind.

fabric-renderer-registries-v1/src/main/resources/fabric-renderer-registries-v1.mixins.json
altrisi: Done.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/BlockEntityRendererRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ * Helper class for registering BlockEntityRenderers.
+ */
+public interface BlockEntityRendererRegistry {
+	BlockEntityRendererRegistry INSTANCE = new BlockEntityRendererRegistryImpl();
+
+	/**
+	 * Register a BlockEntityRenderer for a BlockEntityType. Can be called clientside before the world is rendered.
+	 *
+	 * @param blockEntityType the {@link BlockEntityType} to register a renderer for
+	 * @param blockEntityRendererFactory a {@link class_5614} that creates a {@link BlockEntityRenderer}, called
```
liach: mind update the jd here to refer to the right class?

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/EntityModelLayerRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+import net.fabricmc.fabric.impl.client.rendering.EntityModelLayerImpl;
+import net.fabricmc.fabric.mixin.client.rendering.EntityModelLayersAccessor;
+
+/**
+ * A helpers for registering entity model layers and providers for the layer's textured model data.
+ *
+ * @deprecated Experimental feature, may be removed or changed without further notice: Snapshot feature.
+ */
+@Deprecated
+@ApiStatus.Experimental
```
liach: can we raise this out of the exp state?

Technici4n: Most likely yes. I'm not aware of any issue with this addition, and it should be stable to have for 1.17 now.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/EntityModelLayerRegistry.java
```diff
@@ -58,8 +52,6 @@ private EntityModelLayerRegistry() {
 	}
 
 	@FunctionalInterface
-	@ApiStatus.Experimental
-	@Environment(EnvType.CLIENT)
```
liach: uhh, you removed an annotation incorrectly

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/BlockEntityRendererRegistry.java
altrisi: Done, however, why couldn't it be inlined directly in the API class, like it's done in other classes?

altrisi:
I've made the requested changes: Took EntityModelLayerRegistry out of experimental API and changed the `INSTANCE` pattern to a static method calling another static method in the Impl class.

My question is: Can't that be inlined and have the implementation directly in the API class, given it's a static method in a final class anyway? Other classes in the API already do that, including EntityModelLayerRegistry, that comes from the same module.

Technici4n:
Thanks!

I think inlining is fine if the internal logic is small enough that it doesn't hinder readability of the API class, which should be the case here. To be confirmed by @sfPlayer1.

liach:
i think you can just do that like what the jdk does.

altrisi:
Actually I don't think it's worth it, given the impl classes still have to exist to hide the setup methods and fields that mixins reference.

Will keep it like this unless you think it'd still make sense.

liach:
Sure, feel free to choose whatever impl as long as it is not noted to be inefficient and exposes only our desired contents.

altrisi:
(then this should be ready for reviews, triage request got removed after annotation comment)

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/LivingEntityFeatureRendererRegistrationCallback.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @param entityRenderer the entity renderer
+	 */
+	void registerRenderers(EntityType<? extends LivingEntity> entityType, LivingEntityRenderer<?, ?> entityRenderer, RegistrationHelper registrationHelper, EntityRendererFactory.Context context);
+
+	/**
+	 * A delegate object used to help register feature renderers for an entity renderer.
+	 *
+	 * <p>This is not meant for implementation by users of the API.
+	 */
+	interface RegistrationHelper {
```
Technici4n, line 74: You can add `@ApiStatus.NonExtendable` for this.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/EntityModelLayerRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+import net.fabricmc.fabric.impl.client.rendering.EntityModelLayerImpl;
+import net.fabricmc.fabric.mixin.client.rendering.EntityModelLayersAccessor;
+
+/**
+ * A helpers for registering entity model layers and providers for the layer's textured model data.
+ */
+@Environment(EnvType.CLIENT)
```
Technici4n, line 32: This one is `@Environment(EnvType.CLIENT)` and not the others. Should all of them have it?

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerRegistry.java
```diff
@@ -16,25 +16,18 @@
...
 
 /**
  * A helpers for registering entity model layers and providers for the layer's textured model data.
  *
- * @deprecated Experimental feature, may be removed or changed without further notice: Snapshot feature.
+ * @deprecated This module has been moved into fabric-rendering-v1. Use {@link net.fabricmc.fabric.api.client.rendering.v1.EntityModelLayerRegistry} instead.
  */
-@Deprecated
-@ApiStatus.Experimental
+@Deprecated(forRemoval = true)
```
Technici4n: Not sure about the use of `(forRemoval = true)`, as we don't have plans to remove it for the moment... it should not be a maintenance burden.

Technici4n (commented):
A few minor inquiries, otherwise this PR seems appropriate. I'll do a bit of local testing when I find the time this week.

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityModelLayerRegistry.java
liach: this should be, as jdk now calls forremoval as terminally deprecated

Technici4n: Hmmm, the documentation says that
> `forRemoval=true` indicates that the API is subject to removal in a future release.

So it should be fine to have them. Can be marked as resolved.

altrisi: I assumed this was the case since the issue was about reducing the number of modules, and they can't be reduced if it's not removed at some point in the future.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/LivingEntityFeatureRendererRegistrationCallback.java
altrisi: Done.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/EntityModelLayerRegistry.java
altrisi: Added them, makes sense since they all reference classes only present in the client.

altrisi (commented):
Made the requested changes.

(edit: wow github messed up the replies to the comments in here, see above for the actual conversations)

fabric-rendering-v1/src/testmod/resources/fabric.mod.json
```diff
@@ -10,7 +10,8 @@
   },
   "entrypoints": {
     "client": [
-      "net.fabricmc.fabric.test.rendering.client.WorldRenderEventsTests"
+      "net.fabricmc.fabric.test.rendering.client.WorldRenderEventsTests",
+      "net.fabricmc.fabric.test.renderer.registry.FeatureRendererTest"
```
Technici4n:
```suggestion
      "net.fabricmc.fabric.test.rendering.client.FeatureRendererTest"
```

Technici4n (approved):
Looks fine, and I confirm that the living entity feature renderer registration callback works with that small adjustment. Haven't been able to test the others though.

altrisi:
Bump, is there anything missing?

Player3324:
I'll have a closer look in a bit, `@Deprecated(forRemoval = true)` should be just `@Deprecated` since we technically don't remove the module but leave it as legacy

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/EntityRendererRegistry.java
```diff
@@ -26,7 +26,10 @@
 
 /**
  * Helper class for registering EntityRenderers.
+ *
+ * @deprecated This module has been moved into fabric-rendering-v1. Use {@link net.fabricmc.fabric.api.client.rendering.v1.EntityRendererRegistry} instead
  */
+@Deprecated(forRemoval = true)
 public interface EntityRendererRegistry {
 	EntityRendererRegistry INSTANCE = new EntityRendererRegistryImpl();
```
Player3324: may as well define this inline (anonymous class?), same in BlockEntityRendererRegistry

fabric-renderer-registries-v1/src/main/java/net/fabricmc/fabric/api/client/rendereregistry/v1/LivingEntityFeatureRendererRegistrationCallback.java
```diff
@@ -63,11 +62,26 @@
 	 */
 	void registerRenderers(EntityType<? extends LivingEntity> entityType, LivingEntityRenderer<?, ?> entityRenderer, RegistrationHelper registrationHelper, EntityRendererFactory.Context context);
 
+	private static Event<LivingEntityFeatureRendererRegistrationCallback> createEvent() {
+		var event = EventFactory.createArrayBacked(LivingEntityFeatureRendererRegistrationCallback.class, callbacks -> (entityType, entityRenderer, registrationHelper, context) -> {
```
Player3324: no `var` use please

Player3324 (commented):
Mostly good, thanks and sorry for the delay. Besides the deprecation change request from my above comment there are 2 minor details inline and I think the test mod in the old module should go, it only conflicts with the "new" copy.

altrisi:
Done: Removed old testmod, changed `var` and inlined the old registry implementations.

Also removed the `forRemoval = true` from `@Deprecated`, though I thought they would fit since the issue mentioned "reduce the number of modules in Fabric API", which isn't really possible without removing it at some point. There was also a bit of conversation about it in https://github.com/FabricMC/fabric/pull/1470#discussion_r650574661.

Player3324:
Thanks, it's effectively still reduced by becoming a legacy module. We'll change how they are exposed in the near future, so they don't bother devs anymore while still retaining compat for end users.

Player3324 (approved):

altrisi:
Fixed conflicts with recent additions to rendering-v1.

liach (approved):

liach:
@modmuss50 Mind check it out?

modmuss50 (approved):

kanpov (approved):

fabric-rendering-v1/src/main/resources/fabric-rendering-v1.mixins.json
```diff
@@ -8,7 +8,12 @@
     "MixinBuiltinModelItemRenderer",
     "MixinInGameHud",
     "MixinItemColorMap",
-    "MixinWorldRenderer"
+    "MixinWorldRenderer",
+    "EntityModelLayersAccessor",
+    "EntityModelsMixin",
+    "LivingEntityRendererAccessor",
+    "MixinBlockEntityRenderers",
+    "MixinEntityRenderers"
```
Juuxel, line 16: Preferably this would be sorted alphabetically like in other modules, but this is fine too.

Juuxel (approved):
Nice! This also will resolve #427 if the deprecated module is not kept around forever (which'll probably be the case...).

PR #1474, opened 2021-06-04
Fix Screen API ButtonList in 1.17

### LINKED ISSUE #1473: Screen API ButtonList broken since 21w20a
21w20a changed buttons to be a "drawable child" of screen, added to a screen by calling `addDrawableChild`.

```java
   protected <T extends Element & Drawable & Selectable> T addDrawableChild(T drawableElement) {
      this.drawables.add((Drawable)drawableElement);
      return this.addSelectableChild(drawableElement);
   }

   protected <T extends Element & Selectable> T addSelectableChild(T child) {
      this.children.add(child);
      this.selectables.add((Selectable)child);
      return child;
   }
```

The port of the screen API to 1.17-pre1 assumes buttons to be a "selectable child", not handling `this.drawables` leading to buttons being clickable and functioning but invisible.

While looking into this I found out that the tab order is handled by the order of `this.children`, the render order by the order of `this.drawables` while `this.selectable` has something to do with narration.

jackassmc:
`ButtonList` now properly handles `drawables`, `selectables` and `children`.

Closes https://github.com/FabricMC/fabric/issues/1473

### DIFF
--- a/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
+++ b/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
@@ -22,6 +22,7 @@ import java.util.List;
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.client.gui.Selectable;
+import net.minecraft.client.gui.Drawable;
 import net.minecraft.client.gui.Element;
 import net.minecraft.client.gui.widget.ClickableWidget;
 
@@ -32,56 +33,63 @@ import net.fabricmc.api.Environment;
 @ApiStatus.Internal
 @Environment(EnvType.CLIENT)
 public final class ButtonList extends AbstractList<ClickableWidget> {
-	private final List<Selectable> listeners;
+	private final List<Drawable> drawables;
+	private final List<Selectable> selectables;
 	private final List<Element> children;
 
-	public ButtonList(List<Selectable> listeners, List<Element> children) {
-		this.listeners = listeners;
+	public ButtonList(List<Drawable> drawables, List<Selectable> selectables, List<Element> children) {
+		this.drawables = drawables;
+		this.selectables = selectables;
 		this.children = children;
 	}
 
 	@Override
 	public ClickableWidget get(int index) {
-		return (ClickableWidget) listeners.get(translateIndex(index, false));
+		final int drawableIndex = translateIndex(drawables, index, false);
+		return (ClickableWidget) drawables.get(drawableIndex);
 	}
 
 	@Override
 	public ClickableWidget set(int index, ClickableWidget element) {
-		index = translateIndex(index, false);
-		int prevIndex = listeners.indexOf(element);
+		final int drawableIndex = translateIndex(drawables, index, false);
+		drawables.set(drawableIndex, element);
 
-		if (prevIndex >= 0 && prevIndex != index) {
-			if (prevIndex < index) index--;
-			listeners.remove(prevIndex);
-		}
+		final int selectableIndex = translateIndex(selectables, index, false);
+		selectables.set(selectableIndex, element);
 
-		int childIndex = children.indexOf(element);
-
-		if (childIndex >= 0) {
-			children.set(childIndex, element);
-		}
-
-		return (ClickableWidget) listeners.set(index, element);
+		final int childIndex = translateIndex(children, index, false);
+		return (ClickableWidget) children.set(childIndex, element);
 	}
 
 	@Override
 	public void add(int index, ClickableWidget element) {
-		index = translateIndex(index, true);
-
-		if (listeners.remove(element)) { // ensure no duplicates
+		// ensure no duplicates
+		final int duplicateIndex = drawables.indexOf(element);
+		if (duplicateIndex >= 0) {
+			drawables.remove(element);
+			selectables.remove(element);
 			children.remove(element);
-			index--;
+			if (duplicateIndex <= translateIndex(drawables, index, true)) { 
+				index--;
+			}
 		}
 
-		listeners.add(index, element);
-		this.children.add(element);
+		final int drawableIndex = translateIndex(drawables, index, true);
+		drawables.add(drawableIndex, element);
+
+		final int selectableIndex = translateIndex(selectables, index, true);
+		selectables.add(selectableIndex, element);
+
+		final int childIndex = translateIndex(children, index, true);
+		children.add(childIndex, element);
 	}
 
 	@Override
 	public ClickableWidget remove(int index) {
-		index = translateIndex(index, false);
+		index = translateIndex(drawables, index, false);
 
-		final ClickableWidget removedButton = (ClickableWidget) listeners.remove(index);
+		final ClickableWidget removedButton = (ClickableWidget) drawables.remove(index);
+		this.selectables.remove(removedButton);
 		this.children.remove(removedButton);
 
 		return removedButton;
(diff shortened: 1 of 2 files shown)


fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
```diff
@@ -91,30 +99,29 @@ public ClickableWidget remove(int index) {
...
 	}
 
-	private int translateIndex(int index, boolean allowAfter) {
+	private int translateIndex(List<?> list, int index, boolean allowAfter) {
 		int remaining = index;
 
-		for (int i = 0, max = listeners.size(); i < max; i++) {
-			if (listeners.get(i) instanceof ClickableWidget) {
+		for (int i = 0, max = list.size(); i < max; i++) {
+			if (list.get(i) instanceof ClickableWidget) {
```
liach, line 117: The original impl should have used a list iterator, but you don't have to worry about this.

liach (approved):
Looks right from my last glimpse on the related code. If you can provide like a screenshot then it's better.

jackassmc:
Here's a screenshot of the title screen:
Screenshot_2021-06-05_16-50-31

Here's a screenshot of the title screen with a tooltip, this prevents rendering the armor icon somehow:
Screenshot_2021-06-05_16-50-44

fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
```diff
@@ -32,56 +33,63 @@
...
-		listeners.add(index, element);
-		this.children.add(element);
+		final int drawableIndex = translateIndex(drawables, index, true);
+		drawables.add(drawableIndex, element);
+
+		final int selectableIndex = translateIndex(selectables, index, true);
+		selectables.add(selectableIndex, element);
+
+		final int childIndex = translateIndex(children, index, true);
+		children.add(childIndex, element);
```
Player3324, line 86: The many translateIndex calls look fishy. The class is constructing a button (=ClickableWidget) view on top of the underlying wider List, besides making sure that the additions/removals remain in sync with the other collections.

This means that there should be only one direct underlying List, the one exposed by this class. `translateIndex` should translate for that List exclusively, the other collections can be manipulated arbitrarily, their indices don't have to sync up in any way unless vanilla has further requirements there.

See the existing code which uses `listeners` as the underlying List, thus using `translateIndex` for all accesses to it. Accesses to `children` don't use translation, just presence. `selectables` likely just needs to be handled like `children` was.

jackassmc: I sync all lists because `drawables` determines the render order and `children` determines the tab order. I don't know exactly what `selectables` does but it has something to do with narration so I sync it just to be safe.

Player3324: You can only do this if the lists were in sync before though, otherwise the `index` becomes unrelated to the list in question.

Player3324: What you might be able to do instead is scanning for a common predecessor, but it looks like the API isn't mighty enough to handle this properly and correctly in all cases.

jackassmc:
```java
   protected <T extends Element & Drawable & Selectable> T addDrawableChild(T drawableElement) {
      this.drawables.add((Drawable)drawableElement);
      return this.addSelectableChild(drawableElement);
   }

   protected <T extends Element & Selectable> T addSelectableChild(T child) {
      this.children.add(child);
      this.selectables.add((Selectable)child);
      return child;
   }
```
The lists are somewhat in sync, if there are a lot of Selectables that aren't Drawable and they're before a Drawable and that Drawable is set those Selectables can end up after the Drawable.

My goal is to ensure buttons are correctly positioned relative to other buttons in all lists.

liach: Per screen impl detail, if two elements are present in two of the lists, then they are guaranteed to be in the same order in relative to each other in both lists. Maybe a better approach is to use a list of any object as a backing list, and patch the list of drawable/selectable/elements as the shadows on that backing list.

jackassmc: So what should I change here? I don't understand how buttons can have the same relative order without the searching for buttons done in `translateIndex`.

liach: you can create another list as the final reference for indexes for all types of elements, and the order in that list should be matched by other lists

jackassmc: > create another list as the final reference for indexes for all types of elements

That already exists, it's `children`. It has all `Element`s in it. Then there is `selectables` which has all `Element`s that are also `Selectable` and finally `drawables` which has all `Element`s that are `Selectable` and `Drawable`.

> the order in that list should be matched by other lists

In vanilla the order of these three lists is the same, the only difference is that there are more elements in `children` than in `selectables` and more in `selectables` than in `drawables`.

fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
```diff
@@ -32,56 +33,63 @@
...
 	}
 
 	@Override
 	public ClickableWidget remove(int index) {
-		index = translateIndex(index, false);
+		index = translateIndex(drawables, index, false);
```
liach, line 91: then this translate should happen on children than drawables as we use children as the final reference

fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
liach: then this index call should be fine.

fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java
jackassmc: The order in the rest of the file is `drawables` then `selectables` then `children`, from most specific to least specific.

Changing this to use `children` wouldn't change the way it works because the index is only used once to remove the element at the index: https://github.com/FabricMC/fabric/blob/188baa9af807a9371ac3ce6bce56c68ec833057c/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java#L89-L91

Then the removed element is used to remove from the other lists: https://github.com/FabricMC/fabric/blob/188baa9af807a9371ac3ce6bce56c68ec833057c/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/impl/client/screen/ButtonList.java#L92-L93

So changing it would just make this function inconsistent with the rest of the file.

liach (approved):
should be good. imo we should really stop using such lists.

LambdAurora (approved):

liach:
@sfPlayer1 I think this fix is adequate as of now. A better solution is a new way to handle buttons from the current exposed list, but this is out of the scope of this pr.

Player3324 (approved):
I still think that projecting the index onto other Lists is fragile, but it's apparently fine as long no mod manpulates them directly. We can maybe tweak it another time.

Thanks!

modmuss50 (approved):
LGTM

Player3324:
The PR has a few checkstyle violations, please try a local build and/or configure your IDE with Fabric-API's checkstyle.xml to resolve them.

jackassmc:
I fixed all the checkstyle violations but the workflow needs to be approved again.

PR #1484, opened 2021-06-10
Added support for VillagerEntity task lists.

ByThePowerOfScience:
### Adds:

- `VillagerTaskProvider`: A handler for dynamically adding tasks to a villager's task-list based on their `VillagerProfession`.  Essentially extends the hardcoded `net.minecraft.entity.ai.brain.VillagerTaskListProvider` and makes it able to adapt based on profession.  Two types of tasks can be changed: constant (added to the main list) and random (added to the list passed to that task-list's RandomTask).  Task "category" is chosen based on an enum - each value corresponding to a different `create<X>Tasks` method in `VillagerTaskListProvider`.
- Base!`VillagerTaskProvider`: Allows modders to change the tasks of _every_ villager, regardless of profession.
- `VillagerTaskProviderRegistry`: Registers `VillagerTaskProvider`s, mapped to `VillagerProfession`.  If multiple task-lists are submitted for the same profession, the lists will be merged.
- `VillagerTaskProviderInternals`: Allows `VillagerTaskListProviderInjector` to retrieve entries from the `VillagerTaskProvider` registry at runtime.  The Profession-TaskProvider map is retrieved from `VillagerTaskProviderRegistry` during startup, converted to an `ImmutableMap`, and stored as a static property of this class.
- `WorkerVillagerTask`: A framework for creating simple world-altering `VillagerTask`s.  Includes handler methods for scanning the surrounding area, registering and selecting a random target, and adding a look/walk target.  Includes hooks for run conditions, executing world actions, duration, and other factors that need to be considered when implementing `Task<LivingEntity>`.  The framework is not required; it's just an easier way to implement `Task<LivingEntity>`.
- Extensive documentation for each of these.

### Future Maintenance:

- If a new method is added to `VillagerTaskListProvider`, three things will need to be added:
1. An `@Inject` mixin for that method's "Constant" method (inject into the main list),
2. A `@ModifyArgs` mixin for that method's "Random" tasks (inject into the list passed to `RandomTask`), or add a new `RandomTask` to that list if none is included by default (copy from the other methods of that kind and change the enum), and,
3. A new enum value for that category.

- Potential signature changes are documented in `VillagerTaskListProviderInjector`.

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProvider.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProvider.java
@@ -1,0 +1,616 @@
+package net.fabricmc.fabric.api.object.builder.v1.ai.tasks;
+
+import com.google.common.collect.*;
+import com.mojang.datafixers.util.*;
+import net.minecraft.entity.ai.brain.task.*;
+import net.minecraft.entity.passive.*;
+import net.minecraft.village.*;
+//import org.jetbrains.annotations.*;
+
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.*;
+
+/**
+ * Provides task lists for villager professions. Called when initializing a {@code VillagerEntity}'s {@link net.minecraft.entity.ai.brain.Brain Brain}. Must be registered with {@link VillagerTaskProviderRegistry#register(VillagerProfession, VillagerTaskProvider)} before it can be called.
+ * @see VillagerTaskListProvider Vanilla Source
+ * @see Task Task&lt;VillagerEntity&gt;
+ */
+@SuppressWarnings({"unused"})
+public final class VillagerTaskProvider {
+	public VillagerTaskProvider()
+	{
+		constantTaskMap = new EnumMap<>(TaskType.class);
+		randomTaskMap = new EnumMap<>(TaskType.class);
+	}
+	
+	private final EnumMap<TaskType, List<BiFunction<VillagerProfession, Float, Pair<Integer, ? extends Task<? super VillagerEntity>>>>> constantTaskMap;
+	private final EnumMap<TaskType, List<BiFunction<VillagerProfession, Float, Pair<Task<? super VillagerEntity>, Integer>>>> randomTaskMap;
+	
+	
+	/**
+	 * Adds a <b>Constant</b> task under the specified {@link TaskType} for any {@link VillagerProfession VillagerProfession} using this {@code VillagerTaskProvider}.
+	 * <p><i>Constant</i> tasks are priorities given to a {@code LivingEntity} which are constantly assessed, whereas <i>Random</i> tasks are selected during the {@code Entity's} lifetime.
+	 * <p>Example Usage: <blockquote><pre>addConstantTask(TaskType.CORE, (profession, f) -&gt; Pair.of(1, new VillagerTask(profession, f)));</pre></blockquote>
+	 * @see #addRandomTask(TaskType, BiFunction) addRandomTask
+	 * @see #addBaseConstantTask(TaskType, BiFunction) addBaseConstantTask
+	 * @apiNote {@code PLAY} tasks may only be modified in {@code base} tasks. All calls to this method with {@code TaskType.PLAY} will be redirected to {@link #addBaseConstantTask(TaskType, BiFunction) addBaseConstantTask}.
+	 * @implSpec {@code BiFunction}s must return a {@link Pair} containing the <b>Task</b> and its <b>Weight</b> (as used in a {@link net.minecraft.util.collection.WeightedList WeightedList}).
+	 * In addition, {@code PLAY} tasks must expect the {@code VillagerProfession} argument to be {@code NULL} when calculated.
+	 * @param taskType The {@code TaskType} of the specified task.
+	 * @param dynamicTask {@code BiFunction}, taking the {@code VillagerProfession} of the {@link VillagerEntity} executing the task and the <b>walk speed</b> of the villager as arguments.
+	 */
+	public VillagerTaskProvider addConstantTask(TaskType taskType, BiFunction<VillagerProfession, Float, Pair<Integer, ? extends Task<? super VillagerEntity>>> dynamicTask)
+	{
+		if (taskType == TaskType.PLAY) {
+			addBaseConstantTask(taskType, dynamicTask);
+			
+			return this;
+		}
+		
+		List<BiFunction<VillagerProfession, Float, Pair<Integer, ? extends Task<? super VillagerEntity>>>> list = constantTaskMap.get(taskType);
+		if (list == null)
+			list = new ArrayList<>();
+		
+		list.add(dynamicTask);
+		
+		constantTaskMap.put(taskType, list);
+		return this;
+	}
+	
+	/**
+	 * Adds a <b>Constant</b> task under the specified {@link TaskType} for any {@link VillagerProfession} using this {@code VillagerTaskProvider}.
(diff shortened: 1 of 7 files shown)


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProviderRegistry.java
```diff
@@ -0,0 +1,84 @@
...
+public final class VillagerTaskProviderRegistry {
+	private static final Map<VillagerProfession, VillagerTaskProvider> VILLAGER_TASK_PROVIDER_BUILDER = new HashMap<>();
+	
+	/**
+	 * Adds a {@link VillagerTaskProvider} to the registry, which will be called by the associated {@link VillagerProfession}.
+	 * @param executingProfession The {@code VillagerProfession} that will execute this {@code VillagerTaskProvider}.
+	 * @param taskListProvider The {@code VillagerTaskProvider} to register.
+	 */
+	public static void register(final VillagerProfession executingProfession, final VillagerTaskProvider taskListProvider)
+	{
```
frqnny, line 25: The brackets shouldn’t be like this, I think it might even fail a checkstyle check.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/WorkerVillagerTask.java
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * A framework for creating simple world-affecting {@code VillagerTask}s, including pre-made functions for scanning for targets, etc.
+ * @apiNote Tasks may also extend {@code Task<VillagerEntity>}.
+ * @see FarmerVillagerTask FarmerVillagerTask
+ * @see net.minecraft.entity.ai.brain.task.BoneMealTask BoneMealTask
+ * @see Task
+ * @see VillagerTaskProviderRegistry
+ */
+public abstract class WorkerVillagerTask extends Task<VillagerEntity> {
```
frqnny, line 32: I don’t think this class is right for FAPI. It isn’t supposed to have helper but unneeded classes.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProvider.java
```diff
@@ -0,0 +1,616 @@
...
+	 * @see TaskType#PLAY
+	 * @see TaskType#REST
+	 * @see TaskType#MEET
+	 * @see TaskType#IDLE
+	 * @see TaskType#PANIC
+	 * @see TaskType#PRERAID
+	 * @see TaskType#RAID
+	 * @see TaskType#HIDE
+	 */
+	public enum TaskType {
```
frqnny, line 495: This is problematic as it wouldn’t support adding an Activity to a villager brain. This would defenetly break a mod I would port over. Couldn’t we just track this stuff by using Activity instead of a enum TaskType?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/VillagerTaskProviderInternals.java
```diff
@@ -0,0 +1,116 @@
...
+	
+	private static final Map<VillagerProfession, VillagerTaskProvider> villagerTaskProviderMap;
+	
+	private static final VillagerTaskProvider defaultTaskProvider = new VillagerTaskProvider();
+	
+	private static VillagerTaskProvider getTaskProvider(VillagerProfession profession)
+	{
+		assert(profession != null);
+		
+		if (profession == VillagerProfession.NONE)
```
frqnny, line 26: Also some checkstyle issues. Get the Checkstyle plugin and ensure everything is all right.

frqnny (changes requested):
Overall I think it’s a good idea but it does need some polishing. Thanks!

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/WorkerVillagerTask.java
ByThePowerOfScience: No worries; I can drop it without any hassle.  I like the idea of helper classes, but I understand the necessity of a lightweight API.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProviderRegistry.java
ByThePowerOfScience: I switched to that style for personal use, because I find it easier to read when the header is separate from the body.  I'll run my IDE's code cleanup with the default preset.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProvider.java
ByThePowerOfScience: ...I was not aware of that option.  Give me like 15 minutes.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProviderRegistry.java
frqnny: Check the github actions failed logs to know what needs fixing. It doesn't include everything that needs fixing, but it won't stop failing until it fixes everything. You can also attempt to check locally by trying to build the module and seeing what checkstyle fails are there so you don't spam the commits on the pr.

ByThePowerOfScience: It was a licensing error.  Can I copy the license header from another class?

Technici4n: Just run `gradlew licenseFormatAll`, and you can use `gradlew check` to make sure the build test passes.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/ai/tasks/VillagerTaskProvider.java
ByThePowerOfScience: Huh, funny how my first implementation strategy - intercepting calls to `VillagerTaskListProvider` - ended up being the right option for implementing `Activity`.  What's even funnier is that I can't reuse my old code because too much has changed. :P
Well, at least it's just replacing Mixins this time and not scrapping 95% of my code to use an `EnumMap` instead of individual handler methods, lol.

Technici4n:
Are you planning to do more work on this?

ByThePowerOfScience:
> Are you planning to do more work on this?

Thank you so much for reminding me of this!  I totally forgot about this project.  I'll finish it up this weekend.

EDIT: And by "this weekend" I of course meant "over a month later", what do you mean I procrastinated.

ByThePowerOfScience:
Now that I've started generalizing this API hook and basing it on Activity rather than an external enum, I realized that there's nothing tying it to just Villagers anymore, so I'm in the process of retooling this into a global cross-mod "Brain Task-List Injector".

I still need to work out how to implement the universal Entity compatibility, but I've got most of it down so it shouldn't take too long.  

Closing this PR for now, since the concept has changed from the original.

PR #1486, opened 2021-06-11
Make BlockApiLookup expose the API and context classes

### LINKED ISSUE #1464: API Lookup should expose the API and context type
It's useful for anyone using API API beyond the simple use case.

For example, while trying to dispatch API queries to AE2 parts, I noticed that I need to get the context class to check if it's `AEPartLocation` or `Direction`: https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/5711bbc8ffaeb25ada8149cb5f674d3e699eb3d6/src/api/java/appeng/api/parts/PartApiLookup.java#L38. (warning: most of the code is MCP, but not this file).

In this case I could also ask for the context as an additional parameter, but it's redundant and I think it's a good API API addition. I am opening this to remember. I will update #1352 as well soon:tm:.

Technici4n:
Closes #1464. I had a few minutes to write this, not high priority but at the same time very simple.

### DIFF
--- a/fabric-api-lookup-api-v1/build.gradle
+++ b/fabric-api-lookup-api-v1/build.gradle
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-api-lookup-api-v1"
-version = getSubprojectVersion(project, "1.0.0")
+version = getSubprojectVersion(project, "1.1.0")
 
 moduleDependencies(project, [
 	'fabric-api-base',
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
@@ -217,6 +217,16 @@ public interface BlockApiLookup<A, C> {
 	 */
 	void registerFallback(BlockApiProvider<A, C> fallbackProvider);
 
+	/**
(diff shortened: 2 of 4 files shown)


liach (approved):

modmuss50 (approved):

Player3324 (approved):

Player3324:
Considering it's a mixed class the get prefixes would be required, but it may cause confusion with the regular get()

liach:
Last call this? The compatibility impact should be minimal.

Player3324:
It's too minor for the last call process imo

Technici4n:
Hmmm yes, I think omitting the `get` prefix is better to avoid confusion.

Technici4n:
No need for last call, we just need to agree on the get prefix. Imo both options are equally suitable.

liach:
Related: FabricMC/yarn#2408

Another remark: this api and context classes are immutable.
Also, how about kotlin usages?

Technici4n:
Hmmm let me add the get prefixes then, it will be better for consistency.

Technici4n:
image

Actually no, this is a better separation. Arghhh I don't know 😆

Technici4n:
image

This is with the `get` prefixes... I'm in favor of not adding them, and we can merge this PR then. 👍

liach:
Yeah, I just added that link for referential purposes. Jdk does not use get prefix for such immutable properties mostly, and I agree with your decision of no extra `get`.

Issue #1476, opened 2021-06-05
Non-flammable wood is not respected for tags added to the fuel registry


haykam821:
Unlike vanilla fuels that are backed by tags (such as `#minecraft:wooden_slabs` or `#minecraft:wooden_stairs`), any tag added to the fuel registry using Fabric API's helper is not checked for non-flammable wood, meaning it is much more painful to follow vanilla conventions while adding fuel.

PR #1491, opened 2021-06-12
Create a test mod for the fabric-content-registries-v0 module

haykam821:


Technici4n (approved):
LGTM, does it test every feature of the module?

Juuxel:
I think testing tag-based fuel values is the only thing that's missing from here.

Technici4n:
Might be worth adding that, then?

haykam821:
It does not test `LootEntryTypeRegistry` which is deprecated in favor of a replacement from another module as well. Testing tags is a good idea, though, and would be beneficial for #1476.

Juuxel (approved):

liach (approved):

liach:
On a side note, I hope we have a place where we look up all the desired behaviors as a result of enabling these test mods than having to look into individual code. Another approach may be to have some unit tests in events, but those would be a bit hard to regulate as well.

fabric-content-registries-v0/src/testmod/java/net/fabricmc/fabric/test/registry/ContentRegistryTest.java
```diff
@@ -0,0 +1,40 @@
...
+public final class ContentRegistryTest implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		CompostingChanceRegistry.INSTANCE.add(Items.OBSIDIAN, 0.5F);
+
+		FlammableBlockRegistry.getDefaultInstance().add(Blocks.OBSIDIAN, 2, 3);
+		FlammableBlockRegistry.getDefaultInstance().add(BlockTags.SNOW, 4, 6);
+
+		FuelRegistry.INSTANCE.add(Items.OBSIDIAN, 20 * 3);
+		FuelRegistry.INSTANCE.add(ItemTags.FOX_FOOD, 20 * 6);
```
modmuss50, line 38: Can we add a check on first tick or something to ensure these have been registered, and crash hard if they havent.

modmuss50 (changes requested):
Would perfer some validation that this has actually worked, crashing the game hard if something isnt quite right would be best.

Technici4n:
@haykam821 Any update?

Technici4n:
up?

ghost:
Doesn't this already exist? In fact there are conflicts...

ghost:
Note that Composting and Fuel tests are missing from testmods from all the releases (from 1.17 to 1.19).
All other tests are present.

Technici4n:
Feel free to open a new PR with proper automated testing then! :)

PR #1492, opened 2021-06-12
Fix wrong RegistryOps call

frqnny:
Looks like it's the last one to the old methods. Should fix the datapack issues.


frqnny:
Looks like modmuss already commited this himself.

modmuss50:
Ah sorry, same fix though. 👍

PR #1495, opened 2021-06-13
Fix structure config not applying in every case

### LINKED ISSUE #1494: StructureConfigs need applying
https://github.com/slicedlime/examples/raw/master/vanilla_worldgen.zip 

For some reason, we do not add StructureConfigs to data pack chunk generator. 
Using Mo Structures / any structure mod and the data pack’s chunk generator make my structures not appear. That is because we need to add to them. I can do this later tonight unless someone wants to do it before.

frqnny:
Fixes #1494

The problem was that data packs that add a custom chunk generator will not get structure configs and thus not apply structures to these biomes. This becomes a problem with data packs like the caves data pack. 

The problem was that Structure API added Structure Configs by first using *two* mixins to capture StructuresConfig (the class that holds a map of StructureConfig s to Structures). Then, during the FabricStructureBuilder's register method, it would iterate through a list of StructureConfigs captured by the two mixins and add the structure config. 

Now, we just have a Map of structure features to configs that we add to when we call register, and when it is needed, the map is added to the StructuresConfig at use. This simplifies that part of the code by a lot. It also now depends on lifecycles.

This will need backporting to 1.16.


frqnny:
To reproduce, you can download Mo' Structures and the worldgen datapack.
https://launcher.mojang.com/v1/objects/233aa01c6db99ffcf95594e384e72671ae279ced/CavesAndCliffsPreview.zip
Test with a build of Fabric API and then with this- the testmod when I tested worked and showed the caves and cliffs datapack along with the testmod structure generating.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructureImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+	@Override
+	public void onInitialize() {
+		ServerWorldEvents.LOAD.register((server, world) -> {
+			// Need temp map as some mods use custom chunk generators with immutable maps in themselves.
+			Map<StructureFeature<?>, StructureConfig> tempMap = new HashMap<>(world.getChunkManager().getChunkGenerator().getStructuresConfig().getStructures());
+
+			tempMap.putAll(STRUCTURE_TO_CONFIG_MAP);
+
+			((StructuresConfigAccessor) world.getChunkManager().getChunkGenerator().getStructuresConfig()).setStructures(tempMap);
```
modmuss50: Make the map immutable again before replacing it

liach:
this change looks fantastic! should review on desktop again soon.

liach (approved):

fabric-structure-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -15,6 +15,11 @@
   "authors": [
     "FabricMC"
   ],
+  "entrypoints": {
+    "main": [
+      "net.fabricmc.fabric.impl.structure.FabricStructureImpl"
+    ]
+  },
   "depends": {
```
Technici4n, line 23: Don't forget to add lifecycle events here.

Technici4n (approved):
Tested and it works, injection point seems alright. Don't forget to update the `fabric.mod.json`.

Also... what is this in `FabricStructureBuilder#register`? That looks extremely suspicious.
```java
		// update builtin biomes, just to be safe
		for (Biome biome : BuiltinRegistries.BIOME) {
			BiomeAccessor biomeAccessor = (BiomeAccessor) (Object) biome;
			Map<Integer, List<StructureFeature<?>>> structureLists = biomeAccessor.getStructureLists();

			if (!(structureLists instanceof HashMap)) {
				// not guaranteed by the standard to be a mutable map
				((BiomeAccessor) (Object) biome).setStructureLists(structureLists = new HashMap<>(structureLists));
			}

			// not guaranteed by the standard to be mutable lists
			structureLists.compute(step.ordinal(), (k, v) -> makeMutable(v)).add(structure);
		}
```

Once these two concerns are addressed, it can be merged as far as I am concerned.

fabric-structure-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -15,6 +15,11 @@
...
     "fabricloader": ">=0.8.0",
     "fabric-api-base": "*"
```
liach:
```suggestion
    "fabric-api-base": "*",
    "fabric-lifecycle-events-v1": ">=1.4.4"
```

Technici4n:
My two comments have been addressed, thanks!

liach (approved):

liach:
Actually whether this is a "small change" is somewhat debatable, since this introduces new dependencies and apis to mods relying exclusively on the structures api (i.e. the lifecycle api). But this is an addition (and is largely compatible) and most mods using structures api would use the lifecycle api, so it would be more or less fine.

LambdAurora (approved):

modmuss50 (approved):
Looks good, much cleaner 👍 

Thanks @Technici4n for testing.

Player3324 (approved):
LGTM, thanks!

PR #1498, opened 2021-06-14
Fix ClientPickBlockApplyCallback.EVENT not working as expected Fixes #1493

### LINKED ISSUE #1493: ClientPickBlockApplyCallback.EVENT does not affect the pick block outcome.
### Steps to reproduce:

- Clone the 1.17 Example project
- Paste the following code in onInitialize
```
        ClientPickBlockApplyCallback.EVENT.register((player,result,stack)->{
            return player.getInventory().getStack(9);
        });
```
- Put an Item in to the top left slot of the inventory
- Middle-Mouse-Click something 
### Expected Result:
The item in the top left slot gets moved into your selected hotbar slot
### Actual Result:
Nothing Happens

FriederHannenheim:
Please tell me if I should change anything in the code


FriederHannenheim:
#1493

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientPickBlockCallback.java
```diff
@@ -1,65 +0,0 @@
...
-
-import net.minecraft.entity.player.PlayerEntity;
-import net.minecraft.item.ItemStack;
-import net.minecraft.util.hit.HitResult;
-
-import net.fabricmc.fabric.api.event.Event;
-import net.fabricmc.fabric.api.event.EventFactory;
-
-/**
- * This event handler has been deprecated due to not hooking nicely
```
modmuss50: Unless this was causing a problem when fixing I’d suggest leaving this alone.

modmuss50 (changes requested):

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/client/player/ClientPickBlockCallback.java
FriederHannenheim: Ok I'll leave it in.

modmuss50:
I dont believe this is the correct fix as it does not handle the empty case correctally and doesnt set the correct stack to the other calls. I came up with the following fix based on the old code.

```java
	@ModifyVariable(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;getInventory()Lnet/minecraft/entity/player/PlayerInventory;"), method = "doItemPick", ordinal = 0)
	public ItemStack modifyItemPick(ItemStack stack) {
		MinecraftClient client = (MinecraftClient) (Object) this;
		ItemStack result = ClientPickBlockApplyCallback.EVENT.invoker().pick(client.player, client.crosshairTarget, stack);
		fabric_itemPickCancelled = result.isEmpty();
		return result;
	}

	@Inject(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;getInventory()Lnet/minecraft/entity/player/PlayerInventory;"), method = "doItemPick", cancellable = true)
	public void cancelItemPick(CallbackInfo info) {
		if (fabric_itemPickCancelled) {
			info.cancel();
		}
	}
```
This seems like the propper solution to me, as this still handles the empty item case much better and allows for blocking the call to addBlockEntityNbt.

I would also tweak the test mod to not override all block picking and just in certain cases such as:

```java
		ClientPickBlockApplyCallback.EVENT.register((player, result, stack) -> {
			if (player.getStackInHand(Hand.MAIN_HAND).getItem() == Items.DIAMOND) {
				return new ItemStack(Items.OBSIDIAN);
			}

			if (stack.getItem() == Items.GRASS_BLOCK) {
				return ItemStack.EMPTY;
			}

			return stack;
		});
```

EDIT: I think this might still be wrong as I doesnt really change anything. I will look more when its not late.

Player3324:
Assuming the problem is the `getSlotWithStack` arg (`itemStack3`) still being on the stack for the call, the original fabric-api code should be fixed by adding a shift by -1 (= just `Shift.BEFORE`) in the mixin `@At` specification.

This turns something like vanilla's (assumed)
`aload 5, aload 4, invokevirtual getSlotWithStack` 
into something like 
`aload 5, aload 0, invokevirtual mixin, astore 4, aload 4, invokevirtual getSlotWithStack`
instead of the previous
`aload 5, aload 4, aload 0, invokevirtual mixin, astore 4, invokevirtual getSlotWithStack`
where the arg for getSlotWithStack remained the old stack, not observing the local variable change.

FriederHannenheim:
> I dont believe this is the correct fix as it does not handle the empty case correctally and doesnt set the correct stack to the other calls. I came up with the following fix based on the old code.
> 
> ```java
> 	@ModifyVariable(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;getInventory()Lnet/minecraft/entity/player/PlayerInventory;"), method = "doItemPick", ordinal = 0)
> 	public ItemStack modifyItemPick(ItemStack stack) {
> 		MinecraftClient client = (MinecraftClient) (Object) this;
> 		ItemStack result = ClientPickBlockApplyCallback.EVENT.invoker().pick(client.player, client.crosshairTarget, stack);
> 		fabric_itemPickCancelled = result.isEmpty();
> 		return result;
> 	}
> 
> 	@Inject(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;getInventory()Lnet/minecraft/entity/player/PlayerInventory;"), method = "doItemPick", cancellable = true)
> 	public void cancelItemPick(CallbackInfo info) {
> 		if (fabric_itemPickCancelled) {
> 			info.cancel();
> 		}
> 	}
> ```
> 
> This seems like the propper solution to me, as this still handles the empty item case much better and allows for blocking the call to addBlockEntityNbt 

I'm pretty sure this is not necessary since the item stack is always checked to be non-empty. So there's no need to cancel the method since no further action will happen

Player3324:
The creative mode path appears to do something even if the stack is empty / the index is -1.

Player3324:
From reading the code one of  those cases might be when the hotbar doesn't have an empty slot or even if the current slot isn't empty.

FriederHannenheim:
Alright. I'll try the old method with Shift.BEFORE.

FriederHannenheim:
Alright, it works. Should I make a new pull request so there's not that many useless commits?

Player3324:
Using this one is fine, we're going to squash it anyway.

FriederHannenheim:
There. Now everything is working

Player3324 (approved):
LGTM, thanks!

modmuss50 (approved):

liach (approved):

PR #1507, opened 2021-06-20
Add support for creation of tree types (TrunkPlacerType etc.) in Object Builders API v1

kanpov:
This PR adds functionality for creating and registering registry wrappers primarily used by trees like `TrunkPlacerType`s.

This is **not** a new module, it's an addition to the already existing Object Builders API.

This is my first time contributing to the Fabric project (except for my Wiki contributions) and any feedback, corrections and criticism would be really appreciated!

Thanks for your attention! 

**UPDATE**: this PR will no longer receive updates. I do not understand why the review process must take so long. thanks everyone who has helped me with this. I'm going to make this into a third-party library because this is ridiculous.

**UPDATE**: The library that does this is now available at https://github.com/RedGrapefruit09/Arctree


fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+	/**
+	 * A convenience method to avoid conflicts when using {@link IntProviderType#register}.
+	 *
+	 * @param id Registry {@link Identifier}
+	 * @param codec The {@link Codec} associated with the {@link IntProvider}
+	 * @param <T> The owner {@link IntProvider}
+	 * @return Created and registered {@link IntProviderType}
+	 */
+	public static <T extends IntProvider> IntProviderType<T> registerIntProviderType(Identifier id, Codec<T> codec) {
+		return IntProviderType.register(id.getNamespace() + "__" + id.getPath(), codec);
```
liach:
```suggestion
		return IntProviderType.register(id.toString(), codec);
```

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+	 * Creates a new instance of a {@link BlockStateProviderType}, registers and returns it.<br>
+	 * Uses {@link BlockStateProviderTypeInvoker} under the hood.
+	 *
+	 * @param id Registry {@link Identifier}
+	 * @param codec The {@link Codec} associated with the {@link BlockStateProvider}
+	 * @param <T> The owner {@link BlockStateProvider}
+	 * @return Created and registered {@link BlockStateProviderType}
+	 */
+	public static <T extends BlockStateProvider> BlockStateProviderType<T> registerBlockStateProviderType(Identifier id, Codec<T> codec) {
+		return BlockStateProviderTypeInvoker.invokeRegister(id.getNamespace() + "__" + id.getPath(), codec);
```
liach:
```suggestion
		return BlockStateProviderTypeInvoker.invokeRegister(id.toString(), codec);
```

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+	 * Creates a new instance of a {@link TreeDecoratorType}, registers and returns it.<br>
+	 * Uses {@link TreeDecoratorTypeInvoker} under the hood.
+	 *
+	 * @param id Registry {@link Identifier}
+	 * @param codec The {@link Codec} associated with the {@link TreeDecorator}
+	 * @param <T> The owner {@link TreeDecorator}
+	 * @return Created and registered {@link TreeDecoratorType}
+	 */
+	public static <T extends TreeDecorator> TreeDecoratorType<T> registerTreeDecoratorType(Identifier id, Codec<T> codec) {
+		return TreeDecoratorTypeInvoker.invokeRegister(id.getNamespace() + "__" + id.getPath(), codec);
```
liach:
```suggestion
		return TreeDecoratorTypeInvoker.invokeRegister(id.toString(), codec);
```

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+	 * Creates a new instance of a {@link FoliagePlacerType}, registers and returns it.<br>
+	 * Uses {@link FoliagePlacerTypeInvoker} under the hood.
+	 *
+	 * @param id Registry {@link Identifier}
+	 * @param codec The {@link Codec} associated with the {@link FoliagePlacer}
+	 * @param <T> The owner {@link FoliagePlacer}
+	 * @return Created and registered {@link FoliagePlacerType}
+	 */
+	public static <T extends FoliagePlacer> FoliagePlacerType<T> registerFoliagePlacerType(Identifier id, Codec<T> codec) {
+		return FoliagePlacerTypeInvoker.invokeRegister(id.getNamespace() + "__" + id.getPath(), codec);
```
liach:
```suggestion
		return FoliagePlacerTypeInvoker.invokeRegister(id.toString(), codec);
```

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+	 * Creates a new instance of a {@link TrunkPlacerType}, registers and returns it.<br>
+	 * Uses {@link TrunkPlacerTypeInvoker} under the hood.
+	 *
+	 * @param id Registry {@link Identifier}
+	 * @param codec The {@link Codec} associated with the {@link TrunkPlacer}
+	 * @param <T> The owner {@link TrunkPlacer}
+	 * @return Created and registered {@link TrunkPlacerType}
+	 */
+	public static <T extends TrunkPlacer> TrunkPlacerType<T> registerTrunkPlacerType(Identifier id, Codec<T> codec) {
+		return TrunkPlacerTypeInvoker.invokeRegister(id.getNamespace() + "__" + id.getPath(), codec);
```
liach:
```suggestion
		return TrunkPlacerTypeInvoker.invokeRegister(id.toString(), codec);
```

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingBlock.java
```diff
@@ -0,0 +1,23 @@
...
+ * <p>
+ *     An example of using this implementation:<br>
+ *
+ *     <code>
+ *			public static final SimpleSaplingBlock MY_SAPLING = new SimpleSaplingBlock(new SimpleSaplingGenerator(MY_TREE_CONFIGURED), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));
+ *     </code>
+ * </p>
+ */
+public class SimpleSaplingBlock extends SaplingBlock {
+	public SimpleSaplingBlock(SaplingGenerator generator, Settings settings) {
```
liach: Instead of creating this class, adding an access widener entry to make sapling block's constructor public is better. Then mods depending on the tree api (ideally the object builder api) will just use the access widener from the api and can call the constructor directly.

In addition, users can always create a `new SaplingBlock(generator, settings) {}` to call the protected constructor.

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+ * <p>Most of the time when working with {@link SaplingGenerator}s, you'll only need this functionality.</p>
+ * <p>
+ *     An example of using this implementation:<br>
+ *
+ *     <code>
+ *         public static final MY_SAPLING_GENERATOR = new SimpleSaplingGenerator(MY_TREE_CONFIGURED);
+ *     </code>
+ * </p>
+ */
+public class SimpleSaplingGenerator extends SaplingGenerator {
```
liach: The usefulness of this imo is limited; there is another `LargeTreeSaplingGenerator`, which this cannot cover apparently.

liach (commented):
Good additions; I believe this can just be part of the object builder api, as this does not relate to biome internals (and it doesn't need to depend on the biome api at all)

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
liach: Fyi such code will be piped to `Registry.register`, which converts the `toString()` result back as an identifier. This is a stupid process, I know.

LambdAurora:
> Good additions; I believe this can just be part of the object builder api, as this does not relate to biome internals (and it doesn't need to depend on the biome api at all)

Can we stop putting random stuff in object builder API?
Like come on, there's BlockSettings, Criterion stuff, BlockEntityType builder, Material builder, ModelPredicateProvider, TradeOffer stuff, more entity stuff, villager stuff, and POI?

Player3324:
We have been entertaining the idea of adding global access wideners to solve all cases where the vanilla API is too restrictive from them registering within the same class or package.

The object builder module is kind of the catch all for access fixes, IMO better than a dedicated module for trees..

liach:
> Can we stop putting random stuff in object builder API?

Please be less dreary, thanks! These stuff The object builder does add API to call otherwise-unavailable register methods for different types of objects, such as living entity attributes. All of these trunk placer, block state provider, etc. types have private ctor + register methods, and exposing the register methods as in this PR (but moved into object builder API) Imo is the best. See also #1091, which adds a similar hook to the object builders.

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+ *         public static final BlockStateProviderType<MyBlockStateProvider> MY_BLOCK_STATE_PROVIDER = FabricTreeRegistry.registerBlockStateProviderType(new Identifier("tutorial", "my_block_state_provider"), MyBlockStateProvider.CODEC);
+ *     </code>
+ * </p>
+ * <p>
+ *     Example for {@link IntProviderType}s:<br>
+ *
+ *     <code>
+ *         public static final IntProviderType<MyIntProvider> MY_INT_PROVIDER = FabricTreeRegistry.registerIntProviderType(new Identifier("tutorial", "my_int_provider"), MyIntProvider.CODEC);
+ *     </code>
+ * </p>
```
haykam821: I think it's better to put these examples under their individual methods.

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
```diff
@@ -0,0 +1,125 @@
...
+import net.minecraft.world.gen.stateprovider.BlockStateProviderType;
+import net.minecraft.world.gen.treedecorator.TreeDecorator;
+import net.minecraft.world.gen.treedecorator.TreeDecoratorType;
+import net.minecraft.world.gen.trunk.TrunkPlacer;
+import net.minecraft.world.gen.trunk.TrunkPlacerType;
+
+/**
+ * <p>
+ * 	   A registry for {@link TrunkPlacerType}s, {@link FoliagePlacerType}s, {@link TreeDecoratorType}s,
+ * 	   {@link BlockStateProviderType}s and {@link IntProviderType}s.
```
haykam821: Use an Oxford comma here:

```suggestion
 * 	   {@link BlockStateProviderType}s, and {@link IntProviderType}s.
```

LambdAurora:
> The object builder module is kind of the catch all for access fixes, IMO better than a dedicated module for trees..

The object builder module mostly looks like a trash pile due to bad organization.

Like, clearly it's the most random module of the whole API, it's extremely non-intuitive to modders, and bloat very small mods for no reason.

> Please be less dreary, thanks! These stuff The object builder does add API to call otherwise-unavailable register methods for different types of objects, such as living entity attributes. All of these trunk placer, block state provider, etc. types have private ctor + register methods, and exposing the register methods as in this PR (but moved into object builder API) Imo is the best. See also #1091, which adds a similar hook to the object builders.

Ok, but some of those things could be organized in a way better and intuitive way. How can I guess I need object builder API for `FabricBlockSettings` (which I just learned this instant, I never realized it wasn't in a block-related module).

Like, it's just becoming ridiculous.

Player3324:
An `access-fixes` module would probably be best, global aw or not.

frqnny:
If #1484 gets merged, we could create a fabric-villager-api to add stuff from villagers into there.

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
kanpov: Yea I was a bit confused there

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
kanpov: Okay, I'll do that. Thanks!

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/FabricTreeRegistry.java
kanpov: I'm a Russian learning English (B2), so don't judge me :)

kanpov:
Thanks @liach and @haykam821 for your suggestions!

I'll try to move this into object builders. I disagree with @LambdAurora about the object builders API, although it could have a better name, like what @sfPlayer1 suggested.

kanpov:
I have moved this functionality into Object Builders and renamed this pull request. I've also removed the `SimpleSaplingBlock` and `SimpleSaplingGenerator` temporarily.

I'm going to add an access widener to `SaplingBlock` instead of using a separate class.

I'm not sure about the `SimpleSaplingGenerator`, I don't think it should be removed, but maybe renamed to something like `DefaultSaplingGenerator`? What do you think?

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingBlock.java
Juuxel: Access wideners aren't exposed, though. Only your own AW will be applied at compile time (and deps' at runtime).

kanpov:
I've renamed `FabricTreeRegistry` to `TreeTypeHelper`, added an access widener for `SaplingBlock` and bumped Object Builders API's version to `1.11.0` from `1.10.9`

Juuxel:
The access widener isn't all that useful since Fabric API can't expose its AW to dependents (at least yet), so `SaplingBlock.<init>` will only be widened for fabric-object-builder-api-v1.

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingBlock.java
kanpov: Yeah, IDEs don't support AWs very well, which is its biggest disadvantage.

kanpov:
> The access widener isn't all that useful since Fabric API can't expose its AW to dependents (at least yet), so `SaplingBlock.<init>` will only be widened for fabric-object-builder-api-v1.

I've added back the `SimpleSaplingBlock` as a temporary solution for the time.

No exposed access wideners in APIs is a **huge issue** in my opinion.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/FabricTreeRegistry.java
```diff
@@ -76,6 +47,14 @@
 	 * Creates a new instance of a {@link FoliagePlacerType}, registers and returns it.<br>
 	 * Uses {@link FoliagePlacerTypeInvoker} under the hood.
 	 *
+	 * <p>
+	 *     Example:<br>
+	 *
+	 *     <code>
+	 *         public static final FoliagePlacerType<MyFoliagePlacer> MY_FOLIAGE_PLACER = FabricTreeRegistry.registerFoliagePlacerType(new Identifier("tutorial", "my_foliage_placer"), MyFoliagePlacer.CODEC);
```
liach: Hmm, I don't think the generic diamonds work with generated html. Replace `<code></code>` with `{@code}` instead to avoid this issue. Same for other usages.

liach:
@sfPlayer1 Mind enable workflows for this pr, so we can run checkstyle on github?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/FabricTreeRegistry.java
kanpov: Okay, thanks! I'll do that tomorrow since it's 10 PM in my timezone and I kinda have to sleep.

liach:
In addition, redgrapefruit, please run `./gradlew licenseFormat` in your fabric repo to add license headers for the new classes; the violations are failing the build.

kanpov:
> 
> 
> In addition, redgrapefruit, please run `./gradlew licenseFormat` in your fabric repo to add license headers for the new classes; the violations are failing the build.

Done! The checks can be re-run now, @sfPlayer1.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/FabricTreeRegistry.java
kanpov: Done!

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingBlock.java
kanpov: I've readded this to the PR because AWs aren't exposed. It'll stay as a temporary solution

fabric-tree-api-v1/src/main/java/net/fabricmc/fabric/api/tree/v1/SimpleSaplingGenerator.java
kanpov: I've removed this for now. Maybe will restore later.

kanpov:
I've merged the fork upstream to avoid any possible conflicts in the future. If you have any suggestions for this PR, please share them here.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
```diff
@@ -0,0 +1,39 @@
...
+ * <p>
+ *    Example:<br>
+ *
+ *    {@code
+ *    	public static final SaplingBlock MY_SAPLING = new SaplingBlock(new MySaplingGenerator(), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));
+ *    }
+ * </p>
+ */
+public class SimpleSaplingBlock extends SaplingBlock {
+	protected SimpleSaplingBlock(SaplingGenerator generator, Settings settings) {
```
liach:
```suggestion
	public SimpleSaplingBlock(SaplingGenerator generator, Settings settings) {
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * <p>
+ *    A temporary solution because access wideners cannot be applied to dependents (yet).<br>
+ *    Allows you to create your own {@link SaplingBlock}s.
+ * </p>
+ * <p>
+ *    Example:<br>
+ *
+ *    {@code
+ *    	public static final SaplingBlock MY_SAPLING = new SaplingBlock(new MySaplingGenerator(), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));
```
liach:
```suggestion
 *    	public static final SaplingBlock MY_SAPLING = new SimpleSaplingBlock(new MySaplingGenerator(), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));
```

fabric-object-builder-api-v1/src/main/resources/fabric-object-builder-api-v1.accesswidener
```diff
@@ -1,5 +1,6 @@
 accessWidener	v1	named
 extendable  method    net/minecraft/block/AbstractBlock$Settings    <init>  (Lnet/minecraft/block/Material;Ljava/util/function/Function;)V
 extendable  method    net/minecraft/block/AbstractBlock$Settings    <init>  (Lnet/minecraft/block/Material;Lnet/minecraft/block/MapColor;)V
+extendable  method    net/minecraft/block/SaplingBlock              <init>  (Lnet/minecraft/block/sapling/SaplingGenerator;Lnet/minecraft/block/AbstractBlock$Settings;)V
```
liach: Since this isn't exposed by the api, we should probably revert it.
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
kanpov: My fault. Thanks! (I am so dumb)

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
kanpov: Bruh

fabric-object-builder-api-v1/src/main/resources/fabric-object-builder-api-v1.accesswidener
kanpov: Yeah, `SimpleSaplingBlock` is a temporary solution for the time being.

kanpov:
Thanks @liach for your feedback! I've committed all your suggestions.

The approval for running workflows is annoying tbh. reviews? maybe?

liach (approved):

liach:
@Technici4n @modmuss50 @sfPlayer1  Mind review this simple hook addition?

kanpov:
> @Technici4n @modmuss50 @sfPlayer1 Mind review this simple hook addition?

^

Technici4n:
Looks good, javadoc format is wrong though. Please don't use `<p>` tags for the first paragraph, don't use `</p>` tags at all (they are implicit), and don't indent javadoc. You can check javadoc with `gradlew check`, however that may not catch all of these.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
```diff
@@ -0,0 +1,39 @@
...
+ * <p>
+ *    A temporary solution because access wideners cannot be applied to dependents (yet).<br>
+ *    Allows you to create your own {@link SaplingBlock}s.
+ * </p>
+ * <p>
+ *    Example:<br>
+ *
+ *    {@code
+ *    	public static final SaplingBlock MY_SAPLING = new SimpleSaplingBlock(new MySaplingGenerator(), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));
+ *    }
```
liach:
```suggestion
 * Allows you to create your own {@link SaplingBlock}s.
 *
 * <p>Example:<br>
 * {@code public static final SaplingBlock MY_SAPLING = new SimpleSaplingBlock(new MySaplingGenerator(), FabricBlockSettings.copyOf(Blocks.OAK_SAPLING));}
 *
 * @apiNote This is a temporary solution because access wideners cannot be applied to dependents yet.
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/SimpleSaplingBlock.java
```diff
@@ -0,0 +1,39 @@
...
+ * </p>
```
liach:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+import net.minecraft.world.gen.trunk.TrunkPlacerType;
+
+/**
+ * A registry for {@link TrunkPlacerType}s, {@link FoliagePlacerType}s, {@link TreeDecoratorType}s,
+ * {@link BlockStateProviderType}s, and {@link IntProviderType}s.
+ */
+public final class TreeTypeHelper {
+	/**
+	 * Creates a new instance of a {@link TrunkPlacerType}, registers and returns it.<br>
+	 * Uses {@link TrunkPlacerTypeInvoker} under the hood.
```
liach:
```suggestion
	 * Creates a new instance of a {@link TrunkPlacerType}, registers and returns it.
```
No need to mandate implementation details.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * <p>
+	 *    Example:<br>
+	 *
+	 *    {@code
+	 *          public static final TrunkPlacerType<MyTrunkPlacer> MY_TRUNK_PLACER =	FabricTreeRegistry.registerTrunkPlacerType(new Identifier("example", "my_trunk_placer"), MyTrunkPlacer.CODEC);
+	 *    }
+	 * </p>
```
liach:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 * @param <T> The owner {@link TrunkPlacer}
+	 * @return Created and registered {@link TrunkPlacerType}
+	 */
+	public static <T extends TrunkPlacer> TrunkPlacerType<T> registerTrunkPlacerType(Identifier id, Codec<T> codec) {
+		return TrunkPlacerTypeInvoker.invokeRegister(id.toString(), codec);
+	}
+
+	/**
+	 * Creates a new instance of a {@link FoliagePlacerType}, registers and returns it.<br>
+	 * Uses {@link FoliagePlacerTypeInvoker} under the hood.
```
liach:
```suggestion
	 * Creates a new instance of a {@link FoliagePlacerType}, registers and returns it.
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * <p>
+	 *     Example:<br>
+	 *
+	 *    {@code
+	 *         public static final FoliagePlacerType<MyFoliagePlacer> MY_FOLIAGE_PLACER = FabricTreeRegistry.registerFoliagePlacerType(new Identifier("tutorial", "my_foliage_placer"), MyFoliagePlacer.CODEC);
+	 *     }
+	 * </p>
```
liach, line 69:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * A convenience method to avoid conflicts when using {@link IntProviderType#register}.
+	 *
+	 * <p>
+	 *     Example for {@link IntProviderType}s:<br>
+	 *
+	 *     {@code
+	 *         public static final IntProviderType<MyIntProvider> MY_INT_PROVIDER = FabricTreeRegistry.registerIntProviderType(new Identifier("tutorial", "my_int_provider"), MyIntProvider.CODEC);
+	 *     }
+	 * </p>
```
liach:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 * @param <T> The owner {@link TreeDecorator}
+	 * @return Created and registered {@link TreeDecoratorType}
+	 */
+	public static <T extends TreeDecorator> TreeDecoratorType<T> registerTreeDecoratorType(Identifier id, Codec<T> codec) {
+		return TreeDecoratorTypeInvoker.invokeRegister(id.toString(), codec);
+	}
+
+	/**
+	 * Creates a new instance of a {@link BlockStateProviderType}, registers and returns it.<br>
+	 * Uses {@link BlockStateProviderTypeInvoker} under the hood.
```
liach:
```suggestion
	 * Creates a new instance of a {@link BlockStateProviderType}, registers and returns it.
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * <p>
+	 *     Example for {@link BlockStateProviderType}s:<br>
+	 *
+	 *     {@code
+	 *         public static final BlockStateProviderType<MyBlockStateProvider> MY_BLOCK_STATE_PROVIDER = FabricTreeRegistry.registerBlockStateProviderType(new Identifier("tutorial", "my_block_state_provider"), MyBlockStateProvider.CODEC);
+	 *     }
+	 * </p>
```
liach:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 * Creates a new instance of a {@link TreeDecoratorType}, registers and returns it.<br>
+	 * Uses {@link TreeDecoratorTypeInvoker} under the hood.
+	 *
+	 * <p>
+	 *     Example for {@link TreeDecoratorType}s:<br>
+	 *
+	 *     {@code
+	 *         public static final TreeDecoratorType<MyTreeDecorator> MY_TREE_DECORATOR = FabricTreeRegistry.registerTreeDecoratorType(new Identifier("tutorial", "my_tree_decorator"), MyTreeDecorator.CODEC);
+	 *     }
+	 * </p>
```
liach:
```suggestion
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+	 * @param <T> The owner {@link FoliagePlacer}
+	 * @return Created and registered {@link FoliagePlacerType}
+	 */
+	public static <T extends FoliagePlacer> FoliagePlacerType<T> registerFoliagePlacerType(Identifier id, Codec<T> codec) {
+		return FoliagePlacerTypeInvoker.invokeRegister(id.toString(), codec);
+	}
+
+	/**
+	 * Creates a new instance of a {@link TreeDecoratorType}, registers and returns it.<br>
+	 * Uses {@link TreeDecoratorTypeInvoker} under the hood.
```
liach:
```suggestion
	 * Creates a new instance of a {@link TreeDecoratorType}, registers and returns it.
```

liach (commented):
For javadocs, if you can make param/return use non-capitalized phrases, it's better. Indeed there is no need for indent for examples.

Btw you don't need to include implementation details (like what invoker it calls) in the javadoc.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Creates a new instance of a {@link FoliagePlacerType}, registers and returns it.<br>
+	 * Uses {@link FoliagePlacerTypeInvoker} under the hood.
+	 *
+	 * <p>
+	 *     Example:<br>
+	 *
+	 *    {@code
+	 *         public static final FoliagePlacerType<MyFoliagePlacer> MY_FOLIAGE_PLACER = FabricTreeRegistry.registerFoliagePlacerType(new Identifier("tutorial", "my_foliage_placer"), MyFoliagePlacer.CODEC);
```
warjort: This and other javadoc are using old refactored names.

FabricTreeRegistry -> TreeTypeHelper

warjort:
I don't think the BlockStateProvider and IntProvider registration belongs in a class about Trees.
They both have much more general usage in other world gen feature configuration.

Is there a reason why you can't also include FloatProvider and HeightProvider to complete the set?

warjort:
> Is there a reason why you can't also include FloatProvider and HeightProvider to complete the set?

Looking more closely at it, these are interfaces and so is IntProviderType.
There is no problem with a private constructor like the BlockStateProviderType

So the IntProvider registration method is really unnecessary?

warjort:
I don't see where you are registering your new mixins?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/tree/TreeTypeHelper.java
kanpov: I'm dumb. Sorry. Will change that in a sec

kanpov:
Thanks @warjort and @liach for your reviews! 

Let's go through your points.

1. Fixed the `p` usage
2. Fixed javadocs with `FabricTreeRegistry` instead of `TreeTypeHelper`
3. Registered the mixins in the config
4. Removed the unnecessary `IntProvider` utils
5. There will be no `FloatProvider` and `HeightProvider` utils as you can register them via the static methods

If there're any issues left, please report them. Thanks again for your support.

kanpov:
@sfPlayer1 @modmuss50 this PR is ready for review

fabric-object-builder-api-v1/src/main/resources/fabric-object-builder-v1.mixins.json
```diff
@@ -15,7 +15,11 @@
     "SpawnRestrictionAccessor",
     "TypeAwareTradeMixin",
     "VillagerProfessionAccessor",
-    "VillagerTypeAccessor"
+    "VillagerTypeAccessor",
+    "BlockStateProviderTypeInvoker",
+    "TreeDecoratorTypeInvoker",
+    "TrunkPlacerTypeInvoker",
+    "FoliagePlacerTypeInvoker"
```
liach: Sort them by alphabetical order.

kanpov: Okay, one second please. Imma open that random alphabetical order website.

kanpov: Done! Thanks for reporting!

liach (approved):

RedGrapefruit09:
I give up. I'm just going to make this a library. 

DM me on discord after 2000 years if this gets merged.

bye.

frqnny:
why the op ening and closing

PR #1518, opened 2021-06-23
Add client command source properties

xpple:
Allows the client to access various new fields in `FabricClientCommandSource`, or alternatively alter them with their own Mixin. This functionality is inspired by `ServerCommandSource`.


frqnny:
What would getRotation be used for when you can just do entity.getClientRotation yourself

xpple:
The idea of this PR is to have separate, independent fields. The default values however, are not independent. This means one could modify the `entity` field and not mess with the rotation.

xpple:
I must also note that this PR was preceded by a [conversation on Discord](https://discord.com/channels/507304429255393322/566276937035546624/855932005510545408).

liach:
Imo you should change the design to just include the `Map<String, Object> properties`, and this map should be an immutable one.

The map in the client command source will be immutable, in accordance with the immutable nature of command sources.

The new methods I recommend:
```java
<T> T getMeta(String key);
Map<String, ?> getMetas();
<T> FabricClientCommandSource withMeta(String key, T data); // returns a new client command source
FabricClientCommandSource withMeta(Map<String, ?> extraMeta); // returns a new client command source
```
the entity, position, and rotation info can go to the meta. I don't see why you need to change the world, as a client only loads one world at once.

Player3324:
The get* methods need to be as direct and convenient as possible, otherwise mods will ignore then and fetch the desired data through MinecraftClient or getPlayer, which defeats the point of this PR.

liach:
The `ClientPlayerEntity getPlayer()` needs to be revamped; it should be `AbstractClientPlayerEntity getPlayer()` so it can represent both the main client player and server-sent other players.

If we do add more explicit getters, I would recommend for each getter, a `FabricClientCommandSource withXxx(Xxx xxx)` (like those seen in server command source) for each of those, as to promote mods to use these methods to create immutable instances than using mixin to do hacks and break contracts of immutability.

xpple:
> If we do add more explicit getters, I would recommend for each getter, a `FabricClientCommandSource withXxx(Xxx xxx)` (like those seen in server command source) for each of those, as to promote mods to use these methods to create immutable instances than using mixin to do hacks and break contracts of immutability.

Sure! Thing is I discussed this with Player, and he said `with<>` methods were out of scope for the Fabric API and should be done by the client (what I'm doing right now). Adding the `with<>` methods would make life significantly easier, though.

xpple:
> The `ClientPlayerEntity getPlayer()` needs to be revamped; it should be `AbstractClientPlayerEntity getPlayer()` so it can represent both the main client player and server-sent other players.

Yep, good idea.

xpple:
> The get* methods need to be as direct and convenient as possible, otherwise mods will ignore then and fetch the desired data through MinecraftClient or getPlayer, which defeats the point of this PR.

I get that. I initially removed the `getPlayer` method as I found it to be pointless, but like Haykam said it would break mods which were previously using that method. I added it back while making sure it always returns some player, where `client.player` is a fallback. I guess this is more of a convenience method than an actual getter.

Edit: was this meant as a response to liach? If so I do agree with Player.

xpple:
> Imo you should change the design to just include the `Map<String, Object> properties`, and this map should be an immutable one.
> 
> The map in the client command source will be immutable, in accordance with the immutable nature of command sources.
> 
> The new methods I recommend:
> 
> ```java
> <T> T getMeta(String key);
> Map<String, ?> getMetas();
> <T> FabricClientCommandSource withMeta(String key, T data); // returns a new client command source
> FabricClientCommandSource withMeta(Map<String, ?> extraMeta); // returns a new client command source
> ```

I know that the `ServerCommandSource` is immutable too, but can we agree that their way of modifying the source by instantiating a new object is weird? Instead of modifying one field they essentially recreate the entire thing.

xpple:
> Imo you should change the design to just include the `Map<String, Object> properties`, and this map should be an immutable one.
> 
> The map in the client command source will be immutable, in accordance with the immutable nature of command sources.

What do others think about this solution? In my unexperienced opinion this isn't really convenient.

liach:
Hmm, after looking, the client command source is actually created only once for each client play network handler (each server login/savegame load). As a result, the modifications from this command will affect the source for the next command. So in your mixin, you need to ensure that these data fields are correctly set up and cleared before and after each client command execution/parsing. Good news is that the command source does not need to be immutable and is not concurrent (multithreaded)

Player3324:
My intention was to have only the getters in FabricClientCommandSource in the form of default methods that redirect to the existing getPlayer as best as possible. A 3rd party mod could then subclass ClientCommandSource and reimplement those methods to yield whatever it wants. The entirety of the changes necessary are the following extra methods in FabricClientCommandSource:

```java
default Entity getEntity() {
	return getPlayer();
}

default Vec3d getPosition() {
	return getPlayer().getPos();
}

default Vec2f getRotation() {
	return getPlayer().getRotationClient();
}

default Object getMeta(String key) {
	return null;
}
```

The mod which wants to override e.g. the position would then do something like this:

```java
class MyCcs extends ClientCommandSource implements FabricClientCommandSource {
	private final Vec3d position;

	MyCcs(ClientPlayNetworkHandler networkHandler, MinecraftClient client, Vec3d position) {
		super(networkHandler, client);

		this.position = position;
	}

	public Vec3d getPosition() {
		return position;
	}
}

```
and pass an instance of McCcs to the command.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/mixin/command/client/ClientCommandSourceMixin.java
```diff
@@ -38,6 +44,12 @@
 	@Final
 	private MinecraftClient client;
 
+	private Entity entity = client.player;
+	private Vec3d position = entity.getPos();
+	private Vec2f rotation = entity.getRotationClient();
+	private ClientWorld world = client.world;
+	private Map<String, Object> meta = new HashMap<>();
```
liach: These fields should be removed. The non-stub implementations would happen in mod-defined custom client command sources.

liach (approved):

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
```diff
@@ -56,12 +60,49 @@
 	 *
 	 * @return the player
 	 */
-	ClientPlayerEntity getPlayer();
+	AbstractClientPlayerEntity getPlayer();
```
liach: Hmm, this change may not be backward compatible...

I suggest declare another
```java
default AbstractClientPlayerEntity getTargetPlayer() {
    return getPlayer();
}
```

and possibly deprecate (or not) the existing `getPlayer()` as well.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/mixin/command/client/ClientCommandSourceMixin.java
```diff
@@ -38,6 +39,9 @@
 	@Final
 	private MinecraftClient client;
 
+	private Entity entity = client.player;
+	private ClientWorld world = client.world;
```
liach: Should just revert the changes to client command source mixin.

xpple:
> My intention was to have only the getters in FabricClientCommandSource in the form of default methods that redirect to the existing getPlayer as best as possible. A 3rd party mod could then subclass ClientCommandSource and reimplement those methods to yield whatever it wants. The entirety of the changes necessary are the following extra methods in FabricClientCommandSource:
> 
> ```java
> default Entity getEntity() {
> 	return getPlayer();
> }
> 
> default Vec3d getPosition() {
> 	return getPlayer().getPos();
> }
> 
> default Vec2f getRotation() {
> 	return getPlayer().getRotationClient();
> }
> 
> default Object getMeta(String key) {
> 	return null;
> }
> ```
> 
> The mod which wants to override e.g. the position would then do something like this:
> 
> ```java
> class MyCcs extends ClientCommandSource implements FabricClientCommandSource {
> 	private final Vec3d position;
> 
> 	MyCcs(ClientPlayNetworkHandler networkHandler, MinecraftClient client, Vec3d position) {
> 		super(networkHandler, client);
> 
> 		this.position = position;
> 	}
> 
> 	public Vec3d getPosition() {
> 		return position;
> 	}
> }
> ```
> 
> and pass an instance of McCcs to the command.

Sounds good but I don't see how I can pass an instance of `MyCcs` to a command when the builder methods like `literal()` return something with a type parameter of `FabricClientCommandSource`...

haykam821:
You should be able to cast it.

Player3324:
I don't really agree with the getPlayer changes, seems rather detrimental to me. Often you need the actual player with proper networking etc, substituting this use with any other player-like instance would at best not do anything, more likely just crash from naive blind casts. Almost everything else should be fine with getEntity and potentially some instanceof/casting.

Please make sure to API works for what it's supposed to do before we can merge it.

xpple:
> You should be able to cast it.

Yeah that's what I ended up doing. It's a bit ugly, though.

Regarding the `getPlayer()` method, I think it only extends what there is currently. If one would want to be 100% certain that the returned player is the client's player, `client.player` should probably be used instead.

I am still in the testing phase, whereafter I will fine-tune the changes.

xpple:
Actually, I get this error when I try to cast `FabricClientCommandSource` to my custom command source:
```
java.lang.ClassCastException: class net.minecraft.client.network.ClientCommandSource cannot be cast to class dev.xpple.seedmapper.command.CustomClientCommandSource (net.minecraft.client.network.ClientCommandSource and dev.xpple.seedmapper.command.CustomClientCommandSource are in unnamed module of loader net.fabricmc.loader.launch.knot.KnotClassLoader @4e0ae11f)
```

haykam821:
What is your mixin/interface structure?

xpple:
I pushed my testing to my Fabric mod in [this commit](https://github.com/xpple/SeedMapper/commit/f3ea94babd57eb779d608f8b3f2ebf0895ff3d5b). The source modifier command is the `SourceCommand` class.

liach:
First, thank you for this example!

Imo, instead of casting an arbitrary command source to your custom command source, you should add a method:
```java
public static CustomClientCommandSource of(ClientCommandSource src) {
  if (src instanceof CustomClientCommandSource custom) return custom;
  return new CustomClientCommandSource(/* fill in the parameters */);
}
```
This should resolve the issue you are having.

In addition, I suggest you keep a reference to the parent command source (or a function to its `getMeta` method), so when a key is absent in your source's meta at [here](https://github.com/xpple/SeedMapper/commit/f3ea94babd57eb779d608f8b3f2ebf0895ff3d5b#diff-d4597af3fcc9dbd6343288b8e944f9f86252610c4c254c0adab3dbd068e72b0cR79), you can always query the parent source.

xpple:
Thanks for the tips! I got the first things working:
image

xpple:
I got it all working with [this commit](https://github.com/xpple/SeedMapper/commit/13ee866964787236a34b0098ae884197811fbf09). Both my implementation and the new getters seem to work without issues.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
liach:
```suggestion
	ClientPlayerEntity getPlayer();
```
for binary compatibility (otherwise existing code calling `FabricClientCommandSource.getPlayer` expecting a `ClientPlayerEntity` return would get a no such method error)

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
```diff
@@ -55,13 +58,56 @@
 	 * Gets the player that used the command.
 	 *
 	 * @return the player
+	 * @deprecated Please use {@link #getTargetPlayer()} instead.
 	 */
-	ClientPlayerEntity getPlayer();
+	@Deprecated
+	AbstractClientPlayerEntity getPlayer();
+
+	default AbstractClientPlayerEntity getTargetPlayer() {
```
liach: Mind copy the javadocs over from the other method

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
```diff
@@ -55,13 +58,56 @@
...
 
 	/**
 	 * Gets the world where the player used the command.
 	 *
 	 * @return the world
 	 */
 	ClientWorld getWorld();
+
+	/**
+	 * Gets the meta property that was assigned to this <code>FabricClientCommandSource</code>.
```
liach:
```suggestion
	 * Gets the meta property under {@code key} that was assigned to this source.
	 *
	 * <p>This method should return the same result for every call with the same {@code key}.
```

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
```diff
@@ -55,13 +58,56 @@
 	 * Gets the player that used the command.
 	 *
 	 * @return the player
+	 * @deprecated Please use {@link #getTargetPlayer()} instead.
```
liach:
```suggestion
	 * @deprecated Please use {@link #getTargetPlayer()} for the user, and {@code getClient().player} for the client's player.
```
Now we actually need to distinguish the two cases.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
xpple: Sure

Player3324:
> Regarding the `getPlayer()` method, I think it only extends what there is currently. If one would want to be 100% certain that the returned player is the client's player, `client.player` should probably be used instead.

I don't think this addresses my point, how is there a sufficiently meaningful purpose in widening the scope of `getPlayer` that isn't already covered by `getEntity`?

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
xpple: Indeed.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
xpple: [This](https://github.com/FabricMC/fabric/pull/1518#discussion_r669211123) does not allow other players though, as suggested [by you](https://github.com/FabricMC/fabric/pull/1518#issuecomment-867296548). Do you want me to remove that feature?

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/api/client/command/v1/FabricClientCommandSource.java
xpple: [This](https://github.com/FabricMC/fabric/pull/1518#discussion_r669211123) does not allow other players though, as suggested [here](https://github.com/FabricMC/fabric/pull/1518#issuecomment-867296548) by you. Do you want me to remove that feature?

liach: Sorry, I am dumb! Player says that we can just use `getEntity` instead, so we proably will revert changes to this and remove `getTargetPlayer`; instead we will document this `getPlayer` simply returns `getClient().player` than the actual entity that executes the command, and ask users to do instanceof checks on the `getEntity` result.

xpple: Sure!

xpple:
Should `getPlayer()` be deprecated?

Player3324:
I'd just leave it as is for now.

Player3324 (approved):
LGTM, thanks!

Technici4n (approved):
Simple enough.

modmuss50 (approved):

PR #1540, opened 2021-07-05
Structure Pool Callback

frqnny:
Basically redos #1389 Even though it's exactly the same except FabricStructurePool is a record.
This PR adds a callback to allow players to modify structure pools. 

I was going to make it different but I decided not too- I was thinking of using DynamicRegistryLoadingEvent or whatever it's called, but that loads before so nah. So this is basically just retargetted to 1.17. 

Anyway, the important part is that this is data-pack safe. Repurposed Structures, for example, uses structure pool in json instead of using builtin and they show up using the testmod and RS together. I guess you could test this by making a RS addon? The best use case, though, will be to allow people to add houses to villages. Say you add a profession, you will need a house for that profession, this is the way you add it.


liach (approved):

liach:
Checkstyle is too old. It needs to be bumped to 8.40+ to support records (latest as of comment is 8.44)

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,66 @@
...
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 *
+	 * @param element the element to add
+	 * @param weight  Minecraft handles weight by adding it that amount of times into a list.}
+	 */
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		//adds to elementCounts list; minecraft makes these immutable lists so we replace them with an array list
+		StructurePoolAccessor pool = (StructurePoolAccessor) underlying();
+		if (pool.getElementCounts() instanceof ArrayList) {
```
liach:
```suggestion

		if (pool.getElementCounts() instanceof ArrayList) {
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.structure.pool.StructurePool;
+import net.minecraft.structure.pool.StructurePoolElement;
+
+import net.fabricmc.fabric.mixin.structure.StructurePoolAccessor;
+
+/**
+ * Represents a modifiable structure pool that has several helper methods for modders.
+ */
+public record FabricStructurePool(StructurePool underlying) {
+
```
liach:
```suggestion
```

liach (commented):
checkstyle

frqnny:
ill fix soon don’t worry

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,66 @@
...
+	 * @param weight  Minecraft handles weight by adding it that amount of times into a list.}
+	 */
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		//adds to elementCounts list; minecraft makes these immutable lists so we replace them with an array list
+		StructurePoolAccessor pool = (StructurePoolAccessor) underlying();
+
+		if (pool.getElementCounts() instanceof ArrayList) {
+			pool.getElementCounts().add(Pair.of(element, weight));
+		} else {
+			List<Pair<StructurePoolElement, Integer>> list = new ArrayList<>((pool).getElementCounts());
```
maityyy:
```java
(pool).getElementCounts()
```
???

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/DynamicRegistryManagerMixin.java
```diff
@@ -0,0 +1,46 @@
...
+import net.fabricmc.fabric.api.structure.v1.FabricStructurePool;
+import net.fabricmc.fabric.api.structure.v1.StructurePoolAddCallback;
+
+@Mixin(DynamicRegistryManager.class)
+public abstract class DynamicRegistryManagerMixin {
+	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private static <E> void load(RegistryOps<?> ops, DynamicRegistryManager manager, DynamicRegistryManager.Info<E> info, CallbackInfo ci, RegistryKey<? extends Registry<E>> registryKey) {
+		if (registryKey.equals(Registry.STRUCTURE_POOL_KEY)) {
+			for (E pool : manager.get(registryKey)) {
+				if (pool instanceof StructurePool) {
```
maityyy: can be simplified btw
```java
if (pool instanceof StructurePool p) {
    StructurePoolAddCallback.EVENT.invoker().onAdd(new FabricStructurePool(p));
}
```

frqnny: thanks for these catches

frqnny:
Ooh, got an idea. gonna expose the identifier so they don't have to do #underlying

frqnny:
Updated javadocs too. Should be great now :D

Technici4n:
Looks really nice and concise. Would there be a way to test this, such as a testmod ?

frqnny:
This PR is based off Structurized, a mod by Draylar and mantained by me
Here is slightly outdated TacoCraft adding corn farms to plains villages example 
https://github.com/frqnny/TacoCraft/blob/0491e061b467949ef1615dbef48af3cbed3ade02/src/main/java/io/github/frqnny/tacocraft/init/ModGen.java#L40
Here is golems galore adding its golems to Iron Golem pool to make its golems spawn in villages without breaking village iron farms. 
https://github.com/frqnny/golemsgalore/blob/c0fa0f5f5f017f7ac226ef92b5dc8749c6789fca/src/main/java/io/github/frqnny/golemsgalore/GolemsGalore.java#L39
the test mod and the javadocs have some examples

Technici4n (changes requested):
Please add a testmod.

Every time I consider reviewing this PR I am reminded that I'll have to `gradlew publishToMavenLocal` and clone another mod and test it there. This is too much effort, especially if an update breaks this addition and we need to quickly test it.

The testmod should ideally add a structure, validate that the callback gets invoked, and possibly even inspect the structure pool after server start to make sure the addition succeeded.

frqnny:
Will do sometime this week. 


EDIT:

it is now 9/28, sorry I haven't done it. I can't promise anything so this PR is on hold until I'm able to work on it again.

frqnny:
Just find a desert village and this should appear when testing. Might take a while, using /locatebiome, /locate and /tp (10k blocks away to find new villages)

frqnny:
image

build.gradle
```diff
@@ -175,7 +175,7 @@ allprojects {
 
 	checkstyle {
 		configFile = rootProject.file("checkstyle.xml")
-		toolVersion = "8.43"
+		toolVersion = "9.1"
```
modmuss50: Is there a reason for this update?

fabric-structure-api-v1/build.gradle
```diff
@@ -4,3 +4,6 @@ moduleDependencies(project, [
 		'fabric-lifecycle-events-v1',
 		'fabric-api-base'
 ])
+loom {
```
modmuss50: New line please.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,67 @@
...
+import net.minecraft.structure.pool.StructurePool;
+import net.minecraft.structure.pool.StructurePoolElement;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.mixin.structure.StructurePoolAccessor;
+
+/**
+ * Represents a modifiable structure pool that has several helper methods for modders.
+ */
+public record FabricStructurePool(StructurePool underlying, Identifier id) {
```
modmuss50: Is a record really the right choice here? An interface would be better imo:

- A record (and interface) shouldnt really contain logic
- It makes it easier for someone else (or another module) to re-impliment the API.
- It makes the API surface clear, I dont think its too bad here but would be better as a simple interface.

The impl could still be a record (if it makes sense) but the exposed API should really be an interface imo.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,67 @@
...
+		} else {
+			List<Pair<StructurePoolElement, Integer>> list = new ArrayList<>(pool.getElementCounts());
+			list.add(Pair.of(element, weight));
+			pool.setElementCounts(list);
+		}
+
+		//adds to elements list
+		for (int i = 0; i < weight; i++) {
+			pool.getElements().add(element);
+		}
```
modmuss50: This logic should be "hidden" away in an impl somewhere, moving to an interface will solve this.

modmuss50 (changes requested):

build.gradle
frqnny: it was a merge conflict that I resolved. it'd be complaining about it if I didn't, IDK why the former red is showing that it is 8.43 since I was just updating it to match 9.1

https://github.com/FabricMC/fabric/blob/fa34ed2c10c399ed6c973a56f3fda31cd47811f6/build.gradle#L142

frqnny:
I've gone ahead and updated Structurized to 1.18 and to match with this PR's latest changes. It required to real changes from 1.17 to 1.18, so the big thing is that it's an exact replica of this PR. There is still our new testing material, but I will also post tommorow-ish how we use this mod with my own mods in 1.18 to prove that this PR works. (Structurized 1.3.0 worked in 1.17, Structurized 1.4 works in 1.18) https://github.com/frqnny/structurized

maityyy:
Is this event needed for anything else? I think it's better to convert all the events needed to modify something into something like the BiomeModification API, which is much more convenient (probably).

frqnny:
The event can be used for manipulating villages, bastions, and other structure pools. The problem is that IMO something like BiomeModification would  be over-engineering the entire thing. We only do one event, which should be simple enough for modders (no more than 7 lines). Something similar to Biome Modification is a bit too much for something like a StructurePool as we can just give them the FabricStructurePool in an event and get done with it.

frqnny:
**All I want for Christmas... is**

~~YOUUUUUU~~ this pr to get reviews


jokes aside just a small subtle reminder to take your time and review when you are open, thanks devs for the amazing work and Merry Christmas.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,54 @@
...
+	 *
+	 * @param element the element to add
+	 */
+	void addStructurePoolElement(StructurePoolElement element);
+
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 *
+	 * @param element the element to add
+	 * @param weight  Minecraft handles weight by adding it that amount of times into a list.}
```
Juuxel:
```suggestion
	 * @param weight  Minecraft handles weight by adding it that amount of times into a list
```

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.structure.pool.StructurePoolElement;
+import net.minecraft.util.Identifier;
+
+/**
+ * Represents a modifiable structure pool that has several helper methods for modders.
+ */
+public interface FabricStructurePool {
+	/**
+	 * Adds a new {@link StructurePoolElement} to the {@link StructurePool}.
+	 * See the alternative {@link #addStructurePoolElement(StructurePoolElement, int)} for details.
```
Juuxel, line 29: What details? It has the exact same description excluding this line

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
```diff
@@ -0,0 +1,54 @@
...
+package net.fabricmc.fabric.api.structure.v1;
+
+import net.minecraft.structure.pool.StructurePool;
+import net.minecraft.structure.pool.StructurePoolElement;
+import net.minecraft.util.Identifier;
+
+/**
+ * Represents a modifiable structure pool that has several helper methods for modders.
+ */
+public interface FabricStructurePool {
```
Juuxel, line 26: Maybe `@ApiStatus.NonExtendable`?

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * A callback for newly added structure pools.
+ *
+ * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured data packs
+ * instead of the builtin Minecraft or mod resources.
+ *
+ * <p>Example usage:
+ * <pre>{@code
+ * StructurePoolAddCallback.EVENT.register(structurePool -> {
+ * 	if (structurePool.id().toString().equals("minecraft:village/common/butcher_animals")) {
```
Juuxel: This example is a bit ugly, I'd just add a field with that identifier and compare that instead of the string

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/StructurePoolAddCallback.java
```diff
@@ -0,0 +1,50 @@
...
+package net.fabricmc.fabric.api.structure.v1;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * A callback for newly added structure pools.
+ *
+ * <p><strong>Word of warning</strong>: Mods may be editing on the structure pool from user configured data packs
+ * instead of the builtin Minecraft or mod resources.
```
Juuxel, line 26: Out of scope for this PR, but a possible future improvement would be a method in `FabricStructurePool` to check whether it's from a data pack (some mods might not want to modify user data)

fabric-structure-api-v1/src/main/resources/fabric-structure-api-v1.accesswidener
```diff
@@ -0,0 +1,3 @@
+accessWidener v1 named
+accessible class net/minecraft/util/registry/DynamicRegistryManager$Info
+mutable    field    net/minecraft/structure/pool/StructurePool    elementCounts    Ljava/util/List;
```
Juuxel: No need for this if you mark your `@Accessor` setter with `@Mutable` in `StructurePoolAccessor`

fabric-structure-api-v1/src/testmod/java/net/fabricmc/fabric/test/structure/StructureTest.java
```diff
@@ -64,6 +68,16 @@
 				.adjustsSurface()
 				.register();
 		Registry.register(Registry.STRUCTURE_PIECE, new Identifier("fabric", "test_structure_piece"), PIECE);
+
+		//Basic Test of Callback
+		StructurePoolAddCallback.EVENT.register(structurePool -> LOGGER.info(structurePool.getId().toString()));
```
Juuxel:
```suggestion
		StructurePoolAddCallback.EVENT.register(structurePool -> LOGGER.info("Structure pool {} added", structurePool.getId()));
```

Juuxel (changes requested):

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
haykam821: I'm not sure this is a good parameter description since this is an implementation detail.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructurePoolImpl.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public void addStructurePoolElement(StructurePoolElement element) {
+		addStructurePoolElement(element, 1);
+	}
+
+	@Override
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		//adds to elementCounts list; minecraft makes these immutable lists, so we replace them with an array list
```
deirn: Maybe add a check that `weight > 0`?

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/StructurePoolAccessor.java
```diff
@@ -0,0 +1,38 @@
...
+import com.mojang.datafixers.util.Pair;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import net.minecraft.structure.pool.StructurePool;
+import net.minecraft.structure.pool.StructurePoolElement;
+
+@Mixin(StructurePool.class)
+public interface StructurePoolAccessor {
+	@Accessor(value = "elements")
```
deirn, line 31: Is the explicit target necessary? :thinking:

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
frqnny: Ok, I'll rephrase it a bit.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
frqnny: Doesn't that mean it shouldn't be implemented?
modmuss requested that the class and impl be split to also include an interface in case someone wants to use it.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/StructurePoolAccessor.java
frqnny: It doesn't hurt to have them. Unless there's some strict formatting rule about this, IMO it is better to have them as they make it extra clear what field they are supposed to access.

frqnny:
Thanks for the reviews, this PR has been updated accordingly. Hope to get this merged soon!

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/StructurePoolAccessor.java
```diff
@@ -31,6 +32,7 @@
 	List<StructurePoolElement> getElements();
 
 	@Accessor(value = "elementCounts")
+	@Mutable
```
Juuxel: Should be on the setter I think, but mixin might apply the `final` removal here too (even though it's a bit weird)

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
Juuxel: This API interface + impl class pattern is usually used so that we can change the impl's structure without breaking anything (or even move it to another module). I don't think users gain anything in particular from implementing this itf, but I guess it can stay how it is without the anno.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/StructurePoolAccessor.java
frqnny: Okay, I'll try to see what happens when I remove it.

frqnny: Seems to work fine, now it is just on the setter.

Player3324 (approved):
LGTM with the pending review comments resolved. Thanks!

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/DynamicRegistryManagerMixin.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKey;
+
+import net.fabricmc.fabric.api.structure.v1.StructurePoolAddCallback;
+import net.fabricmc.fabric.impl.structure.FabricStructurePoolImpl;
+
+@Mixin(DynamicRegistryManager.class)
+public abstract class DynamicRegistryManagerMixin {
+	@Inject(method = "load(Lnet/minecraft/util/dynamic/RegistryOps;Lnet/minecraft/util/registry/DynamicRegistryManager;Lnet/minecraft/util/registry/DynamicRegistryManager$Info;)V", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private static <E> void load(RegistryOps<?> ops, DynamicRegistryManager manager, DynamicRegistryManager.Info<E> info, CallbackInfo ci, RegistryKey<? extends Registry<E>> registryKey) {
```
deirn: Since the `DynamicRegistryManager.Info<E> info` isn't used in this inject, you can replace it with `@Coerce Object info` instead, thus making the AW not needed.

[Coerce javadoc](https://jenkins.liteloader.com/job/Mixin/javadoc/org/spongepowered/asm/mixin/injection/Coerce.html)

Juuxel (approved):

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/mixin/structure/DynamicRegistryManagerMixin.java
frqnny: That's cool, thanks for the heads up! Glad we could remove the AW.

deirn (approved):

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructurePool.java
modmuss50: This is fine, I dont believe we use that on other similar interfaces.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructurePoolImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+		//adds to elementCounts list; minecraft makes these immutable lists, so we replace them with an array list
+		StructurePoolAccessor pool = (StructurePoolAccessor) getUnderlyingPool();
+
+		if (pool.getElementCounts() instanceof ArrayList) {
+			pool.getElementCounts().add(Pair.of(element, weight));
+		} else {
+			List<Pair<StructurePoolElement, Integer>> list = new ArrayList<>(pool.getElementCounts());
+			list.add(Pair.of(element, weight));
+			pool.setElementCounts(list);
```
modmuss50: I think it would be best to make the list immutable before setting it back again. Would remove the need to check if its an arrary list.

If peformace is a concern (I dont think it is here?) setting the array back to an immutable one after the callback has been invoked would also be good.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructurePoolImpl.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	@Override
+	public void addStructurePoolElement(StructurePoolElement element, int weight) {
+		if (weight <= 0) {
+			throw new IllegalArgumentException("weight must be positive");
+		}
+
+		//adds to elementCounts list; minecraft makes these immutable lists, so we replace them with an array list
+		StructurePoolAccessor pool = (StructurePoolAccessor) getUnderlyingPool();
```
modmuss50: I would rename this to `poolAccessor` or something, as its hiding the `pool` field.

modmuss50 (approved):
API looks good, just a few minor things I'd change in the impl.

I will add this to last call. 👍

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructurePoolImpl.java
frqnny: All right, now it just temporarily replaces whatever collection with an array list, then sets it back as an Immutable List.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/impl/structure/FabricStructurePoolImpl.java
frqnny: done.

modmuss50: Looks great, thanks 👍

frqnny:
As of the latest commit, the callback is working, sadly a cave ate our cactus farm :(

PR #1541, opened 2021-07-05
Prevent SpriteFinder from returning Missing sprite for wrong atlas

grondag:
SpriteFinder was originally designed for use with block/item rendering. Since then, Mojang has added new sprite atlas textures and each atlas has it's own "missing" sprite.  However, SpriteFinder still returns the block/item missing sprite in all cases.  This sprite will almost certainly not render correctly when used with other atlases, causing confusion. 

This will be only a cosmetic problem in most cases, but mods with atlas-specific mixins or logic can have more negative impacts.
 
This PR causes SpriteFinder to return the correct, atlas-specific missing Sprite.

### DIFF
--- a/fabric-renderer-api-v1/build.gradle
+++ b/fabric-renderer-api-v1/build.gradle
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-renderer-api-v1"
-version = getSubprojectVersion(project, "0.4.3")
+version = getSubprojectVersion(project, "0.4.4")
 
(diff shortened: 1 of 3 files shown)


liach (approved):

Player3324 (approved):
LGTM, thanks!

Technici4n (approved):

Juuxel (approved):

modmuss50 (approved):

PR #1356, opened 2021-03-09
Fabric Transfer API: "fluid only" edition

Technici4n:
# Fabric Transfer API: "fluid only" edition
Building on #1234, this PR adds a new `fabric-transfer-api-v1` for transfer of generic game resources, and uses it to define a "Fluid API" for blocks.

This PR replaces #1278 that I deemed too big to be implemented and reviewed properly. Let's go for quality above quantity.

I am also hosting builds of this module on my "maven" so that it's easier to try it out: https://github.com/Technici4n/Technici4n-maven. Please give it a try, any feedback is appreciated!

## Problem scope
This PR is centered around fluid transfer between blocks: Reading fluids stored in a block, extracting fluids from a block and inserting fluids into a block.

Contrarily to #1278, this PR does NOT cover:
* item transfer between blocks,
* fluid-containing items such as buckets or portable tanks.

I will talk about these things a bit later.

### Transactions
This PR introduces the transaction concept. Essentially, a transaction allows setting checkpoints like in a video game: if something doesn't go as expected, the state of the game reverts to what it was at the previous checkpoint. Game objects that take part in a transaction are commonly referred to as "participants" or "endpoints".

Here is an example that hopefully makes this clearer:
```java
try (Transaction outerTransaction = Transaction.openOuter()) {
    /* (A) some transaction operations */
    try (Transaction nestedTransaction = outerTransaction.openNested()) {
        /* (B) more operations */
        nestedTransaction.commit(); // Validate the changes that happened in this transaction.
                                    // This is a nested transaction, so changes will only be applied if the outer
                                    // transaction is committed too.
    }
    /* (C) even more operations */
    outerTransaction.commit(); // This is an outer transaction: changes (A), (B) and (C) are applied.
}
// If we hadn't committed the outerTransaction, all changes (A), (B) or (C) would have been reverted.
```

[...]

PR #1553, opened 2021-07-09
Fabric Transfer API: item transfer and fluid-containing items.

Technici4n:
This PR expands #1356 to add an item transfer API, and item support for the fluid API. This PR is currently being tested in prod by Modern Industrialization, without issues so far.

If you want to test this, I am hosting experimental builds for in-dev testing:
```groovy
repositories {
    maven {
        name = "Technici4n"
        url = "https://raw.githubusercontent.com/Technici4n/Technici4n-maven/master/"
        content {
            includeGroup "net.fabricmc.fabric-api"
        }
    }
}
dependencies {
    modImplementation("net.fabricmc.fabric-api:fabric-api-lookup-api-v1:1.3.0+b246072f13")
    modImplementation("net.fabricmc.fabric-api:fabric-transfer-api-v1:1.1.0-pre.04+648b8b6c18")
}
```

## Item API
The Item transfer API is very similar to the Fluid transfer API:
* [`ItemVariant`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemVariant.java) is a trivial adaptation of `FluidVariant` to items.
* You will not be surprised to learn that the main interface for item transfer is `Storage<ItemVariant>`, of which instances can be retrieved through [`ItemStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemStorage.java).
* `Inventory`, `SidedInventory` and `PlayerInventory` compatibility is provided by the public [`InventoryStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java) and [`PlayerInventoryStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/PlayerInventoryStorage.java) helpers. Modders can feel free to use them for their needs, and this PR also uses the wrappers directly to register a fallback in `ItemStorage`.

Compatibility with vanilla droppers and hoppers is included as well: (notice how clean the API is to use! :smile:)
* [`DropperBlockMixin`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/DropperBlockMixin.java).
* [`HopperBlockEntityMixin`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/HopperBlockEntityMixin.java).

## Fluid-containing items
This one is a bit more tricky. As noted in #1356:
> There will be two challenges here: designing a context type C that is suitable for all kinds of inventories that may store fluid-containing items, and providing ad-hoc helpers to allow multiple mods to register fluid empty/fill actions for the same item (so that for example mod A could add a honey fluid to fill honey bottles and mod B could add a milk fluid and milk bottles - while allowing bottles to be filled with either fluid).

### `ContainerItemContext`
[`FluidStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorage.java) now contains an `ItemApiLookup<Storage<FluidVariant>, ContainerItemContext>` to expose and query fluid API instances for items.

[`ContainerItemContext`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java) allows fluid-containing items to interface with their containing inventory in a clean way: operations are reflected in the inventory directly so that operations can properly compound within transactions. It has been designed with both player inventories and foreign inventories (think bottling machine) in mind.

### `FluidStorage.CombinedItemApiProvider`
`FluidStorage` additionally contains some events that allow multiple `Storage<FluidVariant>` instances to be offered in a single API query, which is not possible through direct use of API Lookup.
Fabric API uses this to register handlers for buckets and water bottles in a compatible way.

### A few examples
I added the following public base implementations (one of them is used for full buckets):
* [`EmptyItemStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/EmptyItemStorage.java).
* [`FullItemStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/FullItemStorage.java).

Have a look at the implementation of the empty bucket storage as well:
* [`EmptyBucketStorage`](https://github.com/Technici4n/fabric/blob/item-transfer-api/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/fluid/EmptyBucketStorage.java).


Ueaj-Kerman:
If you just want to review this PR, [try this](https://github.com/FabricMC/fabric/pull/1553/files/5a2014037cb2cd4c92b8a63540e78eb7310646f5..b246072f5302e26c63ca2ff56ad832e6fdfff57e)

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * Try to transform as many as possibly of {@linkplain #getItemVariant() the current variant} into another variant.
+	 * That is, extract the old variant, and insert the same amount of the new variant instead.
+	 *
+	 * @param into The variant of the items after the conversion. May not be blank.
+	 * @param maxAmount The maximum amount of items to convert. May not be negative.
+	 * @param transaction The transaction this operation is part of.
+	 * @return A nonnegative integer not greater than maxAmount: the amount that was transformed.
+	 */
+	default long transform(ItemVariant into, long maxAmount, TransactionContext transaction) {
```
Ueaj-Kerman: I would change this into 
`long transform(ItemVariant from, long amount, ItemVariant to, long amount, TransactionContext ctx)`
For example, if an item were to drain water from buckets in the inventory, it may be desirable to keep all the buckets in the same positions in the inventory, in which case it could be called once for each bucket stack, but if stacking is desired, then the function can be called once with the total remainder of buckets.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,161 @@
...
+ * <p>Implementors of item APIs will generally want to:
+ * <ul>
+ *     <li>Query how much of the variant is in the inventory through {@link #getAmount}.</li>
+ *     <li>Transform some of the bound variant into another variant through {@link #transform}.</li>
+ * </ul>
+ */
+// TODO: update javadoc
+// TODO: wrapper around a SingleSlotStorage<ItemVariant>
+// TODO: wrapper around a StackReference maybe?
+public interface ContainerItemContext {
```
Ueaj-Kerman, line 90: Why doesn't this subclass Storage?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,161 @@
...
+	 * This may or may not include the slot of this context.
+	 *
+	 * @return An unmodifiable list containing additional slots of this context.
+	 */
+	List<SingleSlotStorage<ItemVariant>> getInventorySlots();
+
+	/**
+	 * Return the world of this context.
+	 */
+	World getWorld();
```
Ueaj-Kerman: imo this should be left out, if the position isn't provided, the world shouldn't be either. And the world/position isn't always known either. I would instead have an API API for this or something else

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorage.java
```diff
@@ -47,11 +57,55 @@
...
+	 *
+	 * <p>This is appropriate to use when multiple mods could wish to expose the Fluid API for some items,
+	 * for example when dealing with items added by the base Minecraft game such as buckets or empty bottles.
+	 * A typical usage example is a mod adding support for filling empty bottles with a honey fluid:
+	 * Fabric API already registers a storage for empty bottles to allow filling them with water through the event,
+	 * and a mod can register an event handler that will attach a second storage allowing empty bottles to be filled with its honey fluid.
+	 *
+	 * @throws IllegalStateException If an incompatible provider is already registered for the item.
+	 */
+	public static Event<FindForItem> findForItem(Item item) {
```
Ueaj-Kerman: With this there isn't really a good way to filter based on tags, or class or other. filters gooder and best in game

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
```diff
@@ -227,6 +227,13 @@ default A find(World world, BlockPos pos, C context) {
 	 */
 	Class<C> contextClass();
 
+	/**
+	 * Return the provider for the passed block (registered with one of the {@code register} functions), or null if none was registered (yet).
```
Juuxel, line 231:
```suggestion
	 * {@return the provider for the passed block (registered with one of the {@code register} functions), or null if none was registered (yet)}
```
An inline `@return` also puts it as the method return value description.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemApiLookup.java
```diff
@@ -154,6 +155,13 @@
 	 */
 	Class<C> contextClass();
 
+	/**
+	 * Return the provider for the passed item (registered with one of the {@code register} functions), or null if none was registered (yet).
```
Juuxel, line 159:
```suggestion
	 * {@return the provider for the passed item (registered with one of the {@code register} functions), or null if none was registered (yet)}
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/FullItemStorage.java
```diff
@@ -0,0 +1,97 @@
...
+import net.fabricmc.fabric.api.transfer.v1.storage.StoragePreconditions;
+import net.fabricmc.fabric.api.transfer.v1.storage.base.ExtractionOnlyStorage;
+import net.fabricmc.fabric.api.transfer.v1.storage.base.SingleSlotStorage;
+import net.fabricmc.fabric.api.transfer.v1.transaction.TransactionContext;
+
+/**
+ * Base implementation of a storage for a full item.
+ */
+// TODO: more javadoc
+public final class FullItemStorage implements ExtractionOnlyStorage<FluidVariant>, SingleSlotStorage<FluidVariant> {
```
Ueaj-Kerman: We should discuss an elegant solution to requerying. Consider that after any operation on an item, it's Storage may become invalid, so we need a good way to auto-update. I think this is within the capabilities of the current API but I'm not sure.

Juuxel (changes requested):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemStorage.java
```diff
@@ -0,0 +1,59 @@
...
+	private ItemStorage() {
+	}
+
+	static {
+		// Load generic vanilla api fallback
+		ItemStorage.SIDED.registerFallback((world, pos, state, blockEntity, direction) -> {
+			Block block = state.getBlock();
+			Inventory inventory = null;
+
+			if (block instanceof InventoryProvider) {
```
Ueaj-Kerman: we're on java 16 smh
```suggestion
			if (block instanceof InventoryProvider i) {
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/HopperBlockEntityAccessor.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+import org.spongepowered.asm.mixin.gen.Invoker;
+
+import net.minecraft.block.entity.HopperBlockEntity;
+import net.minecraft.inventory.Inventory;
+
+/**
+ * Hopper accessors, for use in {@link HopperBlockEntityMixin}.
```
Juuxel, line 27: Isn't `@Shadow` cleaner in this case if you're not using it outside of the mixin?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemStorage.java
```diff
@@ -0,0 +1,59 @@
...
+public final class ItemStorage {
+	public static final BlockApiLookup<Storage<ItemVariant>, Direction> SIDED =
+			BlockApiLookup.get(new Identifier("fabric:sided_item_storage"), Storage.asClass(), Direction.class);
+
+	private ItemStorage() {
+	}
+
+	static {
+		// Load generic vanilla api fallback
+		ItemStorage.SIDED.registerFallback((world, pos, state, blockEntity, direction) -> {
```
Ueaj-Kerman, line 84: I would recommend just calling HopperBlockEntity#getInventory, that way if any mods have mixed into there or mix into there later or if mojang adds a new inventory getting system thing, compatibility will be maintained. Make BlockApiAccess or whatever it's called more powerful by letting it declare whether it needs the blockstate and block entity, this way we can half-avoid querying the block entity and blockstate, since HopperBlockEntity#getInventory does that internally.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemVariant.java
```diff
@@ -0,0 +1,118 @@
...
+import net.fabricmc.fabric.impl.transfer.item.ItemVariantImpl;
+import net.fabricmc.fabric.api.transfer.v1.storage.TransferVariant;
+
+/**
+ * An immutable count-less ItemStack, i.e. an immutable association of an item and an optional NBT compound tag.
+ *
+ * <p>Do not implement, use the static {@code of(...)} functions instead.
+ */
+@ApiStatus.NonExtendable
+public interface ItemVariant extends TransferVariant<Item> {
```
Ueaj-Kerman, line 43: override the `hashCode` and `equals` method and state that the method is implemented. I found it annoying in my spigot days when this wasn't documented

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/CursorSlotWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+public class CursorSlotWrapper extends SnapshotParticipant<ItemStack> implements SingleSlotStorage<ItemVariant> {
+	private static final Map<ScreenHandler, CursorSlotWrapper> WRAPPERS = new MapMaker().weakKeys().makeMap();
+
+	public static CursorSlotWrapper get(ScreenHandler screenHandler) {
+		return WRAPPERS.computeIfAbsent(screenHandler, CursorSlotWrapper::new);
+	}
+
+	private final ScreenHandler screenHandler;
+
+	private CursorSlotWrapper(ScreenHandler screenHandler) {
```
Ueaj-Kerman, line 40: might wana pass the player as well, iirc in 1.16 the cursor slot was on the player rather than the screen handler. They may change it back, or it'll be important if we backport the api

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/CursorSlotWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+import net.minecraft.screen.ScreenHandler;
+
+import net.fabricmc.fabric.api.transfer.v1.item.ItemVariant;
+import net.fabricmc.fabric.api.transfer.v1.storage.StoragePreconditions;
+import net.fabricmc.fabric.api.transfer.v1.storage.base.SingleSlotStorage;
+import net.fabricmc.fabric.api.transfer.v1.transaction.TransactionContext;
+import net.fabricmc.fabric.api.transfer.v1.transaction.base.SnapshotParticipant;
+
+public class CursorSlotWrapper extends SnapshotParticipant<ItemStack> implements SingleSlotStorage<ItemVariant> {
+	private static final Map<ScreenHandler, CursorSlotWrapper> WRAPPERS = new MapMaker().weakKeys().makeMap();
```
Juuxel: This seems to be the same as the JDK `WeakHashMap`, which is a bit nicer than a 3rd party library class IMO.

Ueaj-Kerman: no it's not the same thing, this one does reference equality. Though not sure if that's actually needed

Technici4n: `WeakHashMap` is not thread-safe, and this can be accessed concurrently from the client and server threads. This is why most maps in the API use `MapMaker`.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/CursorSlotWrapper.java
Technici4n: Will just be regular MC update breakage imo.
Re backporting: it's useless imo cause no mods will support it.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemVariant.java
Technici4n: I'll do it in `TransferVariant` directly.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/HopperBlockEntityAccessor.java
Technici4n: I don't think that works as the functions in `HopperBlockEntityMixin` are all static.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
Technici4n: Not sure tbh, this is slightly harder to read in java form, and makes the javadoc redundant. Also, intellij doesn't seem to properly show it (it shows it like a regular `@return`). Probably a good reason not to do this.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemVariant.java
Technici4n: Done.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: In-place modification can already be done through the slot list.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: Position is terrible, agreed. It just feels wrong.

However `World` makes sense to access persistent state and similar global state I think. Otherwise ender buckets have to rely on lifecycle events to capture server instances, and get the client world through the `MinecraftClient` singleton. Not yet totally sure, quite low cost for players though (the entity contains the world already).

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/FullItemStorage.java
Technici4n: This is already handled by the 3 storage implementations I wrote, will require good javadoc to make sure nobody forgets about it though:
```java
		// Can't insert if the item is not emptyItem anymore.
		if (!context.getItemVariant().isOf(emptyItem)) return 0;
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemStorage.java
Technici4n: Block entities are cached by `BlockApiCache` already, making them very cheap to access in the providers.

If mods have a custom inventory system, they should just register a fallback provider in the lookup. If mojang changes something, the API will require updating but that doesn't sound like a big deal.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorage.java
Technici4n: Perhaps I can add a fallback event to have the same flexibility as API Lookup itself? Zero additional effort! 😆

Technici4n: Done. Maybe this is worth adding to the api api itself.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: I did another pass on this, have a look at the new design!

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/HopperBlockEntityAccessor.java
Juuxel: Ah, alright. I didn't look at the uses carefully enough apparently.

Technici4n: Np ^^

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/CursorSlotWrapper.java
Technici4n: Fyi in 1.16 it didn't require screen handlers (only player entity)... these kinds of changes are impossible to predict, and we should just have as few parameters as possible.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorage.java
Technici4n: Update: not worth it cause the `ContainerItemContext` allows dropping the `ItemStack` parameter for this particular use case.

Technici4n:
I updated this a lot today... the API should be close to final now hopefully!

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
liach: Eh, shouldn't really use intellj bug as an excuse. This both leaves this statement at the beginning of the javadoc and adds a corresponding "return" section in generated javadoc

Technici4n: I find the original javadoc less readable then. Duplicating the description just feels stupid to me, automatic or not.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/FullItemStorage.java
Ueaj-Kerman: There should be a utility class to do this imo, something like:
```java
Storage<FluidKey> firstKey = ItemAPIs.firstKey(myStorage);
Storage<FluidKey> firstKey = ItemAPIs.combineAll(myStorage);
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,192 @@
...
+	/**
+	 * Try to transform as many as possibly of {@linkplain #getItemVariant() the current variant} into another variant.
+	 * That is, extract the old variant, and insert the same amount of the new variant instead.
+	 *
+	 * @param into The variant of the items after the conversion. May not be blank.
+	 * @param maxAmount The maximum amount of items to convert. May not be negative.
+	 * @param transaction The transaction this operation is part of.
+	 * @return A nonnegative integer not greater than maxAmount: the amount that was transformed.
+	 */
+	default long transform(ItemVariant into, long maxAmount, TransactionContext transaction) {
```
Player3324: exchange

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,192 @@
...
+		return 0;
+	}
+
+	/**
+	 * Return the main slot of this context.
+	 */
+	SingleSlotStorage<ItemVariant> getMainSlot();
+
+	/**
+	 * Try to insert overflow items into this context.
```
Player3324: this is not descriptive enough, no idea what it is supposed to do

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -0,0 +1,192 @@
...
+	 *
+	 * @see Storage#insert
+	 */
+	long insertOverflow(ItemVariant itemVariant, long maxAmount, TransactionContext transactionContext);
+
+	/**
+	 * Get additional slots that may be available in this context.
+	 * These may or may not include the main slot of this context.
```
Player3324: this caveat may be too vague, please at least include the rationale in the docs

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorage.java
```diff
@@ -49,11 +67,97 @@
...
+	 *
+	 * <p>This is appropriate to use when multiple mods could wish to expose the Fluid API for some items,
+	 * for example when dealing with items added by the base Minecraft game such as buckets or empty bottles.
+	 * A typical usage example is a mod adding support for filling empty bottles with a honey fluid:
+	 * Fabric API already registers a storage for empty bottles to allow filling them with water through the event,
+	 * and a mod can register an event handler that will attach a second storage allowing empty bottles to be filled with its honey fluid.
+	 *
+	 * @throws IllegalStateException If an incompatible provider is already registered for the item.
+	 */
+	public static Event<CombinedItemApiProvider> combinedItemApiProvider(Item item) {
```
Player3324, line 104: the method name and interface name need improving, I don't have a good one yet

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
```diff
@@ -0,0 +1,112 @@
...
+	/**
+	 * A wrapper around a player inventory.
+	 *
+	 * <p>Note that this is a wrapper around all the slots of the player inventory.
+	 * This may cause direct insertion to insert arbitrary items into equipment slots or other unexpected behavior.
+	 * To prevent this, {@link #offerOrDrop} is recommended for simple insertions.
+	 * {@link #getSlots} can also be used to retrieve a wrapper around a range of slots.
+	 */
+	@ApiStatus.NonExtendable
+	interface Player extends InventoryStorage {
```
Player3324: I'd suffix this with Storage

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
```diff
@@ -0,0 +1,112 @@
...
+	 * Return a wrapper around the inventory of a player.
+	 *
+	 * @see Player
+	 */
+	static Player ofPlayer(PlayerInventory playerInventory) {
+		return (Player) of(playerInventory, null);
+	}
+
+	/**
+	 * Return a wrapper around the cursor slot of a screen handler.
```
Player3324: this should describe what the cursor slot is

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventorySlotWrapper.java
```diff
@@ -0,0 +1,125 @@
...
+
+			if (inventory.isValid(slot, keyStack)) {
+				int inserted = Math.min(keyStack.getMaxCount(), count);
+				this.updateSnapshots(transaction);
+				keyStack.setCount(inserted);
+				inventory.setStack(slot, keyStack);
+				return inserted;
+			}
+		} else if (key.matches(stack)) {
+			int inserted = Math.min(stack.getMaxCount() - stack.getCount(), count);
```
lilybeevee: Merging stacks here doesn't account for `inventory.getMaxCountPerStack()`

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: Better now?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventorySlotWrapper.java
Technici4n:
```java
int count = (int) Math.min(maxAmount, inventory.getMaxCountPerStack());
```
at the beginning of the function accounts for it!

Technici4n: Well no that was wrong in case some items were already in the slot, however my new commit fixes this! Thanks!

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/ItemStorage.java
Technici4n: Not suitable anymore anyway due to transactions requiring special handling. We can't support the full MC API.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/SidedInventorySlotWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+			return slotWrapper.insert(resource, maxAmount, transaction);
+		}
+	}
+
+	@Override
+	public long extract(ItemVariant resource, long maxAmount, TransactionContext transaction) {
+		if (!sidedInventory.canExtract(slotWrapper.slot, resource.toStack(), direction)) {
+			return 0;
+		} else {
+			return slotWrapper.insert(resource, maxAmount, transaction);
```
Juuxel: This is suspicious, shouldn't it be `extract`?

Technici4n: Copy pasta...

Technici4n: Fixed, thanks :smile:

Technici4n:
Updated with the gradle setup to test this PR in-dev.

Player3324 (approved):
LGTM, thanks!

fabric-transfer-api-v1/build.gradle
```diff
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-transfer-api-v1"
-version = getSubprojectVersion(project, "1.0.0")
+version = getSubprojectVersion(project, "1.1.0-pre.05")
```
modmuss50: Can you leave the versions alone please? This is not right.

Technici4n: No worries, this was just for pre-merge testing in case multiple mods make use of it.

modmuss50 (approved):
Looks great 👍 Thanks.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ComposterWrapper.java
```diff
@@ -0,0 +1,206 @@
...
+			location.world.syncWorldEvent(WorldEvents.COMPOSTER_USED, location.pos, increaseSuccessful ? 1 : 0);
+		}
+
+		// Reset after successful commit.
+		increaseProbability = DO_NOTHING;
+	}
+
+	/**
+	 *
+	 */
```
Juuxel: Should this be here?

Juuxel (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: Removed `getWorld()`.

PR #1554, opened 2021-07-11
Fix FuelRegistry not following vanilla checks for valid fuels

### LINKED ISSUE #1476: Non-flammable wood is not respected for tags added to the fuel registry

Juuxel:
Closes #1476.

Unlike adding the fuel items directly, calling `AbstractFurnaceBlockEntity.addFuel` checks that the fuel is not excluded in `#minecraft:non_flammable_wood`. This makes Fabric API fuels match the vanilla behaviour, which is also configurable via data packs.

(I had to use an AW instead of an accessor due to SpongePowered/Mixin#399 - the two `addFuel` overloads broke the Mixin AP.)

### DIFF
--- a/fabric-content-registries-v0/build.gradle
+++ b/fabric-content-registries-v0/build.gradle
@@ -1,5 +1,9 @@
 archivesBaseName = "fabric-content-registries-v0"
-version = getSubprojectVersion(project, "0.2.2")
+version = getSubprojectVersion(project, "0.2.3")
+
+loom {
+	accessWidener = file("src/main/resources/fabric-content-registries-v0.accesswidener")
+}
 
 moduleDependencies(project, [
 		'fabric-api-base',
--- a/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FuelRegistryImpl.java
+++ b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FuelRegistryImpl.java
@@ -117,9 +117,7 @@ public final class FuelRegistryImpl implements FuelRegistry {
 					map.remove(i);
 				}
 			} else {
-				for (Item i : tag.values()) {
-					map.put(i, time);
-				}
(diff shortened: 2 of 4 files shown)


Technici4n (approved):
LGTM, a testmod would be ideal to quickly validate these changes (#1491).

liach (approved):

haykam821 (approved):

frqnny (approved):

fabric-content-registries-v0/build.gradle
```diff
@@ -1,5 +1,9 @@
 archivesBaseName = "fabric-content-registries-v0"
-version = getSubprojectVersion(project, "0.2.2")
+version = getSubprojectVersion(project, "0.2.3")
+
+loom {
+	accessWidener = file("src/main/resources/fabric-content-registries-v0.accesswidener")
```
modmuss50, line 5: This should be a mixin. Using an AW should only be done when required. Especially in large multi project builds like fab api as it will slow things down.

modmuss50 (changes requested):

fabric-content-registries-v0/build.gradle
Juuxel: ...and the requirement is right there in the PR description:

> (I had to use an AW instead of an accessor due to SpongePowered/Mixin#399 - the two addFuel overloads broke the Mixin AP.)

Player3324 (approved):

fabric-content-registries-v0/build.gradle
modmuss50: Ah my bad. This is fine then.

modmuss50 (approved):

PR #1555, opened 2021-07-11
Working tag list registry

### LINKED ISSUE #1467: 1.17 Tag API
With 1.17 coming soon to be released it would be a good idea to bring back [this proof of concept](https://github.com/modmuss50/fabric/tree/FabricRequiredTagListRegistry) and make it an actual feature than just a proof of concept.

This would allow the use of biome tags, or anything which has a registry for (dimensions tag, etc.) and would be a very welcomed feature.

deirn:
Adds ability to register custom tag lists aside from built-in ones.
The mixin also makes the game doesn't crash when registering multiple tag lists with the same `registryKey` and `dataType`, it'll simply return the same list.

Contains test for both `RequiredTagList#add` and `TagRegistry#create` along with JSON only tag that doesn't mentioned in code.

Closes #1467


frqnny (commented):
Nice! Looks awesome. I’d just add some javadoc on the mixin to explaining how  other fapi contributors can add things like biome tags

deirn:
Adding biome tags (or any tags) would be the same way how vanilla did with block tags, calling `RequiredTagList#add`. I think a yarn PR adding javadoc to require this module would work better. The testmod adds a biome tag, btw.

Also speaking of it, 1.17 makes fabric's `TagRegistry` a little weird to use IMO, I actually wanted to deprecate this module and make a v1 but I changed my mind 😅

frqnny:
Yes, this pr does fine. 

Yarn javadocs should not contain references to FAPI or mods, they are just there to explain and document MC code. It is standard for mixins to have javadocs, i.e. biome modification having a big javadoc in one of its mixins

deirn:
@frqnny added docs on the mixin, I hope it's clear enough

deirn:
Marking this as a draft, apparently there's a problem with the tag being loaded far before dynamic registries making biome tags (and probably other) failed to load with biome datapack.

deirn:
Replaced with #1562

PR #1562, opened 2021-07-13
Tag Factory API

### LINKED ISSUE #1467: 1.17 Tag API

deirn:
This PR adds `TagFactory<T>` that can be used to make tags for any given registry, works for both static or dynamic.
This PR also deprecates `TagRegistry` in favor of static factories in `TagFactory`.

### Example
To make biome tags, what would you do is the following
```java
TagFactory<Biome> factory = TagFactory.of(Registry.BIOME_KEY, "tags/biomes");
Tag<Biome> biomeTag = factory.create(new Identifier("test:biomes"));
```

### Problem with dynamic registries
Minecraft calls `ServerResourceManager#reload` before loading dynamic registries, making tags for them failed to load if the tag mentions datapack entries because it's not yet resolved. The workaround that I can think of is to delay loading the dynamic reg tag until the registry finished loading.

### Testmod
The testmod contains a really basic biome JSON, just to make sure the tag loading works.

Replaces #1555 
Closes #1467


fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/TagFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A factory for accessing datapack tags.
+ */
+public interface TagFactory<T> {
+	TagFactory<Item> ITEM = of(ItemTags::getTagGroup);
+	TagFactory<Block> BLOCK = of(BlockTags::getTagGroup);
+	TagFactory<Fluid> FLUID = of(FluidTags::getTagGroup);
+	TagFactory<GameEvent> GAME_EVENT = of(GameEventTags::getTagGroup);
+	TagFactory<EntityType<?>> ENTITY_TYPE = of(EntityTypeTags::getTagGroup);
```
LambdAurora, line 48: Imo, I would offer a default implementation of the biomes tag, it should be something that could get widely adopted, and it could also encourage people to use it.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/FabricTagManagerHooks.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package net.fabricmc.fabric.impl.tag.extension;
+
+import net.minecraft.tag.TagGroup;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKey;
+
+public interface FabricTagManagerHooks {
+	void fabric_addTagGroup(RegistryKey<? extends Registry<?>> registryKey, TagGroup<?> tagGroup);
```
LambdAurora, line 24:
```suggestion
	void fabric$addTagGroup(RegistryKey<? extends Registry<?>> registryKey, TagGroup<?> tagGroup);
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagManager.java
```diff
@@ -0,0 +1,54 @@
...
+	private Map<RegistryKey<? extends Registry<?>>, TagGroup<?>> tagGroups;
+
+	@Inject(method = "<init>", at = @At("TAIL"))
+	private void init(Map<RegistryKey<? extends Registry<?>>, TagGroup<?>> tagGroups, CallbackInfo ci) {
+		// Make it mutable so we can add dynamic registry tags later.
+		this.tagGroups = new HashMap<>(tagGroups);
+	}
+
+	@Override
+	public void fabric_addTagGroup(RegistryKey<? extends Registry<?>> registryKey, TagGroup<?> tagGroup) {
```
LambdAurora, line 51:
```suggestion
	public void fabric$addTagGroup(RegistryKey<? extends Registry<?>> registryKey, TagGroup<?> tagGroup) {
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinMinecraftServer.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.resource.ServerResourceManager;
+import net.minecraft.server.MinecraftServer;
+
+import net.fabricmc.fabric.impl.tag.extension.TagFactoryImpl;
+
+@Mixin(MinecraftServer.class)
+public class MixinMinecraftServer {
```
LambdAurora: This mixin is useless, [in my testing calling the lifecycle event `SERVER_STARTING` is sufficient](https://github.com/LambdAurora/LambdaFoxes/blob/1.17/src/main/java/dev/lambdaurora/lambdafoxes/LambdaFoxes.java#L67).

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
```diff
@@ -0,0 +1,106 @@
...
+	 * {@link RegistryOps#method_36574} for the relevant code.
+	 */
+	public static void loadDynamicRegistryTags(MinecraftServer server) {
+		server.getProfiler().push("fabricDynamicRegistryTagsReload");
+		DYNAMICS.forEach(tagList -> {
+			RegistryKey<? extends Registry<?>> registryKey = tagList.getRegistryKey();
+			Registry<?> registry = server.getRegistryManager().get(registryKey);
+			TagGroupLoader<?> tagGroupLoader = new TagGroupLoader<>(registry::getOrEmpty, tagList.getDataType());
+			TagGroup<?> tagGroup = tagGroupLoader.load(server.getResourceManager());
+			((FabricTagManagerHooks) server.getTagManager()).fabric_addTagGroup(registryKey, tagGroup);
```
LambdAurora:
```suggestion
			((FabricTagManagerHooks) server.getTagManager()).fabric$addTagGroup(registryKey, tagGroup);
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinMinecraftServer.java
deirn: I guess the `runServer` one can be removed, but we still need the datapack reload one though since we wanted to reload the tags before syncing packet and `END_DATAPACK_RELOAD` is called after the sync.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagManager.java
deirn: Some mixin use `$` and other use `_` so I'm not really sure. I just used what existing mixin on this module use.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinMinecraftServer.java
LambdAurora: > I guess the `runServer` one can be removed, but we still need the datapack reload one though since we wanted to reload the tags before syncing packet and `END_DATAPACK_RELOAD` is called after the sync.

In my testing the other one is not needed at all actually, `/reload` doesn't have the same issues as server starting.

The dynamic registry seems to not be rebuilt after tags when a reload happen, or it might be a side-effect of worldgen not being reloadable? Would need more testing.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagManager.java
LambdAurora: iirc the current convention is `$`, it's often used as a separator while `_` is very snake_case-y

Technici4n: I've always used _, just pick whatever.

Technici4n: (if it's an existing module, use the style of the module)

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinMinecraftServer.java
deirn: Looking through LambdaFoxes' code, it doesn't seem to address `Couldn't load tag {} as it is missing following references:` error that thrown when it tries to load dynamic registry tags. I [deliberately](https://github.com/goodasintended/fabric/blob/cba6f29aaefff3765ee9cc8d342c1d3e793f9128/fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinTagManagerLoader.java#L38) disabled loading those tags from `ServerResourceManager#reload` to make it doesn't error so `loadDynamicRegistryTags` needs to be called on both setup and reload. It's a "cosmetic" only error though, I can remove those if you guys have a problem with that part.

LambdAurora: Uh, after further testing, indeed, and it's quite a big error as it will prevent the application of datapacks then.

In my initial testing I did not have tags referencing datapack biomes so I wasn't aware of this detail.

deirn: I think the server would fail only if you make your tag via `RequiredTagList#add`, it'll throw `Missing required tags: ` alongside that error.

LambdAurora: I would move the runServer one to the lifecycle event, and let the other one as a mixin then.

frqnny:
Looks very nice, I will review later. It *would* be nice to build up actual tags into this later (specially for biome) and then allow Biome Modification API to allow you to use a selector with api's. Perhaps adding a util method to check if it is in a tag would be nice?

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/MixinRequiredTagListRegistry.java
```diff
@@ -0,0 +1,42 @@
...
+
+import net.fabricmc.fabric.impl.tag.extension.TagFactoryImpl;
+
+@Mixin(RequiredTagListRegistry.class)
+public class MixinRequiredTagListRegistry {
+	@Inject(method = "getBuiltinTags", at = @At("TAIL"), cancellable = true)
+	private static void getBuiltinTags(CallbackInfoReturnable<Set<RequiredTagList<?>>> cir) {
+		// Add tag lists registered on fabric to the map.
+		Set<RequiredTagList<?>> set = new HashSet<>();
+		set.addAll(cir.getReturnValue());
```
LambdAurora: Why not do `new HashSet<>(cir.getReturnValue())` directly?

It would directly allocate what's needed.

fabric-tag-extensions-v0/src/testmod/java/net/fabricmc/fabric/test/tag/extension/TagExtensionTest.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKey;
+import net.minecraft.world.biome.Biome;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.command.v1.CommandRegistrationCallback;
+import net.fabricmc.fabric.api.tag.TagFactory;
+
+public class TagExtensionTest implements ModInitializer {
+	static final TagFactory<Biome> BIOME_TAGS = TagFactory.of(Registry.BIOME_KEY, "tags/biomes");
```
LambdAurora: Since `TagFactory` has the BIOMES one directly, the testmod should use it directly too :P

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
```diff
@@ -0,0 +1,106 @@
...
+	 * <p>Look at server's {@link Main#main} function calls for {@link ServerResourceManager#reload} and
+	 * {@link RegistryOps#method_36574} for the relevant code.
+	 */
+	public static void loadDynamicRegistryTags(MinecraftServer server) {
+		server.getProfiler().push("fabricDynamicRegistryTagsReload");
+		DYNAMICS.forEach(tagList -> {
+			RegistryKey<? extends Registry<?>> registryKey = tagList.getRegistryKey();
+			Registry<?> registry = server.getRegistryManager().get(registryKey);
+			TagGroupLoader<?> tagGroupLoader = new TagGroupLoader<>(registry::getOrEmpty, tagList.getDataType());
+			TagGroup<?> tagGroup = tagGroupLoader.load(server.getResourceManager());
```
liach: Should we asynchronously call `loadTags` first before we call each `buildGroup`, like what vanilla does?

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/AccessorDynamicRegistryManager.java
```diff
@@ -16,17 +16,19 @@
...
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKey;
 
-@Mixin(FluidTags.class)
-public interface AccessorFluidTags {
-	@Accessor("REQUIRED_TAGS")
-	static RequiredTagList<Fluid> getRequiredTags() {
-		throw new UnsupportedOperationException();
+@Mixin(DynamicRegistryManager.class)
+public interface AccessorDynamicRegistryManager {
```
liach: Better named `DynamicRegistryManagerAccess`

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
deirn: That's the case with function tags but not for regular registry tags, it just calls `load` directly.
I don't think async is necessary for this, calls for `ServerResourceManager#reload` on server setup are immediately followed with `get` ¯\\\_(ツ)\_/¯

liach: Yeah, this doesn't quite affect the validity of these tags. 

I am currently exploring if there is any other fault on the general api structure (such as how you track dynamic registry entries). Beyond that, I think this should be mostly good.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/TagFactory.java
```diff
@@ -0,0 +1,66 @@
...
+	 */
+	static <T> TagFactory<T> of(RegistryKey<? extends Registry<T>> registryKey, String dataType) {
+		return TagFactoryImpl.of(registryKey, dataType);
+	}
+
+	static <T> TagFactory<T> of(Supplier<TagGroup<T>> tagGroupSupplier) {
+		return TagFactoryImpl.of(tagGroupSupplier);
+	}
+
+	Tag<T> create(Identifier id);
```
Juuxel: Tag.Identified would be a nicer return value since you already know the ID and the impl value already is an instance of it.

Juuxel (changes requested):

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/mixin/tag/extension/AccessorDynamicRegistryManager.java
LambdAurora: I thought we named accessor `<base class>Accessor` instead of Access? When did this change?

Technici4n:
I wonder if this shouldn't be a v1 module instead? Whether the other "magic interface" hooks present in tag-extensions-v0 are useful or not needs closer investigation.

LambdAurora:
> I wonder if this shouldn't be a v1 module instead? Whether the other "magic interface" hooks present in tag-extensions-v0 are useful or not needs closer investigation.

Wondered the same thing, would make dropping the deprecated part easier as it's literally dropping the entire v0 module.
Though it would worsen the "too many modules" issue in the meantime.

deirn:
The decision to make `tag-extensions` its own module is a little weird tbh, I'd just make it a part of `object-builders`.

OroArmor:
no object builders needs to be removed. it is the trashcan for unknown apis. its fine to make it its own module

deirn:
For `FabricDataGeneratorTagBuilder`, it's actually not needed anymore since vanilla has it now, opened a yarn PR to [rename it](https://github.com/FabricMC/yarn/pull/2585).

liach:
We should really rename object builder to access api, as the api mainly opens up access to hardcoded elements in vanilla. Any additional behavioral changes to the game though should go to another api. Imo this tag one can stay in another api for now as the post-registry loading of tag may somewhat change vanilla behaviors? (I still need to read through yarn code to determine how it works first)

LambdAurora:
> We should really rename object builder to access api, as the api mainly opens up access to hardcoded elements in vanilla. Any additional behavioral changes to the game though should go to another api. Imo this tag one can stay in another api for now as the post-registry loading of tag may somewhat change vanilla behaviors? (I still need to read through yarn code to determine how it works first)

This definitely modifies Vanilla behavior by delaying some tag loading (else Vanilla would load all tags at the same time), which is required due of the dynamic registry, which needs some tags to be built, and some tags using dynamic registry as lookup will need it to be built too.

Technici4n:
> We should really rename object builder to access api, as the api mainly opens up access to hardcoded elements in vanilla. Any additional behavioral changes to the game though should go to another api. Imo this tag one can stay in another api for now as the post-registry loading of tag may somewhat change vanilla behaviors? (I still need to read through yarn code to determine how it works first)

This would be a breaking change, so can't do that. And it is additionally somewhat offtopic for this issue. Once we get global AWs going, we can look into deprecating it.

Regarding this issue, I think `fabric-tag-api-v1` is fine. Ordering issues with dynamic registries need to be solved though. Regarding biome modifications, we have to be extra careful with ordering. We probably need biome modifications to apply after the biome tags, so they can be used as selectors.

deirn:
> We probably need biome modifications to apply after the biome tags, so they can be used as selectors.

There's another problem:
Currently, built-in biomes would get modified before datapacks being loaded, it'll simply be too late to use tags. 🙃 

Edit: [well that was pretty easy to fix lol](https://github.com/FabricMC/fabric/pull/1571).

deirn:
Changed dynamic reg tags loading point (again) to right after datapacks entries loaded. Biome modifications with tag selector should work™ when #1571 is merged.

liach (approved):

LambdAurora (approved):

Juuxel (approved):

YanisBft (approved):

Player3324 (approved):
LGTM, thanks! The `of` naming for the TagFactory factory methods is not quite right since it is no collection/container, but the only other choice I can think of is `create`, which would conflict with the implementation method.

modmuss50 (approved):

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/api/tag/TagFactory.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * A factory for accessing datapack tags.
+ */
+public interface TagFactory<T> {
+	TagFactory<Item> ITEM = of(ItemTags::getTagGroup);
+	TagFactory<Block> BLOCK = of(BlockTags::getTagGroup);
+	TagFactory<Fluid> FLUID = of(FluidTags::getTagGroup);
+	TagFactory<GameEvent> GAME_EVENT = of(GameEventTags::getTagGroup);
+	TagFactory<EntityType<?>> ENTITY_TYPE = of(EntityTypeTags::getTagGroup);
+	TagFactory<Biome> BIOME = of(Registry.BIOME_KEY, "tags/biomes");
```
haykam821, line 49: Why biomes specifically?

LambdAurora: It was my suggestion as biomes tags can be very useful and should be encouraged to be used in the appropriate situations.

haykam821: It should only be created as needed. That way, no registries other than vanilla ones get special treatment.

LambdAurora: One of the big issue with those non-built-in-tags factories are the path is basically up to the modder, and you cannot have multiple paths.
Hopefully all modders will use the standardized paths, but you can't really guarantee it.
I expect the biomes tags to be widely used in the future so I don't see why not providing it as "built-in". Plus it would encourage the use of those.

Juuxel: Biome tags specifically are also something I've seen requested as a feature and asked about a lot on Discord, so it's good to have them in Fabric API.

haykam821: Surely a modder can use `of(Registry.BIOME_KEY, "tags/biomes")` themselves if they need biome tags. It could even have a simplified overload of just `of(Registry.BIOME_KEY)` to prevent mismatches in the path.

LambdAurora: Looked again at Registry keys, and turns out that not exposing the path in API and creating it from the registry key would be much more safer and avoid modders having different registry base tag paths.

Juuxel: That does sort of run into problems if people have tags for modded registries - "should a non-minecraft namespace be somehow included to prevent conflicts?" etc

LambdAurora: Base tag paths should have those rules:
 - starts with `tags/`
 - if the `registry` field of the `RegistryKey` is not `Registry.ROOT` then append it, if the namespace is not minecraft then append it too as `namespace/path`.
 - append the value field, if the namespace is not minecraft then append it too as `namespace/path`
 - ends with `s`

haykam821: I don't think the `s` requirement will be possible due to exceptions. For example, the `fabric:cactus` registry would have a path of `tags/fabric/cactuss`.

frqnny (approved):

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}
+
+		return of(tagList::getGroup);
+	}
+
+	/**
+	 * Manually load tags for dynamic registries and add the resulting tag group to the tag list.
+	 *
+	 * <p>Minecraft loads the resource manager before dynamic registries, making tags for them failed to load
```
haykam821:
```suggestion
	 * <p>Minecraft loads the resource manager before dynamic registries, making tags for them fail to load
```

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
```diff
@@ -0,0 +1,119 @@
...
+	 * if it mentions datapack entries. The solution is to manually load tags after the registry is loaded.
+	 *
+	 * <p>Look at server's {@link Main#main} function calls for {@link ServerResourceManager#reload} and
+	 * {@link RegistryOps#method_36574} for the relevant code.
```
haykam821, line 88: Use `@see`

haykam821 (changes requested):
I don't think it's a good idea to provide a biome tag factory by default.

fabric-tag-extensions-v0/src/main/java/net/fabricmc/fabric/impl/tag/extension/TagFactoryImpl.java
deirn: IMO the current paragraph is more clear than using `@see`, `Main#main` is a huge function and we want it to be clear what to look for.
image

kanpov (approved):

PR #1564, opened 2021-07-13
Stop DefaultResourcePack scanning the entire classpath

### LINKED ISSUE #66: Datapack tags always replace, even when they should append
**API version:** 0.1.4.71

**Description:**
My mod adds new walls. So I have to add these new walls to the vanilla block tag _minecraft:walls_.
With a "false" value in the "replace" field, my walls should append to the vanilla ones. But instead, they replace them.

Technici4n:
This is a major performance optimization.

Without this PR, on AOF4, the model loading worker thread takes ~17s.
The sampler results include such stack traces:
image

With this PR, the worker thread takes ~11.5s.

Should also fix #66 for good.


fabric-resource-loader-v0/src/main/resources/fabric-resource-loader-v0.mixins.json
```diff
@@ -17,6 +17,7 @@
   "client": [
     "client.ClientBuiltinResourcePackProviderMixin",
     "client.CreateWorldScreenMixin",
+    "client.DefaultClientResourcePackMixin",
```
haykam821, line 20: Was this mixin not used before?

LambdAurora (approved):
imo, I would have used NIO to be consistent with the mod resource loading, but this is good enough.

ghost:
Fyi this doesn't work on Linux however Technici4n is already aware.

LambdAurora:
> Fyi this doesn't work on Linux however Technici4n is already aware.

Uh, how so? That sounds weird to me.

ghost:
Essentially it's missing the `/` before the path begins so `/home/user/.local/share/multimc/libraries/com/mojang/minecraft/1.17/minecraft-1.17-client.jar` is instead `home/user/.local/share/multimc/libraries/com/mojang/minecraft/1.17/minecraft-1.17-client.jar` which is invalid.

frqnny (approved):
thanks for documenting it, but it'd be nice to have javadocs explaining every overwrite but oh well just merge i like saving under one second

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/client/DefaultClientResourcePackMixin.java
```diff
@@ -0,0 +1,61 @@
...
+ * Override the overwritten open() method to correctly load the client's extra assets.
+ */
+@Mixin(DefaultClientResourcePack.class)
+public abstract class DefaultClientResourcePackMixin extends DefaultResourcePackMixin {
+	@Final
+	@Shadow
+	private ResourceIndex index;
+
+	@Override
+	public InputStream open(ResourceType type, Identifier id) throws IOException {
```
liach: Instead of overriding `open` here, can you just patch `findInputStream` and `getInputStream` in `DefaultResourcePack` instead? Those are protected methods which mods and subclasses can use, and they should be patched away from `Class.getResource` `ClassLoader.getResources` calls.

Technici4n: Indeed, much nicer now!

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/DefaultResourcePackMixin.java
```diff
@@ -19,55 +19,110 @@
...
+		} else {
+			// Client pack
+			type = ResourceType.CLIENT_RESOURCES;
+		}
+
+		// Locate MC jar by finding the URL that contains the assets root.
 		try {
-			Enumeration<URL> candidates = DefaultResourcePack.class.getClassLoader().getResources(path);
+			URL assetsRootUrl = DefaultResourcePack.class.getResource("/" + type.getDirectory() + "/.mcassetsroot");
+			URLConnection connection = assetsRootUrl.openConnection();
```
liach, line 68: Need to close the connection after you exit this method to prevent resource leak

Also override/overwrite `close` to close this backing resource pack as well.

liach (changes requested):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/DefaultResourcePackMixin.java
Technici4n: No, it's not closeable.
> Invoking the close() methods on the InputStream or OutputStream of an URLConnection after a request may free network resources associated with this instance, unless particular protocol specifications specify different behaviours for it.

Technici4n: > Also override/overwrite `close` to close this backing resource pack as well.

Well: :smile:

>
```java
 	/**
	 * @author FabricMC
	 * @reason Close the resource pack we redirect resource access to.
	 */
	@Overwrite
	public void close() {
		fabric_mcJarPack.close();
	}
```

liach:
Looks pretty good beyond this resource leak question

Technici4n:
There is no leak! See my comment. 😛

liach (approved):

Leo40Git (approved):

Platymemo (approved):

Player3324 (approved):
LGTM, thanks! The MC Jar origin may be available from loader some day..

Juuxel (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/DefaultResourcePackMixin.java
```diff
@@ -19,55 +19,110 @@
...
-	protected void onFindInputStream(ResourceType resourceType, Identifier identifier, CallbackInfoReturnable<InputStream> callback) {
-		if (DefaultResourcePack.resourcePath != null) {
-			// Fall through to Vanilla logic, they have a special case here.
-			return;
-		}
+public abstract class DefaultResourcePackMixin {
+	/**
+	 * Redirect all resource access to the MC jar zip pack.
+	 */
+	final ResourcePack fabric_mcJarPack = createJarZipPack();
```
modmuss50, line 51: Minor thing, no need to change now. But you can use @Unique on the field without the fabric_ prefix.

modmuss50 (approved):

PR #1571, opened 2021-07-15
Move biome modifications after datapacks loaded

deirn:
This PR restores 1.16 behavior in which biomes would only be modified after datapacks loaded.
Now the behavior will actually match the javadocs.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/modification/RegistryOpsMixin.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/modification/RegistryOpsMixin.java
@@ -37,20 +37,20 @@ import net.fabricmc.fabric.impl.biome.modification.BiomeModificationImpl;
  *     <li>{@link DynamicRegistryManager#create()} is used to create a dynamic registry manager with just
  *     entries from {@link net.minecraft.util.registry.BuiltinRegistries}</li>
  *     <li>Sometimes, Vanilla Minecraft will stop here, and use the {@link DynamicRegistryManager} as-is (examples: server.properties parsing, world creation screen).</li>
- *     <li>{@link RegistryOps#of(DynamicOps, ResourceManager, DynamicRegistryManager)} gets called with the manager, and a
+ *     <li>{@link RegistryOps#method_36574(DynamicOps, ResourceManager, DynamicRegistryManager)} gets called with the manager, and a
  *     resource manager that contains the loaded data packs. This will pull in all worldgen objects from datapacks into the
  *     dynamic registry manager.</li>
  *     <li>After the worldgen objects are pulled in from the datapacks, this mixin will call the biome modification callback.</li>
  *     <li>In most cases, Vanilla will stop here and now use the dynamic registy manager to instantiate a server.</li>
  *     <li>Sometimes, i.e. when using the "re-create world feature", and a datapack throws an error, Vanilla will sometimes
- *     repeat the {@link RegistryOps#of(DynamicOps, ResourceManager, DynamicRegistryManager)} call on the same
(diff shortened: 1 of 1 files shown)


Technici4n:
Haven't had a deep look yet, however this needs testing with:
- Code biomes,
- Datapack biomes.

If both work it should be fine.

frqnny:
I need to look at this since I’m 50% sure this is not doing anything.

frqnny:
image
I'm pretty sure this is not doing anything. ...74 calls ...75, and calling 75's return is basically the same as 74's return. 

I don't see what this does.

deirn:
`method_36575` also get called on `DynamicRegistryManager#create` far before the resource manager is even being created while`method_36574` only called after the resource manager loaded.
We need to hooks to `method_36574` to make biome tags selector a thing.

There's also some redundancy for the current behavior in which if we replace built-in biome with datapack biome it'll get modified two times. And I don't think we need to modify biomes on Create New World screen, right?

frqnny:
Hold on, the problen was that biomes were not being modified at the right time. Is this resolved? If so, I’ll have to check tommorow over the differences in what each calls, to see if this change would do what you just described.

warjort:
method_36574 is the correct place in my opinion.
36575 is also used to create the "built in" skeleton of the dynamic registries.
i.e. the dynamic registries before datapacks are applied.

This means biome modifications will get applied to non datapack biomes before the datapacks are loaded.
Then later these biomes cannot be remodified after the datapacks are loaded.

See the following stacktraces from doing "Create New World", clicking "Data Packs" and then clicking "Done"

First the builtin biomes have their modifications applied (DynamicRegistryManager::create),
this is doing main registries -> dynamic registries:
````
at net.fabricmc.fabric.impl.biome.modification.BiomeModificationImpl.modifyBiomes(BiomeModificationImpl.java:155)
at net.minecraft.util.dynamic.RegistryOps.handler$zdn000$afterCreation(RegistryOps.java:554)
at net.minecraft.util.dynamic.RegistryOps.method_36575(RegistryOps.java:59)
at net.minecraft.util.registry.DynamicRegistryManager.create(DynamicRegistryManager.java:120)
at net.minecraft.client.gui.screen.world.MoreOptionsDialog.loadDatapacks(MoreOptionsDialog.java:326)
````

Then the datapacks are loaded (DynamicRegistryManager::load),
this is doing datapacks -> dynamic registries:
````
at net.minecraft.resource.ReloadableResourceManagerImpl.containsResource(ReloadableResourceManagerImpl.java)
at net.minecraft.util.dynamic.RegistryOps$EntryLoader$1.load(RegistryOps.java:230)
at net.minecraft.util.dynamic.RegistryOps.readSupplier(RegistryOps.java:162)
at net.minecraft.util.dynamic.RegistryOps.method_29762(RegistryOps.java:130)
at com.mojang.serialization.DataResult.lambda$flatMap$10(DataResult.java:138)
at com.mojang.datafixers.util.Either$Left.map(Either.java:38)
at com.mojang.serialization.DataResult.flatMap(DataResult.java:136)
at net.minecraft.util.dynamic.RegistryOps.loadToRegistry(RegistryOps.java:130)
at net.minecraft.util.registry.DynamicRegistryManager.load(DynamicRegistryManager.java:191)
at net.minecraft.util.registry.DynamicRegistryManager.load(DynamicRegistryManager.java:178)
at net.minecraft.util.dynamic.RegistryOps.method_36575(RegistryOps.java:58)
at net.minecraft.util.dynamic.RegistryOps.method_36574(RegistryOps.java:53)
at net.minecraft.client.gui.screen.world.MoreOptionsDialog.loadDatapacks(MoreOptionsDialog.java:328)
````

Changing the mixin to be in 36574 will mean all the biomes are done together after that second entry point.

frqnny (approved):
All right, this change would make sense and hopefully help not running it multiple times.

modmuss50 (approved):
Yes, I think this is correct. Pending testing this is good to merge.

liach (approved):

PR #1581, opened 2021-07-18
Replace chunk cache HashSet with HashMap

### LINKED ISSUE #1580: fabric-lifecycle-events-v1 is slowly leaking memory
https://github.com/FabricMC/fabric/blob/3fe3c3f50c8112272ddd3865ebfa6f2625ac7747/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java#L46

This `HashSet` is storing many unloaded chunks, causing it to run into `OutOfMemoryError` and crash the server with a high playercount and loading many chunks. 
I can provide heap dumps if necessary.

ishland:
Fixes https://github.com/FabricMC/fabric/issues/1580

It seems that Minecraft is using a different WorldChunk instance every time. So this uses a HashMap to workaround that issue. 

No memory leak from there after generating ~1493k chunks with [chunky the pregenerator](https://github.com/pop4959/Chunky). (Although I spotted another memory leak from vanilla)

Pushing this onto a production server with a high playercount will make this problem much easier to reproduce.


fabric-lifecycle-events-v1/src/main/resources/fabric-lifecycle-events-v1.mixins.json
```diff
@@ -3,6 +3,7 @@
   "package": "net.fabricmc.fabric.mixin.event.lifecycle",
   "compatibilityLevel": "JAVA_16",
   "mixins": [
+    "IWorldChunk",
```
Technici4n: This mixin is useless.

Technici4n (changes requested):
I am not sure this changes anything... I can't reproduce the issue in singleplayer at least.

Technici4n:
Thanks for tackling this. I'm not really sure there is an issue to being with, or maybe I should test on a dedicated server.

ishland:
Closed as https://github.com/FabricMC/fabric/issues/1580 isn't valid.

PR #1606, opened 2021-08-02
Optimized assets.

K0-RR:
Lossless compression only - 'oxipng -r -s -o max'


Technici4n:
I fail to see a major improvement here. I would refrain from PR'ing such things.

K0-RR:
Does every improvement need to be major?

Issue #1532, opened 2021-06-30
ArmorRenderingRegistry missing in 1.17


quiqueck:
The `ArmorRenderingRegistry` is missing in **fabric-rendering-v1**.

shedaniel:
Correct, it was merged to 1.16.5 but not 1.17 of Fabric API, drafting a PR...

AzureDoom:
> Correct, it was merged to 1.16.5 but not 1.17 of Fabric API, drafting a PR...

so about that PR? Geckolib needs it for doing it's Armor.

Aelpecyem:
Any progress on this? I was considering drafting a PR by myself otherwise

K0-RR:
I think this can be closed now.

Technici4n:
Implemented by #1607.

PR #1607, opened 2021-08-02
Re-implement ArmorRenderingRegistry

Aelpecyem:
This PR reimplements ArmorRenderingRegistry, fixing #1532. This approach is more minimalist and consistent, mostly matching BuiltinItemRendererRegistry.


ghost:
Fyi, missing license headers, there should be a licenser task you can run to fix this.
If I'm right it is `gradlew :licenseFormatAll`

Technici4n:
Would this be a good chance to get rid of the registry, and move to a static `register` function in `ArmorRenderer`?

Aelpecyem:
I figured I'd keep it consistent with how the BuiltinItemRenderer is registered, but sure

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Armor renderers render worn armor items with custom code.
+ * They may be used to render armor with special models or effects.
+ *
+ * <p>The renderers are registered with {@link ArmorRenderingRegistry#register(ArmorRenderer, Item...)}.
+ */
+@FunctionalInterface
+@Environment(EnvType.CLIENT)
+public interface ArmorRenderer {
+	static final HashMap<Item, ArmorRenderer> RENDERERS = new HashMap<>();
```
Technici4n: This should probably go in an impl class, exposing the map like that makes it part of the public API. My remark was mostly about having a single API class instead of 2!

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * Standard getter method for armor renderers
+	 * @param item	the item, associated with the renderer
+	 * @return the renderer, returns null if no renderer is registered
+	 */
+	@Nullable
+	static ArmorRenderer getRenderer(Item item){
+		return RENDERERS.get(item);
+	}
```
Technici4n: If the map is moved to an impl class, I'm not sure this will still be needed (this function can be moved to the impl class, and called directly by the mixin).

Technici4n (commented):
Minor things. Make sure you also run `gradlew check` to ensure CI passes when the workflow is approved.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
Aelpecyem: > Minor things. Make sure you also run `gradlew check` to ensure CI passes when the workflow is approved.

Is there a way to quickly reformat files according to the guidelines or do I have to do that manually?

Technici4n: Unfortunately, that has to be done manually.

Technici4n:
Hmmm I preferred the `ArmorRenderer` with a static `register` function, as we are moving away from the `INSTANCE` pattern. Sorry for the bad communication I guess? 😅

What we usually do is something like:
```java
// public API class
public interface ArmorRenderer {
    static void register(ArmorRenderer renderer, Item... items) {
        ArmorRenderingImpl.register(renderer, items);
    }

    // regular interface methods
}

// impl class
public class ArmorRenderingImpl { // Feel free to pick a better name
    private static final Map<> RENDERERS = ...;

    public static void register(ArmorRenderer renderer, Item... items) {
        Objects.requireNonNull(renderer, "renderer is null");

        if (items.length == 0) {
            // catch people who forget to register an item.
            throw new IllegalArgumentException();
        }

        for (Item item : items) {
            Objects.requireNonNull(item, "armor item is null");

            if (RENDERERS.putIfAbsent(item, renderer) != null) {
                throw new IllegalArgumentException("Custom armor renderer already exists for " + Registry.ITEM.getId(item));
            }
        }
    }

    // directly referenced by the mixin (not part of the API)
    @Nullable
    public static ArmorRenderer get(Item item) {
        return RENDERERS.get(item);
    }
}
```

Aelpecyem:
I see, one sec then

Aelpecyem:
What would be a better name for the impl class?

Technici4n:
Up to you, it's your PR. 😆 Personally I think `ArmorRenderingImpl` is fine.

Technici4n (approved):
Looks fine, and the testmod works.

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -0,0 +1,86 @@
...
+@FunctionalInterface
+public interface ArmorRenderer {
+	/**
+	 * Registers the armor renderer for the specified items.
+	 * @param renderer	the renderer
+	 * @param items		the items
+	 * @throws IllegalArgumentException if an item already has a registered armor renderer
+	 * @throws NullPointerException if either an item or the renderer is null
+	 */
+	static void register(ArmorRenderer renderer, Item... items) {
```
Juuxel: Maybe use `ItemConvertible` instead of `Item` here? Though I suspect blocks as armor pieces are pretty rare.

fabric-rendering-v1/src/main/resources/fabric-rendering-v1.mixins.json
```diff
@@ -11,5 +11,8 @@
   ],
   "injectors": {
     "defaultRequire": 1
-  }
+  },
+  "mixins": [
+    "MixinArmorFeatureRenderer"
```
Juuxel: This should be in the `client` block as well.

Juuxel (changes requested):

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -0,0 +1,86 @@
...
+	 * @param matrices			the matrix stack
+	 * @param vertexConsumers	the vertex consumer provider
+	 * @param light				packed lightmap coordinates
+	 * @param stack				the item stack of the armor item
+	 * @param model				the model to be rendered
+	 * @param texture			the texture to be applied
+	 */
+	static void renderPart(MatrixStack matrices, VertexConsumerProvider vertexConsumers, int light, ItemStack stack, Model model, Identifier texture) {
+		VertexConsumer vertexConsumer = ItemRenderer.getArmorGlintConsumer(vertexConsumers, RenderLayer.getArmorCutoutNoCull(texture), false, stack.hasGlint());
+		model.render(matrices, vertexConsumer, light, OverlayTexture.DEFAULT_UV, 1.0F, 1.0F, 1.0F, 1.0F);
```
Player3324: You can just pass `1, 1, 1, 1` here

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/ArmorRenderingTests.java
```diff
@@ -0,0 +1,50 @@
...
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.rendering.v1.ArmorRenderer;
+
+public class ArmorRenderingTests implements ClientModInitializer {
+	private BipedEntityModel<LivingEntity> armorModel;
+	private Identifier texture = new Identifier("textures/block/dirt.png");
+
+	@Override
+	public void onInitializeClient() {
+		ArmorRenderer.register((matrices, vertexConsumers, stack, entity, slot, light, model) -> {
```
Player3324, line 37: Please add a few words what this does/what it's supposed to change/look like so someone porting the code can quickly verify correct behavior without guessing from the code.

Player3324 (approved):
LGTM, thanks! 2 minor remarks inline.

Aelpecyem:
ok, this should do it

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/ArmorRendererRegistryImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+	public static void register(ArmorRenderer renderer, ItemConvertible... items) {
+		Objects.requireNonNull(renderer, "renderer is null");
+
+		if (items.length == 0) {
+			throw new IllegalArgumentException("Armor renderer registered for no item");
+		}
+
+		for (ItemConvertible item : items) {
+			Objects.requireNonNull(item, "armor item is null");
```
Technici4n:
```suggestion
			Objects.requireNonNull(item.asItem(), "armor item is null");
```

Juuxel: Ideally both of them, but `item.asItem()` will conveniently NPE if the convertible is null.

liach: If an implementation is bad enough that `asItem` returns `null`, I would expect it to be bad enough that it returns different `Item` instances on different `asItem` calls... Should we keep the retrieved item in a local var hence

Technici4n: Don't think so, this is just to prevent a wrong IDE method fill or similar.

liach (approved):

Aelpecyem:
How long does it usually take for PRs to get approved? A few mods I work on rely on it.

liach:
You can ping modmuss here or on discord for a review on this

modmuss50 (approved):

LambdAurora (approved):

modmuss50:
Will merge in a few hours, I dont think this needs a full week of last call.

PR #1617, opened 2021-08-05
ItemGroup API: Remove return and call super so all others ItemStack's get appended

AlphaMode:


Player3324:
The existing behavior is deliberate, using `appendItems` moves population responsibility to the item group itself instead of the items.

liach (approved):

Technici4n:
What is the use case for this overload?

AlphaMode:
> What is the use case for this overload?

If you wanted to have a normal item group then append items at the end you would not be able to  with this it passes the item group so you can add back the default items

fabric-item-groups-v0/src/main/java/net/fabricmc/fabric/api/client/itemgroup/FabricItemGroupBuilder.java
```diff
@@ -76,6 +77,16 @@ public FabricItemGroupBuilder stacksForDisplay(Consumer<List<ItemStack>> appende
...
 	public FabricItemGroupBuilder appendItems(Consumer<List<ItemStack>> stacksForDisplay) {
+		return appendItems(((itemStacks, itemGroup) -> stacksForDisplay.accept(itemStacks)));
+	}
+
+	/**
+	 * This allows for a custom list of items to be displayed in a tab, this enabled tabs to be created with a custom set of items.
+	 *
+	 * @param stacksForDisplay Add ItemStack's to this list to show in the ItemGroup
+	 * @return a reference to the FabricItemGroupBuilder
+	 */
```
Technici4n: Please explain how this method is different from the other overload (e.g. "one use case is to use the passed ItemGroup to query \<vanilla function\> to add back default items in the tab").

fabric-item-groups-v0/src/main/java/net/fabricmc/fabric/api/client/itemgroup/FabricItemGroupBuilder.java
```diff
@@ -76,6 +77,16 @@ public FabricItemGroupBuilder stacksForDisplay(Consumer<List<ItemStack>> appende
 	 * @return a reference to the FabricItemGroupBuilder
 	 */
 	public FabricItemGroupBuilder appendItems(Consumer<List<ItemStack>> stacksForDisplay) {
```
Technici4n, line 81: Would you mind stating that this bypasses items specifying a creative tab directly, and then explain how you can get this behavior by using the other function?

Technici4n (commented):
I think the change is fine then. The javadoc needs some work though.

fabric-item-groups-v0/src/main/java/net/fabricmc/fabric/api/client/itemgroup/FabricItemGroupBuilder.java
```diff
@@ -70,12 +72,25 @@ public FabricItemGroupBuilder stacksForDisplay(Consumer<List<ItemStack>> appende
...
 	/**
-	 * This allows for a custom list of items to be displayed in a tab, this enabled tabs to be created with a custom set of items.
+	 * Set the item stacks of this item group, by having the consumer add them to the passed list.
+	 * This bypasses {@link Item#appendStacks}. If you want to append stacks from your items, consider using {@linkplain #appendItems(BiConsumer) the other overload}.
 	 *
 	 * @param stacksForDisplay Add ItemStack's to this list to show in the ItemGroup
 	 * @return a reference to the FabricItemGroupBuilder
 	 */
 	public FabricItemGroupBuilder appendItems(Consumer<List<ItemStack>> stacksForDisplay) {
+		return appendItems(((itemStacks, itemGroup) -> stacksForDisplay.accept(itemStacks)));
```
liach:
```suggestion
		return appendItems((itemStacks, itemGroup) -> stacksForDisplay.accept(itemStacks));
```

liach (approved):

fabric-item-groups-v0/src/testmod/java/net/fabricmc/fabric/test/item/group/ItemGroupTest.java
```diff
@@ -35,7 +39,20 @@
...
 				).build();
 
+	private static final ItemGroup ITEM_GROUP_2 = FabricItemGroupBuilder.create(new Identifier("fabric-item-groups-v0-testmod", "test_group_two"))
+			.icon(() -> new ItemStack(Items.REDSTONE))
+			.appendItems((stacks, itemGroup) -> {
+					for(Item item : Registry.ITEM) {
+						if(item.isIn(ItemGroup.FOOD))
+							stacks.add(new ItemStack(item));
+						if(item.isIn(itemGroup))
+							stacks.add(new ItemStack(item));
```
Technici4n:
```suggestion
						if (item.getGroup() == ItemGroup.FOOD || item.getGroup() == itemGroup) {
							stacks.add(new ItemStack(item));
						}
```

Technici4n (approved):
Works as expected, but needs one little change to run without AW.

Technici4n:
FYI you can run `gradlew build` to test that everything builds as expected! This will prevent countless workflow attempts.

Technici4n:
Still waiting for you to fix the build.

Technici4n (approved):

fabric-item-groups-v0/build.gradle
```diff
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-item-groups-v0"
-version = getSubprojectVersion(project, "0.2.10")
+version = getSubprojectVersion(project, "0.2.11")
```
Juuxel: Nitpick: should be a minor version bump to `0.3.0` since it adds a feature.

Juuxel (approved):

Player3324 (approved):
LGTM, thanks!

It'd be nice to use a dedicated interface instead of the BiConsumer since that has sub-par naming/ergonomics and can't be evolved as easily. That's more of a bonus though.

modmuss50 (approved):

PR #1629, opened 2021-08-11
Add tag-based mining level API, fix and deprecate FabricBlockSettings.breakByTool

### LINKED ISSUE #1533: Fabric tool tags don't affect `#minecraft:mineable/*`
Mojang introduced a non-hardcoded mining tool tag system (similar to `FabricBlockSettings.breakByTool`) in 1.17 with its own set of "mineable" tags for different tool types. These don't take into account the Fabric tool tags at all, unless the item itself is a `PickaxeItem` (or any other corresponding `MiningToolItem` - shovels, hoes or axes).

### LINKED ISSUE #1534: [Suggestion] Replace `FabricBlockSettings.breakByTool` by extending the 1.17 vanilla tag system
In 1.17, Mojang added a system for defining the breaking tools and required tool levels with a set of builtin tags. (Previously, this was done with ugly if-statements in tool classes.) This makes the Fabric `breakByTool` system redundant for vanilla tools and vanilla mining levels.

- Custom tool types can be done by passing a custom `mineable` tag for your tool, and passing it into `MiningToolItem`'s constructor. There's no API needed here.
  - Multitools that have multiple tool types can just have a custom `Item.isSuitableFor(BlockState)` implementation, which is reasonable enough to implement manually IMO.
    - Example: `state.isIn(#mineable/pickaxe) || state.isIn(#mineable/shovel)` for a pickaxe-shovel multitool
- Custom mining levels are a bit harder. The tags themselves are hardcoded in `MiningToolItem.isSuitableFor`, so a simple registry of "mining level (int) -> mining level block tag" could be made here.

Discord discussion:
Discord discussion

Juuxel:
Closes #1534. Fixes #1533.

- Deprecated `FabricBlockSettings.breakByTool` in favour of the two following features:
- (1) Added `#fabric:mineable/sword` and `#fabric:mineable/shears` to mirror `#mineable/[pickaxe|axe|shovel|hoe]`
- (2) Added dynamic mining level tags (like vanilla's `#needs_diamond_tool`)
  - The format is `fabric:needs_tool_level_N` where *N* is the wanted mining level as an int (as described by [Yarn's docs](https://github.com/FabricMC/yarn/blob/1.17.1/src/constants/java/net/fabricmc/yarn/constants/MiningLevels.java))
  - They can be used alongside vanilla tags, and when the wanted mining level is higher than what vanilla has tags for (eg. netherite)
- Adds a new, much simpler mining level API containing a simple block state -> required mining level int getter. This does not catch mining levels registered with the deprecated `breakByTool` for modularity purposes!
- Fixes #1533
- Fixes effective checks failing for tagged tool items (that are in `#fabric:your_favourite_tool_tag` but aren't instances of `DynamicToolItem` or `ToolItem`)

This PR does not fix the tool attribute API's tests not actually testing the dynamic tool functionality, which is out of scope.


fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/mixin/mininglevel/ShearsItemMixin.java
```diff
@@ -0,0 +1,56 @@
...
+	private void onGetMiningSpeedMultiplier(ItemStack stack, BlockState state, CallbackInfoReturnable<Float> info) {
+		if (info.getReturnValueF() == 1.0f) { // if not caught by vanilla checks
+			if (state.isIn(FabricMineableTags.SHEARS_MINEABLE)) { // mimics MiningToolItem.getMiningSpeedMultiplier
+				// In vanilla 1.17, shears have three special mining speed multiplier values:
+				//   - cobweb and leaves return 15.0
+				//   - wool returns 5.0
+				//   - vines and glow lichen return 2.0
+				// As the most "neutral" option out of these three,
+				// we'll use 5.0 as it's not extremely fast nor extremely slow.
+				info.setReturnValue(5.0f);
```
liach, line 52: Can we just split the 3 mining speeds into 3 tags? And we specify which tags have priority?

Juuxel: Hmm, that's an interesting idea. I could imagine an additional two tags (`mineable/shears/weak` and `mineable/shears/strong`) that would be used alongside this, but I'd like another opinion before changing this - a single value is still simpler.

Technici4n: There should be a way for blocks to implement this switch themselves imo.

Juuxel: So how should the blocks implement it? I think there are three reasonable options:

1. Tags (my preferable option)
2. A callback (feels like bloat tbh)
3. An interface to implement on blocks (takes out the data-driven aspect of mining levels)

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/impl/mininglevel/MiningLevelManagerImpl.java
```diff
@@ -0,0 +1,89 @@
...
+import net.fabricmc.yarn.constants.MiningLevels;
+
+public final class MiningLevelManagerImpl {
+	private static final Logger LOGGER = LogManager.getLogger("fabric-mining-level-api-v1/MiningLevelManagerImpl");
+	private static final String TOOL_TAG_NAMESPACE = "fabric";
+	private static final Pattern TOOL_TAG_PATTERN = Pattern.compile("^needs_tool_level_([0-9]+)$");
+
+	// A cache of block state mining levels. Cleared by
+	// - MiningLevelCacheInvalidator when tags are reloaded
+	// - ClientPlayNetworkHandlerMixin when tags are synced
```
liach, line 42: Hmm, I feel we should get an event for such syncs, but that is a separate issue from this pr

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/impl/mininglevel/MiningLevelManagerImpl.java
```diff
@@ -0,0 +1,89 @@
...
+					continue;
+				}
+
+				Matcher matcher = TOOL_TAG_PATTERN.matcher(tagId.getPath());
+
+				if (matcher.matches()) {
+					try {
+						int tagMiningLevel = Integer.parseInt(matcher.group(1));
+
+						if (tagMiningLevel < 0) {
```
liach: Don't think a negative number can match this pattern.

liach (approved):

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/impl/mininglevel/MiningLevelManagerImpl.java
Juuxel: Oops, that's true - I had more generic code initially so this is a leftover.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/mixin/mininglevel/ShearsItemMixin.java
Technici4n: Does it make sense for a block's mining speed multiplier to depend on the tool that is used? I feel like blocks should have control over this maybe?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
```diff
@@ -315,7 +315,10 @@ public FabricBlockSettings breakByHand(boolean breakByHand) {
 	/**
 	 * Please make the block require a tool if you plan to disable drops and slow the breaking down using the
 	 * incorrect tool by using {@link FabricBlockSettings#requiresTool()}.
+	 *
+	 * @deprecated Replaced by {@code mineable} tags. See fabric-mining-level-api-v1 for further details.S
 	 */
+	@Deprecated(forRemoval = true)
```
Technici4n, line 321: Technically shouldn't be `forRemoval` I think, unless it's broken beyond salvation?

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/mixin/mininglevel/ShearsItemMixin.java
Technici4n: Perhaps that would be a `FabricBlockSettings`-added setting? Think `BiFunction<BlockState, ItemStack, Double> dynamicMiningSpeedMultiplier` or something like that.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
Juuxel: This PR actually fixes it (it's broken in 1.17) so I'll take out the forRemoval.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/mixin/mininglevel/ShearsItemMixin.java
Juuxel: > Does it make sense for a block's mining speed multiplier to depend on the tool that is used? I feel like blocks should have control over this maybe?

The tool attribute API supports this from the tool side. I can add the block stuff too, but anything touching mining levels also has to go through tool attributes to support dynamic tools - basically adding everything twice.

Technici4n: Imo mining levels and breaking speed are different concerns. A suitable mining level would just delegate to the block settings to grab the speed (in fact that's how tools work). Unfortunately mojang had to do it differently for swords and shears.

Technici4n: Maybe we could go for a simple
```java
@Nullable
Double getShearsMiningSpeedMultiplier(BlockState state);
```
auto-invoking event? And the same for swords?

Technici4n: It sucks that it's hard-coded, so maybe we should have this event have a tool tag parameter? Not sure how useful it'd be, I think having different breaking speeds for different kinds of tools would be useful. Auto-invoking events allow both easy implementation on the Block class, and more flexible implementations with regular event handlers.

Juuxel: Mining levels and breaking speeds are very coupled in the current Fabric API structure since the (broken) object builders method of setting required mining levels is tied to the tool attribute module, which in turn is tied to breaking speeds.

I'm a bit wary of adding mining speed events to this PR since it feels out of scope. That event would also have to be *in* the mining level API (since that's what implements the shear mining levels in this PR), which disagrees with the idea that "mining levels and breaking speed are different concerns".

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
liach: if this fix is hacky, you can still consider just leaving it noop and remove it later with the annotation

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
```diff
@@ -315,7 +315,10 @@ public FabricBlockSettings breakByHand(boolean breakByHand) {
 	/**
 	 * Please make the block require a tool if you plan to disable drops and slow the breaking down using the
 	 * incorrect tool by using {@link FabricBlockSettings#requiresTool()}.
+	 *
+	 * @deprecated Replaced by {@code mineable} tags. See fabric-mining-level-api-v1 for further details.S
```
Juuxel:
```suggestion
	 * @deprecated Replaced by {@code mineable} tags. See fabric-mining-level-api-v1 for further details.
```

ghost (approved):

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/impl/mininglevel/MiningLevelManagerImpl.java
Juuxel: Resolved.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/mixin/mininglevel/ShearsItemMixin.java
Technici4n: Ok, let's just ignore this for now.

Technici4n (approved):
Sorry for taking so long to review this, it looks absolutely amazing.

Player3324 (approved):
LGTM, thanks!

modmuss50 (approved):
Awesome 👏

ramidzkh (approved):

marcus8448 (approved):

PR #1632, opened 2021-08-13
Add flattening, stripping and tilling registries

Juuxel:
Spiritual successor to a small part of #1125 (which I feel should still be resurrected once again, but I'm not the one to do that).

- Adds registries for flattenable blocks (use with shovel), strippable blocks (use with axe) and tillable blocks (use with hoe).
- The registries are close to vanilla, so there is no support for tags (unlike the flammable block registry). If someone needs it, I might look into making them `Block2ObjectMap`s.
- The tillable block registry includes a few extra methods for the most common registration types (block -> block state, block -> block state + drop an item) that use vanilla methods, but also exposes the raw functionality.

### DIFF
--- a/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
+++ b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
@@ -44,7 +44,7 @@ public final class TillableBlockRegistry {
 	 * a block. If the predicate returns {@code true}, an action is executed. Default instances of these can be created
 	 * with these {@link HoeItem} methods:
 	 * <ul>
-	 * <li>usage predicate: {@link HoeItem#usagePredicate(ItemUsageContext)}</li>
+	 * <li>usage predicate for farmland-like behavior: {@link HoeItem#usagePredicate(ItemUsageContext)}</li>
 	 * <li>simple action: {@link HoeItem#getTillingConsumer(BlockState)}</li>
 	 * <li>simple action that also drops an item: {@link HoeItem#getTillingConsumer(BlockState, ItemConvertible)}</li>
 	 * </ul>
--- a/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
+++ b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
@@ -50,7 +50,7 @@ public final class TillableBlockRegistry {
 	 * </ul>
 	 *
 	 * @param input          the input block that can be tilled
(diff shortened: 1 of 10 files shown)


Technici4n (approved):
LGTM, can always be expanded later for custom behavior if necessary.

Player3324 (approved):
LGTM, thanks! As mentioned on Discord refactoring the module is postponed to a future change.

modmuss50 (approved):

warjort:
Maybe HoneycombItem.UNWAXED_TO_WAXED_BLOCKS as well?
But that doesn't look that easy to implement because of the public memoized suppliers.

LambdAurora:
> Maybe HoneycombItem.UNWAXED_TO_WAXED_BLOCKS as well?
> But that doesn't look that easy to implement because of the public memoized suppliers.

That should be an Oxidized API/registry imo.
The issue is it requires asm soup to achieve.

Juuxel:
@warjort That'd be useful but I feel like it's out of scope since this PR deals with tool interactions. (And if the logic is different, it's probably good to have a separate PR for it anyway)

liach (approved):

liach:
> The issue is it requires asm soup to achieve.

@LambdAurora Just curious, do you just mean complex bytecode modification beyond what the current mixin model offers?

Jamalam360 (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+	 * <li>simple action: {@link HoeItem#getTillingConsumer(BlockState)}</li>
+	 * <li>simple action that also drops an item: {@link HoeItem#getTillingConsumer(BlockState, ItemConvertible)}</li>
+	 * </ul>
+	 *
+	 * @param input          the input block that can be tilled
+	 * @param usagePredicate a predicate that filters if the block can be tilled
+	 * @param tillingAction  an action that is executed if the predicate returns {@code true}
+	 */
+	public static void register(Block input, Predicate<ItemUsageContext> usagePredicate, Consumer<ItemUsageContext> tillingAction) {
+		Objects.requireNonNull(input, "input bock cannot be null");
```
haykam821:
```suggestion
		Objects.requireNonNull(input, "input block cannot be null");
```

haykam821 (changes requested):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/TillableBlockRegistry.java
kanpov: **bock**

kanpov (approved):

haykam821 (approved):

PR #1633, opened 2021-08-13
Add sleeping events

Juuxel:
Added 9 entity events related to sleep:
- starting to sleep
- waking up
- checking if a block is a valid bed for an entity (useful for custom beds)
- modifying the sleeping direction (useful for custom beds)
- only for players:
  - an "allow sleep time" event
  - an "allow setting spawn" event
  - an "allow sleeping" event (as in, allow to start sleeping)
  - an "allow nearby monsters" event that can also be used to do custom monster checks (to force a failed monster check)
  - an "allow resetting time" event (whether the player counts towards the minimum for skipping the night)

### DIFF
--- a/fabric-entity-events-v1/build.gradle
+++ b/fabric-entity-events-v1/build.gradle
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-entity-events-v1"
-version = getSubprojectVersion(project, "1.1.0")
+version = getSubprojectVersion(project, "1.2.0")
 
 moduleDependencies(project, [
 		'fabric-api-base'
@@ -7,4 +7,6 @@ moduleDependencies(project, [
 
 dependencies {
 	testmodImplementation project(path: ':fabric-command-api-v1', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-networking-api-v1', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-registry-sync-v0', configuration: 'dev')
 }
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
@@ -34,8 +34,14 @@ import net.fabricmc.fabric.api.event.EventFactory;
 public final class EntitySleepEvents {
 	/**
 	 * An event that checks whether a player can start to sleep in a bed-like block.
+	 * This event only applies to sleeping using {@link PlayerEntity#trySleep(BlockPos)}.
 	 *
-	 * @see PlayerEntity#trySleep(BlockPos)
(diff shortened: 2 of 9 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	/**
+	 * An event that can be used to provide the entity's sleep direction if missing.
+	 *
+	 * <p>This is useful for custom bed blocks that need to determine the sleeping direction themselves.
+	 */
+	public static final Event<ModifySleepingDirection> MODIFY_SLEEPING_DIRECTION = EventFactory.createArrayBacked(ModifySleepingDirection.class, callbacks -> (entity, sleepingPos, sleepingDirection) -> {
+		for (ModifySleepingDirection callback : callbacks) {
+			Optional<Direction> result = callback.modifySleepDirection(entity, sleepingPos, sleepingDirection);
```
Player3324: I'd have done it like so, without early return:

```suggestion
			sleepingDirection = callback.modifySleepDirection(entity, sleepingPos, sleepingDirection);
```

The Optional use for cancellation is a little weird IMO.

Juuxel: Yeah, that's simpler. Cascading direction modifications is a bit weird, still, but I think it should be fine.

Player3324 (approved):
LGTM, thanks! JavaDoc tweaks pending as discussed externally.

fabric-entity-events-v1/build.gradle
```diff
@@ -1,10 +1,12 @@
...
+version = getSubprojectVersion(project, "1.2.0")
 
 moduleDependencies(project, [
 		'fabric-api-base'
 ])
 
 dependencies {
 	testmodImplementation project(path: ':fabric-command-api-v1', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-networking-api-v1', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-registry-sync-v0', configuration: 'dev')
```
liach, line 11: How does this impl depend on registry sync? Assuming the networking api is the prerequisite of registry sync.

Juuxel: It registers a test block and a test item, which need registry sync on a test dedi server. (And also if running with different combinations of test mods that register stuff to prevent ID mixups.)

liach: Oh right this is testmod, my bad, thought it was the api's impl

modmuss50 (approved):
Very nice 👍

Mysterious-Dev (approved):
LGTM

liach (approved):

kanpov (approved):

PR #1634, opened 2021-08-13
Fix typos

fredster33:


modmuss50:
Please do not spam PRs. We generally do not accept these kinds of PRs. Thanks for understanding

PR #1640, opened 2021-08-18
Add biome tag selector

deirn:
This PR adds a helper function into `BiomeSelectors` that accept a biome tag.
Also contains changes from #1571 for it to actually work.

<details>
<summary>A screenshot</summary>

</details>

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
@@ -26,6 +26,7 @@ import com.google.common.collect.ImmutableSet;
 
 import net.minecraft.entity.EntityType;
 import net.minecraft.entity.SpawnGroup;
+import net.minecraft.tag.Tag;
 import net.minecraft.util.registry.BuiltinRegistries;
 import net.minecraft.util.registry.RegistryKey;
 import net.minecraft.world.biome.Biome;
@@ -102,6 +103,15 @@ public final class BiomeSelectors {
 	 */
 	public static Predicate<BiomeSelectionContext> foundInTheEnd() {
 		return context -> context.getBiome().getCategory() == Biome.Category.THEEND;
+	}
+
+	/**
+	 * Returns a biome selector that will match all biomes in the given tag.
+	 *
+	 * @see net.fabricmc.fabric.api.tag.TagFactory#BIOME
+	 */
+	public static Predicate<BiomeSelectionContext> tag(Tag<Biome> tag) {
+		return context -> tag.contains(context.getBiome());
 	}
 
 	/**
--- a/fabric-biome-api-v1/build.gradle
+++ b/fabric-biome-api-v1/build.gradle
@@ -4,3 +4,9 @@ version = getSubprojectVersion(project, "3.1.11")
 minecraft {
     accessWidener = file("src/main/resources/fabric-biome-api-v1.accesswidener")
 }
+
+dependencies {
+	testmodImplementation project(path: ':fabric-api-base', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-resource-loader-v0', configuration: 'dev')
+	testmodImplementation project(path: ':fabric-tag-extensions-v0', configuration: 'dev')
+}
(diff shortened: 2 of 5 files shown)


YanisBft (approved):

LambdAurora (approved):
A dream finally coming true.

modmuss50:
The API change looks good, I think its best to handle #1571 as its own PR. Ill review that now and can try and get this in quite quickly.

modmuss50 (approved):
Approving the changes outside of what https://github.com/FabricMC/fabric/pull/1571 does.

Will try and merge the 2 at the same time, but easier to track the change in its own PR.

Juuxel (approved):

modmuss50:
Quick last call, will merge with the others in a few days 👍

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
```diff
@@ -104,6 +105,15 @@ private BiomeSelectors() {
 		return context -> context.getBiome().getCategory() == Biome.Category.THEEND;
 	}
 
+	/**
+	 * Returns a biome selector that will match all biomes in the given tag.
+	 *
+	 * @see net.fabricmc.fabric.api.tag.TagFactory#BIOME
+	 */
+	public static Predicate<BiomeSelectionContext> tag(Tag<Biome> tag) {
```
liach, line 113: should we enforce tag.identified to prevent accidentally passing tags that expire on reload?

deirn: I don't think that it's matter since biome modifications only called once on server setup.

frqnny (approved):
Amazing! I've always wanted to add this myself. I haven't checked- does Fabric add tags you can reference themselves or is it going to leave it to people to figure out? I think it would be nice to have biomes force to use tags like c:overworld or stuff like that to help modders use this new tags.

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
liach: do modification apply to loaded biomes after each reload? if yes, on setup, a persistent (identified) tag should be passed to address biome tag changes in reloads, as you don't have a chance to update the tag reference manually later. another supplier of tag may work too.

liach:
> Amazing! I've always wanted to add this myself. I haven't checked- does Fabric add tags you can reference themselves or is it going to leave it to people to figure out? I think it would be nice to have biomes force to use tags like c:overworld or stuff like that to help modders use this new tags.

that's already added im #1562

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
liach: An instance: say some the biome tag selector test accidentally just acquired a tag from the tag group for the biome tags; the tag definitely don't update across resource reloads. if a data pack that adds swamp to the tag is enabled, the predicate still will use the outdated tag, and will return `false` for a swamp biome context.

Using `Tag.Identified<Biome>` is just an extra level of safeguard for good practices, which I think its benefits outweigh its restrictions.

deirn: All modifications will only be applied once, right after the dynamic registry is loaded (server setup.) The function won't even get called on `/reload`. Even if it is somehow called multiple times, the individual biome can only be modified once anyway.

deirn:
> > Amazing! I've always wanted to add this myself. I haven't checked- does Fabric add tags you can reference themselves or is it going to leave it to people to figure out? I think it would be nice to have biomes force to use tags like c:overworld or stuff like that to help modders use this new tags.
> 
> that's already added im #1562

I think frqnny was asking about [convention tags](https://fabricmc.net/wiki/tutorial:tags#existing_common_tags) but for biomes, and no, fabric doesn't have built-in tags for it currently.

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectors.java
liach: Hmm, I wonder if this weird fact should be documented or not

liach (approved):

PR #1656, opened 2021-08-21
Fabric Resource Conditions API

### LINKED ISSUE #1644: Recipe loading conditions
We could extend the recipe json format to allow conditionally skipping the loading of some recipes.

Example use case: loading certus quartz crystal -> certus quartz dust grinding recipes when AE2 is loaded.

A possible extension could look like this:
```json5
// Standard MI recipe format
{
  "type": "modern_industrialization:macerator",
  "eu": 2,
  "duration": 200,
  "item_inputs": { "tag": "c:certus_quartz_crystals" },
  "item_outputs": { "item": "appliedenergistics2:certus_quartz_dust" },
// Additional conditions, the recipe will be skipped if the conditions fail.
  "fabric:conditions": { "mod": "appliedenergistics2" }
}
```

The challenge is picking the exact format, designing an API that allows defining custom conditions and providing a set of useful conditions (mod loaded, AND, OR, others?).

Technici4n:
Closes #1644.

Example recipe
```json
{
  "type": "minecraft:crafting_shapeless",
  "ingredients": [
    {
      "item": "minecraft:stick"
    }
  ],
  "result": {
    "item": "minecraft:diamond"
  },
  "fabric:conditions": [
    {
      "condition": "fabric:not",
      "value": {
        "values": [
          "a"
        ]
      }
    }
  ]
}
```
Example condition implementation, with the helper to generate json entries too:
```java
// ID will be used multiple times
	private static final Identifier MODS_LOADED = new Identifier("fabric:mods_loaded");
// Only API we need is the json provider
	public static ConditionJsonProvider modsLoaded(String... modIds) {
		return new ConditionJsonProvider() {
			@Override
			public void serialize(JsonObject object) {
				JsonArray array = new JsonArray();
				for (var modId : modIds) {
					array.add(modId);
				}
				object.add("values", array);
			}

			@Override
			public Identifier getConditionIdentifier() {
				return MODS_LOADED;
			}
		};
	}
// Register the condition
	static {
		ResourceConditions.register(MODS_LOADED, object -> {
			JsonArray array = JsonHelper.getArray(object, "values");

			for (JsonElement element : array) {
				if (element.isJsonPrimitive()) {
					if (!FabricLoader.getInstance().isModLoaded(element.getAsString())) {
						return false;
					}
				} else {
					throw new JsonParseException("Invalid mod id entry: " + element);
				}
			}

			return true;
		});
	}
```

Please let me know what you think of the approach, and which additional conditions should be provided.

TODO:
- [x] Recipe conditions.
- [x] Integration in the datagen API.
- [x] Validating testmod.
- [x] Document the conditions we provide by default.
- [x] Document where we check for conditions.


haykam821 (changes requested):
Conditions should be registered using identifiers instead of strings.

warjort:
> Conditions should be registered using identifiers instead of strings.

That would make the real name of the builtin conditions something like ````minecraft:not````
i.e. ````new Identifier("not")```` from ````"not": {}````

BTW: isn't ````{}```` the same as ````"{"always": ""}```` , i.e. "always" is redundant?

warjort:
> i.e. "always" is redundant?

Actually is does have a use under a different name
````
 "fabric:conditions": {
  "comment": "some memo to describe the purpose of a complicated condition"
}
````

**Comment by warjort** [comment]
I don't know how useful it would be, but you can also add some syntatic sugar by supporting things like:

````"not": false```` instead of  ````"not": {"never": ""}````

 "ex falso quodlibet" :-)

**Comment by warjort** [comment]
````
            Predicate<JsonElement> condition = JsonResourceConditions.get(entry.getKey());

            if (condition == null) {
                throw new IllegalArgumentException("Encountered unknown recipe condition " + entry.getKey());
            } else if (condition.test(entry.getValue())) {
                return true;
            }
````

Shouldn't this error handling be inside the framework? 
Having every implementation do this if they support thirdparty conditions seems error prone.

Something like:
````
boolean JsonResourceConditions::test(String id, JsonElement value) {
    // the above code
}
````

I am also thinking of possible future enhancement if additional cross cutting logic needs to applied when a condition is tested.
e.g. adding externally configured fallback conditions something like:
````
config/fabric_condition_aliases.json
{
  [
  "original":  {"mod_loaded": "foobar"}, 
  "alias": {"mod_loaded": "foobar_forked"}
  ]
)
````
That's just an example, not a serious suggestion. :-)

**Comment by Technici4n** [comment]
These are good suggestions (besides the last one :-P), but I think this API is pending a datagen API, as we will have to integrate it. Perhaps a more complex approach where serializers have to be registered could be useful for datagen, but for that we need a datagen API (and thus global access wideners... hopefully we will have those in a few weeks :)).

**Comment by kanpov** [comment]
What's the status on this PR?

**Comment by Technici4n** [comment]
Waiting for datagen API, which itself is waiting for #1802.

**Comment by AlphaMode** [comment]
@Technici4n Do you plan on working on this soon now that #1824 and #1802 are merged?

**Comment by kanpov** [comment]
> @Technici4n Do you plan on working on this soon now that #1824 and #1802 are merged?

Welp, both of these PRs have already been merged, so...

**Comment by Technici4n** [comment]
Yes, see #api on discord for the moment as there are some early design decisions that need to be made.

**Comment by Technici4n** [comment]
Updated the PR with a new approach, going to continue this over the coming days.

File: fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+	@Nullable
+	public static Predicate<JsonObject> get(Identifier identifier) {
+		return REGISTERED_CONDITIONS.get(identifier);
+	}
+
+	/**
+	 * Check if the passed JSON object either has no {@code fabric:conditions} tag, or all of its conditions match.
+	 * This should be called for objects that may contain a conditions entry.
+	 */
+	public static boolean objectMatchesConditions(JsonObject object) {
```
shartte, line 118: Maybe this? I am not so sure about this...

```suggestion
	public static boolean resourceMatchesConditions(JsonObject object) {
```

Technici4n: The idea is that these conditions can also be used for nested objects, not necessarily the resource itself. For example if you only want some outputs to be produced if a mod is loaded.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/FabricResourceConditions.java
```diff
@@ -0,0 +1,191 @@
...
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.JsonHelper;
+
+import net.fabricmc.loader.api.FabricLoader;
+
+/**
+ * Contains {@link ConditionJsonProvider}s for resource conditions provided by fabric itself.
+ */
+public final class FabricResourceConditions {
```
Juuxel: Maybe `StandardResourceConditions` or `DefaultResourceConditions`? (think JDK `StandardOpenOption`, `StandardCharsets` etc)

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Return the identifier of this condition. This is only for use by {@link #toJson()} to write it.
+	 */
+	Identifier getConditionIdentifier();
+
+	/**
+	 * Serialize the parameters of this condition. This is only for use by {@link #toJson()}.
+	 */
+	void serialize(JsonObject object);
```
Juuxel: Maybe `writeJson` to match yarn's naming patterns?

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,69 @@
...
+		JsonObject jsonObject = new JsonObject();
+		jsonObject.addProperty(ResourceConditions.CONDITION_ID, getConditionIdentifier().toString());
+		this.serialize(jsonObject);
+		return jsonObject;
+	}
+
+	/**
+	 * Return the identifier of this condition. This is only for use by {@link #toJson()} to write it.
+	 */
+	Identifier getConditionIdentifier();
```
Juuxel: `getConditionId` is probably fine too, we don't usually call identifiers by the full word

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * <p>A conditional JSON object must have a {@link #CONDITIONS_KEY} entry, containing an array of condition objects.
+ * The conditions in the array must be satisfied to load the resource.
+ * Each condition object must contain a {@link #CONDITION_ID} entry with the identifier of the condition,
+ * and it may also contain additional data for the condition.
+ * Here is an example of a resource that is only loaded if no mod with id {@code a} is loaded:
+ * <pre>{@code
+ * {
+ *   ... // normal contents of the resource
+ *   "fabric:conditions": [ // array of condition objects
```
Juuxel:
```suggestion
 *   "fabric:load_conditions": [ // array of condition objects
```

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -0,0 +1,128 @@
...
+ * Contains {@link ConditionJsonProvider}s for resource conditions provided by fabric itself.
+ */
+public final class DefaultResourceConditions {
+	private static final Identifier NOT = new Identifier("fabric:not");
+	private static final Identifier AND = new Identifier("fabric:and");
+	private static final Identifier OR = new Identifier("fabric:or");
+	private static final Identifier ALL_MODS_LOADED = new Identifier("fabric:all_mods_loaded");
+	private static final Identifier ANY_MOD_LOADED = new Identifier("fabric:any_mod_loaded");
+	private static final Identifier FLUID_TAGS_POPULATED = new Identifier("fabric:fluid_tags_populated");
+	private static final Identifier ITEM_TAGS_POPULATED = new Identifier("fabric:item_tags_populated");
```
Juuxel, line 45: Hmm, why not block tags? (just curious)

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
```diff
@@ -0,0 +1,176 @@
...
+
+import net.minecraft.resource.JsonDataLoader;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.JsonHelper;
+
+import net.fabricmc.fabric.impl.resource.conditions.ResourceConditionsImpl;
+
+/**
+ * Registration and access to resource loading conditions.
+ * A resource condition is an identified {@code Predicate<JsonElement>} that can decide whether a resource should be loaded or not.
```
Juuxel:
```suggestion
 * A resource condition is an identified {@code Predicate<JsonObject>} that can decide whether a resource should be loaded or not.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLootTableProvider.java
```diff
@@ -52,19 +56,39 @@
 
 	FabricDataGenerator getFabricDataGenerator();
 
+	/**
+	 * Return a new exporter that applies the specified conditions to any loot table it receives.
+	 */
+	default BiConsumer<Identifier, LootTable.Builder> withConditions(BiConsumer<Identifier, LootTable.Builder> exporter, ConditionJsonProvider... conditions) {
+		Preconditions.checkArgument(conditions.length > 0);
```
Juuxel: Should this have an exception message? (Also applies to all the corresponding checks in other methods)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
```diff
@@ -89,4 +93,18 @@ public static void run() throws IOException {
 			dataGenerator.run();
 		}
 	}
+
+	/**
+	 * Used to keep track of conditions associated to generated objects.
+	 */
+	private static final Map<Object, ConditionJsonProvider[]> CONDITIONS_MAP = new IdentityHashMap<>();
```
Juuxel, line 128: This feels a bit cursed, but I can't think of any better solutions either

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A resource condition and its parameters that can be serialized to JSON.
+ */
+public interface ConditionJsonProvider {
```
Juuxel, line 29: `ConditionSerializer`? Although this is already configured to be for a specific condition...

Juuxel (changes requested):
I still believe that a file/directory-level (instead of file content-based) approach would be simpler and easier to use. For one, it allows to exclude entire directories by default without duplicating the same `fabric:load_conditions` block to each JSON file. This works for mods that organise their mod compat files by the supported mod (`data/adorn/loot_tables/blocks/terrestria` etc) It also works automatically for each and every data type.

In any case, here's a review of what you have now.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: This was done to match `RecipeJsonProvider#serialize`. Maybe yarn should be changed here.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: This interface is basically the equivalent of `RecipeJsonProvider` for conditions, that's why I chose that name. :wink:

Juuxel: That makes sense then - recipes have serialisers already, so that is just for datagen. I'd clarify in the doc of this class that this is intended for use with the data generation API specifically.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: Renamed to `writeAdditional`, still not convinced. But `writeJson` is not very descriptive. Maybe `writeParameters`...

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
Technici4n: Didn't feel necessary for recipes, but I just added them because why not.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: Expanded the docs a little.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Juuxel: Oh yeah, it doesn't include the ID. `writeParameters` seems fine then

Technici4n: Went for `writeParameters`.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
Juuxel: It might be necessary for loot tables, actually (iirc they have some tag support)

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
```diff
@@ -0,0 +1,176 @@
...
+public final class ResourceConditions {
+	private static final Map<Identifier, Predicate<JsonObject>> REGISTERED_CONDITIONS = new ConcurrentHashMap<>();
+	/**
+	 * The key fabric uses to identify resource conditions in a JSON object.
+	 */
+	public static final String CONDITIONS_KEY = "fabric:load_conditions";
+	/**
+	 * The key identifying the resource condition's identifier inside a condition object.
+	 */
+	public static final String CONDITION_ID = "condition";
```
Juuxel: Should this also have a `_KEY` suffix like the other key/property name field?

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,69 @@
...
+	 */
+	default JsonObject toJson() {
+		JsonObject jsonObject = new JsonObject();
+		jsonObject.addProperty(ResourceConditions.CONDITION_ID, getConditionId().toString());
+		this.writeParameters(jsonObject);
+		return jsonObject;
+	}
+
+	/**
+	 * Return the identifier of this condition. This is only for use by {@link #toJson()} to write it.
```
Juuxel:
```suggestion
	 * {@return the identifier of this condition} This is only for use by {@link #toJson()} to write it.
```

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A resource condition and its parameters that can be serialized to JSON, meant for use in data generators.
+ */
+public interface ConditionJsonProvider {
+	/**
+	 * Write the passed conditions to a json object.
```
Juuxel:
```suggestion
	 * Write the passed conditions to a JSON object in the {@value ResourceConditions#CONDITIONS_KEY} array.
	 *
	 * @throws IllegalArgumentException if the JSON object already contains that array
```

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
```diff
@@ -0,0 +1,176 @@
...
+ *       }
+ *     }
+ *   ]
+ * }
+ * }</pre>
+ */
+public final class ResourceConditions {
+	private static final Map<Identifier, Predicate<JsonObject>> REGISTERED_CONDITIONS = new ConcurrentHashMap<>();
+	/**
+	 * The key fabric uses to identify resource conditions in a JSON object.
```
Juuxel:
```suggestion
	 * The key ({@value}) Fabric uses to identify resource conditions in a JSON object.
```
This shows the actual key as well in the generated javadoc (without going to the list of constants).

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
```diff
@@ -0,0 +1,176 @@
...
+	 */
+	public static final String CONDITIONS_KEY = "fabric:load_conditions";
+	/**
+	 * The key identifying the resource condition's identifier inside a condition object.
```
Juuxel:
```suggestion
	 * The key ({@value}) identifying the resource condition's identifier inside a condition object.
```

Juuxel (changes requested):

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceConditions.java
Technici4n: Interesting!

liach:
I recall shedaniel's draft had metadata files defined for any resource file (like `.mcmeta` files used by vanilla, also JSON format), so those conditions work for any resource such as command functions than json-only. In addition, your patch can be easily bypassed if someone isn't using json data loader; I don't think tags use the json data loader.

Juuxel:
> I don't think tags use the json data loader

@liach Note that tags don't need it since the values can be marked as optional, in which case they're ignored if not present.

Technici4n:
Message I wrote on discord:
```
So, regarding recipe conditions. There are basically two options:
1) extend some json formats, for example we'd extend the recipe json format with a fabric:conditions: {...}.
2) work at the resourcepack level directly, to allow blacklisting entire directories.
APPROACH 1
+ easy to implement
+ guaranteed not to slow down the resource manager
+ can implement conditions that depend on previous resource reload listeners. For example, you can have a tag_exists_with_at_least_one_entry recipe condition, which is very useful
+ not limited to root json objects; can also be used for sub-objects, e.g. to conditionally remove a loot pool from a loot table
- you need to manually add a conditions entry to each recipe (LARGELY if not entirely mitigated by datagen imo)
- we'll have to patch multiple json formats VS just patching the resource manager
APPROACH 2
+ much easier to use for simple compat recipes if you don't have datagen. Just place a .fabricmeta in the relevant directory and the conditions will be applied to all files in the directory
+ only requires resource manager patches VS having to patch multiple json formats
- much harder to implement
- might slow down the resource manager
- conditions can't depend on previous reload listeners
For better conditions, and because it's a lot easier, I suggest going for approach 1
```
TL;DR 1 is good enough, guaranteed not to be slower and much easier to implement.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @throws IllegalArgumentException if the JSON object already contains that array
+	 */
+	static void write(JsonObject conditionalObject, ConditionJsonProvider @Nullable... conditions) {
+		if (conditions == null) { // no condition -> skip
+			return;
+		}
+
+		Preconditions.checkArgument(conditions.length > 0, "Must write at least one condition."); // probably a programmer error
+		if (conditionalObject.has(ResourceConditions.CONDITIONS_KEY)) throw new IllegalArgumentException("Object already has a condition entry: " + conditionalObject);
```
liach, line 41: Shouldn't we try to merge the conditions if there are already existing ones?

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -0,0 +1,71 @@
...
+ * A resource condition and its parameters that can be serialized to JSON, meant for use in data generators.
+ */
+public interface ConditionJsonProvider {
+	/**
+	 * Write the passed conditions to a JSON object in the {@value ResourceConditions#CONDITIONS_KEY} array.
+	 *
+	 * @throws IllegalArgumentException if the JSON object already contains that array
+	 */
+	static void write(JsonObject conditionalObject, ConditionJsonProvider @Nullable... conditions) {
+		if (conditions == null) { // no condition -> skip
```
liach, line 36: It'd be better if we handle the null and the 0-length array case the same

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: Datagen already merges multiple ConditionJsonProvider[] together.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
Technici4n: The 0-length case is usually a programmer error, while the `null` case is conveniently used to skip unspecified entries in the datagen.

liach: > The 0-length case is usually a programmer error, while the `null` case is conveniently used to skip unspecified entries in the datagen.

Then why not just ask the modders to not call this method if they have nothing to write to the conditions after all

Technici4n: Well I guess modders could make use of it in their own datagen if they need their own conditions

shartte (approved):

Juuxel (approved):

Player3324 (approved):

modmuss50 (approved):

Mysterious-Dev (approved):

modmuss50:
Going to merge and release now. Sorry for the delay ive been busy IRL 👍

LemmaEOF:
https://github.com/cottonmc/libcd

PR #1681, opened 2021-08-28
Sleeping event fixes

### LINKED ISSUE #1679: Fabric EntitySleepEvents issues
Had the very bad surprise of those events breaking my sleeping bags in [Aurora's Decorations](https://github.com/LambdAurora/AurorasDecorations/).

The reason being that Vanilla detects the sleeping direction with the code `Direction direction = this.world.getBlockState(pos).get(HorizontalFacingBlock.FACING);`
But FAPI hardcodes the check to BedBlock: `Direction initial = state.getBlock() instanceof BedBlock ? (Direction) state.get(property) : null;`
FAPI should check if the state contains `HorizontalFacingBlock.FACING` and use it if available instead.

Another nitpick is the injection of `ALLOW_BED`, it injects at RETURN in:
```java
    private boolean isSleepingInBed() {
        return this.getSleepingPosition().map(pos -> this.world.getBlockState(pos).getBlock() instanceof BedBlock).orElse(false);
    }
```
and calls back `getSleepingPosition`, which is not the greatest thing imo, the injection could be in the lambda instead, still at return. The lambda intermediary is `method_18405`.

### LINKED ISSUE #1680: NPE when returning a SleepFailureReason with null message
This is not a Fabric API bug per se, it can happen in vanilla with a special dimension datapack, but with the introduction of `EntitySleepEvents.ALLOW_SLEEPING` this can occur several times.

It can be replicated with: 
```java
EntitySleepEvents.ALLOW_SLEEPING.register((player, pos) -> PlayerEntity.SleepFailureReason.OTHER_PROBLEM);
```
then clicking a bed block at night in the Overworld.

Log (Warn: Mojang Official Mappings): https://gist.github.com/alkyaly/5c6207833bb8315340d158643349b7ca

BedBlock#onUse never null checks for a SFR text, passing null to PlayerEntity#sendMessage.
This could be easily solved with a injection inside the onUse lambda:
(Warn: Mojang Official Mappings): https://github.com/alkyaly/Somnia-Fabric/blob/1.17-Fabric/src/main/java/io/github/alkyaly/somnia/mixin/BedBlockMixin.java#L29-L35

Having the mixin in Fabric API would avoid duplicated code within mods that depend on that event.

Juuxel:
Closes #1679 and fixes #1680. Also clarified what the "sleeping direction" means in jd.

This probably needs backporting to 1.16 as well.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
@@ -296,6 +296,7 @@ public final class EntitySleepEvents {
 	public interface ModifySleepingDirection {
 		/**
 		 * Modifies or provides a sleeping direction for a block.
+		 * The sleeping direction is where a player's head is pointing when they're sleeping.
 		 *
 		 * @param entity            the sleeping entity
 		 * @param sleepingPos       the position of the block slept on
--- a/fabric-entity-events-v1/build.gradle
+++ b/fabric-entity-events-v1/build.gradle
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-entity-events-v1"
-version = getSubprojectVersion(project, "1.2.2")
+version = getSubprojectVersion(project, "1.2.3")
 
 moduleDependencies(project, [
 		'fabric-api-base'
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/BedBlockMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/BedBlockMixin.java
@@ -1,0 +1,23 @@
+package net.fabricmc.fabric.mixin.entity.event;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.block.BedBlock;
+import net.minecraft.entity.player.PlayerEntity;
+
+@Mixin(BedBlock.class)
+abstract class BedBlockMixin {
+	// Synthetic lambda body for Either.ifLeft in onUse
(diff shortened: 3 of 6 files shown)


Technici4n (approved):
LGTM

LambdAurora (approved):

liach (approved):

modmuss50 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
```diff
@@ -296,6 +296,7 @@
 	public interface ModifySleepingDirection {
 		/**
 		 * Modifies or provides a sleeping direction for a block.
+		 * The sleeping direction is where a player's head is pointing when they're sleeping.
```
haykam821, line 299: Does this refer to the first-person perspective direction or the direction the upper half of the player's body is in?

Juuxel: The upper half of the body. The player's head is always pointing up in the sleeping pose, so it's unambiguous here.

haykam821: There are three possible directions based on this description:

1. The direction in which the upper half of the player's body is facing: Upper half of sleeping pose (south)
2. The direction the player faces in the first-person perspective: First-person sleeping pose (north)
3. The direction the player faces in a third-person perspective: Third-person sleeping pose (up)

In addition, the player can face additional directions by entering a bed in a third-person perspective:

Player3324 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntitySleepEvents.java
Player3324: Would be nice to clarify this a little. like "the position the body is pointing at when looking from feet to head" or so, assuming that's correct.

liach:
```suggestion
		 * The sleeping direction is where a player's head is pointing from the feet when they're sleeping.
```

Juuxel, line 299:
```suggestion
		 * The sleeping direction is where a player's head is located from their feet when they're sleeping.
```
What about this?

PR #1685, opened 2021-08-29
Add Transaction#getCurrentUnsafe and fix mixin fields

Technici4n:
`Transaction#getCurrentUnsafe` makes my life a bit easier when implementing support for AE2. It will also be useful for LBA should it attempt to have compatibility.

Adding a `fabric_` prefix will prevent conflicts with older versions of FastTransferLib that used to contain the exact same `ItemMixin` (with a `cachedItemVariant` field). `@Unique` only prevents conflicts with the mixin target, so it is effectively useless in 99% of the cases.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionManagerImpl.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionManagerImpl.java
@@ -50,7 +50,7 @@ public class TransactionManagerImpl {
 		} else if (stack.get(currentDepth).isOpen) {
 			return stack.get(currentDepth);
 		} else {
-			throw new IllegalStateException("May not call getCurrentTransaction() from a close callback.");
+			throw new IllegalStateException("May not call getCurrentUnsafe() from a close callback.");
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/transaction/TransactionManagerImpl.java
```diff
@@ -40,6 +43,17 @@ public Transaction openOuter() {
...
 	}
 
+	@Nullable
+	public TransactionContext getCurrentUnsafe() {
+		if (currentDepth == -1) {
+			return null;
+		} else if (stack.get(currentDepth).isOpen) {
+			return stack.get(currentDepth);
+		} else {
+			throw new IllegalStateException("May not call getCurrentTransaction() from a close callback.");
```
BasiqueEvangelist:
```suggestion
			throw new IllegalStateException("May not call getCurrentUnsafe() from a close callback.");
```

Technici4n: Oops. :smile:

Player3324 (approved):

modmuss50 (approved):

Issue #1684, opened 2021-08-29
Feature request: custom liquid block renderer


FoxSamu:
The `fabric-rendering-fluid` provides an easy and compatible way to assign custom textures to modded fluid blocks. However, when it comes to more customized implementations of fluid blocks, the builtin fluid renderer may not be enough.

As an example, I'll use the idea of a fluid that acts entirely upside down (there could be actual uses for that). The default `FlowingFluid` implementation does not work for this, so this needs a custom implementation of `Fluid`: `UpwardsFlowingFluid`. Implementing and registering the custom fluid block is not an issue, but the vanilla fluid renderer is not made to render fluid blocks upside down. Instead, we want a custom fluid renderer implementation to render the fluid upside down.

Currently, the fluid rendering API allows a modder to register a `FluidRenderHandler` for a custom fluid, which allows to select custom textures for a fluid (bypasses the hardcoded fluid textures in vanilla), or use a custom tint for the fluid:
```java
FluidRenderHandlerRegistry.INSTANCE.register(ModFluids.STILL_FLUID, (level, pos, state) -> new TextureAtlasSprite[] {
    still,
    flow
});

FluidRenderHandlerRegistry.INSTANCE.register(ModFluids.FLOWING_FLUID, (level, pos, state) -> new TextureAtlasSprite[] {
    still,
    flow
});
```
This is the existing functionality of the API. I wanna propose a `CustomFluidRenderer`, which is called instead of the vanilla fluid renderer, as soon as the fluid it is registered for is rendered.
```java
FluidRendererRegistry.INSTANCE.register(ModFluids.STILL_FLUID, ModFluidRenderer.INSTANCE);
FluidRendererRegistry.INSTANCE.register(ModFluids.FLOWING_FLUID, ModFluidRenderer.INSTANCE);
```

~~I somehow failed to clone and set up the Fabric API project locally, but I could make a PR as soon as that works for me.~~ This works now, see #1687.

FoxSamu:
Closing this as this is now a pull request: see #1687

PR #1687, opened 2021-08-29
Custom fluid renderers and enhanced vanilla fluid renderer customization

FoxSamu:
PR for #1684:

> The `fabric-rendering-fluid` provides an easy and compatible way to assign custom textures to modded fluid blocks. However, when it comes to more customized implementations of fluid blocks, the builtin fluid renderer may not be enough.
> 
> As an example, I'll use the idea of a fluid that acts entirely upside down (there could be actual uses for that). The default `FlowingFluid` implementation does not work for this, so this needs a custom implementation of `Fluid`: `UpwardsFlowingFluid`. Implementing and registering the custom fluid block is not an issue, but the vanilla fluid renderer is not made to render fluid blocks upside down. Instead, we want a custom fluid renderer implementation to render the fluid upside down.

```java
FluidRendererRegistry.INSTANCE.registerCustomRenderer(ModFluids.AIR, HiddenFluidRenderer.INSTANCE);
FluidRendererRegistry.INSTANCE.registerCustomRenderer(ModFluids.FLOWING_AIR, HiddenFluidRenderer.INSTANCE);
```

```java
public class HiddenFluidRenderer implements CustomFluidRenderer {
    public static final HiddenFluidRenderer INSTANCE = new HiddenFluidRenderer();

    @Override
    public boolean renderFluid(BlockPos pos, BlockRenderView world, VertexConsumer vertexConsumer, FluidState state) {
        // Mods could implement completely custom rendering here. This could be useful for blocks too, something to consider
        
        // True is returned if any face is tessellated, false otherwise
        return false;
    }
}
```

This PR adds the ability to register custom renderers for modded fluids, or vanilla fluids if wanted. A custom fluid renderer overrides `FluidRenderHandler`s for the fluids the renderer is registered for, but the renderer can still decide to look up the desired fluid textures through the `FluidRenderHandler`.

**Overview of the changes**
- `FluidRenderHandlerRegistry` is now `FluidRenderRegistry`, and `FluidRenderHandlerRegistry` is kept as a deprecated compatibility wrapper.
  - `get` is now `getRenderHandler`, with `get` being kept as deprecated compatibility proxy.
  - `register` is now `registerRenderHandler`, with `register` being kept as deprecated compatibility proxy.
- `FluidRenderRegistry` now allows the registry of `CustomFluidRenderer`s, alongside `FluidRenderHandlers`.
  - Similar `registerCustomRenderer` and `getCustomRenderer` exist now to register/get the custom renderers
- `CustomFluidRenderer.DEFAULT` delegates to the vanilla fluid renderer directly
- ~~A new mixin into `BlockRenderManager` redirects the default fluid renderer to the custom fluid renderer, if one is registered for the fluid being rendered.~~ (Done via existing mixin now)
- A `DefaultLikeFluidRenderer` exists that can be instantiated/inherited for more customized, but default fluid rendering.
- It is now possible to return a glass overlay sprite for a custom fluid by returning a third sprite in the sprite array.
- A `FluidOverlayBlock` interface is added that blocks can implement to enforce an overlay texture instead of a falling fluid texture where the fluid is seen through the block. Comes in handy where blocks cannot extend `TransparentBlock` anymore because they already extend another block type.

**In other words**
- Fluids can now have entirely custom renderers
- Glass overlay sprites can now be specified as part of the fluid render handler
- It is now possible to make any custom block have a fluid overlay texture (instead of the flowing texture)


fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultFluidRenderer.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockRenderView;
+
+import java.util.concurrent.locks.ReentrantLock;
+
+public class DefaultFluidRenderer implements CustomFluidRenderer {
+	public static final DefaultFluidRenderer INSTANCE = new DefaultFluidRenderer();
+
+	// Lazily loaded because FluidRenderer might not yet exist at instantiation time
+	private FluidRenderer defaultRenderer;
```
Technici4n: You need this to be `volatile` for double-checked locking to be safe.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/CustomFluidRenderer.java
```diff
@@ -0,0 +1,38 @@
...
+	 * <p>Note that this method must *only* return {@code true} if at least one face is tessellated. If no faces are
+	 * tessellated this method must return {@code false}.
+	 *
+	 * @param pos            The position in the world, of the fluid to render
+	 * @param world          The world the fluid is in
+	 * @param vertexConsumer The vertex consumer to tessellate the fluid in
+	 * @param state          The fluid state being rendered
+	 * @return Whether anything is tessellated
+	 */
+	boolean renderFluid(BlockPos pos, BlockRenderView world, VertexConsumer vertexConsumer, FluidState state);
```
Technici4n: Is there any reason this is not a method in `FluidRenderHandler`? Something like

```java
default boolean renderFluid(BlockPos pos, BlockRenderView world, VertexConsumer vertexConsumer, FluidState state) {
    // static method that delegates to vanilla, somewhere in impl.
    return FluidRenderHandlerImpl.defaultRenderFluid(pos, world, vertexConsumer, state);
}
```

would work as well I think?

Technici4n: Maybe you should pass the `FluidRenderer` as a parameter? Then the default method body could just call it directly.

Technici4n (commented):
I thought that if you wanted a fluid that flows upside down, you needed a ton of patches to fluid logic. Maybe that's a wrong assumption that I made?

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultFluidRenderer.java
FoxSamu: This entire class is removed now

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/CustomFluidRenderer.java
FoxSamu: Hm... I think the delegation via `FluidRenderHandler` is more stable than directly exposing `FluidRenderer`. Can look into that, but I also added a `DefaultLikeFluidRenderer` now, that imitates vanilla's behaviour but can be inherited to modify certain parameters or behaviours.

FoxSamu: Although, I am not sure how well it works to ship a `DefaultLikeFluidRenderer` that is an entire copy of the default fluid renderer, except with some extra methods to inherit.

Technici4n: This is probably not a good idea due to the maintenance cost that comes with it.

FoxSamu:
> I thought that if you wanted a fluid that flows upside down, you needed a ton of patches to fluid logic. Maybe that's a wrong assumption that I made?

@Technici4n This requires a ton of patches, but I achieved it by just copying, deobfing and modifying `FlowingFluid` myself. Works for most fluid logic. Still though, that's just an example.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/CustomFluidRenderer.java
FoxSamu: I added access to the default implementation via an extra parameter that wraps the `FluidRenderer`, to not directly expose the `FluidRenderer` (because I'm not sure how robust that is, plus it's not documented)

FoxSamu:
That should do it. Feedback is highly appreciated, especially about `DefaultLikeFluidRenderer` (is it useful?). I might also have made some style-related mistakes.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/CustomFluidRenderer.java
FoxSamu: The new `CustomFluidRenderer` interface was intended by the way. This makes it possible to define one render type for fluids with different render handlers.

I can understand the maintenance cost of `DefaultLikeFluidRenderer`. I am fine with removing it if mainenance cost outweighs its usefullness.
However, it could be handy if a lot of modders just want to adjust their fluids more closely, while keeping it look like vanilla, without having to imitate vanilla's code all by themselves or break Fabric's mixins with their own. Plus you avoid a whole bunch of mixins from Fabric that could drain some performance. I might even believe it's not even that hard to maintain, but I'm not sure.
Also, I realised it's in the `impl` package :facepalm:

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
+import net.minecraft.util.math.MathHelper;
+import net.minecraft.util.math.Vec3d;
+import net.minecraft.util.shape.VoxelShape;
+import net.minecraft.util.shape.VoxelShapes;
+import net.minecraft.world.BlockRenderView;
+import net.minecraft.world.BlockView;
+
+public class DefaultLikeFluidRenderer implements CustomFluidRenderer {
```
FoxSamu: Should be in `api` package

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+
+				tessellated = true;
+				heightNW -= 0.001f;
+				heightSW -= 0.001f;
+				heightSE -= 0.001f;
+				heightNE -= 0.001f;
+
+				Vec3d velo = state.getVelocity(world, pos);
+				Sprite sprite;
+				if (velo.x == 0 && velo.z == 0) {
```
FoxSamu: Some overridable method could be added to choose flowing or still textures, and another to determine flow angle.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+
+					direction = Direction.EAST;
+					shouldRender = renderEast;
+				}
+
+				if (shouldRender && !isSideCovered(world, pos, direction, Math.max(heightL, heightR))) {
+					tessellated = true;
+
+					BlockPos near = pos.offset(direction);
+					Sprite sprite = sprites[1];
```
FoxSamu: Again, overridable method to decide between flowing or still texture

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+				}
+			}
+
+			return tessellated;
+		}
+	}
+
+	private void vertex(VertexConsumer vertexConsumer, double x, double y, double z, float red, float green, float blue, float u, float v, int light) {
+		vertexConsumer.vertex(x, y, z)
+					  .color(red, green, blue, 1)
```
FoxSamu: Support alpha?

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+
+public class DefaultLikeFluidRenderer implements CustomFluidRenderer {
+	private static final float HALF_PI = MathHelper.field_29845;
+
+	@Override
+	public boolean renderFluid(BlockPos pos, BlockRenderView world, VertexConsumer vertexConsumer, FluidState state, DefaultBehavior defaultBehavior) {
+		return render(world, pos, vertexConsumer, state);
+	}
+
+	private static final float DEFAULT_FLUID_HEIGHT = 8 / 9f;
```
FoxSamu: Could be changed by overridable method

FoxSamu:
(self review for tomorrow :) )

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/DefaultLikeFluidRenderer.java
ghost: If the render method is native to this class and not referenced by and external classes there is no point specifying it separately if renderfluid is a passthrough to it.

Might as well move the body of render to renderFluid to minimize the code a bit.

ghost (approved):

Technici4n:
After more thinking, here is a list of changes that you should implement to make this a lot simpler:
- Please just add a method to FluidRenderHandler, this will keep the API changes minimal. There is no need to separate sprite/color and rendering because the fluid is custom anyway.
- To delegate to the vanilla fluid renderer, just add a static method to an impl class and call it from the default method body. If people want to call it they can just use `FluidRenderHandler.super.render(...)`. That will be a lot cleaner than passing an opaque parameter which requires a new API type.
- You should get rid of the "default like" implementation, this is too high maintenance. Feel free to expose it in a third party library though, I'm sure it can be useful.
- You should probably replace the interface for transparent blocks by a tag.
- Please expand the testmod so we can validate the changes. Just rendering something on top of the default render should be fine for illustration.

FoxSamu:
Thanks for the feedback! I will work on this 👍

The tag for overlay blocks will be hard though, as this is a client-only API and tags are server-side. Though, the interface is neither a good solution so I will probably have to make a registry instead (like for render types).

FoxSamu:
Also, it could be useful to expose the reload-event of the fluid renderer, could make it easier to load fluid textures without having to use another api to register reload listeners.

Technici4n:
> Also, it could be useful to expose the reload-event of the fluid renderer, could make it easier to load fluid textures without having to use another api to register reload listeners.

Sure, if you can find a nice way to do this why not!

Technici4n:
> The tag for overlay blocks will be hard though, as this is a client-only API and tags are server-side. Though, the interface is neither a good solution so I will probably have to make a registry instead (like for render types).

Hmmm, I have the feeling that we have too many registries for trivial things in fabric. Adding one more makes it worse. I don't know what else can be done though. :(

FoxSamu:
> Hmmm, I have the feeling that we have too many registries for trivial things in fabric. Adding one more makes it worse. I don't know what else can be done though. :(

I don't see any other option here. It could be part of the `FluidRenderHandlerRegistry`. It might also be possible to make it configurable via `FabricBlockSettings` but I'm unsure about that one.

Technici4n:
Yeah a registry for interaction between the fluid renderer and blocks should be fine. If we need to expand the behavior we will be able to expand the registry later. If you go for that, note that new registries should have public static methods. (No public INSTANCE field anymore).

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Creates a fluid render handler with an overlay texture.
+	 *
+	 * @param stillTexture   The texture for still fluid
+	 * @param flowingTexture The texture for flowing/falling fluid
+	 * @param overlayTexture The texture behind glass, leaves and other {@linkplain
+	 *                       FluidRenderHandlerRegistry#setBlockTransparency registered transparent blocks}
+	 */
+	public SimpleFluidRenderHandler(Identifier stillTexture, Identifier flowingTexture, Identifier overlayTexture) {
```
Technici4n:
```suggestion
	public SimpleFluidRenderHandler(Identifier stillTexture, Identifier flowingTexture, @Nullable Identifier overlayTexture) {
```

Technici4n (commented):
API changes look amazing. The `SimpleFluidRenderHandler` is a great addition!
I'll have to take a closer look at the mixin changes.

I recommend trying to keep the javadoc style consistent with the rest of this module. So no alignment of param descriptions, and a period at the end of these descriptions.

You can run `gradlew licenseFormatAll` to add license headers and `gradlew check` to run checkstyle, otherwise this PR will fail CI when it will be enabled.

FoxSamu:
I'll do the cleaning up when I finish the test mod. Will have to force myself to not run the IntelliJ formatter every time (or reconfigure it).

FoxSamu:
I added a couple of fluids to the test mod that demonstrate various different ways of rendering.
image

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java
```diff
@@ -42,16 +46,38 @@
...
+		FluidRenderHandlerRegistryImpl.INSTANCE.onFluidRendererReload(self, waterSprites, lavaSprites, waterOverlaySprite);
 	}
 
 	@Inject(at = @At("HEAD"), method = "render", cancellable = true)
 	public void tesselate(BlockRenderView view, BlockPos pos, VertexConsumer vertexConsumer, FluidState state, CallbackInfoReturnable<Boolean> info) {
+		if (!fabric_customRendering.get()) {
+			// Prevent recursively looking up custom fluid renderers when default behaviour is being invoked
+			fabric_customRendering.set(true);
+			tessellateViaHandler(view, pos, vertexConsumer, state, info);
+			fabric_customRendering.set(false);
```
Technici4n: `try`/`finally`?

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.client.texture.Sprite;
+import net.minecraft.client.texture.SpriteAtlasTexture;
+import net.minecraft.fluid.FluidState;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockRenderView;
+
+/**
+ * A simple fluid render handler that uses and loads sprites given by their
+ * identifiers. Most fluids don't need more than this. The fluid is not tinted.
```
Technici4n: Imo allowing a fixed tint would be very useful. For example all of my mod's fluids are tinted water. :smile:

Technici4n (commented):
Sorry for the delay, I can't review right now, but I should have time on Wednesday / Thursday! If anyone wants to do a first review, go ahead!

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java
FoxSamu: ah yes of course!

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
FoxSamu: True, I realized that might be handy. I added a custom method `SimpleFluidRenderHandler.coloredWater(tint)` just for this, and provided the identifiers of vanilla's textures as constants.

FoxSamu:
A bit of a delay, as I moved on with another project, but didn't forget.

I made the requested changes. I also added a way to register a pair of a still and flowing fluid at once, because I made a mistake regarding this multiple times 😅 (registering only the still fluid). I hope that makes it a bit clearer to users that both fluids must be registered, plus it uses one rendering instance for two fluids which can be faster in terms of caching.

Registering a renderer for a fluid with colored water textures now becomes a one-liner, with no more need to load and cache the fluid sprites via a resource reload listener and no need to implement an interface manually anymore:
```java
// Red vanilla water texture
FluidRenderHandlerRegistry.INSTANCE.register(MyFluids.STILL, MyFluids.FLOWING, SimpleFluidRenderHandler.coloredWater(0xFF0000));
```

Technici4n (approved):
Looks very nice, thanks a lot!

liach (approved):

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
```diff
@@ -41,8 +44,54 @@
...
+	 * @param renderer The FluidRenderHandler.
+	 */
+	default void register(Fluid still, Fluid flow, FluidRenderHandler renderer) {
+		register(still, renderer);
+		register(flow, renderer);
+	}
+
+	/**
+	 * Registers whether a block is transparent or not. When a block is
+	 * tranparent, the flowing fluid texture to the sides of that block is
```
Juuxel:
```suggestion
	 * transparent, the flowing fluid texture to the sides of that block is
```

Player3324 (approved):
LGTM, thanks!

modmuss50 (approved):
Awesome. 👍

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java
```diff
@@ -42,16 +46,41 @@
...
+	private Sprite waterOverlaySprite;
 
 	private final ThreadLocal<FluidRendererHookContainer> fabric_renderHandler = ThreadLocal.withInitial(FluidRendererHookContainer::new);
+	private final ThreadLocal<Boolean> fabric_customRendering = ThreadLocal.withInitial(() -> false);
+	private final ThreadLocal<Block> fabric_neighborBlock = new ThreadLocal<>();
 
 	@Inject(at = @At("RETURN"), method = "onResourceReload")
 	public void onResourceReloadReturn(CallbackInfo info) {
-		FluidRenderHandlerRegistryImpl.INSTANCE.onFluidRendererReload(waterSprites, lavaSprites);
+		FluidRenderer self = FluidRenderer.class.cast(this);
```
Juuxel:
```suggestion
		FluidRenderer self = (FluidRenderer) (Object) this;
```
More idiomatic, at least for Fabric API

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRendererHookContainer.java
```diff
@@ -27,11 +28,34 @@
...
+		if (handler != null) {
+			Sprite[] sprites = handler.getFluidSprites(world, pos, state);
+
+			this.sprites[0] = sprites[0];
+			this.sprites[1] = sprites[1];
+
+			if (sprites.length > 2) {
+				hasOverlay = true;
+				overlay = sprites[2];
+			}
```
Juuxel, line 45: Should there be an `else` here setting `hasOverlay` to false if there are fewer sprites, or does `clear` always handle that?

Juuxel (changes requested):

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRendererHookContainer.java
FoxSamu: The container is cleared every time before use so yes :)

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java
Technici4n: I think this is a smart way to avoid the warning. :smile:

PR #1688, opened 2021-08-29
Add TooltipComponentCallback

Technici4n:
Allows registering a custom mapping from `TooltipData` to `TooltipComponent`.

Unfortunately this is done by vanilla in the static `TooltipComponent#of` and mixin doesn't allow injecting into interface methods, even if they are static. So I used a mixin plugin to overwrite the method through ASM. This is the first time I use ASM, so it may not be optimal. Perhaps injecting right before the exception instead of throwing would be suitable as well, but then I'd have to learn how to do that. :smile:

### DIFF
--- a/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/TooltipComponentCallback.java
+++ b/fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/TooltipComponentCallback.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.rendering.v1;
+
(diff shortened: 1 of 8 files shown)


LambdAurora:
Why in the hell, just inject in the lambda of Screen instead?
But the best solution would be a new version of Fabric loader with the mixin changes, especially since it's merged now...

Technici4n:
That was just an excuse to do some ASM. :smile:

Note that the lambda injection is quite bad for mods that need to create the `TooltipComponent` from some `TooltipData` directly.
Example (mojmap): https://github.com/shedaniel/RoughlyEnoughItems/blob/ee4af3e5469b2f49f8e711e0752a345ec53726c6/fabric/src/main/java/me/shedaniel/rei/fabric/PluginDetectorImpl.java#L118-L134.

Waiting for an up to date loader is probably the best thing to do though.

shedaniel:
you can just make it not replace the method in asm, just inject with asm, mods injecting to the screen method is currently extremely annoying on my end and it remains unknown when loader will be updated, so this is a good compromise.

LambdAurora:
> Note that the lambda injection is quite bad for mods that need to create the `TooltipComponent` from some `TooltipData` directly.
> Example (mojmap): https://github.com/shedaniel/RoughlyEnoughItems/blob/ee4af3e5469b2f49f8e711e0752a345ec53726c6/fabric/src/main/java/me/shedaniel/rei/fabric/PluginDetectorImpl.java#L118-L134.

The thing is I don't see why you would need to create a `TooltipComponent` from some `TooltipData`, it's only useful to render tooltips, which the `Screen` class already offers you the method for.

Technici4n:
Btw we need to discuss the scope of such an extension:
- I think allowing TooltipData -> TooltipComponent conversions is useful for a lot of items. I can certainly use it for MI's barrel items.
- Providing a way for mods to add TooltipComponents to any item is also very useful, but the injection seemed a bit painful.

Being not very familiar with the system, I wanted to implement the former because I have a direct use for it.

LambdAurora:
The last one required a wait to create compounds of tooltip components and datas, similar to what Inspecio does.

PR #1689, opened 2021-08-29
Send block entity update packet along with regular block update packet

PotatoPresident:
Canceling a block break event on a block entity currently resets the block back to its data-less state. For example a steve head, blank sign and banners just disappear.

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinServerPlayerInteractionManager.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinServerPlayerInteractionManager.java
@@ -16,7 +16,6 @@
 
 package net.fabricmc.fabric.mixin.event.interaction;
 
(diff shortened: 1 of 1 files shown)


haykam821 (approved):

Player3324 (commented):
I think this is fine, but it's not a perfect fix - the update packet is not a full representation of the BE data. The BE may yield additional data in toInitialChunkDataNbt or use an external mechanism to transport its data.

It'd need clientside support to roll it back better, but vanilla should also have a solution with its spawn protection - worth checking how that handles the scenario this PR tries to fix.

PotatoPresident:
> It'd need clientside support to roll it back better, but vanilla should also have a solution with its spawn protection - worth checking how that handles the scenario this PR tries to fix.

Just looked, spawn protection doesn't deal with this and just ignores block nbt like the current implementation. 
Its not a perfect fix like you said, but I'm not sure what would be better.

Player3324 (approved):

Player3324:
Thanks for checking. Let's just go with this then, a more comprehensive solution would get very complex quickly.

modmuss50 (approved):

PR #1695, opened 2021-08-30
TooltipComponent API

Technici4n:


warjort:
Why not just the following?
It's alot harder to get wrong and doesn't require all the extra infrastructure of an event.

````
public interface FabricTooltipData extends TooltipData {

    @Environment(EnvType.CLIENT)
    TooltipComponent getTooltipComponent();
}

@Inject(at = @At("HEAD"), method = "method_32635(Ljava/util/List;Lnet/minecraft/client/item/TooltipData;)V", cancellable = true)
private static void injectRenderTooltipLambda(List<TooltipComponent> components, TooltipData data, CallbackInfo ci) {
    if (data instanceof FabricTooltipData fabricTooltipData) {
        TooltipComponent component = fabricTooltipData.getTooltipComponent();
        if (component != null) {
            components.add(1, component);
            ci.cancel();
        }

        // (*) Shouldn't ci.cancel() be here?
    }
}
````

Also, why not inject into ````TooltipComponent.of(TooltipData)```` instead of the ephemeral lambda?
If mojang introduce new logic that uses ````TooltipComponent.of()```` your injection won't get invoked.
It also can't be reused by modded code.

Finally, see the comment (*). Like your original it only cancels if there is a component, but I think it should cancel if the data is recognised. Allowing the TooltipData to optionally produce a component.

**Comment by LambdAurora** [comment]
@warjort The injection to be more correct would require Fabric Loader to be updated with the new mixin version, which is strange that it isn't yet since all the mixin related stuff was merged.

**Comment by modmuss50** [comment]
There are some upstream mixin bugs that are blocking us from updating.

**Comment by Technici4n** [comment]
`@Environment(EnvType.CLIENT)` should really be avoided as much as possible in user code. Imo using a separate class is a lot cleaner (especially when the data can be a one-liner record).

I'll add another event to append tooltip components for an arbitrary stack though.

We can expose a helper that calls the lambda under the hood, can't really do more for now unfortunately, but at least we'll be able to swiftly adjust the implementation yet preserve the API if mojang changes where it's used.

@LambdAurora I don't know if you had design notes regarding this API?

**Comment by modmuss50** [comment]
> @Environment(EnvType.CLIENT) should really be avoided as much as possible in user code

Yes, this is a no-go imo. I do have plans to look at splitting the client and server into diffrent source sets so this pattern wont work going forward.

**Comment by LambdAurora** [comment]
> @LambdAurora I don't know if you had design notes regarding this API?

First iteration was [Inspecio](https://github.com/Queerbric/Inspecio) and recently it became apparent it had the flaw of being made client-only so the non-sidedness of TooltipData wasn't apparent at the time of making the mod.

Having an interface to convert is less spreading out code as an event would though. Like, by reading the TooltipData I see clearly the conversion at the same time, while with an event, it's not as clear (which is one of my issues with the sleeping events, I need events to say that my sleeping bag can be used to sleep, which spreads out the code of the block and gets harder to keep track off).

> I do have plans to look at splitting the client and server into diffrent source sets so this pattern wont work going forward.

@modmuss50 I don't see why Fabric would need to split client and server into different source sets? Wouldn't this be a step back to pre-1.3 but with the exception that obfuscation mappings are unified? I really do not see the need for this.
And I do not see how it would prevent that interface to work if those are split?

**Comment by modmuss50** [comment]
>  I don't see why Fabric would need to split client and server into different source sets?

See: https://github.com/FabricMC/fabric-loom/issues/376 (Looking at this issue it might not be super clear what my plans are)

The idea to have compile time sided access saftey. The idea is to split the mc jar into 2 as well, so your server/common code wont be able to compile against any client side classes. This has been made a lot easier with the 1.17 changes, and it has been confirmed that mojang use a similar setup to develop the game.

It will require quite a huge amount of work (especially in loom), and im not 100% sure if its possible. If you have any input on this please put it on the loom issue so it doesnt get lost.

Even if that doesnt work out, in general keeping client only stuff in its own class/package is something that is highly encoraged and followed throughout fabric and fabric loader.

**Comment by LambdAurora** [comment]
> Even if that doesnt work out, in general keeping client only stuff in its own class/package is something that is highly encoraged and followed throughout fabric and fabric loader.

I agree with this, though I think having stuff spread out heavily in a big codebase is doing more harm than good (and gets very confusing).

Though, I do like the idea of ensuring sided access safety, different source sets often ends up being more annoying to use sadly.

I do wonder if we really have no other way to design an API for tooltip components without spreading it out too much.

**Comment by Technici4n** [comment]
> I do wonder if we really have no other way to design an API for tooltip components without spreading it out too much.

We can do a direct `ItemStack` -> `TooltipComponent` mapping, but you will still have to register an event handler.

**Comment by LambdAurora** [comment]
> We can do a direct `ItemStack` -> `TooltipComponent` mapping, but you will still have to register an event handler.

Bit worried about this one and the implications.

How would it allow to compound components?
How would this even be implemented? There's likely not enough context in the screen lambda.

**Comment by Technici4n** [comment]
> How would it allow to compound components?

Yes, combined components are very easy to implement afaict.
I was thinking of two events, both allowing adding to a `List<>`: one for "own" components, and one for adding components to foreign stacks. This could otherwise be a good use of event phases (#1669).

> How would this even be implemented? There's likely not enough context in the screen lambda.

We'd have to capture the item stack in a ThreadLocal. Not a big problem though.

**Comment by warjort** [comment]
> I agree with this, though I think having stuff spread out heavily in a big codebase is doing more harm than good (and gets very confusing).

It seems strange to me that to ensure something is checked at compile time (whether you are loading client classes in the wrong place), it is replaced with something else that can't be checked at compile time (e.g. whether you registered your event properly).

This just moves the problem adding extra complications and runtime overhead in the process.

**Comment by Technici4n** [comment]
What? The event is client-side only, you can't call it from common code.
Something has to map `ItemStack` (a common class) to `TooltipComponent` (a client-only class). Clearly that something must be at least somewhat client-side only.

**Comment by warjort** [comment]
> What? The event is client-side only, you can't call it from common code.
Something has to map ItemStack (a common class) to TooltipComponent (a client-only class). Clearly that something must be at least somewhat client-side only.

I am not disputing that, it is correct given your assumptions. 
I am questioning your assumptions.

You don't like the ````@Environment(CLIENT)```` because you want to check the client/server distinction at compile time and don't trust mod developers to do it correctly.
The solution of using an event, introduces something else that needs to be tested.
i.e. 
* whether you registered your event correctly
* whether it matches the correct data, 
* whether it doesn't break in the presence of other people's data, etc. 
(it can't be validated at compile time).

My question is what has changed? 
Besides introducing more complications? 
Not to mention that linear search over all listeners everytime you want to convert a TooltipData to a component.

**Review by Mysterious-Dev — APPROVED** [review]

**Review by liach — APPROVED** [review]

**Review by Player3324 — APPROVED** [review]

**Review by modmuss50 — APPROVED** [review]

**Review by shedaniel — APPROVED** [review]

**Review by ghost — COMMENTED** [review]
I thought mod code shouldn't be using `@Environment(EnvType.CLIENT)`

**Review by LambdAurora — APPROVED** [review]

**Comment by LambdAurora** [comment]
> I thought mod code shouldn't be using `@Environment(EnvType.CLIENT)`

Since when?
They can use it and code referencing it will get stripped at runtime.
Not using it on client-only code is considered bad practice.

**Comment by ghost** [comment]
Then why aren't all of fabric's client only events & classes annotated like this... 🤔

**Comment by modmuss50** [comment]
> I thought mod code shouldn't be using @Environment(EnvType.CLIENT)

There is generally no reason to, fabric api/libraries are a little diffrent as it will provide an extra line of defence in case someone uses it accidentally. Arguably its best used as a form of documenation. In a normal mod there is really no need, so I generally dont bother to use it.

> Not using it on client-only code is considered bad practice.

No, I dont think thats true. A good practice is to make sure that your client and server code is split and not mixed in the same classes thus removing the need to use the annoation.

I do have plans to help enfonce this by splitting them into diffrent source sets, but we are a little way off that.

**Comment by LambdAurora** [comment]
> > Not using it on client-only code is considered bad practice.
> 
> No, I dont think thats true. A good practice is to make sure that your client and server code is split and not mixed in the same classes thus removing the need to use the annoation.
> 
> I do have plans to help enfonce this by splitting them into diffrent source sets, but we are a little way off that.

Even if split, I was always told that not annotating with environment annotations is bad practice.

I am still worried about the implications of different source sets, especially in cases where not being allowed to reference a client-only stuff in a common class would result in overly complex and inefficient data structures and algorithms.

**Comment by modmuss50** [comment]
> I am still worried about the implications of different source sets, especially in cases where not being allowed to reference a client-only stuff in a common class would result in overly complex and inefficient data structures and algorithms.

We will have to see, I want to prototype it first to see how it will work, this discussion is for elsewhere (and when I have the energy to rip loom into 2) though 👍

**Comment by liach** [comment]
> I am still worried about the implications of different source sets, especially in cases where not being allowed to reference a client-only stuff in a common class would result in overly complex and inefficient data structures and algorithms.

It won't. You are always supposed to handle client-specific stuff like models as an extension (like vanilla does). Also these are more like business logic than true data structures or algorithms.

File: fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/TooltipComponentCallback.java [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+			if (component != null) {
+				return component;
+			}
+		}
+
+		return null;
+	});
+
+	/**
+	 * Return the tooltip component for the passed data, or null if none is available.
```
Juuxel, line 53: This could also use the inline return added in Java 16, which ["provides content for the first sentence of a method's description, and a 'Returns' section"](https://docs.oracle.com/en/java/javase/16/docs/specs/javadoc/doc-comment-spec.html) (instead of just the desc). But this is fine too.
```suggestion
	 * {@return the tooltip component for the passed data, or null if none is available}
```

Juuxel (approved):

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/TooltipComponentCallback.java
LambdAurora: ~~Hope Intellij will start displaying this correctly soon.~~

Juuxel: It seems to have been fixed in a 2021.1 patch release: https://youtrack.jetbrains.com/issue/IDEA-258058

I think it doesn't still render it quite the way the javadoc tool documentation specifies, but doesn't ignore it anymore either.

liach: @Technici4n Worth it to make this addition?

Technici4n: Don't know, personally I don't particularly like two lines of javadoc saying exactly the same thing.

liach: Hmm, we might need to refer to a common javadoc convention at some point. We currently use this tag extensively in yarn; oracle recommends a return tag even if it means there is no body comment; google style recommends what you do. Anyways we can fix that later.

PR #1699, opened 2021-09-02
Add methods to the FabricBlockEntityTypeBuilder class for adding additional blocks

haykam821:


liach (approved):
Why not, this is a builder anyway

LambdAurora (approved):
Nice.

Though, I personally would prefer to be able to add supported blocks after registration since I have a case where I register new blocks that depend on Registry analysis (which means I never know when everything actually ended to register and let me register my BE type).

ghost:
IMO `List<Block>` should rather be a `Set<Block>`, not that it matters atm but it will be more resiliant to mojang changes, especially if they change `ImmutableSet.copyOf` to `Set.of`.

Technici4n (changes requested):
Testmod please.

liach:
> Testmod please.

If we do add a lot of test mods, this is going to be unmaintainable; unlike unit tests in say junit or so that can be run independently and with selection, test mods aren't, and the judgement of mod behavior is done by human than machine, which further decreases the usefulness of these mods. In addition, this change is clearly small, adding a test mod for this is unrealistic. The better approach is to change one of the test mod block entity type builder to call the new methods instead.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
```diff
@@ -31,23 +35,36 @@
...
 
 	public static <T extends BlockEntity> FabricBlockEntityTypeBuilder<T> create(Factory<? extends T> factory, Block... blocks) {
-		return new FabricBlockEntityTypeBuilder<>(factory, blocks);
+		List<Block> blocksList = new ArrayList<>(blocks.length);
+		Collections.addAll(blocksList, blocks);
+
+		return new FabricBlockEntityTypeBuilder<>(factory, blocksList);
+	}
+
+	public FabricBlockEntityTypeBuilder<T> addBlock(Block block) {
```
Juuxel, line 58:
```suggestion
	/**
	 * Adds a supported block for the block entity type.
	 *
	 * @param block the supported block
	 * @return this builder
	 */
	public FabricBlockEntityTypeBuilder<T> addBlock(Block block) {
```
Hope the code style is correct, did this in the comment editor :smile:

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
```diff
@@ -31,23 +35,36 @@
...
+		this.blocks.add(block);
+		return this;
+	}
+
+	public FabricBlockEntityTypeBuilder<T> addBlocks(Block... blocks) {
```
Juuxel, line 69:
```suggestion
	/**
	 * Adds supported blocks for the block entity type.
	 *
	 * @param blocks the supported blocks
	 * @return this builder
	 */
	public FabricBlockEntityTypeBuilder<T> addBlocks(Block... blocks) {
```

Juuxel (changes requested):
Javadoc please (see suggestions :wink:)

Technici4n:
> The better approach is to change one of the test mod block entity type builder to call the new methods instead.

Yes that's what I was implicitly suggesting. The second commit in particular was added after some review on discord, which could have been avoided by proper testing.

Technici4n:
Will reopen when a testmod is changed to ensure this works as expected.

liach:
We have in-progress and outdated labels. Use them.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
```diff
@@ -31,23 +35,36 @@
...
+		return this;
 	}
 
 	public BlockEntityType<T> build() {
 		return build(null);
 	}
 
 	public BlockEntityType<T> build(Type<?> type) {
-		return BlockEntityType.Builder.<T>create(factory::create, blocks)
+		return BlockEntityType.Builder.<T>create(factory::create, blocks.toArray(new Block[0]))
```
PepperCode1, line 79: Since this is for 1.17, could the new generator accepting method (added in Java 11) be used? `new Block[0]` would become `Block[]::new`.

haykam821:
@Technici4n The test mod now adds a few blocks that are expected to have block entities, which can be tested by clicking them:

Betrayal blocks with success messages in chat

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
haykam821: Fixed in d6051ccc02983ec66cfd541b054748091e3419d0

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
haykam821: Fixed in ca96055b02d12a3c1a428fc12098d3e6f79d98fb

modmuss50 (approved):
Easy enough, I like the test mod textures 👍

LambdAurora:
What happens if I add blocks needing a block entity, but I never know in advance the number of blocks added, thus I never know when to actually register the block entity type?

haykam821:
@LambdAurora This was mainly intended to clean up code that has definite knowledge of the blocks that should be added to the builder.

One of the following two solutions could be employed:

- Allow builders to contain custom predicates for blocks in addition to just a simple set of blocks in a separate pull request
- Use a mixin to modify the block entity type's `blocks` set or change the return value of `supports`.

Juuxel:
FYI you can just extend `BlockEntityType` to provide your own logic ([see Adorn for example](https://github.com/Juuxel/Adorn/blob/b3b786a198455878bb173a6adf01e2e8b0d3e528/common/src/main/kotlin/juuxel/adorn/block/entity/AdornBlockEntityType.kt#L13)) instead of mixins to `supports`.

Player3324 (approved):

modmuss50 (approved):

PR #1700, opened 2021-09-02
Transfer API: Work around vanilla capturing ItemStack references

Technici4n:
Ensure that Inventory wrappers will try to mutate the backing stack as much as possible. In many cases, MC code captures a reference to the ItemStack so we want to edit that stack directly and not a copy whenever we can. Obviously this can't be perfect, but we try to cover as many cases as possible.

This is the kind of change that makes me happy I wrote tests. :smile:


modmuss50 (approved):

Technici4n:
I confirm that this fixes an issue I've been having in an `ItemUseCallback` in Modern Industrialization.

Player3324 (approved):

PR #1726, opened 2021-09-16
Confusing code paths in Keybinding API

isXander:
Keybinding registry's exceptions and returns have been overcomplicated and confused.
**What's Changed?**
```java
if (existingKeyBindings == binding) {
	throw null;
}
``` 
throwing null is not a shorthand but a de-reference of null, causing a `NullPointerException` as a result. This abstracts the exception making. It has been replaced with an `IllegalArgumentException` with a suitable description.
```java
... if (existingKeyBindings.getTranslationKey().equals(binding.getTranslationKey())) {
	throw new RuntimeException("...");
}
``` 
Throwing a `RuntimeException` is very bad because it makes catching fabric specific exceptions difficult. It has instead been replaced with `IllegalArgumentException`. This makes more sense because... it's an illegal argument. You can't register two keybindings with an equal ID.
```java
return moddedKeyBindings.add(binding) ? binding : null;
```
this is completely useless. `add()` always returns true.

### DIFF
--- a/fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
+++ b/fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
@@ -47,6 +47,16 @@ public final class KeyBindingHelper {
 	}
 
 	/**
+	 * Returns if the given keybind has been registered
+	 *
+	 * @param keyBinding the keybinding
+	 * @return if the keybind has been registered
+	 */
+	public static boolean keybindingExists(KeyBinding keyBinding) {
+		return KeyBindingRegistryImpl.doesKeybindingExist(keyBinding);
+	}
+
+	/**
 	 * Returns the configured KeyCode bound to the KeyBinding from the player's settings.
 	 *
 	 * @param keyBinding the keybinding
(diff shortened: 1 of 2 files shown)


frqnny:
Might be a good idea to use the new method in the testmod? Would help us to check if keybind registering worked.

warjort:
> It has instead been replaced with IllegalArgumentException. This makes more sense because... it's an illegal argument. 

There's nothing wrong with the argument. What's wrong is its already registered. It's an IllegalStateException.

fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
```diff
@@ -46,6 +46,16 @@ public static KeyBinding registerKeyBinding(KeyBinding keyBinding) {
 		return KeyBindingRegistryImpl.registerKeyBinding(keyBinding);
 	}
 
+	/**
+	 * Returns if the given keybind has been registered
+	 *
+	 * @param keyBinding the keybinding
+	 * @return if the keybind has been registered
+	 */
+	public static boolean keybindingExists(KeyBinding keyBinding) {
```
liach, line 55: Why do you need this method? In general, you should just choose a specific key translation key that includes your mod id to avoid conflicts.

Technici4n:
The impl cleanup is fine, but why do you need `keybindingExists`? (liach's comment still stands)

fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
isXander: it was an edge case in one of my mods,
my mod would dynamically add keybinds like Action 1, Action 2 and it was a good way to detect what action number to add next

if it needs to be removed thats fine

Technici4n:
Closing due to no answer in a long time.

PR #1730, opened 2021-09-17
Structure API fixes [1.18]

shartte:
These fixes allow the testmod to go ingame for the structure-api and the test structure will spawn correctly in a superflat world (which the changes relate to).

Structures are no longer linked to Biomes via an instance field  in the Biome class. Instead they are linked by registry keys. This means that standard biome structure start rules also apply to superflat worlds, and their biomes (default=PLAINS).

The superflat chunk generator now has its own StructuresConfig object, which only contains a hardcoded list of structures.
Changes the FabricStructureBuilder to allow adding structures to this default list.

### DIFF
--- a/fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructureBuilder.java
+++ b/fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructureBuilder.java
@@ -20,16 +20,14 @@ import java.util.Objects;
 
 import com.google.common.collect.ImmutableList;
 
-import net.minecraft.world.gen.GenerationStep;
 import net.minecraft.util.Identifier;
+import net.minecraft.world.gen.GenerationStep;
 import net.minecraft.world.gen.chunk.StructureConfig;
 import net.minecraft.world.gen.chunk.StructuresConfig;
-import net.minecraft.world.gen.feature.ConfiguredStructureFeature;
 import net.minecraft.world.gen.feature.FeatureConfig;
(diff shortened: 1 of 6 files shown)


modmuss50 (approved):
Minor changes to the api, seems fine to me.

fabric-structure-api-v1/src/main/java/net/fabricmc/fabric/api/structure/v1/FabricStructureBuilder.java
```diff
@@ -113,39 +111,25 @@ private FabricStructureBuilder(Identifier id, S structure) {
...
-	 * superflat worlds.
-	 *
-	 * @see #superflatFeature(FeatureConfig)
+	 * Enables generation of this feature in superflat worlds. Structures will still only generate in biomes
+	 * for which they have been enabled. Superflat worlds will use the plains biome by default.
 	 */
-	public FabricStructureBuilder<FC, S> superflatFeature(ConfiguredStructureFeature<FC, ? extends StructureFeature<FC>> superflatFeature) {
-		Objects.requireNonNull(superflatFeature, "superflatFeature must not be null");
-		this.superflatFeature = superflatFeature;
+	public FabricStructureBuilder<FC, S> enableSuperflat() {
```
ghost, line 128: Imo I would just name this the same as the field name `generateInSuperflat` or rather `generatesInSuperflat`

modmuss50: I want to get this in ASAP so I think im just going to put experimental on this method for now so we can change it later if needed 👍

PR #1732, opened 2021-09-18
Biome API without Overworld for now [1.18]

shartte:
The biome-api module is actually composed of four relatively independent components:
- Biome Modification
- Modifying the default Overworld biome source
- Modifying the default Nether biome source
- Modifying the default End biome source

The PR removes the overworld component while restoring the others. Overworld has changed significantly and now uses the same basic generation approach as the nether. It's quite uncomfortable to add biomes via noise points however, and there probably needs to be a new API for modifying how overworld biomes are generated. There's no reason to keep the rest inactive, however.

Changes to the other bits:
- Biome depth and scale are gone since those related to generation, which has changed in Vanilla
- Structure starts are no longer defined directly in a Biome, which means that biome modification's addStructure, hasStructure, removeStructure do not translate 1:1. The information is now stored in the ChunkGeneratorSettings which is more powerful than before (i.e. PLAINS as part of Overworld can be treated differently from PLAINS as part of a different chunk generator settings). To keep compatibility with the old API, adding/removing a structure start from a biome will operate on all registered chunk generator settings.
- Modifying Nether and End biomes is functionally unchanged
- I took the opportunity to split up the InternalBiomeData godclass into the separate pieces for Nether, End and Overworld biome modification since they're not really related.
- Vanilla now prevents biomes with different generation order for the same feature-type from being used together. This breaks the use of BEACH with any nether biomes in the test mod. I switched this to using THE_END for the demo since the test just demonstrates that a foreign biome now generates in the nether.

Future improvements:

- Overworld API, obviously. There is potential for a low-level API that is 1:1 the same a the Nether biome API (since Overworld is now also multi-noise based). But in addition, since Vanilla uses a higher-level "generator" to generate the multi-noise points for its biomes, we can also provide an API to hook mods into this generator's output, providing a more abstract way of modifying biomes in the Overworld. (The generator still deals in humidity, temperature, mountains, etc.)
- The End uses a simple perlin noise source, but since it now has access to the multi noise source (even if Vanilla ignores it for the end), the noise could be sampled for weirdness to decide which modded biome gets used. This is best left to actual worldgen experts to decide, however.
- Biome modification could allow structures to be added to biomes only for specific chunk generation settings now.


liach (approved):
Api side looks good

Player3324 (approved):
API changes look fine, thanks!

modmuss50 (approved):

PR #1755, opened 2021-10-02
Add two more sleep events

Juuxel:
These are the last things I found when porting Adorn's sofas over to sleep events.

- **This PR de-hardcodes the usual wake up code path from bed blocks to any block matching `ALLOW_BED`.** The two events below are needed for it to work correctly.
  - The code handles player positioning/rotation and occupied marking for you, and it would be a pain to reimplement this in mods.
- `SET_BED_OCCUPATION_STATE` event: this marks the bed as occupied or unoccupied
  - This is needed if the block uses some other method instead of the `BedBlock.OCCUPIED` property
  - The vanilla check is also de-hardcoded from `instance BedBlock` to `BlockState.contains(BedBlock.OCCUPIED)`
- `MODIFY_WAKE_UP_POSITION` event: self-explanatory
  - Works like `MODIFY_SLEEPING_DIRECTION`, so modifies an existing value or provides a new value
  - Needed for blocks where `BedBlock.findWakeUpPosition` doesn't work

### DIFF
--- a/fabric-entity-events-v1/build.gradle
+++ b/fabric-entity-events-v1/build.gradle
@@ -1,5 +1,5 @@
 archivesBaseName = "fabric-entity-events-v1"
-version = getSubprojectVersion(project, "1.2.3")
+version = getSubprojectVersion(project, "1.3.0")
 
 moduleDependencies(project, [
 		'fabric-api-base'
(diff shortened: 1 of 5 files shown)


LambdAurora (commented):
Please put literals into `{@code}` in javadocs and add `@Dynamic` annotations on top of injectors to lambdas.
Else it's all good!

Juuxel:
@LambdAurora I'll add the `{@code}` (though the rest of the class doesn't use it either so I'll add it to everything).

As for the `@Dynamic`, it's intended "for mixin elements whose targets are not available in the original class and are either fabricated or transformed at runtime" (quote from mixin docs), none of which applies to lambda bodies.

LambdAurora:
> As for the `@Dynamic`, it's intended "for mixin elements whose targets are not available in the original class and are either fabricated or transformed at runtime" (quote from mixin docs), none of which applies to lambda bodies.

After a lot of discussion about that in various spaces, `@Dynamic` is also intended for lambdas, and it will suppress all warnings in IDE and at compilation about it not being able to find said method.

Juuxel:
Ah, I'll add it then.

LambdAurora (approved):

liach (approved):
The pr aims to show that non-BedBlock beds now work without crashing. If the test can verify this (still only using existing BedBlock based bed), it would be helpful.

Juuxel:
@liach `TestBedBlock` is not a `BedBlock` (albeit the name can be a bit confusing here), so it verifies non-`BedBlock`s too.

Technici4n (approved):

modmuss50 (approved):

Player3324 (approved):

PR #1756, opened 2021-10-02
Update MixinBlockEntityRenderers.java

Domaman202:


liach (approved):
Sure

Technici4n:
I see no point in this. Better to do it in a related bugfix or feature release, or even not at all.

That was probably not the intention, but I tend to interpret these PRs as trying to get the contributor role in a big open source project. I would really recommend that they be avoided.

PR #1759, opened 2021-10-03
Fix handling of orphaned registry entries

BasiqueEvangelist:
Right now, Fabric API overwrites the IDs of any orphaned registry entries with 0. This is because the code tries to get the id from the registry tag, which it just checked that it doesn't have it.
This PR fixes this, restoring the intended behavior of the code.


modmuss50 (approved):
Nice Spot!

modmuss50:
This is also an issue in the none dynamic registry manager: <https://github.com/FabricMC/fabric/blob/1.17/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java#L227>

BasiqueEvangelist:
> This is also an issue in the none dynamic registry manager: https://github.com/FabricMC/fabric/blob/1.17/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java#L227

Fixed.

liach (approved):
:heart:

PR #1762, opened 2021-10-05
Transfer API: bugfixes and improvements

Technici4n:
- Thanks to @virusek20 who spotted the `SingleVariantItemStorage` bug.
- The change to inventory wrappers works around more vanilla cases. The original motivation for this change is that mining with a powered tool would otherwise cancel the mining if the energy of the tool changes through the transfer API during the mining operation. This is because vanilla captures a reference to the original stack, and if it becomes empty the mining operation is cancelled.
- I also used this opportunity to add filtering to a few `StorageUtil` methods.
- Overrode `PlayerInventoryStorage#insert` to avoid the surprising behavior of blindly inserting into armor and offhand slots:
image


liach (approved):
Looks good on glance. Remember to bump api version

Player3324 (approved):

Juuxel (approved):

modmuss50 (approved):

PR #1777, opened 2021-10-18
Fix #1772

### LINKED ISSUE #1772: KeyBindingHelper removes conflicting vanilla keybinds
If a modded key uses the same keycode as a vanilla keybinding, the vanilla keybinding is removed. This happens because of line 80 in KeyBindingRegistryImpl, which tries to remove previous modded keybinds before readding them. However KeyBinding has equals() overridden, it is true whenever the boundKey is equivalent, this happens whenever the InputUtil.Type and keycode are the same. This means that any vanilla KeyBinding disappears if  a modded keybinding has the same code and type.

Stuff-Stuffs:
This appears to fix #1772  by replacing an ArrayList with an ReferenceArrayList, this replicates the pre 1.17 behaviour of remove duplicate keys based on reference equality.

### DIFF
--- a/fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/impl/client/keybinding/KeyBindingRegistryImpl.java
+++ b/fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/impl/client/keybinding/KeyBindingRegistryImpl.java
@@ -32,7 +32,7 @@ import net.fabricmc.fabric.mixin.client.keybinding.KeyBindingAccessor;
 public final class KeyBindingRegistryImpl {
 	private static final Logger LOGGER = LogManager.getLogger();
 
-	private static final List<KeyBinding> moddedKeyBindings = new ReferenceArrayList<>();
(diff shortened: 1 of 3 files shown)


Player3324:
This definitely needs to document why it is a ReferenceArrayList including what the behavior differences are

fabric-key-binding-api-v1/src/main/java/net/fabricmc/fabric/impl/client/keybinding/KeyBindingRegistryImpl.java
```diff
@@ -31,7 +32,7 @@
 public final class KeyBindingRegistryImpl {
 	private static final Logger LOGGER = LogManager.getLogger();
 
-	private static final List<KeyBinding> moddedKeyBindings = Lists.newArrayList();
+	private static final List<KeyBinding> moddedKeyBindings = new ReferenceArrayList<>();
```
Player3324:
```suggestion
	private static final List<KeyBinding> moddedKeyBindings = new ReferenceArrayList<>(); // ArrayList with identity based comparisons for contains/remove/indexOf etc., required for correctly handling duplicate keybinds
```

Player3324 (approved):
LGTM with the suggested or a similar documentation enhancement. Thanks for the analysis and fix, sorry for the delay..

modmuss50 (approved):

PR #1784, opened 2021-10-22
Full Fabric Fluid Api V1

ghost:
This pull request implements a new module "fabric-fluid-api-v1" that is focused on improving everything about the fluid creation, compared with vanilla one.

I've implemented all the #1136 pull request features, but in a simpler way, plus some additions about the fluid customizations.

## Steps

- [x] Ready-to-use basic abstract FlowableFluid class.
- [x] Ready-to-use basic FluidBlock class.
- [x] Custom fog.
- [x] Custom sounds.
- [x] Pushing strength.
- [x] Splash event.
- [x] Splash event: particles.
- [x] Submerged event.
- [x] Submerged event: entity drowning.
- [x] Touching event.
- [x] Touching event: entity on fire.
- [x] Specify if the fluid can light fire.
- [x] Handling hot and fire damage.
- [x] Handling fall damage.
- [x] Specify if the fluid can wet entities.
- [x] Handling drowning.
- [x] Boats over the fluid.
- [x] Entity swimming.
- [x] Fix some small things...
- [x] Check if everything is ok, and avoid any potential compatibility breaks.
- [x] Cleanup.
- [x] Updated to 1.18. (Checkout #1874)
- [x] Backported to 1.16. (Checkout #1879)

## Additions

* `FabricFlowableFluid` class that extends `FlowableFluid`, implementing some common fluid customizations, and adding new customizations, such as fog, particles, sounds, viscosity, events...
* `FabricFluidTags` class that adds new tags for fluid customizations.
* `FabricFluidBlock` class that extends `FluidBlock`, adding a public constructor, so you don't need anymore to create a new class.

## Extra

* `FluidUtils` class that contains some utilities for fluids.
* `EntityFluidExtensions` interface that implements some fluid-related entity features.
* `LivingEntityFluidExtensions` interface that implements some fluid-related living entity features.
* `ClientPlayerEntityFluidExtensions` interface that implements some fluid-related client-side player entity features.
* `CameraFluidExtensions` interface that implements some fluid-related camera features.
* `SoundParameters` class that incapsulates some sound parameters.
* `UnderfluidSoundLoop` class that implements a sound loop player used when the player is submerged by a fluid.

## Note

* All acquatic creatures cannot swim on custom fluids, even if the fluid is tagged as swimmable.
  The tag is meant for land creatures (like the player and all other land mobs, chicken, wolves...).
  For acquatic creatures the swimming activity (how to swim in custom fluids) must be injected in their brain, so maybe this is a job for mod developers, and not for fabric-api...


ghost:
Little demonstration:

<video src="https://user-images.githubusercontent.com/32025549/138524002-fc2d8428-9784-4d38-8713-b7986be5cc2e.mp4" height="300" />

ghost:
Now boats can travel over custom fluids!  
Here a comparison between **navigable** vs **non-navigable** fluid.

In a **non-navigable** fluid the boat will sink.

Shnupbups:
I feel like perhaps the boolean parts of `ExtendedFlowableFluid`, (that is, `canExinguishFire`, `canLightFire`, `canPreventFallDamage`, `canWet`, `isNavigable`, and `isSwimmable`,) should be Fluid Tags? Then the default implementation of the methods can be to check whether the relevant tags contain `this`.

ghost:
> I feel like perhaps the boolean parts of `ExtendedFlowableFluid`, (that is, `canExinguishFire`, `canLightFire`, `canPreventFallDamage`, `canWet`, `isNavigable`, and `isSwimmable`,) should be Fluid Tags? Then the default implementation of the methods can be to check whether the relevant tags contain `this`.

Done!
Simplified a lot of things.

I wonder how to make the tag `fabric_fluid` mandatory, and throw an exception if is not used, I've already tried to check this on the `FabricFlowableFluid` constructor, but the tag was not already applied to the fluids...

``` java
public FabricFlowableFluid() {
  if (!this.isIn(FabricFluidTags.FABRIC)) {
    throw new RuntimeException("All FabricFlowableFluid must be tagged with " + FabricFluidTags.FABRIC);
  }
}
```

haykam821:
I would recommend adding a check for both `fluid instanceof FabricFlowableFluid` and `fluid.isIn(FabricFluidTags.FABRIC_FLUID_X)` for relevant properties, like how vanilla tags behave. There should not be a tag for all Fabric fluids.

ghost:
Ok the "Fluid Update" is ready. 👍
So this is finally removed from draft and is ready for review.

However, following the footsteps of the biome-modification api, for now, I kept the "experimental" status as a form of precaution, because as a new module it may contains potential undescovered bugs...

PepperCode1:
Does this PR allow creating fluids with an amount of levels other than 8? I see there is `getMaxLevel`, but it doesn't seem to do much because the default `LEVEL` property is still used which always has 8 levels.

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/render/FluidRenderer.java
```diff
@@ -0,0 +1,128 @@
...
+import net.minecraft.util.registry.Registry;
+import net.minecraft.world.BlockRenderView;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.Function;
+
+/**
+ * Provides basic fluid rendering.
+ */
+public class FluidRenderer {
```
PepperCode1: The name of this class is confusing for me. From first glance and without reading the other explanations, I thought this class and its methods were to render a fluid block in the world. To avoid this confusion and a name conflict with vanilla's `FluidRenderer`, I think this class should be renamed to something like `SimpleFluidRenderRegistry`. The methods should be renamed from `render` to `register`, `registerRenderer`, `assignRenderer`, etc.

Edit: I feel like this class doesn't belong in this module and should instead be in `rendering-fluids-v1`. However, that module already has `SimpleFluidRenderHandler`, which has the same functionality as this class, so I don't see the point of keeping this class.

AlphaMode:
Perhaps rename this to FLAMMABLE instead of FIRE_LIGHTER which feels a bit dull
https://github.com/devpelux/fabric/blob/ef4d06981210aee849f3777eaadbc3d206710fdf/fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/tag/FabricFluidTags.java#L38

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/render/FluidRenderer.java
ghost: Ok, thanks for letting me know this.
The problem was that I was 17 commits before and there was no SimpleFluidRenderHandler class.
So I've simply removed this duplicate class.

ghost:
- [x] ~~There seems to be a compilation problem with the most recent updates...~~ Fixed!

modmuss50:
> * [ ]  There seems to be a compilation problem with the most recent updates...

Looks like checkstyle failiures.

modmuss50:
Clearly you have put a huge amount of work into this and I don't want for this to come across as discrediting your work, I can assure you that is not the case! However I don't think this PR in its current state is a great fit to become an official fabric API module.

This PR is quite overwhelming to review and seems to have quite a lot of invasive Mixin's with a huge amount of logic and magic numbers that will be very hard or even impossible to main for years going forward.

One of my worries is changing vanilla behaviour, fabric is used a lot by the technical community who will notice even the smallest change to vanilla behaviour, assuming all of the logic is correct now in this PR, its going to be very hard for someone to update this logic to match newer versions of Minecraft.

I think it would be best to try and split this PR up into more manageable bite sized chunks. Figure out what will be the most useful and what can be done without large invasive changes. This is not to say we don't want all of the changes eventually but taking a slower approach to rolling them out will help keep the risks of their being an issue a lot lower.

It may be a good idea to create your own library that other mods can depend on to get an idea of how well the API and changes in this PR work. This would give you a lot more flexibility to make changes and improvements and then over time integrate the changes into Fabric API. 

I look forward to have a discussion with you to try and figure out a path forward where this can work as an official fabric API module. Thanks a lot for understanding.

ghost:
Ok, this is now closed, because a completed and huge rework is needed, and it would mean data loss and work loss, so is better to restart by zero, in my opinion, and leave this here only for legacy documentation.

The initial goal was to handle fog and pushing, in fact I had also added this as a guide in the wiki, and that was pretty simple, only two classes were edited (Entity and BackgroundRenderer), and almost no editing on existing methods, only adding new ones, (and the setting blocks and entities on fire is handled by the overriding of a method inherited from FlowableFluid: [onRandomClick](https://github.com/devpelux/fabric/blob/f6d46d1c41cea40654872b28d39cf829be6fe1f8/fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java#L250)).

The 90% of this api is covered by managing swimming and breathing on custom fluids, and day by day I also started to think that it was too much, and it was becoming more like a mod than an api, this is the reason why I've added the warning note about the water creatures yesterday...

The plan is this:

**First:** the most simple thing: the fog. Will be added inside the already existing rendering-fluids module because is clearly related to fluid rendering in client side.

**Second:** the basic behaviour, like lit-on-fire, and fluid pushing strength will be added inside the module fluid-api, plus some fluid events, all of these inside FabricFlowableFluid.

**Third:** Everything covering swimming, breathing, and other things, will never be added in fabric-api (not from me at least), but in a future external library, maybe called "FluidExtensionsLib", that anyone, that want to customize these things, could use, this will create a sort of on/off behaviour.

florensie:
I actually already have a library mod that influences large parts of the vanilla swimming behavior, and it seems like the mixins in this API would have massively conflicted with [mine](https://github.com/florensie/ExpandAbility/tree/1.17/common/src/main/java/be/florens/expandability/mixin/swimming). I think for something like this to succeed as an API it would have to separate these big changes into multiple components: rendering, swimming, floating (boats) and possibly colliding (strider).

This would allow finer control over separate parts, like in my case where I want to enable only the swimming feature regardless of what fluid the player is or isn't in. You could then offer some kind of interface that integrates all these components for people who want to create straightforward fluids.

Either way I think this would have to be a standalone library mod, as it would be way too complex to properly maintain in Fabric API. I think there is a good reason that even Forge, the API known for making huge changes and big bloated APIs doesn't even have a solution for this after all these years of its existence.

PR #860, opened 2020-06-30
Add item NBT update animation handler API

Draylar:
This PR is a solution to [issue 856](https://github.com/FabricMC/fabric/issues/856). I would recommend reading the issue for context.

**Design**
i509VCB and I spoke a little bit about what route to take on the API. I originally favored keeping it as an event, mostly so multiple people could have a say in whether an item updated, but that issue seems like a small edge case, so I rolled with an interface instead. Keeping it as an interface also makes this API simpler and smaller. 

When creating an Item class, you can implement `UpdateAnimationHandler` to take control of whether or not the animation update plays every tick (not sure if that's game or render tick, think it's the latter). 

```java
public class PatchedUpdatingItem extends UpdatingItem implements UpdateAnimationHandler {

	public PatchedUpdatingItem(Settings settings) {
		super(settings);
	}

       @Override
	public void inventoryTick(ItemStack stack, World world, Entity entity, int slot, boolean selected) {
		float currentTicks = stack.getOrCreateTag().getFloat("ticks");
		stack.getOrCreateTag().putFloat("ticks", currentTicks + 1);
		super.inventoryTick(stack, world, entity, slot, selected);
	}

	@Override
	public boolean shouldRunAnimationUpdate(ItemStack original, ItemStack updated) {
		return false;
	}
}
```

This code, without the interface, would normally cause the item to spaz out when held in your hand. This is because `ItemStack.areEqual(this.mainHand, itemStack)` returns false, which causes the update animation to play. I counteract this in Inmis by overriding `ItemStack#equals` for my backpack items, which is similar to how this solution works. In this API, I redirect both `areEqual` methods in `HeldItemRenderer#updateHeldItems` (for main-hand & off-hand) and then ask the item if it should update. Note that:
  - The interface method is only called if *both stacks* are the same. 
  - If both items are not the same or they don't implement the interface, the default behavior is used.
  
**Improvements**
The one thing I am still considering is whether or not the method needs more *context*. It might be appropriate to give the method information on whether the item is in the main-hand or off-hand. 

I'm also not *entirely* sold on the method name of `shouldRunAnimationUpdate` because it feels a little bit bulky. I intentionally reversed the logic (boolean is flipped in `HeldItemRendererMixin`) so that the method makes more sense for implementers: *normally* the method would be returning true for equality, which is the opposite of returning true for whether the animation update should play. Having it flipped so the conditions are equal, eg. `shouldSkipAnimationUpdate`, seems weird, if that makes sense.

**Tests**
[...]

PR #1790, opened 2021-10-26
Allow canceling the NBT update animation and block break reset

### LINKED ISSUE #856: Hooks for ItemStack NBT update animation
### Issue

When an `ItemStack`'s NBT updates, a quick "reload" animation is played (which involves the item dipping down and then pulling back up). This [quickly becomes an issue](https://streamable.com/wdpmpl) when you have items that reload NBT *quickly*, like a tick counter item:
```java
Registry.register(
        Registry.ITEM,
        new Identifier("nbtissue", "testitem"),
        new Item(new Item.Settings()) {
            @Override
            public void inventoryTick(ItemStack stack, World world, Entity entity, int slot, boolean selected) {
                float currentTicks = stack.getOrCreateTag().getFloat("ticks");
                stack.getOrCreateTag().putFloat("ticks", currentTicks + 1);
                super.inventoryTick(stack, world, entity, slot, selected);
            }
        }
);
```
([result](https://streamable.com/wdpmpl))

### Context

`HeldItemRenderer#updateHeldItems` is where this animation update takes place. Forge has a hook here that allows you to define whether an item should cause said re-equip animation, and I think providing something similar in the API would be a good idea. 

### Current Solutions

In [Inmis](https://github.com/Draylar/inmis), I [use an entirely cursed solution for getting around this issue](https://github.com/Draylar/inmis/blob/1.16/src/main/java/draylar/inmis/mixin/ItemStackMixin.java). This will, obviously, fail as soon as 2 people try to do it (I claimed it first!!!!). I'm writing this issue up now because several other people are encountering this issue and need a solution.

### Potential Solutions

Forge provides a method in their Item class that allows you to define whether the re-equip animation should occur between stack A and stack B. We could do something similar in usage, like:

```java
FabricItemUpdateAnimations.register(Items.MY_ITEM, (originalStack, newStack) -> {
        return [...]
});
```

An event might also make sense (so multiple people can have a say in whether a diamond sword triggers the reload animation):

```java
ItemUpdateAnimationCallback.event(Items.DIAMOND_SWORD).register((originalStack, newStack) -> {
        return ActionResult.FAIL;
});
```

Technici4n:
Introduce the `FabricItem` interface. In the future the idea is to make it visible on the `Item` class in dev (FabricMC/fabric-loom#496).
- `allowNbtUpdateAnimation` allows cancelling the NBT/count update animation for held items.
- `allowContinuingBlockBreaking` allows continuing block breaking even if the NBT changes.

Supersedes #860. Closes #856.


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/HeldItemRendererMixin.java
```diff
@@ -0,0 +1,65 @@
...
+
+import net.fabricmc.fabric.api.client.item.v1.AllowNbtUpdateAnimation;
+
+@Mixin(HeldItemRenderer.class)
+public class HeldItemRendererMixin {
+	@Shadow
+	private ItemStack mainHand;
+
+	@Shadow
+	private ItemStack offHand;
```
Juuxel, line 42: Are these only used for the animation? :eyes:

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/AllowNbtUpdateAnimation.java
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Return true if this listener allows the update animation to run.
+	 *
+	 * @param player The current player. This may be safely cast to {@link ClientPlayerEntity} in client-only code.
+	 * @param hand The hand. This event applies both to the main hand and the off hand.
+	 * @param originalStack The previous stack.
+	 * @param updatedStack The new stack, of the same item as {@code originalStack}.
+	 * @return False to cancel the update animation, true to allow the animation (unless subsequent listeners cancel it).
+	 */
+	boolean allowNbtUpdateAnimation(PlayerEntity player, Hand hand, ItemStack originalStack, ItemStack updatedStack);
```
liach: I recommend some utilities to easily check the differences in two nbt structures. For instance, if the method wants to prevent if only the `TimeLeft` int nbt is updated, current api makes it quite hard to preclude other factors, such as checking if other parts of the stacks changed, if the nbt diff only limits to this `TimeLeft` nbt, etc.

Technici4n: I thought about using `TriState` instead of `boolean`. Thoughts?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/HeldItemRendererMixin.java
Technici4n: I'll check, but yeah I'm pretty sure. A bit weird, I agree.

Technici4n: Yeah, I'm pretty sure this is correct. These fields are the current rendered hands, so it makes sense to set them directly in case we want to cancel the animation.

Technici4n:
The documentation is a bit misleading, as the count could change while leaving the NBT untouched.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/AllowNbtUpdateAnimation.java
Technici4n: > I thought about using `TriState` instead of `boolean`. Thoughts?

I don't think that would help.

Technici4n: What you probably want is something like `areNbtEqualIgnoreSubtags("energy")`, right?

liach: Hmm, IMO we should allow each handler to register like an NBT subpath that it manages; if the content in that subpath changed, this handler will be queried. This way, if there are two non-updating sub tags from two different mods, one mod will not consider the other mod's sub-tag as a general tag and delegates back to vanilla "show update" result.

For instance, if one mod adds an item age (like to make player explode when holding an item too long) `ExplosionCounter` and another mod adds an `energy`, the explosion counter's handler has no way to know that `energy` is an ignored tag; hence it will continue to propagate unwanted updates.

So the API would be more like:
```java
public interface AllowNbtUpdateAnimation<T extends Tag> {
  static <T extends Tag> void register(Item item, NbtPath<T> /* something like that */ path, AllowNbtUpdateAnimation<T> tag) {}

  boolean allowNbtUpdateAnimation(PlayerEntity player, Hand hand, @Nullable T originalTag, @Nullable T updatedTag);
}
```

Technici4n: An easier way would be to short-circuit on `return true` instead of short-circuiting on `return false`.

Technici4n: Or at least, I think a general subpath handler is totally overkill. There is probably an easier way.

liach: no, then if both some regular and ignored tags are updated, your logic will incorrectly call for a skip.

Technici4n: I think that short-circuiting on `return true`, and having something like `return NbtUtil.areDifferentIgnoreSubtag("energy")` would already go a long way.

liach: but it's still wrong if you have two subtags from two mods to ignore

shartte: Interesting use case. Is that something that actually occurs in the wild?
In the sense that a mod that doesn't own the item changes the item's NBT and that should *not* cause the item update animation to play out? Is that really a realistic use case?

Technici4n: Good question. Using an event instead of an interface like `FabricItem` is also debatable...

Technici4n: Changed the approach to use `FabricItem`.

Technici4n:
I changed the approach to use a new interface called `FabricItem` where we can put general hooks that are only relevant for the current item (otherwise an event is better). The point is to expose it via interface injection in loom whenever it will be possible.

Juuxel:
Without itf injection (so currently), I'd rather add hooks into FabricItemSettings instead of magic interfaces - FabricItemSettings is a good place for this that already exists for similar hooks.

Technici4n:
I guess that would work too indeed, will update the PR in a bit.

Technici4n:
It's now an item setting.

Technici4n:
Back to `FabricItem` after some internal discussion! Also added a second method, which can also be useful for items that update their NBT frequently, and especially for tools that update their NBT while they are mining.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/ClientPlayerInteractionManagerMixin.java
```diff
@@ -0,0 +1,67 @@
...
+	@Final
+	private MinecraftClient client;
+	@Shadow
+	private BlockPos currentBreakingPos;
+	@Shadow
+	private ItemStack selectedStack;
+
+	/**
+	 * Rewrites the isCurrentlyBreaking logic, checking for the steam drill in the
+	 * process. This is done to avoid a more invasive mixin such as a @Redirect.
```
Technici4n: I should update this comment.

Player3324 (approved):

modmuss50 (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/client/ClientPlayerInteractionManagerMixin.java
Technici4n: Done. :)

shartte (approved):

Juuxel (approved):
The API looks good, though I still somewhat dislike magic interfaces without injection - but in any case, this works well too.

PR #1796, opened 2021-10-30
Fix features not generating when added via the biome modification api.

modmuss50:
Follows the vanilla behaviour of adding all the features to this set. Idk why it exists in the first place tbh.


shartte:
Was this seriously just renamed? It was previously a generation-step based list of lists.
I must have been blind before and it still uses a biome based approach then...

Issue #1760, opened 2021-10-03
Transfer API: future improvements


Technici4n:
A list of features for the transfer API, to keep tracks of things that should be added in the future. I will amend it when I find new things.

Planned:
- [ ] Item-containing items API: #1749.

Implemented in #2375: (Not merged yet)
- [x] Make `InsertionOnlyStorage` have no storage views by default.
- [x] Add some equivalent of `FluidUtil.interactTank`. Ideally that should play a sound and not consume the used item in creative.
- [x] Make FluidStorage. GENERAL_COMBINED_PROVIDER final.
- [x] Remove "A view is always tied to a specific transaction, and should not be accessed outside of it." from `StorageView` since it is now outdated.
- [x] Add `copyOrCreateNbt` method to `TransferVariant`.
- [x] Option to color water and lava, but only if requested by at least one mod.

Implemented in #2095:
- [x] Fluid "attributes": viscosity, temperature, light level, filling sound, emptying sound, etc... What should depend on `Fluid`, what should depend on `FluidState` or what should depend on `FluidVariant` is TBD.

Implemented in #2086:
- [x] `Transaction.getCurrentState()` which will return whether a transaction is open, closing, final closing or not existing. This is nicer than catching the ISE in `getCurrentUnsafe()`. 
- [x] Update the javadoc of `Storage#iterator` to say that the returned iterator is invalid after the transaction is closed, instead of forcing the iterator to immediately return false in `hasNext`. Also mention that callers are not allowed to call `remove` on the iterator.
- [x] Change `iterator()` and `iterable()` to use `? extends StorageView<T>` instead of `StorageView<T>` to make it nicer for implementing classes that just want to return a list iterator.
- [x] Definalize `SingleStackStorage#insert` & related methods since implementors may wish to override them.

Implemented in #1801:
- [x] Remove the notBlank check for `ContainerItemContext.withInitial` as it forced users to add a check themselves.
- [x] Clarify that `isResourceBlank()` is equivalent to `getResource().isBlank()` for transfer variants.
- [x] Add a few `FilteringStorage` static methods to wrap an existing storage, while blocking insertion or extraction entirely.
- [x] Deprecate for removal the fluid variant rendering methods that return a single sprite, and replace them with methods that return a `Sprite[]`. This will provide a standard way to retrieve a flowing fluid sprite for pipes, e.g. for use in pipes.
- [x] Add `insertStacking` to `StorageUtil`.
[...]

PR #1801, opened 2021-11-03
Another wave of transfer API improvements

Technici4n:
- ~~Add a water and lava `FluidVariantRenderHandler` to have cool-looking blue and red names (respectively).~~ People don't want it so it will go into my mods instead.
- [x] Remove the notBlank check for `ContainerItemContext.withInitial` as it forced users to add a check themselves.
- [x] Clarify that `isResourceBlank()` is equivalent to `getResource().isBlank()` for transfer variants.
- [x] Add a few `FilteringStorage` static methods to wrap an existing storage, while blocking insertion or extraction entirely.
- [x] Deprecate for removal the fluid variant rendering methods that return a single sprite, and replace them with methods that return a `Sprite[]`. This will provide a standard way to retrieve a flowing fluid sprite for pipes, e.g. for use in pipes.
- [x] Add `insertStacking` to `StorageUtil`.
- [x] Add a fallback to the sided storages that throws an exception if the world is not a server world to guard against incorrect API usage (client worlds are not allowed).

(most of #1760).

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRendering.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRendering.java
@@ -27,10 +27,7 @@ import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.item.TooltipContext;
 import net.minecraft.client.texture.Sprite;
 import net.minecraft.fluid.Fluid;
-import net.minecraft.fluid.Fluids;
 import net.minecraft.text.LiteralText;
-import net.minecraft.text.MutableText;
-import net.minecraft.text.Style;
 import net.minecraft.text.Text;
 import net.minecraft.util.Formatting;
 import net.minecraft.util.math.BlockPos;
@@ -53,22 +50,6 @@ import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariant;
 public class FluidVariantRendering {
 	private static final ApiProviderMap<Fluid, FluidVariantRenderHandler> HANDLERS = ApiProviderMap.create();
 	private static final FluidVariantRenderHandler DEFAULT_HANDLER = new FluidVariantRenderHandler() { };
-
-	static {
-		// Add colored names for vanilla fluids!
-		register(Fluids.WATER, new FluidVariantRenderHandler() {
-			@Override
-			public Text getName(FluidVariant fluidVariant) {
-				return ((MutableText) FluidVariantRenderHandler.super.getName(fluidVariant)).setStyle(Style.EMPTY.withColor(Formatting.BLUE));
-			}
-		});
-		register(Fluids.LAVA, new FluidVariantRenderHandler() {
-			@Override
-			public Text getName(FluidVariant fluidVariant) {
-				return ((MutableText) FluidVariantRenderHandler.super.getName(fluidVariant)).setStyle(Style.EMPTY.withColor(Formatting.RED));
-			}
-		});
-	}
 
 	/**
 	 * Register a render handler for the passed fluid.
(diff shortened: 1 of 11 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
```diff
@@ -40,7 +40,10 @@
 	long extract(T resource, long maxAmount, TransactionContext transaction);
 
 	/**
-	 * @return {@code true} if the {@link #getResource} contained in this storage view is blank, or {@code false} otherwise.
+	 * Return {@code true} if the {@link #getResource} contained in this storage view is blank, or {@code false} otherwise.
```
Juuxel, line 43: This'd be the perfect use case for a certain inline `@return`, but I feel it's not going to happen this time either 😄

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRendering.java
```diff
@@ -50,6 +54,22 @@
...
+			public Text getName(FluidVariant fluidVariant) {
+				return ((MutableText) FluidVariantRenderHandler.super.getName(fluidVariant)).setStyle(Style.EMPTY.withColor(Formatting.BLUE));
+			}
+		});
+		register(Fluids.LAVA, new FluidVariantRenderHandler() {
+			@Override
+			public Text getName(FluidVariant fluidVariant) {
+				return ((MutableText) FluidVariantRenderHandler.super.getName(fluidVariant)).setStyle(Style.EMPTY.withColor(Formatting.RED));
+			}
+		});
```
Juuxel: I'm not really sure if this is fitting for Fabric API, but seems like a nice little eyecandy feature.

Juuxel (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRendering.java
Technici4n: Two mods attempting to do it would lead to a crash for duplicate registrations, unless they catch the resulting exception. Imo it's nicer to have fabric API do it.

shedaniel: Personally I would want a way to not apply the colors, maybe another method would fit for the colored one, but this is just out-of-scope in my opinion

Technici4n: How so? MI's fluids are all colored, and I am PR'ing this to avoid MI setting the fluid variant render handler of vanilla fluids.

Note that LBA used to color water and lava fluid keys too (in fact that's where we got the idea to color MI fluids from).

shartte: I like the colors, but I understand the argument too. Hmmm.
If you want to serve both worlds, offer an "enableColoredVanillaFluidNames()", and if any mod calls it -> colored names!

immibis: What if water is purple and lava is green in my texture pack?

Technici4n: Removed this change from this PR. Will be a config option in MI.

modmuss50 (approved):

Player3324 (approved):

PR #1804, opened 2021-11-05
Improvements to tag-based mining level API

Shnupbups:
This PR introduces improvements upon the tag-based mining level API introduced in #1629, mostly focused around Shears

- Shears now work with mining level tags.
- Vanilla Shears have a mining level of 2, equivalent to that of Iron tools.
- Introduced `fabric:mineable/shears_fast` and `fabric:mineable/shears_slow` block tags, allowing for utilisation of the mining speeds used by leaves and vines respectively - these tags are nested in the already present `fabric:mineable/shears`, so it is not necessary to manually put a block in both. Blocks not present in either fast nor slow will use the standard speed used by wool.
- Added `FabricTool`, an interface implemented by `ToolItem` and `ShearsItem` and extended by `DynamicAttributeTool` - allows for Shears to utilise `ToolMaterial`s, and provides some potentially useful methods.
- Added `FabricShearsItem`, a class that extends `ShearsItem` - provides a simple way to pass a `ToolMaterial` into shears.
- Potentially fixed issues regarding `DynamicAttributeTool`s mining blocks with dynamic mining levels
- Added stuff to the Test Mod which utilises new functionality introduced by this PR


fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricMineableTags.java
```diff
@@ -43,6 +43,16 @@
 	 */
 	public static final Tag.Identified<Block> SHEARS_MINEABLE = register("mineable/shears");
 
+	/**
+	 * Blocks in this tag ({@code #fabric:mineable/shears}) can be effectively mined with shears at a faster than normal rate.
+	 */
+	public static final Tag.Identified<Block> SHEARS_MINEABLE_FAST = register("mineable/shears_fast");
+
+	/**
+	 * Blocks in this tag ({@code #fabric:mineable/shears}) can be effectively mined with shears at a slower than normal rate.
```
Juuxel:
```suggestion
	 * Blocks in this tag ({@code #fabric:mineable/shears_slow}) can be effectively mined with shears at a slower than normal rate.
```

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricMineableTags.java
```diff
@@ -43,6 +43,16 @@
 	 */
 	public static final Tag.Identified<Block> SHEARS_MINEABLE = register("mineable/shears");
 
+	/**
+	 * Blocks in this tag ({@code #fabric:mineable/shears}) can be effectively mined with shears at a faster than normal rate.
```
Juuxel:
```suggestion
	 * Blocks in this tag ({@code #fabric:mineable/shears_fast}) can be effectively mined with shears at a faster than normal rate.
```

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,54 @@
...
+package net.fabricmc.fabric.api.mininglevel.v1;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.item.ToolMaterial;
+import net.minecraft.tag.Tag;
+
+public interface FabricTool {
```
Juuxel, line 30: This interface and its methods should be documented.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,54 @@
...
+		ToolMaterial toolMaterial = getToolMaterial();
+
+		if (toolMaterial != null) {
+			return getToolMaterial().getMiningLevel();
+		} else {
+			return 0;
+		}
+	}
+
+	default boolean isSuitableFor(int miningLevel, BlockState state) {
```
Juuxel, line 76: Is this called anywhere in the mining level API? Any tool attributes stuff should stay in that module - all features here should work without the tool attribute api being present.

Juuxel (changes requested):

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: While it is only called by the tool attributes module, it is necessary for that module's `DynamicAttributeTool`s to work with dynamic mining levels introduced by this module.

I don't see your point exactly - it only being called by another module in no way means that this module relies on the other. Mining Level will work just fine without Tool Attributes.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: Done, let me know if it's sufficient

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Juuxel: My point is that the interface has features that don't work when the module is used alone with just its deps. I guess you could work around that by saying something like "Called in other Fabric API modules" in the JD as an explanation for it, but that feels a bit hacky to me.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+		} else {
+			return 0;
+		}
+	}
+
+	/**
+	 * Returns whether a tool of this type, if it had the mining level provided, would be suitable for mining the provided {@code BlockState}.
+	 * Can be useful in circumstances where {@link net.minecraft.item.Item#isSuitableFor(BlockState)} is not.
+	 * @param miningLevel the mining level of the tool
+	 * @param state the blockstate being mined
```
Juuxel:
```suggestion
	 * @param state the block state being mined
```
Small typo fix

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: > the interface has features that don't work when the module is used alone with just its deps.

What? The feature of the interface works just fine, it still returns exactly as it should.
This method doesn't require tool attributes at all, it's the other way around. Tool attributes requires this method. This method works just fine without tool attributes.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public interface FabricTool {
+	/**
+	 * Returns the {@code ToolMaterial} of this tool.
+	 * Can be {@code null} if the tool does not utilise {@code ToolMaterial}s.
+	 * Defaults to {@code null}.
+	 * @return The tool material
+	 */
+	@Nullable
+	default ToolMaterial getToolMaterial() {
```
Technici4n, line 38: What about dynamic tools?

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/**
+	 * Returns a {@code Tag} of the blocks this tool is effective against.
+	 * Can be {@code null} if the tool does not utilise a tag.
+	 * Defaults to {@code null}.
+	 * @return The effective blocks
+	 */
+	@Nullable
+	default Tag<Block> getEffectiveBlocks() {
```
Technici4n, line 49: What about tools that can work on multiple kinds of blocks, e.g. paxels?

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: Dynamic Tools return null, as per the javadoc.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: These can either return a tag that nests `minecraft:mineable/pickaxe`, `minecraft:mineable/axe` and `minecraft:mineable/shovel`, or return null as per the javadoc.

Technici4n: I'm not sure I see the point of these functions then.

Shnupbups: `getToolMaterial` exists so that things like shears can be tied to a tool material, and thus inherit their mining level from it
`getEffectiveBlocks` exists so `isSuitableFor` and `getMiningSpeedMultiplier` can use them
What is so hard to understand about that?

`FabricTool` isn't intended for making fancy multitools or anything, but simply allowing for the creation of simple vanilla-style tools that aren't pickaxes, axes, shovels, or hoes. This API specifically uses it for swords and shears, but it can potentially be used for other things too.

`DynamicAttributeTool` extends `FabricTool`, and it is then ideal for use for making more fancy tools like paxels and drills and whatnot.

Technici4n: Imo what these methods are being used for is not super clear. Is it just metadata? Or does it have some special behavior? What is the effect of overriding them? Perhaps they could be extended a tiny little bit to replace the dynamic tool module entirely?

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Juuxel: You'd expect this method to at least do something when overridden, when it does precisely nothing since it's not called. I don't really see why the usual item method doesn't suffice here since you also know the mining level from the same interface.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Returns the mining level of this tool.
+	 * Defaults to the mining level of the material returned by {@link #getToolMaterial()}, or 0 if that returns {@code null}.
+	 * @return The mining level
+	 */
+	default int getMiningLevel() {
+		ToolMaterial toolMaterial = getToolMaterial();
+
+		if (toolMaterial != null) {
+			return getToolMaterial().getMiningLevel();
```
ghost: Why not just use the local variable `toolMaterial` here...

ghost (approved):
Looks good imo

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: That's... actually a good point, don't know why I didn't consider that.

I've updated it so now this method is actually used by the regular `isSuitableFor`, and thus will work when overridden.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: lol must have missed that when refactoring some things, fixed

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/TaggedToolsModdedBlocksToolHandler.java
```diff
@@ -41,13 +42,13 @@
...
 	public ActionResult isEffectiveOn(Tag<Item> tag, BlockState state, ItemStack stack, @Nullable LivingEntity user) {
-		if (!(stack.getItem() instanceof DynamicAttributeTool) && !(stack.getItem() instanceof ToolItem)) {
+		if (!(stack.getItem() instanceof FabricTool)) {
 			@Nullable ToolManagerImpl.Entry entry = ToolManagerImpl.entryNullable(state.getBlock());
 
 			if (entry != null) {
 				int miningLevel = 0; // minimum mining level: the tool is tagged but nothing else is said about it
 				int requiredMiningLevel = entry.getMiningLevel(tag);
-				return requiredMiningLevel >= miningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
+				return miningLevel >= requiredMiningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
```
Juuxel: My fix for this is a bit more comprehensive (and you need to fix the merge conflicts anyway).

Also, are all `ToolItem`s `FabricTool`s? If not, the if check above would have a different meaning

Juuxel: Did I comment just as you fixed the conflicts? :eyes:

fabric-mining-level-api-v1/src/main/resources/data/fabric/tags/blocks/mineable/shears.json
```diff
@@ -0,0 +1,7 @@
+{
+  "replace": false,
+  "values": [
+    "#fabric:mineable/shears_fast",
+    "#fabric:mineable/shears_slow"
```
Juuxel, line 5: Does this work when the fast/slow tags are not defined? (In a plain environment without the testmod)

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/TaggedToolsModdedBlocksToolHandler.java
Shnupbups: Haha, yeah I fixed the conflicts pretty quickly.

And yes, `ToolItem` implements `FabricTool` courtesy of `ToolItemMixin`. All `ToolItem`s, `ShearsItem`s, and `DynamicAttributeTool`s are `FabricTool`s.

fabric-mining-level-api-v1/src/main/resources/data/fabric/tags/blocks/mineable/shears.json
Shnupbups: Good shout, fixed

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+
+		if (toolMaterial != null) {
+			return toolMaterial.getMiningLevel();
+		} else {
+			return 0;
+		}
+	}
+
+	/**
+	 * Returns whether a tool of this type, if it had the mining level provided, would be suitable for mining the provided {@code BlockState}.
```
Juuxel:
```suggestion
	 * Returns whether a tool of this type, if it had the mining level provided, would be suitable for mining the provided {@link BlockState}.
	 * Automatically called for subclasses of {@link net.minecraft.item.ToolItem}, {@link net.minecraft.item.SwordItem} and {@link net.minecraft.item.ShearsItem}.
```

fabric-mining-level-api-v1/src/testmod/resources/assets/fabric-mining-level-api-v1-testmod/lang/en_us.json
```diff
@@ -0,0 +1,8 @@
+{
+  "item.minecraft.shears": "Iron Shears",
```
Juuxel, line 2: Not really needed in a testmod

fabric-mining-level-api-v1/src/testmod/java/net/fabricmc/fabric/test/mininglevel/MiningLevelTest.java
```diff
@@ -95,13 +140,28 @@ private static void test() {
...
+		checkMiningSpeed(NEEDS_SHEARS, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
+		checkMiningSpeed(NEEDS_SHEARS_SLOW, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
+		checkMiningSpeed(NEEDS_SHEARS_FAST, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
+		checkMiningSpeed(NEEDS_STONE_SHEARS, Items.SHEARS, NETHERITE_SHEARS, Items.STICK);
+		checkMiningSpeed(NEEDS_DIAMOND_SHEARS, Items.SHEARS, NETHERITE_SHEARS, Items.STICK);
+		checkMiningSpeed(NEEDS_NETHERITE_SHEARS, Items.SHEARS, NETHERITE_SHEARS, Items.STICK);
+
+		checkMiningSpeed(Blocks.VINE, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
+		checkMiningSpeed(Blocks.WHITE_WOOL, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
+		checkMiningSpeed(Blocks.OAK_LEAVES, WOODEN_SHEARS, STONE_SHEARS, Items.SHEARS, DIAMOND_SHEARS, NETHERITE_SHEARS, GOLDEN_SHEARS, Items.STICK);
```
Juuxel, line 164: I don't really see why these should be printed out. Instead, it might be useful to compare the mining speeds like the tool attribute test.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
```diff
@@ -0,0 +1,82 @@
...
+	 * @see net.minecraft.item.Item#isSuitableFor(BlockState)
+	 * @param miningLevel the mining level of the tool
+	 * @param state the block state being mined
+	 * @return Whether the tool is suitable
+	 */
+	default boolean isSuitableFor(int miningLevel, BlockState state) {
```
Juuxel, line 76: Also, why is this needed? You know the mining level from this same interface, so you don't really have to do the "if it had the mining level" guesswork.

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/ModdedToolsVanillaBlocksToolHandler.java
```diff
@@ -55,18 +56,18 @@ private ToolItem getVanillaItem(int miningLevel) {
...
 
 			// Gets the mining level from our modded tool
-			int miningLevel = ((DynamicAttributeTool) stack.getItem()).getMiningLevel(tag, state, stack, user);
+			int miningLevel = tool.getMiningLevel(tag, state, stack, user);
 			if (miningLevel < 0) return ActionResult.PASS;
 
-			ToolItem vanillaItem = getVanillaItem(miningLevel);
-			return vanillaItem.isSuitableFor(state) ? ActionResult.SUCCESS : ActionResult.PASS;
+			FabricTool vanillaItem = (FabricTool) getVanillaItem(miningLevel);
+			return vanillaItem.isSuitableFor(miningLevel, state) ? ActionResult.SUCCESS : ActionResult.PASS;
```
Juuxel, line 70: This doesn't really need to be changed to use the interface's `isSuitableFor` since they return the same value because of the mixin, but this is fine too I guess.

Juuxel (changes requested):

fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/ModdedToolsVanillaBlocksToolHandler.java
Shnupbups: `getVanillaItem` can only return... well, vanilla items (or rather, items it specifically knows about, which happen to be just vanilla items). That is, for example, wooden, stone, iron, diamond, and netherite pickaxes, of mining levels 0, 1, 2, 3, and 4 respectively.
If `getVanillaItem(5)` is called, it will return the highest one it knows about: netherite, which has a mining level of _4_ and not the desired 5.
Thus, if the original `isSuitableFor(state)` were to be called on the item the method returns, in this case it would be checking if a tool with a mining level of _4_ was suitable for the state... which is not what we want.
So, instead, we use `isSuitableFor(miningLevel, state)` to ensure that we are indeed checking for a mining level of 5.

This then justifies the existence of said method.

fabric-mining-level-api-v1/src/main/java/net/fabricmc/fabric/api/mininglevel/v1/FabricTool.java
Shnupbups: See my response to the `ModdedToolsVanillaBlocksToolHandler` comment you made below.

fabric-mining-level-api-v1/src/testmod/java/net/fabricmc/fabric/test/mininglevel/MiningLevelTest.java
Shnupbups: Perhaps, I just wanted something quick and dirty to do the trick for debugging purposes and figured that plenty of other test mods spam the log with prints, so it wouldn't be too terrible.

fabric-mining-level-api-v1/src/testmod/resources/assets/fabric-mining-level-api-v1-testmod/lang/en_us.json
Shnupbups: eh, neither are things like textures, but i figured 'why not'. after all, it doesn't hurt

Technici4n:
I don't think this will ever get merged in its current state. Maybe some bits could be PR'ed separately.

PR #1808, opened 2021-11-06
Fix possible issues with API instances begin null.

### LINKED ISSUE #1806: Cannot invoke "FluidRenderHandlerRegistry.get()" because "FluidRenderHandlerRegistry.INSTANCE" is null
I ran into this weird issue that `FluidRenderHandlerRegistry.INSTANCE` is `null` when searching for water in REI with Fabric API 0.42.0 and I think I know why this is happening.

Here's the relevant code:

https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java#L32-L33

https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java#L43-L44

On first glance, this looks good, REI (Architectury, actually) calls `FluidRenderHandlerRegistry.INSTANCE.get()` which should statically initialize `FluidRenderHandlerRegistry`, which should then statically initialize `FluidRenderHandlerRegistryImpl` before setting the `FluidRenderHandlerRegistry.INSTANCE` field, so how can it be `null`?

This issue happens when `FluidRenderHandlerRegistryImpl` gets statically initialized first, because before `Impl` is getting initialized, the superinterface `FluidRenderHandlerRegistry` will be initialized ([reference](https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.1)), so `FluidRenderHandlerRegistry.INSTANCE` will be set to `FluidRenderHandlerRegistryImpl.INSTANCE`, which at that point is `null`!

Okay, but why does this happen, where is `Impl` getting initialized?

There are currently only 4 references to `Impl`:
https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java#L59

https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java#L85

https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinFluidRenderer.java#L149

https://github.com/FabricMC/fabric/blob/098ef32f377adfd9407233e42d5d6f6a91e519bc/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandler.java#L85

The first one of these will run on resource reload, so basically when reaching the title screen, this explains what I found while debugging: `FluidRenderHandlerRegistry.INSTANCE.get()` will work when run in the client mod initializer (and will fix the issue) but it won't work when run ingame.

## A Bunch of Solutions

I think one quick way to fix this would be to eliminate any direct reference to `FluidRenderHandlerRegistryImpl`, e.g. by replacing
`FluidRenderHandlerRegistryImpl.INSTANCE` with `(FluidRenderHandlerRegistryImpl) FluidRenderHandlerRegistry.INSTANCE`, but this is a very brittle solution.

A more robust fix would be to replace the `FluidRenderHandlerRegistry.INSTANCE` field with a default `getInstance()` method, but this is also an API breaking change.

A non-API-breaking fix could be to remove the `FluidRenderHandlerRegistryImpl.INSTANCE` field altogether and create the instance directly in the interface - this will need a `public` `Impl` constructor.

A non-structural fix could be to statically initialize `FluidRenderHandlerRegistry` as early as possible e.g. running it in Fabric API's client mod initializer, but this could theoretically still break if e.g. another mod tries to access `Impl` even earlier (mod load order, hacks, etc).

modmuss50:
One solution to fix: https://github.com/FabricMC/fabric/issues/1806 See for other suggestions for fixes. 

This will break a few mods that used the impl. (Seems to just be TechReborn in AOF 4, I am releasing a fix now so I think this is a none issue) To reduce the impact of this it may be a better idea to just fix the FluidRenderHandlerRegistry?

### DIFF
--- a/fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
+++ b/fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
@@ -31,6 +31,7 @@ import net.fabricmc.fabric.api.client.networking.v1.ClientPlayNetworking;
 import net.fabricmc.fabric.api.client.screen.ContainerScreenFactory;
 import net.fabricmc.fabric.api.client.screen.ScreenProviderRegistry;
 import net.fabricmc.fabric.api.container.ContainerFactory;
+import net.fabricmc.fabric.api.container.ContainerProviderRegistry;
 import net.fabricmc.fabric.impl.container.ContainerProviderImpl;
 
 public class ScreenProviderRegistryImpl implements ScreenProviderRegistry {
@@ -50,7 +51,7 @@ public class ScreenProviderRegistryImpl implements ScreenProviderRegistry {
 	@Override
 	public <C extends ScreenHandler> void registerFactory(Identifier identifier, ContainerScreenFactory<C> containerScreenFactory) {
 		registerFactory(identifier, (syncId, identifier1, player, buf) -> {
-			C container = ((ContainerProviderImpl) ContainerProviderImpl.INSTANCE).createContainer(syncId, identifier1, player, buf);
+			C container = ((ContainerProviderImpl) ContainerProviderRegistry.INSTANCE).createContainer(syncId, identifier1, player, buf);
 
 			if (container == null) {
 				LOGGER.error("Could not open container for {} - a null object was created!", identifier1.toString());
--- a/fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
+++ b/fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
@@ -34,11 +34,6 @@ import net.fabricmc.fabric.api.container.ContainerFactory;
 import net.fabricmc.fabric.impl.container.ContainerProviderImpl;
 
 public class ScreenProviderRegistryImpl implements ScreenProviderRegistry {
-	/**
-	 * Use the instance provided by ScreenProviderRegistry.
-	 */
-	public static final ScreenProviderRegistry INSTANCE = new ScreenProviderRegistryImpl();
-
 	private static final Logger LOGGER = LogManager.getLogger();
 
 	private static final Map<Identifier, ContainerFactory<HandledScreen>> FACTORIES = new HashMap<>();
@@ -55,7 +50,7 @@ public class ScreenProviderRegistryImpl implements ScreenProviderRegistry {
 	@Override
 	public <C extends ScreenHandler> void registerFactory(Identifier identifier, ContainerScreenFactory<C> containerScreenFactory) {
 		registerFactory(identifier, (syncId, identifier1, player, buf) -> {
-			C container = ContainerProviderImpl.INSTANCE.createContainer(syncId, identifier1, player, buf);
+			C container = ((ContainerProviderImpl) ContainerProviderImpl.INSTANCE).createContainer(syncId, identifier1, player, buf);
 
 			if (container == null) {
 				LOGGER.error("Could not open container for {} - a null object was created!", identifier1.toString());
--- a/fabric-blockrenderlayer-v1/src/main/java/net/fabricmc/fabric/api/blockrenderlayer/v1/BlockRenderLayerMap.java
+++ b/fabric-blockrenderlayer-v1/src/main/java/net/fabricmc/fabric/api/blockrenderlayer/v1/BlockRenderLayerMap.java
@@ -36,7 +36,7 @@ import net.fabricmc.fabric.impl.blockrenderlayer.BlockRenderLayerMapImpl;
  * <p>Client-side only.
  */
 public interface BlockRenderLayerMap {
-	BlockRenderLayerMap INSTANCE = BlockRenderLayerMapImpl.INSTANCE;
+	BlockRenderLayerMap INSTANCE = new BlockRenderLayerMapImpl();
 
 	/**
 	 * Map (or re-map) a block state with a render layer.  Re-mapping is not recommended but if done, last one in wins.
--- a/fabric-blockrenderlayer-v1/src/main/java/net/fabricmc/fabric/impl/blockrenderlayer/BlockRenderLayerMapImpl.java
+++ b/fabric-blockrenderlayer-v1/src/main/java/net/fabricmc/fabric/impl/blockrenderlayer/BlockRenderLayerMapImpl.java
@@ -28,7 +28,7 @@ import net.minecraft.item.Item;
 import net.fabricmc.fabric.api.blockrenderlayer.v1.BlockRenderLayerMap;
 
 public class BlockRenderLayerMapImpl implements BlockRenderLayerMap {
-	private BlockRenderLayerMapImpl() { }
+	public BlockRenderLayerMapImpl() { }
 
 	@Override
 	public void putBlock(Block block, RenderLayer renderLayer) {
@@ -74,8 +74,6 @@ public class BlockRenderLayerMapImpl implements BlockRenderLayerMap {
 			putFluid(fluid, renderLayer);
 		}
 	}
-
-	public static final BlockRenderLayerMap INSTANCE = new BlockRenderLayerMapImpl();
 
 	private static Map<Block, RenderLayer> blockRenderLayerMap = new HashMap<>();
 	private static Map<Item, RenderLayer> itemRenderLayerMap = new HashMap<>();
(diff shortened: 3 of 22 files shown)


fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
```diff
@@ -55,7 +50,7 @@ public void registerFactory(Identifier identifier, ContainerFactory<HandledScree
 	@Override
 	public <C extends ScreenHandler> void registerFactory(Identifier identifier, ContainerScreenFactory<C> containerScreenFactory) {
 		registerFactory(identifier, (syncId, identifier1, player, buf) -> {
-			C container = ContainerProviderImpl.INSTANCE.createContainer(syncId, identifier1, player, buf);
+			C container = ((ContainerProviderImpl) ContainerProviderImpl.INSTANCE).createContainer(syncId, identifier1, player, buf);
```
shedaniel: Shouldn't this be `((ContainerProviderImpl) ContainerProviderRegistry.INSTANCE)`, I think this either compiles to that or it might either cause the same issues

modmuss50: It's basically the same, only reason I didnt do it was becuase I am lazy, but I dont think it makes any diffrence.

I have tested this PR in AOF and it seems to work fine.

Juuxel: Yeah there are many other places in this PR using this pattern. I think it just compiles to the same since the impls don't have their own fields, but that's a bit unclear.

Juuxel (approved):

fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
Fourmisain: Quoting https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.1:

> A reference to a static field (§8.3.1.1) causes initialization of only the class or interface that actually declares it, even though it might be referred to through the name of a subclass, a subinterface, or a class that implements an interface.

So it's fine to write it like this, it's just a little confusing to look at.

shedaniel: Yeah, just a bit confusing, it is fine I guess.

shedaniel (approved):

fabric-containers-v0/src/main/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
modmuss50: Cleaned it up, found another case of it as well. 👍

Fourmisain:
Oh no, Sodium references the `Impl` `INSTANCE` too: <https://github.com/CaffeineMC/sodium-fabric/blob/d18bd93346df563ea29ce8ebafdccad5e2ca6c03/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java#L133>

Here's how it looks with this PR:
ohno

Instead of rendering a river, the console renders infinite `NoSuchFieldError`s.

modmuss50:
I don’t fancy breaking mods like sodium this late into 1.17, it would be a support nightmare we can easily avoid.

Would adding the field back into the impl, but setting it via from the interface work? Could then just deprecate it for removal and get rid of it in 1.18.

Fourmisain:
(Disclaimer: It's a wall of text again!)

> Would adding the field back into the impl, but setting it via from the interface work?

Hm... I'm not sure, so I think what you mean is putting the
`INSTANCE = new FluidRenderHandlerRegistryImpl()`
inside the interface and then the
`INSTANCE = FluidRenderHandlerRegistry.INSTANCE`
inside `Impl`, basically reversing how it is right now, right?

This should work if the `Impl` gets initialized first, but if the interface gets initialized first (if e.g. a mod calls `FluidRenderHandlerRegistry.INSTANCE.register()` from the mod initializer), we should have the issue that now `Impl.INSTANCE` will be `null`.

So it basically just reverses the situation, but the issue is that *sometimes* the interface gets initialized first and *sometimes* it's the `Impl`, depending on what the installed mods do.

### "Non-structural ~~Fix~~ Hack"

As it is right now, it'll only work if the interface gets initialized first.
That's why I suggested the "non-structural fix" of trying to ensure this by forcing it in the client mod initializer.

This could be done by putting some "NOP" (getting some déjà vu) like
```
FluidRenderHandlerRegistry instance = FluidRenderHandlerRegistry.INSTANCE;
```
or
```
FluidRenderHandlerRegistry.INSTANCE.get(Fluids.EMPTY);
```
in `onInitializeClient()`.

It's hacky and ugly at that, but it'll probably do the job well enough to bridge the gap.

### Affected Mods

I did a [GitHub search](https://github.com/search?q=FluidRenderHandlerRegistryImpl&type=Code) earlier to check for mods referencing the `Impl` and the list is actually really short:
Arcane Magic, JustMap (both not updated for 2 years), SimpleFluidTanks (not updated for 1 year), St'ructure Tools (actively developed) and Sodium (and the Iris fork) of course.

None of them reference the `Impl` inside the client mod initializer, so the above fix should be relatively safe.

### Other Ways?

I'm still not sure if there are other "non-structural" solutions.
I had some thoughts about using `static {}` initializer blocks that check/set the fields if they are not set on the other end - except interfaces don't allow those blocks and the fields are `final`.

There *may* be an even hackier fix that sets both `INSTANCE` fields from `FluidRenderHandlerRegistryImpl`'s constructor via reflection - what could go wrong? 😅

Fourmisain:
> There *may* be an even hackier fix that sets both `INSTANCE` fields from `FluidRenderHandlerRegistryImpl`'s constructor via reflection

Actually, could one not use a mutable accessor to set the fields from the constructor?
Something like
```
private FluidRenderHandlerRegistryImpl() {
  FluidRenderHandlerRegistryAccessor.setINSTANCE(this);
}
```
This only feels _somewhat_ illegal, rewriting final fields, leaking uninitialized objects...

Aight, I definitely reached my evil limit for today, gonna head out!

modmuss50:
Setting in the client init seems the best way to get a fix out for this ASAP, and only for the `FluidRenderHandlerRegistry`

> but it'll probably do the job well enough to bridge the gap.

I agree that this is the best solution for 1.17, the changes in this PR can go into 1.18 where I have no care for breaking mods that use impl classes.

Thanks a lot for your help, ill get a PR ready shortly.

modmuss50:
I manually cherrypicked this right into 1.18 here: https://github.com/FabricMC/fabric/commit/8a8e3d81de655bbd219e5bfa7c793ad6c18381a8

Fourmisain:
Got some sleep and now I realize there actually was an easy non-hack solution.

Basically one could have added some singleton pattern to `Impl` solely to set both `INSTANCE` fields, e.g.:

```
public class FluidRenderHandlerRegistryImpl implements FluidRenderHandlerRegistry {
  @Deprecated(forRemoval = true)
  public static final FluidRenderHandlerRegistryImpl INSTANCE = getInstance();

  // Bill Pugh singleton pattern:

  private static class Singleton {
    private static final FluidRenderHandlerRegistryImpl INST = new FluidRenderHandlerRegistryImpl();
  }

  @Deprecated(forRemoval = true)
  public static FluidRenderHandlerRegistryImpl getInstance() {
    return Singleton.INST;
  }

  // alternative (single threaded) pattern:

  // private static FluidRenderHandlerRegistryImpl singleton;
  //
  // public static FluidRenderHandlerRegistryImpl getInstance() {
  //   if (singleton == null) singleton = new FluidRenderHandlerRegistryImpl();
  //   return inst;
  // }
}

public interface FluidRenderHandlerRegistry {
  public static final FluidRenderHandlerRegistryImpl INSTANCE = FluidRenderHandlerRegistryImpl.getInstance();
}
```

This would have worked too, but it also adds to the mess because now there's 3 ways to access the same instance... 😄

PR #1809, opened 2021-11-06
Fix shears being the ultimate mining tool

### LINKED ISSUE #1805: FAPI .42 Broken Ores
WIth the latest Fabric API Update for 1.17.1, Industrial Ore, TechReborn Ore, Things Gleaming Ore are broken and will not allow Pickaxes to drop the ore upon breaking. Using any normally required pickaxe to break these, all it does is break the block. 

If you have WTHIT Enabled, it also says the required harvested tool is a shear. Using a shear, you can get the ore content. 

Vanilla ore seems in tact. only certain modded ore is broken. 

image

image

Juuxel:
Fixes #1805. Most of this PR is testmod changes to catch these types of errors in the future, and the actual fix is a one-liner.

Note that this doesn't `fix` #1810 which is most likely only theoretical and thus not critical.

### DIFF
--- a/fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/TaggedToolsModdedBlocksToolHandler.java
+++ b/fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/TaggedToolsModdedBlocksToolHandler.java
@@ -47,7 +47,7 @@ public class TaggedToolsModdedBlocksToolHandler implements ToolManagerImpl.ToolH
 			if (entry != null) {
 				int miningLevel = 0; // minimum mining level: the tool is tagged but nothing else is said about it
 				int requiredMiningLevel = entry.getMiningLevel(tag);
-				return requiredMiningLevel >= miningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
+				return miningLevel >= requiredMiningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/impl/tool/attribute/handlers/TaggedToolsModdedBlocksToolHandler.java
```diff
@@ -47,7 +47,7 @@ public ActionResult isEffectiveOn(Tag<Item> tag, BlockState state, ItemStack sta
 			if (entry != null) {
 				int miningLevel = 0; // minimum mining level: the tool is tagged but nothing else is said about it
 				int requiredMiningLevel = entry.getMiningLevel(tag);
-				return requiredMiningLevel >= miningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
+				return requiredMiningLevel >= 0 && miningLevel >= requiredMiningLevel ? ActionResult.SUCCESS : ActionResult.PASS;
```
shedaniel: honestly this could probably be simplified to requiredMiningLevel == 0, not sure if it would make it less clear however

Juuxel: Yeah I know, but my formatting preserves the "intention" behind the code. I wonder if the compiler knows how to optimise it...

I guess it could always be `== 0` with the logic explained in a comment if needed?

modmuss50 (approved):
Looks great, thanks for the good tests. Will work on getting this out soon.

PR #1815, opened 2021-11-07
Add an API to enable elytra flight

Technici4n:
- Add an `ALLOW` event to block elytra flight entirely, and a `CUSTOM` event for modded "custom" elytras.
- Add the `FabricElytraItem` interface for the most common use case.
- Add an event to disable cape rendering. **I'd appreciate some testing with this one (just wear the `diamond_elytra` test item) by someone who has a cape.**

Note that if you test this on AOF4 it's not going to work. This is because FallFlyingLib completely overwrites the vanilla logic. One objective of this PR is to replace FallFlyingLib, or at least allow it to call this event.

The mixin injections are not obvious, but being straight injects they should not cause compatibility issues, and they work around existing libraries. I haven't tested it in production yet, but I expect this addition to work on AOF4 with FallFlyingLib (and dependent mods) removed.

TODO:
- [x] Test on AOF4 with FFL removed

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.entity.event.v1;
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events related to elytra flight for living entities. Elytra flight is also known as "fall flying".
+ */
+public final class EntityElytraEvents {
+	/**
+	 * An event to check if elytra flight (both through normal and custom elytras) is allowed.
+	 * All listeners need to return true to allow the entity to fly, otherwise elytra flight will be blocked/stopped.
+	 */
+	public static final Event<Allow> ALLOW = EventFactory.createArrayBacked(Allow.class, listeners -> entity -> {
+		for (Allow listener : listeners) {
+			if (!listener.allowElytraFlight(entity)) {
+				return false;
+			}
+		}
+
+		return true;
+	});
+
+	/**
+	 * An event to grant elytra flight to living entities when some condition is met.
+	 * Will be called when players try to start elytra flight by pressing space in mid-air, and every tick for all flying living entities to check if elytra flight is still allowed.
+	 *
+	 * <p>Items that wish to enable custom elytra flight when worn in the chest equipment slot can simply implement {@link FabricElytraItem} instead of registering a listener.
+	 */
+	public static final Event<Custom> CUSTOM = EventFactory.createArrayBacked(Custom.class, listeners -> (entity, tickElytra) -> {
+		for (Custom listener : listeners) {
+			if (listener.useCustomElytra(entity, tickElytra)) {
+				return true;
+			}
+		}
+
+		return false;
+	});
+
+	static {
+		CUSTOM.register((entity, tickElytra) -> {
+			ItemStack chestStack = entity.getEquippedStack(EquipmentSlot.CHEST);
+
+			if (chestStack.getItem() instanceof FabricElytraItem fabricElytraItem) {
+				return fabricElytraItem.useCustomElytra(entity, chestStack, tickElytra);
+			}
+
+			return false;
+		});
+	}
+
+	@FunctionalInterface
+	public interface Allow {
+		/**
+		 * @return false to block elytra flight, true to allow it (unless another listener returns false)
+		 */
+		boolean allowElytraFlight(LivingEntity entity);
+	}
+
+	@FunctionalInterface
+	public interface Custom {
+		/**
+		 * Try to use a custom elytra for an entity.
+		 * A custom elytra is anything that allows an entity to enter and continue elytra flight when some condition is met.
+		 * Listeners should follow the following pattern:
+		 * <pre>{@code
+		 * EntityElytraEvents.CUSTOM.register((entity, tickElytra) -> {
+		 *     if (check if condition for custom elytra is met) {
+		 *         if (tickElytra) {
+		 *             // Optionally consume some resources that are being used up in order to fly, for example damaging an item.
+		 *             // Optionally perform other side effects of elytra flight, for example playing a sound.
+		 *         }
+		 *         // Allow entering/continuing elytra flight with this custom elytra
+		 *         return true;
+		 *     }
+		 *     // Condition for the custom elytra is not met: don't let players enter or continue elytra flight (unless another elytra is available).
+		 *     return false;
+		 * });
+		 * }</pre>
+		 *
+		 * @param entity     the entity
+		 * @param tickElytra false if this is just to check if the custom elytra can be used, true if the custom elytra should also be ticked, i.e. perform side-effects of flying such as using resources.
+		 * @return true to use a custom elytra, enabling elytra flight for the entity and cancelling subsequent handlers
+		 */
+		boolean useCustomElytra(LivingEntity entity, boolean tickElytra);
+	}
+
+	private EntityElytraEvents() {
+	}
+}
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/ClientPlayerEntityMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/ClientPlayerEntityMixin.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.entity.event.elytra;
+
+import com.mojang.authlib.GameProfile;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Slice;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.network.AbstractClientPlayerEntity;
+import net.minecraft.client.network.ClientPlayerEntity;
+import net.minecraft.client.network.ClientPlayNetworkHandler;
+import net.minecraft.client.world.ClientWorld;
+import net.minecraft.item.Items;
+import net.minecraft.network.packet.c2s.play.ClientCommandC2SPacket;
+
+@SuppressWarnings("unused")
+@Mixin(ClientPlayerEntity.class)
+abstract class ClientPlayerEntityMixin extends AbstractClientPlayerEntity {
+	ClientPlayerEntityMixin(ClientWorld world, GameProfile profile) {
+		super(world, profile);
+		throw new AssertionError();
+	}
+
+	@Shadow
+	@Final
+	private ClientPlayNetworkHandler networkHandler;
+
+	/**
+	 * Call {@link #checkFallFlying()} even if the player is not wearing {@link Items#ELYTRA} to allow custom elytra flight.
+	 */
+	@Inject(at = @At(value = "FIELD", target = "Lnet/minecraft/entity/EquipmentSlot;CHEST:Lnet/minecraft/entity/EquipmentSlot;"), method = "tickMovement", slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;isClimbing()Z"), to = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;checkFallFlying()Z")), allow = 1)
+	void injectElytraStart(CallbackInfo info) {
+		// Note that if fall flying is not ALLOWed, checkFallFlying will return false and nothing will happen.
+		if (this.checkFallFlying()) {
+			networkHandler.sendPacket(new ClientCommandC2SPacket(this, ClientCommandC2SPacket.Mode.START_FALL_FLYING));
+		}
+	}
+}
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.entity.event.elytra;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.entity.EntityType;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.entity.event.v1.EntityElytraEvents;
+
+@SuppressWarnings("unused")
+@Mixin(PlayerEntity.class)
+abstract class PlayerEntityMixin extends LivingEntity {
+	PlayerEntityMixin(EntityType<? extends LivingEntity> entityType, World world) {
+		super(entityType, world);
+		throw new AssertionError();
+	}
+
+	@Shadow
+	public abstract void startFallFlying();
+
+	/**
+	 * Allow the server-side and client-side elytra checks to fail when {@link EntityElytraEvents#ALLOW} blocks flight,
+	 * and otherwise to succeed for elytra flight through {@link EntityElytraEvents#CUSTOM}.
+	 */
+	@SuppressWarnings("ConstantConditions")
+	@Inject(at = @At(value = "FIELD", target = "Lnet/minecraft/entity/EquipmentSlot;CHEST:Lnet/minecraft/entity/EquipmentSlot;"), method = "checkFallFlying()Z", allow = 1, cancellable = true)
+	void injectElytraCheck(CallbackInfoReturnable<Boolean> cir) {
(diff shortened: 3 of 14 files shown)


NotJustAnna (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Shadow
+	public abstract void startFallFlying();
+
+	/**
+	 * Allow the server-side and client-side elytra checks to succeed for elytra flight through {@link AllowElytraFlight}.
+	 */
+	@Inject(at = @At(value = "FIELD", target = "Lnet/minecraft/entity/EquipmentSlot;CHEST:Lnet/minecraft/entity/EquipmentSlot;"), method = "checkFallFlying()Z", require = 1, allow = 1, cancellable = true)
+	void injectElytraCheck(CallbackInfoReturnable<Boolean> cir) {
+		if (AllowElytraFlight.EVENT.invoker().allowElytraFlight(PlayerEntity.class.cast(this), false)) {
```
liach: I recall a cast like `(PlayerEntity) (Object) this` would be optimized by mixin in bytecode when applying the mixin methods. So maybe just that ugly cast.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/ClientPlayerEntityMixin.java
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Allow the client to start flight with custom elytras implemented through {@link AllowElytraFlight}.
+	 */
+	@Inject(at = @At(value = "FIELD", target = "Lnet/minecraft/entity/EquipmentSlot;CHEST:Lnet/minecraft/entity/EquipmentSlot;"), method = "tickMovement", slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;isClimbing()Z"), to = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;checkFallFlying()Z")), require = 1, allow = 1)
+	void injectElytraStart(CallbackInfo info) {
+		ClientPlayerEntity self = ClientPlayerEntity.class.cast(this);
+
+		if (AllowElytraFlight.EVENT.invoker().allowElytraFlight(this, false) && this.checkFallFlying()) {
+			self.networkHandler.sendPacket(new ClientCommandC2SPacket(this, ClientCommandC2SPacket.Mode.START_FALL_FLYING));
```
liach: Should shadow the field instead

liach (approved):
The `ChestItem` sounds a bit weird but I don't think I have a better alternative.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
Technici4n: Good point, while I find that `.class.cast` looks better, it produces slightly worse bytecode.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/AllowElytraFlight.java
```diff
@@ -0,0 +1,96 @@
...
+	 * @param player The player.
+	 * @param tickElytra False if this is just to check if elytra flight is allowed. True if the elytra should be ticked, i.e. damaged.
+	 * @return True to allow elytra flight, and cancel subsequent handlers. False to disallow elytra flight, unless a subsequent listener allows it.
+	 */
+	boolean allowElytraFlight(PlayerEntity player, boolean tickElytra);
+
+	/**
+	 * An interface that can be implemented on an item to enable elytra flight when it is worn in the {@link EquipmentSlot#CHEST} slot.
+	 */
+	interface ChestItem {
```
Juuxel: Maybe `ChestElytraItem`? `ChestItem` sounds a bit too much like an item for a chest (the container) :smile:

Player3324:
I don't mind the event, but maybe the top level class should have a somewhat more generic name to at least also include a potential "elytra flight event" that indicates that/when the actual activity happens (after it has been allowed).

Technici4n:
Renamed the event to `Grant` in case we want to add an event that prevents elytra flight which could be called `Allow`.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,116 @@
...
+		 */
+		boolean grantElytraFlight(LivingEntity livingEntity, boolean tickElytra);
+	}
+
+	/**
+	 * An interface that can be implemented on an item to enable elytra flight when it is worn in the {@link EquipmentSlot#CHEST} slot.
+	 *
+	 * <p>To disable cape rendering when this item is worn (like the vanilla elytra item), have a look at {@code LivingEntityFeatureRenderEvents}.
+	 */
+	public interface FabricElytraItem {
```
Shnupbups: I'd recommend this not be an inner class, but that's purely for aesthetic reasons of preferring `implements FabricElytraItem` over `implements EntityElytraEvents.FabricElytraItem`.

Also, since this itself isn't really an `Item`, shouldn't this just be called `FabricElytra` and not `FabricElytraItem`?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		/**
+		 * A helper to perform the default vanilla elytra tick logic: damage the elytra every 20 ticks, and send a game event every 10 ticks.
+		 */
+		default void doVanillaElytraTick(LivingEntity livingEntity, ItemStack chestStack) {
+			int nextRoll = livingEntity.getRoll() + 1;
+
+			if (!livingEntity.world.isClient && nextRoll % 10 == 0) {
+				if ((nextRoll / 10) % 2 == 0) {
```
Shnupbups: Should there not be some way of letting custom elytra take damage in different tick intervals? Say, once every 5 ticks, every 10 ticks, or even every 40 ticks?

Perhaps a method `int getDamageTickInterval(LivingEntity livingEntity, ItemStack stack)` which returns said interval, to be used here, which defaults to 20?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,116 @@
...
+					chestStack.damage(1, livingEntity, p -> p.sendEquipmentBreakStatus(EquipmentSlot.CHEST));
+				}
+
+				livingEntity.emitGameEvent(GameEvent.ELYTRA_FREE_FALL);
```
Shnupbups: Perhaps some way of letting custom elytra *not* emit the game event? So you could have some sort of 'sneaky' elytra? Or maybe a way of changing the event emitted to a different, custom one?

Perhaps a method `boolean shouldEmitGameEvent(LivingEntity livingEntity, ItemStack stack)` to be used here, which defaults to true? And/or `GameEvent getGameEvent(LivingEntity livingEntity, ItemStack stack)` which defaults to `GameEvent.ELYTRA_FREE_FALL`?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,116 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.world.event.GameEvent;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events related to elytra flight for living entities. Elytra flight is also knows as "fall flying".
+ */
+public class EntityElytraEvents {
```
Shnupbups: Perhaps a `DENY` event so elytra can be conditionally disabled? Say, if I made a custom dimension and didn't want players to be able to use elytra within it?

Shnupbups (approved):
Looks good! Have a few suggestions though.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Shnupbups: Ah, just noticed you yourself mentioned a possible `ALLOW` event.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,116 @@
...
+				return true;
+			}
+
+			return false;
+		}
+
+		/**
+		 * A helper to perform the default vanilla elytra tick logic: damage the elytra every 20 ticks, and send a game event every 10 ticks.
+		 */
+		default void doVanillaElytraTick(LivingEntity livingEntity, ItemStack chestStack) {
```
maityyy: I don't think it's worth copying the vanilla code, as it will make it difficult to update the module to the new version of the game.

This API will break my mod, which completely overrides the elytra damage logic, so this is very important to me.
https://gist.github.com/maityyy/05a7656e363cd347c6e583646a32a0f9

I think if someone needs this functionality, they can use the mixin without any problems, which will not break my mod.

maityyy (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Technici4n: > I'd recommend this not be an inner class, but that's purely for aesthetic reasons of preferring `implements FabricElytraItem` over `implements EntityElytraEvents.FabricElytraItem`.

This was done to avoid spamming API classes for something that is essentially a convenient implementation of the event.

> Also, since this itself isn't really an `Item`, shouldn't this just be called `FabricElytra` and not `FabricElytraItem`?

It's not an item, but it only works if it's implemented on an `Item` subclass, so implementors should be items. Another advantage is that people searching for `ElytraItem` should find this class in the popup.

Does this explain the decisions?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Technici4n: Given how short the logic is, mods should imo just copy/paste it and adjust it. The primary purpose of this method is to illustrate what mods are expected to do in their "elytra tick".

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Technici4n: Here too it is much easier to copy/paste the method and remove this call if custom behavior is desired.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Technici4n: I don't think it's realistically possible atm due to the various injects that mods do. Trying to short-circuit the vanilla path is a bit tough when 3-4 mods already inject into it. Once mods migrate to this event, introducing ALLOW will be a lot easier as we will be the only ones editing vanilla code.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Shnupbups: IMO the less copypasting and code repetition the better.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Shnupbups: Explains them, yes, but I still think calling it `FabricElytra` and having it be a separate class would be better.

Though perhaps, if suffixing such interfaces with `Item` is preferred, I should change my own `FabricTool` in my PR to `FabricToolItem`...

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
NotJustAnna: There's a lot of conflicts in how mixins are written and how tightly they're written together. My mod (Winged) had quite a lot of problems regarding mod compatibility between other mods, solely because of how many mixins ends up being mashed together. I tried to fix this with FFL and ultimately failed, and I ultimately prefer that the API allows for the least amount of conflicts between mods trying to implement Elytras and elytra-related functionality.

NotJustAnna: Also this code don't interfere with the vanilla elytra funcionality, and would only break cases of modded elytras + your mod's feature, which I think is fair enough. Ask modders if they want their custom wings to have compatibility with your mod, and just accept the case that not everyone will want such compatibility.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
liach: > IMO the less copypasting and code repetition the better.

Vanilla minecraft logic unfortunately is just a lot of hardcoding. So we do have to resort to copypasta unfortunately.

Technici4n: It's more that this is a trivial 5 lines function, and I would rather have mods copy/paste the bits they need (again, we are talking about ~5 lines that basically do 2 things) than introduce two extra functions to customize this.

We don't know if mojang is going to change their elytra tick logic, but if they do I'd rather have mods decide if they want the new behavior instead of their old custom behavior, than having to break or extend this default helper.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Juuxel: I agree with Shnupbups here (esp. about the nesting). `implements EntityElytraEvents.FabricElytraItem` is honestly pretty ugly and having a top-level interface would be nicer.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Juuxel: Should probably be `final` to be consistent with other event classes, but doesn't *really* affect anything in this case with the private ctor

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/LivingEntityFeatureRenderEvents.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.client.render.entity.feature.FeatureRenderer;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events related to living entity {@link FeatureRenderer}s.
+ * To register a renderer, see {@link LivingEntityFeatureRendererRegistrationCallback} instead.
+ */
+public class LivingEntityFeatureRenderEvents {
```
Juuxel: Same here

fabric-rendering-v1/src/main/java/net/fabricmc/fabric/api/client/rendering/v1/LivingEntityFeatureRenderEvents.java
```diff
@@ -0,0 +1,53 @@
...
+
+		return true;
+	});
+
+	@FunctionalInterface
+	public interface AllowCapeRender {
+		/**
+		 * @return false to prevent rendering the cape
+		 */
+		boolean allowCapeRender(AbstractClientPlayerEntity clientPlayerEntity);
```
Juuxel:
```suggestion
		boolean allowCapeRender(AbstractClientPlayerEntity player);
```

Juuxel (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
Technici4n: Nevermind, added an `ALLOW` event. 😆

Player3324 (approved):

modmuss50 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,110 @@
...
+		return true;
+	});
+
+	/**
+	 * An event to grant elytra flight to living entities when some condition is met.
+	 * Will be called when players try to start elytra flight by pressing space in mid-air, and every tick for all living entities to check if elytra flight is still allowed.
+	 *
+	 * <p>Items that wish to enable custom elytra flight when worn in the chest equipment slot can simply implement {@link FabricElytraItem} instead of registering a listener.
+	 */
+	public static final Event<Custom> CUSTOM = EventFactory.createArrayBacked(Custom.class, listeners -> (livingEntity, damageElytra) -> {
```
apple502j: `Custom#useCustomElytra` calls the param `tickElytra`.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,110 @@
...
+			if (listener.useCustomElytra(livingEntity, damageElytra)) {
+				return true;
+			}
+		}
+
+		return false;
+	});
+
+	static {
+		CUSTOM.register((livingEntity, damageElytra) -> {
```
apple502j: See above; `damageElytra` vs `tickElytra`

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,110 @@
...
+		 *         // Allow entering/continuing elytra flight with this custom elytra
+		 *         return true;
+		 *     }
+		 *     // Condition for the custom elytra is not met: don't let players enter or continue elytra flight (unless another elytra is available).
+		 *     return false;
+		 * });
+		 * }</pre>
+		 *
+		 * @param livingEntity the entity
+		 * @param tickElytra   false if this just to check if the custom elytra can be used, true if the custom elytra should also be ticked, i.e. use resources.
```
apple502j: "is *this just* to check" - typo?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,110 @@
...
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events related to elytra flight for living entities. Elytra flight is also knows as "fall flying".
```
Juuxel:
```suggestion
 * Events related to elytra flight for living entities. Elytra flight is also known as "fall flying".
```

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/EntityElytraEvents.java
```diff
@@ -0,0 +1,110 @@
...
+			return false;
+		});
+	}
+
+	@FunctionalInterface
+	public interface Allow {
+		/**
+		 * @return false to block elytra flight, true to allow it (unless another listener returns false)
+		 */
+		boolean allowElytraFlight(LivingEntity livingEntity);
```
Juuxel: (`livingEntity` is unnecessarily verbose, but changing it for the sake of bikeshedding is pointless)

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Shadow
+	public abstract void startFallFlying();
+
+	/**
+	 * Allow the server-side and client-side elytra checks to fail when {@link EntityElytraEvents#ALLOW} blocks flight,
+	 * and otherwise to succeed for elytra flight through {@link EntityElytraEvents#CUSTOM}.
+	 */
+	@SuppressWarnings("ConstantConditions")
```
Juuxel, line 47: Hmm, I think these IDEA inspections should rather be suppressed with `// noinspection Xyz` since they're not from javac or APs, but not sure

Juuxel (changes requested):
Looks good, there's just what @apple502j commented as well as a typo.

Technici4n:
Alright, tested on AOF4 with Winged and IR (not sure this one was necessary? probably not) removed, and it works! Both ALLOW and CUSTOM! :)

Juuxel (approved):

Technici4n:
Just squashed to rebase on the buildscript changes from master.

PR #1820, opened 2021-11-11
Fix missing some chunk unloads events leading to a memory leak.

### LINKED ISSUE #1819: Going back and forth between main and outer end island through gateway portal leaks memory on client
java 18(ea build 22)
minecraft 1.17.1
fabric loader 0.12.5
fabric api 0.42.1

and nothing else :p

according to VisualVM by the time i took heap dump this HashSet...

https://github.com/FabricMC/fabric/blob/6d0fac422f959257199eeedfbc224431ce60049d/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/WorldMixin.java#L46

...contained 67140 elements and had retained size of ~1028 megabytes, which doesn't sound right considering that i'm playing on a vanilla'ish(carpetmod) server with view-distance=16

i can share heap dump if it's going to be of any use but this leak seems to be easily reproducible
at least on "my machine™" :D

modmuss50:


Technici4n (approved):

modmuss50:
This also appers to be a possible issue on 1.16

PR #1824, opened 2021-11-13
Fabric Data Generation API

modmuss50:
This PR allows mods to generate JSON data/assets for their mods. See the minimal example bellow:

```java
public class DataGeneratorTestInitializer implements DataGeneratorInitializer {
	@Override
	public void onInitializeDataGenerator(FabricDataGenerator dataGenerator) {
		dataGenerator.addProvider(TestRecipeProvider::new);
	}

	private static class TestRecipeProvider extends FabricRecipeProvider {
		private TestRecipeProvider(FabricDataGenerator dataGenerator) {
			super(dataGenerator);
		}

		@Override
		public void generate(Consumer<RecipeJsonProvider> exporter) {
			offerPlanksRecipe2(exporter, Blocks.ACACIA_PLANKS, ItemTags.ACACIA_LOGS);
		}
	}
}
```

See the test mod here: https://github.com/modmuss50/fabric/blob/datagen/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java for a full example.

There may be more scope for additional data providers and validation down the line, but right now this covers all of the data providers included with vanilla. You can of course create your own data provider if you have custom data.

The datagen can run on the client or server (both headless) through gradle:

```groovy
sourceSets {
	main{
		resources {
			srcDirs += [
				'src/main/generated'
			]
		}
	}
}

loom {
	runs {
		datagen {
			server()

			name "Data Generation"
			vmArg "-Dfabric-api.datagen"
			vmArg "-Dfabric-api.datagen.output-dir=${file("src/main/generated")}"

			runDir "build/datagen"
		}
	}
}
assemble.dependsOn runDatagen
```

I plan to write a wiki page with an more indepth tutorial on how to use it once released.

Please let me know if you see anything wrong/missing. 👍


fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/AccessWidenerGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.nio.file.Paths;
+import java.util.Objects;
+import java.util.function.Predicate;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+
+import net.minecraft.data.server.RecipesProvider;
+
+public class AccessWidenerGenerator {
```
BasiqueEvangelist: Why is the AW generator for the *main* sourceset in the *testmod*?

modmuss50: This is going to move all into gradle, the reason for putting it here is it doesnt need to get shipped with the mod (as its only used by the buildscript)

Player3324 (approved):

Jab125:
I already use datagen, its just very messy [3e2j/goblin-traders-fabric](https://github.com/3e2j/goblin-traders-fabric) has a messy example, i'd be really happy to see this being officially supported

shartte:
I'll try this out in AE2 as soon as I got through the 1.18 porting woes.

modmuss50:
That would be great, I did plan on testing it a little bit in TR (either port TR to 1.18 or backport this PR) but checking it out with AE that has existing data generation would be very useful.

Technici4n:
I want to try this with MI too, with some custom model jsons and some tags to begin with.

Once we have datagen, we can also look into a fabric convention tags module (e.g. add vanilla iron ingots to `c:iron_ingots`, and expose these tags to modders in the API).

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementsProvider.java
```diff
@@ -0,0 +1,78 @@
...
+	private static final Gson GSON = (new GsonBuilder()).setPrettyPrinting().create();
+
+	private final FabricDataGenerator dataGenerator;
+
+	protected FabricAdvancementsProvider(FabricDataGenerator dataGenerator) {
+		this.dataGenerator = dataGenerator;
+	}
+
+	/**
+	 * Use {@link Advancement.Task#build(Consumer, String)} to register and advancement to build.
```
apple502j: typo?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+		this.modContainer = mod;
+		this.strictValidation = strictValidation;
+	}
+
+	/**
+	 * Helper overloaded method to aid with registering a {@link DataProvider} that has a single argument constructor for a {@link FabricDataGenerator}.
+	 *
+	 * @return The {@link DataProvider}
+	 */
+	public <T extends DataProvider> T addProvider(Function<FabricDataGenerator, T> function) {
```
apple502j: Can't we name the param `provider`?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipesProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * Implement this method and then use the range of methods in {@link RecipesProvider} or from one of the recipe json factories such as {@link ShapedRecipeJsonFactory} & {@link ShapelessRecipeJsonFactory}.
+	 */
+	protected abstract void generateRecipes(Consumer<RecipeJsonProvider> exporter);
+
+	@Override
+	public void run(DataCache cache) {
+		Path path = this.root.getOutput();
+		Set<Identifier> generatedRecipes = new HashSet<>();
```
apple502j: Maybe fix the inconsistencies (`Sets.newHashSet();` is used in other files)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,250 @@
...
+	protected abstract void generateTags();
+
+	/**
+	 * Creates a new instance of {@link FabricObjectBuilder} for the given {@link net.minecraft.tag.Tag.Identified} tag.
+	 *
+	 * @param tag The {@link net.minecraft.tag.Tag.Identified} tag to create the builder for
+	 * @return The {@link FabricObjectBuilder} instance
+	 */
+	@Override
+	protected FabricObjectBuilder<T> getOrCreateTagBuilder(Tag.Identified<T> tag) {
```
apple502j: I believe the yarn mapping for this is not great.

Assuming `TagBuilder` refers to `Tag.Builder`, this method does not return `Tag.Builder`. `getTagBuilder` (which actually creates `Tag.Builder`) should be renamed to `getOrCreateTagBuilder` on yarn.

`ObjectBuilder` is also not very good as a name since it's not building an object. (Tag entries have objects in them but ObjectBuilder is not responsible for entry creation, only entry addition.) I'd say it's like `RegistryBoundBuilder` (since that's one big difference between this and the underlying `Tag.Builder`). `getOrCreateTagBuilder` is more like `createBuilder` since `ObjectBuilder` instances are created on every call - and being bound to a registry is just an implementation detail.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,250 @@
...
+		 * Copy the entries from a tag with the {@link Block} type into this item tag.
+		 *
+		 * <p>The {@link Items} tag provider must be constructed with an associated {@link Blocks} tag provider to use this method.
+		 *
+		 * @param blockTag The block tag to copy from.
+		 * @param itemTag The item tag to copy to.
+		 */
+		public void copy(Tag.Identified<Block> blockTag, Tag.Identified<Item> itemTag) {
+			Tag.Builder itemTagBuilder = this.getTagBuilder(itemTag);
+			Tag.Builder blockTagBuilder = Objects.requireNonNull(this.blockTagBuilderProvider, "Pass Block tag provider via constructor to use copy").apply(blockTag);
```
apple502j, line 147: Maybe swap these?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/client/MinecraftMixin.java
```diff
@@ -0,0 +1,43 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.MinecraftClient;
+
+import net.fabricmc.fabric.impl.datagen.FabricDataGenHelper;
+
+@Mixin(MinecraftClient.class)
+public class MinecraftMixin {
```
apple502j: Isn't this `MinecraftClientMixin`?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementsProvider.java
Shnupbups:
```suggestion
	 * Use {@link Advancement.Task#build(Consumer, String)} to register an advancement to build.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
Shnupbups:
```suggestion
	public <T extends DataProvider> T addProvider(Function<FabricDataGenerator, T> provider) {
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
```diff
@@ -0,0 +1,89 @@
...
+import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.entrypoint.EntrypointContainer;
+
+@ApiStatus.Internal
+public class FabricDataGenHelper {
+	private static final Logger LOGGER = LogManager.getLogger();
+
+	/**
+	 * When enabled the dedicated server startup will be hyjacked to run the data generators and then quit.
```
Shnupbups:
```suggestion
	 * When enabled the dedicated server startup will be hijacked to run the data generators and then quit.
```

Mysterious-Dev (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+
+import net.minecraft.data.DataGenerator;
+import net.minecraft.data.DataProvider;
+
+import net.fabricmc.loader.api.ModContainer;
+
+/**
+ * An extension to vanilla's {@link DataGenerator} providing mod specific data, and helper functions.
+ */
+public class FabricDataGenerator extends DataGenerator {
```
BasiqueEvangelist: Is this class supposed to be extendable?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
modmuss50: I agree, the mappings around this are a little odd. Might be worth getting this stuff fixed before merging this PR.

Shnupbups (approved):

oliviathevampire:
Maybe add datagen for sounds and languages like what forge have?

modmuss50:
I can see sounds beign useful, not sure how languages would work however. Need to be careful not to fall into the trap of adding something just becuase forge has it however. Come with a problem and not an existing solution.

Adding new data gens like that is out of scope for this PR, might be worth opening an issue to keep track of other stuff for a follow up PR.

Patbox:
> I can see sounds beign useful, not sure how languages would work however. Need to be careful not to fall into the trap of adding something just becuase forge has it however. Come with a problem and not an existing solution.
> 
> Adding new data gens like that is out of scope for this PR, might be worth opening an issue to keep track of other stuff for a follow up PR.

Languagr could be useful for repeated/similar translations like colored block variants or oxidisable blocks

Technici4n (approved):
Been testing this with MI to generate a few recipes, and it worked perfectly. Going to try other types of resources over time... Thanks a lot for this API! 👍

ghost:
Only complaint so far is the name of `FabricTagProvider.Items` ect since it prevents modders from being able to use un-qualified references to vanilla's `Items` class.

modmuss50:
> Only complaint so far is the name of FabricTagProvider.Items ect since it prevents modders from being able to use un-qualified references to vanilla's Items class.

Thats a good point, I think I started with `Item` but changed it to `Items` without thinking about the other clashes, I will rename it to `ItemTagProvider` (along with the others)

> Languagr could be useful for repeated/similar translations like colored block variants or oxidisable blocks

That does sounds like an intresting use-case, something to think about with further changes down the line. 👍

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipesProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
+
+/**
+ * Extend this class and implement {@link FabricRecipesProvider#generateRecipes}.
+ *
+ * <p>Register an instance of the class with {@link FabricDataGenerator#addProvider} in a {@link net.fabricmc.fabric.api.datagen.v1.DataGeneratorEntrypoint}
+ */
+public abstract class FabricRecipesProvider extends RecipesProvider {
+	private final FabricDataGenerator dataGenerator;
```
Technici4n:
```suggestion
	protected final FabricDataGenerator dataGenerator;
```

Technici4n (changes requested):
Would be nice if these fields could be protected for use in subclasses.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/BlockStateDefinitionProviderMixin.java
```diff
@@ -0,0 +1,117 @@
...
+
+	@Inject(method = "run", at = @At("TAIL"))
+	private void runTail(DataCache cache, CallbackInfo ci) {
+		dataGeneratorThreadLocal.remove();
+	}
+
+	@Inject(method = "method_25738", at = @At("HEAD"), cancellable = true)
+	private static void filterBlocksForProcessingMod(Map<Block, BlockStateSupplier> map, Block block, CallbackInfoReturnable<Boolean> cir) {
+		if (dataGeneratorThreadLocal.get() instanceof FabricDataGenerator dataGenerator) {
+			if (dataGenerator.isStrictValidationEnabled()) {
```
ghost: This should have a `!` before, strict validation is enabled by default...

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTablesProvider.java
```diff
@@ -0,0 +1,102 @@
...
+			generated.add(identifier);
+
+			if (identifier.equals(LootTables.EMPTY)) {
+				continue;
+			}
+
+			biConsumer.accept(entry.getKey(), entry.getValue());
+		}
+
+		if (dataGenerator.isStrictValidationEnabled()) {
```
ghost, line 78: Not sure as to the cause atm but this doesn't seem to work for expanded storage, even when all loot table code is there it complains every block doesn't have a loot table...
Code: <https://github.com/NinjaPhenix/ExpandedStorage/blob/7.3/1.18/fabric-datagen/src/main/java/ninjaphenix/expandedstorage/data/providers/BlockLootProvider.java>

Error:
```
[17:45:55] [Render thread/INFO] (Minecraft) [STDERR]: java.lang.IllegalStateException: Missing loot table(s) for [expandedstorage:netherite_barrel, expandedstorage:obsidian_chest, expandedstorage:obsidian_barrel, expandedstorage:old_wood_chest, expandedstorage:candy_cane_mini_present, expandedstorage:diamond_barrel, expandedstorage:wood_mini_chest, expandedstorage:old_diamond_chest, expandedstorage:present, expandedstorage:red_mini_present_with_sparrow, expandedstorage:diamond_chest, expandedstorage:iron_chest, expandedstorage:netherite_chest, expandedstorage:vanilla_wood_mini_chest, expandedstorage:wood_mini_chest_with_sparrow, expandedstorage:old_obsidian_chest, expandedstorage:wood_chest, expandedstorage:white_mini_present, expandedstorage:red_mini_present, expandedstorage:pumpkin_mini_chest, expandedstorage:pink_amethyst_mini_present, expandedstorage:old_iron_chest, expandedstorage:iron_barrel, expandedstorage:vanilla_wood_mini_chest_with_sparrow, expandedstorage:green_mini_present_with_sparrow, expandedstorage:old_netherite_chest, expandedstorage:gold_barrel, expandedstorage:pumpkin_mini_chest_with_sparrow, expandedstorage:gold_chest, expandedstorage:pink_amethyst_mini_present_with_sparrow, expandedstorage:lavender_mini_present, expandedstorage:white_mini_present_with_sparrow, expandedstorage:candy_cane_mini_present_with_sparrow, expandedstorage:green_mini_present, expandedstorage:lavender_mini_present_with_sparrow, expandedstorage:pumpkin_chest, expandedstorage:old_gold_chest]
```

Pretty sure generated loot table id slightly differs from block id e.g.:
loot table id = `expandedstorage:blocks/wood_chest`
block id = `expandedstorage:wood_chest`

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTablesProvider.java
```diff
@@ -0,0 +1,102 @@
...
+			Set<Identifier> missing = Sets.newHashSet();
+
+			for (Identifier blockId : Registry.BLOCK.getIds()) {
+				if (blockId.getNamespace().equals(dataGenerator.getModId())) {
+					if (!generated.contains(blockId)) {
```
ghost: Generated loot table id never matches the block id so strict validation always fails.

generated loot table id looks like: `expandedstorage:blocks/wood_chest`
block id looks like: `expandedstorage:wood_chest`

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTablesProvider.java
modmuss50: Fixed. Thanks 👍

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+ * @see GameEventTagProvider
+ */
+public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
+	private final String path;
+	private final String name;
+
+	/**
+	 * Construct a new {@link FabricTagProvider}.
+	 *
+	 * <p>Common implementations of this class are provided. For example @see Blocks
```
Shnupbups:
```suggestion
	 * <p>Common implementations of this class are provided. For example @see BlockTagProvider
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+	 * Extend this class to create {@link Block} tags in the "/blocks" tag directory.
+	 */
+	public abstract static class BlockTagProvider extends FabricTagProvider<Block> {
+		public BlockTagProvider(FabricDataGenerator dataGenerator) {
+			super(dataGenerator, Registry.BLOCK, "blocks", "Block Tags");
+		}
+	}
+
+	/**
+	 * Extend this class to create {@link ItemTagProvider} tags in the "/items" tag directory.
```
Shnupbups:
```suggestion
	 * Extend this class to create {@link Item} tags in the "/items" tag directory.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+		 */
+		public void copy(Tag.Identified<Block> blockTag, Tag.Identified<Item> itemTag) {
+			Tag.Builder blockTagBuilder = Objects.requireNonNull(this.blockTagBuilderProvider, "Pass Block tag provider via constructor to use copy").apply(blockTag);
+			Tag.Builder itemTagBuilder = this.getTagBuilder(itemTag);
+			blockTagBuilder.streamEntries().forEach(itemTagBuilder::add);
+		}
+	}
+
+	/**
+	 * Extend this class to create {@link FluidTagProvider} tags in the "/fluids" tag directory.
```
Shnupbups:
```suggestion
	 * Extend this class to create {@link Fluid} tags in the "/fluids" tag directory.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
```diff
@@ -0,0 +1,92 @@
...
+	 */
+	private static final String OUTPUT_DIR = System.getProperty("fabric-api.datagen.output-dir");
+
+	/**
+	 * When enabled providers can enable extra validation, such as ensuring all registry entries have data generated for them.
+	 */
+	private static final boolean STRICT_VALIDATION = System.getProperty("fabric-api.datagen.strict-validation") != null;
+
+	/**
+	 * Filter to a specific modid with this property, useful if dependencies also have data generators.
```
Shnupbups: Use of MOD_ID as separate words in the actual variable name, but modid as one word in the doc is inconsistent. Mod ID should be separate words, with ID capitalized in docs.
```suggestion
	 * Filter to a specific mod ID with this property, useful if dependencies also have data generators.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipesProvider.java
```diff
@@ -0,0 +1,79 @@
...
+			JsonObject jsonObject = provider.toAdvancementJson();
+
+			if (jsonObject != null) {
+				saveRecipeAdvancement(cache, jsonObject, path.resolve("data/" + identifier.getNamespace() + "/advancements/" + provider.getAdvancementId().getPath() + ".json"));
+			}
+		});
+	}
+
+	/**
+	 * Override this method to change the recipe identifier. The default implementation normalises the namespace to the modid.
```
Shnupbups: Don't we use American English? Also mod ID should be separate words, with ID capitalized in docs.
```suggestion
	 * Override this method to change the recipe identifier. The default implementation normalizes the namespace to the mod ID.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+	 * Returns the {@link ModContainer} for the mod that this data generator has been created for.
+	 *
+	 * @return a {@link ModContainer} instance
+	 */
+	public ModContainer getModContainer() {
+		return modContainer;
+	}
+
+	/**
+	 * Returns the modid for the mod that this data generator has been created for.
```
Shnupbups: ModId implies separate words, and they **should** probably be separate words anyway in docs, with ID capitalized.
```suggestion
	 * Returns the mod ID for the mod that this data generator has been created for.
```

Shnupbups (changes requested):
JavaDoc fixes

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementsProvider.java
```diff
@@ -0,0 +1,78 @@
...
+			if (!identifiers.add(advancement.getId())) {
+				throw new IllegalStateException("Duplicate advancement " + advancement.getId());
+			}
+
+			DataProvider.writeToPath(GSON, cache, advancement.createTask().toJson(), getOutputPath(advancement));
+		}
+	}
+
+	private Path getOutputPath(Advancement advancement) {
+		return dataGenerator.getOutput().resolve("data/%s/advancements/%s.json".formatted(advancement.getId().getNamespace(), advancement.getId().getPath()));
```
Technici4n, line 73: Should use the mod id and not the namespace of the advancement.

Technici4n (changes requested):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementsProvider.java
Technici4n: Could also make this method (and similar ones if they exist) protected.

modmuss50: The advancement should have been already created with the mod id?

Technici4n: Yeah just checked, and that's correct, I should create it with the mod id.

Technici4n: image

I was using this method, but if the modid is not right, child advancements will use the `minecraft` namespace to the parent anyway. So no need to change the id indeed.

modmuss50: Yeah, slightly annoying that it takes a string and not an ID, would be more effort than its worth to find a solution to that.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+	 * @return a {@link ModContainer} instance
+	 */
+	public ModContainer getModContainer() {
+		return modContainer;
+	}
+
+	/**
+	 * Returns the mod ID for the mod that this data generator has been created for.
+	 *
+	 * @return a modid
```
Shnupbups: missed one!
```suggestion
	 * @return a mod ID
```

modmuss50: Thanks for that 👍

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -0,0 +1,33 @@
...
+package net.fabricmc.fabric.api.datagen.v1;
+
+/**
+ * An entry point for data generation.
+ *
+ * <p>In {@code fabric.mod.json}, the entrypoint is defined with {@code fabric-datagen} key.</p>
+ *
+ * @see FabricDataGenerator
+ */
+public interface DataGeneratorEntrypoint {
```
Juuxel, line 27:
```suggestion
@FunctionalInterface
public interface DataGeneratorEntrypoint {
```
The default entrypoint adapter supports implementing these via "method reference" strings, so `@FunctionalInterface` would make sense.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementsProvider.java
```diff
@@ -0,0 +1,78 @@
...
+	protected final FabricDataGenerator dataGenerator;
+
+	protected FabricAdvancementsProvider(FabricDataGenerator dataGenerator) {
+		this.dataGenerator = dataGenerator;
+	}
+
+	/**
+	 * Use {@link Advancement.Task#build(Consumer, String)} to register an advancement to build.
+	 */
+	public abstract void generateAdvancement(Consumer<Advancement> consumer);
```
Juuxel, line 54: I'd add better docs here (what this method does in the first place, and that the consumer is used for adding the advancements)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTablesProvider.java
```diff
@@ -0,0 +1,98 @@
...
+	protected final FabricDataGenerator dataGenerator;
+
+	protected FabricBlockLootTablesProvider(FabricDataGenerator dataGenerator) {
+		this.dataGenerator = dataGenerator;
+	}
+
+	/**
+	 * Use the range of {@link BlockLootTableGenerator#addDrop} methods to generate block drops.
+	 */
+	protected abstract void generateBlockLootTables();
```
Juuxel, line 52: This also would benefit from actually describing the method as the first line of JD.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+			parent.add(element);
+			return this;
+		}
+
+		/**
+		 * Add an optional {@link Identifier} to the tag.
+		 *
+		 *
+		 * @return the {@link FabricTagBuilder} instance
+		 */
```
```suggestion
		/**
		 * Add an optional {@link Identifier} to the tag.
		 *
		 * @return the {@link FabricTagBuilder} instance
		 */
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+			((net.fabricmc.fabric.impl.datagen.FabricTagBuilder) builder).fabric_setReplace(replace);
+			return this;
+		}
+
+		/**
+		 * Add a single element to the tag.
+		 *
+		 *
+		 * @return the {@link FabricTagBuilder} instance
+		 */
```
```suggestion
		/**
		 * Add a single element to the tag.
		 *
		 * @return the {@link FabricTagBuilder} instance
		 */
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		/**
+		 * Add another tag to this tag.
+		 *
+		 * @return the {@link FabricTagBuilder} instance
+		 */
+		@Override
+		public FabricTagBuilder<T> addTag(Tag.Identified<T> identifiedTag) {
+			parent.addTag(identifiedTag);
```
```suggestion
		public FabricTagBuilder<T> addTag(Tag.Identified<T> tag) {
			parent.addTag(tag);
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricLootTableProvider.java
```diff
@@ -0,0 +1,76 @@
...
+import net.minecraft.loot.LootTable;
+import net.minecraft.loot.context.LootContextType;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.datagen.v1.provider.FabricBlockLootTablesProvider;
+import net.fabricmc.fabric.api.datagen.v1.provider.SimpleFabricLootTableProvider;
+import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
+
+/**
+ * A base interface for Loot table providers. You should not implement this class directly.
```
Juuxel: Should this really be internal or just an `@ApiStatus.NonExtendable` (or whatever it was) API package interface, since the functionality here is exposed to modders?

Juuxel (changes requested):

fabric-data-generation-api-v1/build.gradle
```diff
@@ -0,0 +1,109 @@
...
+
+import org.objectweb.asm.ClassReader
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.tree.ClassNode
+
+import java.util.zip.ZipFile
+
+task generateAccessWidener() {
+	doLast {
+		File inputJar = loom.mappingsProvider.mappedProvider.baseMappedJar
```
Juuxel, line 59: Also, is Fabric API setting a bad example to modders here by accessing loom internals? 😛

modmuss50: Yes, but this is not an example ;) I might put a comment on it.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricLootTableProvider.java
modmuss50: This is really impl, but api classes extend from it. I will make it NonExtendable and move it to the api.

fabric-data-generation-api-v1/build.gradle
```diff
@@ -0,0 +1,110 @@
...
+moduleDependencies(project, [
+		'fabric-api-base',
+		'fabric-registry-sync-v0',
+		'fabric-networking-api-v1'
+])
+
+sourceSets {
+	testmod {
+		java {
+			srcDirs += [
```
oliviathevampire, line 13: Why is this java and not resources?

modmuss50: Fixed.

Shnupbups (approved):

Juuxel (approved):
Looks great! A (probably) simple future addition could be a preset for biome tag providers, since FAPI adds support for them.

JosephYapYEEEEEEEEEEEEEEEEEEEEEEEEEEEET (approved):

AlphaMode (approved):

liach:
ah sorry, github somehow send me another commit in its feed and didn't notice this. sorry for asking for the review pr again!

PR #1834, opened 2021-11-16
Restored Nether/End- and Biome Modification API

shartte:
Restored functionality back to what it was during the snapshot. And I took the liberty of converting the biome modification API from accessors to AWs since this much more easily exposes the compilation issues that accessors hide.

Mojang essentially split decorators off from ConfiguredFeature and introduced `class_6796`, which is "placed feature" (-> Yarn names this `Registry.PLACED_FEATURE_WORLDGEN`). So in essence, the modification API now requires modders to add registered placed features instead of configured features into biomes.

Glaring omissions right now:

- No overworld biome addition API
- No way to easily modify the material rules used for surface builders (essentially the chunk generator now has a rule-list that says how to generate biomes surfaces)

I tested feature generation in the overworld (custom desert well), that nether and end biomes spawn, and custom sky color modifications.


fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/modification/BiomeModificationContextImpl.java
```diff
@@ -120,109 +112,98 @@ void freeze() {
...
 		public void setTemperature(float temperature) {
-			accessor.setTemperature(temperature);
+			weather.temperature = temperature;
 		}
 
 		@Override
 		public void setTemperatureModifier(Biome.TemperatureModifier temperatureModifier) {
-			Objects.requireNonNull(temperatureModifier);
-			accessor.setTemperatureModifier(temperatureModifier);
+			weather.temperatureModifier = Objects.requireNonNull(temperatureModifier);
```
maityyy, line 129: I think there are not enough custom messages here
```
Objects.requireNonNull("temperatureModifier cannot be null", temperatureModifier);
```

modmuss50 (approved):

PR #1836, opened 2021-11-16
Entity API Lookup

deirn:
`EntityType#create` requires a world instance to call, making `registerSelf` check needs to be made on the server started event, I wonder if it's worth the trouble checking in the first place.

Also, do we want to include some helper method for querying API from `World`+`BlockPos`? That'll be like how Hopper checks for container minecarts (`HopperBlockEntity#getInventoryAt`).

The testmod reuses the `Inspectable`, that's why I moved things around.

### DIFF
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
@@ -45,7 +45,7 @@ import net.fabricmc.fabric.impl.lookup.entity.EntityApiLookupImpl;
  * <p>We need to create the EntityApiLookup. We don't need any context so we use {@link Void}.
  * <pre>{@code
  * public class MyApi {
- *     public static final EntityApiLookup<Leveled, Void> LEVELED_ENTITY = EntityApiLookup.get(new Identifier("mymod:leveled_entity), Leveled.class, Void.class);
+ *     public static final EntityApiLookup<Leveled, Void> LEVELED_ENTITY = EntityApiLookup.get(new Identifier("mymod:leveled_entity"), Leveled.class, Void.class);
  * }
  * }</pre>
  *
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
@@ -66,11 +66,11 @@ import net.fabricmc.fabric.impl.lookup.entity.EntityApiLookupImpl;
  * }
  * MyApi.LEVELED_ENTITY.registerSelf(LEVELED_PIG_ENTITY_TYPE);
  *
- * // Otherwise, registerForTypes can be used.
- * MyApi.LEVELED_ENTITY.registerForTypes((entity, ignored) -> {
+ * // Otherwise, registerForType can be used.
+ * MyApi.LEVELED_ENTITY.registerForType((zombieEntity, ignored) -> {
  *     // Return a Leveled instance for your entity here, or null if there's none.
  *     // The context is Void in this case, so it can be ignored.
- * }, EntityType.PLAYER, EntityType.ZOMBIE);
+ * }, EntityType.ZOMBIE);
  *
  * // Generic fallback, to interface with anything, for example if we want to all other entity level defaults to 1.
  * MyApi.LEVELED_ENTITY.registerFallback((entity, ignored) -> {
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.lookup.v1.entity;
+
+import java.util.function.BiFunction;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityType;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.lookup.entity.EntityApiLookupImpl;
+
+/**
+ * An object that allows retrieving APIs from entities.
+ * Instances of this interface can be obtained through {@link #get}
+ *
+ * <p>When trying to {@link #find} an API for an entity, the provider registered for the entity type will be queried if it exists.
+ * If it doesn't exist, or if it returns {@code null}, the fallback providers will be queried in order.
+ *
+ * <p><h3>Usage Example</h3>
+ * Let's pretend that we have the following interface that we want to attach to entities.
+ * <pre>{@code
+ * public interface Leveled {
+ *     int getLevel();
+ * }
+ * }</pre>
+ *
+ * <p>We need to create the EntityApiLookup. We don't need any context so we use {@link Void}.
+ * <pre>{@code
+ * public class MyApi {
+ *     public static final EntityApiLookup<Leveled, Void> LEVELED_ENTITY = EntityApiLookup.get(new Identifier("mymod:leveled_entity), Leveled.class, Void.class);
+ * }
+ * }</pre>
+ *
+ * <p>Now we can query instances of {@code Leveled}.
+ * <pre>{@code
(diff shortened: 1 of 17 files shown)


fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
```diff
@@ -0,0 +1,176 @@
...
+ * <pre>{@code
+ * public interface Leveled {
+ *     int getLevel();
+ * }
+ * }</pre>
+ *
+ * <p>We need to create the EntityApiLookup. We don't need any context so we use {@link Void}.
+ * <pre>{@code
+ * public class MyApi {
+ *     public static final EntityApiLookup<Leveled, Void> LEVELED_ENTITY = EntityApiLookup.get(new Identifier("mymod:leveled_entity), Leveled.class, Void.class);
```
Juuxel:
```suggestion
 *     public static final EntityApiLookup<Leveled, Void> LEVELED_ENTITY = EntityApiLookup.get(new Identifier("mymod:leveled_entity"), Leveled.class, Void.class);
```

Technici4n (changes requested):
Hi, thanks for tackling Entity API Lookup.

Having that said, I don't think the approach used for blocks and items makes sense for entities: it is often desirable to expose multiple instances of the API for entities. Consider for example a player with a fluid backpack and a fluid hat. You'd ideally want both storages, and most of the time you'd like them to be wrapped in a combined storage.

I would therefore suggest that query would return a `List<T>`, and that each non-null API would be appended to the list. This also implies that all fallback providers would run for every query.
We should specify under which circumstances the list is mutable or not. Imo we can specify that the list is mutable, in case calling code wants to reuse it for some reason.

As far as API usability is concerned, I think providing the following set of query functions should be nice enough to use in practice:
```java
public interface EntityApiLookup<A, C> {
    List<A> findAll(Entity entity, C context);
    @Nullable
    default findFirst(Entity entity, C context) {
        List<A> list = findAll(entity, context);
        return list.size() > 0 ? list.get(0) : null;
    }
    @Nullable
    T find(Entity entity, C context, Function<List<A>, A> combiner) {
        List<A> list = findAll(entity, context);
        return list.size() > 0 ? combiner.apply(list) : null;
    }
}
```

Thoughts? We can discuss this a bit before you implement this change. I'd especially wait for @sfPlayer1's input.

Technici4n:
After some discussion with Player on discord, we can implement the dispatching across multiple providers ourselves under the hood in the cases where it is needed. I have done a similar thing on top of `ItemApiLookup`: look at `ItemStorage#combinedItemApiProvider(Item)` and `ItemStorage#GENERAL_COMBINED_PROVIDER`. This means that you can basically ignore my previous comment.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
```diff
@@ -0,0 +1,176 @@
...
+ * public class LeveledPigEntity extends PigEntity implements Leveled {
+ *     ...
+ * }
+ * MyApi.LEVELED_ENTITY.registerSelf(LEVELED_PIG_ENTITY_TYPE);
+ *
+ * // Otherwise, registerForTypes can be used.
+ * MyApi.LEVELED_ENTITY.registerForTypes((entity, ignored) -> {
+ *     // Return a Leveled instance for your entity here, or null if there's none.
+ *     // The context is Void in this case, so it can be ignored.
+ * }, EntityType.PLAYER, EntityType.ZOMBIE);
```
Technici4n: Should probably use `registerForType` as the example as it will allow using the entity subclass directly (I know this isn't done in BlockApiLookup, if you want to fix it there as well why not).

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/package-info.java
```diff
@@ -59,6 +59,20 @@
  * </ul>
  * </p>
  *
+ * <p><h2>Retrieving APIs from entities</h2>
+ * <ul>
+ *     <li>A query for an entity API takes an entity and additional context of type {@code C},
+ *     and uses that to find an object of type {@code A}, or  {@code null} if there's no such object.</li>
+ *     <li>{@link net.fabricmc.fabric.api.lookup.v1.entity.EntityApiLookup EntityApiLookup&lt;A, C&gt;} instances provide a
+ *     {@link net.fabricmc.fabric.api.lookup.v1.entity.EntityApiLookup#find find()} function that does the query, and registration happens
+ *     primary trough {@link net.fabricmc.fabric.api.lookup.v1.entity.EntityApiLookup#registerSelf registerSelf()} and
```
Technici4n:
```suggestion
 *     primarily through {@link net.fabricmc.fabric.api.lookup.v1.entity.EntityApiLookup#registerSelf registerSelf()} and
```

fabric-api-lookup-api-v1/src/testmod/java/net/fabricmc/fabric/test/lookup/FabricApiLookupTest.java
```diff
@@ -35,9 +36,10 @@
 import net.fabricmc.fabric.test.lookup.api.ItemInsertable;
 import net.fabricmc.fabric.test.lookup.compat.InventoryExtractableProvider;
 import net.fabricmc.fabric.test.lookup.compat.InventoryInsertableProvider;
+import net.fabricmc.fabric.test.lookup.entity.FabricEntityApiLookupTest;
 import net.fabricmc.fabric.test.lookup.item.FabricItemApiLookupTest;
 
-public class FabricApiLookupTest implements ModInitializer {
+public class FabricApiLookupTest implements ModInitializer, ClientModInitializer {
```
Technici4n: Please move the client-only entrypoint and code to a separate class.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/entity/EntityApiLookupImpl.java
```diff
@@ -0,0 +1,162 @@
...
+						throw new IllegalArgumentException(errorMessage);
+					}
+				}
+			});
+		}
+	}
+
+	@Override
+	@Nullable
+	public A find(Entity entity, C context) {
```
Technici4n, line 98: Should this check for `EntityPredicates.VALID_ENTITY.matches(entity)` (i.e. `entity.isAlive()`) ? Not sure, but I noticed that hoppers perform this check. I'd say it's better to add it now and then remove it later if it causes issues.

fabric-api-lookup-api-v1/src/testmod/java/net/fabricmc/fabric/test/lookup/InspectorBlock.java
```diff
@@ -46,4 +52,15 @@ public ActionResult onUse(BlockState state, World world, BlockPos pos, PlayerEnt
 
 		return ActionResult.PASS;
 	}
+
+	@Override
+	public void onSteppedOn(World world, BlockPos pos, BlockState state, Entity entity) {
+		Inspectable inspectable = FabricEntityApiLookupTest.INSPECTABLE.find(entity, null);
+
+		if (inspectable != null) {
+			for (ServerPlayerEntity player : world.getServer().getPlayerManager().getPlayerList()) {
```
Technici4n: `world.getServer()` returns `null` for client-side world. Probably missing a `!world.isClient` check in the if?

Technici4n (commented):
Tested and it (almost) works. Looks fantastic, amazing work! Besides a few remarks, this is good for me. Thanks a lot :)

I don't think methods to query an API in a box are particularly helpful atm, maybe we can consider adding them later.

The javadoc style is not consistent with the rest of the module, but it is consistent with yarn conventions, so I don't mind.

fabric-api-lookup-api-v1/src/testmod/java/net/fabricmc/fabric/test/lookup/InspectorBlock.java
deirn: Oops, I was using block entity tick before I found about `onSteppedOn` :sweat_smile:

Technici4n (approved):
Perfect, thank you once again! 😄

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/entity/EntityApiLookupImpl.java
```diff
@@ -0,0 +1,165 @@
...
+import net.minecraft.util.registry.Registry;
+
+import net.fabricmc.fabric.api.lookup.v1.custom.ApiLookupMap;
+import net.fabricmc.fabric.api.lookup.v1.custom.ApiProviderMap;
+import net.fabricmc.fabric.api.lookup.v1.entity.EntityApiLookup;
+
+public class EntityApiLookupImpl<A, C> implements EntityApiLookup<A, C> {
+	private static final Logger LOGGER = LogManager.getLogger("fabric-api-lookup-api-v1/entity");
+	private static final ApiLookupMap<EntityApiLookup<?, ?>> LOOKUPS = ApiLookupMap.create(EntityApiLookupImpl::new);
+	private static final Map<Class<?>, Set<EntityType<?>>> REGISTERED_SELVES = new HashMap<>();
```
Technici4n, line 46: There is a race condition with this map if entities are registered from multiple threads. I'd recommend putting every access in a synchronized(MAP) block.

deirn:
Honestly, I'm a little confused as to why we need to check the type on `registerSelf` (for all the lookups, not only entity). The `ClassCastException` message should be clear enough. If people ended up using it with non-implementing types then that's on them ¯\\\_(ツ)\_/¯.

Technici4n:
The point is to provide fail-fast checking as we can't rely on the type system for that particular path. With the current impl, any mistake should be caught at init or server startup time, so a single gametest (in mods that make use of the API) will ensure that everything is ok.

Player3324 (approved):

modmuss50 (approved):
Looks great, many thanks 👍

Juuxel (approved):

PR #1839, opened 2021-11-16
Fixes double-invoking of the client start event

shartte:
This is due to vanilla now calling thread.setPriority (thus accessing the field twice).


Player3324 (approved):
to really leverage the atomicboolean properly you'd want to do `if (!startCalled.compareAndSet(false, true)) throw ...`, otherwise there's still a small race

modmuss50 (approved):

shartte:
I replaced the atomic booleans by fields in the testmod. There's no concurrency going on at that point anyway.

Technici4n (approved):

PR #1844, opened 2021-11-17
Add a handful of useful worldgen transitive access-wideners.

modmuss50:
Mostly a good test of the transitive access wideners, and becuase I needed these. ;)

Does bump the min loader to 0.12, and the useage of v2 AW does force loom 0.10, as this PR is 1.18 only its a good place to test before emposing those requirements on 1.17 IMO.


maityyy:
How about `TrunkPlacerType.register()`?

Player3324 (approved):

kanpov:
> How about `TrunkPlacerType.register()`?

I agree, I tried to do so with #1507 , but it got abandoned. Now that we have exposed access wideners, such changes are way easier.

As of now, `TrunkPlacer`s, `FoliagePlacer`s etc. can be implemented into your mod by mixins or with a third-party library, [like mine](https://github.com/RedGrapefruit09/Arctree).
It would be nice to have this built into the API.

kanpov (changes requested):
As @liach has already said, it would be nice for this PR to add support for:

- `TrunkPlacerType.register`
- `FoliagePlacerType.register`
- `TreeDecoratorType.register`

And others, mentioned in #1507 .

modmuss50:
I will have to check but those register methods usually only take a string in. The way that mods should register them is directly with the registry

modmuss50:
Yes, none of those register methods should be used by modders. I will make sure the constructor is public however.

RedGrapefruit09:
> Yes, none of those register methods should be used by modders. I will make sure the constructor is public however.

I partially agree with you here, but the register methods, aside from registering, also create the instances.
The constructors for these classes are private, so `register` is the only way to deal with them.

kanpov (changes requested):
Could we please have `TreeDecoratorType` too? Sorry for being annoying.

modmuss50:
Is it not already there? https://github.com/FabricMC/fabric/pull/1844/files#diff-5a9e2631f07f9bd28b850b333091d172aee1163d1e444fb1ef1301d1c5f52571R82

kanpov:
@modmuss50 ok i'm dumb

kanpov (approved):

kanpov:
@modmuss50 why was this PR reverted?

modmuss50:
It was decided after I merged it that we are going to make a whole module just for generic transtive AWs

kanpov:
> It was decided after I merged it that we are going to make a whole module just for generic transtive AWs

what's the status on that? sounds exciting.

PR #416, opened 2019-10-31
Registry sync packet compression

JamiesWhiteShirt:
As described in #410, we have a problem with the registry sync packet being too large. Specifically, the NBT compound that contains the whole state that is sent to the client can be too large to deserialize. The registries must be very bloated for this to be a problem, but we have seen that it can happen in certain setups.

This packet is compressible, and compressing it should be beneficial even if the registries aren't excessively large. This reduces the size of the "vanilla" registry sync message by an estimated 45%, and in a specific modded case reduces it by an estimated 40%. It should also avoid the exception described in the issue. However, we will not have backward compatibility until we can actually perform some sort of handshake.

Instead of using NBT, this uses the byte buffer directly, which avoids some overhead both in processing and in raw byte size, and allows some compression tricks to improve it even further.

This proposed solution uses some assumptions about the state of the registries, namely:

1. Registry objects are grouped into a few namespaces where each namespace generally has multiple objects. Generally, each mod places multiple objects into its own one namespace.
2. The raw IDs of objects within a namespace are somewhat densely packed. Generally, mods register objects in bulks under the same namespace, where objects in a bulk are assigned raw IDs in series.

For each registry, registry objects are grouped by namespace so that the namespace string does not have to be re-written for each object in the namespace. The objects within each namespace are written in ascending order of raw IDs. The raw IDs within a group are written with differential encoding using varints.

Note that vanilla will by default compress messages using a Deflater. I have been unable to measure the true compressed size, but I have manually deflated the packets and measured their length which should be equivalent.

| Packet | Raw size (bytes) | Deflated size (bytes) |
| - | ---: | ---: |
| Vanilla 19w44a (Old) | 105805 | 27219 |
| Vanilla 19w44a (New) | 57283 | 15030 |
| Vanilla 19w44a + Painting Mod (Old) | 322399 | 66795 |
| Vanilla 19w44a + Painting Mod (New) | 185099 | 39566 |

It is a bit tricky to measure this with mods when targeting 19w44a, as many mods have not been updated, but I tested with Painting Mod, which is notorious for filling the registries.

Note that the differential encoding doesn't affect the raw byte size that much, but without it, deflate is less effective.

Issue #410, opened 2019-10-21
registry sync (1.14.4+) vanilla packet exceeded 


SDUBZ:
Logs: 
Server: http://puu.sh/EkPKJ/2166c5a066.log
Client:  http://puu.sh/EkPLA/6dafd81fa4.log

Screenshots: 
http://puu.sh/Ekm8u/9e325dfe3d.png

Will add more info as needed or if things change

LemmaEOF:
`[00:18:54] [Netty Local Client IO #0/WARN]: Failed to handle packet fabric:registry/sync!
java.lang.RuntimeException: Tried to read NBT tag that was too big; tried to allocate: 2097164bytes where max allowed: 2097152`

Looks like registry sync might need to split up its registry sync into multiple packets if there are too many things registered.

LemmaEOF:
Also, there's a big red flag that so many blocks are logging `Adding <id> to saved/remote registry`, since it looks like [that's only supposed to happen when there's a problem](https://github.com/FabricMC/fabric/blob/b02deb007189239ffe6bcc9e53480ef618a95074/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/MixinIdRegistry.java#L237). I'm not quite sure what's causing it in this scenario.

SDUBZ:
I will update logs etc next time i preform updates etc to keep info current to pack changes etc :)

SDUBZ:
log: http://puu.sh/EvwOa/215996578c.log
wait is this crashing single player now too
Failed to handle packet fabric:registry/sync!
java.lang.RuntimeException: Tried to read NBT tag that was too big; tried to allocate: 2097256bytes where max allowed: 2097152

Now affecting singleplayer

LemmaEOF:
Yeah, the solution to this is to split up sync packets if needed. Shouldn't be much of an issue.

Pyrofab:
Not really, you can cut down a lot on the packet size by using proper compression. Experiments have been made, but sadly I don't think any has made its way into a PR.

natanfudge:
@jamieswhiteshirt?

JamiesWhiteShirt:
The current packet is unnecessarily bloated and should not be using NBT. It was likely made this way for parity with registry persistence. The solution is more likely optimizing it for compression, which I have already experimented with.

[...]

PR #1853, opened 2021-11-23
Registry sync packet compression II

deirn:
The compression method is based on #416, with some changes:
- This PR also groups all registry id with same namespace, alongside grouping the object id.
- This PR sends empty string in place of `minecraft` namespace :tiny_potato:
- This PR changes how to send the bulk rawIds. Since most mods registers all at once, the rawId of those objects would be adjacent to each other (the differential would be 1). We can skip sending those differential varints by sending the difference of the first raw id and the last raw id of the bulk before, alongside the size of each bulk. Then let the client loop and increment the raw id.

This PR also keeps the old NBT-based packet to support old server/client.

Number comparison is with [AOF4](https://www.curseforge.com/minecraft/modpacks/all-of-fabric-4/files/3526987) plus [Concistency+](https://www.curseforge.com/minecraft/mc-mods/consistency/files/3510835), which is a mod that has [a fair amount](https://github.com/ConsistencyPlus/ConsistencyPlus/blob/0.4.1/src/main/java/io/github/consistencyplus/consistency_plus/registry/CPlusBlocks.java) of reg objects ;-)
| Packet | Raw size | Deflated size |
| :-: | -: | -: |
| NBT (Old) | `768102` | `210493` |
| Direct (New) | `400934` | `101688` |

Closes #416 
Edit: had a big brain moment and saved some more bytes lol


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -38,55 +38,70 @@
...
+			return;
+		}
 
-		if (tag == null) {
-			return null;
+		if (ServerPlayNetworking.canSend(player, DirectRegistrySyncPacket.ID)) {
+			sendPacket(player, DirectRegistrySyncPacket.getInstance());
+		} else {
+			LOGGER.warn("Player {} can't receive direct packet, using nbt packet instead", player.getEntityName());
+			sendPacket(player, NbtRegistrySyncPacket.getInstance());
```
liach: Should we consider kicking the client if it cannot accept the legacy registry sync as well, as otherwise the client will have a glitchy experience?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/FabricRegistryInit.java
```diff
@@ -21,10 +21,16 @@
 import net.fabricmc.api.ModInitializer;
 import net.fabricmc.fabric.api.event.registry.RegistryAttribute;
 import net.fabricmc.fabric.api.event.registry.RegistryAttributeHolder;
+import net.fabricmc.fabric.api.networking.v1.ServerPlayConnectionEvents;
 
 public class FabricRegistryInit implements ModInitializer {
 	@Override
 	public void onInitialize() {
+		ServerPlayConnectionEvents.JOIN.register((handler, sender, server) -> {
+			// TODO: If integrated and local, don't send the packet (it's ignored)
```
liach: You can check if this server is integrated and this sender is server's owner

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -38,55 +38,70 @@
...
+import net.fabricmc.fabric.impl.registry.sync.packet.DirectRegistrySyncPacket;
+import net.fabricmc.fabric.impl.registry.sync.packet.NbtRegistrySyncPacket;
+import net.fabricmc.fabric.impl.registry.sync.packet.RegistrySyncPacket;
 
 public final class RegistrySyncManager {
 	static final boolean DEBUG = System.getProperty("fabric.registry.debug", "false").equalsIgnoreCase("true");
-	static final Identifier ID = new Identifier("fabric", "registry/sync");
 	private static final Logger LOGGER = LogManager.getLogger("FabricRegistrySync");
 	private static final boolean DEBUG_WRITE_REGISTRY_DATA = System.getProperty("fabric.registry.debug.writeContentsAsCsv", "false").equalsIgnoreCase("true");
+	private static final boolean FORCE_NBT_SYNC = System.getProperty("fabric.registry.forceNbtSync", "false").equalsIgnoreCase("true");
```
liach: I think we can just use `Boolean.getBoolean("fabric.registry.forceNbtSync")`

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/RegistrySyncPacket.java
```diff
@@ -0,0 +1,33 @@
...
+package net.fabricmc.fabric.impl.registry.sync.packet;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.network.PacketByteBuf;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.registry.sync.map.RegistryMap;
+
+public interface RegistrySyncPacket {
```
JamiesWhiteShirt: It is misleading to call this a registry sync packet. Implementations are not packets but rather strategies for sending and serializing registry sync packets.

liach (commented):
The result of `writeBuffer` should preferably be cached, especially given preparing the direct registry sync buffer is more costly.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/DirectRegistrySyncPacket.java
```diff
@@ -0,0 +1,189 @@
...
+
+					if (adjacentRawIds.containsKey(rawId - 1)) {
+						rawIdGroup = adjacentRawIds.remove(rawId - 1);
+					} else {
+						rawIdGroup = new ArrayList<>();
+					}
+
+					rawIdGroup.add(id.getPath());
+					adjacentRawIds.put(rawId, rawIdGroup);
+				}
```
JamiesWhiteShirt: I see a few ways the producing part of the loop that iterates over the registries could be made much more efficient.

Grouping by namespace into a sorted list of entries in the namespace first would allow you to use a more efficient strategy to find consecutive ranges of raw IDs. See [this](https://stackoverflow.com/a/55914488) for some inspiration. It could represent each range as a slice of the sorted list. It would also limit what needs to be in memory to one namespace at a time.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/map/IdMap.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class IdMap extends Object2IntLinkedOpenHashMap<Identifier> {
+	/**
+	 * Returns an entry pair list sorted by the rawId.
+	 */
+	public List<Entry<Identifier>> sortedEntryList() {
+		return object2IntEntrySet().stream()
+				.sorted(Comparator.comparingInt(Entry::getIntValue))
+				.toList();
+	}
```
JamiesWhiteShirt: I don't think it is a good idea to extend collection classes. This could just as easily be a method somewhere else that takes the ID map as a parameter.

shartte: agreed

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/map/RegistryMap.java
```diff
@@ -0,0 +1,65 @@
...
+package net.fabricmc.fabric.impl.registry.sync.map;
+
+import java.util.LinkedHashMap;
+
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
+
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.util.Identifier;
+
+public class RegistryMap extends LinkedHashMap<Identifier, IdMap> {
```
shartte: Same as above, I'd not extend collection classes and instead use static helpers

shartte:
How are you testing the numbers? I think there might be some other ways to achieve smaller packet sizes by optimizing more for deflate. Can you potentially provide a binary dump of the NBT packet that you used for your tests?

Technici4n:
I think we should just bump the max packet size, or deflate the packet ourselves before encoding the packet. I wouldn't spend too much time on something that zlib will do almost as well.

shartte:
I'd hypothesize that simply reorganizing the data could already be sufficient in decreasing the compressed size by optimizing for deflate a little. But without being able to run the same extreme case benchmark, I can't test that easily.

deirn:
Added log prints to check the buffer size, probable need to revert that commit later.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -90,6 +97,38 @@ private static void sendPacket(ServerPlayerEntity player, RegistrySyncPacket pac
 	}
 
 	public static void receivePacket(ThreadExecutor<?> executor, RegistrySyncPacket packet, PacketByteBuf buf, boolean accept, Consumer<Exception> errorHandler) {
+		if (DEBUG) {
+			LOGGER.info("{} raw size: {}", packet.getClass().getSimpleName(), buf.readableBytes());
+
+			// see PacketDeflater
```
liach: Hmm, doesn't vanilla deflate custom payloads already? Does deflating again decrease sizes?

deirn: I don't see a way to know what the current packet is being deflated in the server, so instead I manually compress it in the client instead. I think that's what  JamiesWhiteShirt did on their comparison. The client uncompresses the buffer again.

liach: Hmm, if you really want compression, should we build the byte buf and send it into a gzip input stream/output stream for better results, though this might be a silly idea?

deirn: No, this is done to mimic the vanilla compression so we can measure it. That part is copied directly from vanilla code.

Patbox:
> I think we should just bump the max packet size, or deflate the packet ourselves before encoding the packet. I wouldn't spend too much time on something that zlib will do almost as well.

This solution doesn't really fix the problem and will cause issues with vanilla clients (things will error only on vanilla clients, which will make debugging require more work)

Patbox:
Looking that you are already refactoring it, it would be nice if you could add some events/etc for excluding entries from syncing. It would be useful for mods wanting to register things server side without requiring mods on client (see [Polymer](https://github.com/Patbox/polymer))

modmuss50:
IMO that’s out of scope for this PR. I have already looked at that here: https://github.com/FabricMC/fabric/pull/1339

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/DirectRegistrySyncPacket.java
deirn: Changed, how about now?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/DirectRegistryPacketSerializer.java
```diff
@@ -0,0 +1,175 @@
...
+
+			for (Identifier regId : regIds) {
+				buf.writeString(regId.getPath());
+
+				Object2IntMap<Identifier> idMap = map.get(regId);
+
+				// Sort object ids by raw id and group them by namespace
+				Map<String, List<Object2IntMap.Entry<Identifier>>> idNamespaceGroups = idMap.object2IntEntrySet().stream()
+						.sorted(Comparator.comparingInt(Object2IntMap.Entry::getIntValue))
+						.collect(Collectors.groupingBy(e -> e.getKey().getNamespace(), LinkedHashMap::new, Collectors.toList()));
```
JamiesWhiteShirt: It is assumed that each group is sorted by the stream being sorted before being collected into group lists. Consider doing this in the opposite order - group by namespace, then sort each group by raw ID. Sorting after grouping would make the intent clearer. It also reduces the large sort into a few independent smaller sorts, which should be slightly more efficient.

deirn: Upon further reading, `SimpleRegistry`, which is what all registry basically is, is using `rawIdToEntry` list as backing for its iterator, making it already sorted by the raw id. Should we just don't bother re-sorting it here? I already used linked map for the id map in `RegistrySyncManager#createAndPopulateRegistryMap` anyway.

JamiesWhiteShirt: Even if it is the case that SimpleRegistry's iteration order is by ascending raw ID, I don't know if it should be considered part of its Iterable contract or just an implementation detail that may change in the future. We would also have to ensure that items continue to be inserted into the linked map in that order in case we ever want to change `createAndPopulateRegistryMap` or the surrounding logic that provides the parameter to `writeBuffer`. If we are going to rely on this, it should at least be documented.

oliviathevampire:
Progress on this? I'm in huge need for it

modmuss50 (approved):
Pending a lot of testing, I think this looks good.

I think there is minimal risk to world saves with this as its just syncing, and not the saving part of it.

Technici4n:
Shouldn't this PR also tackle splitting the packet if it exceeds a certain size? Otherwise we will encounter the same issue again in the future I think.

modmuss50:
Thats a fair point, its technically not a full solution. Is PR a realistic limit that no one will ever get past? Idk.

At the very least it could accept multiple packets so could have 1 large packet for 1 large registry?

Technici4n:
This brings roughly a x2 improvement given the numbers in the description, which is what it is, but not enough to ensure the issue is solved forever.

deirn:
From what I can tell, the 2MB limit that causes #410 is only for `readNbt`, so the direct method won't trigger that. 
Not really sure what `0x800000` number is in `PacketInflater` but that looks like the limit for the packet itself (8MB if so).

deirn:
Hmm, nevermind, `CustomPayloadS2CPacket` limits the packet to 1MB, making it easier to hit :/

JamiesWhiteShirt:
When talking about size limitations, make sure to clarify if you mean compressed size or uncompressed size.

I don't remember the specifics, but there are indeed two different limitations. We've seen registry sync packets containing an NBT payload exceeding 2MB (I believe) uncompressed, which is the limit for `readNbt`. I believe at some point a new overload for `readNbt` appeared which could bypass this limit. It could have worked as a stopgap, but we have a proper solution now.

If there is a limit of 1MB for CustomPayloadS2CPacket, I'm guessing that's 1MB compressed, otherwise we wouldn't be able to hit the `readNbt` limit. The current NBT implementation deflates quite well, so it could easily fit a >2MB uncompressed payload in less than 1MB compressed. Or could the size limit for CustomPayloadS2CPacket have changed since we last observed [this](https://github.com/FabricMC/fabric/issues/410#issuecomment-544383443)?

----

Either way, if we are going to future proof this, I would suggest sending one packet per registry as a natural split.

deirn:
Alright, it now split into multiple packets **only** if it's necessary. The testmod registry won't exceed 1MB though to make the equality check between nbt and direct packet works, so I added a configurable max payload size (defaults to 1MB).

The way it split packets is to slice it to multiple `MAX_PAYLOAD_SIZE`d buffer and reconstruct the whole buffer in the client. It also send an empty buffer to mark the end of the split.

---
> If there is a limit of 1MB for CustomPayloadS2CPacket, I'm guessing that's 1MB compressed,

The size check is performed in the constructor, so it's probably the uncompressed buffer. I think Mojang added those in 1.17, idk why.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/DirectRegistryPacketHandler.java
```diff
@@ -0,0 +1,244 @@
...
+ * <ul>
+ *     <li>Directly write into the buffer instead of using an nbt;</li>
+ *     <li>Group all {@link Identifier} with same namespace together and only send those unique namespaces once for each group;</li>
+ *     <li>Group adjacent rawIds together and only send the difference of the first rawId and the last rawId of the bulk before.
+ *     This is based on the assumption that mods generally register all of their object at once,
+ *     therefore making the rawIds somewhat densely packed.</li>
+ * </ul>
+ */
+public class DirectRegistryPacketHandler extends RegistryPacketHandler {
+	public static final RegistryPacketHandler INSTANCE = new DirectRegistryPacketHandler();
```
JamiesWhiteShirt: Small nitpick: With `DirectRegistryPacketHandler` now being stateful, it should not be singleton-ish. The `DirectRegistryPacketHandler` instance used at runtime should probably be placed elsewhere, such as in a static field in `RegistrySyncManager`.

deirn: Done :+1:

Player3324 (approved):
LGTM, thanks!

PR #1854, opened 2021-11-23
Rework how DynamicAttributeTool user context is provided, and provide it in more circumstances

Technici4n:
One issue with the new implementation is that it's not re-entrant, but I'm not sure the previous one was. I changed it because it was way too hard to reason about. If re-entrancy is needed, I can save a `Stack<LivingEntity>` in the threadlocal, but I am concerned about some edge cases where we would forget to pop said stack. Perhaps a redirect with a `try/finally` should be preferred to the current pairs of hooks? Let me know what you think!

Also removed the unused (and broken) `MixinToolItem`.


Technici4n:
What I mean by redirects, is whether the following `LivingEntityMixin` code should be preferred to the one in the PR:
```java
@Mixin(LivingEntity.class)
public class LivingEntityMixin {
	@Redirect(
			at = @At(
					value = "INVOKE",
					target = "net/minecraft/item/ItemStack;getAttributeModifiers(Lnet/minecraft/entity/EquipmentSlot;)Lcom/google/common/collect/Multimap;"
			),
			method = "getEquipment"
	)
	private Multimap<EntityAttribute, EntityAttributeModifier> redirectGetAttributeModifiers(ItemStack stack, EquipmentSlot slot) {
		DynamicToolContext.push(this);

		try {
			return stack.getAttributeModifiers(slot);
		} finally {
			DynamicToolContext.pop(this);
		}
	}
}
```

modmuss50:
Are there plans to contiune with this for 1.18.1? Or as it was removed in 1.18.2 shall we close this?

Im unsure if its ready/worth getting into 1.18.1?

Technici4n:
Ah, I had forgotten about this. I'll PR a replacement for some of `DynamicAttributeTool` for 1.18.2 (unless someone else beats me to it).

modmuss50:
Great thanks for that, 👍 Just wanted to double check you still didnt want to get this into 1.18.1, I also didnt think it would be worth the effort.

PR #1866, opened 2021-11-26
Add id getter to ApiLookups and expose the BE in BlockApiCache

### LINKED ISSUE #1838: API Lookups should have an ID getter
Similar to how they have an API class and context class getter.

Technici4n:
Fixes #1838. I deprecated `LookupFactory` for removal because this is a very niche API (I couldn't find any use of it on github besides fabric's own uses), so hopefully we can remove it one day:tm:.

~~Will conflict with #1836, and will need updating to also add the identifier getter to `EntityApiLookup`. That's why it's still a draft.~~ Updated.

### DIFF
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
@@ -75,6 +75,21 @@ public interface BlockApiCache<A, C> {
 	BlockEntity getBlockEntity();
 
 	/**
+	 * Return the lookup this cache is bound to.
+	 */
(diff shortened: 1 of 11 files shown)


shartte (approved):
IMHO this is good. the id is helpful for debugging and reporting purposes too

Technici4n:
I just thought about adding a `BlockApiLookup`, `ServerWorld` and `BlockPos` getter to `BlockApiCache` too. :smile:

liach (approved):

modmuss50 (approved):

Player3324 (approved):

Juuxel (commented):
`getId` would be enough instead of `getIdentifier` - we don't have `Registry.getIdentifier` either 😛 (also inline `{@return}` could be nicer than "Return the..." but that's minor)

Technici4n:
@Juuxel done!

Juuxel (approved):

PR #1872, opened 2021-11-28
Fix AttackBlockCallback on the client not sending a packet to the server

Technici4n:
Attacking a chest with a lava bucket (see testmod) fails without this mixin patch. I also removed all the useless cancels and returns after `ci.setReturnValue` in other methods.


shartte:
I'll be honest that this is hard to review without having a flowchart 😓

Technici4n:
My reasoning is:
- Vanilla sends a packet on block attack, and the event doesn't.
- If you consume or succeed, the server doesn't know so you get a desync.
- Adding the packet seems to fix the issue.

:P

modmuss50 (approved):

Technici4n:
Missed a path, this is still broken in creative mode I think. Shoot :(

PR #1890, opened 2021-12-07
Add datagen support for dynamic registry tags

deirn:
Some dynamic registries have a default registry in `BuiltinRegistries`, making it (somewhat) possible to use it in `FabricTagProvider` ctor. This is not the case for dimension type though, it doesn't have a default static registry.

There are two ways to make a support for dynamic registry tag datagen:
- Make our own `DataProvider` and copy all methods from `AbstractTagProvider` to it, besides the registry checking bit.
- Use a fake registry that always returns `true` on `Registry#containsId` calls, practically skipping the registry check.

I decided to go with the second option.

This PR:
- Adds `FabricTagProvider.DynamicRegistryTagProvider` that accepts a `RegistryKey` instead of `Registry`. It checks for whether the registry is dynamic and throws an exception if not.
- Throws an exception if someone use regular `FabricTagProvider` ctor with a registry from `BuiltinRegistries`.
- Adds `add` methods in `FabricTagBuilder` that accepts `Identifier` instead of registry object. This is the only way to add non-optional entries to dynamic registry tags.
- Throws an exception if someone call `add` with a default registry object on a dynamic registry tag provider.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -210,7 +210,7 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 	/**
 	 * An extension to {@link net.minecraft.data.server.AbstractTagProvider.ObjectBuilder} that provides additional functionality.
 	 */
-	public class FabricTagBuilder<T> extends ObjectBuilder<T> {
+	public final class FabricTagBuilder<T> extends ObjectBuilder<T> {
 		private final AbstractTagProvider.ObjectBuilder<T> parent;
 
 		private FabricTagBuilder(ObjectBuilder<T> parent) {
@@ -319,7 +319,7 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 		 *
 		 * @return the {@link FabricTagBuilder} instance
 		 */
-		public final FabricTagBuilder<T> add(Identifier... ids) {
+		public FabricTagBuilder<T> add(Identifier... ids) {
 			for (Identifier id : ids) {
 				add(id);
 			}
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -189,7 +189,7 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 	}
 
 	/**
-	 * Extend this class to create a dynamic registry tags.
+	 * Extend this class to create dynamic registry tags.
 	 */
 	public abstract static class DynamicRegistryTagProvider<T> extends FabricTagProvider<T> {
 		/**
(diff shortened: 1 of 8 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -177,13 +188,32 @@ public GameEventTagProvider(FabricDataGenerator dataGenerator) {
 		}
 	}
 
+	/**
+	 * Extend this class to create a dynamic registry tags.
```
```suggestion
	 * Extend this class to create dynamic registry tags.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -177,13 +188,32 @@ public GameEventTagProvider(FabricDataGenerator dataGenerator) {
...
+			Preconditions.checkArgument(DynamicRegistryManagerAccessor.getInfos().containsKey(registryKey), "Only dynamic registries are supported in this tag provider.");
+		}
+	}
+
 	/**
 	 * An extension to {@link net.minecraft.data.server.AbstractTagProvider.ObjectBuilder} that provides additional functionality.
 	 */
-	public static class FabricTagBuilder<T> extends ObjectBuilder<T> {
-		private final AbstractTagProvider.ObjectBuilder<T> parent;
+	public class FabricTagBuilder<O> extends ObjectBuilder<O> {
```
Juuxel: IMO this `T` -> `O` rename is a bit pointless

fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java
```diff
@@ -111,6 +138,25 @@ protected void generateTags() {
...
+	private static class TestBiomeTagsProvider extends FabricTagProvider.DynamicRegistryTagProvider<Biome> {
+		private TestBiomeTagsProvider(FabricDataGenerator dataGenerator) {
+			super(dataGenerator, Registry.BIOME_KEY, "biomes", "Biome Tags");
+		}
+
+		@Override
+		protected void generateTags() {
+			FabricTagBuilder<Biome> builder = getOrCreateTagBuilder(TagFactory.BIOME.create(new Identifier(MOD_ID, "biome_tag_test")))
+					.add(BiomeKeys.BADLANDS.getValue())
+					.add(BiomeKeys.BASALT_DELTAS.getValue());
```
Juuxel: It could be nice to also have `RegistryKey` overloads so you wouldn't have to do the `getValue` each time.

Juuxel (changes requested):
Seems like a nice addition :+1:

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -210,10 +210,10 @@ protected DynamicRegistryTagProvider(FabricDataGenerator dataGenerator, Registry
 	/**
 	 * An extension to {@link net.minecraft.data.server.AbstractTagProvider.ObjectBuilder} that provides additional functionality.
 	 */
-	public class FabricTagBuilder<O> extends ObjectBuilder<O> {
-		private final AbstractTagProvider.ObjectBuilder<O> parent;
+	public class FabricTagBuilder<T> extends ObjectBuilder<T> {
```
liach: If the self-returning methods are final, better just declare this class final as well (as otherwise subclasses cannot return a more reified self type) and this already has a private constructor.

deirn: Made the class final, some methods still have explicit final because of `@SafeVarargs`.

Juuxel (approved):

modmuss50 (approved):

Technici4n:
Should be `AssertionError`s ideally.

Player3324 (approved):

PR #1894, opened 2021-12-10
Some updates to rendering-fluids module, and added fabric-fluid-api containing some utilities for handling custom fluids

ghost:
This pull requests does 3 things:
1. Adds a new module called "fabric-fluid-api" with two classes `FabricFlowableFluid` that inherits from `FlowableFluid`, implementing some common fluid functions, to avoid reimplementing them and write less code, and a `FabricFluidBlock` class that inherits from `FluidBlock` with a public constructor, plus some utility (static methods) grouped into two classes `FluidInteractions` and `FluidMatching`.
2. Adds the possibility to customize the fog on custom fluids (this requires a dependency on the module "fabric-fluid-api" to get the fog parameters from the fluid, that must inherit from FabricFlowableFluid to customize the fog parameters).
3. Imports the texture registration inside FluidRenderHandlerRegistryImpl: Now there is no need to use `ClientSpriteRegistryCallback.event(PlayerScreenHandler.BLOCK_ATLAS_TEXTURE).register((atlasTexture, registry) -> {...})`, because the used textures will be automatically registered: see this commit for info https://github.com/FabricMC/fabric/commit/27b7162a1dc24ddc3f22a6630527b96cc5c929d8
This avoid to forget to register the textures, and hides this method that in my opinion must be internal.
The implementation avoids duplicate registrations, so if the texture is already registered, will not be registered anymore.
4. Added `public static SimpleFluidRenderHandler coloredLava(int tint)` in `SimpleFluidRenderHandler`.


fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -62,9 +65,7 @@
 	 */
 	public static final Identifier LAVA_FLOWING = new Identifier("block/lava_flow");
 
-	protected final Identifier stillTexture;
-	protected final Identifier flowingTexture;
-	protected final Identifier overlayTexture;
```
modmuss50, line 67: Would this class ever be extended? As these fields are technically exposed by the api to mods that extend this?

fabric-fluid-api-v1/src/main/resources/fabric-fluid-api-v1.mixins.json
```diff
@@ -0,0 +1,10 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.fluid",
+  "compatibilityLevel": "JAVA_16",
+  "mixins": [],
+  "client": [],
+  "injectors": {
+    "defaultRequire": 1
+  }
+}
```
modmuss50: If there are no mixins this can go.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -137,6 +138,59 @@ public static SimpleFluidRenderHandler coloredWater(int tint) {
...
+	public void setOverlayTexture(Identifier id) {
+		textures[2] = id;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Identifier[] getTexturesIds() {
+		return Arrays.stream(textures).filter(Objects::nonNull).toArray(Identifier[]::new);
```
modmuss50: I would do this once in the constructor. Dont really want to be using a stream everytime.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
```diff
@@ -64,6 +66,17 @@ public FluidRenderHandler getOverride(Fluid fluid) {
 	public void register(Fluid fluid, FluidRenderHandler renderer) {
 		handlers.put(fluid, renderer);
 		modHandlers.put(fluid, renderer);
+
+		Identifier[] textures = renderer.getTexturesIds();
+
+		if (textures.length > 0) {
+			ClientSpriteRegistryCallback.event(PlayerScreenHandler.BLOCK_ATLAS_TEXTURE).register((atlasTexture, registry) -> {
```
modmuss50: Registing this was left to the modder before, I think this now has the possibility to start registering textures that werent before. Possibly leading to errors in the log? Please correct me if im wrong. 👍

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,116 @@
...
+		return Optional.of(SoundEvents.ITEM_BUCKET_FILL);
+	}
+
+	/**
+	 * @param entity    The current entity that displays the fog.
+	 * @param tickDelta The time passed from the last tick.
+	 * @param world     The current world.
+	 * @return Fog color (-1 = no fog).
+	 */
+	public abstract int getFogColor(Entity entity, float tickDelta, ClientWorld world);
```
modmuss50: These new methods should have a fabric prefix or suffix to prevent clashes where other mappings have methods with the same name.

modmuss50 (changes requested):
Had a quick look, looks like its on the right track. This will be a lot more maintainable going forward. 👍

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
ghost: Fixed

fabric-fluid-api-v1/src/main/resources/fabric-fluid-api-v1.mixins.json
ghost: Fixed

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
devpelux: Fixed

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
devpelux: Did you mean getFabricFogColor .... ?

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
ghost: Fixed

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
ghost: I've changed the behaviour by restoring the default behaviour and adding specific constructors to SimpleFluidRenderHandler by using a RegistrableTexture object to specify if to register or not the textures.
Please check this ;)

https://github.com/FabricMC/fabric/blob/01c4d2d5a7a3ac890e1647adbd46083553f02104/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java

https://github.com/FabricMC/fabric/blob/01c4d2d5a7a3ac890e1647adbd46083553f02104/fabric-textures-v0/src/main/java/net/fabricmc/fabric/api/client/texture/RegistrableTexture.java

https://github.com/FabricMC/fabric/blob/01c4d2d5a7a3ac890e1647adbd46083553f02104/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java#L71

devpelux:
Sorry for the delay but I had huge commitment!

Another thing... in 22w03a (1.18.2) there is a new method to call to specify the fog shape (spherical or cylindrical), this method doesn't exists in 1.18.1, there is a way to add it only for 1.18.2?

`RenderSystem.setShaderFogShape()` or similar... I don't remember.

It must be added here:
https://github.com/FabricMC/fabric/blob/01c4d2d5a7a3ac890e1647adbd46083553f02104/fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/mixin/client/rendering/fluid/MixinBackgroundRenderer.java#L84

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
ghost: Changed in `getFabric...`

ghost:
Now this is much more compatible, because the default behaviour is the same as in the past, nothing must be changed in existing mods if you don't want.

PR #1904, opened 2021-12-13
Fix missing game test argument types Fixes #1899

### LINKED ISSUE #1899: Missing Argument Types
Whenever a player joins, development servers will emit these errors:

```asciidoc
[16:10:46] [Netty Server IO #8/ERROR] (Minecraft) Could not serialize net.minecraft.command.argument.TestFunctionArgumentType@51f5f7c0 (class net.minecraft.command.argument.TestFunctionArgumentType) - will not be sent to client!
[16:10:46] [Netty Server IO #8/ERROR] (Minecraft) Could not serialize net.minecraft.command.argument.TestClassArgumentType@55935df2 (class net.minecraft.command.argument.TestClassArgumentType) - will not be sent to client!
```

This is caused by the somewhat recent addition of `TestCommand` in https://github.com/FabricMC/fabric/pull/1622

https://github.com/FabricMC/fabric/blob/fe42ded0428aacfc65351f3de58806b08fd7ad3d/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/CommandManagerMixin.java#L41-L43

However, the `ArgumentType`s for the command also need to be added. There is a similar `SharedConstants.isDevelopment` condition in `ArgumentTypes.java` which gates the addition of the `test_argument` and `test_class` command types.

modmuss50:


Plecra (approved):

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/ArgumentTypesMixin.java
```diff
@@ -0,0 +1,48 @@
...
+public abstract class ArgumentTypesMixin {
+	@Shadow
+	public static <T extends ArgumentType<?>> void register(String id, Class<T> argClass, ArgumentSerializer<T> serializer) {
+		throw new AssertionError("Nope.");
+	}
+
+	@Inject(method = "register()V", at = @At("RETURN"))
+	private static void register(CallbackInfo ci) {
+		// Registered by vanilla when isDevelopment is enabled.
+		if (!SharedConstants.isDevelopment) {
```
Juuxel, line 43: You could also redirect the `isDevelopment` field access to always be true here, which duplicates less vanilla code (although it's just two lines)

modmuss50: Going to go with this for now as it follows the other mixins.

voruti, line 43: Why is there a "!", i.e. the test arguments are registered when development is *false*??

modmuss50: This the vanilla constant, and is almost always false, even in a fabric mod dev env. If/when its true vanilla will register these as noted by the comment.

voruti: I see.
I currently have the problem that I want to use BungeeCord, but these two test arguments are unknown to BungeeCord and therefore it generates an error (`IndexOutOfBoundsException: Index 49 out of bounds for length 49` at https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/packet/Commands.java#L595).

voruti: Okay, I think I have found a workaround: If I exclude the gametest, as described here https://issuehint.com/issue/FabricMC/fabric/2315#84183795, the registration of these test arguments seems to be skipped.
I don't know if this has any side effects .

PR #1889, opened 2021-12-06
Datagen stuff

Shnupbups:
Refactor names to match yarn pr https://github.com/FabricMC/yarn/pull/2905

Add TAWs for `BlockStateModelGenerator.blockStateCollector` and `BlockStateModelGenerator.modelCollector`

PR #1905, opened 2021-12-15
Make FabricToolTags Identified tags

Shnupbups:
Allows for using them in datagen without annoying casting

### DIFF
--- a/fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/api/tool/attribute/v1/FabricToolTags.java
+++ b/fabric-tool-attribute-api-v1/src/main/java/net/fabricmc/fabric/api/tool/attribute/v1/FabricToolTags.java
@@ -20,22 +20,23 @@ import net.minecraft.item.Item;
 import net.minecraft.tag.Tag;
 import net.minecraft.util.Identifier;
 
+import net.fabricmc.fabric.api.tag.TagFactory;
 import net.fabricmc.fabric.api.tag.TagRegistry;
 
 /**
  * Tool item tags provided by Fabric.
  */
 public class FabricToolTags {
-	public static final Tag<Item> AXES = register("axes");
-	public static final Tag<Item> HOES = register("hoes");
-	public static final Tag<Item> PICKAXES = register("pickaxes");
-	public static final Tag<Item> SHOVELS = register("shovels");
-	public static final Tag<Item> SWORDS = register("swords");
-	public static final Tag<Item> SHEARS = register("shears");
+	public static final Tag.Identified<Item> AXES = register("axes");
+	public static final Tag.Identified<Item> HOES = register("hoes");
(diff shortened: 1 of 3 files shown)


Juuxel:
This is sadly a breaking change, though I'd be glad to have this.

Shnupbups:
> This is sadly a breaking change, though I'd be glad to have this.

How would this be a breaking change? `Tag.Identified` extends `Tag`, so everything should work just fine? I'm probably missing something obvious though, I am operating on little sleep

Shnupbups:
I mean I can always keep the original `FabricToolTags` class, deprecate it, and make a new one that uses Identified tags? maybe in the mining level api instead of tool attribute api?

BasiqueEvangelist:
> > This is sadly a breaking change, though I'd be glad to have this.
> 
> How would this be a breaking change? `Tag.Identified` extends `Tag`, so everything should work just fine? I'm probably missing something obvious though, I am operating on little sleep

This would break binary compatibility, since the descriptors of the fields changed (`Tag` -> `Tag.Identified`)

Juuxel:
> I mean I can always keep the original FabricToolTags class, deprecate it, and make a new one that uses Identified tags?

I don't think we need a fourth copy of this class in FAPI. And yeah, changing the field descriptor from `Tag` to `Tag$Identified` breaks binary compat.

warjort:
Why not just duplicate the field?

````
// New field with new type
+public static final Tag.Identified<Item> AXES_TAG = register("axes");

// Old field name with old descriptor (just an alias)
public static final Tag<Item> AXES = AXES_TAG;
````

deirn:
What about adding a method in `FabricTagBuilder` that cast it to `Identified`?

Shnupbups:
> What about adding a method in `FabricTagBuilder` that cast it to `Identified`?

At that point you may as well just cast it yourself; `FabricToolTags.identified(FabricToolTags.PICKAXES)` looks just as bad as `(Tag.Identified<Item>) FabricToolTags.PICKAXES`.

For now I think I'll go with what warjort said and just duplicate the field, marking the old ones as deprecated. Only small issue with that is that a not-so-ideal naming scheme has to be used for the new ones in order to keep the old ones.

deirn:
No, not in `FabricToolTags` but in `FabricTagBuilder`, the datagen builder. Any tag from `TagFactory` and `TagRegistry` is already `Identified` anyway. Vanilla static tags too.

Shnupbups:
> No, not in `FabricToolTags` but in `FabricTagBuilder`, the datagen builder. Any tag from `TagFactory` and `TagRegistry` is already `Identified` anyway. Vanilla static tags too.

Ah, right, misread, sorry. That sounds much better! Will do.

Shnupbups:
well in that case i'm gonna close this and add the changes to #1889 (which I'm also probably going to split into two PRs, one for the renames and one for the additions)

PR #1909, opened 2021-12-17
Allow adding vanilla tags to FabricTagBuilder

### LINKED ISSUE #1908: Data generation is run before vanilla tags are populated
Data generation is run before vanilla tags are populated, which causes any tag provider that attempts to create a tag that contains a vanilla tag to crash.

For example, say we want a tag `example:custom_tag` that contains `#minecraft:small_flowers`. This:
```java
getOrCreateTagBuilder(TagFactory.ITEM.create(new Identifier("example", "custom_tag"))).addTag(ItemTags.SMALL_FLOWERS);
```
would crash with `java.lang.IllegalArgumentException: Couldn't define tag example:custom_tag as it is missing following references: #minecraft:small_flowers (from vanilla)`.

This can be mitigated by adding to the vanilla tag yourself, but that's not ideal. For example, this:
```java
getOrCreateTagBuilder(ItemTags.SMALL_FLOWERS).add(Items.POPPY);
getOrCreateTagBuilder(TagFactory.ITEM.create(new Identifier("example", "custom_tag"))).addTag(ItemTags.SMALL_FLOWERS);
```
will work without crashing - but it will leave your mod with an otherwise unnecessary `minecraft:small_flowers` item tag file.

deirn:
Also add `forceAddTag` that always add the tag regardless if it present or not.

Closes #1908

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -28,12 +28,18 @@ import net.minecraft.data.server.AbstractTagProvider;
 import net.minecraft.entity.EntityType;
 import net.minecraft.fluid.Fluid;
 import net.minecraft.item.Item;
+import net.minecraft.tag.BlockTags;
+import net.minecraft.tag.EntityTypeTags;
+import net.minecraft.tag.FluidTags;
+import net.minecraft.tag.GameEventTags;
+import net.minecraft.tag.ItemTags;
 import net.minecraft.tag.Tag;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.registry.Registry;
 import net.minecraft.world.event.GameEvent;
 
 import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
+import net.fabricmc.fabric.impl.datagen.FabricTagEntry;
 
 /**
  * Implement this class (or one of the inner classes) to generate a tag list.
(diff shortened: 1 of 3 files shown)


Technici4n (approved):
LGTM

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -225,11 +231,21 @@ private FabricTagBuilder(ObjectBuilder<T> parent) {
 		/**
 		 * Add another tag to this tag.
 		 *
+		 * <p><b>Note:</b> any vanilla tags can be added to the builder,
+		 * but other tags can only be added if it has a builder registered in the same provider.
```
shartte, line 235: Does datagen run before or after other mods have run through their initializer?
On Forge, other mods do run and register their stuff, making it possible to access the tags and item/block definitions of other mods on the classpath.

modmuss50: Data gen should run after all mods have been initialised, if thats not the case now then something is wrong.

Technici4n: It's not the case for gametests at least, not sure about datagen.

modmuss50: Yes, should be the same for the gametests, might just be an oversight on my part.

deirn: In `AbstractTagProvider#run` it checks if there's a builder for the required tag and it throws an error if not. Not sure how forge datagen works but it looks like it actually loads the resources as if it's in the world?

Shnupbups (approved):

kanpov (approved):

Player3324 (approved):

modmuss50:
Merged in with https://github.com/FabricMC/fabric/pull/1926 as I was unable to push the fixes to the conflicts. Thanks 👍

PR #1911, opened 2021-12-17
Datagen Additions

Shnupbups:
Adds TAWs for `BlockStateModelGenerator.blockStateCollector`, `BlockStateModelGenerator.modelCollector`, `TextureKey.of(String)`, and `TextureKey.of(String, TextureKey)`.
Adds method in `FabricTagBuilder` for adding a non-identified tag, which simply casts the tag to identified as it's highly unlikely that any tags will *not* be identified. Helps with adding old Tag fields from FAPI (like Tool Tags).
Makes `ItemTagProvider#copy` filter out any item entries that don't exist in the item registry.

(supersedes #1889 , this one contains just the additions from that)

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -237,7 +237,7 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 		 * Add another tag to this tag.
 		 *
 		 * @return the {@link FabricTagBuilder} instance
-		 * @throws ClassCastException if tag is not {@link Tag.Identified identified}
+		 * @throws ClassCastException if tag is not {@linkplain Tag.Identified identified}
 		 */
 		public FabricTagBuilder<T> addTag(Tag<T> tag) {
 			parent.addTag((Tag.Identified<T>) tag);
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -237,6 +237,7 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 		 * Add another tag to this tag.
 		 *
 		 * @return the {@link FabricTagBuilder} instance
+		 * @throws ClassCastException if tag is not {@link Tag.Identified identified}
 		 */
 		public FabricTagBuilder<T> addTag(Tag<T> tag) {
 			parent.addTag((Tag.Identified<T>) tag);
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -140,6 +140,8 @@ public abstract class FabricTagProvider<T> extends AbstractTagProvider<T> {
 		 *
 		 * <p>The {@link ItemTagProvider} tag provider must be constructed with an associated {@link BlockTagProvider} tag provider to use this method.
 		 *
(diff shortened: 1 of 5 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -233,6 +233,16 @@ private FabricTagBuilder(ObjectBuilder<T> parent) {
 			return this;
 		}
 
+		/**
+		 * Add another tag to this tag.
+		 *
+		 * @return the {@link FabricTagBuilder} instance
```
Juuxel, line 241: Maybe a `@throws ClassCastException` note about `Tag.Identified` would be good here

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -233,6 +233,17 @@ private FabricTagBuilder(ObjectBuilder<T> parent) {
...
+		 * @throws ClassCastException if tag is not {@link Tag.Identified identified}
```
```suggestion
		 * @throws ClassCastException if tag is not {@linkplain Tag.Identified identified}
```

Juuxel (approved):

deirn (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -146,7 +146,7 @@ public ItemTagProvider(FabricDataGenerator dataGenerator) {
 		public void copy(Tag.Identified<Block> blockTag, Tag.Identified<Item> itemTag) {
 			Tag.Builder blockTagBuilder = Objects.requireNonNull(this.blockTagBuilderProvider, "Pass Block tag provider via constructor to use copy").apply(blockTag);
 			Tag.Builder itemTagBuilder = this.getTagBuilder(itemTag);
-			blockTagBuilder.streamEntries().forEach(itemTagBuilder::add);
+			blockTagBuilder.streamEntries().filter((entry) -> entry.getEntry().canAdd(this.registry::containsId, (id) -> true)).forEach(itemTagBuilder::add);
```
Technici4n, line 151: Sounds a bit unexpected, why would copy() silently ignore some tags?

Shnupbups: Say you have a tag called `c:vines`, containing all the many types of vines in the game. You want both a block tag and an item tag for this.

You have `minecraft:vine`, `minecraft:weeping_vines`, and `minecraft:twisting_vines`, sure. Those all have both block form and item form.

But then there are `minecraft:weeping_vines_plant`, `minecraft:twisting_vines_plant`, `minecraft:cave_vines`, and `minecraft:cave_vines_plant`. Those are all *only* available in block form.

So, when you copy your block tag, containing all 7 vine blocks, to an item tag, you want it to only contain the 3 that actually exist as items. Copying the other 4 ids would result in a crash as those items don't exist.

Many vanilla block tags have item tag equivalents that exclude items that don't exist. Take `minecraft:banners`. The block tag contains all 16 colours of 'standing' banners - and wall banners. The item tag contains just the 16 'standing' banners because the wall banners don't have an item form. (or, rather, their item form is the same as the standing one - the same could be said of the above vines example, where the `_plant` blocks' item forms are the same as their respective 'standard' one (and `cave_vines`' item form is `minecraft:glow_berries`) - but I digress)

Technici4n: Ah ok, makes sense. Could you maybe add to the javadoc that only existing tags are copied?

Shnupbups: Added a note `Any block ids that do not exist in the item registry will be filtered out automatically.` to the javadoc

Technici4n (approved):
LGTM, thank you.

Player3324 (approved):

Mysterious-Dev (approved):

modmuss50 (approved):

modmuss50:
API changes are minimal, I dont think this needs last call 👍

PR #1918, opened 2021-12-20
Support Loader 0.13's multiple mod root paths, optimize ModNioResourcePack

Player3324:
See https://github.com/FabricMC/fabric-loader/pull/585 which is a prerequisite

This also optimizes ModNioResourcePack through early namespace based filtering and removal of streams. The hot path should now only do file system accesses if the namespace is known to be present.

The way namespaces are cached limits hot reloading in dev slightly, it won't react to any namespaces changes without restarting the game. This is however a really obscure scenario and thus not a real problem.


Technici4n:
Interesting optimization (caching by namespace), should have thought of it earlier. Vanilla does it but I'm pretty sure GroupResourcePack doesn't. I am definitely interested in benchmarks of this. :wink:

modmuss50 (approved):
Lightly tested with split mc jars, nothing breaks there. Looks solid 👍

PR #1927, opened 2021-12-23
Use the correct information when throwing Exceptions for failed tests.

jaredlll08:
Currently when you fail a GameTest, you don't really get any useful information.
These tests:
```java
    @GameTest(template = "crafttweaker:empty")
    public void alwaysFail(GameTestHelper helper) {
        
        helper.fail("This will always fail because I told it to, this is useful information to know.");
    }

    @GameTest(template = "crafttweaker:empty")
    private void notAccessible(GameTestHelper helper) {
        
        helper.fail("I'm not accessible, since I'm private.");
    }
```

output:
```xml
        <testcase classname="crafttweaker:empty" name="jsonconvertertest.alwaysfail" time="0.082">
            <failure
                    message="Failed to invoke test method (alwaysFail) in (com.blamejared.crafttweaker.gametest.api.data.JsonConverterTest)"/>
        </testcase>
        <testcase classname="crafttweaker:empty" name="jsonconvertertest.notaccessible" time="0.09">
            <failure
                    message="Failed to invoke test method (notAccessible) in (com.blamejared.crafttweaker.gametest.api.data.JsonConverterTest)"/>
        </testcase>
```

Which doesn't really tell me anything about what is wrong with them.

This PR makes it that those tests produce:
```xml
        <testcase classname="crafttweaker:empty" name="jsonconvertertest.alwaysfail" time="0.111">
            <failure message="This will always fail because I told it to, this is useful information to know."/>
        </testcase>
        <testcase classname="crafttweaker:empty" name="jsonconvertertest.notaccessible" time="0.109">
            <failure
                    message="Failed to invoke test method (notAccessible) in (com.blamejared.crafttweaker.gametest.api.data.JsonConverterTest) because class net.fabricmc.fabric.impl.gametest.FabricGameTestHelper cannot access a member of class com.blamejared.crafttweaker.gametest.api.data.JsonConverterTest with modifiers &quot;private&quot;"/>
        </testcase>
```

Which I'm sure we can all agree is way more useful than before.

Something of note, I am *not* using `FabricGameTest` for my test classes.
Looking at the code, I believe if I was I would get the correct information, but as that interface says, it is optional.

### DIFF
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
@@ -102,8 +102,10 @@ public final class FabricGameTestHelper {
 	public static void invokeTestMethod(TestContext testContext, Method method, Object testObject) {
 		try {
 			method.invoke(testObject, testContext);
-		} catch (IllegalAccessException | InvocationTargetException e) {
-			throw new RuntimeException("Failed to invoke test method (%s) in (%s)".formatted(method.getName(), method.getDeclaringClass().getCanonicalName()), e);
+		} catch (IllegalAccessException e) {
+			throw new RuntimeException("Failed to invoke test method (%s) in (%s) because %s".formatted(method.getName(), method.getDeclaringClass().getCanonicalName(), e.getMessage()), e);
+		} catch(InvocationTargetException e) {
+			throw new RuntimeException(e.getCause().getMessage(), e.getCause());
 		}
 	}
 
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
@@ -104,8 +104,12 @@ public final class FabricGameTestHelper {
 			method.invoke(testObject, testContext);
 		} catch (IllegalAccessException e) {
 			throw new RuntimeException("Failed to invoke test method (%s) in (%s) because %s".formatted(method.getName(), method.getDeclaringClass().getCanonicalName(), e.getMessage()), e);
-		} catch(InvocationTargetException e) {
-			throw new RuntimeException(e.getCause().getMessage(), e.getCause());
+		} catch (InvocationTargetException e) {
+			if (e.getCause() instanceof RuntimeException) {
+				throw (RuntimeException) e.getCause();
+			} else {
+				throw new RuntimeException(e.getCause());
+			}
 		}
 	}
 
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
@@ -105,8 +105,8 @@ public final class FabricGameTestHelper {
 		} catch (IllegalAccessException e) {
 			throw new RuntimeException("Failed to invoke test method (%s) in (%s) because %s".formatted(method.getName(), method.getDeclaringClass().getCanonicalName(), e.getMessage()), e);
 		} catch (InvocationTargetException e) {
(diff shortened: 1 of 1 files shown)


deirn (approved):

Technici4n:
Looks weird, it seems that the real issue is that the test code doesn't print exception causes? Wouldn't having the full stack trace be even more useful?

Juuxel (approved):

shartte:
I also noticed that, yeah. The default reporter isn't all that useful.

I think Mojang relies a lot on in-game reporting (via the lecterns)

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
```diff
@@ -102,8 +102,10 @@ public static void runHeadlessServer(LevelStorage.Session session, ResourcePackM
 	public static void invokeTestMethod(TestContext testContext, Method method, Object testObject) {
 		try {
 			method.invoke(testObject, testContext);
-		} catch (IllegalAccessException | InvocationTargetException e) {
-			throw new RuntimeException("Failed to invoke test method (%s) in (%s)".formatted(method.getName(), method.getDeclaringClass().getCanonicalName()), e);
+		} catch (IllegalAccessException e) {
+			throw new RuntimeException("Failed to invoke test method (%s) in (%s) because %s".formatted(method.getName(), method.getDeclaringClass().getCanonicalName(), e.getMessage()), e);
+		} catch(InvocationTargetException e) {
+			throw new RuntimeException(e.getCause().getMessage(), e.getCause());
```
shartte: I actually think if e.getCause() instanceof RuntimeException re, it should throw re directly

At least that's what Vanilla does

shartte (changes requested):
See my comment about re-throwing e.getCause() in case it's already a runtime exception, like the game test frameworks assertion exceptions are.

jaredlll08:
Unless it is a change made by Fabric, I don't believe the Lecturns show the stack trace.

While it is definitely not something I want to address in this PR, I think it may be worthwhile to look into making a new `TestCompletionListener` that does give the stack trace.

What I think the best solution would be would be to let mods define their own `TestCompletionListener` instead of defaulting to the xml based one, I could see someone running their builds on Team City wanting to use the Team City one for example.

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestHelper.java
```diff
@@ -104,8 +104,12 @@ public static void invokeTestMethod(TestContext testContext, Method method, Obje
 			method.invoke(testObject, testContext);
 		} catch (IllegalAccessException e) {
 			throw new RuntimeException("Failed to invoke test method (%s) in (%s) because %s".formatted(method.getName(), method.getDeclaringClass().getCanonicalName(), e.getMessage()), e);
-		} catch(InvocationTargetException e) {
-			throw new RuntimeException(e.getCause().getMessage(), e.getCause());
+		} catch (InvocationTargetException e) {
+			if (e.getCause() instanceof RuntimeException) {
```
liach: maybe use pattern matching instanceof than call and cast twice

liach (approved):

Technici4n:
> Unless it is a change made by Fabric, I don't believe the Lecturns show the stack trace.
> 
> While it is definitely not something I want to address in this PR, I think it may be worthwhile to look into making a new `TestCompletionListener` that does give the stack trace.
> 
> What I think the best solution would be would be to let mods define their own `TestCompletionListener` instead of defaulting to the xml based one, I could see someone running their builds on Team City wanting to use the Team City one for example.

Yes, that would be neat indeed eventually. For now, this PR is already a net improvement.

Technici4n (approved):

Player3324 (approved):

modmuss50 (approved):

liach:
@modmuss50 should be good to go?

modmuss50:
> @modmuss50 should be good to go?

Yes I know.. hence why I added the ready to merge label. I have been busy the last few days so havent gotten the chance to merge it. Its also quite trivial so there is no major rush?

PR #1928, opened 2021-12-23
Fix attribute modifiers inject limiting other mixins.

jaredlll08:
Since the inject is cancelled, it always runs last, and other mixins aren't able to target the map *after* the inject has been applied.

This is the code output of trying to modify the map after the inject has been applied (in this case, there are 2 `ModifyVariable`s used to show how they chain):
```java
Multimap<Attribute, AttributeModifier> multimap = this.localvar$zli000$modifyMapTwo((Multimap)multimap, slot);
multimap = this.localvar$zlh000$modifyMapOne(multimap, slot);
CallbackInfoReturnable callbackInfo8 = new CallbackInfoReturnable("getAttributeModifiers", true, multimap);
this.handler$zga000$getAttributeModifiers(slot, callbackInfo8, multimap);
return callbackInfo8.isCancelled() ? (Multimap)callbackInfo8.getReturnValue() : multimap;
```
`modifyMapTwo` has a priority of `999` and `modifyMapOne` has a priority of `1001`, so regardless of what priority you use, you can't change the output of `getAttributeModifiers`.

My use-case is being able to nuke all attributes from an item, even if they are "default".


modmuss50:
Please run checkstyle to ensure that the code follows our checkstyle rules. `gradlew checkstyle` most IDE's also have a plugin for checkstyle as well.

modmuss50 (approved):

PR #1945, opened 2022-01-05
Fix Indigo UV rotation

qolq:
When using the `MutableQuadView.BAKE_ROTATE_90` flag, Indigo does a diagonal flip instead of a 90-degree rotation. Since the docs for the flags don't specify whether the rotations should be clockwise or counter-clockwise, I assumed the `BAKE_ROTATE_270` flag is working correctly and the rotations should be clockwise. This fix makes the `BAKE_ROTATE_90` flag apply a 90-degree clockwise rotation. The screenshots below are the results of using the relevant flags.

`BAKE_ROTATE_NONE`
no-rotation

`BAKE_ROTATE_90` (without fix)
90-rotation-bug

`BAKE_ROTATE_90` (with fix)
90-rotation-fix

### DIFF
--- a/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/api/renderer/v1/mesh/MutableQuadView.java
+++ b/fabric-renderer-api-v1/src/main/java/net/fabricmc/fabric/api/renderer/v1/mesh/MutableQuadView.java
@@ -45,7 +45,7 @@ public interface MutableQuadView extends QuadView {
 	int BAKE_ROTATE_NONE = 0;
 
 	/**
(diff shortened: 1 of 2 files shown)


Player3324 (approved):
Math looks good, thanks! Please document the clockwise operation in the constant docs in MutableQuadView. Vanilla uses clockwise rotations in the json declaration, so that makes it a reasonable choice regardless.

modmuss50 (approved):

Technici4n (approved):

qolq:
Updated the `MutableQuadView` docs. Should be good to go I think

Jab125 (approved):

PR #1946, opened 2022-01-05
Fix end return spawn location issue.

KosmX:
## Fix for a serious issue:

When you return from the end, you'll spawn at world-spawn even if you have a valid spawnpoint.

probably it was a typo.


modmuss50 (approved):

modmuss50:
Good spot, will get this out shortly. Thanks 👍

PR #1960, opened 2022-01-19
Migrate to SLF4j and update loom

modmuss50:
Loom 0.11 moves log4j to the runtime only classpath, moving to slf4j is done to follow what mojang are doing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
 	id "eclipse"
 	id "idea"
 	id "maven-publish"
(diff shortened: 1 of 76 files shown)


shartte:
In conjunction with this, I also recommend using this IntelliJ live template to auto-expand "log" into a correct logger declaration:

grafik

shartte (approved):

Player3324 (approved):

Juuxel (approved):

PR #1974, opened 2022-01-28
Update HEADER

Atakku:


haykam821:
The header will not be changed, as per the resolution of #598.

modmuss50:
This will break the build as all of the source headers are wrong.


A possible solution is to use the spotless gradle plugin that allows for dyamic years in the header, the same way as its used in loom. However its not currenly a pressing issue.

PR #1979, opened 2022-01-30
 Fix #1912 and #1972: inventory updates in transactions 

### LINKED ISSUE #1912: Chest Neighbor Update triggers during uncomitted Transfer-API Transactions
When items are being extracted from or inserted into a chest via the Fabric transfer API, the chest will notify its neighbors about the changed comparator signal due to the changed storage. It does this *before* the transaction is comitted. 

This may cause adjacent blocks to react continuously to attempted ("simulated") extracts even though no transaction is ever being comitted. Additionally, blocks that perform any inventory action in the neighbor update will actually crash the game when they try to create a Transaction of their own, since a transaction is already ongoing, but they do not have access to this transaction.

### LINKED ISSUE #1972: SingleStackStorage impl can cause unwanted side effects (ex. furnaces can't ever smelt)
`SingleStackStorage.extract` calls [`setStack`](https://github.com/FabricMC/fabric/blob/e66b59e98c4fc80778b077da360be59d99c03098/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java#L147) and appears to rollback afterwards if the transaction was aborted. However, this does not take into account any side effects of setStack. In the case of furnaces, it resets the cooking timer. A simulated extraction running every tick will completely prevent any furnaces from smelting items. This is exactly what happens [here](https://github.com/Fabricators-of-Create/Create/blob/076236b24846b62200ec25c4b379284b2e7d2015/src/main/java/com/simibubi/create/lib/transfer/item/ItemStorageHandler.java#L75). (Note that this example does not use `simulateExtract` directly, howerver switching to use it instead behaves identically.)

Technici4n:
Fixes the issue by deferring the logic in `setStack` to `onFinalCommit` when necessary.
Includes regression tests.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -203,5 +203,10 @@
 		<module name="AtclauseOrder">
 			<property name="tagOrder" value="@param,@return,@throws,@deprecated"/>
 		</module>
+
+        <!-- Prevent var for all cases other than new instance creation -->
+        <module name="MatchXpath">
+            <property name="query" value="//VARIABLE_DEF[./TYPE/IDENT[@text='var'] and not(./ASSIGN/EXPR/LITERAL_NEW)]"/>
+        </module>
 	</module>
 </module>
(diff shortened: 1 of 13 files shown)


Technici4n:
Converted to draft to prevent accidental merge, since I want to fix #1912 and `SimpleInventory#setStack` too in this PR.

Player3324 (approved):

modmuss50 (approved):

PR #1827, opened 2021-11-14
Allow Built in Resource Packs to have custom names

Jab125:

PR #1983, opened 2022-01-31
Add resource source tracking

Juuxel:
Adds `FabricResource` that contains `getFabricPackSource()` for determining where a resource comes from (vanilla, a mod or a user-provided resource pack). Note: requires changes if #1827 merged since the sources aren't constant anymore

## Why is this useful?

This is a prerequisite for loot table sources in #1241, where we can similarly determine where a loot table comes from. This can be used in Fabric API for de-hardcoding tools like shears in vanilla loot tables - we replace a loot table if its `LootTableSource` is `VANILLA`. This PR's source tracking is used as the base for that.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/FabricResource.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/FabricResource.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


LambdAurora (approved):

liach:
Vanilla in 22w11a has added this mechanism, that each resource is now wrapped into a profile-like object that also carries the pack name of the resource. Does that make this unnecessary for 22w11a and above?

Juuxel:
The resource pack name is not enough if we also want to consider built-in mod resource/data packs. (I'd treat built-in mod resources the same way as vanilla with loot events personally)

Juuxel:
For the time being, I've decided to make this an "internal API" for the loot table module. It's simple enough to promote it in the future if there's wider need for it as an API, not just a component of the loot v2 PR.

PR #1990, opened 2022-02-11
Fix BiomeSelectors.tag

modmuss50:
Thanks @Juuxel for spotting.

PR as I dont plan on doing another release this evening.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
@@ -19,6 +19,7 @@ package net.fabricmc.fabric.api.biome.v1;
 import java.util.List;
 import java.util.Optional;
 
+import net.minecraft.tag.TagKey;
 import net.minecraft.util.registry.RegistryEntry;
 import net.minecraft.util.registry.RegistryEntryList;
 import net.minecraft.util.registry.RegistryKey;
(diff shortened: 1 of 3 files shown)


fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
```diff
@@ -143,4 +144,9 @@ default boolean hasBuiltInStructure(ConfiguredStructureFeature<?, ?> configuredS
 	 * <p>If no dimension options exist for the given dimension key, <code>false</code> is returned.
 	 */
 	boolean canGenerateIn(RegistryKey<DimensionOptions> dimensionKey);
+
+	/**
+	 * Returns true if this biome is in the given {@link TagKey}.
```
Juuxel:
```suggestion
	 * {@return true if this biome is in the given {@link TagKey}}
```

Juuxel (approved):

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
```diff
@@ -143,4 +144,9 @@ default boolean hasBuiltInStructure(ConfiguredStructureFeature<?, ?> configuredS
 	 * <p>If no dimension options exist for the given dimension key, <code>false</code> is returned.
 	 */
 	boolean canGenerateIn(RegistryKey<DimensionOptions> dimensionKey);
+
+	/**
+	 * {@return true if this biome is in the given {@link TagKey}}.
+	 */
+	boolean hasTag(TagKey<Biome> tag);
```
haykam821, line 151: Wouldn't it make more sense to name this `isIn` or `isInTag` to match the tag methods? Biome tags contain biomes, not the other way around.

Juuxel: This doesn't imply that biomes contain biome tags, however, just that they have them. A biome (or a block, item etc) can definitely "have a tag" IMO - "my bronze ingot has the `#c:bronze_ingots` tag". (`isIn` is also fine since it matches yarn)

modmuss50:
Moved this PR into: https://github.com/FabricMC/fabric/pull/2010

PR #2009, opened 2022-02-18
Fix type passed to data pack reload events.

modmuss50:
This is a breaking API change, the API was effectively broken anyway so it makes sense to do this ASAP.


LambdAurora (approved):

Juuxel (approved):
LGTM. Also, isn't this the event that had concerning behaviour even on 1.18? (only triggering for `/reload` and not regular data reloads as a part of the actual server lifecycle) Read down from here for context: https://discord.com/channels/507304429255393322/566276937035546624/934925768490901544

liach:
Is there any mod that uses vanilla loaded data in this event? if yes, keeping the loaded server resource collection may be feasible still

apple502j:
Quick GitHub search shows nobody is actually using the second arg (only using the event to reload configs.)

liach (approved):

PR #2024, opened 2022-02-23
Order of operations

CyborgCabbage:
Fixes a small maths error


Juuxel (approved):

deirn:
This method sounds a bit misleading IMO, as it doesn't really set the range exactly x blocks far. I think we should deprecate this and tell people to use `trackRangeChunks` (or make a new method simply called `trackRange` and deprecate both).

liach (approved):

Player3324 (approved):

modmuss50 (approved):

PR #2029, opened 2022-02-24
Add Access Module

Mysterious-Dev:
This PR add a new module with somes AW to let modder access to register functions

- [x] Villager Profession
- [x] SpawnRestriction

WIP


fabric-access-v1/src/main/resources/fabric-access-v1.accesswidener
```diff
@@ -0,0 +1,11 @@
+accessWidener	v1	named
```
deirn, line 1:
```suggestion
accessWidener	v2	named
```

You need v2 for transitive access widener support.

fabric-access-v1/src/main/resources/fabric-access-v1.accesswidener
```diff
@@ -0,0 +1,11 @@
...
+
+accessible    method    net/minecraft/entity/SpawnRestriction    register    (Lnet/minecraft/entity/EntityType;Lnet/minecraft/entity/SpawnRestriction$Location;Lnet/minecraft/world/Heightmap$Type;Lnet/minecraft/entity/SpawnRestriction$SpawnPredicate;)V
```
deirn, line 3:
```suggestion
transitive-accessible    method    net/minecraft/entity/SpawnRestriction    register    (Lnet/minecraft/entity/EntityType;Lnet/minecraft/entity/SpawnRestriction$Location;Lnet/minecraft/world/Heightmap$Type;Lnet/minecraft/entity/SpawnRestriction$SpawnPredicate;)V
```
Need the prefix so projects that depend on this module can actually access them :P.

deirn (changes requested):

fabric-access-v1/src/main/resources/fabric-access-v1.accesswidener
Mysterious-Dev: It's needed for all actual AW ?

deirn: I guess, yeah? You don't need to prefix it if it is just for internal use.
I suggested one change just for example.

Juuxel:
This was already implemented in an actually functional way (using AW v2 and transitive access wideners) in #2044. These registration methods can become a new PR extending that if needed.

PR #2043, opened 2022-03-05
Add fog customization to rendering-fluids module; Fix #1999.

### LINKED ISSUE #1999: Fluid emptying sound API
This would be sort of similar to the existing fluid rendering APIs - a fluid like lava could provide its own emptying sounds unless they want to use the default (water) sounds.

ghost:
Scope of the pull request:
1. Update the `FluidRenderHandler` by including a `FluidFogHandler` to customize the fog rendered when the player camera is inside the fluid.
2. Inject a new method `getFabricBucketEmptySound` to the `Fluid` class, to get the bucket emptying sound from the fluid itself.

---

Compared to the previous closed (by me) pull request https://github.com/FabricMC/fabric/pull/1894:
`FabricFlowableFluid` with all the unnecessary features has been removed.
`FabricFluidBlock` has been removed because is now useless.
The textures are no more auto-registered (the new usage was too tricky and could deliver unexpected side-effects).
All the other unnecessary features/utilities about fluid-matching and interactions have been removed.


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricFluidBlock.java
```diff
@@ -0,0 +1,37 @@
...
+
+package net.fabricmc.fabric.api.object.builder.v1.block;
+
+import net.minecraft.block.FluidBlock;
+import net.minecraft.fluid.FlowableFluid;
+
+/**
+ * <p>Implements a basic {@link FluidBlock}.</p>
+ *
+ * <p>This class is a wrapper of {@link FluidBlock} with an accessible public constructor.</p>
```
Juuxel: No, these classes just add bloat. This can be a transitive AW instead (along with StairsBlock etc which can all come in one PR imo)

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+
+import net.minecraft.screen.PlayerScreenHandler;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.client.ClientSpriteRegistryCallback;
+
+/**
+ * Implements the registration for fluid textures.
+ */
+public interface FluidTextureRegistry {
```
Juuxel: Shouldn't be an interface since there's no implementation and no instance methods

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Registers the fluid textures.
+	 *
+	 * @param textures Textures to register.
+	 */
+	static void register(Identifier... textures) {
```
Juuxel: IMO people can just use the sprite callback directly without this wrapper method

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
```diff
@@ -0,0 +1,72 @@
...
+	float getFabricFogStart(Entity entity, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * Gets the end point of the fluid fog.
+	 *
+	 * @param entity       Entity submerged by the fluid.
+	 * @param fogType      Type of the fog (SKY or TERRAIN).
+	 * @param viewDistance View distance of the entity.
+	 * @param thickFog     Specifies if the fog is thick.
+	 * @return A float indicating the end point of the fluid fog from the entity.
```
Juuxel: Is this a distance? If so, in what units? That should be clarified here

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
```diff
@@ -0,0 +1,72 @@
...
+	int getFabricFogColor(Entity entity, float tickDelta, ClientWorld world);
+
+	/**
+	 * Gets the start point of the fluid fog.
+	 *
+	 * @param entity       Entity submerged by the fluid.
+	 * @param fogType      Type of the fog (SKY or TERRAIN).
+	 * @param viewDistance View distance of the entity.
+	 * @param thickFog     Specifies if the fog is thick.
+	 * @return A float indicating the start point of the fluid fog from the entity.
```
Juuxel: Same here

ghost: Changed

```java
/**
 * <p>Gets the distance in blocks, from the player camera position, in which the fog is totally opaque.</p>
 * <p>If the end distance is less than the start distance, it will be ignored.</p>
 *
 * @param player       Player submerged by the fluid.
 * @param fogType      Type of the fog (SKY or TERRAIN).
 * @param viewDistance View distance of the player in blocks.
 * @param thickFog     Specifies if the fog is thick.
 * @return A float indicating the end distance of the fluid fog from the player camera, in blocks.
 */
float getFabricFogEnd(Entity player, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
ghost: Changed

```java
/**
 * <p>Gets the distance in blocks, from the player camera position, in which the fog starts rendering.</p>
 * <p>This could be negative, in this case the fog starts partially opaque.</p>
 *
 * @param player       Player submerged by the fluid.
 * @param fogType      Type of the fog (SKY or TERRAIN).
 * @param viewDistance View distance of the player in blocks.
 * @param thickFog     Specifies if the fog is thick.
 * @return A float indicating the start distance of the fluid fog from the player camera, in blocks.
 */
float getFabricFogStart(Entity player, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
ghost: In C# this would have been done with a static class, but java has not static classes...

Maybe this way?

``` java
/**
 * Implements the registration for fluid textures.
 */
public final class FluidTextureRegistry {
	private FluidTextureRegistry() {
	}

	/**
	 * Registers the fluid textures.
	 *
	 * @param textures Textures to register.
	 */
	public static void register(Identifier... textures) {
		ClientSpriteRegistryCallback.event(PlayerScreenHandler.BLOCK_ATLAS_TEXTURE).register((atlasTexture, registry) -> {
			for (Identifier texture : textures) {
				registry.register(texture);
			}
		});
	}
}
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
ghost: The scope is to avoid the copy-paste of the callback method.
This way I think makes more comprehensible that you are actually registering a texture and is aligned with all registration methods for blocks, items...

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricFluidBlock.java
ghost: @modmuss50 what do you think about this?

Juuxel: We've decided on access widening the protected block constructors long ago, also see #2044 which includes an AW for `FluidBlock.<init>`

ghost: Ok removed!

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
```diff
@@ -0,0 +1,74 @@
...
+
+package net.fabricmc.fabric.api.client.render.fluid.v1;
+
+import net.minecraft.client.render.BackgroundRenderer;
+import net.minecraft.client.render.FogShape;
+import net.minecraft.client.world.ClientWorld;
+import net.minecraft.entity.Entity;
+
+/**
+ * Defines a non-transparent fluid with a fog.
```
Juuxel: The javadoc should detail how to use the type (in this case, by implementing it)

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public interface OpaqueFluid {
```
Juuxel: Implementing a client-only interface on common classes is not really Fabric API-like design and it's annoying to spam `@Environment` annotations everywhere to prevent crashes on dedi servers. Probably some sort of registry would be better (maybe extending the existing fluid render handler thing with the fog?)

Juuxel (changes requested):

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
Juuxel: I don't really see what there is to copy-paste? The difference between these two is really small IMO:

```java
// fabric-textures-v0 directly
ClientSpriteRegistryCallback.event(PlayerScreenHandler.BLOCK_ATLAS_TEXTURE)
    .register((atlasTexture, registry) -> registry.register(MY_TEXTURE));

// FluidTextureRegistry
FluidTextureRegistry.register(MY_TEXTURE);
```

It also matches how you'd register textures for any other blocks that go around JSON model loading (that don't just bind that texture directly in a BER). Anyway, others' opinions are welcome as well of course :smile:

(Also, this is not really "aligned with" any registration method for blocks or items - there you define the textures in model files written in json or using this exact callback)

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
Juuxel: Yeah, that's how you'd do it in java

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
ghost: Now is clarified
```java
/**
 * <p>Defines a non-transparent fluid with a fog.</p>
 * <p>This interface must be implemented in each {@link FlowableFluid} class you want to specify fog parameters.</p>
 */
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
devpelux: It's only a simplification to make the registration similar to Registry.register(Registry.BLOCK...) it's obvious that is more 
immediate to learn and clean (i mean the code), that to register a texture you must call a FluidTextureRegistry.register(...), instead to register an event in ClientSpriteRegistryCallback and implement an anonymous interface...

However let's wait other opinions, of course, to maintain or remove this...
Actually it is my obsession to always clean up things 😅

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/OpaqueFluid.java
devpelux: Do you mean a sort of setting object to pass into the FluidRenderHandlerRegistry.INSTANCE.register method?

Like this?

```java
FluidRenderHandlerRegistry.INSTANCE.register(TestFluids.NO_OVERLAY, TestFluids.NO_OVERLAY_FLOWING, new SimpleFluidRenderHandler(
  new Identifier("fabric-rendering-fluids-v1-testmod:block/test_fluid_still"),
  new Identifier("fabric-rendering-fluids-v1-testmod:block/test_fluid_flowing"),
  0xFF5555
),
new FogParameters(.....)
);
```

devpelux: This could be passed inside a setting object inside the FlowableFluid class, but FlowableFluid has no settings, and the constructor is blank, so a class extending FlowableFluid must be created.
I wanted to avoid creating a new "fluid-api" non-client side module for a single class.

ghost: I've fixed this issue, please check.
If possible, add a label "new-module", please.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidTextureRegistry.java
ghost: Ok after thinking about it, considering that no one has ever complained about this, and that is not the job for rendering-fluids, but textures module, I've decided to remove this.

Actually is only a simplification that I've added, trying to cleanup the code... it is my obsession to always try to clean up! 😅

ghost:
This now fixes #1999 by adding "getFabricBucketEmptySound" to get the sound directly from the fluid.

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,87 @@
...
+	 * Gets the color of the fluid fog.
+	 *
+	 * @param player    Player submerged by the fluid.
+	 * @param tickDelta Time passed from the last tick.
+	 * @param world     The current client-side world.
+	 * @return <p>An int value indicating the color of the fluid fog.</p>
+	 * <p>If the color is -1 no fog will be rendered.</p>
+	 * <p>(It could be a hexadecimal value like 0xFFFFFF).</p>
+	 */
+	public abstract int getFabricFogColor(Entity player, float tickDelta, ClientWorld world);
```
Juuxel: Just like the previous approach, this will crash on dedicated servers because it references client-only code in common code.

Since the fog stuff is related to rendering, I'd add them as new methods in `FluidRenderHandler` which seems to be the simplest option. Doesn't require any new classes either.

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/MixinBucketItem.java
```diff
@@ -0,0 +1,58 @@
...
+import net.fabricmc.fabric.api.fluid.v1.FabricFlowableFluid;
+
+@Mixin(BucketItem.class)
+public class MixinBucketItem {
+	@Shadow
+	@Final
+	private Fluid fluid;
+
+	@Inject(method = "playEmptyingSound", at = @At("HEAD"), cancellable = true)
+	private void playEmptyingSound(PlayerEntity player, WorldAccess world, BlockPos pos, CallbackInfo ci) {
```
Juuxel: This mixin can duplicate less of the vanilla code if you just `ModifyVariable` the sound event before the `playSound` call

fabric-fluid-v1/build.gradle
```diff
@@ -0,0 +1,6 @@
+archivesBaseName = "fabric-fluid-v1"
```
Juuxel: This doesn't follow the convention for fabric api modules (`fabric-xyz-api-v1` unless it's some special case like `fabric-xyz-events-v1`)

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,87 @@
...
+	 * @return An enum value indicating the shape of the fluid fog: CYLINDER or SPHERE.
+	 */
+	public abstract FogShape getFabricFogShape(Entity player, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * Gets the sound played when the bucket containing this fluid is emptied.
+	 *
+	 * @return SoundEvent played when the bucket containing this fluid is emptied.
+	 */
+	public Optional<SoundEvent> getFabricBucketEmptySound() {
```
Juuxel: We could also put this directly to `FlowableFluid` with an injected interface like this:

```java
interface FabricFlowableFluid {
  default Optional<SoundEvent> getFabricBucketEmptySound() { return Optional.empty(); }
}
```

fabric-rendering-fluids-v1/src/testmod/java/net/fabricmc/fabric/test/client/rendering/fluid/FabricFluidRenderingTestModClient.java
```diff
@@ -39,9 +40,10 @@ public void onInitializeClient() {
 		FluidRenderHandlerRegistry.INSTANCE.setBlockTransparency(Blocks.SPRUCE_DOOR, true);
 		FluidRenderHandlerRegistry.INSTANCE.setBlockTransparency(Blocks.WARPED_DOOR, true);
 
-		// Red stained glass will have falling fluid textures to the side
+		// Red stained-glass will have falling fluid textures to the side
```
Juuxel: (it's "stained glass" without a hyphen, especially in mc since this is how mojang spells it)

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
deirn: Check [Loom's 0.11 changelog](https://github.com/FabricMC/fabric-loom/releases/tag/0.11) for how to do it.

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/MixinBucketItem.java
ghost: I tried changing the code with your suggestion but, if you want a method like
`public Optional<SoundEvent> getBucketFillSound()`,
changing the variable leads to a problem of what to do if the value is not present.

```java
@ModifyArg(method = "playEmptyingSound", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldAccess;playSound(Lnet/minecraft/entity/player/PlayerEntity;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/sound/SoundEvent;Lnet/minecraft/sound/SoundCategory;FF)V"), index = 2)
private SoundEvent replaceSoundEvent(SoundEvent value) {
	if (fluid instanceof FabricFlowableFluid fabricFlowableFluid) {
		Optional<SoundEvent> emptySound = fabricFlowableFluid.getFabricBucketEmptySound();
		return emptySound.orElse(null);
	}
	return value;
}
```

In this case it causes a NullPointerException.
This because in playEmptyingSound it uses the hardcoded soundEvent with no check for null values.

```java
protected void playEmptyingSound(@Nullable PlayerEntity player, WorldAccess world, BlockPos pos) {
        SoundEvent soundEvent = this.fluid.isIn(FluidTags.LAVA) ? SoundEvents.ITEM_BUCKET_EMPTY_LAVA : SoundEvents.ITEM_BUCKET_EMPTY;
        world.playSound(player, pos, soundEvent, SoundCategory.BLOCKS, 1.0F, 1.0F);
        world.emitGameEvent(player, GameEvent.FLUID_PLACE, pos);
}
```

The only way is leave the inject as is, or change the method to:

```java
@NotNull
public abstract SoundEvent getFabricBucketEmptySound();
```

Forcing to specify a non-null SoundEvent.

Or change

`return emptySound.orElse(null);`

to this

`return emptySound.orElse(SoundEvents.ITEM_BUCKET_EMPTY);`

but in all these way is not possible to specify that the bucket has no sound, and, for the last way, is not correct to play a sound if someone specified null... 🤔

fabric-fluid-v1/build.gradle
ghost: Fixed, just forgotten...

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/MixinBucketItem.java
Juuxel: I'd say default to the water sound like vanilla. (So `return emptySound.orElse(existing);` in a modifyvariable for example) That seems adequate for most modded fluids since they typically behave like coloured water anyway and already use this sound.

For a no-sound bucket, the modder can just override `BucketItem.playEmptyingSound` to do nothing. I feel like that's good enough since I'd imagine most modders want their buckets to have sounds.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
```diff
@@ -39,7 +41,7 @@
 	 * @param fluid The Fluid.
 	 * @return The FluidRenderHandler.
 	 */
-	FluidRenderHandler get(Fluid fluid);
```
devpelux, line 42: This could be null, it causes null pointer exceptions when the fluid is the air! So I've marked as @Nullable

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
```diff
@@ -53,7 +55,7 @@ public FluidRenderHandlerRegistryImpl() {
 	}
 
 	@Override
-	public FluidRenderHandler get(Fluid fluid) {
```
devpelux, line 56: The same as before, this could be null.

fabric-rendering-fluids-v1/src/testmod/java/net/fabricmc/fabric/test/client/rendering/fluid/BiomeBasedFogSettings.java
```diff
@@ -0,0 +1,55 @@
...
+	private final float endRadius;
+
+	public BiomeBasedFogSettings(float startRadius, float endRadius) {
+		this.startRadius = startRadius;
+		this.endRadius = endRadius;
+	}
+
+	@Override
+	public int getFogColor(Camera camera, float tickDelta, ClientWorld world) {
+		return world.getBiome(new BlockPos(camera.getPos())).value().getWaterFogColor();
```
devpelux, line 38: Gets the fog color from the current biome water color.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -70,27 +70,62 @@
```
devpelux, line 70: Here I've added all the constructors override for the fog settings parameter.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidFogSettings.java
```diff
@@ -0,0 +1,92 @@
+/*
```
devpelux, line 1: This is a simple implementation of the fog settings imterface.
IMO almost all modded fluids will use this, with fixed fog color and fixed parameters.

ghost:
~I leaved some comments in the code, just in case...~ nope... github removes them after each commit... 🤔
~Needs a testmod for #1999 this is a known issue.~ done.

fabric-rendering-fluids-v1/src/testmod/java/net/fabricmc/fabric/test/client/rendering/fluid/FabricFluidRenderingTestModClient.java
ghost: fixed

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogSettings.java
```diff
@@ -0,0 +1,73 @@
...
+
+import net.minecraft.client.render.BackgroundRenderer;
+import net.minecraft.client.render.Camera;
+import net.minecraft.client.render.FogShape;
+import net.minecraft.client.world.ClientWorld;
+
+/**
+ * Contains some settings for the fluid fog rendering.
+ */
+public interface FluidFogSettings {
```
deirn: The name `settings` doesn't really fit here IMO, `settings` is more for a value class like vanilla's `AbstractBlock.Settings`. I'd rename it to something like `FluidFogHandler`.

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/MixinBucketItem.java
ghost: i've done by maintaining the default sound into the "orElse" case, this is the more simply way to not change the default behaviour of existing fluids with the interface-injection feature

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogSettings.java
ghost: What name do you think is better here?
I've previously named "FogParameters" honestly i don't know 😅

EDIT: I've not readed FluidFogHandler 😅😅

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
ghost: Done, please check if this solution is good!

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogSettings.java
ghost: Done.

fabric-fluid-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
ghost: Was just a bit tricky but works better than expexted.
I think that the wiki must be updated, because there is no example to use interface injections correctly.

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Implements extra features and configs for {@link FlowableFluid}.
+ */
+public interface FabricFlowableFluid {
+	/**
+	 * <p>Gets the sound played when the bucket containing this fluid is emptied.</p>
+	 * <p>If the return value is <code>Optional.empty()</code>, the following sounds are played:</p>
+	 * <p>- If the fluid extends {@link FlowableFluid} or {@link WaterFluid} is played the <code>ITEM_BUCKET_EMPTY</code> sound.</p>
+	 * <p>- If the fluid extends {@link LavaFluid} is played the <code>ITEM_BUCKET_EMPTY_LAVA</code> sound.</p>
```
Juuxel: Hmm, doesn't it actually just check the lava tag in the code or am I misremembering

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/mixin/fluid/MixinBucketItem.java
```diff
@@ -0,0 +1,42 @@
...
+public class MixinBucketItem {
+	@Shadow
+	@Final
+	private Fluid fluid;
+
+	@ModifyArg(method = "playEmptyingSound", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldAccess;playSound(Lnet/minecraft/entity/player/PlayerEntity;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/sound/SoundEvent;Lnet/minecraft/sound/SoundCategory;FF)V"), index = 2)
+	private SoundEvent modSoundEvent(SoundEvent value) {
+		//Returns the specified value if is present, otherwise does nothing.
+		//Lava and water, not overriding this function, maintain their default behaviour.
+		return ((FlowableFluid) fluid).getFabricBucketEmptySound().orElse(value);
```
Juuxel: Is it actually guaranteed elsewhere in that class to be flowable? If not, I think an instanceof would be better

fabric-fluid-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,34 @@
...
+  "authors": [
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.4.0",
+    "fabric-api-base": "*"
+  },
+  "recommends": {
+    "fabric-rendering-fluids-v1": "*"
+  },
```
Juuxel:
```suggestion
  "suggests": {
    "fabric-rendering-fluids-v1": "*"
  },
```
I don't think there's a need for warnings if rendering-fluids is absent, that'll most likely only happen in cases where it's intended

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -70,27 +70,62 @@
...
+	 * with no tint, with custom fog.
+	 *
+	 * @param stillTexture The texture for still fluid.
+	 * @param flowingTexture The texture for flowing/falling fluid.
+	 * @param overlayTexture The texture behind glass, leaves and other
+	 * {@linkplain FluidRenderHandlerRegistry#setBlockTransparency registered
+	 * transparent blocks}.
+	 * @param fogHandler The fluid fog rendering handler.
+	 */
+	public SimpleFluidRenderHandler(Identifier stillTexture, Identifier flowingTexture, @Nullable Identifier overlayTexture, FluidFogHandler fogHandler) {
```
Juuxel: The fog handler params everywhere should have `@Nullable`

fabric-rendering-fluids-v1/src/main/resources/fabric.mod.json
```diff
@@ -19,7 +19,8 @@
     "fabric-api-base": "*"
   },
   "recommends": {
-    "fabric-textures-v0": "*"
+    "fabric-textures-v0": "*",
+    "fabric-fluid-v1": "*"
```
Juuxel:
```suggestion
    "fabric-fluid-api-v1": "*"
```
also IMO that comment about `suggests` applies here as well (especially since you can make a fully functional custom fluid without that module)

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,41 @@
...
+	 *
+	 * @return SoundEvent played when the bucket containing this fluid is emptied.
+	 */
+	default Optional<SoundEvent> getFabricBucketEmptySound() {
```
deirn: Since `getBucketFillSound` is in the base `Fluid` class, maybe inject this interface to that instead?

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
```diff
@@ -0,0 +1,41 @@
...
+ * Implements extra features and configs for {@link FlowableFluid}.
+ */
+public interface FabricFlowableFluid {
+	/**
+	 * <p>Gets the sound played when the bucket containing this fluid is emptied.</p>
+	 * <p>If the return value is <code>Optional.empty()</code>, the following sounds are played:</p>
+	 * <p>- If the fluid extends {@link FlowableFluid} or {@link WaterFluid} is played the <code>ITEM_BUCKET_EMPTY</code> sound.</p>
+	 * <p>- If the fluid extends {@link LavaFluid} is played the <code>ITEM_BUCKET_EMPTY_LAVA</code> sound.</p>
+	 *
+	 * @return SoundEvent played when the bucket containing this fluid is emptied.
```
deirn:
```suggestion
	 * Gets the sound played when the bucket containing this fluid is emptied.
	 *
	 * <p>If the return value is {@link Optional#empty()}, the following sounds are played:
	 * <ul>
	 *     <li>If the fluid extends {@link FlowableFluid} or {@link WaterFluid}, {@link SoundEvents#ITEM_BUCKET_EMPTY} is played.</li>
	 *     <li>If the fluid extends {@link LavaFluid}, {@link SoundEvents#ITEM_BUCKET_EMPTY_LAVA} is played.</li>
	 * </ul>
```

The convention is to use `<p>` after an empty line and without closing `</p>`.
The `@return` is redundant.

Also, don't forget to import `SoundEvents`.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public interface FluidFogHandler {
+	/**
+	 * Gets the fluid fog color RGB. Alpha is ignored.
+	 *
+	 * @param camera    Camera submerged by the fluid.
+	 * @param tickDelta Time passed from the last tick.
+	 * @param world     Current client world.
+	 * @return <p>An int value indicating the RGB color of the fog.</p>
+	 * <p>If the color is -1, the fog is not rendered.</p>
```
deirn:
```suggestion
	 * @return the color of the fog, or {@code -1} to make it hidden.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	int getFogColor(Camera camera, float tickDelta, ClientWorld world);
+
+	/**
+	 * <p>Gets the distance in blocks, from the camera position, in which the fog starts rendering.</p>
+	 * <p>This could be negative, in this case the fog starts partially opaque.</p>
```
deirn:
```suggestion
	 * Gets the distance in blocks, from the camera position, in which the fog starts rendering.
	 *
	 * <p>This could be negative, in this case the fog starts partially opaque.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
+	 * @param viewDistance Current view distance of the submerged player.
+	 * @param thickFog     Specifies if a thick fog must be rendered.
+	 * @return A float indicating the distance, from the camera position, in which the fog starts rendering.
```
deirn:
```suggestion
```
Redundant.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	float getFogStartRadius(Camera camera, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * <p>Gets the distance in blocks, from the camera position, after which the fog is totally opaque.</p>
+	 * <p>If this is less than the start distance, will be ignored.</p>
```
deirn:
```suggestion
	 * Gets the distance in blocks, from the camera position, after which the fog is totally opaque.
	 *
	 * <p>If this is less than the start distance, will be ignored.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
+	 * @param viewDistance Current view distance of the submerged player.
+	 * @param thickFog     Specifies if a thick fog must be rendered.
+	 * @return A float indicating the distance, from the camera position, after which the fog is totally opaque.
```
deirn:
```suggestion
```
Redundant.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	float getFogEndRadius(Camera camera, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * Gets the shape of the fluid fog.
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
+	 * @param viewDistance Current view distance of the submerged player.
+	 * @param thickFog     Specifies if a thick fog must be rendered.
+	 * @return An enum value indicating the shape of the fluid fog: CYLINDER or SPHERE.
```
deirn:
```suggestion
```
The enum return type is already pretty clear, so I don't think this is necessary.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 * <p>If the color is -1, the fog is not rendered.</p>
+	 */
+	int getFogColor(Camera camera, float tickDelta, ClientWorld world);
+
+	/**
+	 * <p>Gets the distance in blocks, from the camera position, in which the fog starts rendering.</p>
+	 * <p>This could be negative, in this case the fog starts partially opaque.</p>
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
```
deirn:
```suggestion
	 * @param fogType      Type of fog.
```
The enum type is already pretty clear, I don't think it's necessary to put the possible values.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @return A float indicating the distance, from the camera position, in which the fog starts rendering.
+	 */
+	float getFogStartRadius(Camera camera, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * <p>Gets the distance in blocks, from the camera position, after which the fog is totally opaque.</p>
+	 * <p>If this is less than the start distance, will be ignored.</p>
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
```
deirn:
```suggestion
	 * @param fogType      Type of fog.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidFogHandler.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @param thickFog     Specifies if a thick fog must be rendered.
+	 * @return A float indicating the distance, from the camera position, after which the fog is totally opaque.
+	 */
+	float getFogEndRadius(Camera camera, BackgroundRenderer.FogType fogType, float viewDistance, boolean thickFog);
+
+	/**
+	 * Gets the shape of the fluid fog.
+	 *
+	 * @param camera       Camera submerged by the fluid.
+	 * @param fogType      Type of fog (SKY or TERRAIN).
```
deirn:
```suggestion
	 * @param fogType      Type of fog.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandler.java
```diff
@@ -66,6 +66,17 @@ default int getFluidColor(@Nullable BlockRenderView view, @Nullable BlockPos pos
 		return -1;
 	}
 
+	/**
+	 * Get the fluid fog rendering handler.
+	 *
+	 * @param pos The position of the fluid in the world. May be null!
+	 * @param state The current state of the fluid.
+	 * @return The fluid fog rendering handler. May be null!
```
deirn:
```suggestion
```
Redundant.

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidFogHandler.java
```diff
@@ -0,0 +1,92 @@
...
+ * A simple {@link FluidFogHandler} with fixed fluid fog settings.
+ */
+public class SimpleFluidFogHandler implements FluidFogHandler {
+	protected final int color;
+	protected final float startRadius;
+	protected final float endRadius;
+	protected final FogShape shape;
+
+	/**
+	 * Instances a new {@link SimpleFluidFogHandler} with the specified fog settings.
```
deirn:
```suggestion
	 * Creates a new handler with the specified fog settings.
```

fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidFogHandler.java
```diff
@@ -0,0 +1,92 @@
...
+		this.color = color;
+		this.startRadius = startRadius;
+		this.endRadius = endRadius;
+		this.shape = shape;
+	}
+
+	/**
+	 * <p>Instances a new {@link SimpleFluidFogHandler} with the specified fog settings,
+	 * without specifying the shape.</p>
+	 * <p>In this case is used the default shape for all fluids (SPHERE).</p>
```
deirn:
```suggestion
	 * Creates a new handler with the specified fog settings, with {@link FogShape#SPHERE} as its shape.
```

fabric-fluid-api-v1/src/main/java/net/fabricmc/fabric/api/fluid/v1/FabricFlowableFluid.java
ghost: mmm this is a bug... thanks! fixed.

devpelux:
I thought it might be better to completely overwrite `playEmptyingSound` in my opinion, to make the sound completely dependent from the fluid.
Obviously I haven't committed it yet, waiting for some opinions.

```java
/**
 * @author FabricMC
 * @reason Makes the sound totally dependent from the fluid, so modders should never have to mess with this method.
 */
@Overwrite
protected void playEmptyingSound(@Nullable PlayerEntity player, WorldAccess world, BlockPos pos) {
	fluid.getFabricBucketEmptySound().ifPresent(sound -> world.playSound(player, pos, sound, SoundCategory.BLOCKS, 1.0F, 1.0F));
	world.emitGameEvent(player, GameEvent.FLUID_PLACE, pos);
}
```

This is also to get rid of the Optional.empty() issue:

```java
/**
 * Gets the sound played when the bucket containing this fluid is emptied.
 *
 * <p>If the return value is {@link Optional#empty()}, the default sounds are played:
 * <ul>
 *     <li>If the fluid is in {@link FluidTags#LAVA} tag, {@link SoundEvents#ITEM_BUCKET_EMPTY_LAVA} is played.</li>
 *     <li>For other fluids, {@link SoundEvents#ITEM_BUCKET_EMPTY} is played.</li>
 * </ul>
 */
default Optional<SoundEvent> getFabricBucketEmptySound() {
	return Optional.empty();
}
```

deirn:
Some of my suggestion aren't yet resolved, probably because GitHub hides them :P.

ghost:
Well... yes... I've closed this... again! 🤔😅
But imo (and I think in others' opinions too) is better to split this PR in half: one for fog and one for #1999, considering that this issue could be resolved in other ways, and this would lead to duplicated features.

To rollback the #1999 feature, it requires a rollback commit and this will make the commit sequence too much confusing and with no sense (adding a feature, removing it, re-adding it...).


#2101 - Fog
#2102 - Emptying sound

PR #2044, opened 2022-03-05
Create transitive AW module

Juuxel:
- All non-public Block constructors for non-abstract Block classes (autogenerated via ASM)
- Various registries already present in Fabric API (and deprecating the now unnecessary classes/methods)
- Screen handler type creation (SHAPI's "simple" screen handler types, to be specific)


modmuss50 (approved):
Looks great, I suggest we get this sooner rather than later, and then incrementally adding more and deprecating APIs.

Nice job on auto generating the block constructors.

ghost:
Great.
So, for example, with this, I could use `new FluidBlock(...)` without problems?

modmuss50:
> Great. So, for example, with this, I could use `new FluidBlock(...)` without problems?

Yes, thats correct. 👍

Technici4n (approved):

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+				}
+			}
+		}
+}
+
+ClassNode loadClass(Path path) {
+	def node = new ClassNode()
+
+	try (def is = Files.newInputStream(path)) {
+		new ClassReader(is).accept(node, 0)
```
liach: can you add a skip code flag to the reader? would speed up reading significantly i recall.

deirn (approved):

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+task generateAccessWidener {
+	doLast {
+		List<String> lines = file("template.accesswidener").text.lines().collect(Collectors.toCollection { [] })
+		Path inputJar = loom.namedMinecraftProvider.parentMinecraftProvider.mergedJar
+
+		try (def fs = FileSystems.newFileSystem(URI.create("jar:${inputJar.toUri()}"), [create: false])) {
+			generateBlockConstructors(lines, fs)
+		}
+
+		file('src/main/resources/fabric-transitive-access-wideners-v1.accesswidener').text = String.join('\n', lines) + '\n'
```
liach, line 28: should we generate for each build instead? that imo may be more up-to-date

Juuxel: It's also done manually in the data gen module so I think it's fine, but it's easy to change in any case

modmuss50: I agree, its best to be done manually, it would require 2 builds, first one to generate and then a 2nd to use. Without something more specific in loom it will wont be much better than manually. Could possibly have one task to run the regen for all modules, but again not a major issue now.

Juuxel: Hmm, I don't think we actually have to declare this one for loom in the first place since this module doesn't contain any code

fabric-transitive-access-wideners-v1/build.gradle
Juuxel: Added all the skip flags because none of that info is actually needed, just the classes and methods

liach (approved):

Player3324 (approved):

PR #2054, opened 2022-03-07
Add `fabric-registry-sync-v0` to all testmods runtime classpath

deirn:
Some testmod doesn't depend on `fabric-registry-sync-v0`, making it crashes if it registers new objects if you run it individually.
This PR will fix that for all modules by adding it into the runtime classpath. You need to explicitly use `testmodImplementation` if you want to access the API.


modmuss50 (approved):
Easy enough 👍

PR #2070, opened 2022-03-12
Fix optional tag from TagKey datagen

dexman545:
When tags are referenced in another tag's json, it is required to prefix them with `#`.

`FabricTagProvider.FabricTagBuilder#addOptionalTag(net.minecraft.util.Identifier)` uses the relevant vanilla code to do so; however,
`FabricTagProvider.FabricTagBuilder#addOptionalTag(net.minecraft.tag.TagKey<T>)` uses the logic for adding a non-tag id to a tag, causing the `#` to be missing from the generated jsons, and therefore tags that do not function as expected.

Other versions of the API may also be affected.


haykam821 (approved):

modmuss50 (approved):

modmuss50:
I expect this was caused when porting.

Juuxel (approved):

liach (approved):

PR #2078, opened 2022-03-17
Fix the issue of uninstalling a dimension mod or datapack

qouteall:
Minecraft stores custom dimension chunk generators in `level.dat`. To deserialize the chunk generator, it reads the registries (biome registry, chunk generator codec registry, etc.). After uninstalling a dimension mod or a dimension datapack, it fails to deserialize the custom dimensions. And Minecraft cannot load the world. (In 1.18.2, clicking the "Safe mode" button has no effect.)

This is important because uninstalling a dimension mod should be an allowed operation. Forge already contains the fix to this issue. It's easy to fix this issue by removing non-vanilla dimensions from the NBT tag before feeding into DFU and the deserializer. The custom dimensions will be re-added later.

In vanilla, if you have a simple dimension datapack that does define its own thing (for example, it only has a dimension with simple superfalt generation, and does not have any biome or noise settings), then after uninstalling that datapack, the dimension still exists. Vanilla does not give a way to remove custom dimensions. I think this is a wrong behavior. Almost no player want this behavior. After uninstalling a dimension datapack or a dimension mod, its dimension should vanish. This Mixin changes that behavior: it will make the dimensions of uninstalled datapacks vanish normally.

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageBugfixMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageBugfixMixin.java
@@ -34,15 +34,17 @@ import net.minecraft.world.level.storage.LevelStorage;
 /**
  * After removing a dimension mod or a dimension datapack, Minecraft may fail to enter
  * the world, because it fails to deserialize the chunk generator of the custom dimensions in file {@code level.dat}
- * This mixin will remove the custom dimensions from the nbt tag, so the deserializer and DFU cannot see custom dimensions and won't cause errors.
+ * This mixin will remove the custom dimensions from the nbt tag, so the deserializer and DFU cannot see custom
+ * dimensions and won't cause errors.
  * The custom dimensions will be re-added later.
  *
- * <p>This Mixin changes a vanilla behavior that is deemed as a bug (MC-197860). In vanilla, the custom dimension is not removed after uninstalling the dimension datapack.
+ * <p>This Mixin changes a vanilla behavior that is deemed as a bug (MC-197860). In vanilla, the custom dimension
+ * is not removed after uninstalling the dimension datapack.
  * This makes custom dimensions non-removable. Most players don't want this behavior.
  * With this Mixin, custom dimensions will be removed when its datapack is removed.
  */
 @Mixin(LevelStorage.class)
-public class LevelStorageMixin {
+public class LevelStorageBugfixMixin {
 	@SuppressWarnings("unchecked")
 	@Inject(method = "readGeneratorProperties", at = @At("HEAD"))
 	private static <T> void onReadGeneratorProperties(
@@ -51,11 +53,9 @@ public class LevelStorageMixin {
 	) {
 		NbtElement nbtTag = ((Dynamic<NbtElement>) nbt).getValue();
 
-		if (nbtTag instanceof NbtCompound compoundTag) {
-			NbtCompound worldGenSettings = ((NbtCompound) nbtTag).getCompound("WorldGenSettings");
+		NbtCompound worldGenSettings = ((NbtCompound) nbtTag).getCompound("WorldGenSettings");
 
-			removeNonVanillaDimensionsFromWorldGenSettingsTag(worldGenSettings);
-		}
+		removeNonVanillaDimensionsFromWorldGenSettingsTag(worldGenSettings);
 	}
 
 	/**
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
@@ -16,6 +16,11 @@
 
 package net.fabricmc.fabric.mixin.dimension;
 
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 import com.mojang.datafixers.DataFixer;
 import com.mojang.datafixers.util.Pair;
 import com.mojang.serialization.Dynamic;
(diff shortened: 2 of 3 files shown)


fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
```diff
@@ -0,0 +1,83 @@
...
+	@SuppressWarnings("unchecked")
+	@Inject(method = "readGeneratorProperties", at = @At("HEAD"))
+	private static <T> void onReadGeneratorProperties(
+			Dynamic<T> nbt, DataFixer dataFixer, int version,
+			CallbackInfoReturnable<Pair<GeneratorOptions, Lifecycle>> cir
+	) {
+		NbtElement nbtTag = ((Dynamic<NbtElement>) nbt).getValue();
+
+		if (nbtTag instanceof NbtCompound compoundTag) {
+			NbtCompound worldGenSettings = ((NbtCompound) nbtTag).getCompound("WorldGenSettings");
```
liach:
```suggestion
			NbtCompound worldGenSettings = compoundTag.getCompound("WorldGenSettings");
```

liach (approved):
maybe name this mixin class to show it's a bugfix?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java
qouteall: I removed the unnecessary if. And renamed to LevelStorageBugfixMixin

Player3324 (approved):
looks plausible

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageBugfixMixin.java
```diff
@@ -0,0 +1,83 @@
...
+		removeNonVanillaDimensionsFromWorldGenSettingsTag(worldGenSettings);
+	}
+
+	/**
+	 * Removes all non-vanilla dimensions from the tag. The custom dimensions will be re-added later from the datapacks.
+	 */
+	@Unique
+	private static void removeNonVanillaDimensionsFromWorldGenSettingsTag(NbtCompound worldGenSettings) {
+		String[] vanillaDimensionIds =
+				new String[]{"minecraft:overworld", "minecraft:the_nether", "minecraft:the_end"};
```
LoganDark, line 67: Curious, why not check for the `minecraft:` prefix?

qouteall: Checking the prefix is also ok.

qouteall:
I just found the mod Fantasy that do dimension management.
Fantasy use the vanilla dimension registry to store the persistent dimensions [code](https://github.com/NucleoidMC/fantasy/blob/1.18.2/src/main/java/xyz/nucleoid/fantasy/RuntimeWorldManager.java#L32).

This PR will probably break the persistent dimensions of Fantasy 

Update: Fantasy author says it's ok.

qouteall:
It's hard to fix this issue while keeping the vanilla behavior of dimension storage:

As I tested, removing a dimension datapack will cause at lease two different errors.
The first one is in `RegistryEntry.Reference#value()` , when it tries to get the datapack's custom noise settings, it will throw an error
After fixing the first one, the second one is, when it tries to freeze the registry, and found the custom biomes, then it will also throw an error in `SimpleRegistry#freeze` , because during the deserialization of its biome source, it will call `SimpleRegistry#getOrCreateEntry` and then add the custom biome holders into the registry. 
There is a dimension map in level.dat and its deserialized using `UnboundedMapCodec`. `UnboundedMapCodec` used `BaseMapCodec#decode` . `BaseMapCodec#decode` uses `apply2stable` . As I tested, apply2stable will return fail of any of its arguments fail. So the `UnboundedMapCodec` will fail when one dimension's chunk generator fail to deserialize.
I made a `FailSoftMapCodec` https://github.com/qouteall/DimensionDataFix/blob/1.18/src/main/java/qouteall/dimension_data_fix/FailSoftMapCodec.java
That ignores the entries that cannot be deserialized. So one error cannot break the whole process. However, when deserializing a custom dimension's generator, its biome source will be deserialized and it will call `getOrCreateEntry()` for its custom biome. This process has side effects: it will add entries to the biome registry. As these biomes does not exist after uninstalling the datapack, the biome registry will fail to freeze. (That side effect was introduced in 1.18.2)
To summarize this: Deserializing a dimension of an uninstalled datapack will have side effects that cause subsequent failure. Simply making the deserialization fail-soft won't fix the issue.

qouteall:
One way to solve that, is to ask Fantasy to change the way of storing dimensions. Another solution would be to keep the vanilla behavior of storing dimensions and make the deserializer and DFU fail soft. As I descibed, making the deserializer and DFU to fail soft is very hard.

Linguardium:
that seems to be 2 different issues:

the first is that the dimension fails when removed
the second is that the custom biomes fail when removed

are you trying to fix the dimension issue or the custom biome issue? Does the issue persist if a custom biome datapack is removed without utilizing a custom dimension (but while still potentially using the custom world options for the overworld)?

qouteall:
@Linguardium If a datapack only contains biomes and there is no dimension that use these biomes, it will have no issue. But normally a dimension datapack will contain custom biomes.

Linguardium:
> If a datapack only contains biomes and there is no dimension that use these biomes, it will have no issue. But normally a dimension datapack will contain custom biomes.

what if a datapack contains biomes and then the world settings are set to utilize these biomes without a dimension mod? it does sound like this issue is more in line with the structure removal issue rather than a dimension issue

qouteall:
@Linguardium First, a dimension mod is similar to a dimension datapack. This issue applies both to dimension mods and dimension datapacks. 
Second, MC stores dimensions into `level.dat` and it fails to deserialize the dimensions when the biomes/chunk generator types it depends are not present. It's indeed a dimension issue. What do you mean by "structure removal"?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageBugfixMixin.java
liach: What if data packs use `minecraft` namespace for their custom dimensions? Or if data packs remove those vanilla dimensions?

Linguardium:
removing structure adding mods (at least in previous versions) would cause a crash when it tried to process the missing structures in the level.dat.   This seems to be what you are saying is happening with biomes in the biome source/generator, which is unrelated to dimension mods specifically as the biome source/generator can be set to a custom value for vanilla dimensions without a custom dimension

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageBugfixMixin.java
qouteall: As far as I know, datapacks cannot remove vanilla dimensions.

But a datapack can use `minecraft` prefix (although it would be a wrong datapack).

qouteall:
~~@Linguardium It's possible that a mod can add custom structures to overworld and get saved to `level.dat` and then error when the get uninstalled. This pr won't fix that issue.~~

qouteall:
@Linguardium Can you elaborate the missing structure case? Which mod adds structures to overworld and cause custom structures to be saved into `level.dat`? How did it do that?

Linguardium:
I may be misremembering, it may have been failure to handle the chunks correctly. it was back in the 1.16 era but the concept was similar. invalid entries causing issues with proper codec use. removing the structure adding mods left structures that were non-existent which caused a null response from a lookup

qouteall:
Another way of fixing this issue would be to make the deserialization fail-soft and make DFU fail-soft.
When deserializing a registry element through `RegistryElementCodec`, it will call `RegistryLoader.LoaderAccess#load()`, and will call `getOrCreateEntry` which will add a registry entry. But then in `SimpleRegistry#freeze()` it will found that there are some entries that does not reference values and throw an error. I can make it to purge wrong entries when it fail to deserialize a dimension.

qouteall:
A different PR to solve this issue https://github.com/FabricMC/fabric/pull/2088
The registry freeze failure issue can be fixed by purging invalid registry entires (holders).

qouteall:
I heard that Fabric API can be compiled to Spigot and other platforms. Does Spigot utilize the vanilla dimension storage?

Patbox:
> I heard that Fabric API can be compiled to Spigot and other platforms. Does Spigot utilize the vanilla dimension storage?

It can't be...

LoganDark:
> Does Spigot utilize the vanilla dimension storage?

I don't think this should concern us

PseudoDistant (approved):

modmuss50 (approved):
This looks fine, its much less invasive and easier to maintain than #2088

PR #2082, opened 2022-03-18
Update Modrinth plugin version

triphora:
Updates the Modrinth plugin from v1.1.0 to v2.1.1, re: [migration guide on the wiki](https://fabricmc.net/wiki/tutorial:minotaur#updating_from_minotaur_1x_to_2x)

[Minotaur's changelog is here](https://github.com/modrinth/minotaur/compare/v1.1.0...v2.1.1) if you're curious, main changes were moving to an extension and the removal of `addLoader` and `addGameVersion`.

Game version, loader, project version, and token are automatically detected. You may have to bump the Minotaur version too if Loom makes breaking changes to the way that Minotaur [finds the Minecraft dependency](https://github.com/modrinth/minotaur/blob/3f4f065ba8d041497a9e26268acd64858c736ff8/src/main/java/com/modrinth/minotaur/compat/FabricLoomCompatibility.java#L11-L18), but I saw that pinning specific versions was preferred rather than using the latest version like Modrinth recommends.

Tested to work well [on staging](https://staging.modrinth.com/mod/freeeedom/version/0.48.0+local-update-minotaur).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ plugins {
 	id "com.diffplug.spotless" version "6.5.1"
 	id "org.ajoberstar.grgit" version "3.1.0"
 	id "com.matthewprenger.cursegradle" version "1.4.0"
-	id "com.modrinth.minotaur" version "1.1.0"
+	id "com.modrinth.minotaur" version "2.4.3"
 	id "me.modmuss50.remotesign" version "0.3.0" apply false
 }
 
@@ -519,6 +519,7 @@ curseforge {
 
 if (signingEnabled) {
 	project.tasks.curseforge.dependsOn signRemapJar
+	project.tasks.modrinth.dependsOn signRemapJar
 	build.dependsOn signRemapJar
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -448,22 +449,13 @@ task github(dependsOn: (signingEnabled ? signRemapJar : remapJar)) {
 	}
 }
 
-task modrinth(type: com.modrinth.minotaur.TaskModrinthUpload, dependsOn: (signingEnabled ? signRemapJar : remapJar)) {
-	onlyIf {
-		ENV.MODRINTH_TOKEN
-	}
-
-	token = ENV.MODRINTH_TOKEN
```
modmuss50, line 553: Where does the token get passed in now?

modmuss50 (changes requested):

build.gradle
triphora: `$MODRINTH_TOKEN` is the default: <https://github.com/modrinth/minotaur/blob/b27fc1b/src/main/java/com/modrinth/minotaur/ModrinthExtension.java#L27>
If you'd like I can change it to explicitly use it in the buildscript.

Hugman76 (approved):

Mysterious-Dev (approved):

triphora:
Rebased this PR for the 1.19.2 branch.

modmuss50 (approved):

PR #2083, opened 2022-03-18
Update command registration API to reflect 1.19 vanilla changes

Player3324:
There are two ways to go about this, keep a part of v1 and otherwise migrate to v2 or update v1 directly. The command-api-v1 module will receive a major version bump either way.

option 1 with partial legacy bridging - full PR

command api v1 loses its client command part (not bridging friendly enough) and forwards the server command event to v2
command api v2 is a copy of v1 with the new register signature
everything has been adjusted to use v2

option 2 without bridging - only the first commit


Technici4n:
I don't think migrating to v2 is necessary, since the breaking changes are quite small imo, in the sense that they are easy to fix, and shouldn't require fixes in too many places (I expect most mods to only use one command registration callback).

So I'd say first commit only. Thoughts?

LambdAurora:
> I don't think migrating to v2 is necessary, since the breaking changes are quite small imo, in the sense that they are easy to fix, and shouldn't require fixes in too many places (I expect most mods to only use one command registration callback).
> 
> So I'd say first commit only. Thoughts?

Client commands are heavily broken, since now there's a special class for looking up the registry for registry-dependent arguments (item, block, etc), this is going to cause a lot of issues for those.

The old API design did not have a per-world registration callback and only gave a dispatcher.
There can be stuff to support the old dispatcher, but it's kind of messy.

I can understand a v2 for client commands.

Technici4n:
Technically that could just be a v1 major version bump.

LambdAurora:
> Technically that could just be a v1 major version bump.

Then we have to start questioning the relevancy of the `v<version>` system.

Player3324:
the client command api can't be reasonably forwarded from a v1 legacy module to the v2 one due to how brigadier's API works, only the simple event for normal commands can be - which is what the full PR does.

fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/CommandRegistrationCallback.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Callback for when a server registers all commands.
+ *
+ * <p>To register some commands, you would register an event listener and implement the callback.
+ *
+ * <pre><code>
+ * CommandRegistrationCallback.EVENT.register((dispatcher, dedicated) -> {
+ *     // For example, this command is only registered on an integrated server
+ *     if (!dedicated) dispatcher.register(CommandManager.literal("integrated_command").executes(context -> {...}));
+ * })};
```
haykam821, line 37: This code example does not match the new parameters provided.

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/v1/LegacyHandler.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minecraft.server.command.CommandManager.RegistrationEnvironment;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.command.v2.CommandRegistrationCallback;
+
+public final class LegacyHandler implements ModInitializer {
+	@Override
+	@SuppressWarnings("deprecation")
+	public void onInitialize() {
+		CommandRegistrationCallback.EVENT.register((dispatcher, registryAccess, environment) -> net.fabricmc.fabric.api.command.v1.CommandRegistrationCallback.EVENT.invoker().register(dispatcher, environment == RegistrationEnvironment.DEDICATED));
```
haykam821, line 28: Could this lambda method be broken up to multiple lines for readability?

haykam821, line 28: Check `environment.dedicated` so that `RegistrationEnvironment.ALL` is handled properly.

fabric-command-api-v2/src/testmod/java/net/fabricmc/fabric/test/command/CommandTest.java
```diff
@@ -40,11 +41,11 @@
 
 	@Override
 	public void onInitialize() {
-		CommandRegistrationCallback.EVENT.register((dispatcher, dedicated) -> {
+		CommandRegistrationCallback.EVENT.register((dispatcher, registryAccess, environment) -> {
 			// A command that exists on both types of servers
 			dispatcher.register(literal("fabric_common_test_command").executes(this::executeCommonCommand));
 
-			if (dedicated) {
+			if (environment == RegistrationEnvironment.DEDICATED) {
```
haykam821, line 48: Check `environment.dedicated` so that `RegistrationEnvironment.ALL` is handled properly.

fabric-commands-v0/src/main/java/net/fabricmc/fabric/api/registry/CommandRegistry.java
```diff
@@ -38,8 +39,9 @@
 	 * @param consumer  The command provider, consuming {@link CommandDispatcher}.
 	 */
 	public void register(boolean dedicated, Consumer<CommandDispatcher<ServerCommandSource>> consumer) {
-		CommandRegistrationCallback.EVENT.register(((dispatcher, isDedicated) -> {
-			if (dedicated && !isDedicated) {
+		CommandRegistrationCallback.EVENT.register(((dispatcher, registryAccess, environment) -> {
+			if (dedicated && environment == RegistrationEnvironment.INTEGRATED
```
haykam821, line 43: Check `environment.integrated` so that `RegistrationEnvironment.ALL` is handled properly.

fabric-commands-v0/src/main/java/net/fabricmc/fabric/api/registry/CommandRegistry.java
```diff
@@ -38,8 +39,9 @@
...
+					|| !dedicated && environment == RegistrationEnvironment.DEDICATED) {
```
haykam821, line 44: Check `environment.dedicated` so that `RegistrationEnvironment.ALL` is handled properly.

haykam821 (changes requested):

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/v1/LegacyHandler.java
PseudoDistant: I agree with this, readability is very important, be it for making contribution easier, or for making debugging easier. :>

PseudoDistant (changes requested):
Apparently I can't just leave comments, it has to be a full review... bleh

fabric-command-api-v1/src/main/java/net/fabricmc/fabric/impl/command/v1/LegacyHandler.java
Player3324: eclipse's formatter doesn't handle line breaks in the inline form properly, this has to do..

Player3324: The implementation I am using matches the javadoc on v1 better and is the same as the existing implementation, although that can be a little misleading.

PR #2086, opened 2022-03-19
Transfer API improvements 3

### LINKED ISSUE #1998: BlockApiLookup doesn't support BlockEntityTypes with a custom supports override
For example Adorn's [`AdornBlockEntityType`](https://github.com/Juuxel/Adorn/blob/1334e4b6263c25450db6ae6acfb00bec436960bc/common/src/main/kotlin/juuxel/adorn/block/entity/AdornBlockEntityType.kt#L13) overrides `supports` to handle it by the block class instead of block type for mod compat.

`BlockApiLookupImpl` doesn't use the public methods but instead accesses the private `blocks` set - which should be unused since I override its only use - with an accessor and crashes because it's empty.

Since doing a fallback provider automatically for these cases is inefficient, it's probably best to just add some javadoc like

```html
<p>The type is not used directly for detecting the supported blocks and block entities in the world, but it is converted to
its {@linkplain BlockEntityType#blocks} when this method is called.

@throws <whatever the exception currently is> if {@link BlockEntityType#blocks} [of any of the types]/[of the type] is empty
```

Technici4n:
List of changes:
- [x] **(Slightly source-breaking change)** Change the return type of `Storage#iterator` and `Storage#iterable` from `Iterator<StorageView<T>>` to `Iterator<? extends StorageView<T>>` to allow returning a list directly. Most modders shouldn't be affected by this (this only broke one call site in the whole module).
- [x] Precise that using the iterator or a view after the transaction is closed is "undefined behavior". Also specify that calling remove on the iterator is not allowed.
- [x] Add `StorageView#getUnderlyingView` to be able to tell if some views are equal. This is useful to **compute the contents of multiple storage views without duplicates** (see testmod).
- [x] Expose the lifecycle of the transaction manager cleanly with an enum.
- [x] Definalize some methods in `SingleStackStorage` to allow custom implementations of some of them if needed.
- [x] Add a note to `BlockApiLookup` to fix #1998.
- [x] Play the composter empty sound when it is emptied through the transfer API, as a comment in the source code suggests.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
@@ -165,9 +165,9 @@ public interface Storage<T> {
 	 * @return An iterable over the contents of this storage.
 	 * @see #iterator
 	 */
-	@SuppressWarnings("unchecked")
-	default Iterable<StorageView<T>> iterable(TransactionContext transaction) {
-		return () -> (Iterator<StorageView<T>>) iterator(transaction);
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	default Iterable<? extends StorageView<T>> iterable(TransactionContext transaction) {
+		return () -> (Iterator) iterator(transaction);
 	}
 
 	/**
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
@@ -140,6 +140,7 @@ public interface Storage<T> {
 	 * Iterate through the contents of this storage, for the scope of the passed transaction.
 	 * Every visited {@link StorageView} represents a stored resource and an amount.
 	 * The iterator doesn't guarantee that a single resource only occurs once during an iteration.
+	 * Calling {@linkplain Iterator#remove remove} on the iterator is not allowed.
 	 *
 	 * <p>The returned iterator and any view it returns are only valid for the scope of to the passed transaction.
 	 * They should not be used once that transaction is closed.
@@ -152,7 +153,7 @@ public interface Storage<T> {
 	 * the iteration.
 	 *
 	 * @param transaction The transaction to which the scope of the returned iterator is tied.
-	 * @return An iterator over the contents of this storage.
+	 * @return An iterator over the contents of this storage. Calling remove on the iterator is not allowed.
 	 */
 	Iterator<? extends StorageView<T>> iterator(TransactionContext transaction);
 
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
@@ -17,15 +17,14 @@
 package net.fabricmc.fabric.api.transfer.v1.storage;
 
 import java.util.Iterator;
-import java.util.NoSuchElementException;
 
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
-import net.fabricmc.fabric.api.transfer.v1.fluid.base.SingleFluidStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.CombinedStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.ExtractionOnlyStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.InsertionOnlyStorage;
+import net.fabricmc.fabric.api.transfer.v1.storage.base.SingleVariantStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.SingleViewIterator;
 import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
 import net.fabricmc.fabric.api.transfer.v1.transaction.TransactionContext;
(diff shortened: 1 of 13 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/FilteringStorage.java
```diff
@@ -220,5 +220,10 @@ public long getAmount() {
 		public long getCapacity() {
 			return backingView.getCapacity();
 		}
+
+		@Override
+		public StorageView<T> getUnderlyingView() {
+			return backingView;
```
shedaniel: Shouldn't this be calling `backingView.getUnderlyingView()` for preventing doubly filtering storage?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
```diff
@@ -62,4 +62,16 @@
...
+
+	/**
+	 * If this is view is a delegate around another storage view, return the underlying view.
+	 * This can be used to check if two views refer to the same inventory "slot".
+	 * <b>Do not try to extract from the underlying view, or you risk bypassing some checks.</b>
+	 *
+	 * <p>It is expected that two storage views with the same underlying view (a.getUnderlyingView() == b.getUnderlyingView())
+	 * share the same content, and mutating one should mutate the other. However, one of them may allow extraction, and the other may not.
+	 */
+	default StorageView<T> getUnderlyingView() {
```
deirn: Does this have any other use besides for deduplicating views? I think it's better to use a marker type and return that instead (sealed interface maybe? Is that a thing?)

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/FilteringStorage.java
Technici4n: Indeed! Fixed.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
Technici4n: No, it's the only use. I agree that there could be misuse, but this is quite niche, and I think that a marker type would only make the API even more confusing than it already is?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -168,8 +165,9 @@ default long simulateExtract(T resource, long maxAmount, @Nullable TransactionCo
 	 * @return An iterable over the contents of this storage.
 	 * @see #iterator
 	 */
+	@SuppressWarnings("unchecked")
 	default Iterable<StorageView<T>> iterable(TransactionContext transaction) {
```
Player3324: shouldn't this use `? extends StorageView<T>` too?

Player3324 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
Technici4n: I'm not sure, I guess it could? I was thinking that it might be a more breaking change, and I'm not too sure that there is a benefit to it. The point was mainly to allow returning a list of something extending `StorageView<T>` in `iterator`.

Player3324: I think that argument applies a little more directly to iterable() than iterator() when it's backed by e.g. a List? Relying on the default impl is not the most efficient at least.

Technici4n: Good point, indeed! I changed it and it didn't break any call site besides the default impl, so it shouldn't be too breaking.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -165,9 +165,9 @@ default long simulateExtract(T resource, long maxAmount, @Nullable TransactionCo
 	 * @return An iterable over the contents of this storage.
 	 * @see #iterator
 	 */
-	@SuppressWarnings("unchecked")
-	default Iterable<StorageView<T>> iterable(TransactionContext transaction) {
-		return () -> (Iterator<StorageView<T>>) iterator(transaction);
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	default Iterable<? extends StorageView<T>> iterable(TransactionContext transaction) {
+		return () -> (Iterator) iterator(transaction);
```
liach, line 170: Do you need this cast at all?

Technici4n: IntelliJ says that I need it
image

I suspect that it's because the wildcard type cannot be instantiated directly.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
deirn: Hmm, fair enough :+1:

deirn (approved):

modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/StorageView.java
```diff
@@ -62,4 +62,16 @@
 	 * or an estimate of the number of resources that could be stored if this view has a blank resource.
 	 */
 	long getCapacity();
+
+	/**
+	 * If this is view is a delegate around another storage view, return the underlying view.
+	 * This can be used to check if two views refer to the same inventory "slot".
+	 * <b>Do not try to extract from the underlying view, or you risk bypassing some checks.</b>
+	 *
+	 * <p>It is expected that two storage views with the same underlying view (a.getUnderlyingView() == b.getUnderlyingView())
```
Juuxel:
```suggestion
	 * <p>It is expected that two storage views with the same underlying view ({@code a.getUnderlyingView() == b.getUnderlyingView()})
```

Juuxel (approved):

modmuss50:
Manually merged in 1.18.2 👍

PR #2088, opened 2022-03-20
Fix the issue of unloading dimensions v2

qouteall:
Different to https://github.com/FabricMC/fabric/pull/2078  this PR use another way to fix this issue.
It makes dimension deserialization fail-soft and purge the invalid registry entries to prevent failing registry freeze.

As I tested, it can fix the issue of unloading dimension datapacks. But I didn't tested the DFU issue. It may still fail when DFU cannot recognize custom dimensions.
The old PR https://github.com/FabricMC/fabric/pull/2078 does not have any issue from DFU. This PR may have more potential issues than the old PR.

This PR maintains the (probably unwanted) vanilla behavior: if you have a simple dimension datapack that does not define its own things, then its dimension cannot be removed from the world.

The mod https://www.curseforge.com/minecraft/mc-mods/dimension-fix-some-forge-patches-ported is similar to this PR

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


Player3324:
So which solution do you prefer? It looks like 2078 is simpler and I'd say MC-197860 is a indeed a bug that we don't need to retain.

qouteall:
@sfPlayer1 I prefer the first one.

qouteall:
Is there any mod that use datapacks to replace overworld, nether or the end? If so the first PR cannot fix that but the second PR can. It also requires re-adding the missing vanilla dimensions. It's a feature that I am going to add.
The first PR and the second PR can work together.

Player3324:
I think we can worry about that case when it shows being a significant enough issue?

PR #2090, opened 2022-03-20
Add stack-aware getAttributeModifiers and isSuitableFor to FabricItem

Technici4n:
Adds an ItemStack-aware version of `Item#getAttributeModifiers(EquipmentSlot slot)` to `FabricItem`. This functionality used to be available until 1.18.1 as `DynamicAttributeTool#getDynamicModifiers(EquipmentSlot slot, ItemStack stack, @Nullable LivingEntity user)`. I have decided to leave the `@Nullable LivingEntity user` out since:
- due to the way the attribute modifiers are cached and updated, wearer-dependent modifiers are easy to get wrong;
- if absolutely necessary, it is possible to retrieve it by iterating over all current entities, and comparing item stacks with `==`.

Also adds an ItemStack-aware version of `Item#isSuitableFor(BlockState state)`. Together with the previous item, these should be a full replacement for the old dynamic tool API. The testmod shows a dynamic item.

Also implements interface injection for `FabricItem`.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
@@ -16,7 +16,12 @@
 
 package net.fabricmc.fabric.api.item.v1;
 
+import com.google.common.collect.Multimap;
+
 import net.minecraft.client.network.ClientPlayerEntity;
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
@@ -25,9 +30,7 @@ import net.minecraft.util.Hand;
 /**
  * General-purpose Fabric-provided extensions for {@link Item} subclasses.
  *
- * <p>Note: This interface is automatically implemented on all items via Mixin,
- * however it has to be implemented explicitly by modders to be able to override functions.
- * In the future, it is planned that {@code public class Item implements FabricItem} will be visible in a development environment.
+ * <p>Note: This interface is automatically implemented on all items via Mixin and interface injection.
  *
  * <p>Note to maintainers: Functions should only be added to this interface if they are general-purpose enough,
  * to be evaluated on a case-by-case basis. Otherwise they are better suited for more specialized APIs.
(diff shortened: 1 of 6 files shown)


Player3324 (approved):

liach (approved):

AlphaMode:
This would be very useful for tinkers since we have to do this already and we also have to do it for a few other things such as checking if the tool is correct for mining a block so maybe this can he expanded to a few other things?

AlphaMode:
If you want more examples you can look at this class (mojmap) https://github.com/Alpha-s-Stuff/TinkersConstruct/blob/a4b9907f830c568dd86c25085f58f11223f2f0d5/src/main/java/slimeknights/tconstruct/library/tools/item/ModifiableItem.java#L277

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -61,4 +64,18 @@ default boolean allowNbtUpdateAnimation(PlayerEntity player, Hand hand, ItemStac
...
+	 * Return the attribute modifiers to apply when this stack is worn in a living entity equipment slot.
+	 * Stack-aware version of {@link Item#getAttributeModifiers(EquipmentSlot)}.
+	 *
+	 * <p>Note that attribute modifiers are only updated when the stack changes, i.e. when {@code ItemStack.areEqual(old, new)} is false.
+	 *
+	 * @param stack the current stack
+	 * @param slot  the equipment slot this stack is in
+	 * @return the attribute modifiers
+	 */
+	default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers(ItemStack stack, EquipmentSlot slot) {
```
modmuss50, line 79: Should we pre/suffix this with fabric to prevent possible conflicts with other mappings/mods?

Player3324: wouldn't that mean that the api is redundant anyway? there should be no vanilla method with the same signature and purpose

Technici4n:
@AlphaMode do you have a list of "reasonable" methods to provide?

`isSuitableFor` for sure since it is also needed to properly replace the old tool api, but the others don't seem as useful?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
AlphaMode: > @AlphaMode do you have a list of "reasonable" methods to provide?
> 
> `isSuitableFor` for sure since it is also needed to properly replace the old tool api, but the others don't seem as useful?

Yeah the other ones aren't actually a part of the item actual item class.
I would just look around in the item class to add methods that don't pass the ItemStack and make a new variant that does.
Like onItemEntityDestroyed or isDamageable could have a variant that passes the ItemStack

Technici4n: > Like onItemEntityDestroyed or isDamageable could have a variant that passes the ItemStack

Bad examples: `onItemEntityDestroyed` already has the ItemStack since it receives the `ItemEntity`, and `isDamageable` requires too much patching imo. :P

I will only add `getAttributeModifiers` and `isSuitableFor` in this PR, for the others we will see later.

Technici4n:
Updated with a stack-aware version of `isSuitableFor`. This PR should be an appropriate replacement for the old tool API.

modmuss50 (approved):

Juuxel (approved):

PR #2092, opened 2022-03-22
Fix #2091: multiple mods adding to the same tag not working

### LINKED ISSUE #2091: [22w11a] Issue with multiple mods adding blocks to Minecraft:mineable/pickaxe.json
I'm perplexed with this behavior.  I've updated several of my mods from 1.18.2 to 22w11a and I'm noticing that when multiple mods added to minecraft:mineable/pickaxe.json, only one of the mod's entries appear to be in the list.

I tested this on the following mods:
- Resourceful Tools (mine) - https://github.com/kwpugh/resourceful_tools_fabric-1.17/tree/22w11a
- MoreGems (mine) - https://github.com/kwpugh/MoreGems/tree/22w11a
- Gobber (mine) - https://github.com/kwpugh/gobber_fabric-1.17/tree/22w11a
- Copper Equipment (someone else's mod)  - https://github.com/Redy1aye/CopperEquipment/tree/CopperEquipment/src/main/resources/data

If I add just one of these mods to the launcher, it's blocks display the proper tags  on  the f3  screen.

If I add a second of these mod, the first mods blocks no longer have the tags, but the second mod's blocks do.

Add a 3rd mod, only the 3rd's blocks have tags.

I tried different sequences of mods 1st, 2nd, 3rd  and I get the same results.

I have not be able to get any info on whether or not the method adding modded blocks to the mineable/pickaxe had changed with 22w11a, so I am using the same method since 1.18.2.

Here is a sample of the current method;
Screen Shot 2022-03-21 at 2 15 41 PM

I asked my questions in the Fabric Discord but got no relevant responses, hence opening an issue.

Regards,

Kevin

Technici4n:
Fixes #2091.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
@@ -118,7 +118,7 @@ public abstract class GroupResourcePack implements ResourcePack {
 		return this.namespacedPacks.keySet();
 	}
 
-	public void appendResources(NamespaceResourceManagerAccessor manager, Identifier id, List<class_7083> resources) throws IOException {
+	public void appendResources(NamespaceResourceManagerAccessor manager, Identifier id, List<class_7083> resources) {
 		List<ModResourcePack> packs = this.namespacedPacks.get(id.getNamespace());
 
 		if (packs == null) {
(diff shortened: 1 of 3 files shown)


Player3324 (approved):

liach:
Should we remove the group resource pack in favor of one resource pack profile creating multiple packs? That approach is more compatible, as the vanilla code uses stream, which can just become an easy patch to flatmap as opposed to patching all these places and failing silently. We can also make the patched profile to throw on calling to create a resource pack, which is also a fail-fast behavior that defends against potential bugs/incompatibilities.

Player3324:
I'd like to see them gone eventually, it should be enough to patch only the gui instead of having these partial fixups. That's for another time though.

liach:
Indeed, having the gui and commands group multiple bound packs together is the best (somewhat like my idea)

PR #2095, opened 2022-03-24
Add common (client/server) FluidVariant attributes

Technici4n:
This PR introduces fluid variant attributes, that can be useful in a variety of situations for mods using the transfer API.

The new `FluidVariantAttributeHandler` and `FluidVariantAttributes` are designed similarly to the existing `FluidVariantRenderHandler` and `FluidVariantRendering` classes, but they are also available on the server.

I propose the following attributes:
- `getName` and `isLighterThanAir`, replacing the existing rendering `getName` and `fillsFromTop`, making these two also available on the server.
- `getFillSound`, `getEmptySound`, and `getLuminance`: self-explanatory.
- `getTemperature`, `getViscosity`: these are somewhat arbitrary, but they can be useful I think. These and their default values are copied from Forge.

There has been some discussion around allowing an arbitrary number of attributes, for example with `AttributeKey<T>` and `<T> T getAttribute(AttributeKey<T>)`, but imo it is unnecessary.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -32,6 +32,8 @@ import net.fabricmc.api.Environment;
 import net.fabricmc.fabric.api.client.render.fluid.v1.FluidRenderHandler;
 import net.fabricmc.fabric.api.client.render.fluid.v1.FluidRenderHandlerRegistry;
 import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariant;
+import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantProperties;
+import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantPropertyHandler;
 
 /**
  * Defines how {@linkplain FluidVariant fluid variants} of a given Fluid should be displayed to clients.
@@ -44,10 +46,12 @@ import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariant;
 @Environment(EnvType.CLIENT)
 public interface FluidVariantRenderHandler {
 	/**
-	 * Return the name that should be used for the passed fluid variant.
+	 * @deprecated Implement {@link FluidVariantPropertyHandler#getName} instead.
+	 * This function will be removed in a future iteration of the API.
 	 */
+	@Deprecated(forRemoval = true)
 	default Text getName(FluidVariant fluidVariant) {
-		return fluidVariant.getFluid().getDefaultState().getBlockState().getBlock().getName();
+		return FluidVariantProperties.getName(fluidVariant);
 	}
 
 	/**
@@ -120,10 +124,12 @@ public interface FluidVariantRenderHandler {
 	}
 
 	/**
-	 * Return {@code true} if this fluid should fill tanks from top.
+	 * @deprecated Implement {@link FluidVariantPropertyHandler#isGaseous(FluidVariant)} instead.
+	 * This function will be removed in a future iteration of the API.
 	 */
+	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
-		// By default, fluids should be filled from the bottom.
-		return false;
+		// By default, only gaseous fluids should be filled from top.
+		return FluidVariantProperties.isGaseous(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -129,7 +129,7 @@ public interface FluidVariantRenderHandler {
 	 */
 	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
-		// By default, only gaseous fluids should be filled from top.
+		// By default, only fluids lighter than air should be filled from top.
 		return FluidVariantAttributes.isLighterThanAir(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -124,12 +124,12 @@ public interface FluidVariantRenderHandler {
 	}
 
 	/**
-	 * @deprecated Implement {@link FluidVariantAttributeHandler#isGaseous(FluidVariant)} instead.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#isLighterThanAir(FluidVariant)} instead.
 	 * This function will be removed in a future iteration of the API.
 	 */
 	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
 		// By default, only gaseous fluids should be filled from top.
-		return FluidVariantAttributes.isGaseous(fluidVariant);
+		return FluidVariantAttributes.isLighterThanAir(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -124,12 +124,12 @@ public interface FluidVariantRenderHandler {
 	}
 
 	/**
-	 * @deprecated Implement {@link FluidVariantAttributeHandler#isLighterThanAir(FluidVariant)} instead.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#flowsUpwards(FluidVariant)} instead.
 	 * This function will be removed in a future iteration of the API.
 	 */
 	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
 		// By default, only fluids lighter than air should be filled from top.
-		return FluidVariantAttributes.isLighterThanAir(fluidVariant);
+		return FluidVariantAttributes.flowsUpwards(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -124,12 +124,12 @@ public interface FluidVariantRenderHandler {
 	}
 
 	/**
-	 * @deprecated Implement {@link FluidVariantAttributeHandler#flowsUpwards(FluidVariant)} instead.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#isLighterThanAir(FluidVariant)} instead.
 	 * This function will be removed in a future iteration of the API.
 	 */
 	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
 		// By default, only fluids lighter than air should be filled from top.
-		return FluidVariantAttributes.flowsUpwards(fluidVariant);
+		return FluidVariantAttributes.isLighterThanAir(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
@@ -32,8 +32,8 @@ import net.fabricmc.api.Environment;
 import net.fabricmc.fabric.api.client.render.fluid.v1.FluidRenderHandler;
 import net.fabricmc.fabric.api.client.render.fluid.v1.FluidRenderHandlerRegistry;
 import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariant;
-import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantProperties;
-import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantPropertyHandler;
+import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantAttributes;
+import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantAttributeHandler;
 
 /**
  * Defines how {@linkplain FluidVariant fluid variants} of a given Fluid should be displayed to clients.
@@ -46,12 +46,12 @@ import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariantPropertyHandler;
 @Environment(EnvType.CLIENT)
 public interface FluidVariantRenderHandler {
 	/**
-	 * @deprecated Implement {@link FluidVariantPropertyHandler#getName} instead.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#getName} instead.
 	 * This function will be removed in a future iteration of the API.
 	 */
 	@Deprecated(forRemoval = true)
 	default Text getName(FluidVariant fluidVariant) {
-		return FluidVariantProperties.getName(fluidVariant);
+		return FluidVariantAttributes.getName(fluidVariant);
 	}
 
 	/**
@@ -124,12 +124,12 @@ public interface FluidVariantRenderHandler {
 	}
 
 	/**
-	 * @deprecated Implement {@link FluidVariantPropertyHandler#isGaseous(FluidVariant)} instead.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#isGaseous(FluidVariant)} instead.
 	 * This function will be removed in a future iteration of the API.
 	 */
 	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
 		// By default, only gaseous fluids should be filled from top.
-		return FluidVariantProperties.isGaseous(fluidVariant);
+		return FluidVariantAttributes.isGaseous(fluidVariant);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
@@ -64,9 +64,12 @@ public final class FluidConstants {
 	public static final int LAVA_TEMPERATURE = 1300;
 
 	public static final int WATER_VISCOSITY = 1000;
-	// TODO: choose values for these 2
 	public static final int LAVA_VISCOSITY = 6000;
 	public static final int LAVA_VISCOSITY_NETHER = 2000;
+	/**
+	 * For flowable fluids, the viscosity should match VISCOSITY_RATIO * {@code FlowableFluid.getFlowSpeed}.
+	 */
+	public static final int VISCOSITY_RATIO = 200;
 
 	private FluidConstants() {
 	}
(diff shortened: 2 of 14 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantPropertyHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Defines the common properties of {@linkplain FluidVariant fluid variants} of a given Fluid.
+ * Register with {@link FluidVariantProperties#register}.
+ *
+ * <p><b>Experimental feature</b>, we reserve the right to remove or change it without further notice.
+ * The transfer API is a complex addition, and we want to be able to correct possible design mistakes.
+ */
+@ApiStatus.Experimental
+public interface FluidVariantPropertyHandler {
```
Juuxel: "Property handler" sounds something much more generic and more active (think "event handler"). I'd call this `FluidVariantProperties` if it wasn't already in use for the util class.

Technici4n: It's meant to mirror the render handler. Technically it can perform arbitrary logic based on the NBT of the variant, so it is a handler somewhat.

Player3324 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantProperties.java
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	/**
+	 * Return an integer in [0, 15]: the light level emitted by this fluid variant, or 0 if it doesn't naturally emit light.
+	 */
+	public static int getLuminance(FluidVariant variant) {
+		int luminance = getHandlerOrDefault(variant.getFluid()).getLuminance(variant);
+
+		if (luminance < 0 || luminance > 15) {
+			throw new RuntimeException("Invalid luminance %d for fluid variant %s".formatted(luminance, variant));
```
liach: We should probably throw `IllegalStateException` and mention that the handler is broken as well.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantPropertyHandler.java
liach: Maybe also note that `RuntimeException` (or whatever exception you later come up with) will be thrown if the handler returns invalid values

ramidzkh (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantProperties.java
Technici4n: I don't think ISE works: `Signals that a method has been invoked at an illegal or inappropriate time.`

I now mention that the handler is broken.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantPropertyHandler.java
Technici4n: I downgraded the exception to a `LOGGER.warn` since it's not worth crashing the game for this.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
```diff
@@ -120,10 +124,12 @@ default int getColor(FluidVariant fluidVariant, @Nullable BlockRenderView view,
 	}
 
 	/**
-	 * Return {@code true} if this fluid should fill tanks from top.
+	 * @deprecated Implement {@link FluidVariantAttributeHandler#isGaseous(FluidVariant)} instead.
+	 * This function will be removed in a future iteration of the API.
 	 */
+	@Deprecated(forRemoval = true)
 	default boolean fillsFromTop(FluidVariant fluidVariant) {
```
Juuxel: Fwiw, this is slightly distinct from being gaseous - I could imagine some magical *liquids* in mods that float upwards and fill from the top, so this isn't necessarily bad to have with a fallback to gases.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
```diff
@@ -54,6 +57,20 @@ public static long fromBucketFraction(long numerator, long denominator) {
...
+	// ==== FLUID ATTRIBUTES ====
+	// ==========================
+	public static final int WATER_TEMPERATURE = 300;
+	public static final int LAVA_TEMPERATURE = 1300;
+
+	public static final int WATER_VISCOSITY = 1000;
+	public static final int LAVA_VISCOSITY = 6000;
+	public static final int LAVA_VISCOSITY_NETHER = 2000;
+	/**
+	 * For flowable fluids, the viscosity should match VISCOSITY_RATIO * {@code FlowableFluid.getFlowSpeed}.
```
Juuxel:
```suggestion
	 * For flowable fluids, the viscosity should match {@code VISCOSITY_RATIO} * {@link net.minecraft.fluid.FlowableFluid#getFlowSpeed()}.
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
```diff
@@ -54,6 +57,20 @@ public static long fromBucketFraction(long numerator, long denominator) {
 		}
 	}
 
+	// ==========================
+	// ==== FLUID ATTRIBUTES ====
+	// ==========================
+	public static final int WATER_TEMPERATURE = 300;
+	public static final int LAVA_TEMPERATURE = 1300;
```
Juuxel: Maybe document that these are in Kelvin

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	default int getTemperature(FluidVariant variant) {
+		return FluidConstants.WATER_TEMPERATURE;
+	}
+
+	/**
+	 * Return a positive integer, representing the viscosity of this fluid.
+	 * Fluids with lower viscosity generally flow faster than fluids with higher viscosity.
+	 *
+	 * <p>More precisely, viscosity should be 200 * {@code FlowableFluid.getFlowSpeed} for flowable fluids.
```
Juuxel:
```suggestion
	 * <p>More precisely, viscosity should be {@value FluidConstants#VISCOSITY_RATIO} *
	 * {@link net.minecraft.fluid.FlowableFluid#getFlowSpeed()} for flowable fluids.
```
The `@value` tag automatically fills in the constant value

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * Return an integer in [0, 15]: the light level emitted by this fluid, or 0 if it doesn't naturally emit light.
+	 */
+	default int getLuminance(FluidVariant variant) {
+		return variant.getFluid().getDefaultState().getBlockState().getLuminance();
+	}
+
+	/**
+	 * Return a non-negative integer, representing the temperature of this fluid in Kelvin.
+	 * The reference values are 300 for water, and 1300 for lava.
```
Juuxel:
```suggestion
	 * The reference values are {@value FluidConstants#WATER_TEMPERATURE} for water,
	 * and {@value FluidConstants#LAVA_TEMPERATURE} for lava.
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+	default int getTemperature(FluidVariant variant) {
+		return FluidConstants.WATER_TEMPERATURE;
+	}
+
+	/**
+	 * Return a positive integer, representing the viscosity of this fluid.
+	 * Fluids with lower viscosity generally flow faster than fluids with higher viscosity.
+	 *
+	 * <p>More precisely, viscosity should be 200 * {@code FlowableFluid.getFlowSpeed} for flowable fluids.
+	 * The reference values are 1000 for water, 2000 for lava in ultrawarm dimensions (such as the nether), and 6000 for lava in other dimensions.
```
Juuxel:
```suggestion
	 * The reference values are {@value FluidConstants#WATER_VISCOSITY} for water, {@value FluidConstants#LAVA_VISCOSITY_NETHER} for lava in ultrawarm dimensions (such as the Nether),
	 * and {@value FluidConstants#LAVA_VISCOSITY} for lava in other dimensions.
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
```diff
@@ -0,0 +1,159 @@
...
+			TransferApiImpl.LOGGER.warn("Broken FluidVariantAttributeHandler. Invalid luminance %d for fluid variant %s".formatted(luminance, variant));
+			return DEFAULT_HANDLER.getLuminance(variant);
+		}
+
+		return luminance;
+	}
+
+	/**
+	 * Return a non-negative integer, representing the temperature of this fluid in Kelvin.
+	 * The reference values are 300 for water, and 1300 for lava.
```
Juuxel:
```suggestion
	 * The reference values are {@value FluidConstants#WATER_TEMPERATURE} for water,
	 * and {@value FluidConstants#LAVA_TEMPERATURE} for lava.
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		return temperature;
+	}
+
+	/**
+	 * Return a positive integer, representing the viscosity of this fluid variant.
+	 * Fluids with lower viscosity generally flow faster than fluids with higher viscosity.
+	 *
+	 * <p>More precisely, viscosity should roughly be 200 * {@code FlowableFluid.getFlowSpeed}.
```
Juuxel:
```suggestion
	 * <p>More precisely, viscosity should be {@value FluidConstants#VISCOSITY_RATIO} *
	 * {@link net.minecraft.fluid.FlowableFluid#getFlowSpeed()} for flowable fluids.
```
The `@value` tag automatically fills in the constant value

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
```diff
@@ -0,0 +1,159 @@
...
+	 * The reference values are 1000 for water, 2000 for lava in ultrawarm dimensions (such as the nether), and 6000 for lava in other dimensions.
```
Juuxel:
```suggestion
	 * The reference values are {@value FluidConstants#WATER_VISCOSITY} for water, {@value FluidConstants#LAVA_VISCOSITY_NETHER} for lava in ultrawarm dimensions (such as the Nether),
	 * and {@value FluidConstants#LAVA_VISCOSITY} for lava in other dimensions.
```

Juuxel (changes requested):
Mostly javadoc stuff:
- The `{@value}` jd tag is good for constant values in doc comments since it automatically fetches the constants from the field (resistant to code changes)
- Using `{@link}` instead of just `{@code}` lets Mercury (iirc) remap the javadoc for non-yarn mappings

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
Technici4n: I'll rename `isGaseous` to `flowsUpwards` then, since that is the point. (And this info needs to be available on the server too, hence the move to an attribute).

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
Technici4n: Can't reference `getFlowSpeed` directly since it's protected.

Juuxel: You should be able to reference protected methods from links since they are also public API in Java libraries, so the JD tool should work for them just fine. (I'm 99% sure the JDK itself has a link like that)

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+	 * Return the sound corresponding to this fluid being filled.
+	 */
+	default SoundEvent getFillSound(FluidVariant variant) {
+		return variant.getFluid().getBucketFillSound().orElse(SoundEvents.ITEM_BUCKET_FILL);
+	}
+
+	/**
+	 * Return the sound corresponding to this fluid being emptied.
+	 */
+	default SoundEvent getEmptySound(FluidVariant variant) {
```
ghost: Maybe consider to add a modify variable to set this sound inside BucketItem.playEmptyingSound().
This is to avoid to override this method and rewrite vanilla things only for a sound.
The current way could also lead to a situation in which you can set two different sounds for the fluid... 🤔
You set a sound inside playEmptyingSound() and return a different one inside FluidVariant attributes.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+	 * Return the name that should be used for the passed fluid variant.
+	 */
+	default Text getName(FluidVariant fluidVariant) {
+		return fluidVariant.getFluid().getDefaultState().getBlockState().getBlock().getName();
+	}
+
+	/**
+	 * Return the sound corresponding to this fluid being filled.
+	 */
+	default SoundEvent getFillSound(FluidVariant variant) {
```
ghost: Maybe consider to add a modify variable to set this sound inside Fluid.getBucketFillSound().
The current way could also lead to a situation in which you can set two different sounds for the fluid... 🤔
You set a sound inside getBucketFillSound() and return a different one inside FluidVariant attributes.

ghost (changes requested):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
Technici4n: image

Technici4n: Going to ignore this warning since it correctly links it in intellij anyway.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/client/fluid/FluidVariantRenderHandler.java
Technici4n: I went for `isLighterThanAir`.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
Technici4n: The default implementation of `FluidVariantAttributeHandler` should be fine here, if the sound does not depend on the NBT it should be fine to only override Fluid.getBucketFillSound(). I don't think there is a better way. (we can thank vanilla for the asymmetry... :/)

Technici4n: Actually, we could indeed tweak `Fluid#getBucketFillSound` instead. Not sure what is preferable. Maybe it's better to correct the asymmetry as much as possible.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
Technici4n: Done, thanks for the suggestion!

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
Technici4n: Implemented, thanks!

ghost (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/BucketItemMixin.java
```diff
@@ -0,0 +1,59 @@
...
+	private Fluid fluid;
+
+	@ModifyVariable(
+			method = "playEmptyingSound",
+			at = @At("STORE"),
+			index = 4
+	)
+	private SoundEvent hookEmptyingSound(SoundEvent previous) {
+		@Nullable
+		SoundEvent attributeSound = FluidVariantAttributes.getHandlerOrDefault(fluid).getEmptySound(FluidVariant.of(fluid)).orElse(null);
```
liach: Can't we just do something like
```suggestion
		return FluidVariantAttributes.getHandlerOrDefault(fluid).getEmptySound(FluidVariant.of(fluid)).orElse(previous);
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidConstants.java
```diff
@@ -54,6 +59,26 @@ public static long fromBucketFraction(long numerator, long denominator) {
...
+	/**
+	 * Lava temperature, in Kelvin.
+	 */
+	public static final int LAVA_TEMPERATURE = 1300;
+
+	public static final int WATER_VISCOSITY = 1000;
+	public static final int LAVA_VISCOSITY = 6000;
+	public static final int LAVA_VISCOSITY_NETHER = 2000;
+	/**
+	 * For flowable fluids, the viscosity should match {@code #VISCOSITY_RATIO} * {@link FlowableFluid#getFlowSpeed}.
```
Juuxel:
```suggestion
	 * For flowable fluids, the viscosity should match {@code VISCOSITY_RATIO} * {@link FlowableFluid#getFlowSpeed}.
```
Not a link so `#` probably shouldn't be there

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
```diff
@@ -0,0 +1,182 @@
...
+	public static Text getName(FluidVariant variant) {
+		return getHandlerOrDefault(variant.getFluid()).getName(variant);
+	}
+
+	/**
+	 * Return the sound corresponding to a container of this fluid variant being filled if available,
+	 * or the default (water) filling sound otherwise.
+	 */
+	public static SoundEvent getFillSound(FluidVariant variant) {
+		return getHandlerOrDefault(variant.getFluid()).getFillSound(variant).orElse(SoundEvents.ITEM_BUCKET_FILL);
```
Juuxel: Maybe fall back to the fluid's sound here?

Technici4n: Yeah, I guess that should interact correctly with the mixin if it's done here.

Technici4n:
Addressed all comments so far, let me know if you have more. :)

Player3324 (approved):

modmuss50 (approved):

Juuxel (approved):

PR #2098, opened 2022-03-25
Change `PlayerInventoryStorage#drop` to use `PlayerEntity#dropItem`

deirn:
This changes the semantics of `PlayerInventoryStorage#drop` from summoning the item at the player's coord to _throwing_ the item (like when you press `q`) as that's what `PlayerInventory#offerOrDrop` do. Also adds methods that accept `dropItem` arguments.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/PlayerInventoryStorageImpl.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/PlayerInventoryStorageImpl.java
@@ -71,13 +71,13 @@ class PlayerInventoryStorageImpl extends InventoryStorageImpl implements PlayerI
 	}
 
 	@Override
-	public void drop(ItemVariant resource, long amount, TransactionContext tx) {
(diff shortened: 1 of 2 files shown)


Technici4n (approved):

modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/PlayerInventoryStorageImpl.java
```diff
@@ -97,46 +97,45 @@ public void drop(ItemVariant resource, long amount, TransactionContext tx) {
...
 			}
 		}
 
 		@Override
 		protected void onFinalCommit() {
 			// actually drop the stacks
-			for (int i = 0; i < droppedKeys.size(); ++i) {
-				ItemVariant key = droppedKeys.get(i);
+			for (Entry entry : entries) {
+				long reminder = entry.amount;
```
Juuxel: remainder?

Juuxel (approved):

PR #2099, opened 2022-03-26
Fix tags_populated resource conditions in 1.18.2

Technici4n:
Not the cleanest, but it should be acceptable. See also this comment from the PR:
```java
/**
 * Stores the tags deserialized by {@link TagManagerLoader} before they are bound, to use them in the tags_populated conditions.
 * The tags are set at the end of the "apply" phase in {@link TagManagerLoader}, and cleared in {@link net.minecraft.server.DataPackContents#refresh}.
 * If the resource reload fails, the thread local is not cleared and:
 * - the map will remain in memory until the next reload;
 * - any call to {@link #tagsPopulatedMatch} will check the tags from the failed reload instead of failing directly.
 * This is probably acceptable.
 */
```

Also includes a commit by @deirn that moves the gametest entrypoint further up. This prevents loading resources twice, and would have caused the existing tags_populated test to fail, as it should have.

### DIFF
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/server/MainMixin.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/server/MainMixin.java
@@ -34,12 +34,9 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 
 import net.minecraft.resource.ResourcePackManager;
 import net.minecraft.server.Main;
-import net.minecraft.server.SaveLoader;
 import net.minecraft.server.dedicated.EulaReader;
 import net.minecraft.server.dedicated.ServerPropertiesLoader;
 import net.minecraft.util.UserCache;
-import net.minecraft.util.registry.DynamicRegistryManager;
-import net.minecraft.world.SaveProperties;
 import net.minecraft.world.level.storage.LevelStorage;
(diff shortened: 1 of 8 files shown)


deirn (approved):

deirn:
I just noticed that i mixed up gametest with datagen in my commit message 😅 (I blame my sleep schedule)

Player3324 (approved):

modmuss50 (approved):

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/server/MainMixin.java
```diff
@@ -52,8 +49,9 @@ private static boolean isEulaAgreedTo(EulaReader reader) {
...
+	// Inject after resourcePackManager is stored
+	@Inject(method = "main", cancellable = true, locals = LocalCapture.CAPTURE_FAILHARD, at = @At(value = "INVOKE", shift = At.Shift.BY, by = 2, target = "Lnet/minecraft/resource/ResourcePackManager;<init>(Lnet/minecraft/resource/ResourceType;[Lnet/minecraft/resource/ResourcePackProvider;)V"))
```
deirn, line 53: Might want to set `maxShiftBy` to 2 like in #2116 (though I personally don't really care.)

PR #2101, opened 2022-03-27
Add fog customization to rendering-fluids module.

ghost:
### Scope of the PR

Update the `FluidRenderHandler` by including a `FluidFogHandler` to customize the fog rendered when the player camera is inside the fluid.


fabric-rendering-fluids-v1/src/main/java/net/fabricmc/fabric/api/client/render/fluid/v1/SimpleFluidRenderHandler.java
```diff
@@ -99,42 +134,82 @@ public SimpleFluidRenderHandler(Identifier stillTexture, Identifier flowingTextu
...
+
+	/**
+	 * Creates a fluid render handler without an overlay texture,
+	 * with no tint, with custom fog.
+	 *
+	 * @param stillTexture The texture for still fluid.
+	 * @param flowingTexture The texture for flowing/falling fluid.
+	 * @param fogHandler The fluid fog rendering handler.
+	 */
+	public SimpleFluidRenderHandler(Identifier stillTexture, Identifier flowingTexture, @Nullable FluidFogHandler fogHandler) {
```
Technici4n: I think this one clashes with the `(Identifier, Identifier, Identifier)` one when the last parameter is null. I don't think these extra constructors are necessary though? I would say keep the most generic one only, since a custom fog handler is quite niche.

ghost: I've removed the constructors with the fog handler, and added a new fogHandler method to set the handler.
The method returns `this`, so you can use like `FabricBlockSettings.of(Material.METAL).strength(4.0f)`.

`new SimpleFluidRenderHandler(...).fogHandler(handler)`

ghost:
## July 2022 Updates

I tried to add the new 1.19 system to handle fog effects (darkness / blindness), but the result was too much specified and hardcoded, overriding the fog parameters or making impossible to add future effects that will increase the fog radius... the result is that is better to leave this PR unhaltered and more generic as possible, and leave the dev the job to specify what to do with fog.

To generalize it better, i've merged the 3 parameters (the 2 radiuses and the shape) into one object, with one method to retrieve them.
I've also added the `tickDelta` and `skyDarkness` parameters to `FluidFogHandler` methods.
So there are two methods that are basically an extension of `render` and `applyFog` methods in `BackgroundRenderer`.

Now the target is 1.19.1.

I think this is complete, no further changes will be made, except those requested by eventually reviews.

PR #2102, opened 2022-03-27
Fix #1999: Inject getFabricBucketEmptySound() into the Fluid class to get the bucket emptying sound.

### LINKED ISSUE #1999: Fluid emptying sound API

ghost:
This PR propose an implementation of #1999 by injecting the `getFabricBucketEmptySound()` method into the Fluid class to get the bucket emptying sound.  
This is implemented into the transfer api.


Technici4n:
Imo this PR doesn't need to exist (reasoning in #1999 comments). The other one is quite nice though.

ghost:
Closed because of duplication.

PR #2104, opened 2022-04-01
Expose extended screen handler type creation in SH API

Juuxel:
## Changes
- Moved `ExtendedScreenHandlerType` to the API and remove its `final` modifier
  - Creating ESHTs directly matches how "simple" screen handler types should be created now that there's a TAW for it
    **=> prevents users' confusion**
  - Allowing direct creation is also good for multiplatform (Fabric+Forge) projects where you can handle registering and creation very nicely if they're split to be two different things
  - Added documentation
  - It now accepts `ExtendedScreenHandlerType.ExtendedFactory` instead of `ScreenHandlerRegistry.ExtendedClientHandlerFactory` though the interface itself is exactly the same (the latter extends the former now)
- Deprecated `ScreenHandlerRegistry` completely as the creation is now exposed in other ways and registering has always been available in vanilla
- Removed `@Environment(EnvType.CLIENT)` from screen handler factories' `create` methods. This matches vanilla in newer versions where there aren't any specific client-only *members* in otherwise shared classes.
- Expanded the package JD with documentation from `ScreenHandlerRegistry` and new docs as well.

## Motivation
- Simplifies registration in mods like Adorn: [see this change](https://github.com/Juuxel/Adorn/commit/fd36247e2545df7fc7fc88759f5570eee8deb4be)
- Some mods, including the quite popular Architectury API as well as Create Fabric, currently have to either mixin to `Registry` to disable registration or use the current implementation class `ExtendedScreenHandlerType`. This cleans up those use cases.


liach (approved):
Consider a major version bump

Juuxel:
No need for a major version bump since this is backwards compatible. (It'll only break mods like Architectury API that use the impl classes directly)

Player3324 (approved):

Technici4n (approved):

modmuss50 (approved):

PR #2110, opened 2022-04-05
Modernize Indigo

PepperCode1:
- Fix `QuadView#copyTo` not copying enough data
  - The specification (Javadoc) says that this method copies everything except for the material. In reality, it was copying everything except the header bits, which include the cull face, light face, normal set flags, geometry flags, and material, and then marking the target quad's geometry as valid. This did not make sense as the quad data which includes vertex positions was replaced but the header bits which include the geometry were not.
- Refactor `ItemRenderContext`
  - Fix enchantment glint not rendering on item models rendered with this context
  - Make vertex consumer calculation mirror vanilla
  - Improve vertex consumer caching
  - Restructure quad rendering to be more efficient and consistent with block rendering
- Remove unnecessary code including `CompatibilityHelper`
- Add `@Unique` annotations to fields added by Indigo
- Fix typos
- Organize imports

I have tested these changes with Continuity, a mod that heavily utilizes the renderer API, and did not notice any new issues.


AlsoSylv (approved):

maximumpower55 (approved):

frqnny (approved):

Kichura (approved):

isXander (approved):

RaptaG (approved):

fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/QuadViewImpl.java
```diff
@@ -163,7 +163,8 @@ public void copyTo(MutableQuadView target) {
 
 		final MutableQuadViewImpl quad = (MutableQuadViewImpl) target;
 		// copy everything except the material
-		System.arraycopy(data, baseIndex + 1, quad.data, quad.baseIndex + 1, EncodingFormat.TOTAL_STRIDE - 1);
+		System.arraycopy(data, baseIndex, quad.data, quad.baseIndex, EncodingFormat.TOTAL_STRIDE);
+		quad.data[quad.baseIndex + HEADER_BITS] = EncodingFormat.clearMaterial(quad.data[quad.baseIndex + HEADER_BITS]);
```
comp500: Should this preserve the existing material stored in the `MutableQuadViewImpl` rather than clearing it?

comp500 (changes requested):
Looks good, though I wonder if the `@Unique` mixin fields should be renamed to have a consistent naming scheme, since some have a `fabric_` prefix and some don't (I'm not sure what FAPI convention is). Similarly for final fields, should those be capitalised?

fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/QuadViewImpl.java
PepperCode1: I suppose so.

PepperCode1:
> Looks good, though I wonder if the `@Unique` mixin fields should be renamed to have a consistent naming scheme, since some have a `fabric_` prefix and some don't (I'm not sure what FAPI convention is). Similarly for final fields, should those be capitalised?

This can be done, but the convention is that only static final fields are capitalized, which has been done.

comp500 (approved):

Player3324 (approved):

modmuss50 (approved):

Issue #2111, opened 2022-04-06
DynamicRegistryTailCallback


frqnny:
I want to pr a callback that runs after the dynamic registry is set up, and gives you access to the registry manager like DynamicRegistrySetupCallback does.

If anyone has any better names, feel free to comment them.

PR #1031, opened 2020-08-19
Create the fabric-dynamic-registry-api-v1 module

haykam821:
This pull request adds the `fabric-dynamic-registry-api-v1` module. This module is used by implementing `DynamicRegistryProvider` in a `dynamic-registry-provider` entrypoint.

PR #2112, opened 2022-04-06
Add EndDynamicRegistrySetupCallback

frqnny:
Links to #2111 

This small callback adds EndDynamicRegistrySetupCallback
If you are trying to mess with multiple dynamic registries, the safest place is after all of them have loaded. 

I guess you can discuss whether this is necessary?
But to be more precise, this is to facilitate a variety of use cases, such as letting people do something similar to #1540, which sadly was deleted in 1.18. Due to the increasing weirdness of dynamic registries, this implementation doesn't work unless it wants to get really bloated and unnecessary, which would not fit in fabric api.

But I am sure this event will be useful for other use cases.

There is also the case of the name, which I am still unsure of.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEndCallback.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEndCallback.java
@@ -29,7 +29,7 @@ import net.fabricmc.fabric.api.event.EventFactory;
  *
  * <pre>
  *  * {@code
- *  * DynamicRegistryTailCallback.EVENT.register(registryManager -> {
+ *  * DynamicRegistryEndCallback.EVENT.register(registryManager -> {
  *  *     Registry<StructurePool> pools = registryManager.get(Registry.STRUCTURE_POOL_KEY);
  *  *     Registry<StructurePoolProcessorList> lists = registryManager.get(Registry.STRUCTURE_PROCESSOR_LIST_KEY);
  *  *     ...
@@ -38,13 +38,13 @@ import net.fabricmc.fabric.api.event.EventFactory;
  *  * </pre>
  */
 @FunctionalInterface
-public interface DynamicRegistryTailCallback {
+public interface DynamicRegistryEndCallback {
 	void onRegistrationFinish(DynamicRegistryManager registryManager);
 
-	Event<DynamicRegistryTailCallback> EVENT = EventFactory.createArrayBacked(
-			DynamicRegistryTailCallback.class,
+	Event<DynamicRegistryEndCallback> EVENT = EventFactory.createArrayBacked(
+			DynamicRegistryEndCallback.class,
 			callbacks -> registryManager -> {
-				for (DynamicRegistryTailCallback callback : callbacks) {
+				for (DynamicRegistryEndCallback callback : callbacks) {
 					callback.onRegistrationFinish(registryManager);
 				}
 			}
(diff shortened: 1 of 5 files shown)


Trinsdar:
Only other name I could possibly think of is PostDynamicRegistryCallback

Player3324:
we tend to use start/end prefix naming if applicable

frqnny:
Renaming it to DynamicRegistryEndCallback

Player3324:
Sorry, I only added "prefix" after posting the comment initially, see e.g. the tick events in the lifecycle events module. The name probably also needs to include a description of the action, whether it'd be "Load" or something else is tbd.

frqnny:
So how about I rename DynamicRegistrySetupCallback as well to match it? I don't want to break mods that rely on it though

DynamicRegistrySetupCallback  -> StartDynamicRegistrySetupCallback
DynamicRegistryEndCallback -> EndDynamicRegistrySetupCallback

Player3324:
We'd have to keep the old one for compat, I'd just add the "End" one and rename the other one at a later time during a module version bump or so?

PseudoDistant (approved):
I liek

Don't mind me, just passing by. :>

frqnny:
Update on the PR, 
decided to package it in Structurized.


        EndDynamicRegistrySetupCallback.EVENT.register(dynamicRegistries -> {
            Registry<StructurePool> pools = dynamicRegistries.get(Registry.STRUCTURE_POOL_KEY);
            Registry<StructureProcessorList> lists = dynamicRegistries.get(Registry.STRUCTURE_PROCESSOR_LIST_KEY);
            for (StructurePool structurePool : pools) {
                FabricStructurePool pool = new FabricStructurePool(structurePool);
                RegistryEntry<StructureProcessorList> list = lists.getEntry(RegistryKey.of(Registry.STRUCTURE_PROCESSOR_LIST_KEY, new Identifier("empty"))).get();
                if (pool.getUnderlyingPool().getId().toString().contains("village/common/iron_golem")) {
                    pool.addStructurePoolElement(StructurePoolElement.ofProcessedLegacySingle("golemsgalore:variant_1", list).apply(StructurePool.Projection.RIGID), 6);
                    pool.addStructurePoolElement(StructurePoolElement.ofProcessedLegacySingle("golemsgalore:variant_2", list).apply(StructurePool.Projection.RIGID), 4);

                }
            }
        });`



Structurized: https://github.com/frqnny/structurized
Golems Galore: https://github.com/frqnny/golemsgalore

If I pr structurized (in some form) back into FAPI it will very likely use this event and the fact that I am using it in one of my mods proves it works. 

reviews pls

haykam821:
This event would be a good candidate for moving to the `fabric-dynamic-registry-api-v1` module added by #1031.

shartte:
This might break horribly with biome modifications now that 1.19 reuses dynamic registry managers between different worlds.

The current setup is that we _need_ to know the world settings before processing biome modifications to allow mods to freely modify the registry with those in mind (i.e. chunk generator/structure placement settings, etc.).

But as far as I know, this would be called before that happens.

frqnny:
really tired of rewriting this pr every update.

apple502j:
Notes:

- Behavior described by @shartte applies to 22w19a and older, but no longer exists in the release or 1.19-pre1+. However, recreating a world from existing one may reuse DRM in both versions.
- Actually this is not the correct injection point because DRM entries have not been initialized. Better injection point would be `RegistryOps#ofLoaded` (with `ResourceManager`, not `EntryLoader`) at TAIL. Or, if mixin was updated to allow interface injection, toImmutable call.

PR #2129, opened 2022-04-13
Make getKnockbackResistance() work for all materials

Daomephsta:
Vanilla never adds an AttributeModifier for ArmorMaterial.getKnockbackResistance() unless the ArmorMaterial is NETHERITE.
This PR changes that. Any ArmorMaterial that returns > 0.0F from getKnockbackResistance() will cause ArmorItem to add an AttributeModifier for knockback resistance.

I am proposing this for inclusion in fabric-item-api-v1 as armour is a common mod addition, and the required mixin is non-trivial.

There are several other ways the mixin could have been written, but I believe the ModifyVariable is clear, simple, and maintainable. The options are as follows:
* `@Inject` at `TAIL`, copy the entries of `ArmorItem.attributeModifiers` into a new builder, add the knockback resistance modifier if necessary, then build and assign the value to `ArmorItem.attributeModifiers`. This works, but there are easy alternatives that avoid copying.
* `@Inject` at a point where the `builder` local is available, use local capture, and add the knockback resistance modifier if necessary. Avoids copying, but local capture is fragile.
* `@Redirect` the read of `ArmorMaterials.NETHERITE` like so
  ```java
  @Redirect(method = "<init>", at = @At(value = "FIELD", target = "Lnet/minecraft/item/ArmorMaterials;NETHERITE:Lnet/minecraft/item/ArmorMaterial;")) 
  private ArmorMaterial fabric_knockbackResistance(ArmorMaterial material)                                                                           
  {                                                                                                                                                   
      return knockbackResistance > 0.0F ? ArmorMaterials.NETHERITE : material;                                                                        
  }                                                                                                                                                   
  ```
  This is very short, but it isn't easily understandable. If Mojang ever adds other effects to the if statement, it would cause those to be applied to armors with knockback resistance too.
* `@ModifyVariable` the `builder` local to get access to it, then add the knockback resistance modifier if necessary. This avoids copying, it's clear what it does, and it's less sensitive to changes in the LVT than a local capturing `@Inject`.


Technici4n:
Please add a comment to the mixin explaining what it does. You also need to ensure that the various checks pass... you can test locally with `gradlew build` and `gradlew checkstyleMain checkstyleTestmod` for checkstyle specifically.

liach (approved):

Technici4n (commented):
Does this risk breaking existing mods that have a similar inject? I don't see a good workaround though.

Daomephsta:
> Does this risk breaking existing mods that have a similar inject? I don't see a good workaround though.

**Summary:**  
It's highly likely that the breakage would be limited to duplicate tooltip lines. Avoiding that is possible, but it requires the `@Inject` at `TAIL` approach. It's also possible to sanity check instead.

**Full Explanation:**  
The knockback modifier would be duplicated if another mod has a similar injection. However if the modifier UUIDs are the same, the only effect is duplicate tooltip lines, as the code that applies modifiers deduplicates by UUID (first modifier wins).
I checked some of the [existing mods that do this](https://github.com/search?l=Java&q=ArmorItemMixin+knockback&type=Code), they do use the UUID from `ArmorItem.MODIFIERS`, so only the tooltip will be bugged.

All of the existing mods I checked use the `@Inject` at `TAIL` approach (almost always a copy or variant of [this mixin](https://fabricmc.net/wiki/tutorial:armor#adding_knockback_protection)). Because of this the only way I can think of to avoid duplication is to use the same approach with `@Mixin.priority = 1001`, so the code can check for an existing modifier first. 
Alternatively a sanity check can be added to the existing mixin, and the mixin priority set to 1001  
```java
@Shadow private @Final Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers;

@Inject(method = "<init>", at = @At("TAIL"))
private void fabric_knockbackResistanceSanity(CallbackInfo info)
{
    Collection<EntityAttributeModifier> knockbackMods = 
        attributeModifiers.get(EntityAttributes.GENERIC_KNOCKBACK_RESISTANCE);
    if (knockbackMods.size() > 1 && knockbackMods.stream().map(EntityAttributeModifier::getId).distinct().count() < knockbackMods.size())
    {
        // Warning
    }
}
```
Though since the item hasn't been registered yet, the warning would have to identify the instance by the stacktrace element that instantiated it.

Technici4n:
Hmmm, I think it's fine for now with the current mixin then.

Technici4n (approved):

Player3324 (approved):
I think this change is fine. A mod providing custom armor could quite easily override `getAttributeModifiers` to handle this without bytecode edits, but it is fair to assume that most mods don't even realize the need to fix this.

Juuxel (approved):

modmuss50 (approved):

PR #2142, opened 2022-04-18
Fix convention bucket tag ID format

Juuxel:
This makes them consistent with the name format used in Minecraft and other `c` tags like `c:copper_ingots`. (This breaks two conventions: plural instead of singular and suffix instead of directory).

Edit: I created plural-named fields (`WATER_BUCKETS` etc) and deprecated the old ones, as well as moved them down in the source file.


Technici4n (approved):

modmuss50 (approved):

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/ConventionalItemTags.java
```diff
@@ -61,10 +61,10 @@ public class ConventionalItemTags {
...
 	public static final TagKey<Item> POTIONS = register("potions");
 	// Buckets
-	public static final TagKey<Item> WATER_BUCKET = register("bucket/water");
-	public static final TagKey<Item> LAVA_BUCKET = register("bucket/lava");
-	public static final TagKey<Item> MILK_BUCKET = register("bucket/milk");
-	public static final TagKey<Item> EMPTY_BUCKET = register("bucket/empty");
+	public static final TagKey<Item> WATER_BUCKET = register("water_buckets");
+	public static final TagKey<Item> LAVA_BUCKET = register("lava_buckets");
+	public static final TagKey<Item> MILK_BUCKET = register("milk_buckets");
+	public static final TagKey<Item> EMPTY_BUCKET = register("empty_buckets");
```
haykam821: Should these fields be deprecated in favor of fields with plural names?

Technici4n: Could be, but then we'd have the old fields staying around? I guess they could be moved to the bottom of the file.

Juuxel: Both of these suggestions are now in.

Technici4n (approved):
Perfect!

Issue #1635, opened 2021-08-14
[1.17] Creating a custom RenderLayer is high-friction


williewillus:
`RenderLayer.of` was made package-private in 1.17 (and its overload was made private).

Furthermore, because the return type is unnameable, an AW is required if you wanted to use an accessor to invoke it.

This is a lot of hassle for every single mod that does any sort of complex rendering, so I think Fabric API probably should have a call for this.

modmuss50:
This sounds like another great opportunity for global access wideners.

PR #2154, opened 2022-04-21
Move a few object builders to transitive AWs

Technici4n:
List of changes:
- Add TAW for `BlockEntityType.BlockEntityFactory`, which allows using vanilla's builder. **`FabricBlockEntityTypeBuilder` is kept without deprecation (for now).**
- Deprecate `VillagerProfessionBuilder` and replace it by a TAW for the constructor of `VillagerProfession`.
- Deprecate `VillagerTypeHelper` and replace it by TAWs for the constructor of `VillagerType` and the static map `VillagerType.BIOME_TO_TYPE`.
- Add a TAW for `SimpleInventory.stacks` because it is so useful when working with the transfer API.

Note that the testmods from other modules are still using the deprecated `FabricBlockEntityTypeBuilder` because TAWs from other modules are not visible to them.


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
```diff
@@ -32,7 +32,10 @@
  * Fabric's version of BlockEntityType.Builder.
  * Removes the need to use an access widener on the BlockEntityFactory
  *
+ * @deprecated Replaced by access widener for {@link BlockEntityType.BlockEntityFactory}
+ * in Fabric Transitive Access Wideners (v1).
  */
+@Deprecated
```
haykam821: This class provides its own functionality for adding individual blocks that the `BlockEntityFactory` class does not provide.

frqnny: What's the use case, though? Can't you just create the BlockEntityType once you finished building your collection of Blocks?  
I feel like that would be better for modders to do rather than use #addBlock(s) methods.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
```diff
@@ -41,7 +39,11 @@
  * <p>A corresponding <code>IDENTIFIER_PATH.mcmeta</code> file exits in the same directory to define properties such as the {@link VillagerResourceMetadata.HatType HatType} this profession would use.
  *
  * <p>Note this does not register any trades to these villagers. To register trades, add a new entry with your profession as the key to {@link TradeOffers#PROFESSION_TO_LEVELED_TRADE}.
+ *
+ * @deprecated Replaced by access widener for {@link VillagerProfession#VillagerProfession}
+ * in Fabric Transitive Access Wideners (v1).
  */
+@Deprecated
```
frqnny, line 49: IMO we should mark when they will be removed. Is it going to be 1.19 or 1.20?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
Technici4n: > This class provides its own functionality for adding individual blocks that the BlockEntityFactory class does not provide.

Imo that functionality is not worth having in fabric API. If you need it, you can easily emulate it in your mod.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
Technici4n: They will only be removed if vanilla makes heavy breaking changes in this area. Otherwise we will keep them "forever".

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
haykam821: This boilerplate feature comes at barely any cost.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
LambdAurora: Sure love the community project absolutely wanting to be more stable than the game it mods. /s

It becomes ridiculous.

Juuxel: This PR isn't honestly the place to talk about this - we probably have old deprecated APIs from 1.14 and 1.15 still lying around so this isn't much different.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
liach: Mojang's design is to ensure that blocks are initialized before its entities, and this feature may encourage bad coding behavior of messing up initialization order.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
liach: Given mojang nukes stuff often, like text refactors or registry tags, I don't mind nuking more often as long has a clear replacement is offered.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
Juuxel: But this doesn't allow initialising the BET before its blocks are defined anyway since the builder will only be turned into a BET once all the `addBlock`s and such have been done. (And the feature has already been accepted in an earlier PR)

LambdAurora: > Mojang's design is to ensure that blocks are initialized before its entities, and this feature may encourage bad coding behavior of messing up initialization order.

Uh? No?
This is done **before** the block entity type is actually built and registered.

And even then, there's some use cases where adding supported blocks after the block entity type is added is needed, especially in the case where blocks are added through analysis for the registry, which means you never know when you can finally let go and register the block entity type.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
Technici4n: I had recent experiences with AE2 where Forge changed existing APIs unnecessarily, and I'm pretty sure MI is still using deprecated APIs because I haven't updated it to the new ones yet. I think Player had similar experiences. So, we keep APIs as long as maintaining them is reasonable.

If you disagree, then feel free to do it differently in your projects, but this is not going to change in Fabric anytime soon.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
Technici4n: The objective is to eventually create a new module with only the strictly necessary object builders, or migrate its contents to different modules with a clearer purpose, or a combination of both. As such, it's unnecessary to keep these APIs in the non-deprecated modules indefinitely. However, they will still be accessible in object builders v1 until mojang breaks this code.

haykam821: Regarding Mojang's design for how blocks and block entities are initialized, registry freezing allows data requiring both blocks and block entities to be calculated after both have been registered.

liach: We should move this addition functionality to something like `FabricBlockEntityTypeBuilderAccess`

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerProfessionBuilder.java
liach: We probably should start marking the version for deprecation and allow removing them eagerly in contributions if the version differences to the current version is large enough.

Technici4n: > allow removing them eagerly in contributions if the version differences to the current version is large enough

that will be done by the maintainers already, I don't think something needs to be done about it. :P

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/entity/FabricBlockEntityTypeBuilder.java
Technici4n: It's not really a matter of whatever Mojang wants, it's more about not keeping a class in fabric api just for this specific use case. It was originally introduced before TAWs were available, but now we can re-evaluate that decision.

Technici4n:
Reminder to myself to add TAWs for #1635 too.

Technici4n:
Suggestion from Discord:
> `net.minecraft.entity.ai.brain.Activity` and `net.minecraft.entity.ai.brain.sensor.SensorType` should probably get transitive access wideners for their constructors

Technici4n:
Cherry-picked to 1.19, and updated with a clearer PR description.

Technici4n:
FabricBlockEntityTypeBuilder won't be deprecated for now... Let's review and merge this PR now!

Shnupbups (approved):

modmuss50 (approved):

Player3324 (approved):

modmuss50:
Will move this to last call, didnt realise it was ready. Sorry 👍

ghost:
Next waiting for #2341 so i can finally cleanup my accesswidener file 👀

PR #2163, opened 2022-04-26
Add vanilla fallback for gametest structures

LLytho:
Hey!

The PR changes `StructureTestUtilMixin` to allow a fallback to the vanilla structure loading for gametests. The fallback will only happen if the structure defined through `@GameTest` was not found in `gametest/structures/*.snbt`, to ensure old tests are still working. 

Reminder what the fallback is: Using the structure manager from the level to find the `.nbt` structure. If this is not possible, it will take a look at `gameteststructures` in the root directory from the minecraft instance. 

Two reasons for this change: 
- Use already existing structures without the need to convert them into additional `.snbt` files. Also helpful for mods with custom structures.
- In a multiloader project it helps to use one structure format for everything as the default behavior is to load from the structure manager.


modmuss50:
Another idea I had when this issue was first brought to my attention was to add an API that allows mods to specify how they wish to load these structures.

Id have to check again, but I think the vanilla was was very sub-optimal for mods to use, the vanilla structures are no where to be found so IMO it was free game to change the way this worked for fabric. I am not a massive fan of having features that need maintaining just becuase its the forge way, we dont try to be complatible with forge anywhere else.

I am not against this PR, just wanted to get my other idea out there. I also believe you will have some checkstyle issues.

LLytho:
`to add an API that allows mods to specify how they wish to load these structures`
How would this look like? 

`I think the vanilla was was very sub-optimal for mods to use` 
How exactly? As the structure manager holds the resources with namespaces, there should be no problem to load structures. Otherwise I don't really see why choosing the `snbt` format? There is no benefit of using it compared to `nbt` format in which structures are stored normally. 

`I am not a massive fan of having features that need maintaining just becuase its the forge way` 
But this isn't really the case here, as the Fabric mixin removes the vanilla behavior of loading the structure. It has nothing to do with forge. The other Fabric stuff for the mixins are pretty well, where Forge decided to do some weird decision, like splitting up namespace and structure name into different fields in `@GameTest`, but this is probably nothing for this thread. 

`I also believe you will have some checkstyle issues.`
Yeah :D The import ordering was a bit rough to fix :D

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/StructureTestUtilMixin.java
```diff
@@ -38,26 +44,51 @@
...
 
 		try {
 			Resource resource = world.getServer().getResourceManager().getResource(structureId);
-			String snbt;
 
 			try (InputStream inputStream = resource.getInputStream()) {
 				snbt = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
 			}
+		} catch (IOException ignore) {
+			// Ignore
```
frqnny, line 61: Perhaps explain why you are ignoring it, or just log an error so you avoid crashing?

LLytho: An error message which combines an error for all three tries to find the structures is handled here:
https://github.com/FabricMC/fabric/pull/2163/files#diff-68f999bc4b47c492f226b5a47e7b54b644846784de57e477fff655c6ebefb2cdR89

If we would throw the exception there, we would not even try to find the structure in the `StructureManager`.  This is currently the behavior in fabric, which only tries to read from `/gametest/structure/*.snbt` and throws the IOException and exit the process.

Technici4n:
@modmuss50 Mind having a look when you have more time?

Technici4n:
Superseded by #2801

PR #2170, opened 2022-04-28
Create zh_hk.json

SolidBlock-cn:
add language file

### DIFF
--- a/fabric-item-groups-v0/src/main/resources/assets/fabric/lang/lzh.json
+++ b/fabric-item-groups-v0/src/main/resources/assets/fabric/lang/lzh.json
@@ -1,0 +1,3 @@
+{
+  "fabric.gui.creativeTabPage": "卷%2$s之%1$s"
+}
--- a/fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/zh_cn.json
+++ b/fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/zh_cn.json
@@ -1,0 +1,3 @@
+{
+  "pack.source.fabricmod": "Fabric模组"
+}
--- a/fabric-item-groups-v0/src/main/resources/assets/fabric/lang/zh_hk.json
+++ b/fabric-item-groups-v0/src/main/resources/assets/fabric/lang/zh_hk.json
@@ -1,0 +1,3 @@
+{
+  "fabric.gui.creativeTabPage": "第 %d/%d 頁"
+}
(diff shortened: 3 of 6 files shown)


frqnny (approved):

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/zh_cn.json
```diff
@@ -0,0 +1,3 @@
+{
+  "pack.source.fabricmod": "Fabric模组"
```
liach, line 2: Why no space here while space is present between character and words in other languages?

SolidBlock-cn: I'm just following the vanilla convention. In vanilla assets files, `zh-cn.json` usually does not leave spaces between character and words, while `zh-hk.json` and `zh_tw.json` do.

SolidBlock-cn: I'm just following the vanilla convention. In vanilla assets files, `zh-cn.json` usually does not leave spaces between character and words, while `zh-hk.json` and `zh_tw.json` do.

fabric-item-groups-v0/src/main/resources/assets/fabric/lang/lzh.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fabric.gui.creativeTabPage": "卷%2$s之%1$s"
```
liach, line 2: > I'm just following the vanilla convention. In vanilla assets files, `zh-cn.json` usually does not leave spaces between character and words, while `zh-hk.json` and `zh_tw.json` do.

This one has inconsistency with the other `lzh.json` then

SolidBlock-cn: That's also the convention of vanilla's `lzh.json`, where spaces are usually not left. And actually whether to leave space does not matter—both behaviours are acceptable.

liach: So you add space for english but not number? sure... cursed enough

liach (approved):

SolidBlock-cn (commented):
-

Player3324:
Looks like this was forgotten, why close?

PR #2171, opened 2022-04-28
Move deprecated modules, create own maven BOM for the deprecated modules.

modmuss50:
Move deprecated modules to a sub directory.
Remove deprecated modules from the default maven pom
Create a new "fabric-api-deprecated" bom with the deprecated modules.
Updates loom and gradle.
Move to spotless for license header checks.

Developers that wish to depend on the deprecated modules will need to use 

```gradle
modImplementation "net.fabricmc.fabric-api:fabric-api-deprecated:${project.fabric_version}"
```

Should this be backported to 1.18? or just stick to beign for 1.19?

Edit: Tested works as expected.


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         with:
           fetch-depth: 0
       - uses: gradle/wrapper-validation-action@v1
-      - run: ./gradlew check build publishToMavenLocal --stacktrace --parallel --warning-mode=fail
```
modmuss50: The build now has deprecation warnings, waiting on https://github.com/CadixDev/licenser/pull/40 before we can add this back.

I may move to spotless in another PR, it has other featues that we could make use of.

frqnny:
In my opinion there should be no troubles with mods using deprecated apis as long as the legacy module is included in the fat jar on curseforge. Port to 1.18 should be fine

modmuss50:
> In my opinion there should be no troubles with mods using deprecated apis as long as the legacy module is included in the fat jar on curseforge. Port to 1.18 should be fine

Yes, the fat jar on CF still includes all the modules so this does not affect players, the idea behind this is to remove them from the dev env by default. If a mod using one of these deprecated modules updates fabric api they will be unable to compile without manually adding them back.

All of these modules have been deprecated for quite a while so I hope it wont be too painful.

liach (approved):

.github/workflows/build.yml
modmuss50: Moved to spotless, spotless has additional featues that we way want to use in the future.

Player3324 (approved):

PR #2172, opened 2022-04-28
Fabric Data Attachment API

Technici4n:
This API allows attaching arbitrary data to various game objects such as worlds, chunks, entities and block entities. Data can be attached client-side and server-side, but no syncing is provided in the API. The data can optionally be persisted on the server side.

The use case is generally to store data for a specific game object directly with the game object instead of having to store it separately and rely on load/unload events.

This is similar to the "data attachment" part of Forge capabilities, however API queries should be handled by API Lookup as before.

## API overview
There are two classes in the API: `AttachmentType` and `AttachmentSerializer`.

`AttachmentType<A, T>` is the "key" identifying a data attachment of type `A` to game objects of type `T`. It also contains the `get`, `computeIfAbsent`, `set` and `remove` methods that actually allow one to retrieve, update, and remove attached values.

`AttachmentType`s should be stored in `static final` variables, and created using one of the `for*` constructors by passing an identifier, the attachment class, and optionally a serializer:
```java
public interface AttachmentType<A, T> {
	static <A> AttachmentType<A, BlockEntity> forBlockEntity(Identifier identifier, Class<A> attachmentClass, @Nullable AttachmentSerializer<A, ? super BlockEntity> serializer) { ... }
	static <A> AttachmentType<A, WorldChunk> forChunk(Identifier identifier, Class<A> attachmentClass, @Nullable AttachmentSerializer<A, ? super WorldChunk> serializer) { }
	// etc...
```

### Serialization
Attachments types may be registered with an `AttachmentSerializer<A, T>` if they wish to be persisted to NBT. The signature is self-explanatory, and there is also a helper to create a serializer from a DFU Codec:
```java
public interface AttachmentSerializer<A, T> {
	/**
	 * Serialize the value to a new NBT compound.
	 * If {@code null} is returned, the value will not be saved at all.
	 *
	 * @param value The current value of the attachment. Never {@code null}.
	 * @return The serialized attachment, or {@code null} if nothing should be saved.
	 */
	@Nullable
	NbtCompound toNbt(A value);

	/**
	 * Create a new instance from an NBT compound previously created by {@link #toNbt}.
	 * If {@code null} is returned, the instance will not be placed in the attachment target.
	 *
	 * @param target The target of this attachment. This can be used to capture a reference (for example to a host entity).
	 * @param nbt The nbt data of this attachment. Never {@code null}.
	 * @return The deserialized attachment, or {@code null} if no attachment should be loaded.
	 */
	@Nullable
	A fromNbt(T target, NbtCompound nbt);
```

### Using a mutable type
```java
class ManaContainer {
    int mana = 0;
}

public static final AttachmentType<ManaContainer, Entity> MANA_CONTAINER = AttachmentType.forEntity(
    new Identifier("mymod:mana_container"), // identifier for serialization
    ManaContainer.class, // class for type safety
    /* serializer, omitted here */ // or null if the attachment should not be serialized
);

PlayerEntity player = ...;
// Query
ManaContainer maybeContainer = MANA_CONTAINER.get(player);
// Increment mana
MANA_CONTAINER.computeIfAbsent(player, player -> new ManaContainer()).mana++;
// Remove
MANA_CONTAINER.remove(player);

// Entities are always persisted, so no markDirty call is needed.
```

### Using an immutable type
```java
public static final AttachmentType<Integer, Entity> MANA = AttachmentType.forEntity(
    new Identifier("mymod:mana"), // identifier for serialization
    Integer.class, // class for type safety
    AttachmentSerializer.fromCodec(Codec.INT) // codec-based serializers are very easy to obtain
);

PlayerEntity player = ...;
// Query
int currentMana = Objects.requireNonNullElse(MANA.get(player), 0);
// Increment mana
MANA.set(player, Objects.requireNonNullElse(MANA.get(player), 0) + 1);
// Remove
MANA.remove(player);
```

## TODO
- [x] Figure out why chunk attachments are not properly persisted?
- [ ] Make this API experimental (remember to change in the FMJ).
- [ ] Split networking javadoc fix to other PR.
- [ ] Mark internal classes as `@ApiStatus.Internal`.
- [ ] More utility methods in `AttachmentType`?
- [ ] Check that entities do indeed always get saved.
- [ ] Decide whether using `fabric:value` in `AttachmentSerializer.fromCodec` for codecs that don't result in an NbtCompound is acceptable.


maityyy:
I think we need a method that does this for you to reduce the code size
```java
// Increment mana
MANA_CONTAINER.computeIfAbsent(player, ManaContainer::new).mana++;
```
For example:
```java
MANA_CONTAINER.getOrCreate(player).mana++;
```

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTypeImpl.java
```diff
@@ -0,0 +1,84 @@
...
+			AttachmentTypeImpl<T> attachment = new AttachmentTypeImpl<>(identifier, attachmentClass, serializer);
+			newMap.put(identifier, attachment);
+			attachmentsMap = newMap;
+			return attachment;
+		}
+	}
+
+	@Nullable
+	public static AttachmentType<?> get(Identifier identifier) {
+		return attachmentsMap.get(identifier);
```
liach: This is not thread safe without synchronization, that a created attachment type may not be seen when get is called without extra synchronization because one thread has an outdated attachmentMap

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentContainer.java
```diff
@@ -0,0 +1,81 @@
...
+	 * Associate a value to the passed attachment type, and return the previous value or {@code null} if there was no previous value.
+	 */
+	@Nullable
+	<T> T set(AttachmentType<T> attachment, T value);
+
+	/**
+	 * Remove the value associated to the passed attachment type, and return it, or {@code null} if there was no previous value.
+	 */
+	@Nullable
+	<T> T remove(AttachmentType<T> attachment);
```
liach: Imo the default state of an attachment shouldn't be null, but should be regulated by the type.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentContainer.java
```diff
@@ -0,0 +1,81 @@
...
+
+import net.minecraft.nbt.NbtCompound;
+
+import net.fabricmc.fabric.impl.attachment.AttachmentContainerImpl;
+
+/**
+ * Maps {@linkplain AttachmentType attachment types} to attached values.
+ */
+@ApiStatus.NonExtendable
+public interface AttachmentContainer {
```
liach: Wish we can just use sealed, but javac doesn't let us unless we use named modules (jvm lets us though)

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentContainerImpl.java
```diff
@@ -0,0 +1,118 @@
...
+
+		return (T) attachments.remove(attachment);
+	}
+
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	@Override
+	public void writeNbt(NbtCompound nbt, String subtag) {
+		NbtCompound attachmentsTag = new NbtCompound();
+
+		for (Map.Entry<AttachmentType<?>, Object> entry : attachments.entrySet()) {
```
liach: use forEach lambda; identity hash map does not store by entry, so iterating entry set allocates extra objects

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTypeImpl.java
```diff
@@ -0,0 +1,84 @@
...
+	public static <T> AttachmentType<T> create(Identifier identifier, Class<T> attachmentClass, @Nullable Serializer<T> serializer) {
+		Objects.requireNonNull(identifier, "Identifier may not be null.");
+		Objects.requireNonNull(attachmentClass, "Attachment class may not be null.");
+
+		synchronized (AttachmentTypeImpl.class) {
+			if (attachmentsMap.containsKey(identifier)) {
+				throw new IllegalArgumentException("Duplicate AttachmentType creation for identifier " + identifier);
+			}
+
+			// Use copy-on-write here to ensure fast lock-free queries
```
liach: this is not to ensure fast query, but to ensure synchronization can publish updated maps.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTypeImpl.java
Technici4n: Right, I forgot it should be volatile.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
```diff
@@ -0,0 +1,69 @@
...
+
+		return fabric_dataAttachments;
+	}
+
+	@Dynamic
+	@Inject(
+			at = @At(value = "INVOKE", target = "net/minecraft/block/entity/BlockEntity.readNbt(Lnet/minecraft/nbt/NbtCompound;)V"),
+			method = "method_17897" // lambda body in BlockEntity#createFromNbt
+	)
+	public void injectReadNbt(NbtCompound nbt, String id, BlockEntity blockEntity, CallbackInfoReturnable<BlockEntity> cir) {
```
liach: This fails the build.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/EntityMixin.java
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		return fabric_dataAttachments;
+	}
+
+	@Inject(
+			at = @At(value = "INVOKE", target = "net/minecraft/entity/Entity.readCustomDataFromNbt(Lnet/minecraft/nbt/NbtCompound;)V"),
+			method = "readNbt"
+	)
+	public void injectReadNbt(NbtCompound nbt, CallbackInfoReturnable<NbtCompound> cir) {
```
liach: Same here

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/EntityMixin.java
```diff
@@ -0,0 +1,67 @@
...
+		if (fabric_dataAttachments.isEmpty()) {
+			fabric_dataAttachments = null;
+		}
+	}
+
+	@Inject(
+			at = @At(value = "INVOKE", target = "net/minecraft/entity/Entity.writeCustomDataToNbt(Lnet/minecraft/nbt/NbtCompound;)V"),
+			method = "writeNbt"
+	)
+	public void injectWriteNbt(NbtCompound nbt, CallbackInfoReturnable<NbtCompound> cir) {
```
liach: Same here

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTypeImpl.java
Technici4n: I tweaked the wording slightly.

Technici4n:
> I think we need a method that does this for you to reduce the code size
> 
> ```java
> // Increment mana
> MANA_CONTAINER.computeIfAbsent(player, ManaContainer::new).mana++;
> ```
> 
> For example:
> 
> ```java
> MANA_CONTAINER.getOrCreate(player).mana++;
> ```

@maityyy Should it just find a constructor without params via reflection? Or we could also have the `AttachmentType` provide a default value factory.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
Technici4n: Yeah I haven't actually tried running this since I'm still at the prototyping stage... Ignore this for now 😅

liach:
I recommend making the container getter method in `AttachmentTarget` abstract and having another `InternalAttachmnetTarget extends AttachmentTarget` in impl package that defaults the method instead. The `InternalAttachmnetTarget` will be injected into minecraft classes. This is to avoid mod classes implementing `AttachmentTarget` forgetting to override the method somehow.

liach:
> Or we could also have the `AttachmentType` provide a default value factory.

This is what I recommend, and we can pass the owner of the attachment container to the attachment type as well, to facilitate default value computation by context.

Technici4n:
> I recommend making the container getter method in `AttachmentTarget` abstract and having another `InternalAttachmnetTarget extends AttachmentTarget` in impl package that defaults the method instead. The `InternalAttachmnetTarget` will be injected into minecraft classes. This is to avoid mod classes implementing `AttachmentTarget` forgetting to override the method somehow.

I'll think about it, but I think injected interfaces should be part of API. Maybe I'll do it as a nested interface.

Technici4n:
> > Or we could also have the `AttachmentType` provide a default value factory.
> 
> This is what I recommend, and we can pass the owner of the attachment container to the attachment type as well, to facilitate default value computation by context.

We'd have to pass `AttachmentTarget` directly though, and the supplier would have to do an instanceof check?

liach:
> > I recommend making the container getter method in `AttachmentTarget` abstract and having another `InternalAttachmnetTarget extends AttachmentTarget` in impl package that defaults the method instead. The `InternalAttachmnetTarget` will be injected into minecraft classes. This is to avoid mod classes implementing `AttachmentTarget` forgetting to override the method somehow.
> 
> I'll think about it, but I think injected interfaces should be part of API. Maybe I'll do it as a nested interface.

The injected interface has a parent interface in the API, which should suffice. The extra interface serves solely to shut IDE noise when viewing the decompiled minecraft code, and to assure modders extending those classes that they don't need to implement the container getter as the API mixin'd it.

Technici4n:
I'm still not a fan since modders could be relying on it by accident. But I think custom `AttachmentTarget`s are niche enough for this not be too much of an issue? I think Player doesn't even want custom attachments, but IMO it comes "for free". ;)

(By "custom" I mean attachment to custom game objects, i.e. attachment that are not provided by fabric API itself).

liach:
> We'd have to pass `AttachmentTarget` directly though, and the supplier would have to do an instanceof check?

I believe passing `AttachmentTarget` directly suffices, as they have a lot of subclasses and extra type parameter cannot cover those use cases.

If you want to go dangerous, you can declare something like 
```java
public <O extends AttachmentTarget> Builder withDefault(Function<? super O, ? extends T> defaultFactory) {}
```
which would avoid explicit casting and make code look more fluent, but it depends on modders to choose the correct target type `O` and avoid class cast exceptions.

> I'm still not a fan since modders could be relying on it by accident.

The injected interface is already in the impl package, so it shouldn't really be a problem. Having modders implement `AttachmentTarget` but the compiler never tells them that they need to override `getAttachmentContainer` (and IDE won't generate stubs either) is more of a problem.

> But I think custom `AttachmentTarget`s are niche enough for this not be too much of an issue?

If modders shouldn't implement it, document it explicitly and add `@ApiStatus.NonExtendable`, than leaving a dangling interface that can be easily incorrectly implemented.

Technici4n:
> I believe passing `AttachmentTarget` directly suffices, as they have a lot of subclasses and extra type parameter cannot cover those use cases.

Yes, I think you're right.

> If you want to go dangerous, you can declare something like
> 
> ```java
> public <O extends AttachmentTarget> Builder withDefault(Function<? super O, ? extends T> defaultFactory) {}
> ```

Looks quite dangerous indeed. Will probably avoid this. 😄
 
> The injected interface is already in the impl package, so it shouldn't really be a problem. Having modders implement `AttachmentTarget` but the compiler never tells them that they need to override `getAttachmentContainer` (and IDE won't generate stubs either) is more of a problem.

It's not in the impl package? `AttachmentTarget` is injected and it's part of the api package.

> > But I think custom `AttachmentTarget`s are niche enough for this not be too much of an issue?
> 
> If modders shouldn't implement it, document it explicitly and add `@ApiStatus.NonExtendable`, than leaving a dangling interface that can be easily incorrectly implemented.

I mean that they are allowed, but I expect someone implementing attachments for a custom container to be considerate enough to notice that they need to override this method. It will also be an easy-to-fix crash on the first test they do, so while not ideal I think it's "acceptable".

Technici4n:
Other question: Should removing or updating values even be allowed? I think keeping components around forever is a lot simpler, and it shouldn't cause too much overhead... It also simplifies the API surface a lot.

liach:
> It's not in the impl package? AttachmentTarget is injected and it's part of the api package.

No, I mean something like this:
```java
package net.fabricmc.fabric.api.attachment.v1;
// ...
public interface AttachmentTarget {
	AttachmentContainer getAttachmentContainer();
}
```

```java
package net.fabricmc.fabric.impl.attachment;
// ...
public interface InternalAttachmentTarget extends AttachmentTarget {
	@Override
	default AttachmentContainer getAttachmentContainer() {
		throw new UnsupportedOperationException("Implemented via mixin, must be overridden for custom game objects.");
	}
}
```

and then you inject `net.fabricmc.fabric.impl.attachment.InternalAttachmentTarget` onto the minecraft classes.

Technici4n:
To be determined, I'm still not fan of exposing this since modders could end up referencing it by mistake. I really think the exception in the default method is not a huge deal. 😄

Other question: should `AttachmentTarget` include some sort of `markDirty` method too?

liach:
> still not fan of exposing this
Fabric API implementations have tons of internal accessor mixins in impl packages, and you don't fear exposing those though mods can call it freely. Curious.

> Other question: should AttachmentTarget include some sort of markDirty method too?

I'd suggest moving the dirty checking mechanism onto the attachment container. if an attachment needs to do something when it updates, its type should be able to handle it.

Technici4n:
> Fabric API implementations have tons of internal accessor mixins in impl packages, and you don't fear exposing those though mods can call it freely. Curious.

I know but if we use interface injection to inject something in MC source code, we're exposing these interfaces as part of an API imo as it's hard to tell what is impl and what is API. Whereas internal accessor mixins would need to be directly referenced by modders.

liach:
> > Fabric API implementations have tons of internal accessor mixins in impl packages, and you don't fear exposing those though mods can call it freely. Curious.
> 
> I know but if we use interface injection to inject something in MC source code, we're exposing these interfaces as part of an API imo as it's hard to tell what is impl and what is API. Whereas internal accessor mixins would need to be directly referenced by modders.

Sure, I just hope the interface mixing can create explicit dummy implementations so the downstream compiler and IDE won't make noise one day

Technici4n:
I updated the PR with a slightly different approach after some discussion with Player on discord. The support for custom attachments has been removed since it's somewhat useless, and the `IdentityHashMap<AttachmentType<?>, Object>` has been inlined into the attachment targets.

Technici4n:
Updated again after more discussion with Player, `AttachmentTarget` has been removed, and instead the target is now a generic parameter of `AttachmentType`.

Technici4n:
Updated to 1.19, added a nice testmod. Now waiting on @sfPlayer1 for a review of the overall design. Others: feel free to review too.

Technici4n:
Updated to 1.19, added a nice testmod. Now waiting on @sfPlayer1 for a review of the overall design. Others: feel free to review too.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -0,0 +1,211 @@
...
+	 * <p>Serialization will only be used on the logical server.
+	 * Entity data is always persisted, there is no need to call any method after modifying the attachments.
+	 *
+	 * <p>Note that data is not transferred automatically in the following cases, and requires special handling if desired:
+	 * <ul>
+	 *     <li>On the (logical) client and server: when a player dies and respawns.
+	 *     {@code ServerPlayerEvents.COPY_FROM} can be used to copy the data on the server.</li>
+	 *     <li>On the client: when a player changes dimensions.</li>
+	 *     <li>On the client and server: when a non-player entity changes dimensions.</li>
+	 * </ul>
```
Technici4n: Not a fan of this, maybe we should add ways to make this copying easier? I fear that it might make the implementation a lot heavier though.

fabric-data-attachment-api-v1/build.gradle
```diff
@@ -0,0 +1,10 @@
+archivesBaseName = "fabric-data-attachment-api-v1"
+version = getSubprojectVersion(project)
+
+moduleDependencies(project, [
+	'fabric-api-base',
+])
+
+dependencies {
+	testmodImplementation project(path: ':fabric-resource-loader-v0', configuration: 'namedElements')
```
modmuss50: 1.19 with split sources has a new way to specify test deps, look at the other modules.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentsPersistentState.java
```diff
@@ -0,0 +1,51 @@
...
+	public static AttachmentsPersistentState read(ServerWorld world, NbtCompound nbt) {
+		((AttachmentTargetImpl) world).readAttachmentsFromNbt(World.class, nbt);
+		return new AttachmentsPersistentState(world);
+	}
+
+	@Override
+	public boolean isDirty() {
+		// Only write data if there are attachments, or if we previously wrote data.
+		return wasSerialized || ((AttachmentTargetImpl) world).hasSerializableAttachments();
+	}
```
Technici4n, line 48: I tested that this will indeed only cause the data to be written once an attachment is added. So as long as this API is not used for worlds, there is no additional file in the world saves.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTypesMixin.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	@Override
+	@Nullable
+	public <T> T set(AttachmentType<T, ?> type, T value) {
+		if (fabric_dataAttachments == null) {
+			fabric_dataAttachments = new IdentityHashMap<>();
+		}
+
+		return (T) fabric_dataAttachments.put(type, value); // TODO: call markDirty if relevant? (also applies to other places)
```
Technici4n: Here I am thinking of having the API user deal with `markDirty` (or equivalent depending on the attachment target). Thoughts?

fabric-data-attachment-api-v1/build.gradle
Technici4n: Good catch. Removed the testmod dep entirely because resource loader is added automatically.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentsPersistentState.java
modmuss50: Thats good, I was going to ask about, I was unsure if creating the persitent state was enough to write a file. 👍

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
modmuss50: I think without this, its going to cause a number of unexpected bugs. Its not trivial for mods to handle all of these cases on their own either.

I would copy by default, and let the mods remove the data if they dont want that.

Technici4n: We need some sort of copy handler, because the attachments might capture a reference to the Entity. I can add that as an additional parameter to `forEntity`.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -0,0 +1,211 @@
...
+	 *
+	 * @throws IllegalArgumentException If an attachment type with the given identifier already exists for block entities.
+	 */
+	static <A> AttachmentType<A, Entity> forEntity(Identifier identifier, Class<A> attachmentClass, @Nullable Serializer<A, ? super Entity> serializer) {
```
Technici4n: A general concern is that currently the attachments are not synced to the client. While it is debatable whether block entity, chunk and world attachments should be synced, I think it's pretty clear that syncing entity attachments is very useful.

Thoughts on adding optional entity attachment syncing? It might make implementation quite a bit messier (see https://github.com/MinecraftForge/MinecraftForge/pull/8706), but it will definitely make the API more useful I think. Block entity might also not be too hard to add.

Technici4n:
Updated to 1.19.2. A few remarks:

- Syncing entity data to clients is a headache that I don't want to deal with for the moment. However it can easily be added after-the-fact (possibly by someone more familiar with it than me), which means that it can be omitted from the first version. For example, one could imagine the following API extension to register a sync handler:
```java
AttachmentType<Something, Entity> SOMETHING = ...;
AttachmentSyncing.registerForEntity(SOMETHING, new AttachmentSyncHandler<>() { ... });
```
- It turns out that when an entity changes dimension on the server, it keeps any serializable attachments. I need to check the other corner cases, but we probably don't need some to specify an entity copying strategy since this is already handled correctly by saving to NBT and deserializing.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
Technici4n: It turns out that serializable attachments are handled correctly. Player deaths require custom handling using the `COPY_FROM` event, which is imo good (you may not want to keep e.g. mana if the player dies).

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTypesMixin.java
Technici4n: The user has to deal with it, there is no sane way for us to automatically do it.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
Technici4n: I decided that it is too hard for now, but it can be added later if desired. See https://github.com/FabricMC/fabric/pull/2172#issuecomment-1256576448.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentContainer.java
Technici4n: Would conflict with the usual api/impl split of Fabric anyway. :wink:

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentContainer.java
Technici4n: That's going to be too messy when the attachment wants to capture a reference to the entity imo. `computeIfAbsent` should be fine. If this proves to be too bad we can always go for a v2 API... this PR only adds 2 API classes, so not the end of the world.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentContainerImpl.java
Technici4n: In the current version of the PR I don't think I can use forEach without introducing a capturing lambda.

Technici4n:
I just went for a second documentation pass, and I updated the PR description. I resolved all previous comments as they were mostly outdated. Please open new ones if you would like to discuss something.

Waiting for reviews now! :wink:

fabric-data-attachment-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,28 @@
...
+    "homepage": "https://fabricmc.net",
+    "irc": "irc://irc.esper.net:6667/fabric",
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": "*"
```
apple502j, line 19:
```suggestion
    "fabricloader": ">=0.14.9",
    "minecraft": ">=1.19.2"
```

fabric-data-attachment-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "description": "Allows data to be attached to common game objects such as worlds, chunks and entities.",
```
apple502j, line 21:
```suggestion
  "description": "Allows data to be attached to common game objects such as worlds, chunks, and entities.",
```

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
```diff
@@ -0,0 +1,15 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.attachment",
+  "compatibilityLevel": "JAVA_16",
```
apple502j, line 4:
```suggestion
  "compatibilityLevel": "JAVA_17",
```

apple502j (approved):
some minor stuff, otherwise LGTM?

mrjasonn:
This PR can be closed safely as #3476 has implemented a Data Attachment API

PR #2175, opened 2022-04-30
Add ModifyItemAttributeModifiersCallback

### LINKED ISSUE #2120: ItemStack.getAttributeModifiers can't be easily influenced for foreign items
didnt create a crashlog but this is the log file i do have. 

https://pastebin.com/vK5EfqvU

frqnny:
Closes #2120 

Adds an ItemAttributeModifiersCallback

IDK about you guys but I feel shaky about the stuff in ItemStackMixin.

I did add a test and that test is also the example code in the javadoc.

Below is a link to the image showing that the test mod works
https://discord.com/channels/507304429255393322/566276937035546624/970066641817858058

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
@@ -46,13 +46,13 @@ import net.fabricmc.fabric.api.event.EventFactory;
  */
 @FunctionalInterface
 public interface ItemAttributeModifiersCallback {
-	void onGetAttributeModifiers(ItemStack stack, EquipmentSlot slot, Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers);
+	void addAttributeModifiers(ItemStack stack, EquipmentSlot slot, Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers);
 
 	Event<ItemAttributeModifiersCallback> EVENT = EventFactory.createArrayBacked(
 			ItemAttributeModifiersCallback.class,
 			callbacks -> (stack, slot, attributeModifiers) -> {
 				for (ItemAttributeModifiersCallback callback : callbacks) {
-					callback.onGetAttributeModifiers(stack, slot, attributeModifiers);
+					callback.addAttributeModifiers(stack, slot, attributeModifiers);
 				}
 			}
 	);
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
@@ -30,7 +30,8 @@ import net.fabricmc.fabric.api.event.EventFactory;
  * Stack-aware attribute modifier callback for foreign items.
  * Instead of using Mixin to change attribute modifiers in items not in your mod,
  * you can use this event instead, either checking the Item itself or using a tag.
- * Do not use for your own Item classes; see {@link FabricItem}.
+ * This event provides you with a guaranteed mutable map you can put attribute modifiers in.
+ * Do not use for your own Item classes; see {@link FabricItem#getAttributeModifiers} instead.
  * For example, the following code modifies a Diamond Helmet to give you five extra hearts when wearing.
  *
  * <pre>
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import com.google.common.collect.Multimap;
+
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Stack-aware attribute modifier callback for foreign items.
+ * Instead of using Mixin to change attribute modifiers in items not in your mod,
+ * you can use this event instead, either checking the Item itself or using a tag.
+ * Do not use for your own Item classes; see {@link FabricItem}.
+ * For example, the following code modifies a Diamond Helmet to give you five extra hearts when wearing.
+ *
+ * <pre>
+ * {@code
+ *     ItemAttributeModifiersCallback.EVENT.register((stack, slot, attributeModifiers) -> {
+ * 			if (stack.isOf(Items.DIAMOND_HELMET) && slot.getEntitySlotId() == HEAD_SLOT_ID) {
+ * 				attributeModifiers.put(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER);
+ *          }
+ *     });
+ * }
+ * </pre>
+ */
+@FunctionalInterface
+public interface ItemAttributeModifiersCallback {
+	void onGetAttributeModifiers(ItemStack stack, EquipmentSlot slot, Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers);
+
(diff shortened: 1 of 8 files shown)


Technici4n (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Stack-aware attribute modifier callback for foreign items.
+ * Instead of using Mixin to change attribute modifiers in items not in your mod,
+ * you can use this event instead, either checking the Item itself or using a tag.
+ * Do not use for your own Item classes; see {@link FabricItem}.
```
Technici4n: This should probably link the FabricItem method directly.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
```diff
@@ -0,0 +1,58 @@
...
+ * 			if (stack.isOf(Items.DIAMOND_HELMET) && slot.getEntitySlotId() == HEAD_SLOT_ID) {
+ * 				attributeModifiers.put(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER);
+ *          }
+ *     });
+ * }
+ * </pre>
+ */
+@FunctionalInterface
+public interface ItemAttributeModifiersCallback {
+	void onGetAttributeModifiers(ItemStack stack, EquipmentSlot slot, Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers);
```
Technici4n: Could use some javadoc. Mentioning that the Multimap can be updated and is guaranteed to be mutable seems quite valuable in particular.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
frqnny: done, see latest commit

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
frqnny: done, see latest commit

PseudoDistant (approved):
Yes

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
```diff
@@ -0,0 +1,59 @@
...
+ * For example, the following code modifies a Diamond Helmet to give you five extra hearts when wearing.
+ *
+ * <pre>
+ * {@code
+ *     ItemAttributeModifiersCallback.EVENT.register((stack, slot, attributeModifiers) -> {
+ * 			if (stack.isOf(Items.DIAMOND_HELMET) && slot.getEntitySlotId() == HEAD_SLOT_ID) {
+ * 				attributeModifiers.put(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER);
+ *          }
+ *     });
+ * }
```
Juuxel: The indentation here seems to be a bit wonky (mix of tabs and spaces?)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
```diff
@@ -0,0 +1,59 @@
...
+ * </pre>
+ */
+@FunctionalInterface
+public interface ItemAttributeModifiersCallback {
+	void onGetAttributeModifiers(ItemStack stack, EquipmentSlot slot, Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers);
```
Juuxel: Maybe just `addAttributeModifiers` or `modifyAttributeModifiers`? `onGet` sounds like a mixin inject method

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
frqnny: how do i fix (i am out of my home so i will fix later)

Juuxel:
```suggestion
 * {@code
 * ItemAttributeModifiersCallback.EVENT.register((stack, slot, attributeModifiers) -> {
 * 	if (stack.isOf(Items.DIAMOND_HELMET) && slot.getEntitySlotId() == HEAD_SLOT_ID) {
 * 		attributeModifiers.put(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER);
 * 	}
 * });
 * }
```

frqnny: so kind thank you

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ItemAttributeModifiersCallback.java
frqnny: Just changed it to `addAttributeModifiers` per your request.

Technici4n: Technically you can also modify them... The callback could also be called `ModifyItemAttributeModifiers` in that case.

frqnny: Yes, you are right. I will change it to `modifyAttributeModifiers` and `ModifyItemAttributeModifiers`

maityyy:
Does it work correctly when only on the server side?

frqnny:
Maityyy I don't know, any recommendations on how to test

Technici4n:
No idea what happens if it's only done on the server side, but it's imo not out problem :P

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
```diff
@@ -80,7 +83,12 @@ private <T extends LivingEntity> void clearDamager(int amount, T entity, Consume
 			)
 	)
 	public Multimap<EntityAttribute, EntityAttributeModifier> hookGetAttributeModifiers(Item item, EquipmentSlot slot) {
-		return item.getAttributeModifiers((ItemStack) (Object) this, slot);
+		ItemStack stack = (ItemStack) (Object) this;
+		//we need to ensure it is modifiable for the callback
+		Multimap<EntityAttribute, EntityAttributeModifier> attributeModifiers = HashMultimap.create(item.getAttributeModifiers(stack, slot));
+		ModifyItemAttributeModifiersCallback.EVENT.invoker().modifyAttributeModifiers(stack, slot, attributeModifiers);
+		//now we can turn it back to immutable
+		return ImmutableMultimap.copyOf(attributeModifiers);
```
deirn: I don't think it needs to be immutable as vanilla returns a mutable map if the nbt contains `AttributeModifiers` key.

maityyy:
> Maityyy I don't know, any recommendations on how to test

Just change the attributes of any item from the server and see if that item's description is updated on the vanilla client without any mods.

I think it will not work. The server will send a packet of updated attributes and the attributes will be applied to the entity, but the item description will not be updated because the client will not have the new version of the `ItemStack::getAttributeModifiers` method that you are modifying. See `LivingEntity::getEquipmentChanges`.

You can write new attributes to the NBT of the item stack and then the description will update for sure. I hope this happens, because then this callback will become even more useful for developing mods only for the server side.

Technici4n:
@maityyy the callback is not meant for server-side mods, but just to allow mods to modify the return of `IFabricItem#getAttributeModifiers`, for example to tweak vanilla items without having to edit their NBT.

deirn (approved):

maityyy:
Well, with the current PR impl, it looks like I can do something like this, which should work just fine on the server side. But, to be honest, I would like to see this functionality in FAPI, because I do not understand who needs the desync of attributes with the NBT of the item stack.
```java
int HEAD_SLOT_ID = 3;
var MODIFIER_ID = UUID.fromString("blablabla");
var MODIFIER = new EntityAttributeModifier(MODIFIER_ID, "FAPI Test Modifier", 5, Operation.ADDITION);

ModifyItemAttributeModifiersCallback.EVENT.register((stack, slot, modifiers) -> {
	if (stack.isOf(Items.DIAMOND_HELMET) && slot.getEntitySlotId() == HEAD_SLOT_ID) {
		// modifiers.put(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER);
		stack.addAttributeModifier(EntityAttributes.GENERIC_MAX_HEALTH, MODIFIER, HEAD_SLOT_ID);
	}
});
```

TheEpicBlock:
This will not work server-side only. But implementing it like that would require adding things into the nbt, which would get really messy if not done properly. `getAttributeModifiers` is supposed to return the default attributes in case there's no nbt present. I don't think it's Fapi's responsibility to make this work. Polymer or PolyMc have been made for these kinds of use-cases and both should work fine in combination with this pr.

Technici4n:
The default attributes can depend on the current state of the stack or even the player. For example, good luck doing that with vanilla custom attribute modifiers: https://github.com/AztechMC/Modern-Industrialization/blob/master/src/main/java/aztech/modern_industrialization/items/diesel_tools/DieselToolItem.java#L128.

modmuss50 (approved):

PR #2179, opened 2022-05-01
Split client only code into its own sourceset to provide compile time sidedness saftey.

modmuss50:
### What and why

A common source of crashes on modded Minecraft servers comes from modders accidently calling client only code from the client, this PR is another large step towards elimitating that.

This PR has been months in the making and years in the planning, requiring major changes to [Loom](https://github.com/FabricMC/fabric-loom/pull/561) & [Loader](https://github.com/FabricMC/fabric-loader/pull/630). In recent Minecraft versions Mojang has made it easier than ever to cleanly split the jar, going against the status-quo of merging the client and server into one jar.

From the start we have designed Fabric to have a very clear split between client and common (client & server) code. Fabric has always encoraged keeping client only code seprate from the server, this can be seen at a fundamental level with the entrypoints in Loader. Fabric API's have all been designed with this mind.

This PR provides a compile safety net around Fabric API using client only code on the server. Even though there are almost 400 changed files, minimal changes beyond moving the files were required to achieve this in Fabric API, thanks to the effort of all contributors in the past.

These changes should not affect modders or players in anyway, a single "universal" jar is still produced. Im happy to awnswer any questions.

### TODO

- [x] Fix a loom issue causing invalid modclasspath groups.
- [x] Invistage why it fails to build. (I expect another loom issue)
- [x] Ensure that users of other IDEs (Eclipse and Vscode?) can still develop fabric api.
- [x] Testing to make sure nothing broke.

### Future work

- [ ] Loom should re-split Fabric API (and any other mods you depend on) to provide compile time safety for mods using these APIs. Loom already includes the client only entries in the jar's manifest.
- [ ] Enoruage modders to opt-in to compile time sidedness saftey.


warjort:
> Ensure that users of other IDEs (Eclipse and Vscode?) can still develop fabric api.

This has been broken for a while. 
Eclipse builds don't support the circular project references caused by tests referencing a dependency of the main project artifact.
e.g. fabric-api-base(test) -> fabric-lifecycle-events -> fabric-api-base

altrisi:
I develop in Eclipse, and I have no issues with the fabric api project in my IDE right now.

I think I had to change something about circular deps, but I've had no issue with it since.

warjort:
Was it this? 
https://stackoverflow.com/questions/48609810/how-to-solve-a-cyclic-project-dependency-in-eclipse-any-way-to-make-one-of-the
I wonder how it knows what order to (re)build projects?

altrisi:
Something like that yeah, it was just a setting where I did Error -> Warn (and could have set it to info or ignore IIRC). I think it's just that in some build systems that can cause issues and therefore defaults to error, but it hasn't caused any so far for me (granted, I have only touched two or three modules so I don't know if others may fail). I don't think it should though given if it wouldn't be able to resolve the dependencies it wouldn't let you disable it.

modmuss50:
It runs and builds (only through gradle so far)! Had to make one small change to the lifecycle tick event mixins.

modmuss50:
This seems to be working well for me now, would be good for others to clone this down (I'd reccomend to a new dir) and try it themselves.

Last thing to do is to update to a loader with https://github.com/FabricMC/fabric-loader/pull/674

Outside of a fabric api dev env everything should be unchanged.

Technici4n (approved):

LambdAurora:
That was fast to merge.

modmuss50:
It’s been open for 20 days, was becoming a bit of a pain to merge changes in so best to get it out.

Technici4n:
There was quite a bit of testing done for this. For example I ported [one of my mods](https://github.com/Technici4n/Modern-Dynamics/pull/12) to this new system to check that it works fine. If there's an issue, just let us know and we'll try to fix it ASAP.

modmuss50:
Yes, if your wondering why it didnt go through last call, that was becuase there was no changes to any API. It should work identically for anyone consuming it (via a mod dev env or a player).

PR #2189, opened 2022-05-03
Fix #2164: Transfer API performance issues

### LINKED ISSUE #2164: Improve transfer API performance
See https://spark.lucko.me/r99k4902fX.

The `ThreadLocal`s seem to incur some overhead, and that should be addressed. Maybe other things too. Obviously there is a limit to what can be done.

### LINKED ISSUE #2201: Oversized stacks should be handled more gracefully
As per #2196 and its discussion #2199

Code like this
https://github.com/FabricMC/fabric/blob/1d2d03ecd5a4508d14d4bf0647324afef1b8d2bd/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java#L113
can lead to unloadble worlds when the capacity < stack size. 

It will always crash because working item amounts become negative which fails sanity checks elsewhere.
The crash means it can't be fixed in game.

Reasons for this to occur maybe:
* A bug in a mod
* Somebody cheating in an oversize stack
* Installing a mod that increases max stack sizes then removing that mod

The transfer api should handle these oversize stacks more gracefully without crashing.

Technici4n:
- This PR fixes the large performance issue in `InventorySlotWrapper#setStack` due to `ThreadLocal`, as can be seen in https://spark.lucko.me/ZOTanHYIYN. This doesn't fix the problem entirely, but most other issues are to blame on other mods. :P
I have decided to use mixin to directly store that info in the relevant `Inventory` subclasses instead.
image
image
image
- Another optimization is to cache in `ItemVariantImpl` the `ItemStack` used for `Inventory.isValid` and `SidedInventory.canInsert/canExtract` to avoid needing to create a new stack every time.
- I also took the chance to slightly optimize `SingleStackStorage#isResourceBlank`, but that is very minor.
- Also fixes #2201.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
@@ -23,6 +23,7 @@ import org.slf4j.Logger;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
 import net.minecraft.item.Items;
 import net.minecraft.nbt.NbtCompound;
 import net.minecraft.network.PacketByteBuf;
@@ -48,6 +49,10 @@ public class ItemVariantImpl implements ItemVariant {
 	private final Item item;
 	private final @Nullable NbtCompound nbt;
 	private final int hashCode;
+	/**
+	 * Lazily computed, equivalent to calling toStack(1). <b>MAKE SURE IT IS NEVER MODIFIED!</b>
+	 */
+	private volatile @Nullable ItemStack cachedStack = null;
 
 	public ItemVariantImpl(Item item, NbtCompound nbt) {
 		this.item = item;
@@ -134,5 +139,16 @@ public class ItemVariantImpl implements ItemVariant {
 	@Override
 	public int hashCode() {
 		return hashCode;
+	}
+
+	public ItemStack getCachedStack() {
+		ItemStack ret = cachedStack;
+
+		if (ret == null) {
+			// benign data race if called from multiple threads
+			cachedStack = ret = toStack();
+		}
+
+		return ret;
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
@@ -145,7 +145,7 @@ public class ItemVariantImpl implements ItemVariant {
 		ItemStack ret = cachedStack;
 
 		if (ret == null) {
-			// benign data race if called from multiple threads
+			// multiple stacks could be created at the same time by different threads, but that is not an issue
 			cachedStack = ret = toStack();
 		}
 
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java
@@ -123,9 +123,9 @@ public abstract class SingleStackStorage extends SnapshotParticipant<ItemStack> 
 				}
 
 				setStack(currentStack);
-			}
 
-			return insertedAmount;
+				return insertedAmount;
+			}
 		}
 
 		return 0;
@@ -145,9 +145,9 @@ public abstract class SingleStackStorage extends SnapshotParticipant<ItemStack> 
 				currentStack = getStack();
 				currentStack.decrement(extracted);
 				setStack(currentStack);
-			}
 
-			return extracted;
+				return extracted;
+			}
 		}
 
 		return 0;
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleStackStorage.java
@@ -85,7 +85,7 @@ public abstract class SingleStackStorage extends SnapshotParticipant<ItemStack> 
 
 	@Override
 	public boolean isResourceBlank() {
-		return getResource().isBlank();
+		return getStack().isEmpty();
 	}
 
 	@Override
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/SingleVariantStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/SingleVariantStorage.java
@@ -81,9 +81,9 @@ public abstract class SingleVariantStorage<T extends TransferVariant<?>> extends
 				} else {
 					amount += insertedAmount;
 				}
-			}
 
-			return insertedAmount;
+				return insertedAmount;
+			}
 		}
 
 		return 0;
@@ -103,9 +103,9 @@ public abstract class SingleVariantStorage<T extends TransferVariant<?>> extends
 				if (amount == 0) {
 					variant = getBlankVariant();
 				}
-			}
 
-			return extractedAmount;
+				return extractedAmount;
+			}
 		}
 
 		return 0;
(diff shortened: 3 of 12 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
```diff
@@ -48,6 +49,10 @@ public static ItemVariant of(Item item, @Nullable NbtCompound tag) {
 	private final Item item;
 	private final @Nullable NbtCompound nbt;
 	private final int hashCode;
+	/**
+	 * Lazily computed, equivalent to calling toStack(1). <b>MAKE SURE IT IS NEVER MODIFIED!</b>
+	 */
+	private volatile @Nullable ItemStack cachedStack = null;
```
liach, line 55:
```suggestion
	private @Nullable ItemStack cachedStack;
```
Since you already have the data race guard in `getCachedStack()`. `volatile` is expensive.

Technici4n: I think it's still needed since there is no guarantee that the ItemStack reference will be set and properly initialized _after_ the ItemStack is fully constructed.

Technici4n: Maybe @sfPlayer1 has a better suggestion here? Would the following be correct?
```java
private @Nullable ItemStack cachedStack = null;
public ItemStack getCachedStack() {
	ItemStack ret = cachedStack;

	if (ret == null) {
		ret = toStack();
		VarHandle.fullFence(); // Ensure stack is fully constructed before it's assigned to cachedStack
		cachedStack = ret;
	}

	return ret;
}
```

liach: @Technici4n Why can't the cached stack be constructed multiple times, on different threads? Is there a special requirement here that the `cachedStack` must be the same reference than equivalent objects?

If there is no identity requirement, constructing two itemstacks is vastly cheaper than synchronizing or fencing (see benchmark results from openjdk/jdk#6138). In addition, if we do have the identity requirement, we should find a way to get rid of it.

liach: Hmm, I think you can probably use a `storeStoreFence` instead: https://github.com/openjdk/jdk/blob/5a768f75c9cb013edbf6c61e79820bd180cad4ba/test/micro/org/openjdk/bench/vm/fences/SafePublishing.java#L76

liach: @Technici4n Still need to fix this. Using a full fence is still better than volatile, which slows down every read.

Technici4n: Yeah, I'm just not sure it's completely correct. 😅

Technici4n: No idea, I would like Player to have a look if we decide to go for a fence. But I think keeping volatile for now should be fine performance-wise.

Player3324: The recommended VarHandle pattern btw is to use volatile on the field and then do lesser accesses through handles as appropriate

warjort: I am no expert on this, but my understanding is that the suggested fence implementation probably doesn't work?
````
private @Nullable ItemStack cachedStack = null;
public ItemStack getCachedStack() {
	ItemStack ret = cachedStack; // R1

	if (ret == null) {
		ret = toStack();
		VarHandle.fullFence(); // Ensure stack is fully constructed before it's assigned to cachedStack
		cachedStack = ret; // R2
	}

	return ret;
}
````
The issue is that although the ````fullFence()```` will flush everything to the cpu caches and main memory which seems to be same as volatile, the main contract in the java memory model is a "happens before" that the compiler can see/calculate to generate correctly ordered code. This exists for R2 above, but not R1. While it does exist at R1 if cachedStack is volatile.

warjort: I believe the java fence stuff is just a passthrough of the c++ stuff?
https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence if you want to check the rules.
Basically for it work, it needs a synchronize, park, volatile r/w, etc. somewhere on the reader thread competing with the constucting thread which is not true without the volatile on the cachedStack.

liach: This is only to ensure if the `cachedStack` is read to be `non-null` in a benign-race, we can see the nbt data (non-final) written to the stack object. We don't require `cachedStack` to be always visible after it is written (like volatiles) as another thread can just create another cached stack on-demand; but when one thread's cache is seen by another thread, we require its nbt data (non-final, prepared by `toStack()`) to be seen as well, which is what the fence does.

warjort:
````
but when one thread's cache is seen by another thread, we require its nbt data (non-final, prepared by toStack()) to be seen as well, which is what the fence does
````
But I don't think it does, except when you have other heuristics like x86's memory writes are always globally visual once flushed (not true generally in "weaker" architectures).
The issue is that without the "synchronization" somewhere in the reading thread there's no guarantee it runs the loads in the correct order, or even that it does any loads at all, not even from the cpu cache - the compiler is within its rights to generate code that uses established but "out-of-date" values in registers or the cpu uses a different cache/layer.

Maybe you are correct that in this case these bad things can't happen? 

I believe there would be less of a problem if the cache was immutable/static final because the spec requires the compiler to go the extra mile to correctly order and completly construct things, but that is only relevant for things like lazy immutable singletons.

Technici4n: I'm really not sure, so we should probably just keep the volatile. Acquire/release might work, but volatile is a lot easier to write, maintain, and reason about, so I suggest keeping it for now.

liach: I think a store-store fence is enough to create the final semantics. https://github.com/openjdk/jdk/blob/9583e3657e43cc1c6f2101a64534564db2a9bd84/src/java.base/share/classes/java/lang/invoke/ConstantCallSite.java#L52

Notice our fence applies between the write to stack content and the write to `cachedStack` field. On the reader side, we always have that if the stack nbt is read, `cachedStack` pointer is already there.

Since synchronization comes at such a great cost, then I'd recommend not cache unless the stack has complex nbt data if you think fences are unacceptable.

liach: > I'm really not sure, so we should probably just keep the volatile. Acquire/release might work, but volatile is a lot easier to write, maintain, and reason about, so I suggest keeping it for now.

I'd suggest removing caching once and for all. Making a single item stack (especially most item keys don't have nbt) would be cheaper than synchronizing.

Technici4n: Hmm well, we also have the option of using an ItemStack as the backing storage of course.

liach: > Hmm well, we also have the option of using an ItemStack as the backing storage of course.

Uh, I mean to roll back all the changes of `toStack` to `getCachedStack` back to plain `toStack()`.

Technici4n: That doesn't make sense imo, I ran some tests locally and it vastly improved the performance in one of my tests. I don't think volatile is that slow, after all it's used for events.

Technici4n: (Or even API Lookup)

modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java
liach: We can keep this for now. I can write a jcstress test later

PR #2193, opened 2022-05-05
Add registries for villager interactions to fabric-content-registries module

aws404:
This PR adds the following content registries for villager interactions:
- VillagerCollectableRegistry: Allows changing the items which villagers can collect into their inventory
- VillagerCompostableRegistry: Allows changing which items farmer villagers can use with a composter
- VillagerFoodRegistry: Allows changing which items villagers can use as food
- VillagerPlantableRegistry: Allows changing which crops farmer villagers can plant
- VillagerHeroGiftRegisty: Allows adding to the map of hero of the village loot tables

Example uses can be seen in the test mod


frqnny:
Do you think this PR could also cover adding COMPOSTABLES from FarmerWorkTask as well? It would simplify this Mixin for me in my mod: https://github.com/frqnny/TacoCraft/blob/1.19.x/src/main/java/io/github/frqnny/tacocraft/mixin/MixinFarmerWorkTask.java

aws404:
Thats a good catch, I will also seperate into VillagerCollectablesRegisty, VillagerFoodRegistry and VillagerCompostingRegistry

aws404:
I have combined all the registries into one VillagerInteractionRegistries class, and now use ImmutableMap/CollectionRedirect classes to modify most of the maps/collections.

Looking for reviews/opinions on this change.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,111 @@
...
+		return ImmutableCollectionUtils.getAsMutableSet(VillagerEntityAccessor::getGatherableItems, VillagerEntityAccessor::setGatherableItems);
+	}
+
+	private static List<Item> getCompostableRegistry() {
+		return ImmutableCollectionUtils.getAsMutableList(FarmerWorkTaskAccessor::getCompostables, FarmerWorkTaskAccessor::setCompostables);
+	}
+
+	private static Map<Item, Integer> getFoodRegistry() {
+		return ImmutableCollectionUtils.getAsMutableMap(() -> VillagerEntity.ITEM_FOOD_VALUES, VillagerEntityAccessor::setItemFoodValues);
+	}
```
deirn: I'd just inject into `<clinit>` tail and replace the collection with mutable ones.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/GiveGiftsToHeroTaskAccessor.java
```diff
@@ -0,0 +1,34 @@
...
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import net.minecraft.entity.ai.brain.task.GiveGiftsToHeroTask;
+import net.minecraft.util.Identifier;
+import net.minecraft.village.VillagerProfession;
+
+@Mixin(GiveGiftsToHeroTask.class)
+public interface GiveGiftsToHeroTaskAccessor {
+	@Accessor("GIFTS")
+	static Map<VillagerProfession, Identifier> getGifts() {
```
warjort: This is a bad name. If some other mod does the same thing they will likely choose the same (obvious) name.
You then have a conflict because of duplicate method definitions in the same class.

I believe the standard (though not always followed) is to prepend "fabric_", e.g.
https://github.com/FabricMC/fabric/blob/1d2d03ecd5a4508d14d4bf0647324afef1b8d2bd/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/BucketItemAccessor.java#L28

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/FarmerVillagerTaskMixin.java
```diff
@@ -0,0 +1,69 @@
...
+	@Shadow private BlockPos currentTarget;
+
+	@Inject(method = "keepRunning", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/passive/VillagerEntity;getInventory()Lnet/minecraft/inventory/SimpleInventory;"))
+	private void fabric_useRegistry(ServerWorld serverWorld, VillagerEntity villagerEntity, long l, CallbackInfo ci) {
+		SimpleInventory simpleInventory = villagerEntity.getInventory();
+
+		for (int i = 0; i < simpleInventory.size(); ++i) {
+			ItemStack itemStack = simpleInventory.getStack(i);
+
+			if (VillagerPlantableRegistry.INSTANCE.contains(itemStack.getItem())) {
```
warjort: Should this check for an empty stack?
What happens if a mod mistakenly registers "air" as a plantable?
This happens more often than you think if people aren't careful because "air" is both the default item and the empty marker.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
aws404: The way I have seen the items modded into the collections in the past is something along the lines of:
```java
collection = ImmutableList.builder().addAll(collection).add(MyNewItem).build()
``` 
Which if injected after the would completely break that implementation.

The current implemetation ensures that the list will always be mutable before we try to add items, no matter what else messes with it.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/GiveGiftsToHeroTaskAccessor.java
aws404: I have fixed this for all the mixins added in this PR

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/FarmerVillagerTaskMixin.java
aws404: Good catch, I have added an !stack.isEmpty check

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/GiveGiftsToHeroTaskAccessor.java
BasiqueEvangelist: Huh? Why would this be a problem? The accessor is static, and won't Mixin merge accessors to the same field anyway?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemConvertible;
+
+import net.fabricmc.fabric.impl.content.registry.VillagerPlantableRegistryImpl;
+
+/**
+ * Registry of items that farmer villagers can plant on farmland.
+ */
+public interface VillagerPlantableRegistry {
+	VillagerPlantableRegistry INSTANCE = new VillagerPlantableRegistryImpl();
```
Technici4n: We don't use the `INSTANCE` pattern anymore. Could you turn the methods here into static methods?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/FarmerVillagerTaskMixin.java
```diff
@@ -0,0 +1,69 @@
...
+
+import net.fabricmc.fabric.api.registry.VillagerPlantableRegistry;
+
+@Mixin(FarmerVillagerTask.class)
+public class FarmerVillagerTaskMixin {
+	@Nullable
+	@Shadow private BlockPos currentTarget;
+
+	@Inject(method = "keepRunning", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/passive/VillagerEntity;getInventory()Lnet/minecraft/inventory/SimpleInventory;"))
+	private void fabric_useRegistry(ServerWorld serverWorld, VillagerEntity villagerEntity, long l, CallbackInfo ci) {
```
Technici4n: I'm not too fond of this mixin, it is essentially deleting vanilla behavior and replacing it. This might break mods that mixin somewhere in these checks, and most importantly it might break vanilla behavior when updating to a new version. I would prefer a modify variable hook after the if chain, if possible.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,111 @@
...
+
+import net.fabricmc.fabric.impl.content.registry.util.ImmutableCollectionUtils;
+import net.fabricmc.fabric.mixin.content.registry.VillagerEntityAccessor;
+import net.fabricmc.fabric.mixin.content.registry.FarmerWorkTaskAccessor;
+import net.fabricmc.fabric.mixin.content.registry.GiveGiftsToHeroTaskAccessor;
+
+/**
+ * Registries for modifying villager interactions that
+ * villagers have with the world.
+ * The registries are:
```
Technici4n: Dangling comment?

Technici4n (changes requested):
A few comments, the rest looks good.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
aws404: Whoops, fixed.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/FarmerVillagerTaskMixin.java
aws404: After taking another look, I have a new ModifyVariable mixin on the read of the if statement, which should be much safter.
It also leaves the vanilla crops untouched, so they are no longer included in the registry.
Let me know what you think of this solution.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
aws404: Done!

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	/**
+	 * Get all currently registered seed items.
+	 * @return all currently registered seed items.
+	 */
+	public static Set<Item> getItems() {
+		return PLANTABLES.keySet();
+	}
+}
```
Technici4n: Does this make sense to expose given that it doesn't contain the vanilla ones? I think it's useful, so maybe we should have the vanilla ones in the registry for consistency purposes?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public static BlockState getPlantState(ItemConvertible item) {
+		return PLANTABLES.get(item.asItem());
+	}
+
+	/**
+	 * Get all currently registered seed items.
+	 * @return all currently registered seed items.
+	 */
+	public static Set<Item> getItems() {
```
Technici4n, line 110: Should probably be wrapped in `Collections.unmodifiableSet`.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Registers an item to be collectable (picked up from item entity)
+	 * by any profession villagers.
+	 *
+	 * @param item the item to register
+	 */
+	public static void registerCollectable(ItemConvertible item) {
+		Objects.requireNonNull(item, "Item cannot be null!");
```
Technici4n: Maybe check `item.asItem()` instead? It will catch both `item` and `item.asItem()` being null.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,110 @@
...
+		getCollectableRegistry().add(item.asItem());
+	}
+
+	/**
+	 * Registers an item to be use in a composter by farmer villagers.
+	 * @param item the item to register
+	 */
+	public static void registerCompostable(ItemConvertible item) {
+		Objects.requireNonNull(item, "Item cannot be null!");
```
Technici4n: Maybe check `item.asItem()` instead? It will catch both `item` and `item.asItem()` being null.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,110 @@
...
+		getCompostableRegistry().add(item.asItem());
+	}
+
+	/**
+	 * Registers an item to be edible by villagers.
+	 * @param item      the item to register
+	 * @param foodValue the amount of breeding power the item has (1 = normal food item, 4 = bread)
+	 */
+	public static void registerFood(ItemConvertible item, int foodValue) {
+		Objects.requireNonNull(item, "Item cannot be null!");
```
Technici4n: Maybe check `item.asItem()` instead? It will catch both `item` and `item.asItem()` being null.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerInteractionRegistries.java
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+
+	/**
+	 * Registers a hero of the village gifts loot table to a profession.
+	 * @param profession the profession to modify
+	 * @param lootTable  the loot table to associate with the profession
+	 */
+	public static void registerGiftLootTable(VillagerProfession profession, Identifier lootTable) {
+		Objects.requireNonNull(profession, "Profession cannot be null!");
```
Technici4n, line 90: Maybe check for null identifier?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Register an item with an associated to be plantable by farmer villagers.
+	 * For the crop to be harvestable, the block should extend CropBlock, so the
+	 * farmer can test the {@link CropBlock#isMature(BlockState)} method.
+	 * @param item       the seed item
+	 * @param plantState the state that will be planted
+	 */
+	public static void register(ItemConvertible item, BlockState plantState) {
+		PLANTABLES.put(item.asItem(), plantState);
```
Technici4n, line 79: Maybe check for null here?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+		return PLANTABLES.containsKey(item.asItem());
+	}
+
+	/**
+	 * Get the state that is associated with the provided seed item.
+	 * @param item the seed item
+	 * @return the state associated with the seed item
+	 */
+	public static BlockState getPlantState(ItemConvertible item) {
+		return PLANTABLES.get(item.asItem());
```
Technici4n, line 103: Could also use a null check.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/VillagerPlantableRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	/**
+	 * Tests if the item is a registered seed item.
+	 * @param item the item to test
+	 * @return true if the item is registered as a seed
+	 */
+	public static boolean contains(ItemConvertible item) {
+		return PLANTABLES.containsKey(item.asItem());
```
Technici4n, line 93: Could also use a null check.

Technici4n (approved):
A few rather minor comments, and it will be good for me!

Technici4n (approved):

deirn (approved):

Player3324 (approved):

modmuss50 (approved):

PR #2200, opened 2022-05-07
Add ValueToRefrence Function paramter to FabricRegistryBuilder methods

Dragon-Seeker:


apple502j:
This should not be added, because the use of the parameter is deprecated. While the parameter itself in `Registry` methods aren't annotated as such, the methods passed to that parameter, `getRegistryEntry`, are all marked as `@Deprecated`. I believe this is not the "this will be kept, just don't use" type of `@Deprecated` (in contrast, say, `AbstractBlock` methods or `ChunkRegion#toServerWorld`) - all other registries create the entry via the registry, and I believe they exist only because of compatibility. (Also note that this is used by game events, fluid, blocks, items, and entity types, the pre-1.18.2 taggable stuff.)

PR #2217, opened 2022-05-14
Fix wrong mixin in client commands

### LINKED ISSUE #2215: Client Commands no longer work since 22w19a
When typing out a client command in game you'll still receive command suggestions, however when sending the command it'll tell you it's an unknown or incomplete command. To me it looks like the client doesn't intercept the client commands and it gets send to the (integrated) server instead.

apple502j:
Resolves #2215 

`ClientCommandInternals` was changed to take unprefixed command messages (same as `ClientPlayerEntity#sendCommand`). This has some user-visible (although minor) change; the profiler and the debug logs now log unprefixed command.

Tested, works.


Juuxel:
I think it's just better to remove the `/` check from `CCI.executeCommand` instead since that isn't needed anymore.

Technici4n (approved):
Minimal changes, and the author says it works. 😄

Juuxel (approved):

modmuss50 (approved):

modmuss50:
Tested, works great now. Thanks.

Issue #1147, opened 2020-11-03
Discussion: Chat Event


zml2008:
Following up from a discussion this evening about chat events, I'd like to write up what I do with chat in my chat plugin, [featherchat](https://gitlab.com/stellardrift/featherchat) to help guide development of requirements. These steps are roughly in their current order of operation.

- **Chat filter** receive plain chat string, check if it contains any forbidden content that the player does not have permission to override, and cancel the message. A notification is sent to the original sender.
- **Formatting** if the chatter has permission to send formatted messages, parse the original chat message as Text XML, returning a `Text` instance.
- **Layout** Process a template, with message and player name among the parameters, returning a text XML string that gets parsed into a `Text`. This template, by default evaluates to `<{{ name }}> {{ message }}`, but can be customized by the user.
- **Channels** Take the final, formatted messages from both chat and `/me`, and choose the receivers based on an active chat channel. The channel may apply additional styling to the message.
  FeatherChat supports several types of channels by default: global, to people with a permission, to a defined list of members, and to people within a certain distance of the sender.

I'd appreciate feedback from anyone else who has projects that interact with chat! 
/cc @CoolMineman @i509VCB

Technici4n:
@apple502j just saw this, does #2220 cover this stuff?

apple502j:
@Technici4n Partially:

- "Chat filter" should be done using 1.16.4+ `text-filtering-config`. Exposing this is currently not done but I might add in future versions.
- "Formatting" is the upcoming message decorator API.
- "Layout" is the message type dynamic registry in vanilla.
- "Channel" is not done, but should not be hard.

ickerio:
Some sort of implementation like channels where you can specify who a message should be sent to would be awesome

Technici4n:
This should be supported to a sufficient extent by the message API.

PR #2220, opened 2022-05-15
Fabric Message API v1 (experimental)

### LINKED ISSUE #26: Chat API related thing
I need the following chat related things:

- [ ] player sent chat message (Mixin: `Lnet/minecraft/server/network/ServerPlayNetworkHandler;onChatMessage(Lnet/minecraft/server/network/packet/ChatMessageServerPacket;)V`)
- [ ] a way to detect `/say` usage

sending chat messages works via getting a instance of `MinecraftDedicatedServer` which gives the `DedicatedPlayerManager`

but would it be possible to get that server instance from FabricAPI too ?
or maybe through the loader.. it is getting constructed before the mods are loaded
so it could be passed in `onInitializeServer`

### LINKED ISSUE #971: Player Message Event
We need a Player Message Event

apple502j:
# Fabric Message API v1
## Resolves
Resolves #26
Resolves #971
Implements some of #1147

## What is this?
In 22w19a, Mojang introduced a change for modders specifically: "chat decorators". This is already in a state where a mod can use it directly with just one simple mixin. But here is the problem - I wrote "a mod", not "mods". Cross-compatibility will be a huge pain without a central mod providing a way for multiple mods to register chat decorators. This provides exactly that.

When making chat APIs with compatibility in mind, execution order becomes a issue. This PR tries to partially solve it by using the "event phases" in Event API. The API provides two phases - content and styling. The default phase sits in between. I will not try to address mod compatibilities between content-modifying mods, since that'd be a neverending process.

This also adds some events called when chat messages are sent, because it is not easy to get the right injection point.

## What this is NOT
- This **does not provide client-side functionality** (e.g. `ClientChatReceiveEvent`, exposing `ClientChatListener` registering, etc), I'm currently waiting for #2179 
- The API **cannot be used for proper censoring** (unless the server owner is fine with disabling chat signing or users being able to see censored content before sending).
- ~~The API **cannot block messages**. This needs a different callback. Returning `null` in chat decorator will crash the game (and the API provided by this intentionally throws earlier), and returning `Text.empty()` will probably cause an empty text to be sent (untested).~~ Blocking callback added.
- ~~The API **is not blocking-safe**. The decorator is called in the server thread.~~ Can execute asyncly since 1.19-pre1

## Testing
**Fully tested** on dedicated server with MSA, chat preview, and Only Show Secure Chat.

The testmod contains five manual tests:

- Basic (content): messages containing `tater` will have ` :tiny_potato:` appended
- Basic (styling): messages from creative-mode players will be colored in orange
- Logging for sent chat messages to check that the events trigger
- Blocking chat and command messages: messages containing `sadtater` are blocked
- Blocking game messages: death messages are no longer sent


modmuss50:
Handful of checkstyle issues need resolving.

apple502j:
I will remove the caching stuff because it's so hard to make it work right in all situations.

apple502j:
This is basically done. Other things like blocking or client stuff can go later. :winktato:

fabric-chat-api-v1/src/main/java/net/fabricmc/fabric/api/chat/v1/ChatDecoratorEvent.java
```diff
@@ -0,0 +1,101 @@
...
+ * message because it was improperly signed. The most straightforward way of doing this is by
+ * making sure the decorator returns the same result for a given message. Another way to solve this
+ * issue is to cache the decorated message.
+ *
+ * <p>Example of registering a content phase chat decorator:
+ *
+ * <pre><code>
+ * ChatDecoratorEvent.EVENT.register(ChatDecoratorEvent.CONTENT_PHASE, (sender, message) -> {
+ *     // Add smiley face. Has to shallowCopy() to get a MutableText with siblings and styles.
+ *     return message.shallowCopy().append(" :)");
```
modmuss50: Might be best to update these to use your new names from: https://github.com/FabricMC/yarn/pull/3172/

fabric-chat-api-v1/src/main/java/net/fabricmc/fabric/api/chat/v1/ChatDecoratorEvent.java
```diff
@@ -0,0 +1,101 @@
...
+			} else {
+				future = future.thenCompose((decorated) -> decorator.decorate(sender, decorated).thenApply(ChatDecoratorEvent::validateDecoratorResult));
+			}
+		}
+
+		return future == null ? CompletableFuture.completedFuture(message) : future;
+	}, CONTENT_PHASE, Event.DEFAULT_PHASE, STYLING_PHASE);
+
+	private static <T extends Text> T validateDecoratorResult(T value) {
+		return Objects.requireNonNull(value, "chat decorator must not return null");
```
modmuss50: Is it worth including the ChatDecorator classname in this? I can see it beign a slight pain to figure out what is causing it?

fabric-chat-api-v1/src/main/java/net/fabricmc/fabric/api/chat/v1/ServerChatEvents.java
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public static final Event<SendCommandMessage> SEND_COMMAND_MESSAGE = EventFactory.createArrayBacked(SendCommandMessage.class, handlers -> (message, source, typeKey) -> {
+		for (SendCommandMessage handler : handlers) {
+			handler.onSendCommandMessage(message, source, typeKey);
+		}
+	});
+
+	@FunctionalInterface
+	public interface SendChatMessage {
+		void onSendChatMessage(Message<SignedChatMessage> message, ServerPlayerEntity sender, RegistryKey<MessageType> typeKey);
```
modmuss50: Do we want a way to be able to cancel chat messages? I forget if there was a reason for this to not be a thing?

modmuss50 (approved):

liach:
I recommend calling this message event api instead, given we call them message in yarn and this only provide events

apple502j:
@liach Perhaps. In that case the module is `fabric-message-events`?

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ChatDecoratorEvent.java
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * The content phase of the event, passed when registering a chat decorator. Use this when
+	 * the chat decorator modifies the text content of the message.
+	 */
+	public static final Identifier CONTENT_PHASE = new Identifier("fabric-message-api-v1", "content");
+	/**
+	 * The styling phase of the event, passed when registering a chat decorator. Use this when
+	 * the chat decorator only modifies the styling of the message with the text intact.
+	 */
+	public static final Identifier STYLING_PHASE = new Identifier("fabric-message-api-v1", "styling");
```
Player3324: IMO we should use just fabric for the namespace in these, uniqueness requirements are low and makes upgrading easier

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ChatDecoratorEvent.java
```diff
@@ -0,0 +1,111 @@
...
+ *     // Apply orange color to messages sent by server operators
+ *     if (sender != null && sender.server.getPlayerManager().isOperator(sender.getGameProfile())) {
+ *         return CompletableFuture.completedFuture(
+ *             message.copy().styled(style -> style.withColor(0xFFA500)));
+ *     }
+ *     return CompletableFuture.completedFuture(message);
+ * });
+ * </code></pre>
+ */
+public final class ChatDecoratorEvent {
```
Player3324: Since this isn't part of ServerChatEvents and has no similarly identifying name it's implied that it also runs on the client

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatEvents.java
```diff
@@ -0,0 +1,174 @@
...
+
+public final class ServerChatEvents {
+	private ServerChatEvents() {
+	}
+
+	/**
+	 * An event that is called when a player sends a chat message to all players, either via the
+	 * chat hud or the {@code /say} command. Mods can use this event to block the message.
+	 */
+	public static final Event<AllowChatMessage> ALLOW_CHAT_MESSAGE = EventFactory.createArrayBacked(AllowChatMessage.class, handlers -> (message, sender, typeKey) -> {
```
Player3324: It is unclear when this fires, does it act on the inbound side from player packets or just before broadcasting such that it includes mod generated/replayed message? (some concern applies to other events)

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatEvents.java
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		return true;
+	});
+
+	/**
+	 * An event that is called when a command sends a message to all players. Note that if a player
+	 * executed the command, {@link #ALLOW_CHAT_MESSAGE} will also be triggered. To check whether the
+	 * player executed the command, use {@code source.getPlayer() != null}. Mods can use this event
+	 * to block the message.
```
Player3324: this could use a little more info about what "command sends a message" is, maybe an example?

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatEvents.java
```diff
@@ -0,0 +1,174 @@
...
+		for (AllowCommandMessage handler : handlers) {
+			if (!handler.allowCommandMessage(message, source, typeKey)) return false;
+		}
+
+		return true;
+	});
+
+	/**
+	 * An event that is called when a player sends a chat message to all players, either via the
+	 * chat hud or the {@code /say} command.
```
Player3324: chat-hut is unclear, makes it sound like the server gui has something, while it probably refers to chat messages received from the client as typically entered through the chat hud/gui.

send<->allow relation should be documented on both.

not including the `SEND_` prefix in these names could be better, the current ones with the verb sound a lot like something wsill be actively triggered vs getting a passive notification while not adding much.

Player3324 (commented):
IMO the API looks very nice, I have added a bunch of stylistic and comment related comments.

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ChatDecoratorEvent.java
apple502j: Guess I can rename to `ServerChatDecoratorEvent`?

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatEvents.java
apple502j: It is fired just before sending, so this does include the mod-generated message. @sfPlayer1 Should this be renamed or is a javadoc fine?

Player3324: javadoc may be fine for this, but the "send" terminology is a especially misleading for the events below - the client sends and the server sends.. could use "broadcast" instead?

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ChatDecoratorEvent.java
Player3324: yes, maybe Decorate even (verb)?

apple502j: This just exposes `ChatDecorator`, I think `ServerChatDecoratorEvent` is fine.

Player3324 (approved):

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatEvents.java
```diff
@@ -0,0 +1,235 @@
...
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.server.filter.FilteredMessage;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.text.Text;
+import net.minecraft.util.registry.RegistryKey;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ServerChatEvents {
```
liach: `ServerMessageEvents`

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerChatDecoratorEvent.java
```diff
@@ -0,0 +1,111 @@
...
+ *     // Apply orange color to messages sent by server operators
+ *     if (sender != null && sender.server.getPlayerManager().isOperator(sender.getGameProfile())) {
+ *         return CompletableFuture.completedFuture(
+ *             message.copy().styled(style -> style.withColor(0xFFA500)));
+ *     }
+ *     return CompletableFuture.completedFuture(message);
+ * });
+ * </code></pre>
+ */
+public final class ServerChatDecoratorEvent {
```
liach: @apple502j Mind rename this to ServerMessageDecoratorEvent?

apple502j: Well, assuming you're fine with that name, 👍🏼

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageDecoratorEvent.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * <p>This event uses phases to provide better mod compatibilities between mods that add custom
+ * content and styling. Message decorators with the styling phase will always apply after the ones
+ * with the content phase. When registering the message decorator, it is recommended to choose one
+ * of the phases from this interface and pass that to the {@link Event#register(Identifier, Object)}
+ * function. If not given, the message decorator will run in the default phase, which is between
+ * the content phase and the styling phase.
+ *
+ * <p>When implementing a message decorator, it is <strong>very important that the decorator be
+ * idempotent; i.e. return the same text when given the same text (and sender)</strong> -
```
Technici4n: I am not familiar with the vanilla code, but if a function $f$ is idempotent it means that $f(f(x)) = f(x)\  \forall x$, is that really what you meant?

apple502j: "Idempotent" has a slightly different definition in computer science.

Technici4n: https://en.wikipedia.org/wiki/Idempotence

Are you sure? It seems weird to expect that message decorators would run multiple times and that we should be careful about it. To me at least.

apple502j: Hmm, but here "pure" is definitely the word to choose.

PR #2233, opened 2022-05-20
Add Access Widener for EntityModelLayers functions

Mysterious-Dev:
This PR add somes AW for EntityModelLayers functions like `registerMain`, `register`, etc...

That can be usefull for custom entity model in 1.19.


Technici4n:
Two things:
- If you don't use transitive access wideners, they can't be used in dependent mods, which defeats the point.
- This should like go in the transitive access widener module instead.

Mysterious-Dev:
> Two things:
> 
>     * If you don't use transitive access wideners, they can't be used in dependent mods, which defeats the point.
> 
>     * This should like go in the transitive access widener module instead.

This is changed.

Technici4n:
You added them to the wrong file I believe, there should be a template instead. Make sure to add to the template and run the task to update the autogenerated file. This file is autogenerated and should not be edited manually.

Mysterious-Dev:
> You added them to the wrong file I believe, there should be a template instead. Make sure to add to the template and run the task to update the autogenerated file. This file is autogenerated and should not be edited manually.

Done, normaly ^^

Technici4n (approved):
I think this is good now, thanks!

deirn:
All `EntityModelLayers` methods create `EntityModelLayer` with `minecraft` namespace, I don't think we want modders to do that.

From what I can tell, registering `EntityModelLayer` is more involved than calling one of those methods, we need to modify the map `EntityModels#getModels` returns. Fabric API already provide [`EntityModelLayerRegistry`](https://github.com/FabricMC/fabric/blob/1.19/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/EntityModelLayerRegistry.java) for that.

PR #2248, opened 2022-05-23
Fix #2206: Client chunk cache grows unbounded

### LINKED ISSUE #2206: Clientside chunk cache grows unbounded
Tested on 1.18.2, Loader 0.14.4

Continuously generating chunks allows the `Set<WorldChunk> loadedChunks` field, accessed by the interface `net.fabricmc.fabric.impl.event.lifecycle.LoadedChunksCache` (in the chart as `lcc`) and mixed into `ClientWorld` through `net.fabricmc.fabric.mixin.event.lifecycle.client.ClientWorldMixin`, to grow unbounded on the client. On the server chunks are correctly removed via the `LoadedChunksCache` interface method `fabric_markUnloaded`, but this does not seem to happen on the client. Over time this completely saturates the heap. 

_X value on all charts is time in ticks. Chunks are continuously being generated by automatically teleporting the player periodically._

### Memory heap usage vs vanilla and LCC cache size on client and server
image
image

### Memory usage vs client LCC cache size after manually periodically calling `clear` on the set
image

### Memory usage vs client LCC cache size with a more aggressive manual set clearing
image

Technici4n:
Port of #1820 to 1.18.2. I confirmed that the leak originally reported in #1819 still occurred with the current `1.18.2` branch (before applying the fix), and that it doesn't happen anymore after the cherry-pick and the fixup commit.

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java
@@ -33,6 +33,7 @@ import net.minecraft.client.world.ClientWorld;
 import net.minecraft.nbt.NbtCompound;
 import net.minecraft.network.PacketByteBuf;
 import net.minecraft.world.chunk.WorldChunk;
+import net.minecraft.util.math.ChunkPos;
 
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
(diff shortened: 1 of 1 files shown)


Player3324 (approved):

liach:
Will we port to 1.19 too?

Technici4n:
> Will we port to 1.19 too?

Imo we should, otherwise we'll have the same report a third time in a few months. 😄

modmuss50 (approved):

jlin53882:
Will we port to 1.16.5 forge too? It's really important to me, hopefully
This version also has this problem。

PR #2249, opened 2022-05-23
Handle teleporting within the same dimension better.

### LINKED ISSUE #2238: Fabric Dimensions teleportation despawns non-player entities
When using `FabricDimensions#teleport` with a non-player entity, all of the entity's data is copied upon teleportation, including its UUID.
The issue with this is that the original entity still exists at the time of teleportation (and UUID copying), resulting in the entity (or technically both entities) disappearing/despawning completely.
Some solutions would be:
1. The new entity should get a new UUID
2. The old entity should be completely removed before copying over its UUID to the new entity

Tested with the following versions:
**Minecraft:** 1.18.2
**Fabric API:** 0.53.0
**Fabric Dimensions:** 2.1.17

### LINKED ISSUE #2239: Fabric Dimensions slight desync on teleportation on multiplayer
When teleporting a player with `FabricDimensions#teleport` on multiplayer, there's a slight desync on the client side.
For the player, they initially "teleport" to spawn with no items, xp, buffs, etc (basically a "fresh" new player instance), and then get teleported to the specified location a second later with all the items/xp/etc. they originally had.

My guess is that the teleport code runs both on client and server side for some reason, and that on the client side, there's a new player instance with default, non-copied data, and after teleporting client-side, the server syncs the player with a proper teleportation packet.

Tested with the following versions:
**Minecraft:** 1.18.2
**Fabric API:** 0.53.0
**Fabric Dimensions:** 2.1.17

modmuss50:
Fixes #2239 and Fixes #2238

This API was mainly designed to teleport entities between diffrent dimensions, however I think it makes sense to handle the same dimension case better.

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
@@ -19,6 +19,7 @@ package net.fabricmc.fabric.impl.dimension;
 import com.google.common.base.Preconditions;
 
 import net.minecraft.entity.Entity;
+import net.minecraft.server.network.ServerPlayerEntity;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.world.TeleportTarget;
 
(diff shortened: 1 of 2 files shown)


Juuxel (approved):

LordDeatHunter:
Does this account for https://bugs.mojang.com/browse/MC-124177 ?
IIRC, I've seen it same-dimension, tho I might be misremembering.

modmuss50:
This shouldnt affect that same-dimension. And doesnt fix it across dimensions.

PR #2250, opened 2022-05-23
A few datagen fixes

### LINKED ISSUE #2228: Datagen API should be included in prod jars
In MI we're extending `BlockSettings` with functions to specify a `LootTableGenerator` etc, and we can't rely on Fabric's TAWs otherwise it crashes in prod.

### LINKED ISSUE #2240: Datagen: when strict validation is not enabled, item models are not generated from block models
Vanilla however automatically generates an item model, unless `BlockModelGenerators#excludeFromSimpleItemModelGeneration` is called.

It would be nice for fabric to have the same behavior, but it might be a breaking change, or it might require opt-in (which is annoying).

### LINKED ISSUE #2241: Datagen: there should be a way to exclude blocks for strict loot table validation
It doesn't make sense for some blocks like `modern_industrialization:pipe` to have a loot table, so I have to choose between the following 3 options atm:
- No strict validation. But then I lose automatic item model generation from block models (see #2240).
- Generate an empty loot table file. A bit annoying, generates a useless file that could mislead users who look at the jar.
- Override `FabricBlockLootTableProvider#accept` to remove the strict validation check. Requires an extra access widener on `BlockLootTableGenerator.lootTables`.

Technici4n:
1. Closes #2228.
2. Closes #2241.
3. Fixes #2240. Item models are now always automatically generated, but only for blocks whose blockstate jsons are generated. Calling `excludeFromSimpleItemModelGeneration(block)` on the `BlockStateModelGenerator` will opt out of this behavior (regular vanilla datagen feature). **Might require a major version bump since it's a breaking change for mods relying on the previous (wrong) behavior.**

Testing I did, for each commit respectively:
1. I just checked that the jar was indeed included in the built fatjar.
2. Expanded the testmod to check this, work as expected.
3. I tested various combinations of strictness, not registering some blocks, and explicitly disabling the item models with `excludeFromSimpleItemModelGeneration`, everything works as expected now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -375,7 +375,6 @@ sourceSets {
 // These modules are not included in the fat jar, maven will resolve them via the pom.
 def devOnlyModules = [
 		"fabric-gametest-api-v1",
-		"fabric-data-generation-api-v1"
 ]
 
 dependencies {
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java
@@ -16,12 +16,14 @@
 
 package net.fabricmc.fabric.api.datagen.v1.provider;
 
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
 
 import com.google.common.collect.Sets;
 
+import net.minecraft.block.Block;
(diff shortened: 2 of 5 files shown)


modmuss50 (approved):

modmuss50:
Seems fine, will major bump it. But not give it a new module as it was technically a bug,  dev only and small logic change. The breakage is as minimal as its ever going to be.

Technici4n:
Might be nice to rebase instead of squash these commits, since I think only the first two should be backported to 1.18.2. The datagen "breaking change" is a bit annoying this late in the 1.18.2 cycle imo.

Player3324 (approved):

PR #2251, opened 2022-05-23
Fix #2152: End server world tick too early and conditional

### LINKED ISSUE #2152: end world tick event too early and conditional
The mixin seems to be unfortunately placed in 1.18 and potentially other versions, causing the the event to not run if there are no players (contrary to the start world tick). It also runs too early, before entityManager.tick.

Technici4n:
Fix #2152

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientWorldMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientWorldMixin.java
@@ -26,19 +26,17 @@ import net.minecraft.client.world.ClientWorld;
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
 import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientTickEvents;
-import net.fabricmc.fabric.mixin.event.lifecycle.WorldMixin;
 
 @Environment(EnvType.CLIENT)
 @Mixin(ClientWorld.class)
-public abstract class ClientWorldMixin extends WorldMixin {
-	// We override our injection on the clientworld so only the client world's tick invocations will run
-	@Override
-	protected void tickWorldAfterBlockEntities(CallbackInfo ci) {
-		ClientTickEvents.END_WORLD_TICK.invoker().onEndTick((ClientWorld) (Object) this);
-	}
-
+public abstract class ClientWorldMixin {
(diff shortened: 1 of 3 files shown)


fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientWorldMixin.java
```diff
@@ -26,19 +26,17 @@
...
-		ClientTickEvents.END_WORLD_TICK.invoker().onEndTick((ClientWorld) (Object) this);
-	}
-
+public abstract class ClientWorldMixin {
 	@Inject(method = "tickEntities", at = @At("HEAD"))
 	private void startWorldTick(CallbackInfo ci) {
 		ClientTickEvents.START_WORLD_TICK.invoker().onStartTick((ClientWorld) (Object) this);
 	}
+
+	@Inject(method = "tickEntities", at = @At("TAIL"))
```
modmuss50, line 38: Very similar to the 1.19 changes I had to make becuase of splitting. Seems to only differ by this injection target now: https://github.com/FabricMC/fabric/blob/1.19/fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientWorldMixin.java#L38 👍

modmuss50 (approved):

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientWorldMixin.java
Technici4n: Isn't it the exact same inject, just placed differently in the source file and with a different method name?

Technici4n: Seems like you fixed the issue in 1.19 for the server event, I didn't notice! 😮

modmuss50: If I fixed it it was an accident 😂

modmuss50: Ah yeah I did accidentally fix it 👍 Was confused as the method order is swapped.

PR #2261, opened 2022-05-27
Load mod-provided data packs and DRM entries

### LINKED ISSUE #2225: Fabric mods using json for their worldgen isn't loading in the json files on 1.19 snapshots.
The rundown is:
I recently decided to port my mod into the snapshots, and my mod so happens to use json for it's world generation (like structures, features, biome settings, etc). But whenever launching a world, it doesn't load up those json files because the biomes are featureless, locating the structure doesn't work. I'm not really sure if this is either a Minecraft thing or a Fabric thing, but it's worth reporting.

Additional Info:
- On the world creation screen, you can work around the issue by going into the datapack screen and closing out of it so the game verifies it, with that, the world will now load the worldgen folder just fine.
- Worlds loaded from 1.18 will still continue to work just fine

### LINKED ISSUE #2259: Crash at the world creation screen with datapack content added through the Biome API
The game crashes when trying to create a world, if a mod with datapack content that is added through the Biome API is installed.
As a clear example, Promenade features Nether and End biomes and features that are now added by the mod's datapack (see commit https://github.com/DawnTeamMC/Promenade/commit/0e14663a8093a43ce420ed3f0958c8d13bbca925 and this [crash log](https://bytebin.lucko.me/SVmrvc6iGj)) and then added to vanilla generation through the Biome API, which is part of the Fabric API.

In vanilla Minecraft, when you try to create a world, the vanilla builtin worldgen gets loaded, and then opens the screen with the loaded generic content. However the Biome API lets us add to the vanilla generation by adding `RegistryKey`s, but actually requires these keys to represent built-in content so that it gets loaded on this specific screen. This constraint should not exist, and as a fix the Fabric API must tell the game to load the mod datapacks at the same time as builtin Minecraft worldgen content on this screen.

apple502j:
Fixes #2225
Fixes #2259

There were 2 issues in the existing `CreateWorldScreenMixin` mixin, specifically with `create(MinecraftClient, Screen)` method:

- There were two references to `DataPackSettings#SAFE_MODE`. The first one was used to load the DRM, and the second one is passed to the constructor. Previously, only the second one was replaced to include Fabric-provided data packs.
- There were no attempts to load dynamic registry from the provided data packs.

This fixes both issues, plus a code change to try-with-resources `createResourcePack`.

The first issue also affected dedicated servers, and both for test servers, which should be fixed as well.

Biome API testmod now includes a test for adding a placed feature (pile of red concrete) defined in a data pack. Previously this caused a crash.

### DIFF
--- a/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/CreateWorldScreenMixin.java
+++ b/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/CreateWorldScreenMixin.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import com.mojang.datafixers.util.Pair;
 import com.mojang.serialization.JsonOps;
+import org.spongepowered.asm.mixin.Dynamic;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.Unique;
@@ -51,9 +52,6 @@ import net.fabricmc.fabric.mixin.resource.loader.ResourcePackManagerAccessor;
 public class CreateWorldScreenMixin {
 	@Unique
 	private static DataPackSettings defaultDataPackSettings;
-
-	@Unique
-	private static ResourceManager loadingDataPackManager;
 
 	@Shadow
 	private ResourcePackManager packManager;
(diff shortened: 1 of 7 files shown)


fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/CreateWorldScreenMixin.java
```diff
@@ -53,35 +63,46 @@ private static ResourcePackManager onCreateResManagerInit(ResourcePackManager ma
...
-		return new DataPackSettings(enabled, disabled);
+		return (defaultDataPackSettings = new DataPackSettings(enabled, disabled));
+	}
+
+	@ModifyArg(method = "create(Lnet/minecraft/client/MinecraftClient;Lnet/minecraft/client/gui/screen/Screen;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screen/world/CreateWorldScreen;<init>(Lnet/minecraft/client/gui/screen/Screen;Lnet/minecraft/resource/DataPackSettings;Lnet/minecraft/client/gui/screen/world/MoreOptionsDialog;)V"), index = 1)
+	private static DataPackSettings useReplacedDefaultSettings(DataPackSettings dataPackSettings) {
+		return defaultDataPackSettings;
+	}
+
+	@Redirect(method = "method_41854", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/registry/DynamicRegistryManager$Mutable;toImmutable()Lnet/minecraft/util/registry/DynamicRegistryManager$Immutable;"), remap = false)
```
maityyy: Remove `remap = false` to avoid crash outside dev env

[crash-2022-05-27_22.41.42-client.txt](https://github.com/FabricMC/fabric/files/8790471/crash-2022-05-27_22.41.42-client.txt)

maityyy (changes requested):

apple502j:
Seems like this still crashes in dedicated/gametest server. Need to fix this.

Juuxel (approved):

modmuss50 (approved):
Looks fine, I assume this is 1.19 only?

apple502j:
~~The code for dedicated servers need 1.18.2 backport. Client-side code I believe should be fine, but we need some testing. No backport is necessary for 1.18.1 or earlier, as ServerBugfixMixin handles these cases.~~

No backport needed.

maityyy (approved):

PR #2264, opened 2022-05-29
Change client command api to use registration event

jpenilla:
Allows usage of CommandRegistryAccess in client command arguments more easily


fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
```diff
@@ -16,22 +16,25 @@
 
 package net.fabricmc.fabric.api.client.command.v2;
 
+import javax.annotation.Nullable;
```
apple502j: I believe we use the one from jetbrains annotation

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
```diff
@@ -16,14 +16,15 @@
...
 
-import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.DISPATCHER;
 import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.argument;
 import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.literal;
 
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nullable;
```
apple502j: same here

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
```diff
@@ -55,6 +56,15 @@ public final class ClientCommandInternals {
 	private static final Logger LOGGER = LoggerFactory.getLogger(ClientCommandInternals.class);
 	private static final String API_COMMAND_NAME = "fabric-command-api-v2:client";
 	private static final String SHORT_API_COMMAND_NAME = "fcc";
+	private static CommandDispatcher<FabricClientCommandSource> DISPATCHER;
```
apple502j: I don’t think we uppercase nonfinal fields

apple502j: also this should be marked as `@Nullable`

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
```diff
@@ -55,6 +56,15 @@ public final class ClientCommandInternals {
...
+
+	public static void setActiveDispatcher(CommandDispatcher<FabricClientCommandSource> dispatcher) {
```
apple502j: This accepts null, right?

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
```diff
@@ -49,23 +52,31 @@
...
 @Environment(EnvType.CLIENT)
 public final class ClientCommandManager {
+	private ClientCommandManager() {
+	}
+
 	/**
-	 * The command dispatcher that handles client command registration and execution.
+	 * Gets the active command dispatcher that handles client command registration and execution.
+	 *
+	 * <p>May be null when not connected to a server.</p>
```
apple502j: If client commands do work in integrated environment, this is misleading, since you only “connect” to dedicated servers. “not in a world” might work.

apple502j (changes requested):

Player3324 (approved):

Juuxel (approved):
LGTM, this is pretty close to the original design of the client command API before I had to change it to a raw dispatcher 😄

Technici4n (approved):

modmuss50 (approved):
Not ideal to have breaking changes, but seems like its the best way forward.

modmuss50:
Putting this into last call for a little (not the whole week) as this needs to go in ASAP.

apple502j (approved):

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/ClientPlayNetworkHandlerMixin.java
```diff
@@ -41,6 +44,14 @@ abstract class ClientPlayNetworkHandlerMixin {
 	@Final
 	private ClientCommandSource commandSource;
 
+	@Inject(method = "onGameJoin", at = @At("HEAD"))
```
Earthcomputer, line 47: This causes race conditions as it occurs on both the netty thread and the main thread

haykam821: By the way, should this initialization be moved to when the server command tree packet is received? That would also allow the command registry access to be shared between server and client commands.

Issue #2108, opened 2022-04-05
FlammableBlockRegistry ignores tags unless `/reload`


SolidBlock-cn:
When adding a block tag to `FlammableBlockRegistry`, it is ignored, which means blocks in this tag cannot be flammable. If you `/reload`, it can be normal, even if you re-enter the world. But when you re-launch the game, it will happen again.

```
minecraft_version=1.18.1
yarn_mappings=1.18.1+build.22
loader_version=0.13.3
fabric_version=0.46.6+1.18
```

Technici4n:
Can't reproduce in 1.19. Please ensure that you are registering the tag early enough (i.e. in your mod initializer), and let me know if the issue persists.

Technici4n:
Looking at the code, it seems that it should be an issue though... I am a bit confused. 😅

Technici4n:
This is indeed an issue, I think I was unable to reproduce this on Fabric because of the experimental settings warning causing the resources to be loaded a second time... The simplest fix would require #2265 to be merged, so I could use the TagsLoaded event.

PR #2265, opened 2022-05-29
Add SyncDataPackContents and TagsLoaded lifecycle events

Technici4n:
This PR adds two useful lifecycle events.

1. `SyncDataPackContents`. Quite straightforward. I need this event for Modern Dynamics where I have a custom resource reload listener whose contents I want to sync with the client, to display in REI. A similar effect could be achieved using a connection event, and the EndDataPackReload event, however the latter runs too late if REI expects all data to have arrived before the tag and the recipe sync packets.
2. `TagsLoaded`. This event is useful when a tag-dependent cache needs to be invalidated because tags have changed. Not sure if the `DynamicRegistryManager` and the `isClient` parameters are useful, but I thought I might as well provide them. There were two places in Fabric API that I converted to use this event (`FuelRegistry` cache, and the `MiningLevelManagerImpl`). It also turns out that `FuelRegistry` didn't invalidate its cache when a server is started, which could theoretically be an issue when logging out of a singleplayer and then logging back in with a different datapack.


Technici4n:
Interesting prior art for `TagsLoaded`: (mojmap ofc) https://github.com/MinecraftForge/MinecraftForge/pull/8636, I just realized.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/CommonLifecycleEvents.java
```diff
@@ -0,0 +1,44 @@
...
+			callback.tagsLoaded(registries, isClient);
+		}
+	});
+
+	public interface TagsLoaded {
+		/**
+		 * @param registries Up-to-date registries from which the tags can be retrieved.
+		 * @param isClient True if the client just received a sync packet, false if the server just (re)loaded the tags.
+		 */
+		void tagsLoaded(DynamicRegistryManager registries, boolean isClient);
```
Juuxel: This should have a more method-y name. Also, `is` prefix for params goes against yarn a bit

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerLifecycleEvents.java
```diff
@@ -116,6 +129,41 @@ public interface ServerStopped {
...
+		 *     Packet<?> packet = ServerPlayNetworking.createS2CPacket(CHANNEL_NAME, buf);
+		 *
+		 *     // 2) Send to relevant player(s)
+		 *     if (player == null) {
+		 *         // Reload: send to all
+		 *         playerManager.sendToAll(packet);
+		 *     } else {
+		 *         // Player joined: send only to that player
+		 *         player.networkHandler.sendPacket(packet);
+		 *     }
```
Juuxel: This feels weird, like there's two events in one.

Technici4n: Alternatives:
- Having two events.
- Sending the event for each player, and needing to create a new packet every time.

I don't think the alternatives are particularly nicer than this, especially with the extensive javadoc explaining how to use the event.

Technici4n:
Since this is a prerequisite for a clean fix of #2108, I am marking this as medium priority.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerLifecycleEvents.java
Technici4n: Changed to sending the event for each player, the API shape is a lot better, and I think trying to not serialize the packet multiple times is premature optimization. If this proves to be a bottleneck, we can always introduce a specialized event later on.

Player3324 (approved):

modmuss50 (approved):

Juuxel (approved):

PR #2266, opened 2022-05-29
Fix errors not showing up in disconnect messages

Justsnoopy30:
Previously, the code would just send disconnect.genericReason which is empty when no argument is provided, causing blank disconnect messages without the error responsible.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
@@ -63,22 +63,12 @@ abstract class ClientConnectionMixin implements ChannelInfoHolder {
 		this.playChannels = Collections.newSetFromMap(new ConcurrentHashMap<>());
 	}
 
-	@Unique
-	private Throwable caughtException;
-
-	// Must be fully qualified due to mixin not working in production without it
-	@SuppressWarnings("UnnecessaryQualifiedMemberReference")
-	@Inject(method = "Lnet/minecraft/network/ClientConnection;exceptionCaught(Lio/netty/channel/ChannelHandlerContext;Ljava/lang/Throwable;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/network/ClientConnection;send(Lnet/minecraft/network/Packet;Lio/netty/util/concurrent/GenericFutureListener;)V"))
-	private void resendOnExceptionCaught$captureThrowable(ChannelHandlerContext context, Throwable ex, CallbackInfo ci) {
-		caughtException = ex;
-	}
-
 	// Must be fully qualified due to mixin not working in production without it
 	@SuppressWarnings("UnnecessaryQualifiedMemberReference")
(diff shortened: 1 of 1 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
```diff
@@ -63,16 +63,27 @@ private void initAddedFields(NetworkSide side, CallbackInfo ci) {
...
+	private void resendOnExceptionCaught$captureThrowable(ChannelHandlerContext context, Throwable ex, CallbackInfo ci) {
+		caughtException = ex;
+	}
+
 	// Must be fully qualified due to mixin not working in production without it
 	@SuppressWarnings("UnnecessaryQualifiedMemberReference")
 	@Redirect(method = "Lnet/minecraft/network/ClientConnection;exceptionCaught(Lio/netty/channel/ChannelHandlerContext;Ljava/lang/Throwable;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/network/ClientConnection;send(Lnet/minecraft/network/Packet;Lio/netty/util/concurrent/GenericFutureListener;)V"))
 	private void resendOnExceptionCaught(ClientConnection self, Packet<?> packet, GenericFutureListener<? extends Future<? super Void>> listener) {
 		PacketListener handler = this.packetListener;
+		Text disconnectMessage = new TranslatableText("disconnect.genericReason", "Internal Exception: " + caughtException);
```
deirn:
```suggestion
	private void resendOnExceptionCaught(ClientConnection self, Packet<?> packet, GenericFutureListener<? extends Future<? super Void>> listener, ChannelHandlerContext context, Throwable ex) {
		PacketListener handler = this.packetListener;
		Text disconnectMessage = new TranslatableText("disconnect.genericReason", "Internal Exception: " + ex);
```

The above inject is unnecessary as [`Redirect`](https://jenkins.liteloader.com/job/Mixin/javadoc/org/spongepowered/asm/mixin/injection/Redirect.html) can also capture the target method arguments.

deirn (approved):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
liach: Might want to use `ex.getMessage()` as well. Is it really safe to leak the stack trace to the connecting clients?

Technici4n: Agreed, it would likely be better to log the full trace, but only send the message to the client.

Player3324: No ex is fine (involves ex.toString which is just the class name + message) - some exceptions have no message and thus would just show as `null`, which is pretty bad.

Justsnoopy30: > The above inject is unnecessary as [`Redirect`](https://jenkins.liteloader.com/job/Mixin/javadoc/org/spongepowered/asm/mixin/injection/Redirect.html) can also capture the target method arguments.

Oh, I didn't realize.

liach (approved):

Player3324 (approved):

Player3324:
It'd be nice to log the exception with a higher level than vanilla's debug since that's rather unapproachable to server admins and mods can cause all sorts of issues that end up there. This may however require pre-filtering the typical pure network exceptions, which is well beyond this PR.

modmuss50 (approved):

PR #2268, opened 2022-05-30
Add job site and sandstone block/item convention tags

aws404:
Adds the following convention tags for both items/blocks:
* `c:villager_job_sites`
* `c:yellow_sandstone_blocks`
* `c:red_sandstone_blocks`
* `c:sandstone_blocks`
* `c:yellow_sandstone_slabs`
* `c:red_sandstone_slabs`
* `c:sandstone_slabs`
* `c:yellow_sandstone_stairs`
* `c:red_sandstone_stairs`
* `c:sandstone_stairs`


fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/ConventionalBlockTags.java
```diff
@@ -33,7 +33,21 @@ public class ConventionalBlockTags {
...
+	public static final TagKey<Block> RED_SANDSTONE_BLOCKS = register("red_sandstone_blocks");
+	public static final TagKey<Block> SANDSTONE_BLOCKS = register("sandstone_blocks");
+	public static final TagKey<Block> YELLOW_SANDSTONE_SLABS = register("yellow_sandstone_slabs");
+	public static final TagKey<Block> RED_SANDSTONE_SLABS = register("red_sandstone_slabs");
+	public static final TagKey<Block> SANDSTONE_SLABS = register("sandstone_slabs");
+	public static final TagKey<Block> YELLOW_SANDSTONE_STAIRS = register("yellow_sandstone_stairs");
+	public static final TagKey<Block> RED_SANDSTONE_STAIRS = register("red_sandstone_stairs");
+	public static final TagKey<Block> SANDSTONE_STAIRS = register("sandstone_stairs");
+
+	public static final TagKey<Block> JOB_SITE_BLOCKS = register("job_site_blocks");
```
Technici4n: Isn't there already `#minecraft:acquirable_job_site` for this?

aws404: `#minecraft:acquirable_job_site` is a POI tag, not a block tag. You could theoretically use it to get the blocks, however it's not very efficent

Technici4n: Ah, nevermind then. I think a better name would be `#c:villager_job_sites` to explain that it's about villagers. The `_blocks` suffix is unnecessary here imo.

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -112,4 +114,59 @@ private void generateGlassTags() {
...
+				.add(Blocks.CUT_SANDSTONE)
+				.add(Blocks.CHISELED_SANDSTONE)
+				.add(Blocks.SMOOTH_SANDSTONE);
+		getOrCreateTagBuilder(ConventionalBlockTags.RED_SANDSTONE_BLOCKS)
+				.add(Blocks.RED_SANDSTONE)
+				.add(Blocks.CUT_RED_SANDSTONE)
+				.add(Blocks.CHISELED_RED_SANDSTONE)
+				.add(Blocks.SMOOTH_RED_SANDSTONE);
+		getOrCreateTagBuilder(ConventionalBlockTags.SANDSTONE_BLOCKS)
+				.addOptionalTag(ConventionalBlockTags.YELLOW_SANDSTONE_BLOCKS)
```
Technici4n: Why optional btw, just curious?

Technici4n (approved):

haykam821:
I don't think regular sandstone should be referred to as yellow sandstone. Is there a use case for having a combined sandstone tag? Since sandstone and red sandstone are clearly different blocks that wouldn't make sense to be part of the same tag, I think that `c:yellow_sandstone_blocks` and similar tags should be renamed to `c:sandstone_blocks`.

Technici4n:
It is quite common, on Forge at least, to be able to use red sand in recipes that use sand. So would it make sense to be able to use both kinds of sandstone with a common tag? Not sure.

aws404:
> I don't think regular sandstone should be referred to as yellow sandstone. Is there a use case for having a combined sandstone tag? Since sandstone and red sandstone are clearly different blocks that wouldn't make sense to be part of the same tag, I think that `c:yellow_sandstone_blocks` and similar tags should be renamed to `c:sandstone_blocks`.

> It is quite common, on Forge at least, to be able to use red sand in recipes that use sand. So would it make sense to be able to use both kinds of sandstone with a common tag? Not sure.

This was along my trail of thought, happy to change it if the consensus is otherwise though.

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
aws404: I'm not 100% sure, I used optional because thats what everything else uses. I can't see a reason, but I assumed there was one.

dexman545:
As tags are a collection, there is ambiguity if `sandstone` means all sandstones or just the original sandstone. Unless the collection of all sandstones is given some demarcating feature, the original sandstone collections should be marked as `yellow`,  `normal`, or other.

haykam821:
Sandstone isn't referred to as yellow within vanilla Minecraft, so it shouldn't be referred to as yellow within tags.

I think that any tags containing only sandstone and not red sandstone should be prefixed with 'normal' or similar, or the sandstone and red sandstone combined tags should be prefixed with 'all' or similar.

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -112,4 +114,59 @@ private void generateGlassTags() {
...
+				.add(Blocks.FLETCHING_TABLE)
+				.add(Blocks.GRINDSTONE)
+				.add(Blocks.LECTERN)
+				.add(Blocks.LOOM)
+				.add(Blocks.SMITHING_TABLE)
+				.add(Blocks.SMOKER)
+				.add(Blocks.STONECUTTER);
+	}
+
+	private void generateSandstoneBlocksTag() {
```
Juuxel: Nitpick: the name should be plural (`...Tags`)

Juuxel (approved):

Player3324:
I'm not a huge fan of the normal prefix - it's only the color being "normal", but not the structure/naturalness. Uncolored or so might be better. The slabs and stairs lack the prefix while the blocks have it.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/ConventionalItemTags.java
```diff
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.api.tag.convention.v1;
 
+import net.minecraft.block.Block;
```
Player3324: This import is unused, checkstyle hates it ;)

Juuxel:
> The slabs and stairs lack the prefix while the blocks have it.

There are both "normal"/uncoloured/yellow and "all" (unprefixed) versions of the slab and stair tags, though?

Player3324:
Right, I missed that, so only the proper prefix choice question remains.

Technici4n:
@aws404 Sorry for the delay! Could you please change the naming back to yellow? (We had some discussion on discord... Normal doesn't really work, and we couldn't find a better alternative). Should be fine after that.

TiphaineLAURENT:
I don't see a reason to add a special these special sandstone tags. A sandstone stairs should have 3 tags : sandstone (material), stairs (type) and yellow (color) so it would work for recipes which accept sandstone or recipes that only accept yellow sandstone.

TLDR: We should use composite tag not special tag

LambdAurora:
Someone gonna has to explain me how in hell the base sandstone is yellow.

LambdAurora:
I want someone to look at me straight in the eyes and tell me that normal terracotta is actually brown terracotta and that the normal shulker box is actually purple shulker box. While there's already a specific dyed variant for those colors.

image

TiphaineLAURENT:
> Someone gonna has to explain me how in hell the base sandstone is yellow.

If we have to assign it a Minecraft color (https://minecraft.fandom.com/wiki/Dye), it's definitely yellow

LambdAurora:
> If we have to assign it a Minecraft color (https://minecraft.fandom.com/wiki/Dye), it's definitely yellow

It's called sand, not yellow_sand.

Technici4n:
Another update, this time final I hope, after some more discussion on Discord: please change to `uncolored` for the regular sandstone.

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -112,4 +114,59 @@ private void generateGlassTags() {
...
+				.addOptionalTag(ConventionalBlockTags.RED_SANDSTONE_SLABS);
+
+		getOrCreateTagBuilder(ConventionalBlockTags.NORMAL_SANDSTONE_STAIRS)
+				.add(Blocks.SANDSTONE_STAIRS)
+				.add(Blocks.SMOOTH_SANDSTONE_STAIRS);
+		getOrCreateTagBuilder(ConventionalBlockTags.RED_SANDSTONE_STAIRS)
+				.add(Blocks.RED_SANDSTONE_STAIRS)
+				.add(Blocks.SMOOTH_RED_SANDSTONE_STAIRS);
+		getOrCreateTagBuilder(ConventionalBlockTags.SANDSTONE_STAIRS)
+				.addOptionalTag(ConventionalBlockTags.NORMAL_SANDSTONE_SLABS)
```
ghost, line 169: I think this must be stairs, not slabs... 🤔

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/ItemTagGenerator.java
```diff
@@ -312,6 +314,61 @@ private void generateToolTags() {
...
+				.addOptionalTag(ConventionalItemTags.RED_SANDSTONE_SLABS);
+
+		getOrCreateTagBuilder(ConventionalItemTags.NORMAL_SANDSTONE_STAIRS)
+				.add(Items.SANDSTONE_STAIRS)
+				.add(Items.SMOOTH_SANDSTONE_STAIRS);
+		getOrCreateTagBuilder(ConventionalItemTags.RED_SANDSTONE_STAIRS)
+				.add(Items.RED_SANDSTONE_STAIRS)
+				.add(Items.SMOOTH_RED_SANDSTONE_STAIRS);
+		getOrCreateTagBuilder(ConventionalItemTags.SANDSTONE_STAIRS)
+				.addOptionalTag(ConventionalItemTags.NORMAL_SANDSTONE_SLABS)
```
ghost, line 368: I think this must be stairs, not slabs... 🤔

ghost (changes requested):
My feedback is that i agree to change the name to "uncolored" instead of normal, because "normal"  it too much generic and means anything (normal =  no danger; normal = no shape; normal = no food.... etc. this is an example 😅), "uncolored" refers only to colors and is the correct name.

Technici4n:
Replaced by #3051.

PR #2279, opened 2022-06-02
Sculk Sensor Frequency Registry

Shnupbups:
- Mixin'd to `SculkSensorBlock` to make its `FREQUENCIES` map modifiable.
- Created `SculkSensorFrequencyRegistry` with a dedicated method for registering these frequencies, with all necessary checks to avoid issues, and javadoc to assist in understanding.
- Added tests to the test mod, as well as a block that emits a test event, to ensure it all works.

Unfortunately, the mixin currently uses a Redirect. If this is undesired, feel free to suggest other ways of achieving the end result of being able to modify the frequencies map.


fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SculkSensorBlockMixin.java
```diff
@@ -0,0 +1,34 @@
...
+import org.spongepowered.asm.mixin.Dynamic;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import net.minecraft.block.SculkSensorBlock;
+
+@Mixin(SculkSensorBlock.class)
+public class SculkSensorBlockMixin {
+	@Dynamic("Call to Object2IntMaps.unmodifiable in initialization of FREQUENCIES")
```
Technici4n: Why dynamic?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SculkSensorFrequencyRegistry.java
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public static void register(GameEvent event, int frequency) {
+		if (frequency <= 0 || frequency >= 16) {
+			throw new IllegalArgumentException("Attempted to register Sculk Sensor frequency for event "+event.getId()+" with frequency "+frequency+". Sculk Sensor frequencies must be between 1 and 15 inclusive.");
+		}
+
+		int replaced = SculkSensorBlock.FREQUENCIES.getOrDefault(event, 0);
+		SculkSensorBlock.FREQUENCIES.put(event, frequency);
+
+		if (replaced != 0) {
```
Technici4n, line 55: Couldn't you check for putInt != 0 in one call?

Technici4n (approved):
Looks good to me!

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SculkSensorBlockMixin.java
Shnupbups: It's a decorator annotation provided by mixin specifically for providing context for mixins that aren't immediately obvious where they modify. Similar usage is found in this module's `OxidizableMixin` and `HoneycombMixin` when they inject into lambda bodies. I suppose `<clinit>` isn't altogether hard to imagine, but it's certainly less transparent than a mixin to a visible method, and so I felt it somewhat warranted.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SculkSensorFrequencyRegistry.java
Shnupbups: Genuinely forgot that the `put` method of maps returns old values if present, good shout. However I'll keep it separate from the `if` due to the variable then being used in the debug log.

fabric-content-registries-v0/src/main/resources/fabric-content-registries-v0.mixins.json
```diff
@@ -9,7 +9,8 @@
     "MixinAbstractFurnaceBlockEntity",
     "MixinFireBlock",
     "OxidizableMixin",
-    "ShovelItemAccessor"
+    "ShovelItemAccessor",
+    "SculkSensorBlockMixin"
```
Juuxel:
```suggestion
    "SculkSensorBlockMixin",
    "ShovelItemAccessor"
```
Alphabetical ordering

Juuxel (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SculkSensorBlockMixin.java
Technici4n: It should only be used for lambda bodies imo, `<clinit>` is quite well known and probably doesn't require the annotation.

Technici4n: This isn't resolved yet since `@Dynamic` is still there? I don't think it makes sense to have it.

Player3324 (approved):

modmuss50 (approved):

modmuss50:
Looks good, please fix conflicts 👍

PR #2280, opened 2022-06-02
Lazily evaluate end biome overrides.

### LINKED ISSUE #2259: Crash at the world creation screen with datapack content added through the Biome API

modmuss50:
Hopefully fixes #2259

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MixinTheEndBiomeSource.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MixinTheEndBiomeSource.java
@@ -16,6 +16,9 @@
 
 package net.fabricmc.fabric.mixin.biome;
 
+import java.util.function.Supplier;
+
+import com.google.common.base.Suppliers;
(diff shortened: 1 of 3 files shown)


modmuss50:
Superseded by #2282

PR #2296, opened 2022-06-08
Message API: add clientside events

apple502j:
WIP, untested.

This update will bring ~~three~~ two things:

- `ClientMessageEvents`. Like `ServerMessageEvents`, if the message is blocked by the client, it won't be handled. It uses a similar set of callbacks using the same parameter orders with the server-side ones, however command callback obviously does not exist, and `RegistryKey<MessageType> typeKey` is replaced with `MessageType type` due to technical reasons. (Those who need the key can lookup DRM.) Currently there is no Allow callback. This is implemented by adding a custom `ClientChatListener` to `InGameHud#listeners`.
- `ClientMessageHelper`. The event will receive `MessageSender`, but the client might want to resolve the otherwise-empty record; these methods provide UUID-to-entity/player list entry lookup.
- ~~`ServerChatEvents#MODIFY_SENDER_NAME` event. This can be used for coloring player names (note that custom message type might be better), pronouns mods, nicknames, etc. Since the name is not signed, we can modify this with no issue. This will use event phases like `ServerMessageDecoratorEvent`.~~ (postponed due to usability issues)

### DIFF
--- a/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
+++ b/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


apple502j:
Note: MODIFY_SENDER_NAME is postponed. This PR is therefore ready to review.

Technici4n:
So this is not in progress anymore?

apple502j:
@Technici4n Yes

Technici4n:
Seems easy enough. What would be the use case? Would the ability to cancel the message and/or change it would make sense compared to only having "notification" callbacks?

apple502j:
@Technici4n Not sure about the use case for changing the message client-side. Canceling might have a use, but probably later.

Main use case would be mods that ping the user or store messages in other places.

Technici4n:
Sure thing, I assume cancel and modify would be different events running before the ones added by this PR anyway; they could easily be added after the fact.

Technici4n (changes requested):
I'm going to request a small testmod. ;)

apple502j:
Closing; will PR targeting 1.19.1 instead.

PR #2307, opened 2022-06-11
Add TagUtil

### LINKED ISSUE #2276: Addition of a generic `isIn` for use with `TagKeys` that that do not have easy methods of querying
... such as `Enchantment`s.

```java
enchantment -> {
	    var maybeKey = Registry.ENCHANTMENT.getKey(enchantment);
	    return maybeKey.filter(enchantmentRegistryKey ->
	                                   Registry.ENCHANTMENT.entryOf(enchantmentRegistryKey).isIn(tagKey))
	                   .isPresent();
	}
```

This makes using `ConventionalEnchantmentTags` somewhat annoying to implement, and serves as a barrier for other registries to be used with tags.

One solution would be to inject a better way to get the registry key, as well as a method in `TagKey` to query if an object is in that specific tag.

dexman545:
Resolves #2276 

Generic tag queries.

### DIFF
--- a/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
+++ b/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
@@ -34,10 +34,10 @@ public final class TagUtil {
 	}
 
 	/**
-	* See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
-	* registries.
-	* @return if the entry is in the provided tag.
-	*/
+	 * See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
+	 * registries.
+	 * @return if the entry is in the provided tag.
+	 */
 	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
 		return isIn(null, tagKey, entry);
 	}
--- a/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
+++ b/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
@@ -31,6 +31,8 @@ import net.minecraft.util.registry.RegistryKey;
  */
 public class TagUtil {
 	/**
+	* See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
+	* registries.
 	* @return if the entry is in the provided tag.
 	*/
 	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
--- a/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
+++ b/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
@@ -18,7 +18,10 @@ package net.fabricmc.fabric.api.tag.convention.v1;
 
 import java.util.Optional;
 
+import org.jetbrains.annotations.Nullable;
+
 import net.minecraft.tag.TagKey;
+import net.minecraft.util.registry.DynamicRegistryManager;
 import net.minecraft.util.registry.Registry;
 import net.minecraft.util.registry.RegistryEntry;
 import net.minecraft.util.registry.RegistryKey;
@@ -28,11 +31,32 @@ import net.minecraft.util.registry.RegistryKey;
  */
 public class TagUtil {
 	/**
+	* @return if the entry is in the provided tag.
+	*/
+	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
+		return isIn(null, tagKey, entry);
+	}
+
+	/**
+	 * @param registryManager the registry manager instance of the client or server. If the tag refers to entries
+	 *                        within a dynamic registry, such as {@link net.minecraft.world.biome.Biome}s,
+	 *                        this must be passed to correctly evaluate the tag. Otherwise, the registry is found by
+	 *                        looking in {@link Registry#REGISTRIES}.
 	 * @return if the entry is in the provided tag.
 	 */
 	@SuppressWarnings("unchecked")
-	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
-		Optional<? extends Registry<?>> maybeRegistry = Registry.REGISTRIES.getOrEmpty(tagKey.registry().getValue());
+	public static <T> boolean isIn(@Nullable DynamicRegistryManager registryManager, TagKey<T> tagKey, T entry) {
+		Optional<? extends Registry<?>> maybeRegistry;
+
+		if (registryManager != null) {
+			maybeRegistry = registryManager.getOptional(tagKey.registry());
+
+			if (maybeRegistry.isEmpty()) {
+				maybeRegistry = Registry.REGISTRIES.getOrEmpty(tagKey.registry().getValue());
+			}
+		} else {
+			maybeRegistry = Registry.REGISTRIES.getOrEmpty(tagKey.registry().getValue());
+		}
 
 		if (maybeRegistry.isPresent()) {
 			if (tagKey.isOf(maybeRegistry.get().getKey())) {
--- a/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
+++ b/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
@@ -29,7 +29,7 @@ import net.minecraft.util.registry.RegistryKey;
 /**
  * A Helper class for dealing with {@link TagKey}s when their type has no easy way of querying if they are in a tag.
  */
-public class TagUtil {
+public final class TagUtil {
 	/**
 	* See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
 	* registries.
--- a/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
+++ b/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.tag.convention.v1;
+
+import java.util.Optional;
+
+import net.minecraft.tag.TagKey;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryEntry;
+import net.minecraft.util.registry.RegistryKey;
+
+/**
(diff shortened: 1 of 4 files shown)


fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,58 @@
...
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * @return if the entry is in the provided tag.
+	 */
+	public static <T> boolean isIn(TagKey<T> tagKey, RegistryEntry<T> registryEntry) {
```
Technici4n: Why? :D

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * A Helper class for dealing with {@link TagKey}s when their type has no easy way of querying if they are in a tag.
+ */
+public class TagUtil {
+	/**
+	 * @return if the entry is in the provided tag.
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
+		Optional<? extends Registry<?>> maybeRegistry = Registry.REGISTRIES.getOrEmpty(tagKey.registry().getValue());
```
Technici4n: Will this work correctly for dynamic registry entries?

dexman545: This is now supported by the version of this method that takes a dynamic registry.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
dexman545: Completeness :P
It can be removed if needed.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -28,11 +31,32 @@
...
 	@SuppressWarnings("unchecked")
-	public static <T> boolean isIn(TagKey<T> tagKey, T entry) {
-		Optional<? extends Registry<?>> maybeRegistry = Registry.REGISTRIES.getOrEmpty(tagKey.registry().getValue());
+	public static <T> boolean isIn(@Nullable DynamicRegistryManager registryManager, TagKey<T> tagKey, T entry) {
+		Optional<? extends Registry<?>> maybeRegistry;
+
+		if (registryManager != null) {
+			maybeRegistry = registryManager.getOptional(tagKey.registry());
+
+			if (maybeRegistry.isEmpty()) {
```
apple502j: I believe this is redundant.

dexman545: The first one pulls from the dynamic registry, the latter pulls from the static registry. I don't believe the dynamic registry contained the static ones?

apple502j: getOptional does fall back to static ones. getOptionalManaged is the one that does not.

dexman545: that it does, fixed

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,80 @@
...
+import net.minecraft.tag.TagKey;
+import net.minecraft.util.registry.DynamicRegistryManager;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryEntry;
+import net.minecraft.util.registry.RegistryKey;
+
+/**
+ * A Helper class for dealing with {@link TagKey}s when their type has no easy way of querying if they are in a tag.
+ */
+public class TagUtil {
```
haykam821:
```suggestion
public final class TagUtil {
```

haykam821 (changes requested):

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,80 @@
...
+import net.minecraft.tag.TagKey;
+import net.minecraft.util.registry.DynamicRegistryManager;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryEntry;
+import net.minecraft.util.registry.RegistryKey;
+
+/**
+ * A Helper class for dealing with {@link TagKey}s when their type has no easy way of querying if they are in a tag.
+ */
+public final class TagUtil {
```
apple502j, line 36: This needs a private constructor.

fabric-convention-tags-v1/src/testmod/java/net/fabricmc/fabric/test/tag/convention/v1/TagUtilTest.java
```diff
@@ -0,0 +1,56 @@
...
+import net.fabricmc.fabric.api.tag.convention.v1.ConventionalBlockTags;
+import net.fabricmc.fabric.api.tag.convention.v1.ConventionalEnchantmentTags;
+import net.fabricmc.fabric.api.tag.convention.v1.TagUtil;
+
+public class TagUtilTest implements ModInitializer {
+	private static final Logger LOGGER = LoggerFactory.getLogger(TagUtilTest.class);
+
+	@Override
+	public void onInitialize() {
+		ServerLifecycleEvents.SERVER_STARTED.register(server -> {
```
Technici4n, line 39: Could you add a test that tests dynamic registry tags? Currently all of the tests test static tags. This will make porting mistakes easier to detect.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
Technici4n: I'd remove this, the entire point of `TagUtil` is to not need to query the `RegistryEntry`.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,83 @@
...
+	private TagUtil() {
+	}
+
+	/**
+	* See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
+	* registries.
+	* @return if the entry is in the provided tag.
+	*/
```
Technici4n:
```suggestion
	 * See {@link TagUtil#isIn(DynamicRegistryManager, TagKey, Object)} to check tags that refer to entries in dynamic
	 * registries.
	 * @return if the entry is in the provided tag.
	 */
```

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+import net.minecraft.tag.TagKey;
+import net.minecraft.util.registry.DynamicRegistryManager;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryEntry;
+import net.minecraft.util.registry.RegistryKey;
+
+/**
+ * A Helper class for dealing with {@link TagKey}s when their type has no easy way of querying if they are in a tag.
+ */
```
Technici4n: Some extra javadoc would be nice imo. This is just a suggestion/sketch:

```suggestion
/**
 * A Helper class for checking whether a {@link TagKey} contains some entry.
 * This can be useful for {@link TagKey}s whose type has no easy way of querying if they are in a tag.
 *
 * <p>For dynamic registry entries, use {@link #isIn(DynamicRegistryManager, TagKey, Object)} with a non-null dynamic registry manager.
 * For non-dynamic registry entries, the simpler {@link #isIn(TagKey, Object)} can be used.
 */
```

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
Technici4n: (weird javadoc indent)

Technici4n (changes requested):
Looks great, just a few nitpicks.

fabric-convention-tags-v1/src/testmod/java/net/fabricmc/fabric/test/tag/convention/v1/TagUtilTest.java
Technici4n: Nevermind, this is already tested with the `IN_OVERWORLD` biome tag below. Consider this resolved.

Technici4n (approved):
Looks great, and the testmod is working as expected!

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
Technici4n: Good now, thanks a lot!

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * @param registryManager the registry manager instance of the client or server. If the tag refers to entries
+	 *                        within a dynamic registry, such as {@link net.minecraft.world.biome.Biome}s,
+	 *                        this must be passed to correctly evaluate the tag. Otherwise, the registry is found by
+	 *                        looking in {@link Registry#REGISTRIES}.
+	 * @return if the entry is in the provided tag.
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T> boolean isIn(@Nullable DynamicRegistryManager registryManager, TagKey<T> tagKey, T entry) {
```
Technici4n, line 57: Would be nice to add `Objects.requireNonNull` for `tagKey` and `entry`.

fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/TagUtil.java
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		if (maybeRegistry.isPresent()) {
+			if (tagKey.isOf(maybeRegistry.get().getKey())) {
+				Registry<T> registry = (Registry<T>) maybeRegistry.get();
+
+				Optional<RegistryKey<T>> maybeKey = registry.getKey(entry);
+
+				// Check synced tag
+				return maybeKey.filter(registryKey -> registry.entryOf(registryKey).isIn(tagKey))
```
Technici4n: Could I maybe suggest using more explicit code? I'm not very familiar with `filter`, which makes the code harder to read for me than necessary.

Technici4n (approved):

Technici4n (commented):
Question: What if these methods were interface-injected into `TagKey`? Sorry for noticing so late...

apple502j:
Hmm, do we actually use interface injection to add our own methods?

Technici4n:
> Hmm, do we actually use interface injection to add our own methods?

Isn't adding methods to vanilla classes the whole point of interface injection?

dexman545:
I'm not sure II will be helpful for discovery, since all other tag checks are done on other objects, not the tagkey itself.

apple502j:
@Technici4n The main use of interface injection, as far as I'm aware, is to allow subclassses of vanilla class to implement Fabric methods.

Another thing to consider is that using interface injection there completely misses the point of the 1.18.2 tag refactor. `TagKey` is just a key and does not hold the values, unlike `Tag.Identified`. Having `isIn` methods (which, if injected, should be `contains` I think?) makes it look like the key holds the values.

Technici4n:
Yeah, I guess it's not ideal here. Let's go for the separate static helpers then.

modmuss50 (approved):

apple502j (approved):

Player3324 (approved):

PR #2326, opened 2022-06-15
Fix #2108: FlammableBlockRegistry ignores tags unless /reload

### LINKED ISSUE #2108: FlammableBlockRegistry ignores tags unless `/reload`

Technici4n:
Switched to the approach used by the FuelRegistry.


apple502j (approved):
Note to self: do not cause fire in the deserts.

modmuss50 (approved):

PR #2338, opened 2022-06-19
Make screen API crash easier to debug

### LINKED ISSUE #1569: ScreenMixin.ensureEventsAreInitialised() error is outdated and incorrect
https://github.com/FabricMC/fabric/blob/1.17/fabric-screen-api-v1/src/main/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java#L153
> This is usually caused by the screen not calling super.init(Lnet/minecraft/client/MinecraftClient;II)V

The mentioned overload of `init` is final in 1.17, so its no longer possible to miss out the supercall, as the method cannot be overridden in the first place. The error is still useful though, I encountered it due to incorrect async code.
Additionally the error uses the name of the screen's superclass, rather than the name of the screen's class.

apple502j:
Resolves #1569 

The original crash `ScreenMixin` was trying to catch can no longer happen after 1.17. However, the crash still occurs on off-thread `setScreen` as that introduces a race condition. This PR fixes the crash message.

Additionally, this also adds a check for current thread on `setScreen` itself. The vanilla check is only for `isDevelopment` environment and the error was too short. Our error includes the screen class name and the thread name which should make it clear what's going on (calling setScreen on Netty thread/Server thread/etc).

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
@@ -150,7 +150,7 @@ abstract class ScreenMixin implements ScreenExtensions {
 	@Unique
 	private <T> Event<T> ensureEventsAreInitialised(Event<T> event) {
 		if (event == null) {
-			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by the screen not calling super.init(Lnet/minecraft/client/MinecraftClient;II)V", this.getClass().getSuperclass().getName()));
+			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getSuperclass().getName()));
 		}
 
 		return event;
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
@@ -150,7 +150,7 @@ abstract class ScreenMixin implements ScreenExtensions {
 	@Unique
 	private <T> Event<T> ensureEventsAreInitialised(Event<T> event) {
 		if (event == null) {
-			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getSuperclass().getName()));
+			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getName()));
 		}
 
 		return event;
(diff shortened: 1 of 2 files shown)


Technici4n (approved):
Seems sensible to me.

shartte:
I am not sure, why does this one method deserve a thread-check, while all the other API mehods are just as thread-hostile and don't get one? :-P

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
```diff
@@ -150,7 +150,7 @@ public List<ClickableWidget> fabric_getButtons() {
 	@Unique
 	private <T> Event<T> ensureEventsAreInitialised(Event<T> event) {
 		if (event == null) {
-			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by the screen not calling super.init(Lnet/minecraft/client/MinecraftClient;II)V", this.getClass().getSuperclass().getName()));
+			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getSuperclass().getName()));
```
modmuss50: Have you got an example of this beign the case? I think the .getSuperclass() needs removing here though?

apple502j:
@shartte Because some people encountered this situation and it's not easy to debug. Currently, calling `setScreen` on the wrong thread will crash randomly due to `GameRenderer` racing.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
apple502j: Think so. And yes, reaching this is as easy as `new Thread(() -> MinecraftClient.getInstance().setScreen(screen)).start()`

apple502j:
```suggestion
			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getName()));
```

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
```diff
@@ -150,7 +150,7 @@ public List<ClickableWidget> fabric_getButtons() {
 	@Unique
 	private <T> Event<T> ensureEventsAreInitialised(Event<T> event) {
 		if (event == null) {
-			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by the screen not calling super.init(Lnet/minecraft/client/MinecraftClient;II)V", this.getClass().getSuperclass().getName()));
+			throw new IllegalStateException(String.format("[fabric-screen-api-v1] The current screen (%s) has not been correctly initialised, please send this crash log to the mod author. This is usually caused by calling setScreen on the wrong thread.", this.getClass().getName()));
```
modmuss50, line 153: The old reason should still apply here?

modmuss50 (changes requested):

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
apple502j: That method is now final and called by setScreen.

modmuss50 (approved):

PR #2339, opened 2022-06-19
Optimise file exists check for directory based ModNioResourcePacks

modmuss50:
NIO Files.exists is notoriously slow and allocation heavy, when the path is on the computers FS using the legacy .exists API is much faster. This is only an issue when you have a directory based resource pack, zip files are not affected (What most people use anyway)

See:

[Nio vs Io File Absence performance](https://gist.github.com/CodingFabian/422b6cf4bfcdfece54c1)
[Open JDK does the same in 21 places](https://github.com/openjdk/jdk/blob/b544b8b7d43907e93263db31ba3cc6d5951bcaee/src/jdk.jfr/share/classes/jdk/jfr/internal/SecuritySupport.java#L388-L389)
"brings reload times from 30s to 5s or so in testmod client" - Technici4n

Still needs propper testing, but I think it should be good.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
@@ -20,6 +20,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.DirectoryStream;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


LambdAurora (approved):

Technici4n (approved):
This will need backporting to 1.18.

Player3324 (approved):
Concerning that this has such an impact..

liach:
Just curious, how fast is the zip fs provider implementation compared to, say, zipfile for checking existence?

The fact that they throw 2 exceptions in a check alone is shocking enough to me.

modmuss50:
> Just curious, how fast is the zip fs provider implementation compared to, say, zipfile for checking existence?
> 
> The fact that they throw 2 exceptions in a check alone is shocking enough to me.

I dont have comparision numbers, but I know for sure its not as bad as this. The zip fs impl is used a lot with this without any noticable slowdown. These exceptions seen in the profile are beign thrown within the WindowsFileSystemProvider.

There may well be further room for improvement around this, but that will need further investigation.

PR #2341, opened 2022-06-20
Even More TAWs

Shnupbups:
More TAWs - this time for non-datagen stuff.

- `DamageSource` constructor and methods made accessible - supersedes #1085
- `SpawnRestriction` method `register` made accessible - allows `SpawnRestrictionAccessor` mixin to be removed
- `TradeOffers` trade factory inner classes made accessible
- `Material$Builder` methods made accessible (except `lightPassesThrough` method as that would break binary compat in `FabricMaterialBuilder`
- Fixed `generateAccessWideners` task for TAWs module so it can now find the jar
- `AxeItem`, `MusicDiscItem`, and `PickaxeItem` constructors made accessible
- `ItemUsageContext` and `ItemPlacementContext` alternate constructors made accessible - allows passing a `null` PlayerEntity
- `Factory` inner classes of both `CookingRecipeSerializer` and `CuttingRecipeSerializer` made accessible
- `BrewingRecipeRegistry` methods `registerItemRecipe`, `registerPotionType`, and `registerPotionRecipe` made accessible
- `ScoreboardCriterion` `create` methods made accessible
- `ModelPart` inner classes `Vertex` and `Quad` made accessible
- `RenderLayer` `of` method made accessible
- `SensorType` and `Activity` constructors made accessible
- `LivingEntity` methods `damageArmor`, `damageHelmet`, and `damageShield` made accessible
- `ProjectileEntity` constructor made accessible
- `FishingBobberEntity` alternate constructor made accessible - allows passing luck of the sea or lure levels for custom fishing bobbers
- `ServerWorld` method `sendToPlayerIfNearby` made accessible
- `TextColor` method `getHexCode` made accessible
- `TranslatableTextContent` method `getArg` made accessible


Shnupbups:
Most recent commit adds the following:
- Made constructor of `ProjectileEntity` accessible, to allow for subclasses to be constructed
- Made alternate constructor of `FishingBobberEntity` accessible, to allow for custom fishing bobber entity types with non-zero luck of the sea or lure levels
- Made alternate constructors of `ItemUsageContext` and `ItemPlacementContext` accessible, to allow for passing a null player entity
- Made text filtering methods in `ServerPlayNetworkHandler` accessible, so mods can filter text in places other than chat, signs, and books
- Made register methods for `EntityRenderers` and `BlockEntityRendererFactories` accessible, so mods can use those instead of the Fabric API hooks if they want (not sure if this means we should deprecate those hooks, or if they're still useful in some form?)
- Made recipe factory subclasses for `CookingRecipeSerializer` and `CuttingRecipe$Serializer` accessible, so those serializers can be used for datagen (or other thing, I guess)
- Made register methods in `BrewingRecipeRegistry` accessible, so brewing recipes (and even new potion types beyond regular, splash and lingering) can be registered
- Made create/register methods in `ScoreboardCriterion` accessible, so custom scoreboard criteria can be created and registered
- Made subclasses of `ModelPart` accessible, I had a look around GitHub and saw quite a few mods do this, so I'm sure it's useful
- Made a few other miscellaneous methods accessible - getting the item for an `AbstractMinecartEntity`, getting the hex code of a `TextColor`, getting the arg of a `TranslatableTextContent`, and a method in `ServerWorld` for sending a packet to a player if they're near a specific position

Technici4n:
Can you add the following?
image

Should also probably have a look at forge's AT and decide for each entry if we want to have it or not.

Shnupbups:
Added your suggestions - and also a ton of methods related to entities.
Almost every private or package private entity method made accessible, as well as a few protected ones that are nice to have.
I also made a version that includes almost all protected entity methods, but felt that may be excessive...
In fact, this may be considered excessive by some, so let me know what you think, if I should cut it down a notch.
If you think it's too much, then tell me which ones I should keep.

Technici4n:
Imo we should only have the essential TAWs, otherwise we will end up with every protected and private method, which is definitely not the point.

haykam821:
It would be interesting to run a scraper on GitHub's search results for access wideners to find to most commonly-used access wideners.

Shnupbups:
Removed a majority of the entity methods again

fabric-transitive-access-wideners-v1/template.accesswidener
```diff
@@ -24,4 +24,104 @@ transitive-accessible field net/minecraft/loot/LootPool bonusRolls Lnet/minecraf
...
+transitive-accessible method net/minecraft/item/ItemPlacementContext <init> (Lnet/minecraft/world/World;Lnet/minecraft/entity/player/PlayerEntity;Lnet/minecraft/util/Hand;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/hit/BlockHitResult;)V
+
+# Text filtering
+transitive-accessible method net/minecraft/server/network/ServerPlayNetworkHandler filterText (Ljava/lang/Object;Ljava/util/function/Consumer;Ljava/util/function/BiFunction;)V
+transitive-accessible method net/minecraft/server/network/ServerPlayNetworkHandler filterText (Ljava/lang/String;Ljava/util/function/Consumer;)V
+transitive-accessible method net/minecraft/server/network/ServerPlayNetworkHandler filterTexts (Ljava/util/List;Ljava/util/function/Consumer;)V
+
+# Registering custom renderers
+transitive-accessible method net/minecraft/client/render/entity/EntityRenderers register (Lnet/minecraft/entity/EntityType;Lnet/minecraft/client/render/entity/EntityRendererFactory;)V
+transitive-accessible method net/minecraft/client/render/block/entity/BlockEntityRendererFactories register (Lnet/minecraft/block/entity/BlockEntityType;Lnet/minecraft/client/render/block/entity/BlockEntityRendererFactory;)V
```
Technici4n: Fabric has an API for this to accomodate for late registration (afaict?). So, not sure if they should be added.

fabric-transitive-access-wideners-v1/template.accesswidener
```diff
@@ -24,4 +24,104 @@ transitive-accessible field net/minecraft/loot/LootPool bonusRolls Lnet/minecraf
...
+transitive-accessible method net/minecraft/entity/LivingEntity damageHelmet (Lnet/minecraft/entity/damage/DamageSource;F)V
+transitive-accessible method net/minecraft/entity/LivingEntity damageShield (F)V
+
+# Armor stand methods
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity isSlotDisabled (Lnet/minecraft/entity/EquipmentSlot;)Z
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity equip (Lnet/minecraft/entity/player/PlayerEntity;Lnet/minecraft/entity/EquipmentSlot;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/Hand;)Z
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity setSmall (Z)V
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity setShowArms (Z)V
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity setHideBasePlate (Z)V
+transitive-accessible method net/minecraft/entity/decoration/ArmorStandEntity setMarker (Z)V
```
Technici4n: Why is this useful?

fabric-transitive-access-wideners-v1/template.accesswidener
```diff
@@ -24,4 +24,104 @@ transitive-accessible field net/minecraft/loot/LootPool bonusRolls Lnet/minecraf
...
+
+# Miscellaneous entity methods and constructors
+transitive-accessible method net/minecraft/entity/decoration/ItemFrameEntity getAsItemStack ()Lnet/minecraft/item/ItemStack;
+transitive-accessible method net/minecraft/entity/projectile/ProjectileEntity <init> (Lnet/minecraft/entity/EntityType;Lnet/minecraft/world/World;)V
+transitive-accessible method net/minecraft/entity/projectile/PersistentProjectileEntity asItemStack ()Lnet/minecraft/item/ItemStack;
+transitive-accessible method net/minecraft/entity/projectile/FishingBobberEntity <init> (Lnet/minecraft/entity/EntityType;Lnet/minecraft/world/World;II)V
+transitive-accessible method net/minecraft/entity/vehicle/AbstractMinecartEntity getItem ()Lnet/minecraft/item/Item;
```
Technici4n: Why?

Technici4n (commented):
How do you decide which TAWs to add? Imo we should be rather conservative when adding them.

fabric-transitive-access-wideners-v1/template.accesswidener
Shnupbups: I suppose it's only really useful to a select crowd (mostly server side devs), so I've removed it again.

fabric-transitive-access-wideners-v1/template.accesswidener
Shnupbups: Decided to remove the `getAsItemStack`, `asItemStack`, and `getItem` methods again. Kept the two constructors.

fabric-transitive-access-wideners-v1/template.accesswidener
```diff
@@ -24,4 +24,92 @@ transitive-accessible field net/minecraft/loot/LootPool bonusRolls Lnet/minecraf
...
+
+# Entity constructors
+transitive-accessible method net/minecraft/entity/projectile/ProjectileEntity <init> (Lnet/minecraft/entity/EntityType;Lnet/minecraft/world/World;)V
+transitive-accessible method net/minecraft/entity/projectile/FishingBobberEntity <init> (Lnet/minecraft/entity/EntityType;Lnet/minecraft/world/World;II)V
+
+# Miscellaneous other methods
+transitive-accessible method net/minecraft/server/world/ServerWorld sendToPlayerIfNearby (Lnet/minecraft/server/network/ServerPlayerEntity;ZDDDLnet/minecraft/network/Packet;)Z
+transitive-accessible method net/minecraft/text/TextColor getHexCode ()Ljava/lang/String;
+transitive-accessible method net/minecraft/text/TranslatableTextContent getArg (I)Lnet/minecraft/text/StringVisitable;
+
```
ghost, line 104: Could you add this?
tawsi

This is useful to add blocks that could be excaved with shovel, like dirt path.


`transitive-accessible field net/minecraft/item/ShovelItem PATH_STATES Ljava/util/Map;` should work.

ghost (changes requested):

fabric-transitive-access-wideners-v1/template.accesswidener
OroArmor: This feature is already in the content registry module

Shnupbups: As Oro said, you can add new entries to this through Fabric API already. Same with Axe stripping/scraping, Hoe tilling, and Honeycomb waxing.
I'm hesitant to widen fields anyway - can't entirely explain why, I just feel that methods and classes are more "clean" to widen, especially transitively.

fabric-transitive-access-wideners-v1/template.accesswidener
Shnupbups: Yeah I couldn't tell if the FAPI version had any inherent benefits over the vanilla one, and so figured it couldn't hurt to widen the vanilla one. If it's better for modders to use the FAPI one to avoid issues or something, then I suppose I can remove this. Will need further input.

fabric-transitive-access-wideners-v1/src/main/resources/fabric-transitive-access-wideners-v1.accesswidener
```diff
@@ -24,6 +24,94 @@ transitive-accessible field net/minecraft/loot/LootPool bonusRolls Lnet/minecraf
...
+
+# Creating custom block materials
+transitive-accessible method net/minecraft/block/Material$Builder lightPassesThrough ()Lnet/minecraft/block/Material$Builder;
+transitive-accessible method net/minecraft/block/Material$Builder burnable ()Lnet/minecraft/block/Material$Builder;
+transitive-accessible method net/minecraft/block/Material$Builder destroyedByPiston ()Lnet/minecraft/block/Material$Builder;
+transitive-accessible method net/minecraft/block/Material$Builder blocksPistons ()Lnet/minecraft/block/Material$Builder;
+
+# Item constructors
+transitive-accessible method net/minecraft/item/AxeItem <init> (Lnet/minecraft/item/ToolMaterial;FFLnet/minecraft/item/Item$Settings;)V
+transitive-accessible method net/minecraft/item/MusicDiscItem <init> (ILnet/minecraft/sound/SoundEvent;Lnet/minecraft/item/Item$Settings;)V
```
Shnupbups: Worth noting too that this will need to be updated in 1.19.1 to the new constructor that adds an int to the end for the disc's runtime. Would be nice if there was some way to automatically update access wideners to new mappings, similar to (or part of) the migrateMappings task.

ghost:
Any news about completing this?

ghost (approved):

Technici4n:
> Any news about completing this?

Waiting for #2154.

fabric-transitive-access-wideners-v1/src/main/resources/fabric-transitive-access-wideners-v1.accesswidener
Shnupbups:
```suggestion
transitive-accessible method net/minecraft/item/MusicDiscItem <init> (ILnet/minecraft/sound/SoundEvent;Lnet/minecraft/item/Item$Settings;I)V
```

Technici4n (approved):

modmuss50 (approved):

Player3324 (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricMaterialBuilder.java
```diff
@@ -43,8 +43,9 @@ public FabricMaterialBuilder pistonBehavior(PistonBehavior behavior) {
 		return this;
 	}
 
+	@Override
 	public FabricMaterialBuilder lightPassesThrough() {
```
Juuxel: Does this keep binary compat? The vanilla method is package-private so this didn't override it before, meaning it got the name `lightPassesThrough` in the obf jar (I confirmed this). Now it'll probably result in `method_15812` in the obf jar.

Shnupbups: I actually posed this question in the PR description, but it seems no one noticed that...
I'm not sure, and if it doesn't then I don't know how to go about it short of straight up removing the access widener.

modmuss50: Good spot, yes I think will will cause issues. Like you said I think removing the AW is the only option?

Shnupbups: Removed the widener for that method. Remind me to add it back in for the 1.20 cycle.

modmuss50: Great, I think that sounds reasonable. 👍

PR #2344, opened 2022-06-20
Fix Indigo AO calculation

### LINKED ISSUE #1475: Fabric API breaks underwater smooth lighting
Using Fabric API breaks the [underwater smooth lighting fix](https://bugs.mojang.com/browse/MC-68129) that was implemented in 1.17-pre2.

Affects Minecraft 1.17-pre2+ & API 0.34.8-0.57.3 and is fixed in API 0.58.0 for all Minecraft releases.

### LINKED ISSUE #1708: Tinted glass shadow/AO issue
In at least the 0.4 Fabric API for 1.17.1, there appears to be issues with either shadows or ambient occlusion in specific situations using tinted glass.

**Setup:**
For my setup, I'm running Minecraft 1.17.1, Fabric Loader 11.6 and only 0.4 Fabric API with no resource packs.

This is an example of it working as expected with FAPI

However, in this arrangement, there are clear lines where there shouldn't be, it should be smooth.

PepperCode1:
- Ensure `calcFastVanilla` and `computeFace` have parity with vanilla
- Ensure `BlockState`s are not queried more than once for the same position
- Replace deprecated `FabricLoader#getConfigDirectory` call with `FabricLoader#getConfigDir`
- This fix should be backported to 1.18.2.
- Fixes #1475
- Fixes #1708

### DIFF
--- a/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
+++ b/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
@@ -23,8 +23,8 @@ import java.io.IOException;
 import java.util.Locale;
 import java.util.Properties;
 
-import org.slf4j.LoggerFactory;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import net.fabricmc.api.ClientModInitializer;
 import net.fabricmc.fabric.api.renderer.v1.RendererAccess;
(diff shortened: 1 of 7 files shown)


Technici4n (commented):
I'm honestly not qualified enough to understand the changes, and I think that's the case of most of the team, so since this is an impl-only change we will probably have to play around with this and see if the resulting render looks ok.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
```diff
@@ -89,7 +89,7 @@ private static TriState asTriState(String property) {
 	}
 
 	static {
-		File configDir = new File(FabricLoader.getInstance().getConfigDirectory(), "fabric");
+		File configDir = FabricLoader.getInstance().getConfigDir().resolve("fabric").toFile();
```
modmuss50, line 92: Could possibly move this all to use NIO path's, but it makes no real difference, so dont worry about it for now 👍

Player3324 (approved):
LGTM, thanks!

modmuss50 (approved):

PR #2345, opened 2022-06-20
Fix custom dimension not getting loaded on first world load in 1.19

deirn:
- Port ServerBugfixMixin to 1.19, fixes the issue on dedicated server.
- `WorldPresets#createDefaultOptions` doesn't load custom dimension, causing this issue. The workaround is to parse (?) `GeneratorOptions` with custom dimension jsons. Also apparently when you add a new datapack in selection screen the issue won't happen.

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/ServerBugfixMixin.java
@@ -16,9 +16,22 @@
 
 package net.fabricmc.fabric.mixin.dimension;
 
+import com.mojang.serialization.DynamicOps;
+import com.mojang.serialization.Lifecycle;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
+import org.spongepowered.asm.mixin.injection.Redirect;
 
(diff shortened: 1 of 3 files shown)


apple502j (approved):

Technici4n:
BTW, since we were tired of these issues in AE2, we simply create the `ServerWorld` directly in 1.19. Would recommend dimension mods to do that because mojang's system is a PITA to use, and still causes various issues when a mod is removed, etc... Mojmap warning: https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/master/src/main/java/appeng/mixins/spatial/MinecraftServerMixin.java.

deirn:
That's cursed :P. The mod removal issue should already be fixed in #2078.

modmuss50 (approved):

TelepathicGrunt:
Has this been tested in single player as well? The issue is occurring with my bumblezone dimension mod for 1.19 MC with 0.56.3 Fabric API. To reproduce, put on a dimension mod, create a new single player world, do `/execute in` and see the modded dimension is not loaded in. Exit the world and then re-enter and now the dimension is loaded in.

deirn:
> Has this been tested in single player as well?

Yes, it's how it got doscovered in the first place :P.

thiamine128:
Now data packs about dimensions are loaded correctly when creating a default world.
But the test mod of dimension API crashed when creating a super flat world. It seems that the data packs about dimensions are not loaded when creating a super flat(as well as large biomes and amplified) world for the first time.

Technici4n:
@thiamine128 Please open a new issue if there is still something to fix. This PR is closed and nobody is really going to look at it again.

PR #2349, opened 2022-06-21
Make disconnected screen reason text scrollable

deirn:
Mostly happens when regsync failed caused by modpack update.


fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/DisconnectedScreenMixin.java
```diff
@@ -0,0 +1,94 @@
...
+			if (scroll < actualReasonHeight - reasonHeight) {
+				fillGradient(matrixStack, startX, y + reasonHeight - 10, endX, y + reasonHeight, 0, 0xFF000000);
+			}
+		}
+
+		return y + reasonHeight;
+	}
+
+	@Override
+	public void fabric_mouseScrolled(double mouseX, double mouseY, double amount) {
```
modmuss50: If im not mistake you should just able to override onMouseScrolled here, and remove the need for the 2nd mixin + interface?

modmuss50: Fixed my self as I had it cloned down for testing.

modmuss50 (approved):
Great, nice idea.

Player3324:
The change is good, but maybe we should move it to some sort of utility/enhancement module? networking-api seems a bit inappropriate

modmuss50:
Yeah.. We dont really have anything like this right now, could there be anything similar in the future?

Player3324:
I believe so, e.g. around networking (e.g. error induced disconnect are a pain to investigate atm) but maybe also other areas. I'd go for a generic module for little enhancements, especially ones that are perceivable by the end user, but don't warrant explicit opt-in/out.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/DisconnectedScreenMixin.java
deirn: Overriding `Screen`s method in the mixin is pretty much the same as `@Overwrite` so I avoided it.

modmuss50: Not really, as its adding a new method, not overriding anything. Can revert my change if you think its best?

Juuxel (approved):

Technici4n:
Can't we keep this in networking API as long as the logic is not required anywhere else? It's impl-only, so we can freely move it around.

Player3324:
For now maybe, would like to move it eventually though to not bundle optional "behavior" changes with an api.

Player3324 (approved):

PR #2353, opened 2022-06-23
Fix the bitwise comparison for BAKE_NORMALIZED

RichieSams:
The bitwise comparison is doing the opposite of what we want. It's currently checking for the *absence* of the BAKE_NORMALIZED flag. And if the flag is absent, it is normalizing. This should do the opposite.


apple502j (approved):

Player3324 (approved):

modmuss50 (approved):

modmuss50:
Going to release into 1.19 now, will chery pick into 1.18 to release alongside the next batch of 1.18 changes. (Not worth doing a whole release just for this fix). Thanks 👍

RichieSams:
Awesome! Thank you as well :)

Player3324:
Sorry this was actually wrong, the flag is meant to indicate already normalized UVs, as per the javadoc and it using the past tense normalized. The default is to expect 16th.

Issue #2315, opened 2022-06-12
Fabric API for 1.19 is incompatible with dev servers using Fabric API started by fabric-loom


jakobkmar:
The following setup is incompatible:

*Server:*
dev server started using fabric-loom `runServer`, only fabric-api added using modImplementation - the mod itself I used during my test is just a placeholder and does nothing

*Client:*
latest fabric api loaded

Issue:
![image](https://user-images.githubusercontent.com/52456572/173247376-760eef24-2164-49a2-b66f-08ab7b31424c.png)

Other reports in the discord also mentioned this exact error being logged:
```
[15:12:37] [Render thread/ERROR]: Registry remapping failed!
net.fabricmc.fabric.impl.registry.sync.RemapException: Received ID map for minecraft:command_argument_type contains IDs unknown to the receiver!
 - minecraft:test_class
 - minecraft:test_argument
```

warjort:
I assume the client is not in a development environment?

The issue is the fabric gametest mod does different things in a development environment, see for example here:
https://github.com/FabricMC/fabric/blob/1.19/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/ArgumentTypesMixin.java

You can exclude the gametest mod in the development environment like this:
````
  modImplementation("net.fabricmc.fabric-api:fabric-api:${project.fabric_version}") {
      exclude group: "net.fabricmc.fabric-api", module: "fabric-gametest-api-v1"
    }
````

haykam821:
Should Fabric API consider virtualizing argument types so that server-only argument types can be used?

warjort:
> Should Fabric API consider virtualizing argument types so that server-only argument types can be used?

See ````CommandTreeS2CPacket.ArgumentNode.write()```` and ````CommandTreeS2CPacket.readArgumentBuilder()```` where it uses the rawId of the registered argument serializers to serialize/deserialize the arguments.
Any mismatch between those ids is going to cause very broken behaviour.

There is this old PR https://github.com/FabricMC/fabric/pull/1339 which was mentioned on a similar issue https://github.com/FabricMC/fabric/issues/2168 about block entities.

haykam821:
Virtualization would mean all argument types would appear to clients as string argument types or similar, potentially only if the client does not have the argument type. Raw IDs would no longer become an issue as a result.

warjort:
[...]

PR #2356, opened 2022-06-23
Fix registry sync namespace check

### LINKED ISSUE #2348: Fabric Registry Sync marks command argument type registry as modded, even if it's not
It's caused by vanilla using `brigadier` namespace for some entries.

Related line: https://github.com/FabricMC/fabric/blob/1.19/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/MixinSimpleRegistry.java#L58

apple502j:
Resolves #2348 
Should also work as a bandaid fix for #2315 (although I bet this would just make the bug more obscure)

### DIFF
--- a/fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/RegistrySyncTest.java
+++ b/fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/RegistrySyncTest.java
@@ -109,6 +109,15 @@ public class RegistrySyncTest implements ModInitializer {
 		Validate.isTrue(RegistryAttributeHolder.get(fabricRegistry).hasAttribute(RegistryAttribute.SYNCED));
 		Validate.isTrue(!RegistryAttributeHolder.get(fabricRegistry).hasAttribute(RegistryAttribute.PERSISTED));
 
+		SimpleRegistry<String> fakeVanillaRegistry = FabricRegistryBuilder.createSimple(String.class, new Identifier("registry_sync", "fake_vanilla_registry"))
+				.attribute(RegistryAttribute.SYNCED)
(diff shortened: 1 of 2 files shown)


modmuss50:
Test seems to be failing: `Attempted to register object test twice! (at raw IDs 0 and 1 )`

fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/RegistrySyncTest.java
```diff
@@ -109,6 +109,15 @@ public void onInitialize() {
...
 		Validate.isTrue(!RegistryAttributeHolder.get(fabricRegistry).hasAttribute(RegistryAttribute.PERSISTED));
 
+		SimpleRegistry<String> fakeVanillaRegistry = FabricRegistryBuilder.createSimple(String.class, new Identifier("registry_sync", "fake_vanilla_registry"))
+				.attribute(RegistryAttribute.SYNCED)
+				.buildAndRegister();
+
+		Registry.register(fakeVanillaRegistry, new Identifier("minecraft", "test"), "test");
+		Registry.register(fakeVanillaRegistry, new Identifier("brigadier", "test"), "test");
+
+		Validate.isTrue(!RegistryAttributeHolder.get(fakeVanillaRegistry).hasAttribute(RegistryAttribute.MODDED));
```
modmuss50: This test wont work, as all registries created through the builder are marked as modded. I cannot think of a good way to "unit" test this. I will fix and manuall test. 👍

modmuss50 (approved):
Change works well, tested with a fabric api server with gametest module, and a prod fabric api client. 👍

Player3324 (approved):

PR #2363, opened 2022-06-28
Fix incorrect check in GlobalReceiverRegistry

### LINKED ISSUE #2359: Wrong Logic In GlobalReceiverRegistry
The check here is wrong:
https://github.com/FabricMC/fabric/blob/f11b62ad061f64269324c9f1c659fc495ae5e828/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java#L71

See:#2358
It should be ````if (replaced)````. This wasn't noticed before because most people register their receivers in the mod initialiser.

apple502j:
Fixes #2359 
Lock issue needs to be investigated later, not now since it doesn't seem to be actually causing a bug.
Also, this file has never been modified since the networking v1 PR... wow.


modmuss50 (approved):

Player3324 (approved):
may be nice to rename the variable to inserted to reduce confusion (also at the non-global equivalent)

Technici4n (approved):

PR #2367, opened 2022-06-28
Enable parallel builds by default.

modmuss50:
Update remotesign to a parallel capable version.
Set org.gradle.parallel.threads in actions as we are IO bound a lot of time (signing, and publishing to maven). (Might be best to only set for releases)


modmuss50:
Merged manually into 1.19.1 didnt help much sadly. Should aid in dev a bit tho when you have a few cores going spare.

modmuss50:
I will manually cherry pick this back, now working well in the 1.19.1 branch. 👍

PR #2369, opened 2022-06-30
Methods in `TheEndBiomeData` that query the intended type.

quiqueck:
When implementing a custom BiomeSource for the End, we need to know what the intended placement of an End-biome was. This PR adds Methods to `TheEndBiomeData` that are similar to `NetherBiomes.canGenerateInNether`.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
@@ -105,16 +105,36 @@ public final class TheEndBiomes {
 	}
 
 	/**
-	 * Returns true if the given biome was added in the end, considering the Vanilla end biomes,
+	 * Returns true if the given biome was added as a main end Biome in the end, considering the Vanilla end biomes,
 	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
 	 */
-	public static boolean canGenerateAsEndBiome(RegistryKey<Biome> biome){
-		return TheEndBiomeData.canGenerateAsEndBiome(biome);
+	public static boolean canGenerateAsMainIslandBiome(RegistryKey<Biome> biome){
+		return TheEndBiomeData.canGenerateAsEndBiomeIn(BiomeKeys.THE_END, biome);
+	}
+
+	/**
+	 * Returns true if the given biome was added as a small end islands Biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean canGenerateAsSmallIslandsBiome(RegistryKey<Biome> biome){
+		return TheEndBiomeData.canGenerateAsEndBiomeIn(BiomeKeys.SMALL_END_ISLANDS, biome);
+	}
+
+	/**
+	 * Returns true if the given biome was added as a Highland Biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean canGenerateAsHighlandsBiome(RegistryKey<Biome> biome){
+		return TheEndBiomeData.canGenerateAsEndBiomeIn(BiomeKeys.END_HIGHLANDS, biome);
 	}
 
 	/**
 	 * Returns true if the given biome was added as midland biome in the end, considering the Vanilla end biomes,
 	 * and any biomes added to the End as midland biome by mods.
+	 * @param biome The biome to search for
 	 */
 	public static boolean canGenerateAsEndMidlands(RegistryKey<Biome> biome){
 		return TheEndBiomeData.canGenerateAsEndMidlands(biome);
@@ -123,6 +143,7 @@ public final class TheEndBiomes {
 	/**
 	 * Returns true if the given biome was added as barrens biome in the end, considering the Vanilla end biomes,
 	 * and any biomes added to the End as barrens biome by mods.
+	 * @param biome The biome to search for
 	 */
 	public static boolean canGenerateAsEndBarrens(RegistryKey<Biome> biome){
 		return TheEndBiomeData.canGenerateAsEndBarrens(biome);
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
@@ -108,23 +108,23 @@ public final class TheEndBiomes {
 	 * Returns true if the given biome was added in the end, considering the Vanilla end biomes,
 	 * and any biomes added to the End by mods.
 	 */
-	public static boolean isIntendedForEndBiome(RegistryKey<Biome> biome){
-		return TheEndBiomeData.isIntendedForEndBiome(biome);
+	public static boolean canGenerateAsEndBiome(RegistryKey<Biome> biome){
+		return TheEndBiomeData.canGenerateAsEndBiome(biome);
 	}
 
 	/**
 	 * Returns true if the given biome was added as midland biome in the end, considering the Vanilla end biomes,
 	 * and any biomes added to the End as midland biome by mods.
 	 */
-	public static boolean isIntendedForEndMidlands(RegistryKey<Biome> biome){
-		return TheEndBiomeData.isIntendedForEndMidlands(biome);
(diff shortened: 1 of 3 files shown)


apple502j:
I am mot sure if the "intended" naming is good.

quiqueck:
We could use `canGenerateAs` instead (similar to `NetherBiomes.canGenerateInNether`).

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/TheEndBiomeData.java
```diff
@@ -90,6 +90,30 @@ public static void addEndBarrensReplacement(RegistryKey<Biome> highlands, Regist
...
+	public static boolean isIntendedForEndMidlands(RegistryKey<Biome> biome){
+		return END_MIDLANDS_MAP.containsKey(biome);
+	}
+
+	/**
+	 * Returns true if the given biome was added as barrens biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End as barrens biome by mods.
+	 */
+	public static boolean isIntendedForEndBarrens(RegistryKey<Biome> biome){
+		return END_BARRENS_MAP.containsKey(biome);
```
deirn: This doesn't seem to be the correct way to do it, the midlands and barrens map has the *highland* biome as the key.

quiqueck: Yes, just realized that here quiqueck/BCLib#13

apple502j:
I think it would be better if WeightedPicker stored the entries as sets, etc. @deirn what do you think?

deirn:
> We could use canGenerateAs instead (similar to NetherBiomes.canGenerateInNether).

It's still too verbose imo, I'd just named it like the `add` methods, e.g. `isMidlandsBiome` or so.

> I think it would be better if WeightedPicker stored the entries as sets, etc. @deirn what do you think?

I'm not really sure how `WeightedPicker` works tbh, but it looks like it needs to be a list?

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/WeightedPicker.java
```diff
@@ -100,6 +101,16 @@ WeightedEntry<T> search(final double target) {
...
 
+	/**
+	 * Returns any {@link WeightedEntry} that contains the passed element.
+	 *
+	 * @param element The Element you are looking for
+	 * @return The result of the search
+	 */
+	Optional<WeightedEntry<T>> getAny(T element){
+		return entries.stream().filter(w -> w.entry==element).findAny();
+	}
```
deirn: Instead of this, you might want to look into changing `TheEndBiomeData#ADDED_BIOMES` into a `Map<RegistryKey<Biome>, RegistryKey<Biome>>` instead, with vanilla biome as its value.  Then you can check if the value is equal to the vanilla biome.

[Patch file](https://gist.github.com/deirn/d8eb50934e666616b2982d10817c659e)

quiqueck: Yeah, that is much cleaner. I'll include the suggested patch...

quiqueck:
> > I think it would be better if WeightedPicker stored the entries as sets, etc. @deirn what do you think?
> 
> I'm not really sure how `WeightedPicker` works tbh, but it looks like it needs to be a list?

I think the `search`-method needs a list for deterministic results.

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
```diff
@@ -103,4 +103,49 @@ public static void addMidlandsBiome(RegistryKey<Biome> highlands, RegistryKey<Bi
 	public static void addBarrensBiome(RegistryKey<Biome> highlands, RegistryKey<Biome> barrens, double weight) {
 		TheEndBiomeData.addEndBarrensReplacement(highlands, barrens, weight);
 	}
+
+	/**
+	 * Returns true if the given biome was added as a main end Biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean isMainIslandBiome(RegistryKey<Biome> biome){
```
deirn:
```suggestion
	public static boolean isMainIslandBiome(RegistryKey<Biome> biome) {
```

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
```diff
@@ -103,4 +103,49 @@ public static void addMidlandsBiome(RegistryKey<Biome> highlands, RegistryKey<Bi
...
+		return TheEndBiomeData.ADDED_BIOMES.get(biome) == BiomeKeys.THE_END;
+	}
+
+	/**
+	 * Returns true if the given biome was added as a small end islands Biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean isSmallIslandsBiome(RegistryKey<Biome> biome){
```
deirn:
```suggestion
	public static boolean isSmallIslandsBiome(RegistryKey<Biome> biome) {
```

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
```diff
@@ -103,4 +103,49 @@ public static void addMidlandsBiome(RegistryKey<Biome> highlands, RegistryKey<Bi
...
+		return TheEndBiomeData.ADDED_BIOMES.get(biome) == BiomeKeys.SMALL_END_ISLANDS;
+	}
+
+	/**
+	 * Returns true if the given biome was added as a Highland Biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean isHighlandsBiome(RegistryKey<Biome> biome){
```
deirn:
```suggestion
	public static boolean isHighlandsBiome(RegistryKey<Biome> biome) {
```

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
```diff
@@ -103,4 +103,49 @@ public static void addMidlandsBiome(RegistryKey<Biome> highlands, RegistryKey<Bi
...
+		return TheEndBiomeData.ADDED_BIOMES.get(biome) == BiomeKeys.END_HIGHLANDS;
+	}
+
+	/**
+	 * Returns true if the given biome was added as midland biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End as midland biome by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean isMidlandsBiome(RegistryKey<Biome> biome){
```
deirn:
```suggestion
	public static boolean isMidlandsBiome(RegistryKey<Biome> biome) {
```

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java
```diff
@@ -103,4 +103,49 @@ public static void addMidlandsBiome(RegistryKey<Biome> highlands, RegistryKey<Bi
...
+		return TheEndBiomeData.ADDED_BIOMES.get(biome) == BiomeKeys.END_MIDLANDS;
+	}
+
+	/**
+	 * Returns true if the given biome was added as barrens biome in the end, considering the Vanilla end biomes,
+	 * and any biomes added to the End as barrens biome by mods.
+	 * @param biome The biome to search for
+	 */
+	public static boolean isBarrensBiome(RegistryKey<Biome> biome){
```
deirn:
```suggestion
	public static boolean isBarrensBiome(RegistryKey<Biome> biome) {
```

deirn (approved):
Few checkstyle issues, other than that, LGTM.

apple502j:
@quiqueck The suggestion was to use both a set and a list. Anyway, seems like it was solved.

Technici4n (commented):
Could there be a case where a biome replaces multiple end biomes, and does that need to be accounted for?

Looks good otherwise.

apple502j (changes requested):
See the review above - as far as I understand one biome can in fact replace multiple biomes.

PR #2375, opened 2022-07-02
Transfer API improvements 4

### LINKED ISSUE #2325: Remove TransactionContext parameter from Storage#exactView
I forgot to do it in my other PR. Nobody uses this feature I think, so this should be fine for 1.19 still (or if you are using it: please don't use it for the moment in 1.19 until I make this breaking change).

Technici4n:
New features:
- Add `ContainerItemContext.forPlayerInteraction` that can be used to avoid mutating a player's hand stack in creative mode.
- Add `FluidStorageUtil`, at the moment it only contains `interactWithStorage`. This is a convenient helper to make interacting with a tank via `onUse` trivial, while ensuring the correct behavior (playing the sound, not mutating the stack if the player is in creative mode, etc). The provided functionality is roughly equivalent to Forge's `FluidUtil.interactWithFluidHandler`.
- Improve single variant base implementations to reduce boilerplate code for API users. More precisely:
  - Add `writeNbt` method in `SingleVariantStorage`.
  - Introduce the `SingleFluidStorage` and `SingleItemStorage` subclasses that override `getBlankVariant()` and also provide a matching `readNbt`.
  - Add a way to create a `SingleFluidStorage` without needing a subclass: `SingleFluidStorage.withFixedCapacity(long capacity, Runnable onChange)`.
- Add `FluidVariantAttributes#enableColoredVanillaFluidNames`. This method can be called by mods that wish to have colored water and lava names. If nobody calls this method, the names are shown in white. **This doesn't influence vanilla behavior, it only affects mods that call `FluidVariantAttributes.getName(variant)`**.
- Add `TransferVariant.getOrCopyNbt`, a few people asked for it.

Cleanup changes:
- Deprecate the transaction-aware `Storage.getExactView` and introduce an overload that doesn't require a transaction parameter. I forgot to remove the transaction parameter in #2139. Fixes #2325.
- Remove outdated comment in `StorageView` saying that views are bound to a transaction. This was also forgotten in #2139.
- Make `FluidStorage.GENERAL_COMBINED_PROVIDER` final. This was an oversight, there is no reason for the field not to be final.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
@@ -19,8 +19,12 @@ package net.fabricmc.fabric.api.transfer.v1.fluid;
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.fluid.Fluids;
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
 import net.minecraft.sound.SoundCategory;
 import net.minecraft.sound.SoundEvent;
+import net.minecraft.sound.SoundEvents;
 import net.minecraft.util.Hand;
 
 import net.fabricmc.fabric.api.transfer.v1.context.ContainerItemContext;
@@ -57,10 +61,11 @@ public final class FluidStorageUtil {
 		if (handStorage == null) return false;
 
 		// Try to fill hand first, otherwise try to empty it.
-		return moveWithSound(storage, handStorage, player, true) || moveWithSound(handStorage, storage, player, false);
+		Item handItem = player.getStackInHand(hand).getItem();
+		return moveWithSound(storage, handStorage, player, true, handItem) || moveWithSound(handStorage, storage, player, false, handItem);
 	}
 
-	private static boolean moveWithSound(Storage<FluidVariant> from, Storage<FluidVariant> to, PlayerEntity player, boolean fill) {
+	private static boolean moveWithSound(Storage<FluidVariant> from, Storage<FluidVariant> to, PlayerEntity player, boolean fill, Item handItem) {
 		for (StorageView<FluidVariant> view : from) {
 			if (view.isResourceBlank()) continue;
 			FluidVariant resource = view.getResource();
@@ -81,6 +86,14 @@ public final class FluidStorageUtil {
 					transferTransaction.commit();
 
 					SoundEvent sound = fill ? FluidVariantAttributes.getFillSound(resource) : FluidVariantAttributes.getEmptySound(resource);
+
+					// Temporary workaround to use the correct sound for water bottles.
+					// TODO: Look into providing a proper item-aware fluid sound API.
+					if (resource.isOf(Fluids.WATER)) {
+						if (fill && handItem == Items.GLASS_BOTTLE) sound = SoundEvents.ITEM_BOTTLE_FILL;
+						if (!fill && handItem == Items.POTION) sound = SoundEvents.ITEM_BOTTLE_EMPTY;
+					}
+
 					player.playSound(sound, SoundCategory.BLOCKS, 1, 1);
 
 					return true;
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.transfer.v1.fluid;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.sound.SoundCategory;
+import net.minecraft.sound.SoundEvent;
+import net.minecraft.util.Hand;
+
+import net.fabricmc.fabric.api.transfer.v1.context.ContainerItemContext;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+import net.fabricmc.fabric.api.transfer.v1.storage.StorageView;
+import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
+
+/**
+ * Helper functions to work with fluid storages.
+ *
+ * <p><b>Experimental feature</b>, we reserve the right to remove or change it without further notice.
+ * The transfer API is a complex addition, and we want to be able to correct possible design mistakes.
+ */
+@ApiStatus.Experimental
+public final class FluidStorageUtil {
+	/**
+	 * Try to make the item in a player hand "interact" with a fluid storage.
+	 * This can be used when a player right-clicks a tank, for example.
+	 *
+	 * <p>More specifically, this function tries to find a fluid storing item in the player's hand.
+	 * Then, it tries to fill that item from the storage. If that fails, it tries to fill the storage from that item.
+	 *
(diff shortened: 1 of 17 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
```diff
@@ -0,0 +1,96 @@
...
+
+			try (Transaction transferTransaction = Transaction.openOuter()) {
+				// check how much can be inserted
+				long accepted = to.insert(resource, maxExtracted, transferTransaction);
+
+				// extract it, or rollback if the amounts don't match
+				if (accepted > 0 && view.extract(resource, accepted, transferTransaction) == accepted) {
+					transferTransaction.commit();
+
+					SoundEvent sound = fill ? FluidVariantAttributes.getFillSound(resource) : FluidVariantAttributes.getEmptySound(resource);
```
Juuxel, line 88: Hmm, this doesn't take into account that the item might have sounds of its own (in vanilla for example, bottles have different sounds than buckets when you interact with cauldrons)

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/TransferVariant.java
```diff
@@ -92,6 +92,14 @@ default NbtCompound copyNbt() {
 		return nbt == null ? null : nbt.copy();
 	}
 
+	/**
+	 * Return a copy of the tag of this variant, or a new compound if this variant doesn't have a tag.
+	 */
+	default NbtCompound copyOrCreateNbt() {
+		NbtCompound nbt = getNbt();
+		return nbt == null ? new NbtCompound() : nbt;
```
Juuxel:
```suggestion
		return nbt == null ? new NbtCompound() : nbt.copy();
```

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Technici4n: That's a general issue with #2095 then I think, I wasn't aware of that.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/TransferVariant.java
Technici4n: I'm an idiot. 😓

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Technici4n: I don't think there is much that we can do about this. Imo it is a "cosmetic" oversight (i.e. it's not essential), and addressing it would require too much effort.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/InsertionOnlyStorage.java
```diff
@@ -38,4 +42,9 @@ default boolean supportsExtraction() {
 	default long extract(T resource, long maxAmount, TransactionContext transaction) {
 		return 0;
 	}
+
+	@Override
+	default Iterator<StorageView<T>> iterator() {
```
Technici4n, line 47: The overrides in `TrashingStorage` (in testmod) and `TopStorage` (in `ComposterWrapper`) can now be removed.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Juuxel: I don't think it would be too much - or barely any - effort to add two `FluidVariantAttribute` methods (fill/empty sound) that accept an `ItemStack` and default to the non-item stack overloads.

Technici4n: Well, you'd expect a modded bottle of some kind to make the same sound as a vanilla bottle, so you'd need some way to group sounds together (I think).

Juuxel: The only way to solve that is an event design, but working for vanilla bottles at the very least would be good.

IMO interacting with a modded tank should be able to sound the same as interacting with cauldrons without having to [hardcode bottles](https://github.com/Juuxel/Adorn/blob/2aaf03632b5b1679a8b6f0dce76e658ab107856e/common/src/main/kotlin/juuxel/adorn/block/entity/KitchenSinkBlockEntity.kt#L69-L71).

Technici4n: Hmmm yeah, hardcoding bottles is not ideal.

modmuss50 (approved):
👍

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Technici4n: Hardcoding bottles for now until a better solution is designed.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
```diff
@@ -52,6 +56,13 @@ public static void register(Fluid fluid, FluidVariantAttributeHandler handler) {
 		}
 	}
 
+	/**
+	 * Enable blue- and red-colored names for water and lava respectively.
+	 */
+	public static void enableColoredVanillaFluidNames() {
+		coloredVanillaFluidNames = true;
+	}
```
Juuxel, line 67: Is this kind of enable-only behaviour really in scope for Fabric API? It feels like it should be a config option at the very least, so modpacks could disable it if they want to.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -159,14 +159,19 @@ default long simulateExtract(T resource, long maxAmount, @Nullable TransactionCo
...
+		return null;
+	}
+
+	/**
+	 * @deprecated Use and implement the overload without the transaction parameter.
+	 */
+	@Deprecated(forRemoval = true)
+	@Nullable
 	default StorageView<T> exactView(TransactionContext transaction, T resource) {
 		return null;
```
Juuxel: Should this delegate to the new overload now?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/base/SingleFluidStorage.java
```diff
@@ -0,0 +1,74 @@
...
+			}
+		};
+	}
+
+	@Override
+	protected final FluidVariant getBlankVariant() {
+		return FluidVariant.blank();
+	}
+
+	public void readNbt(NbtCompound compound) {
```
Juuxel: Should have jd that explains it uses the same format as the writeNbt from the other class. The parameter should also be `nbt` instead of `compound` imo

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/base/SingleItemStorage.java
```diff
@@ -0,0 +1,48 @@
...
+ * The transfer API is a complex addition, and we want to be able to correct possible design mistakes.
+ */
+@ApiStatus.Experimental
+public abstract class SingleItemStorage extends SingleVariantStorage<ItemVariant> {
+	@Override
+	protected final ItemVariant getBlankVariant() {
+		return ItemVariant.blank();
+	}
+
+	public void readNbt(NbtCompound compound) {
```
Juuxel: Same here

Juuxel (changes requested):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributes.java
Technici4n: It exists the replace a mixin I have in MI, as the behavior is not achievable otherwise because fabric already defines a `FluidVariantRenderHandler`. Mods can make this configurable if they want to by not calling the function.

PR #2377, opened 2022-07-03
Add development guidelines (CONTRIBUTING.md)

Technici4n:
**[Rendered View](https://github.com/Technici4n/fabric/blob/development-guidelines/CONTRIBUTING.md)**

This PR adds a `CONTRIBUTING.md` document encoding Fabric API development guidelines. This will ensure that everyone can check what our contribution standards are. The goal is to ultimately make writing and reviewing Fabric PRs a better experience.

This document has been in the works for a long time. I want to give a special thanks to @i509VCB and @Vaerian for their important contributions to the draft that was made over a year ago, and the rest of team who provided valuable feedback.

I would like to ask those who are interested in contributing to Fabric API and have some time to read this document, and comment on this PR if something is not clear.

Note that there are still a few TODO items, I will be addressing them over the coming days.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,7 +94,6 @@ Fabric API makes strong backwards compatibility guarantees, by which contributor
     - This pattern may be used if the backing implementation may be changed by another mod (such as the renderer api). Even then there should be static methods to access the standard api facilities.
 - Avoid unnecessary use of generics.
     - Unless Vanilla Minecraft mandates the use of generics, or a good reason exists to use generics.
-    - In particular, use of self types should be avoided.
 - Avoid optionals in return values, fields and parameters.
     - Where possible, you should prefer a `@Nullable T`.
     - If vanilla exposes optionals in return types, then returning an optional is fine.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 This document describes the development guidelines for Fabric API. It may be amended at any time. Therefore you should refer to the development guidelines when working on any contributions.
 
-Following these guidelines should ensure your contributions to Fabric API are quick to review, consistent with other code in Fabric API and well thought out. This document should not be seen completely as a strict ruleset but instead the thought process that a Fabric API maintainer would consider during the design, implementation and review of contributions to Fabric API.
+Following these guidelines should ensure your contributions to Fabric API are quick to review, consistent with other code in Fabric API and well thought out. This document should not be seen completely as a strict ruleset but instead the thought process that a Fabric team member would consider during the design, implementation and review of contributions to Fabric API.
 
 Old code or parts thereof might not yet be up to the standards defined by these guidelines. When working with old code, try to adhere to these guidelines, but don't bulk update legacy code to match them. The team will handle updating older code to match the newer standards when appropriate.
 
@@ -23,46 +23,37 @@ It is highly recommended that an issue be opened or a message be posted in the o
 
 The rest of this document is split in the following categories:
 
-**General design considerations**: Broad guidelines to keep in mind when writing APIs for Fabric.
-
-**API conventions**: Guidelines for common API patterns.
-
-**Implementation guidelines**: Guidelines to keep in mind when writing implementation code.
-
-**Documentation**: Guidelines for writing documentation. Extensive documentation of the offered features is essential for usability.
-
-**Structure of Fabric API**: Organization of code within Fabric API.
-
-**Code formatting**: Specific Java code formatting standards.
-
-**Testing**: Guidelines for writing testmod code. Tests showing that the submitted feature is working should be included in the PR.
-
-**Pull Request checklist**: Smaller things to keep in mind when submitting a Pull Request.
-
-TODO: once on github, check if the formatting still matches and try to link to the relevant sections
+- **General design considerations**: Broad guidelines to keep in mind when writing APIs for Fabric.
+- **API conventions**: Guidelines for common API patterns.
+- **Implementation guidelines**: Guidelines to keep in mind when writing implementation code.
+- **Documentation**: Guidelines for writing documentation. Extensive documentation of the offered features is essential for usability.
+- **Structure of Fabric API**: Organization of code within Fabric API.
+- **Code formatting**: Specific Java code formatting standards.
+- **Testing**: Guidelines for writing testmod code. Tests showing that the submitted feature is working should be included in the PR.
+- **Pull Request checklist**: Smaller things to keep in mind when submitting a Pull Request.
 
 ## General design considerations
 
 When designing an API addition, the following goals should be kept in mind:
 - Simplicity: Additions should be
-    - easy to use
-    - inherently hard to misuse or protected against misuse
-    - reasonably self contained
-    - sufficiently capable, yet not excessively loaded with niche features
-    - suitable for simple implementations
-    - not overly abstract
-- Familiarity: Additions should reuse or extend existing designs/patterns in other parts of Fabric API or vanilla
-- Extensibility: Additions should be open for future expansion without needing to deprecate anything
+    - easy to use,
+    - inherently hard to misuse or protected against misuse,
+    - reasonably self contained,
+    - sufficiently capable, yet not excessively loaded with niche features,
+    - suitable for simple implementations,
+    - not overly abstract.
+- Familiarity: Additions should reuse or extend existing designs/patterns in other parts of Fabric API or vanilla.
+- Extensibility: Additions should be open for future expansion without needing to deprecate anything.
 - Portability: Additions should not be too closely coupled with specific vanilla code. They should
-    - avoid exposure of unimportant aspects
-    - project into the future and consider potential future vanilla development
-    - avoid auxiliary libraries
-    - be portable for the sake of simplicity and extensibility
+    - avoid exposure of unimportant aspects,
+    - project into the future and consider potential future vanilla development,
+    - avoid auxiliary libraries,
+    - be portable for the sake of simplicity and extensibility.
 - Performance: Additions should
-    - be fast to initialize and execute
-    - have low or zero allocation rate
-    - have low resident memory use
-    - keep extreme uses in mind (nuke, bulk command execution, spamming etc)
+    - be fast to initialize and execute,
+    - have low or zero allocation rate,
+    - have low resident memory use,
+    - keep extreme uses in mind (nuke, bulk command execution, spamming etc).
 
 One important consideration when designing an API is spending some time thinking about the API in isolation without any influence from underlying implementation and vanilla code. Considering the API as a single entity rather than a part of a larger implementation, akin a regular user of the API may help find missing coverage, bad return values or parts of the API that could be improved.
 
@@ -115,23 +106,22 @@ Fabric API makes strong backwards compatibility guarantees, by which contributor
 - Transitive Access Wideners (TAWs) should be used to expose access to private or protected members in vanilla classes.
     - Most TAWs should go in the dedicated `fabric-transitive-access-wideners-v1` module.
         - Remember to add the `transitive-` prefix, otherwise dependent mods will not see the access modifications.
-        - Some TAWs such as Block subclass constructors are automatically generated. Make sure that you don't edit the `.accesswidener` file directly. Rather edit the template file (`.accesswidener.template`) and run the `gradlew generateAccessWidener` to update the generated file.
+        - Some TAWs such as Block subclass constructors are automatically generated. Make sure that you don't edit the `.accesswidener` file directly. Rather edit the template file (`.accesswidener.template`) and run the `gradlew generateAccessWidener` task to update the generated file.
         - Large amounts of TAWs for a specific purpose can be included in another module, as is the case for the data generation API, for example.
     - Do **not** expose TAWs for functions that take a `String` identifier.
-        - This makes it too easy to forget the mod ID namespace, so the ID would often end up in the vanilla `minecraft` namespace.
+        - This makes it too easy to forget the mod ID namespace, so the identifier would often end up in the vanilla `minecraft` namespace.
         - In general, keep the API guidelines in mind when deciding whether something should be a TAW.
 - Interface injection (i.e. making a minecraft class or interface extend a Fabric interface) should be considered over separate static helpers.
-    - Interface injection requires both a `fabric.mod.json` custom value, and a mixin to actually implement the interface at runtime.
+    - Interface injection requires both a `fabric.mod.json` custom value to make it visible in Minecraft source code, and a mixin to actually implement the interface at runtime.
     - Injected interfaces should have **no abstract methods**.
         - Methods that are guaranteed to be implemented via a mixin to a vanilla class should contain a default body that throws an error. For example:
         ```java
-        default injectedMethod() {
+        default void injectedMethod() {
 		    throw new UnsupportedOperationException("Implemented via mixin");
         }
         ```
         - Never use interface injection to add methods that modders must implement. Rather define a subclass or subinterface in Fabric API.
 - Builders can be used instead of constructors or factory methods with large amounts of parameters.
-    - TODO: any specific builder guidelines?
 
 ### API class modifiers and member visibility
 
@@ -214,7 +204,7 @@ public final class FooEvents {
     - The methods should be named in line with the action of the event, such as `entryAdded(...)`.
     - Method names for notification events should be prefixed with `on`.
       For example, a `DataLoad` event would have an `onDataLoad` method.
-    - Events that may allow or block some action should start with `ALLOW_`. For example, an event to cancel player death might be called `ALLOW_PLAYER_DEATH`.
+    - Events that may allow or block some action should start with `Allow`. For example, an event to cancel player death might be called `AllowPlayerDeath`, with method name `allowDeath`.
 - `Event<>` fields and methods.
     - The field or method exposing the `Event<>` object should be named similarly to the callback interface.
 - The `fabric-lifecycle-events-v1` module is a good example of event naming standards.
@@ -230,7 +220,7 @@ public final class FooEvents {
 
 ### Simplicity
 
-- Simple code that is easy to debug and reason is generally preferable to the shortest possible implementation.
+- Simple code that is easy to debug and reason about is generally preferable to the shortest possible implementation.
 - Limited duplication can be better than indirection, unless the code is complex or used several times.
 - Indirections might make the code harder to read, and should be weighed against their benefits. Examples include:
     - Lambda composition (forEach, streams).
@@ -316,11 +306,8 @@ Do not increment versions when writing a pull request. Version increments will b
 Every module should in its `fabric.mod.json` declare dependencies for:
 - `"fabricloader": ">=x.y.z"`, where `x.y.z` is the version used at the time the module is added.
 - Other used modules, for example `"fabric-api-base": "*"` if events are used. No explicit version needs to be specified.
-- TODO what about minecraft and java version dependencies?
+- Minecraft and Java version dependencies do not need to be specified.
 - In general, version ranges in module dependencies should be optimistic, omitting an upper bound until it is known.
-
-Every module should declare dependencies for fabric-loader no less than `>=0.6.1`, for any other used modules like fabric-api-base for events and for minecraft if it uses any type in MC or one of its libraries.
-TODO: should update this, we're using star dependencies most of the time!
 
 
 ### Packages
@@ -337,19 +324,19 @@ Inside the relevant sourceset, all Fabric API code should be in a subpackage of 
 - For API only: add the module major version with a v prefix, for example `.v1`
 - Further subpackages can be added as needed, all singular.
 
-TODO: add a few examples?
+A good example is the Lifecycle Events (v1) module.
 
 
 ### Mixins
 
 These guidelines should be followed with regards to a mixin's visibility and naming:
 
-| | Accessors | Mixin |
-| -------- | -------- | -------- |
-| Naming   | **Target**Accessor | **Target**Mixin (May include `Client/Server` or `Legacy` prefix if needed) |
-| Visibility | public | package-private* |
+|            | Accessors          | Mixin                                                                      |
+|------------|--------------------|----------------------------------------------------------------------------|
+| Naming     | **Target**Accessor | **Target**Mixin (May include `Client/Server` or `Legacy` prefix if needed) |
+| Visibility | public             | package-private*                                                           |
 
-\* A mixin may be public if a sub class extends a super mixin in a different package. Example: `abstract class ServerWorldMixin extends WorldMixin`
+\* A mixin may be public if a subclass extends a super mixin in a different package. Example: `abstract class ServerWorldMixin extends WorldMixin`
 
 The organization of mixins with a package is dependent on the type of module.
 - Generally if there are a small amount of mixins, then having all mixins in the same package is fine.
@@ -385,15 +372,23 @@ These are less strict guidelines for the code style, but you should generally ob
   - Generally keep consistent with the current standard inside of the class you are editing.
 
 ## Testing
-TODO: this section needs updating for gametests
+Including testing code when submitting new features is essential, both to demonstrate the feature, and to ensure that it works correctly.
+Testing code should not be in the regular source sets, it belongs to so-called "test mods".
 
-Test mods should provide the necessary content to debug an issue with the api manually. If possible, it should also try to verify the api+implementation is working - especially after migrating to other Minecraft versions. Test mods can also be helpful in order to illustrate how to use a module as a fallback for wiki pages that have yet to be written. Testmods are located in the `testmod` source set. Example: `fabric-lifecycle-events-v1/src/testmod/`
+- Test mods are located in the `testmod` source set. Example: `fabric-lifecycle-events-v1/src/testmod/`
+- Test mods should provide the necessary content to debug an issue with the api manually.
+  - Test mods can also be helpful in order to illustrate how to use a module as a fallback for wiki pages that have yet to be written.
+- If a test mod can be partially automated then it is encouraged to implement an automatically failing test if something goes wrong.
+  - This allows issues with the implementation or porting issues to be detected immediately.  
+  - For example, if a test mod checks if commands are registered on the server and the commands are not on the `CommandDispatcher` then the test should fail. 
+  - Good places to run the automated checks include:
+    - The mod initializer if applicable.
+    - In a listener for `ServerLifecycleEvents.SERVER_STARTED` if a server instance is required.
+    - A game test.
+  - Test failures should always throw an `AssertionError` explaining what condition was not met during the test.
 
-If a testmod can be partially automated then it is encouraged to implement an automatically failing test if something goes wrong. For example if a testmod checks if commands are registered on the server and the commands are not on the CommandDispatcher then the test should fail.
-
-Test failures should always throw an `AssertionError` explaining what condition was not met during the test.
-
-Fabric API pull requests should be tested in the dev environment and in production (on both a client and dedicated server). TODO: reminder of how to build
+Fabric API pull requests should be tested in the dev environment and in production (on both a client and dedicated server).
+The `gradlew build` command can be used to produce the Fabric API fatjar, located in `builds/libs/`.
 
 #### Common Mistakes
 One highly likely cause of a production failure is the use of `remap=false` in a mixin. If `remap=false` is used, you need to verify the mixin works in dev and production. Most likely the mixin will not work in production.
@@ -425,4 +420,4 @@ The supported values for the `ModuleLifecycle` are:
 - `"deprecated"`
 
 ### Testing
-See the Testing section above. Make sure to describe in the pull request body how you tested your code, and include a relevant testmod with your pull request.
+See the Testing section above. Make sure to describe in the pull request body how you tested your code, and include relevant test mod code with your pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,14 +23,14 @@ It is highly recommended that an issue be opened or a message be posted in the o
 
 The rest of this document is split in the following categories:
 
-- **General design considerations**: Broad guidelines to keep in mind when writing APIs for Fabric.
-- **API conventions**: Guidelines for common API patterns.
-- **Implementation guidelines**: Guidelines to keep in mind when writing implementation code.
-- **Documentation**: Guidelines for writing documentation. Extensive documentation of the offered features is essential for usability.
-- **Structure of Fabric API**: Organization of code within Fabric API.
-- **Code formatting**: Specific Java code formatting standards.
-- **Testing**: Guidelines for writing testmod code. Tests showing that the submitted feature is working should be included in the PR.
-- **Pull Request checklist**: Smaller things to keep in mind when submitting a Pull Request.
+- [**General design considerations**](#general-design-considerations): Broad guidelines to keep in mind when writing APIs for Fabric.
+- [**API conventions**](#api-conventions): Guidelines for common API patterns.
+- [**Implementation guidelines**](#implementation-guidelines): Guidelines to keep in mind when writing implementation code.
+- [**Documentation**](#documentation): Guidelines for writing documentation. Extensive documentation of the offered features is essential for usability.
+- [**Structure of Fabric API**](#structure-of-fabric-api): Organization of code within Fabric API.
+- [**Code formatting**](#code-formatting): Specific Java code formatting standards.
+- [**Testing**](#testing): Guidelines for writing testmod code. Tests showing that the submitted feature is working should be included in the PR.
+- [**Pull Request checklist**](#checklist-before-submitting-a-pull-request): Smaller things to keep in mind when submitting a Pull Request.
 
 ## General design considerations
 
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):
Great start, this is something we can contiune to add to and adapt as required. 👍

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Experimental modules might be broken if necessary.
+    - Changes that affect compilation but not the resulting bytecode are allowed in experimental modules, for example some generic changes.
+    - Even if an API is marked as experimental, avoid breaking the ecosystem of mods using it.
+        - Changes that will cause hard crashes in many mods when they are released should be delayed until the next majorly breaking Minecraft version update.
+- There might not be a way around a breaking change.
+    - If a major oversight is noticed in an API, it might be broken to address the oversight. Such cases require careful consideration, and should always be discussed with the team beforehand.
+    - Remember that deprecating code is always an option.
+
+### Targeted Minecraft versions
+- Fabric API does not have strict Minecraft version support policies, but rathers supports what is feasible and what the community is interested in.
```
apple502j:
```suggestion
- Fabric API does not have strict Minecraft version support policies, but rather supports what is feasible and what the community is interested in.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+        ```java
+        default injectedMethod() {
+		    throw new UnsupportedOperationException("Implemented via mixin");
+        }
+        ```
+        - Never use interface injection to add methods that modders must implement. Rather define a subclass or subinterface in Fabric API.
+- Builders can be used instead of constructors or factory methods with large amounts of parameters.
+    - TODO: any specific builder guidelines?
+
+### API class modifiers and member visiblity
```
apple502j:
```suggestion
### API class modifiers and member visibility
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Add module name subpackage. It might contain multiple parts separated by `.`.
+- For API only: add the module major version with a v prefix, for example `.v1`
+- Further subpackages can be added as needed, all singular.
+
+TODO: add a few examples?
+
+
+### Mixins
+
+These guidelines should be followed with regards to a mixin's visiblity and naming:
```
apple502j:
```suggestion
These guidelines should be followed with regards to a mixin's visibility and naming:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+
+| | Accessors | Mixin |
+| -------- | -------- | -------- |
+| Naming   | **Target**Accessor | **Target**Mixin (May include `Client/Server` or `Legacy` prefix if needed) |
+| Visiblity | public | package-private* |
```
apple502j:
```suggestion
| Visibility | public | package-private* |
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Limited duplication can be better than indirection, unless the code is complex or used several times.
+- Indirections might make the code harder to read, and should be weighed against their benefits. Examples include:
+    - Lambda composition (forEach, streams).
+    - Method calls.
+    - Method splitting.
+    - Recursion.
+    - Complex class hierarchies.
+- Functional operators (i.e. interfaces) should be designed for maximal readability, and only used when necessary.
+    - Ask for concrete objects if applicable.
+    - Don't include superflous methods, for example replace shouldApply+apply with a single apply that returns success.
```
apple502j:
```suggestion
    - Don't include superfluous methods, for example replace shouldApply+apply with a single apply that returns success.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Use the weakest suitable interface or class when exposing anything without harming expected uses.
+    - However, prefer `Collection` over `Iterable`.
+- Non-trivial reused (directly or derived) inline constants should be moved to static final fields.
+- Javac shouldn't produce any warnings in its default configuration.
+
+### Thread safety
+
+Since Minecraft has two primary threads, the render and server threads, APIs dealing with shared state need to consider thread safety. This will usually require consideration on a case-by-case basis.
+
+- A cache shared between the render and the server threads may for example wish to use locking for unfrequent accesses, or otherwise a copy-on-write strategy to ensure lock-free reads and thread-safe writes.
```
apple502j:
```suggestion
- A cache shared between the render and the server threads may for example wish to use locking for infrequent accesses, or otherwise a copy-on-write strategy to ensure lock-free reads and thread-safe writes.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Note that experimental modules should be in the root folder of the Fabric API repository, as they are expected to be stabilized eventually.
+
+### Versioning, dependencies
+
+The initial version of a module should **always** be `1.0.0`, never 0.x for non-legacy modules.
+Do not increment versions when writing a pull request. Version increments will be applied by maintainers after the pull request is merged.
+
+Every module should in its `fabric.mod.json` declare dependencies for:
+- `"fabricloader": ">=x.y.z"`, where `x.y.z` is the version used at the time the module is added.
+- Other used modules, for example `"fabric-api-base": "*"` if events are used. No explicit version needs to be specificed.
```
apple502j:
```suggestion
- Other used modules, for example `"fabric-api-base": "*"` if events are used. No explicit version needs to be specified.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+| -------- | -------- | -------- |
+| Naming   | **Target**Accessor | **Target**Mixin (May include `Client/Server` or `Legacy` prefix if needed) |
+| Visiblity | public | package-private* |
+
+\* A mixin may be public if a sub class extends a super mixin in a different package. Example: `abstract class ServerWorldMixin extends WorldMixin`
+
+The organization of mixins with a package is dependent on the type of module.
+- Generally if there are a small amount of mixins, then having all mixins in the same package is fine.
+- Client only and dedicated server only mixins should be moved into `client/server` subpackages.
+- If a module is generally complex or has multiple distinct parts, multiple mixins for each class target may exist assuming said mixins are seperated into subpackages by feature.
```
apple502j:
```suggestion
- If a module is generally complex or has multiple distinct parts, multiple mixins for each class target may exist assuming said mixins are separated into subpackages by feature.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+- Single space after `if`, `for`, `do`, `break` unless followed by `;`.
+- No space after `(` or before `)`.
+- Statements should use the block form except for single-line if statements that use a short condition and nothing else.
+- Ternary operator use only within a single line total.
+
+Fabric API uses the Gradle checkstyle plugin. If you run `gradle check` any style errors will be logged in the output. The style errors can also be viewed in a generated webpage by the gradle task.
+
+### Style guidelines
+
+These are less strict guildelines for the code style, but you should generally obey these:
```
apple502j:
```suggestion
These are less strict guidelines for the code style, but you should generally obey these:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,428 @@
...
+    - These examples should be written for developers that have a knowledge of java and basic knowledge of the Minecraft codebase.
+- A brief description belongs in the 1st paragraph, with further paragraphs separated by blank lines and starting with `<p>`. Javadoc for methods may follow with another blank line before describing parameters, return values and exceptions with the appropriate tags. All accessible members should be described appropriately.
+- Good documentation doesn't only explain what something is, but (as appropriate) why it exists, what are the intended use cases, what use cases something is not suitable for and any semantics that need to be kept in mind.
+    - In particular, tricky implementation details should be explained by a few comments when appropriate.
+- Direct references to classes or members should use `@link` tag, unspecific further reading elsewhere may use `@see`.
+- Deprecated elements should describe and reference the replacement or alternative in Javadoc ``@deprecated`. The ``@deprecated` tag has to be last, everything else may be removed while adding it.
```
apple502j:
```suggestion
- Deprecated elements should describe and reference the replacement or alternative in Javadoc `@deprecated`. The `@deprecated` tag has to be last, everything else may be removed while adding it.
```

apple502j (changes requested):

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+
+Fabric API generally is not meant for bugfixes, performance improvements or gameplay changes, and exceptions need careful evaluation.
+
+It is highly recommended that an issue be opened or a message be posted in the official Discord server, to discuss whether a feature is in scope, before it is designed or implemented. Design choices may also warrant prior discussion to avoid wasting time.
+
+## Structure of the guidelines
+
+The rest of this document is split in the following categories:
+
+- **General design considerations**: Broad guidelines to keep in mind when writing APIs for Fabric.
```
liach:
```suggestion
- [**General design considerations**](#general-design-considerations): Broad guidelines to keep in mind when writing APIs for Fabric.
```
we can add such links, same for other categories below,

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+    - However, prefer `Collection` over `Iterable`.
+- Non-trivial reused (directly or derived) inline constants should be moved to static final fields.
+- Javac shouldn't produce any warnings in its default configuration.
+
+### Thread safety
+
+Since Minecraft has two primary threads, the render and server threads, APIs dealing with shared state need to consider thread safety. This will usually require consideration on a case-by-case basis.
+
+- A cache shared between the render and the server threads may for example wish to use locking for infrequent accesses, or otherwise a copy-on-write strategy to ensure lock-free reads and thread-safe writes.
+- Read-only access from multiple threads does not need synchronization.
```
liach, line 268: I hope you add another entry for lazy initialization (like benign race), such as:
```java
private Stuff cachedStuff;

private Stuff stuff() {
  var stuff = this.cachedStuff;
  if (stuff != null) return stuff;
  return this.cachedStuff = compute();
}
```

Frequently, people will write code like
```java
private Stuff stuff() {
  if (this.cachedStuff != null) return this.cachedStuff;
  return this.cachedStuff = compute();
}
```
which reads `cachedStuff` field twice and is subject to races, as the first read may be non-null but the second one may.

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+    - They should be placed at the very bottom of the class to not hurt readability.
+- Access modifiers for fields and methods should be as strict as possible.
+    - If a method is intended to only be for use by mods implementing an api, a `protected` method should suffice.
+
+### Annotations
+- Nullable members, parameters or return values must be annotated with the `@Nullable` annotation.
+    - Any member, parameter or return value that is not marked as nullable can be assumed to be nonnull. The `@NotNull` annotation should never be used, it is implicit.
+    - Introducing custom types might be appropriate (no need to specify `List<@Nullable String>` everywhere).
+- Deprecated API members should have the `@Deprecated` annotation.
+    - Avoid specifying `forRemoval = true`, the functionality will be supported as long as possible.
```
liach, line 141: Not that we usually intend to support such deprecated APIs; the true intention is that we usually sunset an API with a whole module, and we only use `forRemoval = true` if we know this API must be removed as the current module lives and before we move to a new module, which is exceedingly rare.

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+- Module names should usually be suffixed by `-api`.
+    - Modules whose primary purpose is not interaction with their API do not need this suffix. For example, `fabric-transitive-access-wideners-v1` or `fabric-convention-tags-v1`.
+- Module names should always be suffixed by a major version (`-v1`, `-v2`, etc).
+    - The major version starts at `v1` for new functionality, unless they replace a module with equivalent functionality, in which case the version is incremented.
+    - `vn` and `vn+1` module names need not match exactly. For example, `fabric-loot-table-api-v1` was replaced by `fabric-loot-api-v2`.
+
+### Deprecated modules
+Modules that are entirely deprecated are in the `deprecated/` folder. All of their API classes should be deprecated, and their `fabric.mod.json` should have the deprecated lifecycle marker. (See PR checklist below).
+
+A module may only be deprecated if all of its functionality is also provided by a non-deprecated module.
```
liach: ... or those non-migrated functionalities are determined to be no longer feasible

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Experimental modules
+Modules whose design is hard to evaluate might go through an experimental phase, allowing for relaxed backwards compatibility requirements, as described in the Backwards compatibility section.
+
+Writers of experimental modules need to consider the following additional requirements:
+- All API classes should be marked as `@ApiStatus.Experimental`. Do not use `@Deprecated` for experimental modules.
```
liach:
```suggestion
- All API classes should be marked as `@ApiStatus.Experimental`. Do not use `@Deprecated` for experimental modules to generate compiler warnings.
```

CONTRIBUTING.md
Technici4n: Both are correct provided that `cachedStuff` is `volatile`. The reason we use an intermediate variable is to avoid two `volatile` reads, which are more expensive than a single `volatile` read. I'll add something in that area.

apple502j: Are there actually APIs where this matters, btw?

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+    - If vanilla exposes optionals in return types, then returning an optional is fine.
+- Avoid requiring the user to cast to a subtype if possible.
+    - Adding methods to vanilla types can be done via interface injection.
+
+### API design patterns to consider
+
+- Transitive Access Wideners (TAWs) should be used to expose access to private or protected members in vanilla classes.
+    - Most TAWs should go in the dedicated `fabric-transitive-access-wideners-v1` module.
+        - Remember to add the `transitive-` prefix, otherwise dependent mods will not see the access modifications.
+        - Some TAWs such as Block subclass constructors are automatically generated. Make sure that you don't edit the `.accesswidener` file directly. Rather edit the template file (`.accesswidener.template`) and run the `gradlew generateAccessWidener` task to update the generated file.
```
Juuxel:
```suggestion
        - Some TAWs such as Block subclass constructors are automatically generated. Make sure that you don't edit the `.accesswidener` file in `src` directly. Rather edit the template file (`template.accesswidener`) and run the `gradlew generateAccessWidener` task to update the generated file.
```
See the file name here: https://github.com/FabricMC/fabric/blob/1.19/fabric-transitive-access-wideners-v1/template.accesswidener

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+        - Large amounts of TAWs for a specific purpose can be included in another module, as is the case for the data generation API, for example.
+    - Do **not** expose TAWs for functions that take a `String` identifier.
+        - This makes it too easy to forget the mod ID namespace, so the identifier would often end up in the vanilla `minecraft` namespace.
+        - In general, keep the API guidelines in mind when deciding whether something should be a TAW.
```
Juuxel:
```suggestion
      - In general, keep the API guidelines in mind when deciding whether something should be a TAW.
```
This isn't really specific to string IDs

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+- Interface injection (i.e. making a minecraft class or interface extend a Fabric interface) should be considered over separate static helpers.
+    - Interface injection requires both a `fabric.mod.json` custom value to make it visible in Minecraft source code, and a mixin to actually implement the interface at runtime.
+    - Injected interfaces should have **no abstract methods**.
+        - Methods that are guaranteed to be implemented via a mixin to a vanilla class should contain a default body that throws an error. For example:
+        ```java
+        default void injectedMethod() {
+		    throw new UnsupportedOperationException("Implemented via mixin");
+        }
+        ```
```
Juuxel: Fixed funky indentation on the middle line:
```suggestion
        ```java
        default void injectedMethod() {
        	throw new UnsupportedOperationException("Implemented via mixin");
        }
        ```
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Implementation guidelines
+
+### Simplicity
+
+- Simple code that is easy to debug and reason about is generally preferable to the shortest possible implementation.
+- Limited duplication can be better than indirection, unless the code is complex or used several times.
+- Indirections might make the code harder to read, and should be weighed against their benefits. Examples include:
+    - Lambda composition (forEach, streams).
+    - Method calls.
```
Juuxel: Method calls are perfectly normal Java code and not particularly hard to read?

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+    - Method splitting.
+    - Recursion.
+    - Complex class hierarchies.
+- Functional operators (i.e. interfaces) should be designed for maximal readability, and only used when necessary.
+    - Ask for concrete objects if applicable.
+    - Don't include superfluous methods, for example replace shouldApply+apply with a single apply that returns success.
```
Juuxel:
```suggestion
    - Don't include superfluous methods. For example, replace `shouldApply`+`apply` with a single `apply` that returns success.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+
+- A cache shared between the render and the server threads may for example wish to use locking for infrequent accesses, or otherwise a copy-on-write strategy to ensure lock-free reads and thread-safe writes.
+- Read-only access from multiple threads does not need synchronization.
+
+## Documentation
+
+- Every API class should carry a Javadoc block explaining its purpose and reference related classes. Example code in the primary class of any major feature should outline the use, including related vanilla invocations/registrations/etc as applicable to provide an idea of how to start.
+    - Parameters, implementation bodies, etc... may be omitted as appropriate, since the examples are not meant as to be fully working implementations.
+    - The examples should be more akin to pseudocode with a checklist for the process and pointers to everything needed.
+    - These examples should be written for developers that have a knowledge of java and basic knowledge of the Minecraft codebase.
```
Juuxel:
```suggestion
    - These examples should be written for developers that have a knowledge of Java and basic knowledge of the Minecraft codebase.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+Fabric API is organized in different modules. Each module is located in a specific folder matching the module id. For example under `fabric-item-api-v1/` for the Item API (v1).
+
+### Module naming conventions
+- Module names should be named after the exposed functionality.
+    - Consider future developments when naming a module: they might later be expanded.
+- Module names should usually be suffixed by `-api`.
+    - Modules whose primary purpose is not interaction with their API do not need this suffix. For example, `fabric-transitive-access-wideners-v1` or `fabric-convention-tags-v1`.
+- Module names should always be suffixed by a major version (`-v1`, `-v2`, etc).
+    - The major version starts at `v1` for new functionality, unless they replace a module with equivalent functionality, in which case the version is incremented.
+    - `vn` and `vn+1` module names need not match exactly. For example, `fabric-loot-table-api-v1` was replaced by `fabric-loot-api-v2`.
```
Juuxel:
```suggestion
    - `vn` and `vn+1` module names need not match exactly. For example, `fabric-loot-tables-v1` was replaced by `fabric-loot-api-v2`.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+- No blank lines at the start or end of a block, i.e. no blank line after `{` or before `}`.
+- Blank line after input (parameter) validation, if present.
+- New line after `{` unless empty  (`{}` or `{ }`), never before.
+- Line length usually below 120 columns, sometimes up to 160 if beneficial for readability.
+- Single space after `if`, `for`, `do`, `break` unless followed by `;`.
+- No space after `(` or before `)`.
+- Statements should use the block form except for single-line if statements that use a short condition and nothing else.
+- Ternary operator use only within a single line total.
+
+Fabric API uses the Gradle checkstyle plugin. If you run `gradle check` any style errors will be logged in the output. The style errors can also be viewed in a generated webpage by the gradle task.
```
Juuxel:
```suggestion
Fabric API uses the Gradle checkstyle plugin. If you run `gradle check` any style errors will be logged in the output. The style errors can also be viewed in a generated webpage by the Gradle task.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+Fabric API pull requests should be tested in the dev environment and in production (on both a client and dedicated server).
+The `gradlew build` command can be used to produce the Fabric API fatjar, located in `builds/libs/`.
+
+#### Common Mistakes
+One highly likely cause of a production failure is the use of `remap=false` in a mixin. If `remap=false` is used, you need to verify the mixin works in dev and production. Most likely the mixin will not work in production.
+
+## Checklist before submitting a pull request
+
+### Apply license headers
+There is a gradle task that can automate this for you. Simply run `gradlew spotlessApply`.
```
Juuxel:
```suggestion
There is a Gradle task that can automate this for you. Simply run `gradlew spotlessApply`.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+  - Good places to run the automated checks include:
+    - The mod initializer if applicable.
+    - In a listener for `ServerLifecycleEvents.SERVER_STARTED` if a server instance is required.
+    - A game test.
+  - Test failures should always throw an `AssertionError` explaining what condition was not met during the test.
+
+Fabric API pull requests should be tested in the dev environment and in production (on both a client and dedicated server).
+The `gradlew build` command can be used to produce the Fabric API fatjar, located in `builds/libs/`.
+
+#### Common Mistakes
```
Juuxel:
```suggestion
#### Common mistakes
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+        - Exceptions are made for experimental APIs, where deprecated for removal functionality might be removed after sufficient time has passed.
+    - A javadoc `@deprecated` comment should be added to deprecated members with migration instructions.
+- Experimental API classes should all have the `@ApiStatus.Experimental` annotation.
+    - The annotation is not required on individual members, unless the class itself is not experimental.
+- `@ApiStatus.NonExtendable` should be used for API interfaces or classes that modders must not implement or override, but can't be `final` for some reason.
+    - Adding methods to such interfaces or classes is not a breaking change.
+- `@ApiStatus.Internal` is recommended for implementation classes to avoid IDE autocompletion suggesting them. There should be no need for internal methods or classes in the public API package.
+
+### Naming
+- All names should follow the yarn naming standards.
```
Juuxel:
```suggestion
- All names should follow the [Yarn naming standards](https://github.com/FabricMC/yarn/blob/HEAD/CONVENTIONS.md).
```

CONTRIBUTING.md
Juuxel: Including "by a non-deprecated module **or vanilla Minecraft itself**" might be good, some loot tables v1 api was later added to the game and dropped from v2.

CONTRIBUTING.md
liach: This is a point of concern in the API implementation, especially that there are server and client threads

apple502j: I don't see APIs with this use pattern intended to be called on both sides, but guess there can be.

Technici4n: You can find it in the depths of the transfer API: https://github.com/FabricMC/fabric/blob/bdca8858a1b0d5e7885081a27d92e8c3754a9c03/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/ItemVariantImpl.java#L144-L153

It's also used in the content registries:
https://github.com/FabricMC/fabric/blob/bdca8858a1b0d5e7885081a27d92e8c3754a9c03/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FlammableBlockRegistryImpl.java#L49-L70
https://github.com/FabricMC/fabric/blob/bdca8858a1b0d5e7885081a27d92e8c3754a9c03/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FuelRegistryImpl.java#L51-L59

CONTRIBUTING.md
IMB11: Actually, you're incorrect.

According to Oxford:

> Capitalise the first word of the title, and all words within the title except articles (a/ an/the), prepositions (to/on/for etc) and conjunctions (but/and/or etc).

Technici4n: The rest of the document is written with only the first words of the titles capitalized. It looks better like this than with most words capitalized. ;)

IMB11: Fair

Juuxel: Yep, I suggested this for consistency with the rest of the document (I personally couldn't care less about which style is used as long as they're not mixed, which looks ugly)

CONTRIBUTING.md
```diff
@@ -0,0 +1,423 @@
...
+    - For example, an event that both cancels and notifies of an action will produce false notifications (notified but later canceled). It would be preferable to have an `AllowXxx` event for cancellation, and then a `BeforeXxx` event for notification.
+
+## Implementation guidelines
+
+### Simplicity
+
+- Simple code that is easy to debug and reason about is generally preferable to the shortest possible implementation.
+- Limited duplication can be better than indirection, unless the code is complex or used several times.
+- Indirections might make the code harder to read, and should be weighed against their benefits. Examples include:
+    - Lambda composition (forEach, streams).
```
Juuxel: Nitpick: `forEach` on `Iterable`s doesn't really compose anything, not does it make anything harder to read. I still agree it should be avoided for other reasons though (no `break`/`continue`, possibly slower in hot code)

CONTRIBUTING.md
Technici4n: I think that this part is fine as-is: `Avoid specifying` means that in some edge cases we can still allow it (vs `Never specify` which would always forbid it).

CONTRIBUTING.md
Technici4n: I added the note about vanilla MC.

> ... or those non-migrated functionalities are determined to be no longer feasible

I don't think this is fully accurate, as we should try to provide a reasonable alternative. After all, if the functionality was added in the first place, it was probably for a good reason. So I suggest not to add this, and we can discuss exceptions on a case-by-case basis.

CONTRIBUTING.md
Technici4n: The intent is already conveyed by `Method splitting`, so I'm removing this one.

CONTRIBUTING.md
Technici4n: Replaced with `Lambda methods`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+    - In doubt, prefer PR'ing to newer stable versions.
+- Backporting PRs for older versions will generally be accepted, depending on how many changes were required.
+
+### Discouraged API design patterns
+
+- Avoid the old `Api.INSTANCE` pattern in favor of static methods.
+    - This pattern may be used if the backing implementation may be changed by another mod (such as the renderer api). Even then there should be static methods to access the standard api facilities.
+- Avoid unnecessary use of generics.
+    - Unless Vanilla Minecraft mandates the use of generics, or a good reason exists to use generics.
+    - In particular, use of self types should be avoided.
```
Juuxel: Why?

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+        - Some TAWs such as Block subclass constructors are automatically generated. Make sure that you don't edit the `.accesswidener` file in `src` directly. Rather edit the template file (`template.accesswidener`) and run the `gradlew generateAccessWidener` task to update the generated file.
+        - Large amounts of TAWs for a specific purpose can be included in another module, as is the case for the data generation API, for example.
+    - Do **not** expose TAWs for functions that take a `String` identifier.
+        - This makes it too easy to forget the mod ID namespace, so the identifier would often end up in the vanilla `minecraft` namespace.
+    - In general, keep the API guidelines in mind when deciding whether something should be a TAW.
+- Interface injection (i.e. making a minecraft class or interface extend a Fabric interface) should be considered over separate static helpers.
+    - Interface injection requires both a `fabric.mod.json` custom value to make it visible in Minecraft source code, and a mixin to actually implement the interface at runtime.
+    - Injected interfaces should have **no abstract methods**.
```
Juuxel, line 115: Might be good to include the rationale here (that javac will throw a fit if it can't find the implementation of an itf method on a class iirc)

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+        ```
+        - Never use interface injection to add methods that modders must implement. Rather define a subclass or subinterface in Fabric API.
+- Builders can be used instead of constructors or factory methods with large amounts of parameters.
+
+### API class modifiers and member visibility
+
+- Classes in Fabric API should be `final` classes unless the class exposed in the API is explicitly meant for extension.
+- `private` constructors should be used in API classes unless the class is explicitly meant to be instantiated.
+    - This only applies to modder-facing classes, i.e. classes in the `net.fabricmc.fabric.api` subpackage. See below for package structure.
+    - They should be placed at the very bottom of the class to not hurt readability.
```
Juuxel: Placing them at the very top also doesn't hurt readability imo (JDK code does this)

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+
+### Naming
+- All names should follow the [Yarn naming standards](https://github.com/FabricMC/yarn/blob/HEAD/CONVENTIONS.md).
+- If a class only contains getter methods, the `get` prefix may be omitted for methods. The `get` prefix may also be omitted where it is not appropriate.
+- Accessor mixins should be named **Target**Accessor, other mixins should be named **Target**Mixin, where `Target` is the target class name. More details in the mixin section below.
+
+### Events
+
+- Events should not be used if there is only one subscriber, like a handler in a registered unique namespace.
+- Events should be produced and fully usable without object allocation.
```
Juuxel:
```suggestion
- Events should be produced and fully usable without creating data holder objects for inputs.
```
Events sometimes do need to allocate objects, for example when they use `TypedActionResult`

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+- Events should use dedicated callback interfaces.
+    - Callback interfaces should be `@FunctionalInterface`s.
+    - Callback methods should be uniquely named such that a handler can implement multiple at once.
+        - Avoid words that are already clear from the parameters. For example, prefer `onStartTick(MinecraftServer)` over `onStartServerTick(MinecraftServer)`. The interface should still be named `StartServerTick`.
```
Juuxel:
```suggestion
        - Avoid words that are already clear from the parameters. For example, prefer `onStartTick(MinecraftServer)` over `onStartServerTick(MinecraftServer)`. The interface should still be named `StartServerTick` if "server" is not already implied by the containing class.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+
+    // Holder class is not meant for instantiation.
+    private ExampleEvents() {
+    }
+}
+```
+
+#### Event naming
+
+- Callback interfaces.
```
Juuxel:
```suggestion
- Callback interfaces:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+    - Callback interfaces should be named using present tense. For example, `ChunkUnload` and not `ChunkUnloaded`.
+    - The methods should be named in line with the action of the event, such as `entryAdded(...)`.
+    - Method names for notification events should be prefixed with `on`.
+      For example, a `DataLoad` event would have an `onDataLoad` method.
+    - Events that may allow or block some action should start with `Allow`. For example, an event to cancel player death might be called `AllowPlayerDeath`, with method name `allowDeath`.
+- `Event<>` fields and methods.
```
Juuxel:
```suggestion
- `Event<>` fields and methods:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+    - For example, an event that both cancels and notifies of an action will produce false notifications (notified but later canceled). It would be preferable to have an `AllowXxx` event for cancellation, and then a `BeforeXxx` event for notification.
+
+## Implementation guidelines
+
+### Simplicity
+
+- Simple code that is easy to debug and reason about is generally preferable to the shortest possible implementation.
+- Limited duplication can be better than indirection, unless the code is complex or used several times.
+- Indirections might make the code harder to read, and should be weighed against their benefits. Examples include:
+    - Lambda methods (forEach, streams).
```
Juuxel:
```suggestion
    - Lambda methods (`forEach`, streams).
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    return stuff;
+}
+```
+- Read-only access from multiple threads does not need synchronization.
+
+## Documentation
+
+- Every API class should carry a Javadoc block explaining its purpose and reference related classes. Example code in the primary class of any major feature should outline the use, including related vanilla invocations/registrations/etc as applicable to provide an idea of how to start.
```
Juuxel:
```suggestion
- Every API class should carry a Javadoc comment explaining its purpose and reference related classes. Example code in the primary class of any major feature should outline the use, including related vanilla invocations/registrations/etc as applicable to provide an idea of how to start.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,437 @@
...
+    - Parameters, implementation bodies, etc... may be omitted as appropriate, since the examples are not meant as to be fully working implementations.
+    - The examples should be more akin to pseudocode with a checklist for the process and pointers to everything needed.
+    - These examples should be written for developers that have a knowledge of Java and basic knowledge of the Minecraft codebase.
+- A brief description belongs in the 1st paragraph, with further paragraphs separated by blank lines and starting with `<p>`. Javadoc for methods may follow with another blank line before describing parameters, return values and exceptions with the appropriate tags. All accessible members should be described appropriately.
+- Good documentation doesn't only explain what something is, but (as appropriate) why it exists, what are the intended use cases, what use cases something is not suitable for and any semantics that need to be kept in mind.
+    - In particular, tricky implementation details should be explained by a few comments when appropriate.
+- Direct references to classes or members should use `@link` tag, unspecific further reading elsewhere may use `@see`.
```
Juuxel:
```suggestion
- Direct references to classes or members should use the `@link` and `@linkplain` tags, unspecific further reading elsewhere may use `@see`.
```

Juuxel (commented):
This should be the last round of comments from me :tiny_potato:

CONTRIBUTING.md
Technici4n: Well, obviously we don't want to force modders to add methods beyond what vanilla requires. :P

CONTRIBUTING.md
Technici4n: It's a general design guideline. While it mainly applies for inputs, it could also apply for outputs in some cases (e.g. pass a `Consumer` parameter instead of returning a `List`).

CONTRIBUTING.md
Technici4n: Well, they lead to many problems and should be avoided for that reason.

CONTRIBUTING.md
Juuxel: Well yeah, that too. But the javac thing applies when you're injecting into a final class or a class that no one will ever want to extend (LootTable)

CONTRIBUTING.md
Juuxel: > many problems

...that doesn't really answer anything.

CONTRIBUTING.md
Juuxel: The current version reads to me like "you should never allocate objects when designing events", which does not sound like a soft general guideline

CONTRIBUTING.md
Technici4n: I don't have a specific example in mind, but I had issues when using them in the past in AE2. `IAEStack<T extends IAEStack<T>>` used to be generic hell.

CONTRIBUTING.md
Technici4n: Should be ok now.

CONTRIBUTING.md
Juuxel: Self types are fine - for example vanilla's loot builders use them without them being generic hell. Using generics (including self types) is one thing, going overboard with them is another.

Technici4n: I removed this line since it's already contained in `Avoid unnecessary use of generics.`

Juuxel (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,439 @@
...
+One highly likely cause of a production failure is the use of `remap=false` in a mixin. If `remap=false` is used, you need to verify the mixin works in dev and production. Most likely the mixin will not work in production.
+
+## Checklist before submitting a pull request
+
+### Apply license headers
+There is a Gradle task that can automate this for you. Simply run `gradlew spotlessApply`.
+
+### Run checks
+- The `gradlew check` task runs all the style checks and game tests.
+- Code style can be checked with the faster `gradlew checkstyleMain checkstyleTestmod`.
```
deirn, line 419: Maybe create a new task that run all the checkstyle tasks?
```gradle
task checkstyle {
	dependsOn tasks.withType(Checkstyle)
	subprojects.forEach {
		dependsOn it.tasks.withType(Checkstyle)
	}
}
```

Technici4n: @modmuss50 what do you think?

modmuss50: Yeah, something along those lines sounds reasonable.

PR #2381, opened 2022-07-06
Add Map-Lookup Storage Impl

Phoupraw:
Add `ExactViewStorage`, a storage abstract class that provides a variant-to-slot map.
Add `ExactViewStorage.ItemImpl`, a simple implement to `ExactViewStorage<ItemVariant>`.


Technici4n:
As discussed on Discord, I think it's still too early to include such an implementation in Fabric.

PR #2387, opened 2022-07-10
Fix custom dimension not loaded on world preset other than default

deirn:
Fix https://github.com/FabricMC/fabric/pull/2345#issuecomment-1179474099

Loads the DRM again before the integrated server starts and replace the `GeneratorOptions` with newly parsed one.
I tried capturing the DRM and `RegistryOps` instance from `create` and `applyDataPacks` lamdas and using it to parse but it only works for default preset ¯\\\_(ツ)\_/¯.

Dedicated server is not affected.

Also for yarn contributors: `GeneratorOptionsHolder` doesn't seem to be the correct name as it holds more than a `GeneratorOptions`.


apple502j (approved):

apple502j:
Would like to see this backported to 1.18.2, if applicable. Should apply cleanly to 1.19.1.

modmuss50 (approved):

PR #2388, opened 2022-07-10
Add scaledWidth and scaledHeight parameters to the HudRenderCallback

Kaupenjoe:
When using the `HudRenderCallback` usually people want to add something onto the HUD. The position of custom HUD elements is always dependent on the scaled dimensions of the screen. Currently you would have to do get the scaled dimensions through `MinecraftClient` like this:
```
MinecraftClient client = MinecraftClient.getInstance();
int width = client.getWindow().getScaledWidth();
int height = client.getWindow().getScaledHeight();
```
By having access to the scaled height and width via the method parameter, this step can be dropped and it makes the code a bit cleaner.


deirn:
You can't just add new parameter to the method, that would break all mods that uses it. You need to make a new event with different interface, I don't think it's worth it tbh.

modmuss50:
Yes, this is breaking, I dont think it needs to be passed through as you said its easy enough to access through `MinecraftClient`. So I don't see this beign needed. 👍

Kaupenjoe:
Alright, closing this, fair enough. I don't think a new interface and event is worth it for the tiny change.

PR #2390, opened 2022-07-13
Add type specific pack format to `pack.mcmeta`

### LINKED ISSUE #2151: [Feature Request] Separate pack versions in a pack.mcmeta file
In 1.18.2 data packs' pack version was bumped from 8 to 9, making the data and resource version different for the first time.
There is only one `pack_format` in a `pack.mcmeta`
```
{
  "pack": {
    "pack_format": 8,
    "description": "Fabric Resource Loader Test Builtin Pack."
  }
}
```
But we can create a pack with data and assets, so shouldn't we be able to separate data and assets in the `pack_version`?
Example:
```
{
  "pack": {
    "fabric:data_format": 9,
    "fabric:assets_format" 8,
    "description": "Fabric Resource Loader Test Builtin Pack."
  }
}
```
This issue basically mirrors https://github.com/MinecraftForge/MinecraftForge/issues/8563

deirn:
Closes #2151

Adds optional `fabric:resource_pack_format` and `fabric:data_pack_format` entry to specify client resource and datapack format respectively.


apple502j (approved):

Jab125 (approved):

Technici4n:
I wonder how useful this really is compared to having two separate builtin packs. Forge needed it because mods explicitly have to specify a pack format for the mod resources (directly in the jar), but I don't think we need it.

LambdAurora:
What about just getting a custom resource pack profile implementation and always make the pack compatible in the screen instead?

Like, that's already a thing with the mod's own resources internally, the pack format is automatically set in an always compatible way.
And considering this is for built-in resource packs primarily, to me that makes the most sense..

Technici4n:
I believe this is not useful, see my comment above.

PR #2402, opened 2022-07-18
Add original armor model to ArmorRenderer [Breaking change for 1.19.1]

AlphaMode:


modmuss50 (changes requested):
This is an API breaking change.

AlphaMode:
Already working on a work around similar to worldrender context

PR #2407, opened 2022-07-19
ModResourcePackUtil: Properly handle special chars in mod name

apple502j:
Discovered while auditing `String.format`.

`ModResourcePackUtil#openDefault` in resource loader is used to provide `pack.mcmeta` fallback when one does not exist in a registered pack. This previously constructed the file contents via `String.format`. This contains a mod's name, which can include special characters (usually a quote, but sometimes backslashes).

Note that it's not easy to properly add those special characters in `fabric.mod.json`: due to Gradle's templating system used in many setups to substitute versions, backslashes are interpreted during build-time jar processing. Of course, the build result must still be a valid JSON. Therefore, to use `\ "` as the mod's name `" \\" \\\\"` must be written instead. Since this is not well-known I expect few people would encounter this issue in the first place.

If they do, however, and this method is called, then the method is expected to return the stream of a valid JSON. However, they don't, for a couple of reasons:

1. The current escaping only considers quotes. However, trailing backslashes can cause the string-terminating quote to be escaped.
2. Actually, this escaping doesn't work at all, because of [a quirk with `String#replaceAll`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#replaceAll(java.lang.String,java.lang.String)). Either the second argument [should've been quoted](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Matcher.html#quoteReplacement(java.lang.String)) or it should've been [`String#replace` instead](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#replace(java.lang.CharSequence,java.lang.CharSequence)).

Instead of trying to fix the escaping, I just made it use Gson, which is more readable. Note that [the other code](https://github.com/FabricMC/fabric/blob/1.19/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java#L49) still uses string replacement since it's unaffected.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
@@ -77,17 +77,20 @@ public final class ModResourcePackUtil {
 		switch (filename) {
 		case "pack.mcmeta":
 			String description = Objects.requireNonNullElse(info.getName(), "");
-
-			JsonObject pack = new JsonObject();
-			pack.addProperty("pack_format", type.getPackVersion(SharedConstants.getGameVersion()));
-			pack.addProperty("description", description);
-			JsonObject metadata = new JsonObject();
-			metadata.add("pack", pack);
-
-			return IOUtils.toInputStream(GSON.toJson(metadata), Charsets.UTF_8);
+			String metadata = serializeMetadata(type.getPackVersion(SharedConstants.getGameVersion()), description);
+			return IOUtils.toInputStream(metadata, Charsets.UTF_8);
 		default:
 			return null;
 		}
+	}
+
+	public static String serializeMetadata(int packVersion, String description) {
+		JsonObject pack = new JsonObject();
+		pack.addProperty("pack_format", packVersion);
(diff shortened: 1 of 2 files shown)


apple502j:
Question: should I add a testmod for this? That'd require renaming the testmod to include special characters, which is ugly and potentially breaks stuff expecting well-formed preprocessed FMJ file.

Note that this is tested locally.

LambdAurora:
> Question: should I add a testmod for this? That'd require renaming the testmod to include special characters, which is ugly and potentially breaks stuff expecting well-formed preprocessed FMJ file.

What about splitting the method to get the string of the JSON and feed that into the JSON parser to check if it's well-formed in the testmod?

Could then feed lot of different input strings as if you were unit testing.
Or like look into bringing some kind of unit testing into Fabric API?
Feels like this method of testing is particularly adapted to this situation.

apple502j:
Yeah, that's a good idea. Maybe I'll do that, don't have that much time right now.

apple502j:
Oops, it's failing. Seems like testmod has a bug. Fixing soon

LambdAurora (approved):

Technici4n (approved):
LGTM

modmuss50 (approved):

PR #2408, opened 2022-07-20
Allow client commands to open a screen without thread schedule

deirn:
Right now to open a screen via a command you would need to use some sort of thread schedule like `Timer`:
```java
new Timer().schedule(new TimerTask() {
    @Override
    public void run() {
        client.execute(() -> client.setScreen(screen));
    }
}, 1);
```
This is caused by ChatScreen calling `client.setScreen(null)` when enter key is pressed. This PR would make that unnecessary.


Juuxel:
Fyi you don't have to use a timer, a simple `client.send` call wrapping the screen setter is enough. I don't think we really need this pr since using that method is easy enough.

deirn:
Ooh, I somehow missed it 😅 thanks, and never mind about this PR then.

PR #2409, opened 2022-07-20
Add a cancelable CrosshairRenderCallback event

ghost:
This PR adds a new event that is called before crosshair render.

Final test video: [Link](https://streamable.com/y154cf)

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/CrosshairRenderCallback.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/CrosshairRenderCallback.java
@@ -1,0 +1,37 @@
+package net.fabricmc.fabric.api.client.rendering.v1;
+
+import net.fabricmc.fabric.api.event.Event;
+
+import net.fabricmc.fabric.api.event.EventFactory;
+
+import net.minecraft.client.gui.DrawableHelper;
+import net.minecraft.client.util.math.MatrixStack;
+
+public interface CrosshairRenderCallback {
+
+	Event<CrosshairRenderCallback> EVENT = EventFactory.createArrayBacked(CrosshairRenderCallback.class, listeners -> ((matrices, scaledWidth, scaledHeight, zOffset) -> {
+		boolean result = true;
+		for (CrosshairRenderCallback callback : listeners) {
+			if (!callback.onCrosshairRender(matrices, scaledWidth, scaledHeight, zOffset)) {
+				result = false;
+			}
+		}
+		return result;
+	}));
+
+	/**
+	 * Called before crosshair is rendered.
+	 * Crosshair rendering can be canceled if returned {@code false}.
+	 *
+	 * @param matrices The MatrixStack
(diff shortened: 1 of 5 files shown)


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/CrosshairRenderCallback.java
```diff
@@ -0,0 +1,53 @@
...
+	 * @param scaledWidth Scaled width of the game window.
+	 * @param scaledHeight Scaled height of the game window.
+	 * @param zOffset The z offset
+	 * @return {@code false} if crosshair should be disabled, otherwise {@code true}.
+	 */
+	boolean onCrosshairRender(MatrixStack matrices, int scaledWidth, int scaledHeight, int zOffset);
+
+	default void drawTexture(MatrixStack matrices, int x, int y, int zOffset, int u, int v, int width, int height) {
+		DrawableHelper.drawTexture(matrices, x, y, zOffset, u, v, width, height, 256, 256);
+	}
```
deirn: Eh, I think this is unnecessary. How many times one would call this anyway?

deirn (changes requested):
Run the `checkstyleClient` task, there are few styling issues.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/CrosshairRenderCallback.java
ghost: Yeah, now that I have an another look at it, it doesn't even seem like it's helping much either. Removed.

Technici4n (commented):
One thing I'm noticing is that this event can't differentiate between canceling crosshair rendering entirely, and replacing the vanilla render by something else.

ghost:
I made the event in a way that you can add something else to the vanilla crosshair (like vanilla attack cooldown) or replace/cancel it. But either way all listeners are called to render. From what I understand you are talking about a way to completely cancel it without rendering anything. If that's the case how should I go with it's implementation. I can add a method to the event interface that returns true by default and if any of the listeners return it false, it can just cancel crosshair rendering completely before any listeners are called to render.

LambdAurora:
> replace/cancel it

The issue with this is what if there's 2 mods that conditionally replace/cancel it?
They could both render the cursor without knowing about each other.
And that's where there's the issue.

"But Aurora, if I stop execution of the listeners when cancelled, what about some other use cases?"

Event phases got you covered, they'll allow mods to order themselves to render in a more proper manner I guess?
but I'm also doubting it's entirely sufficient.

ghost:
I'm pretty much gambling with the pr with my last commit, it's not the most efficient but I think I covered all the blind spots. Now only one listener can cancel vanilla crosshair, any other listeners canceling it won't be rendered. You can cancel all rendering by calling ```cancelAll()```. If you don't want any other listeners to render something under specific conditions you can add your render by calling ```renderCancelOthers(...)``` and only yours will be rendered.
Also added a small test to testmod. Showcasing the test [video](https://streamable.com/y154cf) on streamable.

PR #2411, opened 2022-07-21
Message API: 1.19.1 changes

apple502j:
## Changes
- **BREAKING**: Provide `MinecraftServer` in `AllowGameMessage` and `GameMessage` events, since they do not have a parameter that allows server access.
- Update javadoc to reflect 1.19.1-pre6 change removing the pure decorator requirement.
- Update testmod to test non-pure decorator and reflect other changes.

## Testing
Tested on dev env (without signing).

## To Do
There is one remaining issue: the current `AllowChatMessage`/`AllowCommandMessage` implementation likely fails in authenticated environments, as the message chain will be split if the sender sends another message after a blocked message, disconnecting the sender. If that is true, there are 2 potential fixes, both with pros and cons:

1. Send the message header to all players, without body. This is what they do for messages fully censored using text filtering. Pros: small change, compatible. Cons: all clients receive the sender's UUID and message hash, which might not be expected in some cases.
2. Move the mixin point before unpacking. This is what they do for messages that fail validation, etc. Pros: does not have to send message header. Cons: requires API-breaking change and introduces inconsistency with unprefixed `ChatMessage`/`GameMessage` events.

We first have to check that this is actually a bug (and not a theoretical issue), and try those fixes. Unfortunately it's not easy for me to test these, as my computer isn't beefy enough to run both the dedicated server and the client, and I do not have the second account for testing situations with other players.

### DIFF
--- a/fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageDecoratorEvent.java
+++ b/fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageDecoratorEvent.java
@@ -41,10 +41,12 @@ import net.fabricmc.fabric.api.event.EventFactory;
  * function. If not given, the message decorator will run in the default phase, which is between
  * the content phase and the styling phase.
  *
- * <p>When implementing a message decorator, it is <strong>very important that the decorator be
- * pure; i.e. return the same text when called multiple times for the same arguments (message and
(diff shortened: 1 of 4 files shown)


apple502j:
They released rc2, we can fix allow events later.

modmuss50 (approved):
LGTM, I agree with fixing possible bugs seprately, want to get these breaking changes out soon.

Player3324 (approved):

PR #2412, opened 2022-07-22
Make `DEFAULT_ENABLED` works with client resource pack

deirn:
Right now client resource pack with `DEFAULT_ENABLED` only works when `options.txt > resourcePacks` is an empty array, which only happens when the resource pack selection screen is never opened.

This PR makes it work by tracking pack ids in an NBT file (`<gameDir>/data/fabricDefaultResourcePacks.dat`):
- If there is NO value with matching resource pack id, add it to the enabled packs and the tracker file.
- If there is a matching value and pack id, do not add it to the enabled packs and let the options value decides if it is enabled or not.
- If there is a value without matching pack id (e.g. because the mod is removed), remove it from the tracker file so that it would be enabled again if added back later.

Also possible bug fix (not really tested):
`Fabric Mods` client resource pack _technically_ only gets added to the default packs when there's at least one mod that has a file in the `assets` directory, but since the resource loader module itself has its icon in the `assets` it gets added regardless.

### DIFF
--- a/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
+++ b/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
@@ -16,9 +16,20 @@
 
 package net.fabricmc.fabric.mixin.resource.loader.client;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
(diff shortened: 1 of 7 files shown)


modmuss50 (approved):

modmuss50:
Changes look fine, some sort of test might be nice, not too sure how feasible it would be however.

modmuss50:
Changes look fine, some sort of test might be nice, not too sure how feasible it would be however.

apple502j (approved):

Player3324 (approved):

modmuss50:
Manually cherry-picked into the 1.19.2 branch. Thanks a lot 👍

PR #2426, opened 2022-07-27
Fix: broadcast message headers when chat event is cancelled

Gegy:
This PR resolves an issue where cancelling chat messages through `ServerMessageEvents` could cause clients to disconnect due to a broken chat signing chain.
When a message was hidden through the event, the header for that chat message would not be distributed to other clients. Because each new chat message is signed in relation to the previous message, when that player next sends a new message, clients will detect that it is now referencing another message that it never received. This causes the disconnect for all receiving players.

This PR instead distributes the headers for these cancelled messages to ensure the chat chains stay consistent. 🙂


modmuss50 (approved):
Great, thanks for this. 👍 Will get it out ASAP

PR #2429, opened 2022-07-29
Fix POI types not being registered properly

apple502j:
`PointOfInterestType` must be registered in `PointOfInterestTypes.POI_STATES`, otherwise the deserializer ignores the serialized POI and re-creates it. `PointOfInterestHelper` now adds the states to the set. (See also MC-254634)

Forwardport needed to 1.19.1, backport to 1.18.2 is not needed


modmuss50 (approved):

modmuss50:
Manually cherry-picked into the 1.19.2 branch. Thanks a lot 👍

PR #2435, opened 2022-07-31
Add more item convention tags.

IMB11:
Adds the following item tags:

`c:meat`
`c:poisonous_foods`
`c:raw_meat`
`c:cooked_meat`
`c:raw_fish`
`c:cooked_fish`
`c:books`
`c:sugary_foods`
`c:mushrooms`
`c:seeds`
`c:fruits`


fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/api/tag/convention/v1/ConventionalItemTags.java
```diff
@@ -90,6 +90,15 @@ public class ConventionalItemTags {
 	public static final TagKey<Item> WHITE_DYES = register("white_dyes");
 	public static final TagKey<Item> YELLOW_DYES = register("yellow_dyes");
 
+	// Note: FISH is not here because it is covered by minecraft:item/fishies
```
dvitski:
```suggestion
	// Note: FISH is not here because it is covered by minecraft:item/fishes
```

IMB11:
Converting this PR to draft to add more convention tags.

Noaaan:
Fruits is not mentioned. I would suggest adding a tag for `vegetables`

IMB11:
That's a good point actually, I'll add vegetables

Shnupbups:
Are Glow Berries not fruits? What about both Golden Apples? What about Glistering Melon Slices, do they count even if they can't be eaten? What about Melon Blocks, do they count? Pumpkins are technically fruits, but do they count? I dunno.

Would a `c:golden_foods` tag be worth having, containing Golden Apples, Enchanted Golden Apples, and Golden Carrots? (and maybe Glistering Melon Slices? again, can't be eaten though...)
Could have `c:golden_apples` too, containing just the two Golden Apples. Then have that tag nested within `c:golden_foods`.
What about `c:stews` containing Mushroom Stew, Rabbit Stew, Suspicious Stew, and Beetroot Soup?
`c:maps` containing Map and Filled Map?
`c:berries` containing Sweet Berries and Glow Berries?
`c:helmets`, `c:chestplates`, `c:leggings`, and `c:boots`?
Then those should probably be nested inside `c:head_wearables`, `c:chest_wearables`, `c:legs_wearables`, and `c:feet_wearables` (with carved pumpkin and mob heads also in the head one, and elytra also in the chest one)
`c:horse_armor` and `c:shields`?

PR #2446, opened 2022-08-05
Rename modid to "fabric-api" and provide "fabric". Update loom

modmuss50:
- Rename modid to "fabric-api" and provide "fabric"
- Update loom
- Update loader
- Update gradle

Small looking change, but will need significant testing.


src/main/resources/fabric.mod.json
```diff
@@ -1,6 +1,9 @@
 {
   "schemaVersion": 1,
-  "id": "fabric",
+  "id": "fabric-api",
+  "provide": [
```
deirn: Pretty sure it's `provides`, not `provide` :P

modmuss50: Opps. You can tell I haven't tested it yet 👍

SilverAndro:
Might want to mark this as superseding/closing #1263

PR #2448, opened 2022-08-06
Message API: fix outdated and incorrect javadoc

apple502j:
- Fix incorrect commands listed as caller. `/msg` is private message and does not "broadcast". `/tellraw` does not sign the content.
- Replace "Message" with "message body" when referencing to the event
- Add the note that events can only block the body and message headers are still broadcast
- Fix incorrect/outdated code example


modmuss50 (approved):

modmuss50:
Cherrypicked into 1.19.2 👍

PR #2450, opened 2022-08-06
Rename and validate Mixin class names.

modmuss50:
Uses the new loom mixin validaition task. Removes a few un-used or duplicate mixins.

High priority as it will be a slight pain for PRs / merging.


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -43,24 +44,35 @@
...
+import net.fabricmc.fabric.impl.registry.sync.RegistrySyncManager;
 import net.fabricmc.fabric.impl.registry.sync.RemapException;
 import net.fabricmc.fabric.impl.registry.sync.RemapStateImpl;
 import net.fabricmc.fabric.impl.registry.sync.RemappableRegistry;
 
 @Mixin(SimpleRegistry.class)
-public abstract class MixinIdRegistry<T> extends Registry<T> implements RemappableRegistry, ListenableRegistry<T> {
+public abstract class SimpleRegistryMixin<T> extends Registry<T> implements RemappableRegistry, ListenableRegistry<T> {
+	@Unique
+	private static final Logger FARBIC_LOGGER = LoggerFactory.getLogger("FabricRegistrySync");
```
Juuxel: "Farbic"? Isn't there already another logger below anyway?

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MultiNoiseUtilMultiNoiseSamplerMixin.java
```diff
@@ -28,7 +28,7 @@
 import net.fabricmc.fabric.impl.biome.MultiNoiseSamplerHooks;
 
 @Mixin(MultiNoiseUtil.MultiNoiseSampler.class)
-public class MixinMultiNoiseSampler implements MultiNoiseSamplerHooks {
+public class MultiNoiseUtilMultiNoiseSamplerMixin implements MultiNoiseSamplerHooks {
```
Juuxel, line 31: For cases like these I think the task should allow to just use the inner name too, but doesn't really matter here

modmuss50: The yarn name might be questionable here, generally I dont think an inner class needs the outer name? For now Ill leave this.

PR #2456, opened 2022-08-09
add fix to AFTER_PLAYER_CHANGE_WORLD

CalebBassham:
Adds a workaround to still call the AFTER_PLAYER_CHANGE_WORLD event when a player is moving from the end to the overworld.


apple502j:
Not sure if this is the best idea, two different things should not trigger the same event.

CalebBassham:
I mean, as long as the event is only called once when a player changed worlds, I would say that this is fine. I would not expect the average person using the api to expect that AFTER_PLAYER_CHANGE_WORLD does not apply to changing worlds from the end to the overworld.

Technici4n:
If the player is killed and resurrected, wouldn't some sort of player resurrect event make more sense for this?

CalebBassham:
They aren't really killed. They are just removed from all worlds while they are on the credits screen.

PR #2464, opened 2022-08-13
Support stack aware recipe remainders

AlphaMode:
This PR allows modders to add stack aware recipe remainders
This is useful for mods that adds custom buckets that support different fluids

This address the long lasting issue #50


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@Mixin(AbstractFurnaceBlockEntity.class)
+public class AbstractFurnaceBlockEntityMixin {
+	private static DefaultedList<ItemStack> capturedInventory;
+
+	@Inject(method = "tick", at = @At("HEAD"))
+	private static void getStackCraftingRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci) {
+		capturedInventory = ((AbstractFurnaceBlockEntityAccessor)blockEntity).getInventory();
```
Juuxel: Can't you shadow since it's in the same class?

Juuxel (commented):
I'd avoid those `@Overwrite`s if possible. This is quite a lot considering there are **none** currently in Fabric API.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
AlphaMode: No, its static and you can't capture locals on redirects/modify var

AlphaMode:
> I'd avoid those `@Overwrite`s if possible. This is quite a lot considering there are **none** currently in Fabric API.

For tthese two ones I don't see another way of doing this since they are in for loops and I can't capture the locals with a redirect so this is my only choice unless someone has a better solution I can implement

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
Juuxel: I don't see how the annotation type matters? This should work afaik:
```java
@Shadow private Whatever inventory;

@Inject(...)
private ...() {
    capturedInventory = blockEntity.inventory;
}
```

AlphaMode: image
Do you mean something like this?

Juuxel: Yeah, I think that should work. It looks a bit cleaner than a separate accessor just for a mixin to the same class.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,31 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
 	default boolean isSuitableFor(ItemStack stack, BlockState state) {
 		return ((Item) this).isSuitableFor(state);
 	}
+
+	/**
+	 * Determines if the item will have a leftover item after it's been used.
+	 * Example: Using lava in a furnace as fuel,
+	 * Stack-aware version of {@link Item#hasRecipeRemainder()}.
```
apple502j:
```suggestion
	 * Determines if the item will leave a new item stack after being consumed in a recipe.
	 * (This is also known as "recipe remainder".)
	 * For example, using a lava bucket in a furnace as fuel will leave an empty bucket.
	 *
	 * <p>This is a stack-aware version of {@link Item#hasRecipeRemainder()}.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,31 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
...
+	 *
+	 * @param stack the current stack
+	 * @return true if the item has a recipe remainder
```
apple502j:
```suggestion
	 * @return {@code true} if the item has a recipe remainder
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,31 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
...
+	 */
+	default boolean hasRecipeRemainder(ItemStack stack) {
+		return ((Item) this).hasRecipeRemainder();
+	}
+
+	/**
+	 * Return a leftover item for use in recipes
+	 * Stack-aware version of {@link Item#getRecipeRemainder()}.
```
apple502j:
```suggestion
	 * Returns a leftover item stack after {@code stack} is consumed in a recipe.
	 * (This is also known as "recipe remainder".)
	 * For example, using a lava bucket in a furnace as fuel will leave an empty bucket.
	 *
	 * <p>This is a stack-aware version of {@link Item#getRecipeRemainder()}.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,31 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
...
+	 *
+	 * @param stack the old {@link ItemStack}
```
apple502j:
```suggestion
	 * @param stack the consumed {@link ItemStack}
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,31 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
...
+	 * @return the leftover item
```
apple502j:
```suggestion
	 * @return the leftover item stack
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,47 @@
...
+package net.fabricmc.fabric.api.item.v1;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+
+public interface FabricItemStack {
+	/**
+	 * Determines if the item will have a leftover item after it's been used.
+	 * Example: Using lava in a furnace as fuel,
+	 * Stack-aware version of {@link Item#hasRecipeRemainder()}.
```
apple502j:
```suggestion
	 * Determines if this item stack will leave a new item stack after being consumed in a recipe.
	 * (This is also known as "recipe remainder".)
	 * For example, using a lava bucket in a furnace as fuel will leave an empty bucket.
	 *
	 * <p>This is a stack-aware version of {@link Item#hasRecipeRemainder()}.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,47 @@
...
+	 *
+	 * @return true if the item has a recipe remainder
```
apple502j:
```suggestion
	 * @return {@code true} if the item stack has a recipe remainder
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	default boolean hasRecipeRemainder() {
+		return ((ItemStack) this).getItem().hasRecipeRemainder((ItemStack) this);
+	}
+
+	/**
+	 * Return a leftover item for use in recipes
+	 * Stack-aware version of {@link Item#getRecipeRemainder()}.
```
apple502j:
```suggestion
	 * Returns a leftover item stack after this item stack is consumed in a recipe.
	 * (This is also known as "recipe remainder".)
	 * For example, using a lava bucket in a furnace as fuel will leave an empty bucket.
	 *
	 * <p>This is a stack-aware version of {@link Item#getRecipeRemainder()}.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,47 @@
...
+	 *
+	 * @return the leftover item
```
apple502j:
```suggestion
	 * @return the leftover item stack
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,53 @@
...
+	protected DefaultedList<ItemStack> inventory;
+	private static DefaultedList<ItemStack> capturedInventory;
+
+	@Inject(method = "tick", at = @At("HEAD"))
+	private static void getStackCraftingRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci) {
+		capturedInventory = ((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).inventory;
+	}
+
+	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z", ordinal = 2))
+	private static boolean setStackCraftingRemainder(ItemStack itemStack) {
```
apple502j: ~~Can't we capture the args in `Redirect`? We shouldn't need a static field for this.~~
Also, convention is to name it `stack`.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,53 @@
...
+		if (itemStack.hasRecipeRemainder()) {
+			capturedInventory.set(1, itemStack.getRecipeRemainder());
+			return true;
```
apple502j:
```suggestion
			return false;
```

Returning `true` calls the vanilla code and overwrites the result.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,53 @@
...
+	private static DefaultedList<ItemStack> capturedInventory;
+
+	@Inject(method = "tick", at = @At("HEAD"))
+	private static void getStackCraftingRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci) {
+		capturedInventory = ((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).inventory;
+	}
+
+	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z", ordinal = 2))
+	private static boolean setStackCraftingRemainder(ItemStack itemStack) {
+		if (itemStack.hasRecipeRemainder()) {
```
apple502j: Needs testing with a furnace smelting wet sponge using multiple buckets. Note that the furnace must be fueled by other fuel first and the buckets must be inserted with command. The code here suggests that the empty bucket is always replaced; the redirected `isEmpty()` check, from what I can see, makes it only replace on the last bucket, but I couldn't reproduce it on vanilla.

~~Also, `itemStack` may be empty as this is called after `decrement`, which needs documenting. (Decrementing cannot turn the stack into `ItemStack#EMPTY`, however, as `item` field is final.)~~

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BannerDuplicateRecipeMixin.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.BannerDuplicateRecipe;
+import net.minecraft.util.collection.DefaultedList;
+
+@Mixin(BannerDuplicateRecipe.class)
+public class BannerDuplicateRecipeMixin {
+	@Unique
+	private ItemStack capturedItemStack;
+
+	@Inject(method = "getRemainder(Lnet/minecraft/inventory/CraftingInventory;)Lnet/minecraft/util/collection/DefaultedList;", at = @At(value = "JUMP", target = "Lnet/minecraft/inventory/CraftingInventory;getStack(I)Lnet/minecraft/item/ItemStack;"), locals = LocalCapture.CAPTURE_FAILHARD)
```
apple502j: ~~Suggestion: Redirect `ItemStack#getItem` call and do all logics there. We can then return `Items.AIR`, as they should not have a recipe remainder. Same applies to `BookCloningRecipeMixin`.~~

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BrewingStandBlockEntityMixin.java
```diff
@@ -0,0 +1,54 @@
...
+
+import net.minecraft.block.entity.BrewingStandBlockEntity;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.collection.DefaultedList;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@Mixin(BrewingStandBlockEntity.class)
+public class BrewingStandBlockEntityMixin {
```
apple502j: Seems like this is the best way.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RecipeMixin.java
```diff
@@ -0,0 +1,50 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Recipe;
+import net.minecraft.util.collection.DefaultedList;
+
+import net.fabricmc.fabric.impl.item.RecipeRemainderHandler;
+
+@Mixin(Recipe.class)
+public interface RecipeMixin<C extends Inventory> {
+	@Inject(method = "getRemainder", at = @At(value = "INVOKE", target = "Lnet/minecraft/inventory/Inventory;getStack(I)Lnet/minecraft/item/ItemStack;"), locals = LocalCapture.CAPTURE_FAILHARD)
```
apple502j, line 37: ~~Hmm, could we use the `getItem` trick?~~

apple502j (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
AlphaMode: image
No it doesn't it prevents the vanilla code from running

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,36 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
...
+	 * Determines if the item will leave a new item stack after being consumed in a recipe.
+	 * (This is also known as "recipe remainder".)
+	 * For example, using a lava bucket in a furnace as fuel will leave an empty bucket.
+	 *
+	 * <p>This is a stack-aware version of {@link Item#hasRecipeRemainder()}.
+	 *
+	 * @param stack the current stack
+	 * @return {@code true} if the item has a recipe remainder
+	 */
+	default boolean hasRecipeRemainder(ItemStack stack) {
```
Technici4n: This should be removed and replaced by simply returning `ItemStack.EMPTY` in `getRecipeRemainder`.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/RecipeRemainderHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+package net.fabricmc.fabric.impl.item;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.ItemStack;
+
+@ApiStatus.Internal
+public class RecipeRemainderHandler {
+	public static ThreadLocal<ItemStack> capturedRecipeItemStack = new ThreadLocal<>();
```
Technici4n:
```suggestion
	public static final ThreadLocal<ItemStack> capturedRecipeItemStack = new ThreadLocal<>();
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BannerDuplicateRecipeMixin.java
```diff
@@ -0,0 +1,52 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.inventory.CraftingInventory;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.BannerDuplicateRecipe;
+import net.minecraft.util.collection.DefaultedList;
+
+@Mixin(BannerDuplicateRecipe.class)
```
Technici4n: This mixin is unneeded and should be removed.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BookCloningRecipeMixin.java
```diff
@@ -0,0 +1,52 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.inventory.CraftingInventory;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.BookCloningRecipe;
+import net.minecraft.util.collection.DefaultedList;
+
+@Mixin(BookCloningRecipe.class)
```
Technici4n: This mixin is unneeded and should be removed.

Technici4n (changes requested):

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/CustomDamageTest.java
```diff
@@ -55,5 +55,11 @@ public Text getName(ItemStack stack) {
 			int v = stack.getOrCreateNbt().getInt("weird");
 			return super.getName(stack).shallowCopy().append(" (Weird Value: " + v + ")");
 		}
+
+		@Override
+		public ItemStack getRecipeRemainder(ItemStack stack) {
+			stack.setDamage(stack.getDamage() + 1);
```
Technici4n: You shouldn't be modifying the stack imo.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RecipeMixin.java
```diff
@@ -40,11 +40,11 @@ default void captureStack(C inventory, CallbackInfoReturnable<DefaultedList<Item
...
 	@Redirect(method = "getRemainder", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/Item;hasRecipeRemainder()Z"))
 	private boolean hasStackRemainder(Item instance) {
-		return RecipeRemainderHandler.capturedRecipeItemStack.get().hasRecipeRemainder();
+		return true;
 	}
 
 	@Redirect(method = "getRemainder", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/collection/DefaultedList;set(ILjava/lang/Object;)Ljava/lang/Object;"))
 	private Object getStackRemainder(DefaultedList<ItemStack> inventory, int index, Object element) {
-		return inventory.set(index, RecipeRemainderHandler.capturedRecipeItemStack.get().getRecipeRemainder());
+		return inventory.set(index, RecipeRemainderHandler.capturedRecipeItemStack.get().getRecipeRemainder().copy());
```
Technici4n: The copy shouldn't be necessary.

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/CustomDamageTest.java
ghost: May be better:

```java
ItemStack moreDamaged = stack.copy();
moreDamaged.setDamage(stack.getDamage() + 1);
return moreDamaged;
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,49 @@
...
+	@Inject(method = "tick", at = @At("HEAD"))
+	private static void getStackCraftingRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci) {
+		capturedInventory = ((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).inventory;
+	}
+
+	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z", ordinal = 2))
+	private static boolean setStackCraftingRemainder(ItemStack itemStack) {
+		capturedInventory.set(1, itemStack.getRecipeRemainder().copy());
+		return true;
+	}
```
ghost: This fixes the bug that overwrites entire stacks, and uses the vanilla way to put the remainder.

PS. Maybe if there is no space in the fuel slot the remainder should be dropped? 🤔
(Note that vanilla just replaces, and not drops, the item)

```suggestion
	private static boolean setStackCraftingRemainder(ItemStack stack) {
		//Gets the remainder.
		ItemStack remainder = stack.getRecipeRemainder().copy();

		//Decrements the stack like vanilla do.
		stack.decrement(1);

		//Insert the remainder if the stack is empty.
		if (stack.isEmpty()) {
			capturedInventory.set(1, remainder);
		}

		return true;
	}
```

ghost (changes requested):
Just a fix.

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/CustomDamageTest.java
```diff
@@ -55,5 +55,11 @@ public Text getName(ItemStack stack) {
...
+			return stack;
```
ghost: Item should disappear if has the max damage.
```suggestion
			if (stack.getDamage() < stack.getMaxDamage()) {
				ItemStack moreDamaged = stack.copy();
				moreDamaged.setDamage(stack.getDamage() + 1);
				return moreDamaged;
			}

			return ItemStack.EMPTY;
```

ghost (changes requested):
Fix two

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
ghost: This is the same behavior as vanilla do.

[](https://user-images.githubusercontent.com/32025549/188224151-c122684e-48d1-4778-895e-f469d504d79f.mp4)

ghost (changes requested):
Suggestion:
Add an item with a custom remainder to the fuel registry for testing the furnace, maybe the weird pickaxe itself.

You must add the content registry module as dependency for the testmod into the build.gradle file to be able to use the `FuelRegistry`.
``` java
archivesBaseName = "fabric-item-api-v1"
version = getSubprojectVersion(project)

moduleDependencies(project, [
		'fabric-api-base'
])

dependencies {
	testmodImplementation project(path: ':fabric-content-registries-v0', configuration: 'namedElements')
}
```

```java
Item item = Registry.register(Registry.ITEM, new Identifier("fabric-item-api-v1-testmod", "weird_pickaxe"), new WeirdPick());
FuelRegistry.INSTANCE.add(item, 200);
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
AlphaMode: Seems to work as intended

ghost (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+
+public interface FabricItemStack {
```
Juuxel, line 30: Needs a top-level jd comment like FabricItem

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -91,4 +91,18 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getAttributeModifiers
 	default boolean isSuitableFor(ItemStack stack, BlockState state) {
 		return ((Item) this).isSuitableFor(state);
 	}
+
+	/**
+	 * Returns a leftover item stack after {@code stack} is consumed in a recipe.
+	 * (This is also known as "recipe remainder".)
```
Juuxel:
```suggestion
	 * (This is also known as a "recipe remainder".)
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -0,0 +1,35 @@
...
+	/**
+	 * Returns a leftover item stack after this item stack is consumed in a recipe.
+	 * (This is also known as "recipe remainder".)
```
Juuxel:
```suggestion
	 * (This is also known as a "recipe remainder".)
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RecipeMixin.java
```diff
@@ -0,0 +1,50 @@
...
+	default void captureStack(C inventory, CallbackInfoReturnable<DefaultedList<ItemStack>> cir, DefaultedList<ItemStack> defaultedList, int i) {
+		RecipeRemainderHandler.capturedRecipeItemStack.set(inventory.getStack(i));
+	}
+
+	@Redirect(method = "getRemainder", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/Item;hasRecipeRemainder()Z"))
+	private boolean hasStackRemainder(Item instance) {
+		return true;
+	}
+
+	@Redirect(method = "getRemainder", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/collection/DefaultedList;set(ILjava/lang/Object;)Ljava/lang/Object;"))
```
Juuxel: Could this use an arg modifier on the `set` call instead? It'd be a bit less prone to conflicts than a full redirect.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,59 @@
...
+	@Shadow
+	protected DefaultedList<ItemStack> inventory;
+	private static DefaultedList<ItemStack> capturedInventory;
+
+	@Inject(method = "tick", at = @At("HEAD"))
+	private static void getStackCraftingRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci) {
+		capturedInventory = ((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).inventory;
+	}
+
+	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z", ordinal = 2))
```
Juuxel: A slice would be better for MC updates than an ordinal in a pretty lengthy method. (Especially since it would target a different call already in 1.19)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,59 @@
...
+	private static boolean setStackCraftingRemainder(ItemStack stack) {
+		//Gets the remainder.
+		ItemStack remainder = stack.getRecipeRemainder();
+
+		//Decrements the stack like vanilla do.
+		stack.decrement(1);
```
Juuxel: Note for 1.19 porting (this was originally a review comment but I didn't notice this PR targetted an outdated MC version):

If unchanged, this will decrement the stack twice. 1.19 removed one of the `isEmpty` calls so this redirector will target the "inner" `isEmpty` after vanilla decrements the stack, if I'm reading the source correctly.

Also, returning true from this redirector in 1.19 would mean that the vanilla logic runs *in addition to* Fabric logic. That would unnecessarily set the recipe remainder twice for items using the vanilla remainder method - not that it probably matters much.

Pseudocode to explain the differences:
```java
if (stack is not empty) { // in 1.18.2, a method call; in 1.19, a local variable reference
    // decrement item stack
    if (stack is now empty) { // this is always an isEmpty call
      // set remainder with vanilla logic
    }
}
```

Juuxel (changes requested):
The mixins should be package-private: https://github.com/FabricMC/fabric/blob/1.19.2/CONTRIBUTING.md#mixins

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
ghost: ### Mmmm, I suggest this:

1. Inject here and get a copy of the item stack before it is decremented and eventually turns empty, to get a reference for obtaining the remainder from the stack.
2. Get the remainder from the stack saved at point 1 and overwrite the vanilla set method with the remainder.
Idk what mixin type is this, you should change the value setted with the set method.

(See 1 and 2 in the images below)

---
Immagine 2022-09-05 024855  
---
Immagine 2022-09-05 025006  
---

apple502j: Actually this is exactly the same as my review (yes, I also forgot this is 1.19)

see https://github.com/FabricMC/fabric/pull/2464#discussion_r950873025

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -41,8 +42,11 @@ private static void getStackCraftingRemainder(World world, BlockPos pos, BlockSt
 		capturedInventory = ((AbstractFurnaceBlockEntityMixin) (Object) blockEntity).inventory;
 	}
 
-	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z", ordinal = 2))
-	private static boolean setStackCraftingRemainder(ItemStack stack) {
+	@Redirect(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isEmpty()Z"), slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;decrement(I)V")))
+	private static boolean setStackCraftingRemainder(ItemStack oldStack) {
+		// We need the item stack before it gets decremented.
+		ItemStack stack = capturedInventory.get(1);
```
Juuxel: Will this really work? Since it's the same stack instance as the modified one, it'll already have been decremented as far as I can tell.

AlphaMode: Some testing seems to show it works fine.

Juuxel: It doesn't - the stack is decremented ⇒ it becomes empty ⇒ `getItem` returns `minecraft:air` ⇒ `FabricItemStack.getRecipeRemainder` returns `ItemStack.EMPTY`.

https://user-images.githubusercontent.com/6596629/188511189-1073029d-47f5-4f3a-b27b-886a38c04b9c.mp4

ghost (changes requested):
Cleaned the mixin and made more compatible with future releases.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,68 @@
...
+		stack.decrement(1);
+
+		//Insert the remainder if the stack is empty.
+		if (stack.isEmpty()) {
+			capturedInventory.set(1, remainder);
+		}
+
+		return false;
+	}
+}
```
Juuxel: @devpelux No need for a no-op redirect either, you can just inject at the `decrement` call and store the remainder - even more compatible!

ghost:
Sorry i deleted my review before Juuxel comment, however i will do a pr in the source fork. 😅

ghost (approved):
For me this is complete (except a missing license 🤔).

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/RecipeRemainderHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+package net.fabricmc.fabric.impl.item;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.ItemStack;
+
+@ApiStatus.Internal
+public class RecipeRemainderHandler {
+	public static final ThreadLocal<ItemStack> capturedRecipeItemStack = new ThreadLocal<>();
```
modmuss50, line 25: This doesnt need its own class does it? Put it in the mixin.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,51 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Recipe;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@Mixin(AbstractFurnaceBlockEntity.class)
+abstract class AbstractFurnaceBlockEntityMixin {
+	@Unique
+	private static ItemStack remainder;
```
modmuss50, line 38: I see no reason for this to be static, could lead to possible issues if people are using mods that try to make things multithreaded.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BrewingStandBlockEntityMixin.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.collection.DefaultedList;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@Mixin(BrewingStandBlockEntity.class)
+class BrewingStandBlockEntityMixin {
+	@Unique
+	private static ItemStack capturedItemStack;
```
modmuss50, line 38: Again make it not static

fabric-item-api-v1/build.gradle
```diff
@@ -4,3 +4,7 @@ version = getSubprojectVersion(project)
 moduleDependencies(project, [
 		'fabric-api-base'
 ])
+
+dependencies {
+	testmodImplementation project(path: ':fabric-content-registries-v0', configuration: 'namedElements')
+}
```
modmuss50, line 10: Update this branch to 1.19.2, this is the old way of defining test mod deps.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,51 @@
...
+	@Inject(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;decrement(I)V"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private static void getRemainder(World world, BlockPos pos, BlockState state, AbstractFurnaceBlockEntity blockEntity, CallbackInfo ci, boolean bl, boolean bl2, ItemStack stack, Recipe recipe, int i, Item item) {
+		//Gets the remainder before the decrement.
+		remainder = stack.getRecipeRemainder();
+	}
+
+	@ModifyArg(method = "tick", index = 1, at = @At(value = "INVOKE", target = "Lnet/minecraft/util/collection/DefaultedList;set(ILjava/lang/Object;)Ljava/lang/Object;"))
+	private static Object setRemainder(Object oldStack) {
+		//Puts the remainder into the inventory.
+		return remainder;
```
modmuss50, line 49: Save remainder in a local var, and then set the field to null.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/BrewingStandBlockEntityMixin.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	@Redirect(method = "craft", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/Item;hasRecipeRemainder()Z"))
+	private static boolean hasStackRecipeRemainder(Item instance) {
+		return true;
+	}
+
+	@ModifyVariable(method = "craft", at = @At(value = "STORE"), index = 4)
+	private static ItemStack createStackRecipeRemainder(ItemStack old) {
+		return capturedItemStack.getRecipeRemainder();
```
modmuss50, line 52: Save capturedItemStack in a local var, and then set the field to null before returning.

modmuss50 (changes requested):

modmuss50:
Removed the `next` tag untill this has been updated to 1.19.2

ghost:
Any news? @AlphaMode

AlphaMode:
Superseded by #2556

PR #2474, opened 2022-08-22
fabric-dimensions-v1: replace custom target with ThreadLocal

teddyxlandlee:
Considering some optimizing mod would use multi-thread algorithm on entity calculation, single-thread custom target getting/setting might be not thread-safe, making the entities teleporting to wrong places.

Replacing it with `ThreadLocal` might solve this potential problem. `Entity.moveToWorld` is single-thread, so using `ThreadLocal` won't miss the data.

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
@@ -29,7 +29,7 @@ public final class FabricDimensionInternals {
 	/**
 	 * The target passed to the last call to {@link FabricDimensions#teleport(Entity, ServerWorld, TeleportTarget)}.
 	 */
-	private static TeleportTarget currentTarget;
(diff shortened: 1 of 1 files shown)


apple502j:
We should instead store this in entity field (assuming that works?)

I'll try to PR this somewhere this week.

teddyxlandlee:
> We should instead store this in entity field (assuming that works?)

Well, that's a great idea. But teleporting is a continuous process, and at one time a thread can only run for one entity, which is enough.

So there's no need to modify current codebase.

Technici4n:
Imo it's fair enough to assume that there's only one active server world at a given time. Multithreading mods are inherently broken, and shouldn't be used.

teddyxlandlee:
Of course only one server world is active at one time. But there could be more than 1 thread calculating entities in order to solve Mojang's annoying optimization. Some mod can do this, and I don't wanna see it breaking Fabric API or any essential module.

apple502j:
This API assumes that it is called from the server thread (with the assumption that there is a server thread per server) and crashes otherwise. I don't believe this is what we should support, esp with potential performance concerns ThreadLocal gives.

PR #2478, opened 2022-08-23
Allow Message to Player Event

BlueZeeKing:
# Message API ALLOW_MESSAGE_TO_PLAYER

## Resolves

Implements some of #1147 

## What is this?

This adds the ALLOW_MESSAGE_TO_PLAYER event to ServerMessageEvents. It's similar to the `ALLOW_COMMAND_MESSAGE` and `ALLOW_CHAT_MESSAGE`  but it operates on a per-player basis. It applies to any messages a player sends, including commands like `/msg`, `/say`, and `/me` and messages sent as normal chat messages through the GUI.

## What does this not do?

This API does not apply to messages sent by anything other than a player.

## Testing

This mod uses the same mixins that the former Message Channels used so all that testing is still useful. I further tested it using the current API with the `/msg`, `/say`, and `/me` commands. It was also tested on sending messages from things other than a player using `/execute`. The testmod blocks the player from seeing any message with their username in it.


fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/MessageChannels.java
```diff
@@ -0,0 +1,94 @@
...
+ *     MessageChannels.setChannel(player.getUuid(), new Identifier("example", "channel1"));
+ * </code></pre>
+ *
+ * <p>An example of putting a player back into the global channel:
+ *
+ * <pre><code>
+ *     MessageChannels.setChannel(player.getUuid(), MessageChannels.GLOBAL_CHANNEL);
+ * </code></pre>
+ */
+public class MessageChannels {
```
haykam821:
```suggestion
public final class MessageChannels {
```

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/MessageChannels.java
```diff
@@ -0,0 +1,94 @@
...
+	private MessageChannels() {
+	}
+
+	/**
+	 * The default identifier all players start in.
+	 */
+	public static final Identifier GLOBAL_CHANNEL = new Identifier("fabric-message-api-v1", "global");
+
+	private static final HashMap<UUID, Identifier> playerChannels = new HashMap<>();
```
haykam821: This field should follow static naming conventions.

haykam821 (changes requested):
The documentation makes a few grammatical mistakes and does not follow the convention of prefixing `@param` and `@return` descriptions with lowercase letters.

apple502j (changes requested):
I believe message channels are a bit out of scope. Mods can implement it easily now that AllowChatMessage event exists.

That said, if we do provide it, then we should use that event instead of mixin. Also how would message channel work with commands, particularly ones /executed?

apple502j:
Ah, this is per-player. Then we need a new event, yeah. That said we should provide this as generic event, not usage-specific one.

BlueZeeKing:
> Ah, this is per-player. Then we need a new event, yeah. That said we should provide this as generic event, not usage-specific one.

I think you're right. I can keep the mixins the same and add the event.

apple502j (commented):
Note to self: this does send header packet, so that's good

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageEvents.java
```diff
@@ -88,6 +90,25 @@ public final class ServerMessageEvents {
...
+	/**
+	 * An event triggered when a command or chat message is sent. Listeners on this event
+	 * are triggered for every player receiving that message. If any listener attached returns
+	 * {@code false} the message will not be sent to that specific player.
+	 *
+	 * <p>If a listener returns {@code false}, the server will not broadcast the message to that
+	 * player and no further listeners will be checked and the message will be called. The
+	 * {@link #CHAT_MESSAGE} or {@link #COMMAND_MESSAGE} are still run.
+	 *
+	 * <p>This also applies to {@see net.minecraft.server.PlayerManager#broadcast()}
```
apple502j:
```suggestion
	 * An event triggered when a player sends a chat message. Listeners on this event
	 * are triggered for every player receiving that message. If any listener attached returns
	 * {@code false} the message body will not be sent to that specific player. However,
	 * due to client-side limitations, the header (which includes the sender profile)
	 * will always be sent. Note that this only applies to messages sent by a player,
	 * not other entities (via {@code /execute}).
	 *
	 * <p>If a listener returns {@code false}, the server will not broadcast the message body
	 * to that player and no further listeners will be run.
	 *
	 * <p>This event is triggered after {@link #CHAT_MESSAGE}, {@link #COMMAND_MESSAGE},
	 * or their allow variant. If those events blocked the message body from being sent,
	 * then this event will not be triggered.
```

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageEvents.java
```diff
@@ -190,6 +211,25 @@ public interface AllowCommandMessage {
...
 
+	@FunctionalInterface
+	public interface AllowMessageToPlayer {
+		/**
+		 * Called when a player is going to be sent a message either from the chat GUI or a
+		 * command (including player specific commands such as {@code /msg}). Returning false
+		 * prevents only this player from receiving the message. This does not block the
+		 * {@link #COMMAND_MESSAGE} or {@link #CHAT_MESSAGE} from sending.
+		 *
+		 * <p>This only applies to messages sent by a player.w
```
apple502j:
```suggestion
		 * Called when a player is going to be sent a message either from the chat GUI or a
		 * command (including player specific commands such as {@code /msg}). Returning {@code false}
		 * prevents only this player from receiving the message body. However,
		 * due to client-side limitations, the header (which includes the sender profile)
		 * will always be sent.
		 *
		 * <p>This only applies to messages sent by a player, not other entities (via {@code /execute}).
```

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/api/message/v1/ServerMessageEvents.java
```diff
@@ -190,6 +211,25 @@ public interface AllowCommandMessage {
...
+		 *
+		 * @param message the broadcast message with message decorators applied; use {@code message.getContent()} to get the text
+		 * @param sender the uuid of the player that sent the message
```
apple502j:
```suggestion
		 * @param sender the UUID of the player that sent the message
```

fabric-message-api-v1/src/main/java/net/fabricmc/fabric/mixin/message/ServerPlayerEntityMixin.java
```diff
@@ -0,0 +1,43 @@
...
+import net.fabricmc.fabric.api.message.v1.ServerMessageEvents;
+
+@Mixin(ServerPlayerEntity.class)
+public class ServerPlayerEntityMixin {
+	@Inject(method = "sendChatMessage", at = @At(value = "HEAD"), cancellable = true)
+	public void sendChatMessage(SentMessage message, boolean filterMaskEnabled, MessageType.Parameters params, CallbackInfo ci) {
+		if (message instanceof SentMessage.Profileless) return;
+
+		SignedMessage signedMessage = ((SentMessageChatAccessor) message).getMessage();
+		UUID sender = ((SentMessageChatAccessor) message).getMessage().createMetadata().sender();
```
apple502j:
```suggestion
		UUID sender = signedMessage.signedHeader().sender();
```

apple502j (changes requested):

PR #2486, opened 2022-08-26
Refactor Dimension API and fix a bug

apple502j:
Closes #2474

- Refactors dimension API to store the current target in an entity's field instead of a static field. While memory leak is not a concern, there is a potential issue as described in 2474; however, I'm pretty sure that cannot happen without calling impl code.
- Fixes a bug where end spawn platform gets created when teleporting to the End, if the entity is not a player. This fix also consolidates the 3 mixins into one.
- Add a testmod for teleporting to any dimension.


IMB11:
> Fixed bug when platform is generated.

Pretty sure that's vanilla

apple502j:
@mineblock11 This is inconsistent with players, which if teleported using this API do not generate the obsidian platform.

modmuss50 (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/Teleportable.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.dimension;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.world.TeleportTarget;
+
+public interface Teleportable {
```
Technici4n, line 25: I'd recommend adding `@ApiStatus.Internal`.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionInternals.java
```diff
@@ -23,32 +23,18 @@
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.world.TeleportTarget;
 
-import net.fabricmc.fabric.api.dimension.v1.FabricDimensions;
-
 public final class FabricDimensionInternals {
```
Technici4n, line 28: Could also have `@ApiStatus.Internal`.

Technici4n (approved):
Good on principle.

PR #2492, opened 2022-08-28
API Lookup API: fail safe when block-BE desync occurs

apple502j:
There is an extremely rare chance that the block entity is different from the one associated with the current block state. For example, if the block state is of a chest but the block entity is a mob spawner, the `(T)` cast below would otherwise fail when a mod registers an API for chests since it tries to cast `MobSpawnerBlockEntity` to `ChestBlockEntity`.

See also: `BlockWithEntity#checkType` https://maven.fabricmc.net/docs/yarn-1.19.2+build.8/net/minecraft/block/BlockWithEntity.html#checkType(net.minecraft.block.entity.BlockEntityType,net.minecraft.block.entity.BlockEntityType,net.minecraft.block.entity.BlockEntityTicker)

### DIFF
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
@@ -218,7 +218,16 @@ public interface BlockApiLookup<A, C> {
 	 */
 	@SuppressWarnings("unchecked")
 	default <T extends BlockEntity> void registerForBlockEntity(BiFunction<? super T, C, @Nullable A> provider, BlockEntityType<T> blockEntityType) {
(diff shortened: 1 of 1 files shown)


Juuxel (approved):

Technici4n:
As I said on Discord, I don't think mods should be expected to deal with such desyncs everywhere and silently failing when appropriate, so I don't think Fabric API should guard against it.
image

Your thoughts?

apple502j:
The "root cause" AFAIK is a vanilla issue, but I don't know the exact detail and I don't think that should be in FAPI. (I do know that some worldgen glitches caused that in the past and that vanilla game almost never does unchecked cast for BlockEntity.)

Mods generally don't care, but there's also little reason to not merge this. Maybe this is a wontfix, you can decide.

Technici4n:
Now that the BE is set as soon as the block is set everywhere, I think these checks are leftovers from legacy code.

Closing as I don't think this is a good practice. (Might reconsider later if these issues start occurring again).

PR #2493, opened 2022-08-29
Fabric Config API v0 

IMB11:
**TODO:**

- [ ] Config Serialization
- [ ] Structure Definition
   + [ ] Annotations
   + [ ] Manual Definition
- [ ] UI Widget Wrappers
- [ ] Testmod

<hr>

Discussion on the Fabric discord has brought up numerous questions on a centralised configuration API that should be part of Fabric API.

A few points that were recommended:

- NO Automatic Config Screens

Instead of automatic config screens being created, config widgets should be provided (`FabricBooleanConfigWidget` as ex.) to ease creation of config screens.

- Swappable Serializers

The API by default will provide TOML, JSON and YML Serializers, the users may extend the `AbstractConfigSerializer` class to create their own Serializers.

- Config classes being registered at `FabricConfigRegistry` 

- Config files should be found in the config folder, as usual, in the format where the mod ID is the file name.

<hr>

How will this work?

> Will update this with the test mod examples on Saturday, this is a draft PR for a reason because it's still WIP.

1) Users can either use annotations, or define their config structure via other means.

**Annotations**:

Users will create a class and mark it using the `FabricConfig` annotation, inside this class, all fields will be serialized into the config file. You can exclude fields and non-static sub classes with the `NoConfig` annotation.

**Manual Defining Structure**:

Unsure - [TBD] - Open for discussion 

2) Users will then register this class to the `FabricConfigRegistry` during mod init - like so:

The user would also specify the Serializer, by default this would be JSON.

```java
FabricConfigRegistry.register(new Identifier(...), MyConfigClass.class, new YMLSerializer());
```

4) To access the config instance, the user would call:

```java
FabricConfigRegistry.get<MyConfigClass>(new Identifier(...));
```
The user could wrap this call as a static method in their config class for ease of access.

<hr>

**FAQ**

I'll add to this list as questions come in via Discord and on this PR.

Q: Isn't this a copy of forge?

This has been asked a few times when proposing a config API, just because forge has a built-in config API, does not mean that Fabric API can't have one.

Q: There are plenty of other user created config APIs, why would we even need to add one to Fabric API?

User created config APIs have their benefits and disadvantages, but to manage mod compatability, it would be best to have a centralised API that is part of Fabric API already.

Q: Why don't you use java properties format? 

Java properties do not support complex objects like JSON, TOML, and YML.

These three formats are already established in modding, with JSON being used majorly by existing fabric mods, TOML being used by multiplatform mods and YML by ports of spigot and Bukkit plugins for compatability with old configs.

> Modified readme.md, just so I could open the PR for discussion, this will be overriden by a force push on the first commit (Which should probably be by Saturday, unable to push right now.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+
+
 # Fabric API
 
 Essential hooks for modding with Fabric.


IMB11:
Heavily a draft, please drop ideas and suggestions if you have any. Would be very helpful.

enjarai:
I often find myself using more advanced features of libraries like Gson. If this is to be widely adopted, it'd probably be important to have support for things like type adapters and nested objects.

On a seperate note, wouldn't this implementation limit the creation of config instances to one per config class? It might be more desirable to use `Identifier`s here for registration and retrieval.

IMB11:
> I often find myself using more advanced features of libraries like Gson. If this is to be widely adopted, it'd probably be important to have support for things like type adapters and nested objects.
> 
> On a seperate note, wouldn't this implementation limit the creation of config instances to one per config class? It might be more desirable to use `Identifier`s here for registration and retrieval.

Updated register/get methods concept to use identifiers.

I honestly am not that familiar with GSON adapters/advanced features, I'll take a look and see what applies to this API.

kwpugh:
Question: Will the new Config API be able to have server-only or client-only config files?  I have often encountered situations where client and server  have different settings in their respective files, leading to unpredictable results.  This may or may not be a stupid question so I apologize in advance if it is.

Daomephsta:
This PR needs more justification of its design decisions.  

> NO Automatic Config Screens

Why not? Config screens are more user-friendly than editing a file directly, and automatic generation means more mods will have config screens.

>  Users can either use annotations, or define their config structure via other means.

If the config structure can be defined without the annotation API, why is the annotation API present? What benefits does it provide, and how do they outweigh the increased maintenance and other costs?

Why an annotation based API specifically? What benefits does it have over alternatives?

> Q: There are plenty of other user created config APIs, why would we even need to add one to Fabric API?
>
> User created config APIs have their benefits and disadvantages, but to manage mod compatability, it would be best to have a centralised API that is part of Fabric API already.

This boils down to "to manage mod compatibility", which is broad and vague. 
What does managing mod compatibility mean? Why does a 1st party config API do so better than 3rd party APIs?

Note that I'm playing devil's advocate, these questions don't reflect my ideas of what a config API needs.

IMB11:
That's a pretty decent question, this could probably be done via entrypoints:

Register server config in Dedicated server entrypoint.

Register client config in client entrypoint.

Daomephsta:
Now for my actual opinions
> Config files should be found in the config folder, as usual, in the format where the mod ID is the file name

This is a good default, but it's often cleaner or even necessary for a mod to have multiple config files. I've split a config file before because one part needed to be loaded before the game was ready to load a different part.

IMB11:
> This PR needs more justification of its design decisions.  
> 
> > NO Automatic Config Screens
> 
> Why not? Config screens are more user-friendly than editing a file directly, and automatic generation means more mods will have config screens.
> 
> >  Users can either use annotations, or define their config structure via other means.
> 
> If the config structure can be defined without the annotation API, why is the annotation API present? What benefits does it provide, and how do they outweigh the increased maintenance and other costs?
> 
> Why an annotation based API specifically? What benefits does it have over alternatives?
> 
> > Q: There are plenty of other user created config APIs, why would we even need to add one to Fabric API?
> >
> > User created config APIs have their benefits and disadvantages, but to manage mod compatability, it would be best to have a centralised API that is part of Fabric API already.
> 
> This boils down to "to manage mod compatibility", which is broad and vague. 
> What does managing mod compatibility mean? Why does a 1st party config API do so better than 3rd party APIs?
> 
> Note that I'm playing devil's advocate, these questions don't reflect my ideas of what a config API needs.

The general consensus was that automatic config screens are repetitive and would make the API more complex than nessecary - however, thinking about it, it's probably better to have automatic config screens.

Automated config screens would also add the following questions:

- How will the user show the screen to the player?
- Will they have to depend on mod menu to do so?
- How would tooltips, descriptions and other key information on config values be managed?

1st party config APIs define the standards, unlike 3rd party APIs, standards defined by 1st party won't just vanish or be randomly marked deprecated - that's really it tbh, you wouldn't have 200+ mods installed that use 6 different config APIs.

The annotation API would be more efficient and simpler to understand. Some people may prefer annotations over having to define the structure via a builder or whatever is implemented for manual definition.

Annotations would probably allow automated definition, for cases where the structure doesn't really matter as long as configurations are editable without bother.

Manual definition would be useful for dynamic configurations, where for example: You have config values for specific players, eg: `Player372_color=red`

itsmiir:
i'm personally against the idea of default config screens; i believe that should be the responsibility of dedicated config mods. my opinion is that adding a config screen is roughly equivalent to a screen like ModMenu, and is outside the scope of FAPI.

haykam821:
Codecs should have first-class support.

IMB11:
> i'm personally against the idea of default config screens; i believe that should be the responsibility of dedicated config mods. my opinion is that adding a config screen is roughly equivalent to a screen like ModMenu, and is outside the scope of FAPI.

This is the key debate on auto config screen. FAPI should provide an instance of a config screen, but it should never modify vanilla to make the screen accessible - it should be up to the developer to show the screen.

IMB11:
I'm going to rewrite the PR description to better explain the new structure of the API which I've thought of from the feedback given

itsmiir:
> Now for my actual opinions
> 
> > Config files should be found in the config folder, as usual, in the format where the mod ID is the file name
> 
> 
> 
> This is a good default, but it's often cleaner or even necessary for a mod to have multiple config files. I've split a config file before because one part needed to be loaded before the game was ready to load a different part. 

i agree. flexibility around config paths also makes things like managing config profiles easier. i do think that mods should always have some entry in the the `config/modid` format, if only to point to where the rest of the files are.

IMB11:
I think we agreed on the discord to split this up into loader, and make API provide screen related stuff. Closing

kwpugh:
I see this is closed, is the config api going to happen?

PR #2504, opened 2022-09-04
Fix crash when replacing SimpleRegistry contents

Benonardo:
A deeper analysis can be found at https://discord.com/channels/507304429255393322/807617284129423370/1015895367168036874


apple502j (approved):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -186,7 +186,7 @@ public void setPre(int id, RegistryKey<T> registryId, T object, Lifecycle lifecy
 			RegistryEntry.Reference<T> oldObject = idToEntry.get(registryId.getValue());
 
 			if (oldObject != null && oldObject != object) {
```
Juuxel: Isn't there still a bug here since `oldObject` (a `RegistryEntry.Reference<T>`) can never be identical to a `T`?

apple502j (approved):

modmuss50 (approved):
A testmod would be nice, but this seems fine anyway 👍

PR #2509, opened 2022-09-06
ModNioResourcePack: don't extend AbstractFileResourcePack

JFronny:
When a mod comes across an AbstractFileResourcePack, it should be safe to assume that the pack is a resource pack based on a File (as the name suggests).
That is not true in this implementation as the attached "base" File is null.
This PR adjusts ModNioResourcePack to no longer extend AbstractFileResourcePack and instead implement the small number of features it provided manually.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
@@ -42,7 +42,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import net.minecraft.resource.AbstractFileResourcePack;
+import net.minecraft.resource.ResourcePack;
 import net.minecraft.resource.ResourceType;
+import net.minecraft.resource.metadata.ResourceMetadataReader;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.InvalidIdentifierException;
 
@@ -51,9 +53,9 @@ import net.fabricmc.fabric.api.resource.ResourcePackActivationType;
 import net.fabricmc.loader.api.ModContainer;
 import net.fabricmc.loader.api.metadata.ModMetadata;
 
-public class ModNioResourcePack extends AbstractFileResourcePack implements ModResourcePack {
+public class ModNioResourcePack implements ResourcePack, ModResourcePack {
 	private static final Logger LOGGER = LoggerFactory.getLogger(ModNioResourcePack.class);
-	private static final Pattern RESOURCE_PACK_PATH = Pattern.compile("[a-z0-9-_.]+");
+	private static final Pattern RESOURCE_PACK_PATH = Pattern.compile("[a-z\\d-_.]+");
 	private static final FileSystem DEFAULT_FS = FileSystems.getDefault();
 
 	private final Identifier id;
@@ -94,8 +96,6 @@ public class ModNioResourcePack extends AbstractFileResourcePack implements ModR
 	}
 
 	private ModNioResourcePack(Identifier id, String name, ModMetadata modInfo, List<Path> paths, ResourceType type, AutoCloseable closer, ResourcePackActivationType activationType) {
-		super(null);
-
 		this.id = id;
 		this.name = name;
 		this.modInfo = modInfo;
@@ -183,7 +183,6 @@ public class ModNioResourcePack extends AbstractFileResourcePack implements ModR
 		return !namespaces.get(type).contains(filename.substring(prefixLen, nsEnd));
 	}
 
-	@Override
 	protected InputStream openFile(String filename) throws IOException {
 		InputStream stream;
 
@@ -204,13 +203,17 @@ public class ModNioResourcePack extends AbstractFileResourcePack implements ModR
 	}
 
 	@Override
-	protected boolean containsFile(String filename) {
-		if (ModResourcePackUtil.containsDefault(modInfo, filename)) {
-			return true;
+	public InputStream openRoot(String fileName) throws IOException {
+		if (!fileName.contains("/") && !fileName.contains("\\")) {
+			return this.openFile(fileName);
+		} else {
+			throw new IllegalArgumentException("Root resources can only be filenames, not paths (no / allowed!)");
 		}
+	}
 
-		Path path = getPath(filename);
-		return path != null && Files.isRegularFile(path);
+	@Override
+	public InputStream open(ResourceType type, Identifier id) throws IOException {
+		return openFile(getFilename(type, id));
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -51,9 +53,9 @@
 import net.fabricmc.loader.api.ModContainer;
 import net.fabricmc.loader.api.metadata.ModMetadata;
 
-public class ModNioResourcePack extends AbstractFileResourcePack implements ModResourcePack {
+public class ModNioResourcePack implements ResourcePack, ModResourcePack {
 	private static final Logger LOGGER = LoggerFactory.getLogger(ModNioResourcePack.class);
-	private static final Pattern RESOURCE_PACK_PATH = Pattern.compile("[a-z0-9-_.]+");
+	private static final Pattern RESOURCE_PACK_PATH = Pattern.compile("[a-z\\d-_.]+");
```
apple502j: Why this change? Won't this allow non-ASCII paths? (like Persian numerals)

JFronny: Idea said it was equivalent

Juuxel: ...why would you change it then? 🤔

JFronny: You mean why I reverted that change?
It's not really related to this PR and idea might just be recommending the change for international compatibility
The reason why I changed it in the first place was that it was recommended by Idea.

Juuxel: I was wondering more about why you changed it originally since the ascii pattern is perfectly clear, but yeah it makes sense to revert it here

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -51,7 +53,7 @@
 import net.fabricmc.loader.api.ModContainer;
 import net.fabricmc.loader.api.metadata.ModMetadata;
 
-public class ModNioResourcePack extends AbstractFileResourcePack implements ModResourcePack {
+public class ModNioResourcePack implements ResourcePack, ModResourcePack {
```
apple502j, line 59: Since we're editing this file, could you put an `@ApiStatus.Internal` here?

JFronny: I think the fact that it is in the impl package implies that, but sure

apple502j (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -183,7 +185,6 @@ private boolean hasAbsentNs(String filename) {
 		return !namespaces.get(type).contains(filename.substring(prefixLen, nsEnd));
 	}
 
-	@Override
 	protected InputStream openFile(String filename) throws IOException {
```
modmuss50: Can this be private now?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -204,13 +205,17 @@ protected InputStream openFile(String filename) throws IOException {
...
 	@Override
-	protected boolean containsFile(String filename) {
-		if (ModResourcePackUtil.containsDefault(modInfo, filename)) {
-			return true;
+	public InputStream openRoot(String fileName) throws IOException {
+		if (!fileName.contains("/") && !fileName.contains("\\")) {
+			return this.openFile(fileName);
+		} else {
+			throw new IllegalArgumentException("Root resources can only be filenames, not paths (no / allowed!)");
 		}
```
modmuss50: [Nit] Would be slightly cleaner/easier to read to do `if(fileName.contains("/") || fileName.contains("\\")` and then throw. Remove the need for the else.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -252,11 +257,30 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+	}
+
 	@Override
 	public Set<String> getNamespaces(ResourceType type) {
 		return namespaces.getOrDefault(type, Collections.emptySet());
 	}
 
+	@Override
+	public <T> T parseMetadata(ResourceMetadataReader<T> metaReader) throws IOException {
+		try (InputStream is = openRoot("pack.mcmeta")) {
```
modmuss50: Can use `openFile`, no need to validate the path here as we know its fine.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -290,4 +314,8 @@ private static boolean exists(Path path) {
 		// NIO Files.exists is notoriously slow when checking the file system
 		return path.getFileSystem() == DEFAULT_FS ? path.toFile().exists() : Files.exists(path);
 	}
+
+	private static String getFilename(ResourceType type, Identifier id) {
+		return type.getDirectory() + "/" + id.getNamespace() + "/" + id.getPath();
+	}
```
modmuss50: Use String.format same as vanilla, or even an access widener to the vanilla method might be nicer.

modmuss50 (changes requested):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
JFronny: I use the same String.format used in vanilla now

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
JFronny: OK

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
JFronny: Seems to be the case

modmuss50 (approved):

Technici4n (approved):
Had a look at `AbstractFileResourcePack` and this seems fine.

modmuss50:
No need for last call as its not an API change, I will test with a handful of mods before merging.

PR #2518, opened 2022-09-11
Item Group API: refactor & add more javadoc

apple502j:
- Add class javadoc, including code usage example and note on creative inventory search.
- Fix consistency issues with the javadoc.
- Fix "This is used to add an icon *to to* the item group." typo in javadoc.
- Make `identifier` field `final`.
- Add argument null checks.
- When expanding the array, use `System.arraycopy` for copying.


Technici4n (approved):
Looks good, not sure about "Sets" instead of "Set" for example, but it doesn't matter.

modmuss50 (approved):

Juuxel:
@Technici4n The present tense 3rd person wording (like "sets") is what [Yarn's conventions suggest in their examples](https://github.com/FabricMC/yarn/blob/1.19.2/CONVENTIONS.md#javadocs) and what for example the JDK uses.

PR #2437, opened 2022-08-01
Add LandPathNodeTypesRegistry

ghost:
# Content

> A registry for associations between blocks and path node types, for land entities.

This PR adds a new registry to the content registries module, this registry is useful when some mod adds a new block that cause damage and want to flag this block to be avoided by entities.

In vanilla this is done by getting the node type for the block while calculating the path for the entity (every block is a node).

To establish which node is a specific block is done in `LandPathNodeMaker.getCommonNodeType(BlockView world, BlockPos pos)`.  
This is basically a series of ifs, so normally you should add a mixin to this method with an if to give a `PathNodeType` for your block.

This PR adds a way to do this common thing inside Fabric API.  
I use this way in a mod that i'm developing.

``` java
public class LandPathNodeMaker extends PathNodeMaker {
[...]
    protected static PathNodeType getCommonNodeType(BlockView world, BlockPos pos) {
        BlockState blockState = world.getBlockState(pos);

// INJECTED HERE ANOTHER IF AND RETURN

        Block block = blockState.getBlock();
        Material material = blockState.getMaterial();
        if (blockState.isAir()) {
            return PathNodeType.OPEN;
        } else if (!blockState.isIn(BlockTags.TRAPDOORS) && !blockState.isOf(Blocks.LILY_PAD) && !blockState.isOf(Blocks.BIG_DRIPLEAF)) {
            if (blockState.isOf(Blocks.POWDER_SNOW)) {
                return PathNodeType.POWDER_SNOW;
            } else if (blockState.isOf(Blocks.CACTUS)) {
                return PathNodeType.DAMAGE_CACTUS;
            } else if (blockState.isOf(Blocks.SWEET_BERRY_BUSH)) {
                return PathNodeType.DAMAGE_OTHER;
            } else if (blockState.isOf(Blocks.HONEY_BLOCK)) {
                return PathNodeType.STICKY_HONEY;
            } else if (blockState.isOf(Blocks.COCOA)) {
                return PathNodeType.COCOA;
            } else {
                FluidState fluidState = world.getFluidState(pos);
                if (fluidState.isIn(FluidTags.LAVA)) {
                    return PathNodeType.LAVA;
                } else if (inflictsFireDamage(blockState)) {
                    return PathNodeType.DAMAGE_FIRE;
                } else if (DoorBlock.isWoodenDoor(blockState) && !(Boolean)blockState.get(DoorBlock.OPEN)) {
                    return PathNodeType.DOOR_WOOD_CLOSED;
[...]

PR #2519, opened 2022-09-11
Add way to specify the PathNodeType for a block, add missing tests in content registries testmod.

ghost:
This is the same as https://github.com/FabricMC/fabric/pull/2437
See discussion in https://github.com/FabricMC/fabric/pull/2437

- [x] ~Lithium causes crash because overwrites the mixin.~ (fixed on fabric side by lowering mixin priority to 999, needs an update on lithium side to remove the overwrite mixin (see [PR #392](https://github.com/CaffeineMC/lithium-fabric/pull/392))
- [x] Separate advanced (with world and pos) from main registry, and add a way to check if a block is registered with advanced, or simple, node type customization (this is useful for performance mods).

# Summary

This pr adds a way to specify that a specific block will cause damage, and entities should avoid that block.
In vanilla this is done by getting the node type for the block while calculating the path for the entity (every block is a node).

You can also specify what an entity will think aboud the block when finds it in a neighbor position in his path, while traveling the world terrain.

# Testing

Dead Bush is marked as dangerous, with the same value as berry bush, every entity (except foxes and bees) avoid it.

https://user-images.githubusercontent.com/32025549/182221967-e24de458-94d0-47d3-8ca9-6e513223cdfb.mp4


fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Registers a {@link PathNodeType} for the specified block.
+	 * This will override the default block behavior.
+	 * For example, you can make a safe block as dangerous and vice-versa.
+	 * Duplicated registrations for the same block will replace the previous registration.
+	 *
+	 * @param block              Block to register.
+	 * @param nodeType           {@link PathNodeType} to associate with the block.
+	 *                           (Pass {@code null} to not specify a node type and use the default behavior)
+	 * @param nodeTypeIfNeighbor {@link PathNodeType} to associate to the block, if is a neighbor block in the path.
```
kevinthegreat1: Sentence grammar and consistency of "associate with"

```suggestion
	 * @param nodeTypeIfNeighbor {@link PathNodeType} to associate with the block, if it is a neighbor block in the path.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+import org.slf4j.LoggerFactory;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.ai.pathing.PathNodeType;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockView;
+
+/**
+ * A registry to associate specific node types to blocks.
```
kevinthegreat1: Maybe this is clearer in that it shows you can only assign one path node type to a block.
Also consistency of "associate with"

```suggestion
 * A registry to associate a block with a specific path node type.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/PathNodeTypeProvider.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A functional interface that provides the {@link PathNodeType}, given the block state and position.
+ */
+@FunctionalInterface
+public interface PathNodeTypeProvider {
+	/**
+	 * Gets the {@link PathNodeType} for the specified block at the specified position.
+	 *
+	 * <p>You can specify what to return if the block is a direct target of an entity path,
+	 * or a neighbor block that the entity will find in the path.
```
kevinthegreat1: Maybe this is clearer and more consistent with the previous line.

```suggestion
	 * or a neighbor block of the entity path.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/PathNodeTypeProvider.java
```diff
@@ -0,0 +1,51 @@
...
+	 *
+	 * <p>For example, for a cactus-like block you should specify {@link PathNodeType#DAMAGE_CACTUS} if the block
+	 * is a direct target ({@code neighbor == false}) to specify that an entity should not pass through or above
+	 * the block because it will cause damage, and {@link PathNodeType#DANGER_CACTUS} if the cactus will be found
+	 * as a neighbor block in the entity path ({@code neighbor == true}) to specify that the entity should not get
+	 * close to the block because is dangerous.
```
kevinthegreat1: Made sentence easier to read and added it

```suggestion
	 * <p>For example, for a cactus-like block you should use {@link PathNodeType#DAMAGE_CACTUS} if the block
	 * is a direct target in the entity path ({@code neighbor == false}) to specify that an entity should not pass through
	 * or above the block because it will cause damage, and you should use {@link PathNodeType#DANGER_CACTUS}
	 * if the block is a neighbor block in the entity path ({@code neighbor == true}) to specify that the entity should not
	 * get close to the block because it is dangerous.
```

kevinthegreat1 (commented):
Suggested some grammar changes

ghost:
@kevinthegreat1 i've changed almost all files, please re-check again for grammar errors.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FabricAbstractBlock.java
```diff
@@ -0,0 +1,105 @@
...
+	@Nullable
+	default PathNodeType getPathNodeType(BlockState state) {
+		return null;
+	}
+
+	/**
+	 * Gets the {@link PathNodeType} for the block if is in a closest neighbor position
+	 * that the entity will find in his path.
+	 *
+	 * <p>This returns the {@link PathNodeType} if the block is a direct target in an entity path.
```
kevinthegreat1: I think this is more clear. Also shouldn't this be neighbor here?

```suggestion
	 * Gets the {@link PathNodeType} for the block if it is a direct neighbor position
	 * to an entity path (directly next to a block that the entity will pass through or above).
	 *
	 * <p>This returns the {@link PathNodeType} if the block is a direct neighbor in an entity path.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FabricAbstractBlock.java
```diff
@@ -0,0 +1,105 @@
...
+		 */
+		@Nullable
+		default PathNodeType getPathNodeType() {
+			AbstractBlock.AbstractBlockState state = (AbstractBlock.AbstractBlockState) this;
+			return state.getBlock().getPathNodeType(state.asBlockState());
+		}
+
+		/**
+		 * Gets the {@link PathNodeType} for the block state if is in a directly neighbor position
+		 * that the entity will find in his path.
```
kevinthegreat1:
```suggestion
		 * Gets the {@link PathNodeType} for the block state if it is a direct neighbor position
	     * to an entity path (directly next to a block that the entity will pass through or above).
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/FabricAbstractBlock.java
```diff
@@ -0,0 +1,105 @@
...
+	 * via Mixin and interface injection.
+	 *
+	 * <p>Note to maintainers: Functions should only be added to this interface if they are general-purpose enough,
+	 * to be evaluated on a case-by-case basis. Otherwise, they are better suited for more specialized APIs.
+	 */
+	interface FabricAbstractBlockState {
+		/**
+		 * Gets the {@link PathNodeType} for the block state.
+		 *
+		 * @return the {@link PathNodeType} specified for the block state.
```
kevinthegreat1: Clarify that this is for direct targets of a path

```suggestion
		 * Gets the {@link PathNodeType} for the block state if it is a direct target in an entity path.
		 *
		 * @return the {@link PathNodeType} specified for the block state.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/LandPathNodeMakerMixin.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.ai.pathing.LandPathNodeMaker;
+import net.minecraft.entity.ai.pathing.PathNodeType;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockView;
+
+@Mixin(value = LandPathNodeMaker.class, priority = 999)
+public class LandPathNodeMakerMixin {
+	/**
+	 * Overrides the node type for the specified position in a path.
```
kevinthegreat1: Clarify?

```suggestion
	 * Overrides the node type for the specified position, if the position is a direct target in a path.
```

kevinthegreat1 (commented):
I guess here are some grammar suggestions.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/LandPathNodeMakerMixin.java
```diff
@@ -0,0 +1,57 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.ai.pathing.LandPathNodeMaker;
+import net.minecraft.entity.ai.pathing.PathNodeType;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.BlockView;
+
+// Applied a bit earlier than other mods to make possible to optimize default vanilla behavior
```
kevinthegreat1:
```suggestion
// Applied a bit earlier than other mods to ensure changes and optimizations to default vanilla behavior
```

Technici4n:
Pepper commented the following on Discord:
> Will a method with BlockPos and level be added?
I read that not passing those allows the type to be cached, which is true, but what if that is not enough context? Both paths should exist.

Thoughts?

ghost:
I have an idea about this, i will push for review tomorrow.
- [x] ~Will a method with BlockPos and level be added?~ reworked including this too

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/LandPathNodeMakerMixin.java
```diff
@@ -0,0 +1,61 @@
...
+			cir.setReturnValue(nodeType);
+		}
+	}
+
+	/**
+	 * Overrides the node type for the specified position, if the position is found as neighbor block in a path.
+	 */
+	@Inject(method = "getNodeTypeFromNeighbors", at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/world/BlockView;getBlockState(Lnet/minecraft/util/math/BlockPos;)Lnet/minecraft/block/BlockState;"), locals = LocalCapture.CAPTURE_FAILHARD, cancellable = true)
+	private static void getNodeTypeFromNeighbors(BlockView world, BlockPos.Mutable pos, PathNodeType nodeType, CallbackInfoReturnable<PathNodeType> cir, int i, int j, int k, int l, int m, int n, BlockState state) {
+		PathNodeTypeContext ctx = new PathNodeTypeContext(state, world, pos);
```
modmuss50: This is now allocating a new object, as perf is a concern here I wonder if there is a better way that could avoid the allocation? Or am I just being pedantic?

ghost: I've reworked by mixing the two solutions: made possible to understand if a node type is dynamic or static and used hash table instead of injected interfaces.

ghost:
#### Brief resume of the last changes (20 September)
I've used the previous solution (with a registry instead of injected interfaces), but now there are two registration methods (for dynamic and static registration) + some new methods to check if a block is registered and if it is registered as static or dynamic.  
I've also reverted by using dead bush for testing (this is only a detail).

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,304 @@
...
+		//Gets the node type from the registered provider.
+		if (provider instanceof StaticPathNodeTypeProvider staticProvider) {
+			return staticProvider.getPathNodeType(state, neighbor);
+		}
+
+		//If no valid provider is found returns null.
+		return null;
+	}
+
+	/**
```
Technici4n, line 144: All the methods below this exist mostly for Lithium, so I think they should be kept to a minimal API surface.
I wonder if we couldn't replace them all by a single `@Nullable PathNodeTypeProvider getProvider(Block block)` ? I'm not sure that calling such a method from outside the class would be allowed.

ghost: Immagine 2022-10-01 011821

I don't want to expose `PathNodeTypeProvider` to avoid further extensions that could break everything.  
I thought... maybe a better way is to replace them with a single method `int getRegistrationType(Block block)` that returns 0 if the block is not registered, 1 if is registered static, 2 if is registered dynamic.  
These values may be enum constants.  
And yes i think only lithium will use this method.

I've also considered that this can lead to a double search (if you want to use the checker method first, and then the getter method after), however with the lithium caching system with flags this must be done only one time when the block is find for the first time after server loading.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,277 @@
...
+			return dynamicProvider.getPathNodeType(state, world, pos, neighbor);
+		} else if (provider instanceof StaticPathNodeTypeProvider staticProvider) {
+			//This can return null too.
+			return staticProvider.getPathNodeType(state, neighbor);
+		}
+
+		//If no valid provider is found returns null.
+		return null;
+	}
+
```
Technici4n, line 143: I think we can have a lower-level API here, just because I don't like all the boilerplate that we need:

(sketch)
```java
/**
 * Return the implementation of the provider registered for a specific block, or {@code null} if nothing is registered.
 *
 * <p>If non-null, the returned object can either be a {@link StaticPathNodeTypeProvider} or a {@link DynamicPathNodeTypeProvider} at the moment depending on what was registered.
 * More kinds of providers might be added if the API is expanded in the future, so make sure not to fail without exceptions if another type of object is returned.
 */
@Nullable
public static Object getProvider(Block block) {
    Objects.requireNonNull(block);
    return NODE_TYPES.get(block);
}
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,277 @@
...
+		}
+
+		//If no valid provider is found returns NONE.
+		return ProviderType.NONE;
+	}
+
+	/**
+	 * General provider, this is a marker interface.
+	 */
+	private interface PathNodeTypeProvider {
```
Technici4n: Would it be sensible to seal this interface?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
ghost: Ok done.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
ghost: Done.
And i learned a new thing. 👍

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -0,0 +1,223 @@
...
+	 * <p>Tip: This method is intended to be used in any cases in which you need to get
+	 * the raw provider for the block, if you need the {@link PathNodeType} for the block state instead,
+	 * you can simply use {@link #getPathNodeType}.
+	 *
+	 * @param block Current block.
+	 * @return the {@link PathNodeTypeProvider} registered for the specified block,
+	 * or {@code null} if no provider is registered for the block.
+	 */
+	@Nullable
+	public static PathNodeTypeProvider getPathNodeTypeProvider(Block block) {
```
Technici4n, line 165: Looks great. Could you maybe add a warning saying that we reserve the right to add more subinterfaces, and that users of this method must take that into account?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/LandPathNodeTypesRegistry.java
```diff
@@ -142,11 +142,17 @@ public static PathNodeType getPathNodeType(BlockState state, BlockView world, Bl
...
+	 * Gets the raw {@link PathNodeTypeProvider} registered for the specified block.
 	 *
 	 * <p>If no {@link PathNodeTypeProvider} is registered for the block, it returns {@code null}.
 	 *
-	 * <p>Tip: This method is intended to be used in any cases in which you need to get
+	 * <p>Note 1: {@link PathNodeTypeProvider} is a marker interface with no methods,
+	 * so you need to cast the result to a subtype, in order to get something from it.
+	 * Currently, the result can be of {@link StaticPathNodeTypeProvider} or {@link DynamicPathNodeTypeProvider}.
+	 * Note that more kinds of providers might be added if the API is expanded in the future,
+	 * so make sure not to fail without exceptions if another type of object is returned.
```
Technici4n: Nitpick: the double `not`/`without` negation makes it positive again.

Technici4n (approved):
Thanks a lot for your efforts. 😅

modmuss50 (approved):
Looks great, thanks a lot for reworking this 👍

modmuss50:
Thanks a lot for this @devpelux and especially for putting in the effort to sort out the issues we found last min when we first merged this. 👍

PR #2521, opened 2022-09-11
Fabric Data Generation: FabricSoundProvider

IMB11:
#### **`sounds.json` Entry Helper | [SoundEventEntry](https://github.com/mineblock11/fabric/blob/data-providerathon/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/helpers/SoundEventEntry.java)**

Mojang does not specifically provide codecs for the sounds.json format, since the `sounds.json` provider (FabricSoundsProvider) required a class to store data on entries, I have provided a few codecs that may be useful to modders because the entry helper utilizes the same structure defined by Minecraft. 

More info on structure on [Minecraft Wiki](https://minecraft.fandom.com/wiki/Sounds.json)

#### `FabricSoundsProvider`

> tbd, still in progress

### DIFF
--- a/fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BiomeTagGenerator.java
+++ b/fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BiomeTagGenerator.java
@@ -22,7 +22,7 @@ import net.minecraft.world.biome.Biome;
 import net.minecraft.world.biome.BiomeKeys;
 
 import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
-import net.fabricmc.fabric.api.datagen.v1.provider.data.FabricTagProvider;
+import net.fabricmc.fabric.api.datagen.v1.provider.FabricTagProvider;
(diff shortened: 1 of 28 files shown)


Juuxel:
I'd avoid the breaking changes, the restructuring doesn't seem necessary here.

IMB11:
Its getting very large though - are package restructuring considered breaking changes? Functionality stays the same?

Juuxel:
Yeah, they're breaking changes - running mods built against older versions would crash and compiling old code against the new version would fail.

edit: Ah yes, github is breaking again

IMB11:
You've just gotta update the imports, unless that causes problems - which it shouldn't?

Juuxel:
Yeah, they're breaking changes - running mods built against older versions would crash and compiling old code against the new version would fail.

IMB11:
Ah I see - okay, i'll sort it out then

IMB11:
Reverted

apple502j:
You can probably use the new package style for new stuff.

IMB11:
Closing, will re-PR for hacktoberfest because why not? :D

modmuss50:
I don’t think that’s a good idea. You loose all of the comment history. Makes it harder to find the discussion in the future.

PR #2525, opened 2022-09-13
Split vanilla play packets if it's too big

deirn:
Resolves `Packet too big (is x, should be less than 8388608)` error, most commonly caused by `SynchronizeRecipesS2CPacket` becoming too big because of too many recipes, but any packet can cause this, especially if NBT is involved.

`CustomPayload*Packet` (mod packets) are not affected by this PR and will still error if it exceeds the limit (1mb for S2C, 32kb for C2S), should Fabric also handle splitting those?

### DIFF
--- a/fabric-networking-api-v1/build.gradle
+++ b/fabric-networking-api-v1/build.gradle
@@ -6,7 +6,8 @@ moduleDependencies(project, [
 ])
 
 testDependencies(project, [
-	':fabric-command-api-v2',
-	':fabric-lifecycle-events-v1',
-	':fabric-key-binding-api-v1'
+		':fabric-command-api-v2',
+		':fabric-lifecycle-events-v1',
+		':fabric-key-binding-api-v1',
+		':fabric-screen-handler-api-v1'
 ])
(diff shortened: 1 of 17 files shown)


apple502j:
This is hard to maintain and does clearly change the behavior that oversized packets kick the player/are discarded. (Those are ones that I believe we should keep. We aren't Spigot.)
We should split recipe packets only, if at all, and if we do we need to make it compatible with proxies. Oversized NBT has other problems that are more complicated.

modmuss50:
I agree, I think this is out of scope of what Fabric API should be doing. As @apple502j pointed out this raises some major compability concerns with 3rd party tools. As well as possibly breaking compaiblity with vanilla or older fabric api clients.

I would be more open to ensuring our networking API provides the tools to allow a seperate mod achieve this. I also dont think splitting specific packets like the recipes is a good solution either. Ideally this would be fixed by mojang (Its quite possible to hit the limit with a datapack in vanilla).

Technici4n:
> I also dont think splitting specific packets like the recipes is a good solution either. Ideally this would be fixed by mojang (Its quite possible to hit the limit with a datapack in vanilla).

We can't just wait for mojang to fix stuff if our users are having an issue. There's a lot more recipes in modded than in the typical vanilla + datapacks scenario. IMO

modmuss50:
Many thanks for the PR, but im going to close it as it is out of scope for API. There are issues with some vanilla packets growing too large with many mods, im not quite sure what the best solution is. Sadly I dont think this is it.

PR #2533, opened 2022-09-15
Add ServerEntityEvents.EquipmentChange event

### LINKED ISSUE #990: Add event for when a LivingEntity equips or unequips an item
This could be useful for e.g. stats that depend on current equipment and need to be recalculated when they change.

modmuss50:
Very simple event on the LivingEntity equipment change, needed to remove a mixin from my (and other people's) mod :)

This is only fired on the server and also fires when the entity joins the world.

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
@@ -101,6 +101,6 @@ public final class ServerEntityEvents {
 
 	@FunctionalInterface
 	public interface EquipmentChange {
-		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack nextStack);
+		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack currentStack);
 	}
 }
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
@@ -17,6 +17,9 @@
 package net.fabricmc.fabric.api.event.lifecycle.v1;
 
 import net.minecraft.entity.Entity;
+import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.util.profiler.Profiler;
 
@@ -75,6 +78,17 @@ public final class ServerEntityEvents {
 		}
 	});
 
+	/**
+	 * Called when an Entity's equipment is changed.
+	 *
+	 * <p>This event is also called when the entity joins the world.
+	 */
+	public static final Event<EquipmentChange> EQUIPMENT_CHANGE = EventFactory.createArrayBacked(ServerEntityEvents.EquipmentChange.class, callbacks -> (livingEntity, equipmentSlot, previous, next) -> {
+		for (EquipmentChange callback : callbacks) {
+			callback.onChange(livingEntity, equipmentSlot, previous, next);
+		}
+	});
+
 	@FunctionalInterface
 	public interface Load {
 		void onLoad(Entity entity, ServerWorld world);
@@ -83,5 +97,10 @@ public final class ServerEntityEvents {
 	@FunctionalInterface
 	public interface Unload {
 		void onUnload(Entity entity, ServerWorld world);
+	}
+
+	@FunctionalInterface
+	public interface EquipmentChange {
+		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack nextStack);
 	}
 }
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
@@ -79,7 +79,7 @@ public final class ServerEntityEvents {
 	});
 
 	/**
-	 * Called during {@link LivingEntity#tick()} after the Entity's equipment has been changed.
+	 * Called during {@link LivingEntity#tick()} if the Entity's equipment has been changed.
 	 *
 	 * <p>This event is also called when the entity joins the world.
 	 */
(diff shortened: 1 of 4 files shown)


fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
```diff
@@ -84,4 +98,9 @@ public interface Load {
 	public interface Unload {
 		void onUnload(Entity entity, ServerWorld world);
 	}
+
+	@FunctionalInterface
+	public interface EquipmentChange {
+		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack nextStack);
```
Technici4n:
```suggestion
		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack currentStack);
```

Could also use a javadoc comment maybe to explain that this method is called once per slot etc.

Technici4n (approved):
Looks perfect !

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
```diff
@@ -75,6 +78,17 @@ private ServerEntityEvents() {
 		}
 	});
 
+	/**
+	 * Called during {@link LivingEntity#tick()} if the Entity's equipment has been changed.
```
haykam821: This documentation could clarify 'changed' as using the `ItemStack.areEqual` method to compare item, count, and NBT.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
```diff
@@ -84,4 +98,9 @@ public interface Load {
 	public interface Unload {
 		void onUnload(Entity entity, ServerWorld world);
 	}
+
+	@FunctionalInterface
+	public interface EquipmentChange {
+		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack currentStack);
```
haykam821, line 105: While the current injection point is problematic for this use case, would it be considered in-scope for this event or another event to potentially reject equipment changes? [Plasmid's `GameRuleType.MODIFY_ARMOR`](https://github.com/NucleoidMC/plasmid/blob/87c4bb18ae02f8d11eafd561a2421bdbbc60e3a0/src/main/java/xyz/nucleoid/plasmid/mixin/game/rule/ServerPlayNetworkHandlerMixin.java#L56-L96), for example, could be greatly simplified in such a case.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/LivingEntityMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerEntityEvents;
+
+@Mixin(LivingEntity.class)
+public abstract class LivingEntityMixin {
+	@Inject(method = "getEquipmentChanges", at = @At(value = "INVOKE", target = "Ljava/util/Map;put(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private void getEquipmentChanges(CallbackInfoReturnable<@Nullable Map<EquipmentSlot, ItemStack>> cir, Map<EquipmentSlot, ItemStack> changes, EquipmentSlot[] slots, int slotsSize, int slotIndex, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack nextStack) {
+		ServerEntityEvents.EQUIPMENT_CHANGE.invoker().onChange((LivingEntity) (Object) this, equipmentSlot, previousStack, nextStack);
```
haykam821:
```suggestion
	private void getEquipmentChanges(CallbackInfoReturnable<@Nullable Map<EquipmentSlot, ItemStack>> cir, Map<EquipmentSlot, ItemStack> changes, EquipmentSlot[] slots, int slotsSize, int slotIndex, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack currentStack) {
		ServerEntityEvents.EQUIPMENT_CHANGE.invoker().onChange((LivingEntity) (Object) this, equipmentSlot, previousStack, currentStack);
```

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerEntityLifecycleTests.java
```diff
@@ -58,6 +58,12 @@ public void onInitialize() {
 			}
 		});
 
+		ServerEntityEvents.EQUIPMENT_CHANGE.register((livingEntity, equipmentSlot, previousStack, nextStack) -> {
+			if (PRINT_SERVER_ENTITY_MESSAGES) {
+				logger.info("[SERVER] Entity equipment change: Entity: {}, Slot {}, Previous: {}, Next {} ", livingEntity, equipmentSlot.name(), previousStack, nextStack);
```
haykam821:
```suggestion
		ServerEntityEvents.EQUIPMENT_CHANGE.register((livingEntity, equipmentSlot, previousStack, currentStack) -> {
			if (PRINT_SERVER_ENTITY_MESSAGES) {
				logger.info("[SERVER] Entity equipment change: Entity: {}, Slot {}, Previous: {}, Current {} ", livingEntity, equipmentSlot.name(), previousStack, currentStack);
```

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
modmuss50: I think this is out of scope for this PR but could definally be a good candidate for something in a new PR. I have only had a quick look but it may be better to also have a client side component for an event that prevents slot usage.

Technici4n:
Note: closes #990.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
modmuss50: Im not too sure this is that useful to know tbh, and might be best kept as an implimentation detail (In code that isnt ours) rarther than something defined in API.

haykam821: I think 'changed' is too vague for a low-level event like this one. From just reading the summary of this event, I thought that this event would only be invoked if equipment was replaced with another stack entirely. Instead, this event is invoked if any equipment stack mutates at all.

modmuss50: Good point, I have updated this to make it more clear 👍

PR #2534, opened 2022-09-15
Add DataGeneratorEntrypoint.getEffectiveModId

modmuss50:
This change is needed to allow other mods to generate data for a different modid. My usecase is that I have a `techreborn-datagen` mod that is not shipped in production builds that generates data for the `techreborn` mod.

This is needed as the Fabric data gen api has a number of places where it uses the modid in the data generation logic such as: https://github.com/FabricMC/fabric/blob/1.19.2/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/ModelProviderMixin.java#L119

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
@@ -35,8 +35,8 @@ public interface DataGeneratorEntrypoint {
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
 
 	/**
-	 * Returns the modid of the mod the data is being generated for.
-	 * A {@code null} return will run the data generator using the modid that registered the current entrypoint.
+	 * Returns the mod ID of the mod the data is being generated for.
+	 * A {@code null} return will run the data generator using the mod ID that registered the current entrypoint.
 	 *
 	 * @return a {@link String} or {@code null}
 	 */
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
@@ -16,6 +16,10 @@
 
 package net.fabricmc.fabric.api.datagen.v1;
 
+import org.jetbrains.annotations.Nullable;
+
+import net.fabricmc.loader.api.ModContainer;
+
 /**
  * An entry point for data generation.
  *
@@ -31,4 +35,15 @@ public interface DataGeneratorEntrypoint {
 	 * @param fabricDataGenerator The {@link FabricDataGenerator} instance
 	 */
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
+
+	/**
+	 * Return the {@link ModContainer} of the mod the data is being generated for.
+	 * A null return will run the data generation using the {@link ModContainer} that registered the current entrypoint.
+	 *
+	 * @return a {@link ModContainer} or null
+	 */
+	@Nullable
+	default ModContainer getEffectiveModContainer() {
+		return null;
+	}
 }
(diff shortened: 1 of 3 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -31,4 +35,15 @@ public interface DataGeneratorEntrypoint {
...
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
+
+	/**
+	 * Return the {@link ModContainer} of the mod the data is being generated for.
+	 * A null return will run the data generation using the {@link ModContainer} that registered the current entrypoint.
+	 *
+	 * @return a {@link ModContainer} or null
+	 */
+	@Nullable
+	default ModContainer getEffectiveModContainer() {
```
modmuss50: On second thoughts it may be better to just return a nullable string of the modid, and find the mod container in the impl.

Technici4n (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -31,4 +33,15 @@ public interface DataGeneratorEntrypoint {
 	 * @param fabricDataGenerator The {@link FabricDataGenerator} instance
 	 */
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
+
+	/**
+	 * Return the modid of the mod the data is being generated for.
+	 * A null return will run the data generation using the modid that registered the current entrypoint.
+	 *
+	 * @return a {@link String} or null
```
apple502j:
```suggestion
	 * Returns the modid of the mod the data is being generated for.
	 * A {@code null} return will run the data generator using the modid that registered the current entrypoint.
	 *
	 * @return a {@link String} or {@code null}
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -31,4 +33,15 @@ public interface DataGeneratorEntrypoint {
 	 * @param fabricDataGenerator The {@link FabricDataGenerator} instance
 	 */
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
+
+	/**
+	 * Return the modid of the mod the data is being generated for.
+	 * A null return will run the data generation using the modid that registered the current entrypoint.
```
apple502j: We should document the behavior for nonexistent mod IDs (right now they crash with `RuntimeException`.)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -31,4 +33,15 @@ public interface DataGeneratorEntrypoint {
 	 * @param fabricDataGenerator The {@link FabricDataGenerator} instance
 	 */
 	void onInitializeDataGenerator(FabricDataGenerator fabricDataGenerator);
+
+	/**
+	 * Returns the modid of the mod the data is being generated for.
+	 * A {@code null} return will run the data generator using the modid that registered the current entrypoint.
```
Juuxel:
```suggestion
	 * Returns the mod ID of the mod the data is being generated for.
	 * A {@code null} return will run the data generator using the mod ID that registered the current entrypoint.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
```diff
@@ -111,8 +112,16 @@ private static void runInternal() {
...
 
 			try {
-				FabricDataGenerator dataGenerator = new FabricDataGenerator(outputDir, entrypointContainer.getProvider(), STRICT_VALIDATION);
-				entrypointContainer.getEntrypoint().onInitializeDataGenerator(dataGenerator);
+				final DataGeneratorEntrypoint entrypoint = entrypointContainer.getEntrypoint();
+				final String effectiveModId = entrypoint.getEffectiveModId();
+				ModContainer modContainer = entrypointContainer.getProvider();
+
+				if (effectiveModId != null) {
+					modContainer = FabricLoader.getInstance().getModContainer(effectiveModId).orElseThrow(() -> new RuntimeException("Failed to find effective mod container for modid (%s)".formatted(effectiveModId)));
```
Juuxel:
```suggestion
					modContainer = FabricLoader.getInstance().getModContainer(effectiveModId).orElseThrow(() -> new RuntimeException("Failed to find effective mod container for mod id (%s)".formatted(effectiveModId)));
```

Juuxel (approved):

PR #2542, opened 2022-09-22
Support namespaced armor textures.

### LINKED ISSUE #96: ArmorMaterial identifier is not mod dependent
I just noticed that when registering new armor materials, their name is not an `Identifier`.
This leads to the strange situation that when implementing own armor sets Minecraft will always search `minecraft:textures/models/armor` for their 3d texture.

Shouldn't Fabric API have an interface for this or at least gain the correct namespace by the item's id?

modmuss50:
I didnt realise this wasnt a thing already untill someone asked me about it! This PR simply supports using a namespace for armor textures in custom ArmorMaterial's.

An alternative would be to extend or interface injection ArmorMaterial to add a method that returns the full `Identifier` however we decided this is a bit overkill.

 Closes #96


Technici4n (approved):
A welcome fix, I'm going to use this in MI as soon as it's released. :smile:

PR #2544, opened 2022-09-23
Fabric Data Fixer API (v1)

apple502j:
Yet another data fixer API proposal. This time, it's a port of data fixer API from QSL, with some internal changes and new methods:

- Unlike QSL, this API saves the data version in a NBT compound `_FabricDataVersions`.
- Current data version can be specified in the `fabric.mod.json` file. (I suggest adding this field in FMJv2.)
- `FabricDataFixerBuilder#build()` overload that uses bootstrap executor.

Files with ported code are marked with `// From QSL.` comment and a special license header.

Fully tested.

## Future Additions?
- [ ] Block entity ID
- [ ] Entity type ID


fabric-data-fixer-api-v1/src/main/java/net/fabricmc/fabric/api/datafixer/v1/EmptySchema.java
```diff
@@ -0,0 +1,56 @@
...
+
+	// Ensure the schema stays empty.
+	@Override
+	public void registerType(boolean recursive, DSL.TypeReference type, Supplier<TypeTemplate> template) {
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	protected Map<String, Type<?>> buildTypes() {
+		return Object2ObjectMaps.emptyMap();
```
Juuxel, line 54: Surely just `Map.of` would be enough since it's empty?

fabric-data-fixer-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,41 @@
...
+  "icon": "assets/fabric-data-fixer-api-v1/icon.png",
+  "contact": {
+    "homepage": "https://fabricmc.net",
+    "irc": "irc://irc.esper.net:6667/fabric",
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC",
+    "QuiltMC"
```
Juuxel, line 17:
```suggestion
    "FabricMC"
```
We haven't marked individual contributors as authors either, so I think just FabricMC is fine

fabric-data-fixer-api-v1/HEADER-PORT
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2016-2022 FabricMC
```
Technici4n, line 2: This requires a QuiltMC copyright line too.

fabric-data-fixer-api-v1/src/main/java/net/fabricmc/fabric/api/datafixer/v1/FabricDataFixes.java
```diff
@@ -0,0 +1,175 @@
...
+		return FabricDataFixesInternals.getModDataVersion(nbt, modId);
+	}
+
+	/**
+	 * Checks if the data fixer registry is frozen. Data fixers cannot be registered
+	 * after the registry gets frozen.
+	 *
+	 * @return {@code true} if frozen, or {@code false} otherwise.
+	 */
+	@Contract(pure = true)
```
Juuxel, line 171: Note: Not a pure function since its value is not constant, but that doesn't seem to be exactly what [`Contract.pure` implies](https://github.com/JetBrains/java-annotations/blob/30daf99ad1a0fa0a75edc61a434efdfa2ebc52f5/common/src/main/java/org/jetbrains/annotations/Contract.java#L70-L89)

Technici4n (commented):
I suppose this will need some discussion on how to organize copyright headers as it's the first port from QSL, and I assume there will be more in the future.

Juuxel (commented):
Going to review the main parts of this PR later, but here are some initial comments

fabric-data-fixer-api-v1/src/main/java/net/fabricmc/fabric/api/datafixer/v1/EmptySchema.java
Technici4n: Also, I don't see the point of using an `Object2Object` map instead of a `HashMap` (except for the custom strategy, or the lack of CME detection? :P).

apple502j: This is copied verbatim - I believe there is no difference. I suppose the vanilla code returns `Object2ObjectMap` (which is apparently faster, if it has contents.)

apple502j:
Note about the scope:

- QSL's API seems to be derived from older PRs in this repo.
- This datafixer API solves data fixing for modded contents, such as renaming a modded block or changing NBT structure. However, this provides no integration for vanilla schemas. In vanilla, all block entities and entities must be registered in a schema, otherwise there is a risk of [world corruption](https://github.com/apple502j/UpdateBlockEntity) when vanilla schema changes. There isn't a good way around this.
- This also does not fix the issue with modded stuff using vanilla schema, and vanilla schema updating.

DataFixer is fundamentally incompatible with the idea of modding - but I do believe that if used correctly, it is one of the most powerful tools we have.

Juuxel:
> This also does not fix the issue with modded stuff using vanilla schema, and vanilla schema updating.

This is a pretty big limitation, considering it makes this mostly usable for breaking mod updates within a single MC version.

apple502j:
Ok, this has a very big flaw:

- Modded block entities must register the schema, that certain serialized fields are for item stacks (AKA inventories).
- There is 1 data fixer per each mod.
- The schemas are registered per data fixer, and only the vanilla schema + data fixer's custom schema are considered when updating.
- So, if anyone puts a modded, to-be-renamed item in a modded inventory from another mod, it won't update, causing world corruption.
- To prevent that, you have to register that schema to all of the data fixers. Wait a minute, isn't that the current state, just with 1 schema for vanilla?

Closing, this API is practically useless.

PR #2545, opened 2022-09-23
[Lifecycles] Client Startup Finished Event

isXander:
## Summary
This event is called when the initial resource reload has completed and the splash-screen is beginning to fade to the title screen.

## Why is it needed?
It may be necessary for some mods to do post-startup tasks like displaying a toast. It guarantees an environment of safety where everything is loaded and the game has actually started up fully. An example of it's use is in [Zoomify](https://github.com/isXander/Zoomify/blob/1.19/src/main/kotlin/dev/isxander/zoomify/Zoomify.kt#L143).

### DIFF
--- a/fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
+++ b/fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
@@ -51,6 +51,18 @@ public final class ClientLifecycleEvents {
 		}
 	});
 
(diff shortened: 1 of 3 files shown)


Technici4n:
As modmuss said on discord:
> Your use case seems better covered by ScreenEvents.AFTER_INIT?

If you agree you can close this PR, otherwise feel free to let us know why. 🙂

isXander:
I feel a dedicated event here would be best because it's easier to find, if I want an event to fire when the game starts, I immediately look to lifecycles.

Technici4n:
If there already exists a general purpose event, I don't see a reason to introduce a more specific one. The documentation issue could be addressed with a wiki page listing all the existing events.

PR #2549, opened 2022-09-25
Build using Ubuntu 22.04

modmuss50:
Update GHA to ubuntu 22.04 in an attempt to fix release issues. PR to let actions run to ensure it still builds.


modmuss50:
This was reverted after merging.

PR #2554, opened 2022-09-27
Create parent directory when writing test report

SquidDev:
This is useful when writing test reports to a child directory, such as `$buildDir/reports/tests/gametest.xml` instead.

The alternative option here would be to put this call inside a mixin to `XmlReportingTestCompletionListener`. Creating a subclass felt more elegant, but happy to use a mixin if preferred!

### DIFF
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/SavingXmlReportingTestCompletionListener.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/SavingXmlReportingTestCompletionListener.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.gametest;
+
+import java.io.File;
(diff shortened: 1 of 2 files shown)


fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/SavingXmlReportingTestCompletionListener.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
+import net.minecraft.test.XmlReportingTestCompletionListener;
+
+public class SavingXmlReportingTestCompletionListener extends XmlReportingTestCompletionListener {
```
modmuss50: Make this final. Can also be package private.

SquidDev: Ahh good catch, thank you!

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/SavingXmlReportingTestCompletionListener.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
+import net.minecraft.test.XmlReportingTestCompletionListener;
+
+final class SavingXmlReportingTestCompletionListener extends XmlReportingTestCompletionListener {
```
Technici4n, line 32: Looks good. This should have a comment explaining why it was added.

Technici4n (approved):

modmuss50 (approved):

apple502j (approved):

Juuxel (approved):

modmuss50:
I will merge this in alongside the next bunch of changes 👍

SquidDev:
❤️ Thanks all for reviewing :)

SquidDev:
Wonderful, thank you so much!

Issue #2503, opened 2022-09-04
sounds.json and lang.json support in fabric-data-generation-api-v1


moltenwolfcub:
The title is quite self-explanatory but as far as I have looked I have been unable to find a way to use this API([fabric-data-generation-api-v1](https://github.com/FabricMC/fabric/tree/1.19.2/fabric-data-generation-api-v1)) to generate the `sounds.json` file with custom SoundEvents or any language file(`lang/en_us.json`, `lang/es_es.json`...). Being a mod developer in both fabric and forge, I've use the `SoundDefinitionsProvider` and `LanguageProvider` data providers in forge - [example of sound usage](https://github.com/moltenwolfcub/crafted_cuisine/blob/master/FORGE/src/main/java/com/moltenwolfcub/crafted_cuisine/datagen/ModSoundsProvider.java),  [Example of Lang usage](https://github.com/moltenwolfcub/crafted_cuisine/blob/master/FORGE/src/main/java/com/moltenwolfcub/crafted_cuisine/datagen/ModLanguageProvider.java#L18) - these have worked well for me but I can't find anything in this API or anywhere else in Fabric that does the same without having to copy-paste many files from forge and adapt them to the Fabric environment. If I have just missed something that already exists, Sorry for wasting your time and I'd appreciate letting me know where it is and how to use it. Thanks.

IMB11:
Maybe I'll take a look at this

IMB11:
Lang is already a PR

moltenwolfcub:
Oh thanks. I shall await this Pull Request to be Merged.

Pschsch:
Any updates on sounds data generator?

apple502j:
Resolved by #4560

PR #2558, opened 2022-09-29
Allow sound instances to play custom audio streams

SquidDev:
There's a couple of mods out there which play dynamically generated sounds, rather than reading from an ogg file. For instance, [Etched](https://github.com/MoonflowerTeam/etched) allows streaming audio from an arbitrary URL and [CC: Restitched](https://github.com/cc-tweaked/cc-restitched) allows players to generate music data and stream it to the client.

Minecraft's sound engine supports custom audio streams. However, as there is no API exposed for this, mods must use mixins. Unfortunately the sound code is fairly complex and, due to its multi-threaded and callback-based nature, awkward to mixin to. For instance (note, most of these examples are using MojMap):

 - [Etched uses `@Redirect`](https://github.com/MoonflowerTeam/etched/blob/8d459ef623a869f6a87cd57f4d634127dd3db7b2/fabric/src/main/java/gg/moonflower/etched/core/mixin/fabric/client/SoundEngineMixin.java#L33-L38). This is the most "correct" method (and is what this PR also does), but is incompatible with any other mod.
 - [CC: Restitched `@Inject`s into the `play` method and a callback lambda](https://github.com/cc-tweaked/cc-restitched/blob/55625c72516fda8290820ec57fdbc994bf06dc1c/src/main/java/dan200/computercraft/fabric/mixin/MixinSoundEngine.java#L29-L50). This code is incorrect, due to the multi-threaded nature of the sound code (though works well enough in practice).
 - Some mods reimplement some (or all) of the `play` method (i.e. [Infinidisc](https://github.com/DenimsMods/Infinidisc/blob/045dd7190b16c51795ee1ee6abaf162d3d6ca10c/src/main/java/denimred/infinidisc/mixin/SoundEngineMixin.java#L53), [music-box](https://github.com/CmdrJane/music-box/blob/c008602ae615af66882c61a4e81e42561c16cbdc/src/main/java/ru/aiefu/musicbox/mixin/SoundEngineMixin.java#L60)).

Given several mods want to provide custom `AudioStream`s, but doing so in a compatible way is tricky to get right, it makes sense for Fabric to provide an API everyone can hook into.

This PR adds a interface `FabricSoundInstance`, injected into `SoundInstance`. This provides a new method `getAudioStream`, which can be overridden to return a custom audio stream. The default implementation implements the current behaviour of just reading an `.ogg` file from resource packs.

## Further notes
 - The obvious alternative here would be to add an event, instead of providing a new interface. The current approach felt more natural, as whenever you want this functionality, you're going to be sub-classing `SoundInstance` anyway. Happy to go for either.

 - This does still require `SoundInstance.getSound` to resolve to an actual file, even if that file isn't read by the audio stream. This is definitely a little clunky; I'm just not sure if "fixing" this is worth the additional implementation cost. Thoughts welcome!

 - I've added a new module `fabric-sound-api-v1` as nowhere else felt natural. Obviously let me know if I've got anything wrong with adding the new module :).

 - It appears that injected interfaces only work when registered in the main source set. I'm not sure if this is intentional or not?


modmuss50:
> It appears that injected interfaces only work when registered in the main source set.

I dont think thats intentional, most likely something I can fix in loom. Ideally client only modules should have everything in the client sourceset as this will prevent the need to split the jar in a mod dev env. I will take a look into this.

Had a quick look at the PR, it looks great at first glance. I will take a deeper look soon. 👍

modmuss50:
This should be done by default, something I can fix in a future loom version, for now you should be able to add the following to the sub projects build.gradle:

```gradle
loom {
	interfaceInjection {
		interfaceInjectionSourceSets.add sourceSets.client
	}
}
```

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.util.Identifier;
+
+/**
+ * General purpose Fabric-provided extensions to {@link SoundInstance}.
+ *
+ * <p>This interface is implicitly implemented on all {@link SoundInstance}s via a mixin and interface injection.
+ */
+public interface FabricSoundInstance {
+	/**
+	 * Load the audio stream for this sound.
```
apple502j:
```suggestion
	 * Loads the audio stream for this sound.
```

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
```diff
@@ -0,0 +1,49 @@
...
+	 *
+	 * <p>By default this will load {@literal .ogg} files from active resource packs. It may be overridden to provide a
```
apple502j:
```suggestion
	 * <p>By default, this will load {@code .ogg} files from active resource packs. It may be overridden to provide a
```

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
```diff
@@ -0,0 +1,49 @@
...
+	 * custom {@link AudioStream} implementation which provides audio from another source, such as over the network or
+	 * driven from user input.
+	 *
+	 * @param loader          The default sound loader, capable of loading {@literal .ogg} files.
+	 * @param id              The resolved sound ID, equal to {@link SoundInstance#getSound()}'s location.
```
apple502j, line 83:
```suggestion
	 * @param id              The resolved sound ID, equal to {@link SoundInstance#getSound()}'s.
```

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
```diff
@@ -0,0 +1,49 @@
...
+	 * @param repeatInstantly Whether this sound should loop. This is true when the sound
+	 *                        {@linkplain SoundInstance#isRepeatable() is repeatable} and has
+	 *                        {@linkplain SoundInstance#getRepeatDelay() no delay}.
+	 * @return The loaded audio stream.
```
apple502j:
```suggestion
	 * @return the loaded audio stream
```

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/sound/SoundSystemMixin.java
```diff
@@ -0,0 +1,42 @@
...
+public class SoundSystemMixin {
+	@Redirect(
+			method = "play(Lnet/minecraft/client/sound/SoundInstance;)V",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/client/sound/SoundLoader;loadStreamed(Lnet/minecraft/util/Identifier;Z)Ljava/util/concurrent/CompletableFuture;"
+			)
+	)
+	private CompletableFuture<?> getStream(SoundLoader loader, Identifier location, boolean looping, SoundInstance sound) {
+		return sound.getAudioStream(loader, location, looping);
```
apple502j:
```suggestion
	private CompletableFuture<?> getStream(SoundLoader loader, Identifier id, boolean looping, SoundInstance sound) {
		return sound.getAudioStream(loader, id, looping);
```

fabric-sound-api-v1/src/client/resources/fabric-sound-api-v1.mixins.json
```diff
@@ -0,0 +1,12 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.client.sound",
+  "compatibilityLevel": "JAVA_16",
```
apple502j:
```suggestion
  "compatibilityLevel": "JAVA_17",
```

fabric-sound-api-v1/src/testmod/java/net/fabricmc/fabric/test/sound/client/SineSound.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraft.sound.SoundCategory;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.math.Vec3d;
+
+class SineSound extends AbstractSoundInstance {
+	protected SineSound(Vec3d position) {
+		super(new Identifier(ClientSoundTest.MOD_ID, "sine_wave"), SoundCategory.BLOCKS, SoundInstance.createRandom());
+		x = position.x;
+		y = position.y;
+		z = position.z;
```
apple502j:
```suggestion
	protected SineSound(Vec3d pos) {
		super(new Identifier(ClientSoundTest.MOD_ID, "sine_wave"), SoundCategory.BLOCKS, SoundInstance.createRandom());
		this.x = pos.x;
		this.y = pos.y;
		this.z = pos.z;
```

fabric-sound-api-v1/src/testmod/java/net/fabricmc/fabric/test/sound/client/SineSound.java
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	@Override
+	public CompletableFuture<AudioStream> getAudioStream(SoundLoader loader, Identifier location, boolean repeatInstantly) {
```
apple502j:
```suggestion
	public CompletableFuture<AudioStream> getAudioStream(SoundLoader loader, Identifier id, boolean repeatInstantly) {
```

fabric-sound-api-v1/src/testmod/java/net/fabricmc/fabric/test/sound/client/SineStream.java
```diff
@@ -0,0 +1,56 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.test.sound.client;
+
+import java.nio.ByteBuffer;
+
+import javax.sound.sampled.AudioFormat;
```
apple502j, line 21: Just making sure, this won't cause a mess in macOS right?

apple502j (changes requested):
Also why are you shipping the empty ogg farm as part of the API (and not testmod)?

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
SquidDev: `SoundInstance.getSound()` returns a `Sound`, not an `Identifier`, so I don't think this makes much sense.

SquidDev:
> Also why are you shipping the empty ogg farm as part of the API (and not testmod)?

As every mod that provides a custom audio stream needs to reference a dummy sound it made sense to provide it in the API, rather than every mod having to bundle their own - much like fabric-gametest-api-v1's empty structure.

fabric-sound-api-v1/src/testmod/java/net/fabricmc/fabric/test/sound/client/SineStream.java
SquidDev: I don't have a Mac to test, but I'd be surprised if it doesn't work. It's just part of the standard JRE (`java.desktop` module, same as Swing) and given that vanilla MC uses it, I feel we'd have seen issues much earlier.

modmuss50: Just tested it on my ARM mac and this works as expected. I dont think there is any concern here as its not an AWT/swing class. The vanilla usage proves this is a none issue.

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
apple502j: Well, maybe `'s ID`.

apple502j:
@SquidDev Then the existence should be documented somewhere, along with `Identifier` of the empty sound exposed as a field.

fabric-sound-api-v1/src/client/java/net/fabricmc/fabric/api/client/sound/v1/FabricSoundInstance.java
SquidDev: I went for `location` because that's the name of the field & getter `Sound`. ID is _fine_, but I don't think it indicates exactly what is being referred to here as well.

SquidDev: I think part of the problem here is that it's important to distinguish between the id of the `SoundInstance` (which corresponds to the id in `sounds.json` and `SoundEvent`), and the id of the actual `Sound` being played (which corresponds to a path in a resource pack). Separating the two by having different nomenclature for both feels helpful.

SquidDev:
> Then the existence should be documented somewhere, along with Identifier of the empty sound exposed as a field.

:+1:  Is it acceptable just to add it as a constant to `FabricSoundInstance`? Not sure there's a more sensible location until #2503 is resolved.

Technici4n (approved):
Looks good to me!

modmuss50:
> Is it acceptable just to add it as a constant to FabricSoundInstance?

Yes, I think that would be a good idea 👍

modmuss50 (approved):
Looks good to me, just need to document the existence of the empty sound, and why is required.

SquidDev:
I felt it was probably worth while adding an example, as the need for `sounds.json` is a little awkward and thus worth being explicit in. Happy to revert if it's Too Much :).

SquidDev:
Hah, I should know better than to commit without running checkstyle!

I don't know if there's any interest in adding a pre-commit hook which runs it automatically? Or is that just going to get too irritatingly slow for people?

Technici4n:
More documentation is always welcome. :)

Not sure about the pre-commit hook as I've never used them myself, but it looks like it might end up being rather annoying ? We always wait for CI to pass before merging anything.

Juuxel:
IMO it's too slow since running `gradlew checkstyleMain checkstyleTestmod` (or even just the main one) builds all modules and that takes a while even with `--parallel`.

modmuss50 (approved):

modmuss50:
I agree, spotless would also need to be ran making commits very slow. I would reccomend using the checkstyle plugin in your IDE to make it clear when you have an issue.

Added to last call, thanks 👍

SquidDev:
Thank you everyone :).

PR #2564, opened 2022-10-01
Resource conditions: support all tags

apple502j:
Should we deprecate registry-specific datagen methods?

### DIFF
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
@@ -16,6 +16,8 @@
 
 package net.fabricmc.fabric.api.resource.conditions.v1;
 
+import java.util.Objects;
+
 import com.google.common.base.Preconditions;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
@@ -43,7 +45,8 @@ public interface ConditionJsonProvider {
 		JsonArray array = new JsonArray();
 
 		for (ConditionJsonProvider condition : conditions) {
-			if (condition != null) array.add(condition.toJson());
+			Objects.requireNonNull(condition, "condition cannot be null");
+			array.add(condition.toJson());
 		}
 
 		conditionalObject.add(ResourceConditions.CONDITIONS_KEY, array);
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
@@ -43,7 +43,7 @@ public interface ConditionJsonProvider {
 		JsonArray array = new JsonArray();
 
 		for (ConditionJsonProvider condition : conditions) {
-			array.add(condition.toJson());
+			if (condition != null) array.add(condition.toJson());
 		}
 
 		conditionalObject.add(ResourceConditions.CONDITIONS_KEY, array);
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
@@ -90,24 +90,30 @@ public final class DefaultResourceConditions {
 
 	/**
 	 * Create a condition that returns true if each of the passed block tags exists and has at least one element.
+	 * @deprecated Use {@link #tagsPopulated} instead.
 	 */
 	@SafeVarargs
+	@Deprecated
 	public static ConditionJsonProvider blockTagsPopulated(TagKey<Block>... tags) {
 		return ResourceConditionsImpl.tagsPopulated(BLOCK_TAGS_POPULATED, false, tags);
 	}
 
 	/**
 	 * Create a condition that returns true if each of the passed fluid tags exists and has at least one element.
+	 * @deprecated Use {@link #tagsPopulated} instead.
 	 */
 	@SafeVarargs
+	@Deprecated
 	public static ConditionJsonProvider fluidTagsPopulated(TagKey<Fluid>... tags) {
 		return ResourceConditionsImpl.tagsPopulated(FLUID_TAGS_POPULATED, false, tags);
 	}
 
 	/**
 	 * Create a condition that returns true if each of the passed item tags exists and has at least one element.
+	 * @deprecated Use {@link #tagsPopulated} instead.
 	 */
 	@SafeVarargs
+	@Deprecated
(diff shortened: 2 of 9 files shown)


fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
```diff
@@ -93,7 +93,7 @@ public void writeParameters(JsonObject object) {
 	}
 
 	@SafeVarargs
-	public static <T> ConditionJsonProvider tagsPopulated(Identifier id, TagKey<T>... tags) {
+	public static <T> ConditionJsonProvider tagsPopulated(Identifier id, boolean includeRegistry, TagKey<T>... tags) {
```
Technici4n, line 96: You should maybe check that all the tag keys have the same registry, just in case.

Technici4n (commented):
I was thinking of allowing keys from different registries but it doesn't really make sense, it's better to combine them with the and condition.

Technici4n:
> Should we deprecate registry-specific datagen methods?

I would actually suggest defaulting to the item registry for the new condition, and deprecating the other 3. The item registry is the primary target, so it makes sense for it to be special-cased.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
apple502j: I don't think that compiles anyway.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
```diff
@@ -43,7 +43,7 @@ static void write(JsonObject conditionalObject, ConditionJsonProvider @Nullable.
 		JsonArray array = new JsonArray();
 
 		for (ConditionJsonProvider condition : conditions) {
-			array.add(condition.toJson());
+			if (condition != null) array.add(condition.toJson());
```
Technici4n: Hmmm? No condition is supposed to be null, it's the vararg that is nullable (otherwise it would have been `@Nullable ConditionJsonProvider...`)

fabric-resource-conditions-api-v1/src/testmod/resources/data/fabric-resource-conditions-api-v1-testmod/recipes/tags_populated.json
```diff
@@ -0,0 +1,20 @@
...
+  "result": {
+    "item": "minecraft:diamond"
+  },
+  "fabric:load_conditions": [
+    {
+      "condition": "fabric:tags_populated",
+      "values": [
+        "fabric-resource-conditions-api-v1-testmod:test_condition"
+      ],
+      "registry": "minecraft:item"
```
Technici4n: Can you try without this line to ensure it works fine? :)

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ConditionJsonProvider.java
apple502j: Ah, makes sense. Changing this to requireNonNull.

Technici4n: If you want - otherwise the crash is pretty straightforward imo. :smile:

Technici4n (approved):

Juuxel:
Is the fact that the new condition defaults to items actually documented somewhere? I'd say that's important info for users, as well as the format of the condition (which I think isn't documented currently *at all*).

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/package-info.java
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Provides a way of conditionally loading JSON-based resources. This can be used with
+ * recipes, loot tables, advancements, predicates, and item modifiers. Conditions are
```
Technici4n: This is only by default for fabric, but mods are free to use conditions too for whatever they want. I use them in Modern Dynamics for the datapack upgrade specifications. :wink:

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -44,7 +44,10 @@ public final class DefaultResourceConditions {
 	private static final Identifier TAGS_POPULATED = new Identifier("fabric:tags_populated");
 
 	/**
-	 * Create a NOT condition: returns true if its child condition is false, and false if its child is true.
+	 * Creates a NOT condition that returns true if its child condition is false, and false if its child is true.
+	 *
+	 * @apiNote This condition's ID is {@code fabric:not}, and takes one property, {@code value},
+	 * which is a condition provider.
```
Technici4n: I don't think `condition provider` is correct, but rather `condition`: a condition is the result of serialization a provider to JSON. Same for the other functions below.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -44,7 +44,10 @@ public final class DefaultResourceConditions {
 	private static final Identifier TAGS_POPULATED = new Identifier("fabric:tags_populated");
 
 	/**
-	 * Create a NOT condition: returns true if its child condition is false, and false if its child is true.
+	 * Creates a NOT condition that returns true if its child condition is false, and false if its child is true.
+	 *
+	 * @apiNote This condition's ID is {@code fabric:not}, and takes one property, {@code value},
```
Technici4n, line 49: Would be nice to be able to use `@value` but it probably doesn't work for IDs.

Technici4n (changes requested):

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/package-info.java
Technici4n: (Note: I am suggesting a clarification to the javadoc)

Technici4n (approved):

modmuss50 (approved):

PR #2573, opened 2022-10-10
Add `AFTER_DEATH` and `ALLOW_DAMAGE` events; generalise `ALLOW_DEATH` to living entities

### LINKED ISSUE #2551: Add Entity Death Event ideally with source of the death
Title kinda says it all, looks like Fabric doesn't have an event for when any kind of entity dies.

Technici4n:
Add a new `ServerLivingEntityEvents` class, with the following events:
- `ALLOW_DAMAGE` to block damage entirely, for example because the player is wearing a "creative mode" armor (my use case).
- `ALLOW_DEATH` to prevent entity death. It is the exact same event as `ServerPlayerEvents.ALLOW_DEATH`, but generalised to living entities. The old event in `ServerPlayerEvents.ALLOW_DEATH` is now deprecated.
- `AFTER_DEATH` to react on entity death. Closes #2551.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
@@ -25,7 +25,7 @@ import net.fabricmc.fabric.api.event.EventFactory;
 /**
  * Various server-side only events related to living entities.
  */
-public class ServerLivingEntityEvents {
+public final class ServerLivingEntityEvents {
 	/**
 	 * An event that is called when a living entity is going to take damage.
 	 *
@@ -110,5 +110,8 @@ public class ServerLivingEntityEvents {
 		 * @param damageSource the source of the fatal damage
 		 */
 		void afterDeath(LivingEntity entity, DamageSource damageSource);
+	}
+
+	private ServerLivingEntityEvents() {
 	}
 }
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
@@ -28,8 +28,8 @@ import net.fabricmc.fabric.api.event.EventFactory;
 public final class ServerLivingEntityEvents {
 	/**
 	 * An event that is called when a living entity is going to take damage.
-	 *
-	 * <p>Mods can cancel this to prevent the damage entirely.
+	 * This is fired from {@link LivingEntity#damage}, before armor or any other mitigation are applied.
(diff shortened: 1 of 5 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.damage.DamageSource;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Various server-side only events related to living entities.
+ */
+public class ServerLivingEntityEvents {
```
Juuxel: Should be final with a private constructor like the other event holder classes

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
```diff
@@ -0,0 +1,117 @@
...
+	 *
+	 * <p>Vanilla checks for entity health {@code <= 0} each tick (with {@link LivingEntity#isDead()}), and kills if true -
+	 * so the entity will still die next tick if this event is cancelled.
+	 * It's assumed that the listener will do something to prevent this, for example, if the entity is a player:
+	 * <ul>
+	 *     <li>a minigame mod teleporting the player into a 'respawn room' and setting their health to 20.0</li>
+	 *     <li>a mod that changes death mechanics switching the player over to the mod's play-mode, where death doesn't apply</li>
+	 * </ul>
+	 */
+	public static final Event<AllowDeath> ALLOW_DEATH = EventFactory.createArrayBacked(AllowDeath.class, callbacks -> (player, damageSource, damageAmount) -> {
```
modmuss50: Should be `entity` not `player`

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+
+		return true;
+	});
+
+	/**
+	 * An event that is called when a living entity dies.
+	 */
+	public static final Event<AfterDeath> AFTER_DEATH = EventFactory.createArrayBacked(AfterDeath.class, callbacks -> (player, damageSource) -> {
```
modmuss50: Should be `entity` not `player`

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
Technici4n: I blame copilot!

modmuss50 (approved):

PR #2576, opened 2022-10-13
Resource loader: mark impl as internal, fix deprecated loader API use

apple502j:
This marks the rest of impl classes as internal, marks `ResourceManagerHelper` as non-extendable, fixes use of deprecated Loader APIs, and fixes javadocs.

Originally included changes on keyed reloaders (support for predicate/modifier) but excluded in this PR. In v1 we should make this use event phases.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
@@ -18,7 +18,7 @@ package net.fabricmc.fabric.impl.resource.loader;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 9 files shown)


Juuxel:
Should the internal annotation be on the entire package instead of the individual classes?

> If a package is marked with this annotation, all its containing classes are considered internal.

<https://javadoc.io/doc/org.jetbrains/annotations/latest/org/jetbrains/annotations/ApiStatus.Internal.html>

apple502j:
Ah, that could be better. Maybe I should use that instead.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
```diff
@@ -65,10 +65,9 @@ public static ResourceManagerHelperImpl get(ResourceType type) {
 	 * @return {@code true} if successfully registered the resource pack, else {@code false}
 	 * @see ResourceManagerHelper#registerBuiltinResourcePack(Identifier, ModContainer, String, ResourcePackActivationType)
 	 * @see ResourceManagerHelper#registerBuiltinResourcePack(Identifier, ModContainer, ResourcePackActivationType)
-	 * @see ResourceManagerHelper#registerBuiltinResourcePack(Identifier, String, ModContainer, boolean)
 	 */
 	public static boolean registerBuiltinResourcePack(Identifier id, String subPath, ModContainer container, String displayName, ResourcePackActivationType activationType) {
-		String separator = container.getRootPath().getFileSystem().getSeparator();
+		String separator = container.getRootPaths().get(0).getFileSystem().getSeparator();
```
liach, line 70: Is this algorithm right? i.e. is it possible for builtin resource packs to be stored in non-first root path?

Technici4n: It is, but this is only for the FS separator. Packs split across multiple root paths should be handled correctly (in the resource pack impl).

apple502j: Yeah, all filesystems should have the same file separator, so doesn't matter.

Technici4n:
@apple502j Mind fixing the build? :)

apple502j:
This can be merged to 1.19.2, seems like the 22w42a port refactors this completely making this unnecessary.

PR #2583, opened 2022-10-15
Remove any infinite recursion from SpriteFinderImpl to prevent StackOverflowExceptions

mo0dss:


Technici4n:
Hi, would you mind explaining in detail what this PR does, which issue it solves, etc?

mo0dss:
https://github.com/PepperCode1/Continuity/issues/120#issuecomment-1242888566

Technici4n:
It is not immediately obvious how your change addresses the root issue, can you please elaborate on that?

PS: I suspect the actual issue is caused by machine precision issues.

mo0dss:
Without this PR all resourcepacks get unloaded due the exception that gets thrown

[without.log](https://github.com/FabricMC/fabric/files/9791934/without.log)

[with.log](https://github.com/FabricMC/fabric/files/9791940/with.log)

apple502j:
We acknowledge the reason behind making such change, but I'd like to see what the change actually does. Rendering code is complicated and needs extensive comments/documentation.

Could you tell us how the old and new code works, perhaps? Also do you know the real reason behind stack overflow? (The real problem, I believe, stems from other bugs.)

Technici4n:
Could you also provide the zip of the mods you used for testing? This would help us reproduce the issue and validate the fix.

mo0dss:
Ah okay thats understandable haha, in this pr the node and quadrants are specified in a more specific way, the finder class contains an array that is expected to have 4 quadrants (lh, hl, ll, hh) of nodes that will again specify another pair of quadrants for sprites instead, resulting in an exact 4 * 4, while in the previous code contains 4 objects that dont specify clear enough if its a sprite or a node, 

for instance lets say were trying to find a sprite at 0.5, 0.5
in the old code itll check for the quadrant in that area, and return that quadrant if its a sprite but if it isnt a sprite itll search that node to find a sprite in that area, and thats where the issue occurs, because if that node contains another node instead of a sprite, itll once more check for a quadrant in that area returning another sprite, or another node while the recursion size should be at max 16, the cause of that is mostly because of `addInner` creating only a node after a sprite has been added twice, the moment that sprite gets called again it will check if the existing quadrant is a node, if not itll create a new one and add the existing quadrant as a sprite, if the sprite gets used again for that quadrant while its a node, it will add more nodes, resulting in the stackoverflow, hopefully i explained that well lol

mo0dss:
> Could you also provide the zip of the mods you used for testing? This would help us reproduce the issue and validate the fix.

https://drive.google.com/file/d/1gsTDpcH_4NqT2c7jYricDhViZiKiN3L1/view?usp=sharing

apple502j:
We've identified the problematic mod - DashLoader - and is working on fixes. Closing, as we'll implement fail-safe measures instead. (The code here is fine.)

Technici4n:
The recursion is a feature. Removing it entirely prevents the stack overflow, but it also breaks functionality of the SpriteFinder. However, thanks to your .zip, I was able to reproduce and fix the issue. You can find my fix at #2584.

apple502j:
(note, any recursion code can be rewritten to not use recursion, but yes.)

PR #2586, opened 2022-10-15
Fabric custom ingredient API

### LINKED ISSUE #116: Custom Ingredient API.
Ingredient is final, and thus extending it to cover NBTs is impossible without lots of hackery.

Proposal: As it is non-trivial to get right, Fabric should implement it by providing a public abstract class CustomIngredient that can be used to create an Ingredient, + serialization logic with types.

Questions: 

* Should it be extended as such, or should mods simply lie/simplify for recipe previews and provide custom matching/crafting logic?
* Is there anything else to keep in mind?

Technici4n:
This API allows modders to define custom `Ingredient`s, with more advanced matching logic than what vanilla allows.

Custom ingredients need to implement the `CustomIngredient` interface, and also provide and register a `CustomIngredientDeserializer`. No overriding of `Ingredient` is needed.

Have a look at the API here: https://github.com/Technici4n/fabric/tree/custom-ingredient-api/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient.

The default conditions can be found here: https://github.com/Technici4n/fabric/blob/custom-ingredient-api/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java.

Here is how the default conditions are implemented: https://github.com/Technici4n/fabric/tree/custom-ingredient-api/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/builtin.

Closes #116.

TODO:
- [x] Decide on exact API and impl structure.
- [x] Add partial/strict NBT, and and difference ingredients.
- [x] JSON and network serialization and deserialization tests.
- [x] Shapeless recipe matching tests.
- [x] `isEmpty` checks tests.
- [x] Write javadoc.
- [x] Test what happens with vanilla clients. See https://github.com/FabricMC/fabric/pull/2586#issuecomment-1355037420
- [x] Rename to Recipe API.


AlphaMode:
A possible NBT ingredient would also be useful

Technici4n:
> A possible NBT ingredient would also be useful

I have the following ingredients in mind, just haven't written them yet: (matching Forge's)
- strict and partial NBT ingredient
- and ingredient
- difference ingredient

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,20 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+
+import net.fabricmc.fabric.impl.ingredient.CustomIngredientImpl;
+
+// TODO: what about getMatchingItemIds ?
+public interface CustomIngredient {
+	boolean matchesStack(ItemStack stack);
+	// TODO: match yarn name? this name derived from old 1.16 yarn Recipe#getPreviewInputs() is more accurate
+	ItemStack[] getPreviewStacks();
```
jaredlll08: I feel the name `getPreviewStacks` gives the wrong impression, the method is used in `Recipe#isEmpty` to determine if a recipe is is incomplete.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/FabricIngredient.java
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	@Nullable
+	default CustomIngredient getCustomIngredient() {
+		return null;
+	}
+
+	default boolean ignoresNbt() {
+		return true;
+	}
```
jaredlll08: I'm not too sure how fabric api deals with this, but should these not have the `fabric_` prefix?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,20 @@
+package net.fabricmc.fabric.api.ingredient.v1;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+
+import net.fabricmc.fabric.impl.ingredient.CustomIngredientImpl;
+
+// TODO: what about getMatchingItemIds ?
+public interface CustomIngredient {
+	boolean matchesStack(ItemStack stack);
```
jaredlll08: An Ingredient is basically a `Predicate<ItemStack>` so why not have this as
```suggestion
	boolean test(ItemStack stack);
```

jaredlll08 (commented):
Looking good so far, however there is a pretty big issue that isn't being dealt with by this PR, ShapelessRecipe.

ShapelessRecipe uses `Ingredient#getMatchingItemIds` to determine if a recipe matches or not, and it is pretty fast doing that, however it gains it's speed by not calling `Ingredient#test` and instead uses the integer id of the Item (see `RecipeMatcher#getItemId`), so to make this PR work with ShapelessRecipes, it needs to mixin ShapelessRecipe and change it to call `Ingredient#test`, which is not as simple as that may sound.

A ShapelessRecipe is basically a [Bipartite Graph](https://en.wikipedia.org/wiki/Bipartite_graph) where each input item needs to match an Ingredient, however multiple items may match the same ingredient.

So a recipe like this (In CraftTweaker format since I don't know the json format for nbt):

```zenscript
craftingTable.addShapeless("name", <item:minecraft:dirt>, 
[
<item:minecraft:diamond_sword>.withEnchantment(<enchantment:minecraft:unbreaking>),
<item:minecraft:diamond_sword>.withEnchantment(<enchantment:minecraft:sharpness>)
]);
```

Which requires 2 Diamond Swords, one with Unbreaking and one with Sharpness, needs to make sure that if a player puts:

```zenscript
[
<item:minecraft:diamond_sword>.withEnchantment(<enchantment:minecraft:unbreaking>).withEnchantment(<enchantment:minecraft:sharpness>),
<item:minecraft:diamond_sword>.withEnchantment(<enchantment:minecraft:sharpness>)
]
```

Into a crafting grid, it doesn't try and use the Sword with Unbreaking to fill the Ingredient that asks for Sharpness (assuming partial nbt matching).

To solve this you need to implement something like the [Hopcroft–Karp algorithm](https://en.wikipedia.org/wiki/Hopcroft%E2%80%93Karp_algorithm), which produces how many ItemStacks match the Ingredients.

There may be other solutions to the problem, but I know that Hopcroft-Karp solves it.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/FabricIngredient.java
Technici4n: For injected interfaces, the status quo is to not add the `fabric_` prefix. Personally. I think that it looks quite ugly, and given fabric API's privileged status we can request other mods to adjust their methods in case of a name conflict.

Technici4n:
Thanks a lot for your detailed explanation.

I decided to go for the simpler augmenting paths algorithm for the maximum cardinality bipartite matching problem, which should look like this : https://github.com/Technici4n/fabric/blob/custom-ingredient-api/fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/ShapelessMatch.java. I haven't tested it yet though, but I will definitely add tests similar to what you suggested.

I also renamed `matchesStack` and `getPreviewStacks` to `test` and `getMatchingStacks` respectively, and provided some javadoc for `CustomIngredient`. I hope the javadoc is clear enough - there's some tricky details that might not always be obvious. I also borrowed your name for `requiresTesting` which I like better than `ignoresNbt` - hopefully you don't mind. :smile:

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
Technici4n: Renamed to `getMatchingStacks`.

Technici4n:
I just tested the following cases:
- Server with Fabric ingredient API, vanilla client, no recipe using custom ingredients: can connect fine.
- Server with Fabric ingredient API, vanilla client, some recipes using custom ingredients: can't connect, with the following error:
image
**Is this fine?**
- Client/server with both Fabric ingredient API and [Faux ingredient extension API](https://github.com/FauxMC/IngredientExtensionAPI): packet writing sending works, but I used the 1.19.2 version of the ingredient extension API, so I get a crash due to other reasons. I think the two APIs are compatible. (Though obviously the ingredient extension API won't be needed anymore once this PR is merged, but I want a smooth transition).

This PR is now ready for reviews!

jaredlll08:
> Client/server with both Fabric ingredient API and [Faux ingredient extension API](https://github.com/FauxMC/IngredientExtensionAPI): packet writing sending works, but I used the 1.19.2 version of the ingredient extension API, so I get a crash due to other reasons. I think the two APIs are compatible. (Though obviously the ingredient extension API won't be needed anymore once this PR is merged, but I want a smooth transition

We have no intention of continuing development on the Faux Ingredient Extension API once this PR is merged, so even if it wasn't compatible, it shouldn't be an issue since neither APIs should share a version.

Technici4n:
My thinking is that you might release a 1.19.3 version of your Extension API before this PR gets merged in 1.19.3. (I have no idea when it will be merged, it's going to require quite a bit of review).

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			ci.cancel();
+		}
+	}
+
+	/**
+	 * @author FabricMC
+	 * @reason Support custom ingredient network deserialization.
+	 */
+	@Overwrite
+	public static Ingredient fromPacket(PacketByteBuf buf) {
```
Patbox: This could be replaced with an injection

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			),
+			method = "write",
+			cancellable = true
+	)
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	private void injectWrite(PacketByteBuf buf, CallbackInfo ci) {
+		// Done here instead of writing the ingredient in the subclass for compat with faux ingredient extension API
+		CustomIngredient customIngredient = getCustomIngredient();
+
+		if (customIngredient != null) {
```
Patbox: You could check with networking api if client has required fabric version for better compatibility. Also it could be used to skip ingredients unsupported by client

Shnupbups:
Would it be out of scope to add a way to give these custom ingredients a recipe remainder that overrides the one in their item's settings?
So one could, for example, make a "chocolate milk bucket" recipe that takes a milk bucket and cocoa beans, without the milk bucket leaving behind an empty bucket by overriding the milk bucket's remainder in just that recipe.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Technici4n: Could be but I don't see the point since I'm basically overwriting the entire method.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Technici4n: I can't skip ingredients that are not supported by the client due to how reloading works, but I can kick clients that don't have the ingredient API if the recipe packet contains custom ingredients. I'll try that.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Patbox: You could allow vanilla code to run, for better compat for things that inject there. And for vanilla code reading that, you can set packbufers index to one before reading

Technici4n:
> Would it be out of scope to add a way to give these custom ingredients a recipe remainder that overrides the one in their item's settings? So one could, for example, make a "chocolate milk bucket" recipe that takes a milk bucket and cocoa beans, without the milk bucket leaving behind an empty bucket by overriding the milk bucket's remainder in just that recipe.

I think it would be out of scope. Some of the problems:
- Ingredients are used in many places, so properly supporting this would require a lot of patches.
- The patches are very invasive as they require all remainder logic to go through the ingredient.
- For recipes, we don't even know which ingredients were matched where in the getRecipeRemainder method, so we'd need to recompte the match before we apply custom remainders.
I think it's best left to other mods (for crafting recipes, it's "easy" to implement on top of this API with a custom recipe type).

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Technici4n: So I can "unread" the int? That would work better indeed.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Patbox: Because that method is written on networking thread, you can get the player, see Nucleoid's Packet Tweaker library

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Patbox: Yep, just store old index before reading and restore it if size is different than -1

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Technici4n: Look at the way reload is implemented. It's using a single packet for all connected clients.

Patbox: Recipes are still written in separate bytebufs per player (packet only contains list of recipes), so that would still work (polymer works thanks to that for example)

Technici4n: Ahhh I see, I'll have a look. Thanks!

Should I just skip invalid recipes, perhaps with a console warning, or disconnect the player?

Patbox: Since fabric contains these by default, just skipping recipes would cause least confusion. alternative would be just writing empty ingredient (0 as size and nothing else). As for warning, it could get rather spammy so maybe only in dev/needs to be enabled by hand?

Patbox: To be fair looking at it now, writing result of getMatchingStacks for client not supporting required type would be also a good alternative

Technici4n: It might lead to ugly desyncs though, where the client thinks it can craft something but the server disagrees. Maybe we should write a barrier block with a warning in the name, Forge-style. :joy:

Patbox: Well, adding some random nbt (with hard to get with commands key/value) would resolve auto matching

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+ *     "fabric:type": "<identifier of the serializer>",
+ *     // extra ingredient data, dependent on the serializer
+ * }
+ * }</pre>
+ *
+ * @see CustomIngredientSerializer
+ */
+public interface CustomIngredient {
+	/**
+	 * Check if a stack matches this ingredient.
```
apple502j:
```suggestion
	 * Checks if a stack matches this ingredient.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @param stack the stack to test
+	 * @return true if the stack matches this ingredient, false otherwise
```
apple502j:
```suggestion
	 * @return {@code true} if the stack matches this ingredient, {@code false} otherwise
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 *     <li>The ingredient should try to return at least one stack with each accepted {@link Item}.
+	 *     This allows mods that inspect the ingredient to figure out which stacks it might accept.</li>
+	 * </ul>
+	 *
+	 * <p>Note: no caching needs to be done by the implementation, this is already handled by the ingredient itself.
+	 */
+	ItemStack[] getMatchingStacks();
+
+	/**
+	 * Return whether this ingredient always requires {@linkplain #test direct stack testing}.
```
apple502j: Consistency? (Use "Returns" or inline return)

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * <p>If false, {@linkplain #test testing this ingredient} with an item stack must be equivalent to checking whether
```
apple502j:
```suggestion
	 * <p>If {@code false}, {@linkplain #test testing this ingredient} with an item stack must be equivalent to checking whether
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 * the item stack's item is contained in the ingredient's {@linkplain #getMatchingStacks() list of matching stacks}.
```
apple502j:
```suggestion
	 * the item stack's item is included in the ingredient's {@linkplain #getMatchingStacks() list of matching stacks}.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 * In that case, optimized matching logic can be used, for example using {@link Ingredient#getMatchingItemIds()}.
+	 *
+	 * <p>If true, the ingredient must always be tested using {@link #test(ItemStack)}.
+	 * Note: Fabric patches some vanilla systems such as shapeless recipes to account for this.
```
apple502j:
```suggestion
	 * <p>If {@code true}, the ingredient must always be tested using {@link #test(ItemStack)}.
	 * Note that Fabric patches some vanilla systems such as shapeless recipes to account for this.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @return true if this ingredient ignores NBT data when matching stacks, false otherwise
```
apple502j: 🤦🏼 I think you mean the opposite

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredient.java
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	boolean requiresTesting();
+
+	/**
+	 * {@return the serializer for this ingredient}
+	 *
+	 * <p>The serializer must have be registered using {@link CustomIngredientSerializer#register}.
```
apple502j:
```suggestion
	 * <p>The serializer must have been registered using {@link CustomIngredientSerializer#register}.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Serializer for a {@link CustomIngredient}.
+ *
+ * <p>All instances must be registered using {@link #register} for deserialization to work.
+ *
+ * @param <T> the type of the custom ingredient
+ */
+public interface CustomIngredientSerializer<T extends CustomIngredient> {
+	/**
+	 * Register a custom ingredient serializer, using the {@linkplain CustomIngredientSerializer#getIdentifier() serializer's identifier}.
```
apple502j:
```suggestion
	 * Registers a custom ingredient serializer, using the {@linkplain CustomIngredientSerializer#getIdentifier() serializer's identifier}.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+	 *
+	 * @throws IllegalArgumentException if the serializer is already registered
+	 */
+	static void register(CustomIngredientSerializer<?> serializer) {
+		CustomIngredientImpl.registerSerializer(serializer);
+	}
+
+	/**
+	 * {@return the custom ingredient serializer registered with the given identifier, or {@code null} if none is registered}.
```
apple502j:
```suggestion
	 * {@return the custom ingredient serializer registered with the given identifier, or {@code null} if there is no such serializer}.
```

The current sentence reads like it returns null when there is no serializer at all, which is incorrect.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+		return CustomIngredientImpl.getSerializer(identifier);
+	}
+
+	/**
+	 * {@return the identifier of this serializer}.
+	 */
+	Identifier getIdentifier();
+
+	/**
+	 * Deserialize the custom ingredient from a JSON object.
```
apple502j:
```suggestion
	 * Deserializes the custom ingredient from a JSON object.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	T read(JsonObject json);
```
apple502j: Yarn convention is deserialize/serialize. "Read/write" is used for mutable objects reading from/writing to an object.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Serialize the custom ingredient to a JSON object.
```
apple502j:
```suggestion
	 * Serializes the custom ingredient to a JSON object.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	void write(JsonObject json, T ingredient);
```
apple502j: See above. Also, should this be `JsonObject write(T ingredient)` instead perhaps?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Deserialize the custom ingredient from a packet buffer.
```
apple502j:
```suggestion
	 * Deserializes the custom ingredient from a packet buffer.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	T read(PacketByteBuf buf);
+
+	/**
+	 * Serialize the custom ingredient to a packet buffer.
```
apple502j:
```suggestion
	 * Serializes the custom ingredient to a packet buffer.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+import net.minecraft.nbt.NbtHelper;
+import net.minecraft.recipe.Ingredient;
+
+import net.fabricmc.fabric.impl.ingredient.builtin.AndIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.DifferenceIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.NbtIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.OrIngredient;
+
+/**
+ * Factory methods for the custom ingredients directly provided by fabric.
```
apple502j:
```suggestion
 * Factory methods for the custom ingredients directly provided by Fabric API.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public final class DefaultCustomIngredients {
+	/**
+	 * Create an ingredient that matches what its sub-ingredients all match.
```
apple502j:
```suggestion
	 * Creates an ingredient that matches when its sub-ingredients all match.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 *     "ingredients": [
+	 *         // sub-ingredient 1,
+	 *         // sub-ingredient 2,
+	 *         // etc...
+	 *     ]
+	 * }
+	 * }</pre>
+	 */
+	public static Ingredient and(Ingredient... ingredients) {
+		for (Ingredient ing : ingredients) Objects.requireNonNull(ing);
```
apple502j: Have error message? (Also, should we document that empty is banned?)

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+
+		return new AndIngredient(ingredients).toVanilla();
+	}
+
+	/**
+	 * Create an ingredient that matches what any of its sub-ingredients matches.
```
apple502j:
```suggestion
	 * Creates an ingredient that matches when any of its sub-ingredients matches.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 *     "ingredients": [
+	 *         // sub-ingredient 1,
+	 *         // sub-ingredient 2,
+	 *         // etc...
+	 *     ]
+	 * }
+	 * }</pre>
+	 */
+	public static Ingredient or(Ingredient... ingredients) {
+		for (Ingredient ing : ingredients) Objects.requireNonNull(ing);
```
apple502j: See above

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+
+		return new OrIngredient(ingredients).toVanilla();
+	}
+
+	/**
+	 * Create an ingredient that matches if its base ingredient matches, and its subtracted ingredient does NOT match.
```
apple502j:
```suggestion
	 * Creates an ingredient that matches if its base ingredient matches, and its subtracted ingredient <strong>does not</strong> match.
```

Why CAPS when we can **strong**?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 * {
+	 *     "fabric:type": "fabric:difference",
+	 *     "base": // base ingredient,
+	 *     "subtracted": // subtracted ingredient
+	 * }
+	 * }</pre>
+	 */
+	public static Ingredient difference(Ingredient base, Ingredient subtracted) {
+		Objects.requireNonNull(base);
+		Objects.requireNonNull(subtracted);
```
apple502j: Error message?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+
+		return new DifferenceIngredient(base, subtracted).toVanilla();
+	}
+
+	/**
+	 * Create an ingredient that wraps another ingredient to also check for stack NBT.
```
apple502j:
```suggestion
	 * Creates an ingredient that wraps another ingredient to also check for stack NBT.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 * This check can either be strict (the exact NBT must match) or non-strict (the ingredient NBT must be a subset of the stack NBT).
+	 *
+	 * <p>Note: refer to {@link NbtHelper#matches} for the non-strict matching.
```
apple502j:
```suggestion
	 * <p>See {@link NbtHelper#matches} for the non-strict matching.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 *
+	 * <p>The JSON format is as follows:
+	 * <pre>{@code
+	 * {
+	 *    "fabric:type": "fabric:nbt",
+	 *    "base": // base ingredient,
+	 *    "nbt": // NBT tag to match, either in JSON directly or a string representation (default: null),
```
apple502j: Document that null means the absence of NBT in strict, and prohibited in non-strict?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 *    "strict": // whether to use strict matching (default: false)
+	 * }
+	 * }</pre>
+	 */
+	public static Ingredient nbt(Ingredient base, @Nullable NbtCompound nbt, boolean strict) {
+		Objects.requireNonNull(base);
```
apple502j: Error message?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+
+		return new NbtIngredient(base, nbt, strict).toVanilla();
+	}
+
+	/**
+	 * Create an ingredient that matches the passed template stack, NBT included.
```
apple502j:
```suggestion
	 * Creates an ingredient that matches the passed template stack, including NBT.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 * This check can either be strict (the exact NBT must match) or non-strict (the template NBT must be a subset of the stack NBT).
+	 *
+	 * <p>Note: refer to {@link NbtHelper#matches} for the non-strict matching.
```
apple502j:
```suggestion
	 * <p>See {@link NbtHelper#matches} for the non-strict matching.
```

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	public static Ingredient nbt(ItemStack stack, boolean strict) {
+		Objects.requireNonNull(stack);
```
apple502j: Error message?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/FabricIngredient.java
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * {@return the backing {@link CustomIngredient} of this ingredient if it's custom, {@code null} otherwise}.
+	 */
+	@Nullable
+	default CustomIngredient getCustomIngredient() {
+		return null;
+	}
+
+	/**
+	 * @see CustomIngredient#requiresTesting()
```
apple502j: Should this document the behavior for vanilla?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/CustomIngredientImpl.java
```diff
@@ -0,0 +1,122 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.ingredient.v1.CustomIngredient;
+import net.fabricmc.fabric.api.ingredient.v1.CustomIngredientSerializer;
+import net.fabricmc.fabric.impl.ingredient.builtin.AndIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.DifferenceIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.NbtIngredient;
+import net.fabricmc.fabric.impl.ingredient.builtin.OrIngredient;
+
+@ApiStatus.Internal
```
apple502j: Remove.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/ShapelessMatch.java
```diff
@@ -0,0 +1,96 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+
+/**
+ * Helper class to perform a shapeless recipe match when ingredients that require testing are involved.
+ *
+ * <p>The problem to solve is a maximum cardinality bipartite matching, for which this implementation uses the augmenting path algorithm.
+ * This has good performance in simple cases, and sufficient O(N^3) asymptotic complexity in the worst case.
+ */
+@ApiStatus.Internal
```
apple502j: Remove.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/builtin/AndIngredient.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.ingredient.v1.CustomIngredientSerializer;
+
+@ApiStatus.Internal
```
apple502j: Remove.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/builtin/AndIngredient.java
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		return true;
+	}
+
+	@Override
+	public ItemStack[] getMatchingStacks() {
+		List<ItemStack> previewStacks = new ArrayList<>();
+
+		if (ingredients.length > 0) {
```
apple502j: Seems unnecessary due to check below. Comment instead?

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/builtin/CombinedIngredient.java
```diff
@@ -0,0 +1,115 @@
...
+	protected CombinedIngredient(Ingredient[] ingredients) {
+		if (ingredients.length == 0) {
+			throw new IllegalArgumentException("AND or OR ingredient must have at least one sub-ingredient");
+		}
+
+		this.ingredients = ingredients;
+	}
+
+	@Override
+	public boolean requiresTesting() {
```
apple502j: I believe this is incorrect, at least for OR.

If ingredient A does not require testing and ingredient B requires testing, the whole ingredient needs testing, otherwise B is never matched.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	public static Ingredient nbt(Ingredient base, @Nullable NbtCompound nbt, boolean strict) {
+		Objects.requireNonNull(base);
+
+		return new NbtIngredient(base, nbt, strict).toVanilla();
+	}
+
+	/**
+	 * Create an ingredient that matches the passed template stack, NBT included.
+	 * This check can either be strict (the exact NBT must match) or non-strict (the template NBT must be a subset of the stack NBT).
```
apple502j: Document that this crashes with NBT-less stack and non-strict mode. Or rework the specs on this.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/builtin/OrIngredient.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.ingredient.v1.CustomIngredientSerializer;
+
+@ApiStatus.Internal
```
apple502j: Remove.

fabric-ingredient-api-v1/src/main/resources/fabric-ingredient-api-v1.accesswidener
```diff
@@ -0,0 +1,5 @@
+accessWidener   v1  named
```
apple502j:
```suggestion
accessWidener   v2  named
```

fabric-ingredient-api-v1/src/main/resources/fabric-ingredient-api-v1.mixins.json
```diff
@@ -0,0 +1,12 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.ingredient",
+  "compatibilityLevel": "JAVA_16",
```
apple502j:
```suggestion
  "compatibilityLevel": "JAVA_17",
```

fabric-ingredient-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "mixins": [
+    "fabric-ingredient-api-v1.mixins.json"
+  ],
+  "accessWidener": "fabric-ingredient-api-v1.accesswidener",
+  "depends": {
+    "fabricloader": ">=0.14.9"
```
apple502j:
```suggestion
    "fabricloader": ">=0.14.11"
```

fabric-ingredient-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "description": "Creation of custom ingredients",
```
apple502j: Should be more descriptive.

apple502j (changes requested):

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
Technici4n: Then `RecipeSerializer` is wrong.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
Technici4n: It's nicer not to need to create a new object imo (and the JsonObject already contains the serializer ID by then, iirc).

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/impl/ingredient/builtin/CombinedIngredient.java
Technici4n: 🤦

fabric-ingredient-api-v1/src/main/resources/fabric.mod.json
Technici4n: I just downgraded to `0.14.10` because that's what Fabric API is using at the moment.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/mixin/ingredient/IngredientMixin.java
Technici4n: Done: https://github.com/FabricMC/fabric/pull/2586/commits/9a154dafb484bca11ed4f243b36758756c356ac2.

fabric-ingredient-api-v1/src/main/java/net/fabricmc/fabric/api/ingredient/v1/CustomIngredientSerializer.java
apple502j: Ah, in that case keep it as is.

Technici4n:
I just implemented a graceful fallback mechanism to make this API work nicely for clients that don't support (some or all) custom ingredients. Summary follows:
- On login, server requests the client's available ingredients.
- The client answers by listing the serializer IDs it supports.
- The server stores that where it's relevant.
- When an ingredient is serialized, we lookup which IDs are supported by the client. If we can lookup and the client doesn't support the ingredient, we send it as a vanilla stack list ingredient instead. Otherwise, we send the custom ingredient as usual.

List of cases to test:
- [x] Server has ingredient API, client doesn't have the API.
- [x] Server has ingredient API, client has the API but doesn't support some ingredients.
- [x] Server has ingredient API, client has the API and all the used ingredients.

Technici4n:
After some consideration, this new module will be renamed Fabric Recipe API v1, and the ingredient stuff will be moved to the `net.fabricmc.fabric.api.recipe.v1.ingredient` package, in preparation for possible other kinds of recipe format extensions.

Technici4n:
image

Works fine on vanilla clients now (ignoring the recipe book) !

Technici4n:
Partial support on clients also works: (debug output on server, removed some ingredients on client)
```
[15:28:36] [Server thread/INFO]: Technici4n joined the game
Wrote fabric:nbt as a vanilla ingredient
Wrote fabric:and as a custom ingredient
Wrote fabric:difference as a vanilla ingredient
Wrote fabric:nbt as a vanilla ingredient
Wrote fabric:or as a custom ingredient
```

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,143 @@
...
+	 *         // sub-ingredient 1,
+	 *         // sub-ingredient 2,
+	 *         // etc...
+	 *     ]
+	 * }
+	 * }</pre>
+	 *
+	 * @throws IllegalArgumentException if the array is empty
+	 */
+	public static Ingredient and(Ingredient... ingredients) {
```
modmuss50: Nit: Maybe call this `all`

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
```diff
@@ -0,0 +1,143 @@
...
+	 *         // sub-ingredient 1,
+	 *         // sub-ingredient 2,
+	 *         // etc...
+	 *     ]
+	 * }
+	 * }</pre>
+	 *
+	 * @throws IllegalArgumentException if the array is empty
+	 */
+	public static Ingredient or(Ingredient... ingredients) {
```
modmuss50: Nit: Maybe call this `any`?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -0,0 +1,143 @@
...
+	public static final int PACKET_MARKER = -1;
+
+	static final Map<Identifier, CustomIngredientSerializer<?>> REGISTERED_SERIALIZERS = new ConcurrentHashMap<>();
+
+	public static void registerSerializer(CustomIngredientSerializer<?> serializer) {
+		Objects.requireNonNull(serializer.getIdentifier(), "CustomIngredientSerializer identifier may not be null.");
+
+		if (REGISTERED_SERIALIZERS.putIfAbsent(serializer.getIdentifier(), serializer) != null) {
+			throw new IllegalArgumentException("CustomIngredientSerializer with identifier " + serializer.getIdentifier() + " already registered.");
+		}
```
modmuss50, line 54: Nit: check if the key exists before adding, as a caller can catch the `IllegalArgumentException` to overwrite.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -0,0 +1,143 @@
...
+
+	@Nullable
+	public static CustomIngredientSerializer<?> getSerializer(Identifier identifier) {
+		Objects.requireNonNull(identifier, "Identifier may not be null.");
+
+		return REGISTERED_SERIALIZERS.get(identifier);
+	}
+
+	static {
+		CustomIngredientSerializer.register(AndIngredient.SERIALIZER);
```
modmuss50: Question, is this OK class loading wise? Might be better moved to an init entrypoint to be sure its registered?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -0,0 +1,143 @@
...
+		customIngredient.getSerializer().write(json, coerceIngredient());
+		return json;
+	}
+
+	@Override
+	public boolean isEmpty() {
+		// We don't want to resolve the matching stacks,
+		// as this might cause the ingredient to use outdated tags when it's done too early.
+		// So we just return false when the matching stacks haven't been resolved yet (i.e. when the field is null).
+		// TODO: this is a bit hacky, can we not do better?
```
modmuss50: When is this called to early? Would hard crashing be better?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientSync.java
```diff
@@ -0,0 +1,87 @@
...
+ * <p><ul>
+ *     <li>Each client sends a packet with the set of custom ingredients it supports.</li>
+ *     <li>We store that set inside the {@link PacketEncoder} using {@link PacketEncoderMixin}.</li>
+ *     <li>When serializing a custom ingredient, we get access to the current {@link PacketEncoder},
+ *     and based on that we decide whether to send the custom ingredient, or a vanilla ingredient with the matching stacks.</li>
+ * </ul>
+ */
+public class CustomIngredientSync implements ModInitializer {
+	public static final Identifier PACKET_ID = new Identifier("fabric", "custom_ingredient_sync");
+	public static final int PROTOCOL_VERSION = 0;
```
modmuss50: Nit: Start with 1 for the first version? Not a major issue though.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
Technici4n: Could also be `intersection`, not sure what's best.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
Technici4n: Could also be `union`, not sure what's best.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
modmuss50: Java's Stream class uses `allMatch` and `anyMatch` I think it would make sense to follow it's lead. I dont think the `match` part is needed in this context however?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
Technici4n: It works fine because the various ingredients don't reference `CustomIngredientImpl` (as they're using only the API).

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/DefaultCustomIngredients.java
62832: I'd argue for `any`/`all` since they're simpler. `union` and `intersection` seem a bit out of place as more set-theoretical terms.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
Technici4n: That's why it's a `putIfAbsent`. :smile:

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
modmuss50: Ah, this is fine. just seen that `CustomIngredientSerializer.get` actually calls `CustomIngredientImpl.getSerializer`. 👍 Might still be a tiny bit cleaner in the mod entrypoint, but not a major issue.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
modmuss50: Ah yeah, good point. This is fine 👍

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
Technici4n: image

For example when used in a shapeless recipe. Vanilla tries to filter out empty ingredients for some reason, but tags haven't been bound yet, so the ingredient should not be resolved.

I don't know if we need to do anything about this, and in my testing this was never really an issue, so I'd suggest we leave it as-is for now. I removed the TODO.

Technici4n:
Just rebased to fix the `gradle.properties` conflict, and switched `getMatchingStacks` to a `List<ItemStack>`.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientSync.java
```diff
@@ -0,0 +1,87 @@
...
+ * <p><ul>
+ *     <li>Each client sends a packet with the set of custom ingredients it supports.</li>
+ *     <li>We store that set inside the {@link PacketEncoder} using {@link PacketEncoderMixin}.</li>
+ *     <li>When serializing a custom ingredient, we get access to the current {@link PacketEncoder},
+ *     and based on that we decide whether to send the custom ingredient, or a vanilla ingredient with the matching stacks.</li>
+ * </ul>
+ */
+public class CustomIngredientSync implements ModInitializer {
+	public static final Identifier PACKET_ID = new Identifier("fabric", "custom_ingredient_sync");
+	public static final int PROTOCOL_VERSION = 1;
```
Technici4n: We should rather have the server send the max supported protocol id, and the client answer back with the protocol it chooses. (e.g. in the future if we add protocol version 2 the server will send 1 or 2 depending on whether it supports it, and the client will reply with the highest protocol version that itself and the server support).

Technici4n: Done

modmuss50:
I expect out of scope for this PR, but having a condition for a custom ingredient might be useful.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/PacketEncoderMixin.java
```diff
@@ -0,0 +1,77 @@
...
+
+	@Inject(
+			at = @At(
+					value = "INVOKE",
+					target = "net/minecraft/network/Packet.write(Lnet/minecraft/network/PacketByteBuf;)V"
+			),
+			method = "encode(Lio/netty/channel/ChannelHandlerContext;Lnet/minecraft/network/Packet;Lio/netty/buffer/ByteBuf;)V"
+	)
+	private void capturePacketEncoder(ChannelHandlerContext channelHandlerContext, Packet<?> packet, ByteBuf byteBuf, CallbackInfo ci) {
+		CustomIngredientSync.CURRENT_SUPPORTED_INGREDIENTS.set(fabric_supportedCustomIngredients);
```
modmuss50, line 54: Does this need to be done for all packets? Checking the Packet is an instance of the specific recipe packet may be better?

modmuss50 (approved):
Look great. Before merging I will make sure to thoroughly test the networking stuff with vanilla and modded clients.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/PacketEncoderMixin.java
Technici4n: The instanceof check will likely be more expensive than the ThreadLocal set.

PR #2591, opened 2022-10-16
Create `ko_kr` translations for `resource-loader` and `item-groups`

Dqlsu:
Added Korean translation.

### DIFF
--- a/fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/ko_kr.json
+++ b/fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/ko_kr.json
@@ -1,0 +1,4 @@
+{
+  "pack.source.fabricmod": "Fabric 모드",
+  "pack.source.builtinMod": "%s에 내장됨"
+}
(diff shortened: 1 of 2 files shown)


fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/ko_kr.json
```diff
@@ -0,0 +1,4 @@
+{
+  "pack.source.fabricmod": "Fabric 모드",
+  "pack.source.builtinMod": "%s에 내장됨"
```
ghost, line 3: Are you sure this works?
The %s maybe needs a space after.

apple502j: I don't think we need a space there.

apple502j (approved):

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/ko_kr.json
ghost: Is not displayed, however I think this is a vanilla problem not displaying long strings.

Immagine 2022-10-16 190214
Immagine 2022-10-16 185254

ghost (approved):

modmuss50 (approved):

PR #2593, opened 2022-10-18
Edit `ko-kr` translations for `resource-loader`

Dqlsu:


Technici4n:
Why is the old translation not good?

Dqlsu:
It has the same meaning, but the current translation is a bit more natural.

ghost (approved):
Immagine 2022-10-19 152443
Immagine 2022-10-19 153035


I agree because vanilla translation for "built-in" is the same, infact i didn't liked the previous one.

Technici4n (approved):

PR #2599, opened 2022-10-20
22w42a 🐫

modmuss50:
# 22w42a port

This PR ports Fabric API to 22w42a. There are a number of changes that affect fabric api on a lower level as well as higher level changes that will affect most modders. Big thanks to @shartte for helping with this port. Also thanks to the yarn contributors for the mappings.

Fabric loader 0.14.10 or later is required.

## Item Groups

22w42a changed the way ItemGroups work. `FabricItemGroupBuilder` has been removed and replaced with `FabricItemGroup`. The following snippet of code shows how to create an item group, and add the `TEST_ITEM` to it. This allows for much finer grain control over the ordering of your item groups.

```java
private static final ItemGroup ITEM_GROUP = new FabricItemGroup(new Identifier(MOD_ID, "test_group")) {
		@Override
		public ItemStack createIcon() {
			return new ItemStack(Items.DIAMOND);
		}

		@Override
		protected void addItems(FeatureSet enabledFeatures, Entries entries) {
			entries.add(TEST_ITEM);
		}
	};
```

If you wish to add items to vanilla's groups, or ones added via a mod you can do so with Fabric's new event. Unlike 1.19.2 you also have fine grain control over the ordering of other item groups.

```java
ItemGroupEvents.modifyEntriesEvent(ItemGroups.BUILDING_BLOCKS).register((content) -> {
		content.add(TEST_ITEM);

		// Add a diamond before the oak fence
		content.addBefore(Blocks.OAK_FENCE, Items.DIAMOND);
		// Add an emerald after the oak door.
		content.addAfter(Blocks.OAK_DOOR, Items.EMERALD);
});
```

## Biome API changes

Previously dynamic registries would source their content from both the `BuiltInRegistries` and data packs. Now the built-in registries are only used for data generation,
and content at runtime is only loaded from data packs.

Mods that previously used JSON objects for these registries exclusively are unaffected by this change. Mods that registered their worldgen content to `BuiltinRegistries` will now have
to generate JSON files and include them in their mod assets. A convenience data provider has been added (`FabricBuiltinRegistriesProvider`) 
that will automatically export JSON files for all objects in `BuiltinRegistries` belonging to the current mod. This means for mods that previously
did built-in worldgen, this should be the only required change to continue working.

The biome modification API will continue to work as before, and will throw more detailed error messages when world-gen 
objects existing only in `BuiltinRegistries` but not in data packs are used.

The following data types are affected by this:

* Dimension Types
* Biomes
* Message Types
* Configured Carvers
* Configured Features
* Placed Features
* Structures
* Structure Sets
* Structure Processor Lists
* Structure Pools
* Chunk Generator Settings
* Noise
* Density Functions
* World Presets
* Flat Level Generator Presets

## Data Generation

Minor breaking changes, recipes now require a `RecipeCategory`

# Texture API

The sprites that should be baked into texture atlases are no longer discovered by querying all `UnbakedModel` during the model baking process. Instead, the `SpriteAtlasManager` will independently search for textures using patterns in all available resource packs and bake them into specific texture atlases.

For example, `BakedModelManager` defines the block atlas as:

* All textures in the `block` folder (or any subfolder)
* All textures in the `item` folder (or any subfolder)
* All textures in the `entity/conduit` folder (or any subfolder)
* The texture at `BellBlockEntityRenderer.BELL_BODY_TEXTURE`
* The texture at `EnchantingTableBlockEntityRenderer.BOOK_TEXTURE`

In Fabric, the `ClientSpriteRegistryCallback` event has been modified to allow new rules to be appended for any texture atlas.

For example, the following would add all textures from the subfolder `part` (recursively) to the atlas (from *any* resource pack/namespace), and also the texture called "ae2:models/skyblockchest" specifically.

```java
ClientSpriteRegistryCallback.event(SpriteAtlasTexture.BLOCK_ATLAS_TEXTURE).register((resourceManager, sprites) -> {
  class_7766.method_45834(resourceManager, "part", sprites::put);
  class_7766.method_45830(resourceManager, new Identifier("ae2:models/skyblockchest"), sprites::put);
});
```

## UnbakedModel Changes

UnbakedModels no longer have texture dependencies (see Texture API), so the corresponding method was removed without a replacement.
Textures must either be in block/ or item/ to be discovered for the block atlas, or the Texture API needs to be used to add them to the atlas.

In addition, Vanilla changed how parent models are loaded by JSON models. It introduced `method_45785` to allow a model to load its parent models.
This method is only called for models that are initially added to the model load list. If your mod subclasses `UnbakedModel` and uses nested JSON models,
it should always call `method_45785` on its own model dependencies. The used JSON models would not be able to load their parents otherwise.

Example implementation:

```java
class CustomUnbakedModel implements UnbakedModel {
[...]

  @Override
  default void method_45785(Function<Identifier, UnbakedModel> loader) {
    for (Identifier dependency : getModelDependencies()) {
      loader.apply(dependency).method_45785(loader);
    }
  }
}
```


fabric-message-api-v1/src/main/java/net/fabricmc/fabric/mixin/message/PlayerManagerMixin.java
```diff
@@ -42,15 +42,13 @@ public abstract class PlayerManagerMixin {
...
-	public abstract void sendMessageHeader(SignedMessage message, Set<ServerPlayerEntity> except);
-
 	@Inject(method = "broadcast(Lnet/minecraft/network/message/SignedMessage;Lnet/minecraft/server/network/ServerPlayerEntity;Lnet/minecraft/network/message/MessageType$Parameters;)V", at = @At("HEAD"), cancellable = true)
 	private void onSendChatMessage(SignedMessage message, ServerPlayerEntity sender, MessageType.Parameters params, CallbackInfo ci) {
 		if (!ServerMessageEvents.ALLOW_CHAT_MESSAGE.invoker().allowChatMessage(message, sender, params)) {
-			if (!message.headerSignature().isEmpty()) {
-				sendMessageHeader(message, Set.of());
-			}
+			// TODO 22w42a
+//			if (!message.headerSignature().isEmpty()) {
```
Gegy: These shouldn’t be needed at all anymore - the protocol is now able to safely handle gaps in players’ message chains 🙂

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
```diff
@@ -56,81 +52,58 @@ public GroupResourcePack(ResourceType type, List<ModResourcePack> packs) {
...
-		}
-
-		Set<Identifier> resources = new HashSet<>();
-
 		for (int i = packs.size() - 1; i >= 0; i--) {
 			ResourcePack pack = packs.get(i);
-			Collection<Identifier> modResources = pack.findResources(type, namespace, prefix, predicate);
 
-			resources.addAll(modResources);
+			pack.findResources(type, namespace, prefix, arg);
```
Gegy, line 79: Not sure if it’s an issue, but question: is it guaranteed that mod packs within a group won’t have intersecting resources? Might need to filter those duplicates before invoking the consumer 🙂

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
```diff
@@ -242,7 +236,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
 
 						try {
 							Identifier id = new Identifier(namespace, nsPath.relativize(file).toString().replace(separator, "/"));
-							if (predicate.test(id)) ids.add(id);
+							visitor.accept(id, () -> Files.newInputStream(file));
```
Gegy: A sneaky change happened here: packs now need to list their mcmeta files, too

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
LambdAurora: As Gegy said, this absolutely needs filtering to de-dupe as there's no guarantee that mod resource packs won't have intersecting resources in the same group!

shartte: I am not sure that vanilla does this either, however. Anyway, one more item on the big pile of "needs to be checked" things.

LambdAurora: Vanilla doesn't have the dupe issue because that method exists only on resource packs that have a single root so the calls are always returning unique entries.

And ResourceManager has a very different signature, meaning when looking at resource packs it'll keep the one that has the highest priority.

shartte: This method is also used in building the resource stack which returns a list of resources with the same identifier. Should that not return the asset from all mods that have it?

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseGameTest.java
```diff
@@ -18,13 +18,11 @@
...
 import org.spongepowered.asm.mixin.MixinEnvironment;
 
-import net.minecraft.test.GameTest;
 import net.minecraft.test.TestContext;
 
-import net.fabricmc.fabric.api.gametest.v1.FabricGameTest;
-
 public class FabricApiBaseGameTest {
-	@GameTest(templateName = FabricGameTest.EMPTY_STRUCTURE)
+	// TODO 22w42a
```
apple502j: Why is this disabled?

fabric-item-group-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -16,16 +16,15 @@
     "FabricMC"
   ],
   "depends": {
-    "fabricloader": ">=0.6.0",
-    "minecraft": ">=1.15-",
+    "fabricloader": ">=0.14.9",
```
apple502j, line 19:
```suggestion
    "fabricloader": ">=0.14.10",
```

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/BootstrapMixin.java
```diff
@@ -29,6 +29,7 @@
 import net.minecraft.fluid.Fluids;
 import net.minecraft.item.Items;
 import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.SimpleRegistry;
```
apple502j: Is this import used?

fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/ServerResourcePackProviderMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.resource.ServerResourcePackProvider;
+import net.minecraft.resource.ResourcePackProfile;
+
+import net.fabricmc.fabric.impl.resource.loader.ModResourcePackCreator;
+
+@Mixin(ServerResourcePackProvider.class)
```
apple502j, line 31: I feel like this should mixin to `DefaultClientResourcePackProvider`.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/MainMixin.java
```diff
@@ -20,15 +20,15 @@
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Redirect;
 
-import net.minecraft.resource.DataPackSettings;
+import net.minecraft.class_7712;
 import net.minecraft.server.Main;
 
 import net.fabricmc.fabric.impl.resource.loader.ModResourcePackUtil;
 
 @Mixin(Main.class)
```
apple502j: I don't think this Mixin applies.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/TestServerMixin.java
```diff
@@ -16,34 +16,19 @@
...
-import net.minecraft.resource.DataPackSettings;
-import net.minecraft.resource.ResourceManager;
+import net.minecraft.class_7712;
 import net.minecraft.test.TestServer;
-import net.minecraft.util.dynamic.RegistryOps;
-import net.minecraft.util.registry.DynamicRegistryManager;
 
 import net.fabricmc.fabric.impl.resource.loader.ModResourcePackUtil;
 
 @Mixin(TestServer.class)
```
apple502j, line 34: Don't think this is necessary.

apple502j (changes requested):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
LambdAurora: ....
Damn it Mojang!

Gegy: Hmm, true! It does look like stack building will work correctly when listing all resource stacks. Single resource lookups will still only yield that first entry, though- might be a weird inconsistency (also from previous behaviour).

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/FabricApiBaseGameTest.java
modmuss50: Ah, had to disable to get it building when the gametest module wasnt enabled, I will fix later 👍

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/GroupResourcePack.java
shartte: I kinda view this method as "visitResources" and the last argument as the visitor.
There are a few vanilla methods that just put the visited resources in a map. So in that case, the *last* visited resource for a given Identifier will win. We should absolutly verify: in which order does Vanilla visit the resource packs (high prio -> low prio? or reverse) and potentially at least match that deterministically (alphabetical order by mod-id maybe?)

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java
shartte: Yes they do, I ran into the problem that skipping .mcmeta files in findResources actually breaks animated textures.

shartte: Or are you talking about pack.mcmeta?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/MainMixin.java
shartte: This was reworked

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/TestServerMixin.java
shartte: It turned out it was, but was reworked anyway. -> Enabling all resource packs like Vanilla does will enable them in the wrong order leading to Vanilla overriding mod resources.

modmuss50:
This should now be ready to merge very shortly. I expect there will be stuff broken, wrong and missing but we can fix that as people find the issues.

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemConvertible;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemStack;
+import net.minecraft.resource.featuretoggle.FeatureSet;
+
+/**
+ * This class allows the entries of {@link ItemGroup item groups} to be modified by the events in {@link ItemGroupEvents}.
```
Juuxel:
```suggestion
 * This class allows the entries of {@link ItemGroup item groups} to be modified by the events in {@link ItemGroupEvents}.
```
```suggestion
 * This class allows the entries of {@linkplain ItemGroup item groups} to be modified by the events in {@link ItemGroupEvents}.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+		}
+		case PARENT_TAB_ONLY -> addAfter(afterLast, newStacks, displayStacks);
+		case SEARCH_TAB_ONLY -> addAfter(afterLast, newStacks, searchTabStacks);
+		}
+	}
+
+	/**
+	 * Adds stacks after an existing stack in the group, or at the end, if the stack isn't in the group.
+	 *
+	 * @param afterLast  Add <code>newStacks</code> after the last group entry matching this stack (compared using {@link ItemStack#canCombine}).
```
Juuxel:
```suggestion
	 * @param afterLast  Add {@code newStacks} after the last group entry matching this stack (compared using {@link ItemStack#canCombine}).
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks  The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks  The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * See {@link #addAfter(ItemStack, Collection, net.minecraft.item.ItemGroup.StackVisibility)}.
+	 */
+	public void addAfter(ItemStack afterLast, Collection<ItemStack> newStacks) {
+		addAfter(afterLast, newStacks, ItemGroup.StackVisibility.PARENT_AND_SEARCH_TABS);
+	}
+
+	/**
+	 * Adds stacks after an existing item in the group, or at the end, if the item isn't in the group.
+	 *
+	 * @param afterLast  Add <code>newStacks</code> after the last entry of this item in the group.
```
Juuxel:
```suggestion
	 * @param afterLast  Add {@code newStacks} after the last entry of this item in the group.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks  The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks  The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+		}
+		case PARENT_TAB_ONLY -> addAfter(afterLast, newStacks, displayStacks);
+		case SEARCH_TAB_ONLY -> addAfter(afterLast, newStacks, searchTabStacks);
+		}
+	}
+
+	/**
+	 * Adds stacks after the last group entry matching a predicate, or at the end, if no entries match.
+	 *
+	 * @param afterLast  Add <code>newStacks</code> after the last group entry matching this predicate.
```
Juuxel:
```suggestion
	 * @param afterLast  Add {@code newStacks} after the last group entry matching this predicate.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks  The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks  The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * See {@link #addBefore(ItemStack, Collection, net.minecraft.item.ItemGroup.StackVisibility)}.
+	 */
+	public void addBefore(ItemStack beforeFirst, Collection<ItemStack> newStacks) {
+		addBefore(beforeFirst, newStacks, ItemGroup.StackVisibility.PARENT_AND_SEARCH_TABS);
+	}
+
+	/**
+	 * Adds stacks before an existing item in the group, or at the end, if the item isn't in the group.
+	 *
+	 * @param beforeFirst Add <code>newStacks</code> before the first entry of this item in the group.
```
Juuxel:
```suggestion
	 * @param beforeFirst Add {@code newStacks} before the first entry of this item in the group.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks   The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks   The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+		}
+		case PARENT_TAB_ONLY -> addBefore(beforeFirst, newStacks, displayStacks);
+		case SEARCH_TAB_ONLY -> addBefore(beforeFirst, newStacks, searchTabStacks);
+		}
+	}
+
+	/**
+	 * Adds stacks before an existing stack to the group, or at the end, if the stack isn't in the group.
+	 *
+	 * @param beforeFirst Add <code>newStacks</code> before the first group entry matching this stack (compared using {@link ItemStack#canCombine}).
```
Juuxel:
```suggestion
	 * @param beforeFirst Add {@code newStacks} before the first group entry matching this stack (compared using {@link ItemStack#canCombine}).
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks   The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks   The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+		}
+		case PARENT_TAB_ONLY -> addBefore(beforeFirst, newStacks, displayStacks);
+		case SEARCH_TAB_ONLY -> addBefore(beforeFirst, newStacks, searchTabStacks);
+		}
+	}
+
+	/**
+	 * Adds stacks before the first group entry matching a predicate, or at the end, if no entries match.
+	 *
+	 * @param beforeFirst Add <code>newStacks</code> before the first group entry matching this predicate.
```
Juuxel:
```suggestion
	 * @param beforeFirst Add {@code newStacks} before the first group entry matching this predicate.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param newStacks   The stacks to add. Only {@link #isEnabled(ItemStack) enabled} stacks will be added.
```
Juuxel:
```suggestion
	 * @param newStacks   The stacks to add. Only {@linkplain #isEnabled(ItemStack) enabled} stacks will be added.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/IdentifiableItemGroup.java
```diff
@@ -14,21 +14,19 @@
...
- * that should be processed prior to this sprite.
- *
- * <p>Best used in conjunction with {@link net.fabricmc.fabric.api.event.client.ClientSpriteRegistryCallback}.
+ * This interface is automatically implemented on all {@link ItemGroup} via Mixin and interface injection.
  */
-public interface DependentSprite {
+public interface IdentifiableItemGroup {
 	/**
-	 * @return A set of all sprites that should be loaded before this sprite.
+	 * @return The unique identifier of this {@link ItemGroup}.
```
Juuxel:
```suggestion
	 * {@return the unique identifier of this {@link ItemGroup}}
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupAccessor.java
```diff
@@ -14,16 +14,17 @@
...
-import net.minecraft.resource.ResourcePackSource;
+import net.minecraft.item.ItemGroup;
 
-@Mixin(FileResourcePackProvider.class)
-public interface FileResourcePackProviderAccessor {
-	@Accessor("source")
-	ResourcePackSource getResourcePackSource();
+@Mixin(ItemGroup.class)
+public interface ItemGroupAccessor {
+	@Accessor("index")
```
Juuxel:
```suggestion
	@Accessor
```
The field name is not needed since it matches the setter

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupsAccessor.java
```diff
@@ -14,29 +14,28 @@
...
-	@Override
-	public void fabric_expandArray() {
-		ItemGroup[] tempGroups = GROUPS;
-		GROUPS = new ItemGroup[GROUPS.length + 1];
+	static void setGroups(ItemGroup[] groups) {
+		throw new AssertionError();
+	}
 
-		System.arraycopy(tempGroups, 0, GROUPS, 0, tempGroups.length);
+	@Invoker("asArray")
```
Juuxel:
```suggestion
	@Invoker
```

fabric-item-group-api-v1/src/client/resources/fabric-item-groups-v1.client.mixins.json
```diff
@@ -1,6 +1,6 @@
 {
```
Juuxel: This file should be called fabric-item-group-api-v1.client.mixins.json

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupsMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemGroups;
+
+@Mixin(ItemGroups.class)
+public class ItemGroupsMixin {
+	@Inject(method = "asArray", at = @At("HEAD"))
+	private static void asArray(ItemGroup[] groups, CallbackInfoReturnable<ItemGroup[]> cir) {
+		// Ensure that all item groups have a none null id.
```
Juuxel:
```suggestion
		// Ensure that all item groups have a nonnull id.
```
(or "non-null" but javax annotations uses this spelling)

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupsMixin.java
```diff
@@ -0,0 +1,38 @@
...
+		for (ItemGroup group : groups) {
+			if (group.getId() == null) {
+				throw new NullPointerException("Item group %s has a null id".formatted(group.getClass()));
```
Juuxel:
```suggestion
				throw new NullPointerException("Item group %s has a null id".formatted(group.getClass().getName()));
```
"Item group class mymod.MyItemGroup has..." is a bit ugly

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupsMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemGroups;
+
+@Mixin(ItemGroups.class)
+public class ItemGroupsMixin {
```
Juuxel:
```suggestion
abstract class ItemGroupsMixin {
```
Mixins should be package-private according to fabric conventions

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupMixin.java
```diff
@@ -0,0 +1,95 @@
...
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.itemgroup.v1.IdentifiableItemGroup;
+import net.fabricmc.fabric.api.itemgroup.v1.FabricItemGroupEntries;
+import net.fabricmc.fabric.api.itemgroup.v1.ItemGroupEvents;
+import net.fabricmc.fabric.impl.itemgroup.ItemGroupEventsImpl;
+import net.fabricmc.fabric.impl.itemgroup.MinecraftItemGroups;
+
+@Mixin(ItemGroup.class)
+public class ItemGroupMixin implements IdentifiableItemGroup {
```
Juuxel:
```suggestion
abstract class ItemGroupMixin implements IdentifiableItemGroup {
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/impl/itemgroup/MinecraftItemGroups.java
```diff
@@ -0,0 +1,60 @@
...
+	public static final Identifier HOTBAR_ID = new Identifier("minecraft:hotbar");
+	public static final Identifier SEARCH_ID = new Identifier("minecraft:search");
+	public static final Identifier TOOLS_ID = new Identifier("minecraft:tools");
+	public static final Identifier COMBAT_ID = new Identifier("minecraft:combat");
+	public static final Identifier CONSUMABLES_ID = new Identifier("minecraft:consumables");
+	public static final Identifier CRAFTING_ID = new Identifier("minecraft:crafting");
+	public static final Identifier SPAWN_EGGS_ID = new Identifier("minecraft:spawn_eggs");
+	public static final Identifier INVENTORY_ID = new Identifier("minecraft:inventory");
+
+	public static final Map<ItemGroup, Identifier> MAP = new ImmutableMap.Builder<ItemGroup, Identifier>()
```
Juuxel: Maybe something more descriptive like `GROUP_ID_MAP`?

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/impl/itemgroup/ItemGroupHelper.java
```diff
@@ -0,0 +1,50 @@
...
+	public static void appendItemGroup(FabricItemGroup itemGroup) {
+		for (ItemGroup existingGroup : ItemGroups.GROUPS) {
+			if (existingGroup.getId().equals(itemGroup.getId())) {
+				throw new IllegalStateException("Duplicate item group: " + itemGroup.getId());
+			}
+		}
+
+		final int index = ItemGroups.GROUPS.length;
+		final ItemGroup[] itemGroups = Stream.concat(Arrays.stream(ItemGroups.GROUPS), Stream.of(itemGroup))
+				.toArray(ItemGroup[]::new);
```
Juuxel: Don't we have the apache commons lib that has a util method for adding an element to an array? Doesn't really matter though tbh

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
```diff
@@ -0,0 +1,59 @@
...
+
+import net.minecraft.item.ItemGroup;
+import net.minecraft.resource.featuretoggle.FeatureSet;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.impl.itemgroup.ItemGroupEventsImpl;
+
+public class ItemGroupEvents {
```
Juuxel:
```suggestion
public final class ItemGroupEvents {
```
Should also have a private constructor, and it needs class jd

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
```diff
@@ -0,0 +1,59 @@
...
+			callback.modifyEntries(group, featureSet, entries);
+		}
+	});
+
+	public static Event<ModifyEntries> modifyEntriesEvent(ItemGroup itemGroup) {
+		return modifyEntriesEvent(itemGroup.getId());
+	}
+
+	public static Event<ModifyEntries> modifyEntriesEvent(Identifier identifier) {
+		return ItemGroupEventsImpl.getOrCreateModifyEntriesEvent(identifier);
```
Juuxel:
```suggestion
	public static Event<ModifyEntries> modifyEntriesEvent(Identifier groupId) {
		return ItemGroupEventsImpl.getOrCreateModifyEntriesEvent(groupId);
```
Just "identifier" doesn't really describe *what* it identifies

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
```diff
@@ -0,0 +1,59 @@
...
+
+	@FunctionalInterface
+	public interface ModifyEntries {
+		void modifyEntries(FeatureSet featureSet, FabricItemGroupEntries entries);
+	}
+
+	@FunctionalInterface
+	public interface ModifyEntriesAll {
+		void modifyEntries(ItemGroup group, FeatureSet featureSet, FabricItemGroupEntries entries);
+	}
```
Juuxel: Do we really need two separate interfaces for this? IMO `ModifyEntries` could just have the group param

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/IdentifiableItemGroup.java
```diff
@@ -14,21 +14,19 @@
...
 
+import net.minecraft.item.ItemGroup;
 import net.minecraft.util.Identifier;
 
 /**
- * Implement this interface on a Sprite to declare additional dependencies
- * that should be processed prior to this sprite.
- *
- * <p>Best used in conjunction with {@link net.fabricmc.fabric.api.event.client.ClientSpriteRegistryCallback}.
+ * This interface is automatically implemented on all {@link ItemGroup} via Mixin and interface injection.
```
Juuxel:
```suggestion
 * This interface is automatically injected to {@link ItemGroup}.
```
This is how it's done in existing code (loot api), mixins are an impl detail

fabric-item-group-api-v1/src/testmod/resources/fabric.mod.json
```diff
@@ -1,12 +1,12 @@
 {
   "schemaVersion": 1,
-  "id": "fabric-item-groups-v0-testmod",
-  "name": "Fabric Item Groups (v0) Test Mod",
+  "id": "fabric-item-groups-v1-testmod",
+  "name": "Fabric Item Groups (v1) Test Mod",
```
Juuxel:
```suggestion
  "id": "fabric-item-group-api-v1-testmod",
  "name": "Fabric Item Group API (v1) Test Mod",
```
The mod id in the code is already correct btw

Juuxel (changes requested):

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupMixin.java
```diff
@@ -41,13 +45,42 @@ public class ItemGroupMixin implements IdentifiableItemGroup {
...
+		// Convert the entries to lists
+		var mutableDisplayStacks = new LinkedList<>(displayStacks);
+		var mutableSearchTabStacks = new LinkedList<>(searchTabStacks);
+		var content = new ItemGroupContent(featureSet, mutableDisplayStacks, mutableSearchTabStacks);
+
+		final Event<ItemGroupEvents.ModifyContent> modifyEntriesEvent = ItemGroupEventsImpl.getModifyEntriesEvent(getId());
 
 		if (modifyEntriesEvent != null) {
-			modifyEntriesEvent.invoker().modifyItems(featureSet, entries);
+			modifyEntriesEvent.invoker().modifyContent(featureSet, content);
```
AlphaMode: Why aren't we passing the search field here couldn't that be useful for modders?

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
shartte: I thought about it, but then all of your event listeners that target a specific group will have a useless 1st parameter

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupMixin.java
shartte: No it wouldn't be useful. The modified lists are stored/cached in the ItemGroup and used in both queries with search=true and search=false.

fabric-item-group-api-v1/src/testmod/resources/fabric.mod.json
modmuss50: The mixin config name is also wrong, will all be fixed in my next commit.

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/impl/itemgroup/ItemGroupHelper.java
modmuss50: ArrayUtils seems to do the job, thanks 👍

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/IdentifiableItemGroup.java
```diff
@@ -24,7 +24,7 @@
  */
 public interface IdentifiableItemGroup {
 	/**
-	 * {@return the unique identifier of this {@link ItemGroup}}
+	 * {@return the unique identifier of this {@link ItemGroup}}.
```
Juuxel, line 27: Does this lead to an extra dot at the end?

modmuss50: Checkstyle fails without it.

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
Juuxel: This is still unresolved

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/IdentifiableItemGroup.java
Juuxel: I know but that's not what I was asking 😛

modmuss50: I should learn to read one day...

Just checked, no it doesnt.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBuiltinRegistriesProvider.java
```diff
@@ -0,0 +1,113 @@
...
+	 * data generation.
+	 */
+	public static Function<FabricDataGenerator, DataProvider> forCurrentMod() {
+		return fabricDataGenerator -> new FabricBuiltinRegistriesProvider(
+				fabricDataGenerator.getOutput(),
+				e -> e.getValue().getNamespace().equals(fabricDataGenerator.getModId())
+		);
+	}
+
+	public void run(DataWriter writer) {
```
apple502j, line 74: Missing `@Override`?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBuiltinRegistriesProvider.java
```diff
@@ -0,0 +1,113 @@
...
+
+			if (optional.isPresent()) {
+				DataProvider.writeToPath(cache, optional.get(), path);
+			}
+		} catch (IOException var6) {
+			LOGGER.error("Couldn't save element {}", path, var6);
+		}
+	}
+
+	public String getName() {
```
apple502j, line 112: Same here.

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/IdentifiableItemGroup.java
Juuxel: TIL, then it's all good.

Juuxel (approved):

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroup.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.item.ItemGroups;
+import net.minecraft.text.Text;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.itemgroup.ItemGroupHelper;
+
+/**
+ * Extend this abstract class to create a new {@link ItemGroup} for a mod.
+ *
+ * <p>Each new instance of this class is automatically appended to {@link ItemGroups#GROUPS}
```
apple502j:
```suggestion
 * <p>Each new instance of this class is automatically appended to {@link ItemGroups#GROUPS}.
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+	public FabricItemGroupEntries(FeatureSet featureSet, List<ItemStack> displayStacks, List<ItemStack> searchTabStacks) {
+		this.featureSet = featureSet;
+		this.displayStacks = displayStacks;
+		this.searchTabStacks = searchTabStacks;
+	}
+
+	/**
+	 * @return The currently enabled feature set.
+	 */
+	public FeatureSet getFeatureSet() {
```
apple502j:
```suggestion
	public FeatureSet getEnabledFeatures() {
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -0,0 +1,464 @@
...
+		case PARENT_TAB_ONLY -> addBefore(beforeFirst, newStacks, displayStacks);
+		case SEARCH_TAB_ONLY -> addBefore(beforeFirst, newStacks, searchTabStacks);
+		}
+	}
+
+	/**
+	 * @return True if the item of a given stack is enabled in the current {@link FeatureSet}.
+	 * @see Item#method_45382
+	 */
+	public boolean isEnabled(ItemStack stack) {
```
apple502j: Why is this a public API?

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
```diff
@@ -0,0 +1,68 @@
...
+		return modifyEntriesEvent(itemGroup.getId());
+	}
+
+	public static Event<ModifyEntries> modifyEntriesEvent(Identifier groupId) {
+		return ItemGroupEventsImpl.getOrCreateModifyEntriesEvent(groupId);
+	}
+
+	@FunctionalInterface
+	public interface ModifyEntries {
+		void modifyEntries(FeatureSet featureSet, FabricItemGroupEntries entries);
```
apple502j: What is this `featureSet`?

fabric-item-group-api-v1/src/main/resources/fabric-item-group-api-v1.mixins.json
```diff
@@ -0,0 +1,14 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.itemgroup",
+  "compatibilityLevel": "JAVA_16",
```
apple502j:
```suggestion
  "compatibilityLevel": "JAVA_17",
```

fabric-item-group-api-v1/src/client/resources/fabric-item-group-api-v1.client.mixins.json
```diff
@@ -1,6 +1,6 @@
 {
   "required": true,
-  "package": "net.fabricmc.fabric.mixin.item.group.client",
+  "package": "net.fabricmc.fabric.mixin.itemgroup.client",
   "compatibilityLevel": "JAVA_16",
```
apple502j:
```suggestion
  "compatibilityLevel": "JAVA_17",
```

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
modmuss50: This seems correct to me.

modmuss50: 2nd thoughts this seems to make more sense.

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java
modmuss50: Good spot, Ill remove this. I dont think many mods will make use of this, and if they do they can easily get it from FabricItemGroupEntries.

Issue #2598, opened 2022-10-20
An Event that Triggers on a Block Place but after the world registers the block is there


ZestyBlaze:
Trying to make a new Golem for Fabric but there is no Event within the Fabric API or other libs that triggers specifically after the world registers that a block has been placed (To be run after the pumpkin or whatever the head block of the golem is)

Note: In Discussion with Tech about how it would be implemented

PR #2601, opened 2022-10-22
Add PlayerPlaceBlockEvents v1

ZestyBlaze:
Add the first draft of PlayerPlaceBlockEvents

After a discussion with Tech, I've written a draft of PlayerPlaceBlockEvents with the events `ALLOW`, `BEFORE`, `AFTER` and `CANCELLED`
I've tested it all in an external environment and it should all work well

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
@@ -1,0 +1,72 @@
+package net.fabricmc.fabric.api.event.player;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemPlacementContext;
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public class PlayerPlaceBlockEvents {
+	public static final Event<Before> BEFORE = EventFactory.createArrayBacked(Before.class,
+			(listeners) -> ((context) -> {
+				for (Before listener : listeners) {
+					ActionResult result = listener.beforeBlockPlace(context);
+					if(result != ActionResult.PASS) {
+						return result;
+					}
+				}
+				return ActionResult.PASS;
+			})
+	);
+
+	public static final Event<After> AFTER = EventFactory.createArrayBacked(After.class,
+			(listeners) -> ((world, entity, pos, state) -> {
+				for (After listener : listeners) {
+					listener.afterBlockPlace(world, entity, pos, state);
+				}
+			})
+	);
+
+	public static final Event<Cancelled> CANCELLED = EventFactory.createArrayBacked(Cancelled.class,
+			(listeners) -> ((world, entity, pos, state) -> {
+				for (Cancelled listener : listeners) {
+					listener.cancelledBlockPlace(world, entity, pos, state);
+				}
+			})
+	);
+
+	public static final Event<Allow> ALLOW = EventFactory.createArrayBacked(Allow.class,
+			(listeners) -> ((entity) -> {
+				for (Allow listener : listeners) {
+					if(!listener.allowBlockPlace(entity)) {
+						return false;
+					}
+				}
+				return true;
+			})
+	);
+
+	@FunctionalInterface
+	public interface Before {
+		ActionResult beforeBlockPlace(ItemPlacementContext context);
+	}
+
+	@FunctionalInterface
+	public interface After {
+		void afterBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
+	}
+
+	@FunctionalInterface
+	public interface Cancelled {
+		void cancelledBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
+	}
+
+	@FunctionalInterface
+	public interface Allow {
+		boolean allowBlockPlace(LivingEntity entity);
+	}
+}
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BlockItemMixin.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BlockItemMixin.java
@@ -1,0 +1,30 @@
+package net.fabricmc.fabric.mixin.event.interaction;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.item.BlockItem;
+import net.minecraft.item.ItemPlacementContext;
+import net.minecraft.item.ItemUsageContext;
+import net.minecraft.util.ActionResult;
+
+import net.fabricmc.fabric.api.event.player.PlayerPlaceBlockEvents;
+
+@Mixin(BlockItem.class)
+public class BlockItemMixin {
+	@Inject(method = "useOnBlock", at = @At("HEAD"), cancellable = true)
+	private void beforeBlockPlace(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir) {
+		if(PlayerPlaceBlockEvents.ALLOW.invoker().allowBlockPlace(context.getPlayer())) {
+			ActionResult result = PlayerPlaceBlockEvents.BEFORE.invoker().beforeBlockPlace(new ItemPlacementContext(context));
+			if (result != ActionResult.PASS) {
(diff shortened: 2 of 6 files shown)


warjort:
I don't think mixing into BlockItem.useOnBlock() is correct.
Anybody can subclass BlockItem and override that method without calling super.useOnBlock().
Even if they do, your cir.setReturnValue() is just going to return to their subclass implementation and not necessarily stop all the processing.
I think a better place to put this this event as it stands would probably be ItemStack.useOnBlock() with a check for 
Item instanceof BlockItem?

Kind of off topic: 
Your original issue is with creating golems. https://github.com/FabricMC/fabric/issues/2598
These need special handling.
e.g. see CarvedPumpkinBlock.canDispense() used in DispenserBehavior
Although that's probably a subject for a different pull request? e.g. being able register BlockPatterns that create entities in CarvedPumpkinBlock or maybe even any block?

Technici4n:
I'll check how forge does it soon (and also this PR), but in principle modders can fire the event themselves for their custom logic, and that's one of the reasons to have it as a fabric api event.

warjort:
````I'll check how forge does it soon````
Their event is effectively called from the ItemStack. They have redirected the code to a helper method.

They also call it from `EndermanEntity.tick()` and `FrostWalkerEnchantment.freezeWater()`
 where they do the `setBlockState()` directly.
This is because their event is for an entity placing a block rather than just a player.

File: fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java [hunk 1]
```diff
@@ -0,0 +1,72 @@
...
+
+	public static final Event<Cancelled> CANCELLED = EventFactory.createArrayBacked(Cancelled.class,
+			(listeners) -> ((world, entity, pos, state) -> {
+				for (Cancelled listener : listeners) {
+					listener.cancelledBlockPlace(world, entity, pos, state);
+				}
+			})
+	);
+
+	public static final Event<Allow> ALLOW = EventFactory.createArrayBacked(Allow.class,
```
Technici4n, line 42: This should be before the other events as it fires before them.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
```diff
@@ -0,0 +1,72 @@
...
+		void afterBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
+	}
+
+	@FunctionalInterface
+	public interface Cancelled {
+		void cancelledBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
+	}
+
+	@FunctionalInterface
+	public interface Allow {
```
Technici4n, line 69: Should be before the `Before` interface. Should also include more context to decide whether placement is appropriate - probably as much context as `Before` since they are fired at the same place.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
```diff
@@ -0,0 +1,72 @@
...
+
+	public static final Event<After> AFTER = EventFactory.createArrayBacked(After.class,
+			(listeners) -> ((world, entity, pos, state) -> {
+				for (After listener : listeners) {
+					listener.afterBlockPlace(world, entity, pos, state);
+				}
+			})
+	);
+
+	public static final Event<Cancelled> CANCELLED = EventFactory.createArrayBacked(Cancelled.class,
```
Technici4n, line 34: We use American spelling - that would be `canceled` with a single L.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BlockItemMixin.java
```diff
@@ -0,0 +1,30 @@
...
+import net.fabricmc.fabric.api.event.player.PlayerPlaceBlockEvents;
+
+@Mixin(BlockItem.class)
+public class BlockItemMixin {
+	@Inject(method = "useOnBlock", at = @At("HEAD"), cancellable = true)
+	private void beforeBlockPlace(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir) {
+		if(PlayerPlaceBlockEvents.ALLOW.invoker().allowBlockPlace(context.getPlayer())) {
+			ActionResult result = PlayerPlaceBlockEvents.BEFORE.invoker().beforeBlockPlace(new ItemPlacementContext(context));
+			if (result != ActionResult.PASS) {
+				PlayerPlaceBlockEvents.CANCELLED.invoker().cancelledBlockPlace(context.getWorld(), context.getPlayer(), context.getBlockPos(), context.getWorld().getBlockState(context.getBlockPos()));
```
Technici4n, line 22: The idea with `CANCELED` is that it fires when `ALLOW` rejects the event - so here it should instead be sent at the end of the mixin, when `ALLOW` returns false.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BlockMixin.java
```diff
@@ -0,0 +1,24 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(Block.class)
+public class BlockMixin {
+	@Inject(method = "onPlaced", at = @At("HEAD"))
+	private void onPlaced(World world, BlockPos pos, BlockState state, LivingEntity placer, ItemStack itemStack, CallbackInfo ci) {
+		PlayerPlaceBlockEvents.AFTER.invoker().afterBlockPlace(world, placer, pos, state);
```
Technici4n, line 22: I'm not too sure this is the optimal injection point, as it will cause the event to fire a bit early (e.g. before the player's stack is decremented). Injecting at the very end of `BlockItem#place` would be better I think (right before the `return ActionResult.success(world.isClient);`).

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
```diff
@@ -0,0 +1,72 @@
...
+						return false;
+					}
+				}
+				return true;
+			})
+	);
+
+	@FunctionalInterface
+	public interface Before {
+		ActionResult beforeBlockPlace(ItemPlacementContext context);
```
Technici4n, line 55: I wonder if this is the optimal parameter choice... An AE2 cable would for example place the AE2 multipart block, which is really not obvious from the `ItemPlacementContext`.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerPlaceBlockEvents.java
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	@FunctionalInterface
+	public interface After {
+		void afterBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
+	}
+
+	@FunctionalInterface
+	public interface Cancelled {
+		void cancelledBlockPlace(World world, LivingEntity entity, BlockPos pos, BlockState state);
```
Technici4n, line 65: The exact name matching the conventions established in `PlayerBlockBreakEvents` should be `onBlockPlaceCanceled`.

Technici4n (changes requested):
OK, I looked more into this. I knew this wasn't trivial, but I didn't think it would be that tricky. 😅

- Forge does this in a very strange way: they record all block state changes in the world while an item is being used, and then fire an event to undo the block state changes. On top of being a very invasive patch, I think this could also cause correctness issues because it's not possible to fully undo a block state change.
- I am not sure that supporting non-players is particularly useful, especially since they're such special cases (enderman and particularly frost walker). They could likely benefit from dedicated events or mixins in the mods that need them - I don't think it's a concern for this PR. However, if we decide that we want the event to apply to players, the callbacks need to take a `PlayerEntity` argument instead of a `LivingEntity` argument.
- We will need some special logic to notify the client if the server decides that the block couldn't actually be placed. I would recommend checking `InteractionEventsRouter`.
- The test mod should be modified a bit to test cancellation a bit better, and observe its side effects. For example, canceling on the server side only when trying to place a cactus block would allow testing how the client behaves in that case.

warjort:
> Forge does this in a very strange way: they record all block state changes in the world while an item is being used, and then fire an event to undo the block state changes. On top of being a very invasive patch, I think this could also cause correctness issues because it's not possible to fully undo a block state change.

My understanding is that forge supports an extra event where you can be notified of all block changes caused by a block placement, including "recursive" changes. e.g. redstone updates or fences connecting

If you veto/cancel this event it will revert all the blocks.

It does this by remembering all changes made by calls to World.setBlockState() when placing a block. 
This includes things like the original NBT of changed/removed block entities.

Technici4n:
Oh right, they also store the NBT, I had missed that. I'm not sure we should have that system - it seems quite invasive and error-prone to me.

apple502j:
Can't we inject to `ServerPlayerInteractionManager`?

Technici4n:
How do you know if a specific item will place a block though?

PR #2608, opened 2022-10-26
22w42a RegSync changes

apple502j:
While this is ABI compatible, I'd still like to call this a breaking change because old code certainly crashes. (See javadoc)

~~`EndDynamicRegistrySetupCallback` can be postponed if we need quicker fix for the bugs.~~ Actually I should probably postpone this

- Fix `DynamicRegistrySetupCallback` being called too late
- Remove unused code
- Mark impl as internal


Technici4n (commented):
Would be nice to have some tests that validate the changed behavior in case this gets changed again in an upcoming snapshot.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistrySetupCallback.java
```diff
@@ -29,13 +29,18 @@
...
+ *         RegistryEntryAddedCallback.event(biomes).register((rawId, id, object) -> {
+ *             // Do something
+ *         });
  *     });
  * });
  * }
  * </pre>
+ *
+ * <p><strong>Important Note</strong>: The passed dynamic registry manager might not
+ * contain the registry. Use {@link DynamicRegistryManager#getOptional} to
```
Technici4n: This is due to the multiple layers, right? In that case, it would be good to explain this somewhere.

Technici4n (changes requested):
Looks pretty good. One small clarification request.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistrySetupCallback.java
apple502j: yep

Technici4n (approved):
Maybe a link to a class referencing the various layers too?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Contains the Registry Sync implementation.
+ */
+@ApiStatus.Internal
+package net.fabricmc.fabric.impl.registry.sync;
```
modmuss50, line 21: Question, I wonder if putting a package-info.java marked with `@ApiStatus.Internal` in api-base in `net.fabricmc.fabric.impl` would cover all the modules? (not something for this PR)

modmuss50: Oh, I guess it wont popagate to the sub packages (based on the next files in the PR).

modmuss50 (approved):

PR #2611, opened 2022-10-27
1.19.3 resource loader changes

apple502j:
Fixes various issues with 22w42a porting and other linter yelling stuff. Should also make "This world is experimental" warning disappear for modded worlds, as mods aren't experimental.

PR #2613, opened 2022-10-28
Resource Loader: use Text for display name

apple502j:
Based on #2611 (merge that PR first!)

This refactors resource loader to use `Text` instead of `String` as the display name. This allows them to be translated. The old String based API is deprecated.


Technici4n:
This is not super urgent so I'll wait for the prerequisite PR to be merged before reviewing this one ;)

modmuss50 (approved):

PR #2615, opened 2022-10-28
Generate package-info files for all impl and mixin packages

Juuxel:
The generated files have `@ApiStatus.Internal` annotations for IDE warnings if mods try to access those classes directly.

Example output:

```java
// [License header here]

/**
 * Implementation code for fabric-loot-api-v2.
 */
@ApiStatus.Internal
package net.fabricmc.fabric.impl.loot;

import org.jetbrains.annotations.ApiStatus;
```


modmuss50 (approved):
Great idea 👍 LGTM

gradle/package-info.gradle
```diff
@@ -0,0 +1,65 @@
...
+
+	for (def sourceSet in [sourceSets.main, sourceSets.client]) {
+		sourceRoots.from sourceSet.java.srcDirs
+	}
+
+	header = rootProject.file('HEADER')
+	outputDir = file('src/generated/java')
+}
+
+sourceSets.main.java.srcDir generateImplPackageInfos.outputDir
```
modmuss50: I think this might need splitting for the main/client source sets as its not going to be correctly denoting the package-infos for client only packages. Thus partly breaking splitting the jars in a mod dev env.

Juuxel: Done without any workarounds now that #2617 has been merged

modmuss50:
Please see build failuire.

Juuxel:
Fixed, the issue was the root build.gradle adding the module sources to `javadoc` but voiding the `Buildable` metadata by converting them to plain `Set<File>`s.

modmuss50 (approved):
👍

Issue #2616, opened 2022-10-28
Multiple modules sometimes share the same package


Juuxel:
This is not correct in a world with modules (I know we don't use them, but it's still a bad practice). It also complicates things like #2615, which assume a one-to-one correspondence from a package with classes to a Fabric API module.

This happens with some pairs of v0/v1 modules:
- Networking: `impl.networking`
- Rendering: `impl.client.rendering`
- Lifecycle events: `impl.event.lifecycle`

PR #2617, opened 2022-10-29
Split all duplicated packages

Juuxel:
This consists of two parts:
<ol><li>splitting packages duplicated between v0/v1 modules (#2616)
<ul><li>here I used the package naming convention from fabric-command-api-v1 where the v0 impl packages have a v0 subpackage</ul>
<li>splitting packages duplicated between client and main in the same module
<ul><li>if a client impl/mixin package already existed, the classes were moved there or to a subpackage
<li>otherwise I followed the newer convention of <code>impl.client.ModuleName</code> (as opposed to <code>impl.ModuleName.client</code>)
</ul></ol>

This is a prerequisite for #2615 since both issues cause problems with duplicated package-info files (either in the same sources jar or in multiple compiled jars)

### DIFF
--- a/deprecated/fabric-networking-v0/src/client/java/net/fabricmc/fabric/api/network/ClientSidePacketRegistry.java
+++ b/deprecated/fabric-networking-v0/src/client/java/net/fabricmc/fabric/api/network/ClientSidePacketRegistry.java
@@ -23,7 +23,7 @@ import net.minecraft.network.Packet;
 import net.minecraft.util.Identifier;
 import net.minecraft.network.PacketByteBuf;
 
-import net.fabricmc.fabric.impl.networking.ClientSidePacketRegistryImpl;
+import net.fabricmc.fabric.impl.networking.v0.ClientSidePacketRegistryImpl;
 
 /**
  * The client-side packet registry.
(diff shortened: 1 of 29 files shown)


deprecated/fabric-networking-v0/src/client/java/net/fabricmc/fabric/api/network/ClientSidePacketRegistry.java
```diff
@@ -23,7 +23,7 @@
 import net.minecraft.util.Identifier;
 import net.minecraft.network.PacketByteBuf;
 
-import net.fabricmc.fabric.impl.networking.ClientSidePacketRegistryImpl;
+import net.fabricmc.fabric.impl.client.networking.v0.ClientSidePacketRegistryImpl;
```
modmuss50, line 26: I think this may be better as `net.fabricmc.fabric.impl.networking.v0.client`

Juuxel: My name matches the current guidelines for package naming (and neither of these passed the old checkstyle)

modmuss50: Yep, I was wrong. Thanks 👍

modmuss50 (approved):

PR #2622, opened 2022-10-31
Remove maven specific jar

modmuss50:
This has not be required for years as loom removes nested jars from mod dependencies. This means that the normal jar (that is uploaded to CF and GitHub) will now also be on maven.


modmuss50:
Merged into https://github.com/FabricMC/fabric/pull/2632

PR #2646, opened 2022-11-09
Added Client Message Events

kevinthegreat1:
Added Client Message Events including

- AllowSendChatMessage, SendChatMessage
- AllowSendCommandMessage, SendCommandMessage
- AllowReceiveChatMessage, ReceiveChatMessage
- AllowReceiveGameMessage, ReceiveGameMessage


fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.network.message.SignedMessage;
+import net.minecraft.text.Text;
+
+import net.fabricmc.fabric.api.message.v1.ClientMessageEvents;
+
+@Mixin(MessageHandler.class)
+public abstract class MessageHandlerMixin {
+
+	@Inject(method = "processChatMessageInternal", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/ChatHud;addMessage(Lnet/minecraft/text/Text;Lnet/minecraft/network/message/MessageSignatureData;Lnet/minecraft/client/gui/hud/MessageIndicator;)V", ordinal = 0), cancellable = true)
+	private void fabric_message_api_onSignedChatMessage(MessageType.Parameters params, SignedMessage message, Text decorated, GameProfile sender, boolean onlyShowSecureChat, Instant receptionTimestamp, CallbackInfoReturnable<Boolean> cir) {
```
modmuss50: No need to prefix the inject method names with `fabric_message_api_`

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,62 @@
...
+			fabric_message_api_onChatMessage(params.applyChatDecoration(filtered), message, sender, params, receptionTimestamp, cir);
+		}
+	}
+
+	@Inject(method = "method_45745", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/ChatHud;addMessage(Lnet/minecraft/text/Text;)V"), cancellable = true)
+	private void fabric_message_api_onProfilelessChatMessage(MessageType.Parameters params, Text decorated, Instant receptionTimestamp, CallbackInfoReturnable<Boolean> cir) {
+		fabric_message_api_onChatMessage(decorated, null, null, params, receptionTimestamp, cir);
+	}
+
+	private void fabric_message_api_onChatMessage(Text message, @Nullable SignedMessage signedMessage, @Nullable GameProfile sender, MessageType.Parameters params, Instant receptionTimestamp, CallbackInfoReturnable<Boolean> cir) {
```
modmuss50: Just prefix with fabric_, I think you can also put @Unique on it

fabric-message-api-v1/src/client/resources/fabric-message-api-v1.client.mixins.json
```diff
@@ -0,0 +1,13 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.message",
+  "compatibilityLevel": "JAVA_16",
+  "mixins": [
+  ],
```
modmuss50: Nit: Remove

fabric-message-api-v1/src/client/resources/fabric-message-api-v1.client.mixins.json
```diff
@@ -0,0 +1,13 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.message",
+  "compatibilityLevel": "JAVA_16",
```
modmuss50: Nit: Might as well use JAVA_17

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * prevents the message from being sent and the {@link #SEND_CHAT_MESSAGE} event
+		 * from triggering.
+		 *
+		 * @param message the message that will be sent to the server
+		 * @return {@code true} if the message should be sent, otherwise {@code false}
+		 */
+		boolean allowSendChatMessage(String message);
+	}
+
+	public interface AllowSendCommandMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * prevents the command from being sent and the {@link #SEND_COMMAND_MESSAGE} event
+		 * from triggering.
+		 *
+		 * @param message the command that will be sent to the server
+		 * @return {@code true} if the command should be sent, otherwise {@code false}
+		 */
+		boolean allowSendCommandMessage(String message);
+	}
+
+	public interface SendChatMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * Called when the client sends a chat message,
+		 * typically from a client GUI. Is not called when {@linkplain
+		 * #ALLOW_SEND_CHAT_MESSAGE chat messages are blocked}.
+		 *
+		 * @param message the message that being sent to the server
+		 */
+		void onSendChatMessage(String message);
+	}
+
+	public interface SendCommandMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * which includes whenever the player executes a vanilla command
+		 * or any other server-side commands. Is not called when {@linkplain
+		 * #ALLOW_SEND_COMMAND_MESSAGE command messages are blocked}.
+		 *
+		 * @param message the command that being sent to the server
+		 */
+		void onSendCommandMessage(String message);
+	}
+
+	public interface AllowReceiveChatMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * @param signedMessage      the signed message received from the server (may be null)
+		 * @param sender             the sender of the message (may be null)
+		 * @param params             the parameters of the message
+		 * @param receptionTimestamp the timestamp when the message was received
+		 * @return {@code true} if the message should be displayed, otherwise {@code false}
+		 */
+		boolean allowReceiveChatMessage(Text message, @Nullable SignedMessage signedMessage, @Nullable GameProfile sender, MessageType.Parameters params, Instant receptionTimestamp);
+	}
+
+	public interface AllowReceiveGameMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * {@link net.minecraft.client.network.ClientPlayerEntity#sendMessage(Text, boolean) ClientPlayerEntity.sendMessage(message, overlay)}.
+		 *
+		 * @param message the message received from the server
+		 * @param overlay whether the message will be displayed in the action bar
+		 * @return {@code true} if the message should be displayed, otherwise {@code false}
+		 */
+		boolean allowReceiveGameMessage(Text message, boolean overlay);
+	}
+
+	public interface ReceiveChatMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
...
+		 * @param message            the message received from the server
+		 * @param signedMessage      the signed message received from the server (may be null)
+		 * @param sender             the sender of the message (may be null)
+		 * @param params             the parameters of the message
+		 * @param receptionTimestamp the timestamp when the message was received
+		 */
+		void onReceiveChatMessage(Text message, @Nullable SignedMessage signedMessage, @Nullable GameProfile sender, MessageType.Parameters params, Instant receptionTimestamp);
+	}
+
+	public interface ReceiveGameMessage {
```
modmuss50: Missing @FunctionalInterface

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,256 @@
+package net.fabricmc.fabric.api.message.v1;
```
modmuss50: Missing package header, I see a few checkstyle issues as well. Please make sure this builds.

modmuss50 (changes requested):

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,283 @@
...
+import net.minecraft.network.message.SignedMessage;
+import net.minecraft.text.Text;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Contains client-side events triggered when sending or receiving messages.
+ */
+public class ClientMessageEvents {
```
apple502j: Make it final, also add private constructor

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,283 @@
...
+			if (!listener.allowReceiveGameMessage(message, overlay)) {
+				return false;
+			}
+		}
+
+		return true;
+	});
+
+	/**
+	 * An event triggered when the client received a chat message,
```
apple502j: "receives"?

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,283 @@
...
+	 * {@linkplain #ALLOW_RECEIVE_CHAT_MESSAGE chat messages are blocked}.
+	 */
+	public static final Event<ReceiveChatMessage> RECEIVE_CHAT_MESSAGE = EventFactory.createArrayBacked(ReceiveChatMessage.class, listeners -> (message, signedMessage, sender, params, receptionTimestamp) -> {
+		for (ReceiveChatMessage listener : listeners) {
+			listener.onReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp);
+		}
+	});
+
+	/**
+	 * An event triggered when the client received a chat message,
```
apple502j: Same

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,283 @@
...
+		 * @param message the message that will be sent to the server
+		 * @return {@code true} if the message should be sent, otherwise {@code false}
+		 */
+		boolean allowSendChatMessage(String message);
+	}
+
+	@FunctionalInterface
+	public interface AllowSendCommandMessage {
+		/**
+		 * An event triggered when the client is about to send a command,
```
apple502j: Called...

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/message/ClientPlayNetworkHandlerMixin.java
```diff
@@ -0,0 +1,49 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.network.ClientPlayNetworkHandler;
+
+import net.fabricmc.fabric.api.message.v1.ClientMessageEvents;
+
+@Mixin(ClientPlayNetworkHandler.class)
```
apple502j: Should set priority to make sure client commands don't get handled

apple502j (changes requested):

kevinthegreat1:
Ok, should be fixed. Also [this](https://github.com/FabricMC/fabric/pull/2618) should be ready for merge.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/message/ClientPlayNetworkHandlerMixin.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.minecraft.client.network.ClientPlayNetworkHandler;
+
+import net.fabricmc.fabric.api.message.v1.ClientMessageEvents;
+
+/**
+ * Mixin to {@link ClientPlayNetworkHandler} to listen for sending messages and commands.
+ * Priority set to 800 to inject before {@code fabric-command-api} so that this api will be called first.
+ */
+@Mixin(value = ClientPlayNetworkHandler.class, priority = 800)
```
apple502j: I mean, run it after client commands (the docs say "server-side command")

kevinthegreat1:
I think this event should capture all messages and commands sent/executed by the player, including client commands. This gives a way to override all messages and commands. I have updated the docs to reflect this behavior.

apple502j (approved):

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,301 @@
...
+		 * which is whenever the player executes a command
+		 * including client commands registered with {@code fabric-command-api}.
+		 * Returning {@code false} prevents the command from being sent and
+		 * the {@link #SEND_COMMAND_MESSAGE} event from triggering.
+		 *
+		 * <p>Also, this is called before client commands are checked and executed.
+		 * Therefore, client commands registered with {@code fabric-command-api}
+		 * will not be executed if {@code false} is returned.
+		 *
+		 * @param message the command that will be sent to the server
```
Juuxel: Does this have the initial slash? I feel that's a good thing to document somewhere

fabric-message-api-v1/src/client/resources/fabric-message-api-v1.client.mixins.json
```diff
@@ -0,0 +1,11 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.message",
+  "compatibilityLevel": "JAVA_17",
+  "client": [
+    "ClientPlayNetworkHandlerMixin", "MessageHandlerMixin"
```
Juuxel:
```suggestion
    "ClientPlayNetworkHandlerMixin",
    "MessageHandlerMixin"
```

Juuxel (changes requested):

kevinthegreat1:
I applied the suggestions.

apple502j:
@kevinthegreat1 Can you test if the PR still works with the latest prerelease?

kevinthegreat1:
@apple502j Still works in 1.19.3-rc1. I don't expect them to make changes to chat for at least a while.

kevinthegreat1:
I don't know why the build failed. Also, I can't find a rerun button in the details now. Can somebody rerun the checks?

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,305 @@
...
+	public interface SendCommandMessage {
+		/**
+		 * Called when the client sends a command,
+		 * which is whenever the player executes a command
+		 * including client commands registered with {@code fabric-command-api}.
+		 * Is not called when {@linkplain #ALLOW_SEND_COMMAND_MESSAGE command messages are blocked}.
+		 * The command string does not include a slash at the beginning.
+		 *
+		 * <p>This will also be called when executing a client command
+		 * registered with {@code fabric-command-api}.
```
ghost: This is mentioned for the second time in this javadoc.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,305 @@
...
+		 * Called when the client is about to send a command,
+		 * which is whenever the player executes a command
+		 * including client commands registered with {@code fabric-command-api}.
+		 * Returning {@code false} prevents the command from being sent and
+		 * the {@link #SEND_COMMAND_MESSAGE} event from triggering.
+		 * The command string does not include a slash at the beginning.
+		 *
+		 * <p>Also, this is called before client commands are checked and executed.
+		 * Therefore, client commands registered with {@code fabric-command-api}
+		 * will not be executed if {@code false} is returned.
```
ghost: This is mentioned for the second time in this javadoc.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
```diff
@@ -0,0 +1,305 @@
...
+
+	@FunctionalInterface
+	public interface ReceiveGameMessage {
+		/**
+		 * Called when the client receives a game message,
+		 * which is any message sent by the server. Is not called when
+		 * {@linkplain #ALLOW_RECEIVE_GAME_MESSAGE game messages are blocked}.
+		 *
+		 * <p>Overlay is whether the message will be displayed in the action bar.
+		 * Use {@link #ALLOW_RECEIVE_GAME_MESSAGE to toggle overlay}.
```
ghost:
```suggestion
		 * Use {@link #ALLOW_RECEIVE_GAME_MESSAGE} to toggle overlay.
```

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,82 @@
...
+	private void fabric_onProfilelessChatMessage(MessageType.Parameters params, Text decorated, Instant receptionTimestamp, CallbackInfoReturnable<Boolean> cir) {
+		fabric_onChatMessage(decorated, null, null, params, receptionTimestamp, cir);
+	}
+
+	@Unique
+	private void fabric_onChatMessage(Text message, @Nullable SignedMessage signedMessage, @Nullable GameProfile sender, MessageType.Parameters params, Instant receptionTimestamp, CallbackInfoReturnable<Boolean> cir) {
+		if (!ClientMessageEvents.ALLOW_RECEIVE_CHAT_MESSAGE.invoker().allowReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp)) {
+			cir.setReturnValue(false);
+			return;
+		}
```
ghost: Should log the message and set the last process time regardless of the event

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,82 @@
...
+	private void fabric_onOverlayGameMessage(Text message, boolean overlay, CallbackInfo ci) {
+		fabric_onGameMessage(message, overlay, ci);
+	}
+
+	@Inject(method = "onGameMessage", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/ChatHud;addMessage(Lnet/minecraft/text/Text;)V"), cancellable = true)
+	private void fabric_onGameMessage(Text message, boolean overlay, CallbackInfo ci) {
+		if (!ClientMessageEvents.ALLOW_RECEIVE_GAME_MESSAGE.invoker().allowReceiveGameMessage(message, overlay)) {
+			ci.cancel();
+			return;
+		}
```
ghost: Should log the message regardless of the event.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,82 @@
...
+			cir.setReturnValue(false);
+			return;
+		}
+
+		ClientMessageEvents.RECEIVE_CHAT_MESSAGE.invoker().onReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp);
+	}
+
+	@Inject(method = "onGameMessage", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/InGameHud;setOverlayMessage(Lnet/minecraft/text/Text;Z)V"), cancellable = true)
+	private void fabric_onOverlayGameMessage(Text message, boolean overlay, CallbackInfo ci) {
+		fabric_onGameMessage(message, overlay, ci);
```
ghost: Message is not logged if it's an overlay message. So also change this when you update the pr.

ghost (changes requested):
Besides the javadoc changes:
* Don't immediately set the return value when `ALLOW_RECIEVE_CHAT_MESSAGE` returns false, setting the return value ends the method, before minecraft logs the message to it's internal log for security purposes and sets the last process time stamp. Fabric should handle these if event returns false.
* First one applies to `ALLOW_RECIEVE_GAME_MESSAGE` as well.

Also `CANCELLED` events would be nice. 
Or invoke `RECEIVE` events regardless of `ALLOW` events and pass the result of the allow event to the receive event for context.

ghost (changes requested):
Besides the javadoc changes:
* Don't immediately set the return value when `ALLOW_RECIEVE_CHAT_MESSAGE` returns false, setting the return value ends the method, before minecraft logs the message to it's internal log for security purposes and sets the last process time stamp. Fabric should handle these if event returns false.
* First one applies to `ALLOW_RECIEVE_GAME_MESSAGE` as well.

Also `CANCELLED` events would be nice. 
Or invoke `RECEIVE` events regardless of `ALLOW` events and pass the result of the allow event to the receive event for context.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
kevinthegreat1: I don't see a problem with having this twice. The whole point is to have the Javadoc for the event and the interface similar, so looking at either Javadoc will give you the same information. This is also the case in the existing `ServerMessageEvents`.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
kevinthegreat1: The chat log is only used for chat reporting. If the player doesn't see the chat message, should we still add it to the log? I think this needs more discussion.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
ghost: I didn't mean the event. I was referring to line 225, which says that the event includes client commands. Then it is repeated here, just phrased differently.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
ghost: I thought logging every possible message was minecraft's default behavior but there isn't a case where message isn't sent. So it only logs the sent messages. You are right, I don't think it's neccessary.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
ghost: Same here, referring to line 193.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientMessageEvents.java
kevinthegreat1: Sorry, I misunderstood you. I removed the duplicate.

ghost (approved):

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.network.ClientPlayNetworkHandler;
+
+import net.fabricmc.fabric.api.client.message.v1.ClientSendMessageEvents;
+
+/**
+ * Mixin to {@link ClientPlayNetworkHandler} to listen for sending messages and commands.
+ * Priority set to 800 to inject before {@code fabric-command-api} so that this api will be called first.
```
Technici4n, line 31: Did you test this specifically? Might need a higher priority cause we want our `HEAD` to be applied last wrt. the other `HEAD` injects... needs to be confirmed by testing.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/MessageHandlerMixin.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	@Inject(method = "onGameMessage", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/InGameHud;setOverlayMessage(Lnet/minecraft/text/Text;Z)V"), cancellable = true)
+	private void fabric_onOverlayGameMessage(Text message, boolean overlay, CallbackInfo ci) {
+		fabric_onGameMessage(message, overlay, ci);
+	}
+
+	@Inject(method = "onGameMessage", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/hud/ChatHud;addMessage(Lnet/minecraft/text/Text;)V"), cancellable = true)
+	private void fabric_onGameMessage(Text message, boolean overlay, CallbackInfo ci) {
```
Technici4n: You should be able to use a `JUMP` injection point to inject before the `if (overlay)`. Not sure if it's worth it though. Otherwise you can specify two selectors in the `at` to only have a single method.

Technici4n (changes requested):
Would it be useful to have a way to modify sent or received messages? The notification events aren't doing much atm.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
kevinthegreat1: I'm not sure what you mean. What is wrong with having this mixin applied first?

kevinthegreat1:
Modifying a received chat message seems to be difficult. The signature and narration would be wrong if we change the message. Is it ok if we only changed the displayed message and leave the narration? Also, the player might be seeing different messages than the ones in the log which causes all sorts of issues with chat reporting.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
Technici4n: If my understanding is correct:

```
after first mixin:

<-- method start
<-- HEAD inject at priority 800
<-- method body

after second mixin

<-- method start
<-- HEAD inject at priority 1000
<-- rest of the method (i.e. previous inject and then body)
```

So to be sure I recommend checking the generated classes using `-Dmixin.debug.export=true`.

Technici4n:
> Also, the player might be seeing different messages than the ones in the log which causes all sorts of issues with chat reporting.

Ah right, that's a good point.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
kevinthegreat1: The HEAD inject with priority 800 (allowSendCommandMessage) is executed first.
Screenshot 2022-12-25 at 11 57 09

Technici4n: All good then.

kevinthegreat1:
I have added the ability to modify messages. Currently, modified received chat messages are marked with the `Modified by Server` indicator and modified received game messages are not changed and mark with `Game Message` indicator. This is not really accurate and to make it accurate we will have to add our own indicator like `Chat/Game Message Modified by client`.

kevinthegreat1:
I have not tested this on multiplayer servers. How do I build a jar for the test mod?

Technici4n:
> I have not tested this on multiplayer servers. How do I build a jar for the test mod?

`gradlew remapTestmodJar` I think?

Technici4n (changes requested):
The modification events should have modify in the name I think. So `GAME` -> `MODIFY_GAME`, `Game` -> `ModifyGame`, `onReceiveGameMessage` -> `modifyReceivedGameMessage` or something like that.

Do we need plain notification events that get called with the modified message? (Don't think so).

apple502j (changes requested):
Modification should internally cancel the message (like when Social Interactions hides the message). Otherwise this can mess up message chain.

Technici4n:
In that case, maybe it's best to leave out modification entirely (for now at least) ?

kevinthegreat1:
Currently, it should keep the original message in the chain. When reporting, you should see the unmodified message.

kevinthegreat1:
> The modification events should have modify in the name I think. So `GAME` -> `MODIFY_GAME`, `Game` -> `ModifyGame`, `onReceiveGameMessage` -> `modifyReceivedGameMessage` or something like that.
> 
> Do we need plain notification events that get called with the modified message? (Don't think so).

The main problem is that people won't know which one to use if they just want to listen for messages. Thats why I didn't add `MODIFY` in front.

kevinthegreat1:
Works as intended in multiplayer. Modified secure chat messages are marked with modified indicator when modified on receive. Already modified messages or game messages are not changed. Chat reports use original unmodified and signed messages.

apple502j (changes requested):
Please, either remove modification, or cancel the message. Having a "modified" message show up in hud while being logged in chat log as unmodified is very bad.

(by "cancelling" I mean early-returning `false` in `processChatMessageInternal`)

kevinthegreat1:
Just to be clear: the unmodified and original message is logged in the chat log, not the modified message.

Technici4n:
Yeah, that's quite problematic indeed. I think modification should be removed entirely. 😓

kevinthegreat1:
I think if we remove it, mod devs would use less thoughtful ways to achieve this and cause more trouble for users.

kevinthegreat1:
I don't see a big problem with the current implementation. Players see the original unmodified message when hovering over the message and when reporting.

apple502j:
@kevinthegreat1 I believe the best solution is, as I said, to cancel and resend. If you're busy I can open a PR with that. What do you think?

kevinthegreat1:
I think the best solution is what we have currently. There will be gaps in the log and lack of context when reporting if we cancel the modified messages. Players see the original message when reporting. If you really don't think this is good I can update the pr.

Technici4n (approved):

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientReceiveMessageEvents.java
```diff
@@ -0,0 +1,229 @@
...
+	 * An event triggered when the client receives a chat message,
+	 * which is any message sent by a player. Is not called when
+	 * {@linkplain #ALLOW_CHAT chat messages are blocked}.
+	 * Mods can use this to modify the message.
+	 *
+	 * <p>Secure chat messages modified with this event will be marked with modified indicator.
+	 * Messages already with the modified indicator or game messages are not affected.
+	 * The unmodified message is used in chat reporting.
+	 */
+	public static final Event<Chat> CHAT = EventFactory.createArrayBacked(Chat.class, listeners -> (message, signedMessage, sender, params, receptionTimestamp) -> {
```
Technici4n, line 107: Should probably be `MODIFY_CHAT` since the message can be modified. (and the function name would be `modifyReceivedChatMessage`).

Technici4n (changes requested):

apple502j (changes requested):
I think this breaks custom message types with different narrations.

kevinthegreat1:
It seems like the best solutions are always narrate the original message or don't narrate modified messages. Do you have any suggestions?

Technici4n:
Hmmm, both options are not ideal. There's also the option to remove modification.

kevinthegreat1:
Is it fine to just leave it there and maybe add a warning to mods? The modification feature doesn't affect anything when it's not used.

apple502j:
Modification is a security-sensitive part of the API, and I still believe the only way to do it is to cancel and send as system message.

I don't remember the specific details on how chat reporting works in 1.19.3, but it should be still true that clients send a message acknowledgment to the server, which constitute part of "evidences" signed and sent in a chat report. If the user sees a message modified by the client, the user didn't acknowledge the original message.

Sure, we could do what others are doing (see QuiltMC/quilt-standard-libraries#250) and implement a big chat API that is capable of doing everything. It might have made sense in 1.18, but now that the system is more complicated, exposing a bigger API surface is probably more of a maintenance burden than a benefit, especially here where an update is released within hours of snapshot release. There are not that much uses in client-side modifications as far as I can see. Plus, when an API is too costly to maintain here but is still useful, there's still an option to release it as a third-party API (like with Terrablender).

If we're keeping the modification functionality, it should just be a wrapper of cancellation event; if a message is modified, cancel and add ourselves. (Also, should this modify the original message or the decorated message? I prefer the former one, and perhaps using `Pair` of `Text` and `MessageType` so that narration works properly.)

kevinthegreat1:
Yes. I agree with the potential security issues related to users not actually seeing the message and the maintenance cost. I did not think of this. I am removing modification for received chat messages. There is now a message in the docs directing mods to use the allow event and add modified messages to the hud manually.

Looking at the quilt implementation, it seems like they're not dealing with all the char reporting. They're modifying the packet when it gets received. It will probably get handled the same way as chat messages modified by the server.

fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientReceiveMessageEvents.java
```diff
@@ -0,0 +1,227 @@
...
+		for (Chat listener : listeners) {
+			listener.onReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp);
+		}
+	});
+
+	/**
+	 * An event triggered when the client receives a chat message,
+	 * which is any message sent by the server. Is not called when
+	 * {@linkplain #ALLOW_CHAT chat messages are blocked}.
+	 * Mods can use this to listen or modify the message.
```
Technici4n: This needs to be split into separate `MODIFY_GAME` and `GAME` events for modification and notification respectively.

kevinthegreat1: Should I split send message and command events to notify and modify too?

Technici4n: Yeah.

kevinthegreat1:
There are separate events for listening and modifying now.

apple502j (approved):

modmuss50 (approved):

Technici4n (approved):
Works well, thank you!

PR #2647, opened 2022-11-09
22w45a

modmuss50:
Item Group changes are the most impactful for mods. Basic usage bellow (Full docs will come once into pre-releases)

```java
 private static final ItemGroup ITEM_GROUP = FabricItemGroup.builder(new Identifier(MOD_ID, "test_group"))
   .icon(() -> new ItemStack(Items.DIAMOND))
   .entries((enabledFeatures, entries, operatorEnabled) -> {
      entries.add(TEST_ITEM);
   })
   .build();
```


fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupMixin.java
```diff
@@ -18,42 +18,48 @@
...
 
 	@Shadow(aliases = "field_40860")
 	private ItemStackSet searchTabStacks;
 
+	@Unique
+	private int page = -1;
+
+	@Unique
+	@Nullable
+	private UUID fallbackUUIDCache;
```
Jab125: Make it more apparent what mod the mixin came from
```suggestion
	@Unique
	private int fabric_page = -1;

	@Unique
	@Nullable
	private UUID fabric_fallbackUUIDCache;
```

Jab125 (changes requested):

fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/mixin/itemgroup/ItemGroupMixin.java
Jab125: and all references need changing as well

oliviathevampire (approved):
looks great!

PR #2650, opened 2022-11-11
Biome API: modify Nether biomes a bit earlier

apple502j:
PlayerManager now keeps its own copy of DRM, so modify nether biomes just before that. This mixin now applies just before biome modification kicks in.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MinecraftServerMixin.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MinecraftServerMixin.java
@@ -16,33 +16,37 @@
 
 package net.fabricmc.fabric.mixin.biome;
 
+import java.net.Proxy;
+
+import com.mojang.datafixers.DataFixer;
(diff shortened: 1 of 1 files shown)


fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/MinecraftServerMixin.java
```diff
@@ -16,33 +16,37 @@
...
 import net.minecraft.util.registry.DynamicRegistryManager;
 import net.minecraft.util.registry.Registry;
+import net.minecraft.util.registry.RegistryKeys;
 import net.minecraft.world.dimension.DimensionOptions;
+import net.minecraft.world.level.storage.LevelStorage;
 
 import net.fabricmc.fabric.impl.biome.NetherBiomeData;
 
-@Mixin(MinecraftServer.class)
+@Mixin(value = MinecraftServer.class, priority = 990)
```
modmuss50, line 42: A comment explaining why the priority is needed might be a good idea? Not a major issue tho.

modmuss50 (approved):

PR #2658, opened 2022-11-14
Resource Conditions: add feature_enabled

apple502j:
Adds a new resource condition to check whether a feature is enabled.

Mojang: please make FeatureFlag a record

Tested, works.


modmuss50 (approved):

Juuxel (approved):

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
```diff
@@ -204,4 +209,41 @@ public static boolean tagsPopulatedMatch(JsonObject object, RegistryKey<? extend
 
 		return true;
 	}
+
+	public static ConditionJsonProvider featureEnabled(Identifier id, final FeatureFlag feature) {
```
Technici4n: You can simplify the id retrieval code as follows:
```java
			// Uses parameter: FeatureFlag... flags
			FeatureManager manager = FeatureFlags.FEATURE_MANAGER;
			FeatureSet set = manager.featureSetOf(flags);
			List<Identifier> identifiers = manager.toId(set).stream().sorted().toList();
```
This checks for the feature universe, and doesn't require the accessor mixin.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -146,6 +148,17 @@ public static <T> ConditionJsonProvider tagsPopulated(TagKey<T>... tags) {
...
 	}
 
+	/**
+	 * Creates a condition that returns true if the passed feature is enabled.
+	 * @param feature the feature to check for
+	 *
+	 * @apiNote This condition's ID is {@code fabric:feature_enabled}, and takes one property:
+	 * {@code feature}, which is the ID of the feature flag to check.
+	 */
+	public static ConditionJsonProvider featureEnabled(FeatureFlag feature) {
```
Technici4n: I would like to see this condition renamed to `features_enabled` and accept multiple features (like the others above).

EDIT: the function name should of course be `featuresEnabled`

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
```diff
@@ -204,4 +209,41 @@ public static boolean tagsPopulatedMatch(JsonObject object, RegistryKey<? extend
...
+			@Override
+			public void writeParameters(JsonObject object) {
+				object.addProperty("feature", feature.toString());
+			}
+		};
+	}
+
+	public static FeatureSet currentFeature = FeatureFlags.DEFAULT_ENABLED_FEATURES;
+
+	public static boolean featureEnabledMatch(JsonObject object) {
```
Technici4n: Similarly, you can convert a list of identifiers to a `FeatureSet` as follows:

```java
			manager.featureSetOf(identifiers, id -> {
				throw new JsonParseException("Unknown feature flag: " + id);
			});
```

This will allow removing the other mixin.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
```diff
@@ -204,4 +209,41 @@ public static boolean tagsPopulatedMatch(JsonObject object, RegistryKey<? extend
...
+			}
+
+			@Override
+			public void writeParameters(JsonObject object) {
+				object.addProperty("feature", feature.toString());
+			}
+		};
+	}
+
+	public static FeatureSet currentFeature = FeatureFlags.DEFAULT_ENABLED_FEATURES;
```
Technici4n: I would prefer a `ThreadLocal` that is cleared similarly to the `ThreadLocal` for tags above, for thread-safety and consistent behavior.

Technici4n (changes requested):

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
apple502j: Fair enough. Was worried it is invoked in other threads but it seems to be fine.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/mixin/resource/conditions/DataPackContentsMixin.java
```diff
@@ -39,4 +46,12 @@ public class DataPackContentsMixin {
 	public void hookRefresh(DynamicRegistryManager dynamicRegistryManager, CallbackInfo ci) {
 		ResourceConditionsImpl.clearTags();
```
Technici4n, line 47: Could you also clear the feature set ThreadLocal ? (And move the mixin's javadoc to the `hookRefresh` function.

apple502j: There should be no need. Feature flags are immutable during server lifetime, and server loading always starts with reload method.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java
```diff
@@ -204,4 +208,37 @@ public static boolean tagsPopulatedMatch(JsonObject object, RegistryKey<? extend
 
 		return true;
 	}
+
+	public static ConditionJsonProvider featuresEnabled(Identifier id, final FeatureFlag... features) {
+		final Set<Identifier> ids = FeatureFlags.FEATURE_MANAGER.toId(FeatureFlags.FEATURE_MANAGER.featureSetOf(features));
```
Technici4n: This should be sorted to ensure that the order in which they're generated is consistent.

Technici4n (approved):

modmuss50 (approved):

PR #2667, opened 2022-11-19
Entity Selector Option API

apple502j:
Adds an entity selector option API.

### Is this a port from (insert another API library)?
No. The underlying design might be similar (after all, this is just a wrapper method), but the code is written entirely by me. This also adds an easy way to make a non-repeatable selector.

### Is this tested?
Yes, actually this is tested using GameTest!

### DIFF
--- a/fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
+++ b/fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
@@ -30,12 +30,16 @@ public interface FabricEntitySelectorReader {
 	 * @param key the key of the flag
 	 * @param value the value of the flag
 	 */
-	void setCustomFlag(Identifier key, boolean value);
+	default void setCustomFlag(Identifier key, boolean value) {
+		throw new UnsupportedOperationException("Not implemented");
+	}
 
 	/**
 	 * Gets the value of the flag.
 	 * @param key the key of the flag
 	 * @return the value, or {@code false} if the flag is not set
 	 */
-	boolean getCustomFlag(Identifier key);
+	default boolean getCustomFlag(Identifier key) {
+		throw new UnsupportedOperationException("Not implemented");
+	}
 }
--- a/fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
+++ b/fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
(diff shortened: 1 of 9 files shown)


fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
```diff
@@ -0,0 +1,45 @@
...
+ * or non-repeatable entity selector option.
+ */
+public interface FabricEntitySelectorReader {
+	/**
+	 * Sets a flag.
+	 * @param key the key of the flag
+	 * @param value the value of the flag
+	 */
+	default void setCustomFlag(Identifier key, boolean value) {
+		throw new UnsupportedOperationException("Not implemented");
```
Technici4n:
```suggestion
		throw new UnsupportedOperationException("Implemented via mixin");
```

fabric-command-api-v2/src/main/java/net/fabricmc/fabric/api/command/v2/FabricEntitySelectorReader.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	/**
+	 * Gets the value of the flag.
+	 * @param key the key of the flag
+	 * @return the value, or {@code false} if the flag is not set
+	 */
+	default boolean getCustomFlag(Identifier key) {
+		throw new UnsupportedOperationException("Not implemented");
```
Technici4n:
```suggestion
		throw new UnsupportedOperationException("Implemented via mixin");
```

fabric-command-api-v2/src/main/java/net/fabricmc/fabric/mixin/command/EntitySelectorOptionsAccessor.java
```diff
@@ -0,0 +1,33 @@
...
+import org.spongepowered.asm.mixin.gen.Invoker;
+
+import net.minecraft.command.EntitySelectorOptions;
+import net.minecraft.command.EntitySelectorReader;
+import net.minecraft.text.Text;
+
+@Mixin(EntitySelectorOptions.class)
+public interface EntitySelectorOptionsAccessor {
+	@Invoker
+	static void callPutOption(String id, EntitySelectorOptions.SelectorHandler handler, Predicate<EntitySelectorReader> condition, Text description) {
```
Technici4n, line 31: Funny how the two classic types of options are called "call options" and "put options" in finance. :smile:

Technici4n (changes requested):

Technici4n (approved):

deirn:
I noticed that you can use Identifier as the key, but it needs to be quoted for it to work. Should we do that instead of replacing it with underscores? Might need to hack around the autocomplete suggestion though.

apple502j:
@deirn I don't think that's very useful.

modmuss50 (approved):

PR #2668, opened 2022-11-19
Load modded translations on the server

SquidDev:
This is an alternative version of #1501. This aims to solve the same problems, but in a slightly more stripped down way.

## Motivation
Vanilla Minecraft initially loads translations through `ClassLoader.getResourceAsStream`, rather than using the currently loaded resource packs. While these translations are later replaced on the client, this does not happen on the server, meaning modded language keys are never translated.

This is especially annoying when in the server console. For instance, even on a server with _just_ fabric-resource-loader-v0-testmod loaded, it's very easy to start seeing untranslated strings:

```
> /datapack list
There are 2 data packs enabled: [vanilla (built-in)], [Fabric Mods (pack.source.fabricmod)]
There are 1 data packs available: [fabric-resource-loader-v0-testmod/test2 (pack.source.builtinMod)]
```

## Changes
This PR loops through all currently loaded mods, reads `/assets/lang/{mod_id}/en_us.json` and adds its entries to the currently loaded language map.

Unlike the vanilla language loader, we use a `HashMap`, meaning that duplicate language keys are allowed. Mods are iterated in load-order, with later mods overriding earlier ones. I don't think this behaviour should be relied on, just noting it for completeness.

## Scope
This PR is intentionally limited in scope: its aiming to fix a flaw in Fabric API and mimic vanilla behaviour, rather than adding a more fleshed-out language system. Support for server-side resource packs or alternative languages are (IMO) better served by mods like [Server Translations](https://github.com/NucleoidMC/Server-Translations).

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
@@ -53,7 +53,7 @@ class LanguageMixin {
 		Map<String, String> map = new HashMap<>(cir.buildOrThrow());
 
 		for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
-			loadModLanguage(mod, map::put);
+			if (!mod.getMetadata().getType().equals("builtin")) loadModLanguage(mod, map::put);
 		}
 
 		return ImmutableMap.copyOf(map);
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.resource.loader;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.JsonParseException;
+import org.slf4j.Logger;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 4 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
```diff
@@ -0,0 +1,77 @@
...
+
+	@Shadow
+	@Final
+	public static String DEFAULT_LANGUAGE;
+
+	@Redirect(method = "create", at = @At(value = "INVOKE", target = "Lcom/google/common/collect/ImmutableMap$Builder;build()Lcom/google/common/collect/ImmutableMap;"))
+	private static ImmutableMap<String, String> create(ImmutableMap.Builder<String, String> cir) {
+		Map<String, String> map = new HashMap<>(cir.buildOrThrow());
+
+		for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
```
modmuss50, line 55: Check that the mod type is not "builtin", this will prevent it from trying to load the languages from loader, minecraft and java. This can be checked by:

`mod.getMetadata().getType().equals("builtin")`

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/LanguageMixin.java
```diff
@@ -0,0 +1,77 @@
...
+class LanguageMixin {
+	@Shadow
+	@Final
+	private static Logger LOGGER;
+
+	@Shadow
+	@Final
+	public static String DEFAULT_LANGUAGE;
+
+	@Redirect(method = "create", at = @At(value = "INVOKE", target = "Lcom/google/common/collect/ImmutableMap$Builder;build()Lcom/google/common/collect/ImmutableMap;"))
```
modmuss50, line 51: This isnt wrong, but it seems like it could be done with an inject before the builder.build(), and capture the BiConsumer local.

SquidDev: I definitely agree that the `@Redirect` (and double copy) is a little nasty.

The problem with using the existing `BiConsumer` is that `ImmutableMap.Builder.put` doesn't allow duplicates. Some mods overwrite others' language keys: hence the need to convert to a `HashMap` and then back to an `ImmutableMap`

modmuss50: Ah, that makes sense. Good spot 👍

modmuss50 (approved):
Looks good, nice and simple 👍 

I cannot see a problem with running this on the client as well.

Juuxel (approved):

thecatcore (approved):

SquidDev:
Amazing, thank you all!

fabric-resource-loader-v0/src/testmod/java/net/fabricmc/fabric/test/resource/loader/LanguageTestMod.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.fabricmc.api.ModInitializer;
+
+public class LanguageTestMod implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		testTranslationLoaded();
+	}
+
+	private static void testTranslationLoaded() {
```
modmuss50, line 29: This is crashing on the client, I think im going to make the testmod and mixin only apply on the server.

We really really need some client test coverage on CI.

SquidDev: Oh, apologies! I thought I'd tested on the client, but clearly not got that bit right.

Thank you for fixing it :).

PR #2659, opened 2022-11-14
Fix #2657: Transfer API edge case with bad isValid overrides

Technici4n:
Pretty self-explanatory.

PR #2669, opened 2022-11-19
Fix shulker boxes accepting other shulkers when using a directionless `InventoryStorage`

SquidDev:
This is a bug caused by vanilla being weird and mods doing Silly Things(TM), so happy for it to be closed if it's out of scope.

## Background
CC: Tweaked/CC: Restitched add programmable computers to Minecraft, allowing you to write code in game. One of the features this mod provides is the ability to interact with inventories with user-written programs.

When a computer interacts with an inventory, it's useful for it to have a "complete" view of the whole inventory, rather than a per-side view. In order to achieve this, we attempt to obtain a directionless view of an inventory with `InventoryStorage.of(blockEntity, null)` before falling back to the directional block lookup API[^1].

## The Bug
When `InventoryStorge` is not given a direction, it only checks validity of inserts using `Inventory.isValid` and _not_ `SidedInventory.canInsert`.

This is generally the correct thing to do (while `SidedInventory.canInsert` does accept a `null` side, its semantics are unclear[^2] and so I think best avoided). However, shulker boxes only override `canInsert` and not `isValid`, meaning a directionless `InventoryStorage` allows moving shulker boxes inside other shulker boxes!

This patch updates shulkers to also implement `isValid`, preventing this behaviour.

[^1]: Maybe the solution here is just to add a directionless version of the block lookup API, but that feels like a much harder task :D.

[^2]: For instance, composters don't accept inserts on the `null` side.


modmuss50:
Do you know of any way to trigger this issue in vanilla/is there an MC bug for this? We want to be careful not to fix any bugs like this in vanilla that people may depend on.

Technici4n (changes requested):
I'm fine with patching it in the transfer API, but I agree that patching it like this would risk a behavior change.

You should be able to patch it in `InventorySlotWrapper` with an instanceof, similarly to #2659.

Nice test BTW. :+1:

SquidDev:
> Do you know of any way to trigger this issue in vanilla/is there an MC bug for this? 

I don't _think_ so, but definitely agree with your concerns. Will follow Technici4n's advice and move this to the slot wrapper.

Thanks both!

Technici4n (approved):

modmuss50 (approved):

PR #2670, opened 2022-11-19
API for brewing recipes which use Ingredients instead of Items

Daomephsta:
`BrewingRecipeRegistry` uses Ingredients internally, but the registration methods only support items. This PR adds a new API class with counterparts that accept Ingredients instead of Items, allowing tags to be used.

This does not include an Ingredient counterpart to `net.minecraft.recipe.BrewingRecipeRegistry.registerPotionType(Item)`.
The backend code does support Ingredients, but I'm unsure if it's useful to have multiple items that represent a single potion type (e.g. splash, regular, lingering)


fabric-content-registries-v0/src/testmod/resources/fabric-content-registries-v0-testmod.mixins.json
```diff
@@ -0,0 +1,11 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.test.mixin.content.registry",
+  "compatibilityLevel": "JAVA_16",
```
kevinthegreat1: Let's use java 17
```suggestion
  "compatibilityLevel": "JAVA_17",
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FabricBrewingRecipeRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Counterpart of {@link BrewingRecipeRegistry} with methods that allow adding recipes which use Ingredients instead of Items.
+ */
+public final class FabricBrewingRecipeRegistry {
+	private FabricBrewingRecipeRegistry() {
+	}
+
+	/**
+	 * Register a recipe for brewing one potion type into another (e.g. regular to splash).
+	 * Only one registration is necessary for all potions of the input type to be brewable into the output type.
```
kevinthegreat1: I feel like there should be documentation on how to register the new potion item (call registerPotionType).
```suggestion
	 * Only one registration is necessary for all potions of the input type to be brewable into the output type.
	 * A new potion type should be registered with {@link BrewingRecipeRegistry#registerPotionType(Item)} before calling this.
```

fabric-content-registries-v0/src/testmod/resources/fabric-content-registries-v0-testmod.mixins.json
Daomephsta: This does nothing. The mixin registered by this config does not require any Mixin features that require Java 17.

> As new versions of Java are released, new features are added to Mixin to support those new language features. The compatibilityLevel property can be used to elevate Mixin's operating level to the required level for the feature set you wish to use.
> ....
> compatibilityLevel should always be set to the highest level required by your mixins.

<https://github.com/SpongePowered/Mixin/wiki/PSA-Forward-Compatibility-Features-in-Mixin#2-mixin-config-property-compatibilitylevel>

apple502j: I believe the contribution guide suggests using the latest version available in e.g. loader version, mixin java version, etc.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/BrewingRecipeRegistryAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import net.minecraft.item.Item;
+import net.minecraft.potion.Potion;
+import net.minecraft.recipe.BrewingRecipeRegistry;
+
+@Mixin(BrewingRecipeRegistry.class)
+public interface BrewingRecipeRegistryAccessor {
```
Technici4n: Since there's already an AW file in the module, you can switch that accessor mixin for an AW.

Daomephsta: Is there a contribution guideline for this that I've missed?
Otherwise we're just getting into the access wideners vs accessor mixins debate.

Daomephsta: Is there a contribution guideline for this that I've missed?

fabric-content-registries-v0/src/testmod/resources/fabric-content-registries-v0-testmod.mixins.json
Daomephsta: Can't find anything in the contribution guide about that

modmuss50: No, doesnt look like its specificially mentioned, we should add it. It's similar to the bit about the loader version "is the version used at the time the module is added." In relality it doesnt really matter, but might as well do it for new modules.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/BrewingRecipeRegistryAccessor.java
Technici4n: AWs are cleaner than mixins, however they require the module to have its own jar, which makes the build slower. But if there's already an AW for other reasons then it's better to use the AW (less class spam). This isn't in the contribution guidelines though I think, maybe they should be amended?

Technici4n (approved):
LGTM

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FabricBrewingRecipeRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Register a recipe for brewing one potion type into another (e.g. regular to splash).
+	 * Only one recipe is necessary for all potions of the input type to be brewable into the output type using the ingredient.
+	 * Use {@link BrewingRecipeRegistry#registerPotionType(Item)} to register new potion types.
+	 * @param input the input potion type (e.g. regular potion)
+	 * @param ingredient the required ingredient (e.g. gunpowder)
+	 * @param output the output type  (e.g. splash potion)
+	 * @see BrewingRecipeRegistry#registerItemRecipe(Item, Item, Item)
+	 */
+	public static void registerItemRecipe(PotionItem input, Ingredient ingredient, PotionItem output) {
```
modmuss50, line 43: The vanilla method takes an 'Item' and not a 'PotionItem' as a pram and then validates that it is a `PotionItem` I think it may be best to mimick this behaviour?

All of the fields in vanilla's `Items` class as an `Item` so you would need to cast it to use it in this method reducing/removing the benefit of using the more specific type.

`BrewingRecipeRegistry.Recipe` seems to allow for any `Item` type.

Daomephsta: Requiring a cast for vanilla potions isn't clean, but neither is a method that declares a less specific type but actually requires a more specific type. My original thought was it's better to detect an incorrect type at compile time or with IDE error checking, rather than at runtime.

> `BrewingRecipeRegistry.Recipe` seems to allow for any `Item` type.

If you're proposing allowing for brewing any item into any item, that looks like it would only require small alterations to `net.minecraft.block.entity.BrewingStandBlockEntity#isValid` & `net.minecraft.screen.BrewingStandScreenHandler.PotionSlot#matches`.

Technici4n: modmuss was mentioning that the natural extension of the vanilla methods would be to do the instanceof check at the beginning of the method

Technici4n: > neither is a method that declares a less specific type but actually requires a more specific type

I agree that both options are suboptimal, but here it's obvious from looking at the method's body that you need a potion item (could also be said explicitly in the javadoc). In doubt we generally mimic mojang's decisions even if they're not optimal.

modmuss50: On 2nd thoughts requiring the cast might be better, it makes it ovioubs to the modder that this needs to be a PotionItem. Either way its going to crash, it might be less likely to crash if the modder has think about it and add the cast themselves.

However as this method is a version of the vanilla on matching it may be better? Im happy with either tbh its a small nit pick.

> If you're proposing allowing for brewing any item into any item
Not really was more of a question to see if this a was a good idea or not? Seems like its out of scope for this PR.

However looking at `BrewingStandScreenHandler.PotionSlot#matches` do we not need this to do an instance of check as well?

Technici4n: Either is fine for me as well, your choice Daomephsta.

modmuss50: Lets stick with what is here 👍

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FabricBrewingRecipeRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+	private FabricBrewingRecipeRegistry() {
+	}
+
+	/**
+	 * Register a recipe for brewing one potion type into another (e.g. regular to splash).
+	 * Only one recipe is necessary for all potions of the input type to be brewable into the output type using the ingredient.
+	 * Use {@link BrewingRecipeRegistry#registerPotionType(Item)} to register new potion types.
+	 * @param input the input potion type (e.g. regular potion)
+	 * @param ingredient the required ingredient (e.g. gunpowder)
+	 * @param output the output type  (e.g. splash potion)
```
Juuxel:
```suggestion
	 * @param output the output type (e.g. splash potion)
```

Juuxel (approved):

PR #2674, opened 2022-11-20
Fix crash in 1.19.3 stack remainder test port

apple502j:


modmuss50 (approved):
Bad merge from me, thanks 👍

PR #2675, opened 2022-11-20
Add ItemTooltipDataCallback event

ghost:
Added an event that is fired when `ItemStack#getTooltipData` is called. After the event, if possible data list is wrapped with a `BundledTooltipData` which is later converted to `BundledTooltipComponentImpl` and rendered.

### DIFF
--- a/fabric-item-api-v1/build.gradle
+++ b/fabric-item-api-v1/build.gradle
@@ -2,7 +2,8 @@ archivesBaseName = "fabric-item-api-v1"
 version = getSubprojectVersion(project)
 
 moduleDependencies(project, [
-		'fabric-api-base'
+		'fabric-api-base',
+		'fabric-rendering-v1'
 ])
 
 testDependencies(project, [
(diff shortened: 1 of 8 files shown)


Technici4n:
Hi, would you mind explaining what exactly this adds, and for which use cases?

ghost:
Minecraft's tooltip data implementation is pretty strict, it only allows one tooltip data and getter method must be overridden to return a custom data. Through this event mods can provide custom tooltip data's to be rendered in the stack's tooltip however they like. 

Mod dependent example: A mod has a fluid container implementation for it's items through an interface and wants to render the contained fluid, so it has to override the same method for each item unless they extend the class. With this event they can simply add their data with a instance of check.

Mod independent example: A mod can add weights to all items, like Light and Heavy. And to fancily display this on the items, it can listen to this event and render a feather or an anvil based on items weight.

These might be uncommon scenarios, and may be this event just doesn't fit inside fabric's scope, but it's just an idea of mine.

ghost:
Here is the test. 
2 spaces between components might be more than needed.

Technici4n:
Thanks for the explanation, I think this feature is in scope. I will take some time to think about how exactly it should be implemented, but I think an event is a good approach. :)

ghost:
Removed the spacing between components to match vanilla, though they are still far apart from each other.

<details>
  <summary>New Preview</summary>
   
</details>

PR #2678, opened 2022-11-20
Add automated client smoke tests.

modmuss50:
Currently we have a number of automated tests that run on a server (auto test server and game tests), but we have no automated testing of the client. This PR is a proof of concept showing how we might be able to run the full client within github actions. This also runs the game with intermediary names, the same as a "production" client.

This PR is currently able to load the client, create a new creative world and gracefully save and exit. This is done by automating the UI. Screenshots are also generated and uploaded after the test completes.

I uploaded a short video showing the tests running on my local machine: https://www.youtube.com/watch?v=CdfCJUZNgo8

This isnt designed to be used by other mods and is only included in Fabric API's test mods.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -351,9 +351,12 @@ loom {
 		}
 		autoTestServer {
 			inherit testmodServer
-
 			name "Auto Test Server"
-
+			vmArg "-Dfabric.autoTest"
+		}
+		autoTestClient {
+			inherit testmodClient
+			name "Auto Test Client"
 			vmArg "-Dfabric.autoTest"
 		}
 	}
--- a/build.gradle
+++ b/build.gradle
@@ -378,7 +378,7 @@ dependencies {
 
 import net.fabricmc.loom.util.OperatingSystem
 
-// This is very far beyond loom's API if you copy this, your on your own.
+// This is very far beyond loom's API if you copy this, you're on your own.
 task runProductionAutoTestClient(type: JavaExec, dependsOn: [remapJar, remapTestmodJar]) {
 	classpath.from configurations.productionRuntime
 	mainClass = "net.fabricmc.loader.impl.launch.knot.KnotClient"
--- a/build.gradle
+++ b/build.gradle
@@ -363,6 +363,49 @@ loom {
 }
 test.dependsOn runGametest
 
+configurations {
+	productionRuntime {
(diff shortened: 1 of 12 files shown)


modmuss50:
Incase you were wondering, GHA gets ~6FPS 😄 
in_game_overworld

build.gradle
```diff
@@ -351,15 +351,66 @@ loom {
...
+}
+
+dependencies {
+	productionRuntime "net.fabricmc:fabric-loader:${project.loader_version}"
+	productionRuntime "net.fabricmc:intermediary:${project.minecraft_version}"
+}
+
+import net.fabricmc.loom.util.OperatingSystem
+
+// This is very far beyond loom's API if you copy this, your on your own.
```
Jab125:
```suggestion
// This is very far beyond loom's API if you copy this, you're on your own.
```

SquidDev:
:+1: This is really cool to see!

Not sure how useful this is, but also recently been having a stab at client tests (https://github.com/cc-tweaked/CC-Tweaked/pull/1219, though not sure how useful the diff actually is. Also, Mojmap warning). One pattern I've found pretty useful is writing client-side tests using vanilla's gametest framework.

Each client-side test is run in its own batch, and so has exclusive control of the player. Tests can then teleport the player around, make the player perform interactions, and then run assertions on the client or server. I've an example test [here](https://github.com/cc-tweaked/CC-Tweaked/blob/737d8a2585c7ebbb28fe02a49ac38cccf8c72375/projects/common/src/testMod/kotlin/dan200/computercraft/gametest/Computer_Test.kt#L90) (again, Mojmap warning) - apologies for the Kotlin DSL I've built on top of everything :D:.

That all said, while I think this is useful for content mods, possibly less helpful for library mods/Fabric API - the sorts of things which need testing are quite different! Not sure! :)

modmuss50:
Thanks :) 

What you have looks great, and think is similar to what this could be if we wanted to expand it beyond a smoke test and/or provide an API for other mods to use. Im a little on the fence if this is out of scope or not for Fabric API. Having seen how well what I have so far works it does give me more confidence in opening this up for other mods.

Like you said I cannot think of too many cases were we could write meaningful client tests in fabric API, the main thing I was after to was to just check it doesnt crash basically.

PR #2681, opened 2022-11-21
Add DataGeneratorEntrypoint.buildRegistry

modmuss50:
Used to allow data generation against custom dynamic registry entries such as biomes. Im not too happy with the way I build up the merge RegistryBuilder, I may revist this.


Technici4n (changes requested):
Can I ask for the following structure to be used in the testmod, or close to that? This is probably what many people are going to reference, so it would be best to move all of the creation code out of the datagen entrypoint and make it more idiomatic. What do you think?

image

Technici4n:
I think the new API is fine (just need to cleanup the testmod as I said above).

Not sure about the exact implementation of merging, but that can easily be improved later on without API breaks, so I wouldn't worry too much about it. :+1:

modmuss50:
@Technici4n I have cleaned up the testmod a little, hopefully its a bit more clear now. 👍

Technici4n (commented):
Thanks, looks nicer to me!

One last question: Minecraft initializes the RegistryLookup offthread (that's why it's a CF). Shouldn't we be doing the same? (Or did I misread the code?)

modmuss50:
Done, I dont think its really needed but we might as well use it.

Technici4n (approved):

PR #2682, opened 2022-11-21
Networking API: thread safe receiver registrations

apple502j:
No breaking change, major deprecations.

## Rationale
Quite a lot of mods misuse the Networking API. They call methods of the game within the network IO thread instead of server/render thread. This is very problematic:

- The network thread could cause a CME, disconnecting the user.
- Or the opposite, which is much worse: the server thread crashes with CME.
- Game could malfunction in a weird way.
- Fastutil objects do not check concurrent modifications, causing object corruption that is potentially exploitable. (See "Any Falling Block", an example of such bug in 1.12.2)

This isn't a rare occurrence - if you do a code search it is very easy to find hundreds of buggy code. And unlike the other threading bug (reading packet buf in server/render thread), this doesn't always crash.

## Solution
TODO


deprecated/fabric-containers-v0/src/client/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
```diff
@@ -63,14 +63,14 @@ public <C extends ScreenHandler> void registerFactory(Identifier identifier, Con
...
-		ClientPlayNetworking.registerGlobalReceiver(ContainerProviderImpl.OPEN_CONTAINER, (client, handler, buf, responseSender) -> {
+		ClientPlayNetworking.registerGlobalReceiver(ContainerProviderImpl.OPEN_CONTAINER, (buf, responseSender, runner) -> {
 			Identifier identifier = buf.readIdentifier();
 			int syncId = buf.readUnsignedByte();
 
 			// Retain the buf since we must open the screen handler with it's extra modded data on the client thread
 			buf.retain();
 
-			client.execute(() -> {
+			runner.run((client) -> {
```
modmuss50, line 73: Im not sure `runner.run` is ideal. I think it should be made more clear what it really does. context.schedule or something?

liach:
I suggest letting mods register one packet -> data object function and one data object + client/server player consumer, like for resource API, which is less error-prone and has cleaner user code. The original API should be kept as it offers max functionality possible, but renamed to like 'registerAdvancedGlobalReceiver' etc.

apple502j:
I'm not going to make a breaking change here, and I don't think the packet object design works well/is easier to use.

deprecated/fabric-containers-v0/src/client/java/net/fabricmc/fabric/impl/client/container/ScreenProviderRegistryImpl.java
apple502j: Yeah we can change it. Other suggestions?

modmuss50:
A higher level API that functions with Codec's may be an idea? Im not too sure how much it would get used though?

apple502j:
Codec is a bit complex for most uses. Though if we want packet object design (like in vanilla) I can add it.

Poll: should we adopt the packet object design?

liach:
The packet object design doesn't need you to write one lambda within another and already has an analogue in resource api. It's also much like vanilla packets having a constructor called on event loop and apply called on engine thread.

also i still suggest keeping the original api as it is useful for other purposes.

apple502j:
@liach Applying Deprecated annotation is the best way to indicate discouragement. (Notice there is no forRemoval.) For the unregister method, the return value could be changed to a simple boolean to not expose backing lambda (no mods use it anyway).

Technici4n:
After some consideration, I think this approach is good. We just need to bikeshed naming and how exactly we encourage people to move to the safer alternatives a bit. :smile:

I'm not a fan of packet objects cause they introduce a lot of boilerplate.

Technici4n (changes requested):
After some thinking, I'd rather use packet objects.

apple502j:
Will make a new PR.

PR #2684, opened 2022-11-21
Fix modded item groups not having id's assigned

AlphaMode:
Currently FabricItemGroupBuilderImpl doesn't do anything with the id. I assume this is a oversight and made a pr to fix it.
Currently I just GROUP_ID_MAP mutable and append the group and id to that. other solutions are possible. but I think this one has the smallest impact.


apple502j:
Hmm, can't we store it in a field of the instance?

modmuss50:
Good spot, definally an oversight from one of the ports.

modmuss50:
Many thanks for the PR, due to the urgency of this, I pushed the fix here: https://github.com/FabricMC/fabric/commit/8790b57d8c1cb37c0f9fc0f5a14e9a39f97d167d

PR #2686, opened 2022-11-23
Add `BakedModel` unwrapping API

### LINKED ISSUE #2600: Model Unwrapping API
Some existing mods like Continuity wrap `BakedModel`s to achieve effects like connected and emissive textures on models that the mod itself did not add. Some other mods need to access the root model that is not wrapped and is of a known class. Continuity and likely others use `ForwardingBakedModel` to wrap models, but this class does not have a getter for the base model. An example of a mod that needs to access the root model is Modern Industrialization, which at the moment uses reflection to access `ForwardingBakedModel`'s internal field.

Furthermore, with the changes of 1.19.3 comes a vanilla refactor that requires Continuity to wrap every single `BakedModel` instead of only a select few. This raises compatibility concerns, but a model unwrapping API can solve these. My original thought was to add the API to Continuity, but having it standardized like the facade API would be much easier to use.

My current idea of how this API would work is an interface such as
```java
public interface WrappedBakedModel {
    BakedModel unwrap();

    default BakedModel getRootModel() {
        BakedModel model = this;
        while (model instanceof WrappedBakedModel wrapped) {
            model = wrapped.unwrap();
        }
        return model;
    }
}
```
`ForwardingBakedModel` can implement this interface if needed, but using `ForwardingBakedModel` instead of an interface as the API directly seems like a bad idea because multiple inheritance is not possible in Java.

Technici4n:
Closes #2600.

Replaces this kind of ugly code from MI:
image

### DIFF
--- a/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/model/WrapperBakedModel.java
+++ b/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/model/WrapperBakedModel.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.renderer.v1.model;
+
+import net.minecraft.client.render.model.BakedModel;
+
+/**
+ * An interface to be implemented by models that wrap and replace another model, such as {@link ForwardingBakedModel}.
+ * This allows mods to access the wrapped model without having to know the exact type of the wrapper model.
+ *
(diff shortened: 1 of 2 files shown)


Juuxel (approved):

modmuss50:
Is this for 1.19.3 only?

Technici4n:
It's version independent and could be backported if we wanted to, but it's also not critical... More of a long-term improvement over the current workarounds. @PepperCode1 what do you think?

PepperCode1:
I think it should be added to 1.19.2 and 1.18.2 just in case.

modmuss50 (approved):

modmuss50:
Should technically go in last call, but prob doesnt need the whole week 👍

fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/model/WrapperBakedModel.java
```diff
@@ -0,0 +1,58 @@
...
+ * An interface to be implemented by models that wrap and replace another model, such as {@link ForwardingBakedModel}.
+ * This allows mods to access the wrapped model without having to know the exact type of the wrapper model.
+ *
+ * <p>If you need to access data stored in one of your {@link BakedModel} subclasses,
+ * and you would normally access the model by its identifier and then cast it:
+ * call {@link #unwrap(BakedModel)} on the model before, in case another mod is wrapping your model to alter its rendering.
+ */
+public interface WrapperBakedModel {
+	/**
+	 * Return the wrapped model, if there is one at the moment, or {@code null} otherwise.
```
Juuxel, line 33:
```suggestion
	 * {@return the wrapped model, if there is one at the moment, or {@code null} otherwise}
```

Technici4n: I didn't do this for consistency with the rest of the module.

PR #2689, opened 2022-11-23
Add coverage task

modmuss50:
Adds a coverage task used to genrate a JaCoCo coverage report (See build output). Datagen is not included, but this does include the client run, and gametests.

Im unsure if this is worth runing for every build as it can be a little slow. I think its worth adding even if it needs to be ran manually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,3 +54,26 @@ jobs:
         with:
           name: Test Screenshots
           path: run/screenshots
+
+  coverage:
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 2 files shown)


Technici4n:
I would maybe not run it in the CI, as it is also a bit of a waste. It's nice to have the option to run it locally to evaluate how we're doing wrt. tests though.

modmuss50:
Yeah, I agree and thats my plan. Just giving it a quick test on GHA to see how slow it is :D

modmuss50:
Took 16 mins (3 on my local machine) So yeah run it manually when needed 👍

PR #2690, opened 2022-11-24
Add ItemTooltipDataCallback event

ghost:
Ehm.. So I changed the branch name. I didn't think it would close the pr. #2675
Anyway, this is the same as before.

> Added an event that is fired when ItemStack#getTooltipData is called. After the event, if possible data list is wrapped with a BundledTooltipData which is later converted to BundledTooltipComponentImpl and rendered.


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipDataCallback.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.List;
+
+import net.minecraft.client.item.TooltipData;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public interface ItemTooltipDataCallback {
```
kevinthegreat1: I would suggest combining this with the `ItemTooltipCallback` class. So you would have one class with two event fields and two functional interfaces. `ServerMessageEvents` in `fabric-message-api` is an example.

I also suggest adding documentation about this event, like an explanation of what this does. Also, you should explain the difference compared to `ItemTooltipCallback` and what cases this api should be used for.

fabric-item-api-v1/src/client/java/net/fabricmc/fabric/mixin/item/client/ItemStackMixin.java
```diff
@@ -16,24 +16,44 @@
...
 
 @Mixin(ItemStack.class)
 public abstract class ItemStackMixin {
 	@Inject(method = "getTooltip", at = @At("RETURN"))
 	private void getTooltip(PlayerEntity entity, TooltipContext tooltipContext, CallbackInfoReturnable<List<Text>> info) {
 		ItemTooltipCallback.EVENT.invoker().getTooltip((ItemStack) (Object) this, tooltipContext, info.getReturnValue());
 	}
+
+	@Inject(method = "getTooltipData", at = @At("RETURN"), cancellable = true)
+	private void getTooltipData(CallbackInfoReturnable<Optional<TooltipData>> cir) {
```
kevinthegreat1, line 46:
```suggestion
	private void fabric_getTooltipData(CallbackInfoReturnable<Optional<TooltipData>> cir) {
```

fabric-item-api-v1/src/client/java/net/fabricmc/fabric/mixin/item/client/ItemStackMixin.java
```diff
@@ -16,24 +16,44 @@
...
 import net.minecraft.text.Text;
 
 import net.fabricmc.fabric.api.client.item.v1.ItemTooltipCallback;
+import net.fabricmc.fabric.api.client.item.v1.ItemTooltipDataCallback;
+import net.fabricmc.fabric.impl.item.BundledTooltipData;
 
 @Mixin(ItemStack.class)
 public abstract class ItemStackMixin {
 	@Inject(method = "getTooltip", at = @At("RETURN"))
 	private void getTooltip(PlayerEntity entity, TooltipContext tooltipContext, CallbackInfoReturnable<List<Text>> info) {
```
kevinthegreat1, line 41:
```suggestion
	private void fabric_getTooltip(PlayerEntity entity, TooltipContext tooltipContext, CallbackInfoReturnable<List<Text>> info) {
```

kevinthegreat1 (commented):
I think it would be better to have the mod id in the mixins.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipDataCallback.java
Technici4n: I think the way this PR does it is fine. The XxxCallback format is fine when the event is standalone.

fabric-item-api-v1/src/client/java/net/fabricmc/fabric/mixin/item/client/ItemStackMixin.java
Technici4n: This is unneeded for callback injectors, as mixin already prefixes them. They're nice to have for stack traces, but not mandatory.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipDataCallback.java
Technici4n: I agree with documentation comment: the interface should have a comment explaining the high-level point of the event (even though it's obvious for us, it might not be for beginners), and any useful implementation details such as when it's fired.

Technici4n: Important things to document:
- the normal tooltip data is already in the list by the time this event is called.
- all the tooltip components in the list will be rendered, in order.

Technici4n:
Just noticed one thing: this shouldn't be in the client package.

Technici4n:
Actually, I just noticed one thing:

If this is an event, what's the point of using `TooltipData` at all? We could just go for a client-only event to combine `TooltipComponent`s... Or is vanilla making this harder than it should be? 🤔

ghost:
By default implementation of vanilla you don't have access to the stack while constructing components, they are handled in different methods.

<details>
  <summary>Code block from Screen</summary>

```java
    protected void renderTooltip(MatrixStack matrices, ItemStack stack, int x, int y) {
        this.renderTooltip(matrices, this.getTooltipFromItem(stack), stack.getTooltipData(), x, y);
    }

    public void renderTooltip(MatrixStack matrices, List<Text> lines, Optional<TooltipData> data, int x, int y) {
        List<TooltipComponent> list = (List)lines.stream().map(Text::asOrderedText).map(TooltipComponent::of).collect(Collectors.toList());
        data.ifPresent((datax) -> {
            // TooltipComponentCallback is fired here
            list.add(1, TooltipComponent.of(datax));
        });
        this.renderTooltipFromComponents(matrices, list, x, y);
    }
```
</details>

We can't really change anything without basically overwriting the one with the stack.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/client/item/v1/ItemTooltipDataCallback.java
```diff
@@ -0,0 +1,61 @@
...
+
+import net.minecraft.client.item.TooltipData;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Event for adding more tooltip data to a stack's tooltip.
+ * Called when an item stack's tooltip is rendered.
```
kevinthegreat1: The only thing I would suggest is mentioning the difference to `TooltipComponentCallback`. Maybe state that this does not require overriding `Item.getTooltipData` for every single item. For single item or custom items, you should still use `TooltipComponentCallback`, is that correct?

ghost: Yeah, it is correct, if their data only affects one item, they should override the vanilla method. I'll mention it.

ghost:
Apperently I don't know how to request reviews. Why does it remove others when requesting from someone?

ghost:
By the way I would like to move `TooltipComponentCallback#EVENT` invoker from `Screen` to `TooltipComponent.of`. It's more generalized. And makes the event compatible with mods that don't require fabric api, as they'll use that method to get components.

Technici4n:
> By the way I would like to move `TooltipComponentCallback#EVENT` invoker from `Screen` to `TooltipComponent.of`. It's more generalized. And makes the event compatible with mods that don't require fabric api, as they'll use that method to get components.

Yes, that's a good idea. Probably fine for this PR.

kevinthegreat1 (approved):

Technici4n (changes requested):
I think this is a very simple and useful addition. A few requests:
- Can you change `BundledTooltipComponent/Data` to something that doesn't look like it would be for bundles. Perhaps `ListTooltipComponent/Data` would work.
- Why is this not in the rendering API? It could go there next to the existing `TooltipComponentCallback` event.

Technici4n:
> Why is this not in the rendering API? It could go there next to the existing TooltipComponentCallback event.

I guess you placed it next to the `ItemTooltipCallback`, which makes sense. So maybe leave it in the item API, but move it to the client source set?

ghost:
Well I put it in the item API because it's where `ItemTooltipCallback` is. 
Also with my last commit I moved everything to client sourceset to match the change in #2755 
and refactored the mixin package on the client sourceset to match the standard.
`mixin.item.client` -> `mixin.client.item`

ghost:
> > Why is this not in the rendering API? It could go there next to the existing TooltipComponentCallback event.
> 
> I guess you placed it next to the `ItemTooltipCallback`, which makes sense. So maybe leave it in the item API, but move it to the client source set?

Ok I don't know how I missed this entire message but I did exactly what you asked anyway :D

ghost:
~~Workflow fail is not related to me. It fails to download something and throws a `FileNotFoundException`.~~ Nevermind it succeeded.

Technici4n (approved):

Technici4n (changes requested):
Sorry for the change of plans, but I have a (I think) better idea. A big problem with the current tooltip system is that it offers poor ordering capabilities between text and component tooltips. So what if we had the following design instead? What do you think?

It should be possible to hook this up by capturing the `ItemStack` instance as a field in the relevant `Screen` methods.

```java
public interface ModifyTooltipContentCallback {
    // called right before a tooltip is rendered, already contains the item's tooltip text and tooltip data
    // stack is EMPTY if this tooltip is not for a stack
    void modifyTooltipContent(List<Entry> entries, ItemStack stack, Screen screen, TextRenderer textRenderer, int mouseX, int mouseY);

    sealed interface Entry { }
    record Text(Text text, boolean shouldWrap) implements Entry { }
    record Component(ClientTooltipComponent component) implements Entry { }
}
```

This would go in the rendering API, and we could deprecate the existing `ItemTooltipCallback`.

A side-effect would be that we finally get proper opt-in tooltip wrapping.

ghost:
I am in favor of a change like this but I think the implementation is a little restrictive, even with current events you don't really have any control over where you are adding components. I think I can mockup a pr in a day or two for may take on an implementation if you'd like to see.

Also I'm not really fond of the wrapper `Entry` interface, I think text wrapping function should be implemented through interface injection to `Text` class. With methods like `doWrap(boolean)` and `shouldWrap()`. Then convert to a different `TooltipComponent` if they should be wrapped.

Patbox:
Interface injection into Text would be confusing, as it is used in many places that have different rules about wrapping. Better alternative would be replacing list with custom class like ItemGroup do for example

ghost:
> Interface injection into Text would be confusing, as it is used in many places that have different rules about wrapping. Better alternative would be replacing list with custom class like ItemGroup do for example

Ah yes, I forgot everything is Text. By the way I was just trying something similar to your suggestion, using a custom list for stack tooltips.

ghost:
@Technici4n I did a pretty small thing, can you check it and let me know what you think. [The branch](https://github.com/Sideroo/fabric/tree/tooltips)

In summary, I replaced the list in `ItemStack#getTooltip` with a custom one that directly stores `TooltipComponent`'s inside itself. Texts are converted to components via a new event (only in some places, will check for other applicable places). I believe this event can be used for checking if the text should be wrapped and if so return a custom component. Added `ItemTooltipEvents` invoked in different stages of `ItemStack#getTooltip`.

Technici4n:
Why not use event phases for ordering?

I think requiring an explicit wrapping parameter isn't too bad, it forces the modder to choose and generally be aware of wrapping, while keeping it opt-in. We could add a record constructor that sets wrapping to `true` if you think it'd too verbose otherwise.

I went for the list because I thought it would be the simplest way to do what I wanted.

And I went for the sealed interface because we fundamentally can have two types of entries. This is still a bit awkward at the moment, but it will get nicer with record switched and patterns in a few java versions. :)

I think my solution is likely the simplest, but of course feel free to suggest anything else, and in any case thanks for tackling this. :)

Technici4n:
Just saw your message, I'll be able to check what you did in ~20 mins.

ghost:
> Just saw your message, I'll be able to check what you did in ~20 mins.

No need to rush, It's late for me anyways. I can't check back till morning.

Technici4n:
A few comments:
- I don't really see the point of `TextToComponentCallback` ? Is that for tooltip wrapping ? Cause the point of wrapping is to be reasonably automatic, and I don't want it to be more complicated than passing some boolean value. Requiring a separate event is quite annoying as that requires a lot of extra work for modders, and I think we want all wrapped tooltips to be wrapped in the same way, which doesn't work if every modder does it their way.
- The approach of working directly at the `ItemStack#getTooltip` level is quite interesting. The `TooltipList` should definitely not be in the API though, it's too hacky imo.
- We need a reasonable way to inspect tooltips, that's why I don't like immediately converting the `Text` to a `ClientTooltipComponent`. The simple solution is to delay this conversion so that it happens after the event(s).
- This won't work nicely for non-item tooltips, but those could likely use a separate event with a somewhat similar system? At least the phases wouldn't be needed there.
- `ItemStack#getTooltip` is missing some context such as the current screen and the mouse position. These might be useful to have in a more general event, but maybe not for item stacks? What if item tooltips went through 2 events (item-specific one injecting in `getTooltip` and more general tooltip-rendering event)?

ghost:
* Yes, the `TextToComponentCallback` event looks unnecessary.
* I don't think returning an extension of the previously returned value is hacky.
* Wanting to delay `Text` to `ClientTooltipComponent` is reasonable, maybe instead of an internal `ClientTooltipComponent` list we can use list of a wrapper interface like you suggested above.
* I agree with the idea of having two events.

I'll look deeper into this when I have some time.

Technici4n:
> I don't think returning an extension of the previously returned value is hacky.

The concept in general is not, but extending `ArrayList` to also add support for other kinds of entries is extremely hacky. I also don't think remove will work correctly if someone tries to remove something from the middle of the list. This makes me think that my original suggestion might be the simplest to implement. :P

ghost:
I updated [the branch](https://github.com/Sideroo/fabric/tree/tooltips) again. I believe this is somewhat closer to what you were looking for. If the feature is likely to happen I can open a pr for it after I make it more fabric-like.

> * The approach of working directly at the `ItemStack#getTooltip` level is quite interesting. The `TooltipList` should definitely not be in the API though, it's too hacky imo.

Did you mean the `api` package by this? I think I misunderstood you and thought you meant the class is too hacky. Now that I am re-reading this I think you mean it shouldn't be directly exposed in the events and should be in the `impl` package? Am I right?

Technici4n:
I meant the `api` package, but I'd say it's very hacky and doesn't properly handle removals in general. I think this general approach doesn't work very nicely with removals or modification of tooltip entries in general.

What I would do instead if I wanted to insert a `ClientTooltipComponent` in a very specific position using one of these events:
- Add a marker `Text` that I could replace later. Maybe something like `Text.literal("mymod:tooltip_component_marker")`.
- I would replace the marker `Text` by a proper `ClientTooltipComponent` later on in the more general tooltip event that I suggested.

Sadly this is not the cleanest solution as far as API goes, but it's the most compatible one with what vanilla is doing I believe?

So, TLDR of what I suggest:
- Pass a `List<Text>` to the various hooks in `ItemStack#getTooltip`.
- Add an event to modify the tooltip similar to what I suggested in https://github.com/FabricMC/fabric/pull/2690#pullrequestreview-1217901703.

Of course feel free to disagree, but I think the `TooltipList` with its current design is trying too much to hack around vanilla, and this won't work nicely with existing `ItemTooltipCallback` listeners.

LambdAurora:
> [...], but move it to the client source set?

TooltipData are not client-only so I'm not entirely sure it makes sense to have the TooltipData part be client-only.

Technici4n:
`TooltipData` only exists because `TooltipComponent` can't be used in common classes; it's only used on the client side.

PR #2695, opened 2022-11-25
Add access widener for SoundEvent.of

### LINKED ISSUE #2694: SoundEvent turned private
In 1.19.3-pre2 the SoundEvent class is private and thus you are unable register SoundEvents as described here https://fabricmc.net/wiki/tutorial:sounds
The static constructor methods of SoundEvent also seem to be protected. The SoundEvents class can register them due to being in the same package but as a modder it seems like we need to use AW or a Mixin now.

modmuss50:
Fixes #2694

### DIFF
--- a/fabric-transitive-access-wideners-v1/src/main/resources/fabric-transitive-access-wideners-v1.accesswidener
+++ b/fabric-transitive-access-wideners-v1/src/main/resources/fabric-transitive-access-wideners-v1.accesswidener
@@ -113,7 +113,7 @@ transitive-accessible method net/minecraft/village/VillagerType <init> (Ljava/la
 # Access stack list in SimpleInventory
 transitive-accessible field net/minecraft/inventory/SimpleInventory stacks Lnet/minecraft/util/collection/DefaultedList;
 
-# Creating custom Sound event's
+# Creating custom sound events
 transitive-accessible method net/minecraft/sound/SoundEvent of (Lnet/minecraft/util/Identifier;F)Lnet/minecraft/sound/SoundEvent;
 transitive-accessible method net/minecraft/sound/SoundEvent of (Lnet/minecraft/util/Identifier;)Lnet/minecraft/sound/SoundEvent;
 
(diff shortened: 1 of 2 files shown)


apple502j (approved):

fabric-transitive-access-wideners-v1/template.accesswidener
```diff
@@ -113,4 +113,8 @@ transitive-accessible method net/minecraft/village/VillagerType <init> (Ljava/la
 # Access stack list in SimpleInventory
 transitive-accessible field net/minecraft/inventory/SimpleInventory stacks Lnet/minecraft/util/collection/DefaultedList;
 
+# Creating custom Sound event's
```
Juuxel:
```suggestion
# Creating custom sound events
```

fabric-transitive-access-wideners-v1/src/main/resources/fabric-transitive-access-wideners-v1.accesswidener
```diff
@@ -113,6 +113,10 @@ transitive-accessible method net/minecraft/village/VillagerType <init> (Ljava/la
 # Access stack list in SimpleInventory
 transitive-accessible field net/minecraft/inventory/SimpleInventory stacks Lnet/minecraft/util/collection/DefaultedList;
 
+# Creating custom Sound event's
```
Shnupbups:
```suggestion
# Creating custom sound events
```

modmuss50: Good thinking, we should add something to check that there arent any generated file chages 🤔

Juuxel: Seems simple enough, I could make a PR for a task to check that the real generated file matches the output

oliviathevampire (approved):

Shnupbups (approved):

PR #2696, opened 2022-11-25
Pass DynamicRegistryView to DynamicRegistrySetupCallback

apple502j:
This introduces `DynamicRegistryView`, which is an context object passed to `DynamicRegistrySetupCallback`. This solves issues with passing a raw `DynamicRegistryManager`, such as `get` method crashing the game.

### Is this a breaking change?
Yes. However, this shouldn't affect most mods; the usage of this API is very low (note, though that they are popular mods). Additionally the `getOptional` method signature is intact.

### Why getOptional (despite the guideline)?
To make it consistent with `DynamicRegistryManager`, and to make it clear that it is more often than not that this doesn't return the registry.

### Why construct and pass a context object?
First, this used `DynamicRegistryManager` as a context object before. (The method doesn't have access to DRM anymore since a few weeks ago.) Second, adding a context object means we can extend it for common uses.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/DynamicRegistryViewImpl.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/DynamicRegistryViewImpl.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/DynamicRegistryViewImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+import net.minecraft.registry.DynamicRegistryManager;
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+
+import net.fabricmc.fabric.api.event.registry.DynamicRegistryView;
+import net.fabricmc.fabric.api.event.registry.RegistryEntryAddedCallback;
+import net.fabricmc.fabric.api.event.registry.RegistryEntryRemovedCallback;
+
+public class DynamicRegistryViewImpl implements DynamicRegistryView {
```
modmuss50: Nit: final

modmuss50 (approved):
Not super happy about a breaking change in the pre's but better than never.

Change looks good tho.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/DynamicRegistryViewImpl.java
Juuxel:
```suggestion
public final class DynamicRegistryViewImpl implements DynamicRegistryView {
```

Technici4n (approved):

PR #2699, opened 2022-11-26
TAW: Check that generated access widener matches expected output

Juuxel:
See https://github.com/FabricMC/fabric/pull/2695#discussion_r1032199744.

This prevents builds with outdated access wideners when the MC version changes, or when template.accesswidener is modified.

Also simplifies the code a bit - `text.lines().collect(Collectors.toCollection { [] })` for getting a mutable list of the lines in a file was replaced with Groovy's `readLines` extension method wrapped in a new `ArrayList` for safety.

### DIFF
--- a/fabric-transitive-access-wideners-v1/build.gradle
+++ b/fabric-transitive-access-wideners-v1/build.gradle
@@ -14,19 +14,44 @@ import java.nio.file.FileSystem
 import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Path
-import java.util.stream.Collectors
+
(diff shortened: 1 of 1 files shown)


modmuss50:
This is a good start, however there are a few modules that have generated TAWs and/or resources. This only covers one module. Running the `generateResources` task during the github action run and then ensuing that git has no changed files may be an easier way to do this?

Juuxel:
That's true but needs some more work with the scripting (ofc individually doing my thing for all modules would take more time still) and this can be used locally for validation which is a plus.

I might still PR the clean-ups from here though, no reason to use streams here.

PR #2703, opened 2022-11-26
Resource Condition: `fabric:never`

ghost:
Fabric equivalent of `forge:false`. Prevents a resource from being loaded.

Most common use case would be mods disabling specific vanilla recipes that don't go along with their mod.

Also added a helper data gen method,`FabricRecipeProvider#disabled`, and moved `ResourceConditionsImpl#featuresEnabled` directly under other json providers, before condition implementations, to match the rest of the file.

By the way, `DefaultResourceConditions#never` can be a field since it has no properties. Wanted to ask this because there isn't any other conditions like this.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipeProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipeProvider.java
@@ -25,6 +25,7 @@ import java.util.function.Consumer;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 import com.google.gson.JsonObject;
+import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.data.DataProvider;
 import net.minecraft.data.DataWriter;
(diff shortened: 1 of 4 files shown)


haykam821:
I don't see the use case as recipes can be disabled through other means already. Also, wouldn't this condition just cause the recipe to default to whatever was loaded in a lower priority data pack?

JieningYu:
> I don't see the use case as recipes can be disabled through other means already. Also, wouldn't this condition just cause the recipe to default to whatever was loaded in a lower priority data pack?

for example: use a datapack to remove a recipe

apple502j:
Can't we use resource filter in `pack.mcmeta`?

haykam821:
I agree with @apple502j; using filters should be the recommended way to accomplish this goal. For example, `pack.mcmeta` could include the following to remove the beacon recipe:

```json
{
	"filter": {
		"block": [
			{
				"namespace": "minecraft",
				"path": "recipes/beacon\\.json"
			}
		]
	}
}
```

Technici4n:
How do these filters scale if you need to filter out 100s of recipes?

BTW, this can also be accomplished by using the `or` condition with no inner conditions.

ghost:
Oh I didn't know that filtering thing. If that's possible, then yeah this isn't really necessary.

PR #2705, opened 2022-11-27
Check stack size in FabricItemGroupEntries

modmuss50:
Vanilla checks the stack size later in ItemGroup.EntriesImpl and crashes. This PR adds the same check to FabricItemGroupEntries, resuting in a much more useful stacktrace to help find the cause of empty stacks in an item group.

I found this when porting one of my mods to 1.19.3


fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/FabricItemGroupEntries.java
```diff
@@ -472,4 +488,20 @@ private static void addAfter(ItemConvertible anchor, Collection<ItemStack> newSt
...
+		}
+	}
+
+	private static void checkStack(ItemStack stack) {
+		if (stack.isEmpty()) {
+			throw new IllegalArgumentException("Cannot add empty stack");
+		}
+
+		if (stack.getCount() != 1) {
+			throw new IllegalArgumentException("Stack size must be exactly 1");
```
apple502j: This should have the item's string representation.

Technici4n (approved):

PR #2706, opened 2022-11-27
Add WorldPresetProvider to data-gen module

ghost:
Since a few versions, world presets have been datapackable, so why not have a data provider for it.


ghost:
Custom world presets are not a commonly used feature so just tell me if worldgen related data generation is not in scope.  Because if this gets committed, I might also pr others. (Dimension types, carvers, features, etc.)

modmuss50:
This should already be covered by the `FabricDynamicRegistryProvider`. Is this not the case?

ghost:
It is indeed the case 🤦‍♂️

PR #2709, opened 2022-11-28
fabric-resource-loader-v0: make language files under all namespaces accessible

teddyxlandlee:
### Why I PR this

In client, language files can be loaded in any namespaces, even if it is not of any mod id.

For example, a mod identified as `cute-octopus` can include both `assets/cute-octopus/lang/*.json` and `assets/cute-octopus-extra/lang/*.json`.

Currently, when mods like `cute-octopus` running on a dedicated server, only `assets/cute-octopus/lang/en_us.json` can be loaded. However, things in `assets/cute-octopus-extra` should also be loaded, as modders might have their own reason to split their language files into different namespaces, which is legal on the client.

This pull request introduces a way to load language files in alternative namespaces on the dedicated server.

Due to backwards compatibility with mods, alternative namespaces are automatically detected in which the `en_us.json` is loaded when server language is to be loaded.

### Testmod

A language file in alternative namespace (`fabric-resource-loader-v0-testmod-test1`) is loaded by the resource loader on dedicated server.

You can find its corresponding test in `LanguageTestMod.java`.


Technici4n:
I'd prefer if the server language logic was adjusted to match that of the client. I think that would be a simpler approach.

teddyxlandlee:
So @Technici4n do you prefer listing possible namespaces in each `ModContainer`, and load them respectively?

Technici4n:
I'd suggest scanning for all `language/en_us.json` files.

Technici4n:
Ah, now I see what you mean. Yes, listing namespaces might indeed be the easiest way to do that with good performance. My point is that users should not have to configure anything for their lang files to be loaded on the server. :)

Technici4n:
To read the namespaces, you can probably make this function `public` and use it:
https://github.com/FabricMC/fabric/blob/70063eb9397d2a82209d3c048881952ab3cae94d/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java#L110.

teddyxlandlee:
Is it acceptable to move the logical code from `mixin` to `impl`? They're like fifty lines occupation.

Technici4n:
As you prefer.

teddyxlandlee:
@Technici4n Done

fabric-resource-loader-v0/src/testmod/resources/fabric.mod.json
```diff
@@ -8,6 +8,11 @@
   "depends": {
     "fabric-resource-loader-v0": "*"
   },
+  "custom": {
```
Technici4n: This isn't needed anymore. Looks good otherwise!

teddyxlandlee:
Testmods & checkstyles have been passed. Ready to merge🚀

Technici4n (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ServerLanguageUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import net.minecraft.resource.ResourceType;
+import net.minecraft.util.Language;
+
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+
+public class ServerLanguageUtil {
```
modmuss50: Nit: final and private constructor

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ServerLanguageUtil.java
```diff
@@ -0,0 +1,52 @@
...
+			final Map<ResourceType, Set<String>> map = ModNioResourcePack.readNamespaces(mod.getRootPaths(), mod.getMetadata().getId());
+
+			for (String ns : map.get(ResourceType.CLIENT_RESOURCES)) {
+				mod.findPath(ASSETS_PREFIX + ns + "/lang/" + Language.DEFAULT_LANGUAGE + ".json")
+						.filter(Files::isRegularFile)
+						.ifPresent(paths::add);
+			}
+		}
+
+		return paths;
```
modmuss50: Nit: `Collections.unmodifiableCollection`

modmuss50 (approved):
Looks good, just 2 small nit picks that would be nice to fix 👍

Juuxel (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ServerLanguageUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.util.Language;
+
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+
+public class ServerLanguageUtil {
+	private static final String ASSETS_PREFIX = ResourceType.CLIENT_RESOURCES.getDirectory() + '/';
+
+	public static Collection<Path> getModLanguageFiles() {
+		Set<Path> paths = new HashSet<>();
```
apple502j: I believe this needs to be LinkedHashSet

apple502j (changes requested):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ServerLanguageUtil.java
teddyxlandlee: Since it's going to be ordered, why not using `ArrayList`?

teddyxlandlee:
> https://github.com/FabricMC/fabric/blob/d3e715aa7e0d1df037ad70f54576066759aeb3b1/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModNioResourcePack.java#L110
> .

Is it "bad" to make anything package-private in Fabric API?  
In other words: do I need to change it protected/public?

Technici4n:
Why would package private be bad?

apple502j (approved):

teddyxlandlee:
Is it worth backporting to older versions like 1.18.2?

PR #2719, opened 2022-11-30
Add DynamicRegistryRegistry and DynamicRegistryFinalizeCallback

### LINKED ISSUE #1012: Hooks for adding custom registries into DynamicRegistryManager
Currently, the list of registry infos in `DynamicRegistryManager` is hardcoded; it would be nice to be able to add our own entries to that list.

### LINKED ISSUE #2111: DynamicRegistryTailCallback

ghost:
PR overview:

* `DynamicRegistryRegistry` : A helper class for registering new dynamic registries. (Name debatable)
* `DynamicRegistryFinalizeCallback` : An event fired just before DRM is immutablized.
* Updated test mod, moved dynamic registry related stuff to it's own initializer.

Closes #1012 
Closes #2111
Replaces #2112

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
@@ -40,8 +40,8 @@ import net.fabricmc.fabric.mixin.registry.sync.RegistryLoaderAccessor;
  * <p>NOTE: These registries are not reloadable.
  *
  * <p><b>Experimental feature</b>, dynamic registries are a pretty new feature, and Mojang
- * can make any changes to these registries during snapshots. Since we cannot guarantee this
- * will work in the next snapshots, we reserve the right to remove or change it without further notice.
+ * can make any changes to their implementation in the upcoming versions. Since we cannot guarantee this
+ * will work in the next versions, we reserve the right to remove or change it without further notice.
  *
  * @apiNote The path used for {@code key} is used as path for resource search as well,
  * so {@code "fabric-api:dynamic_data"} would be searched in {@code "resources/data/modid/dynamic_data/"},
@@ -57,10 +57,10 @@ public class DynamicRegistryRegistry {
 	 * @param codec The codec used for deserialization
 	 * @throws IllegalStateException if key path clashes with an already registered entry
 	 */
-	public static <T> RegistryLoader.Entry<T> register(RegistryKey<? extends Registry<T>> key, Codec<T> codec) {
+	public static <T> void register(RegistryKey<? extends Registry<T>> key, Codec<T> codec) {
 		Objects.requireNonNull(key, "Registry key cannot be null!");
 		Objects.requireNonNull(codec, "Codec cannot be null!");
-		return register(new RegistryLoader.Entry<>(key, codec));
+		register(new RegistryLoader.Entry<>(key, codec));
 	}
 
 	/**
@@ -69,12 +69,11 @@ public class DynamicRegistryRegistry {
 	 * @param entry The entry
 	 * @throws IllegalStateException if key path clashes with an already registered entry
 	 */
-	public static <T> RegistryLoader.Entry<T> register(RegistryLoader.Entry<T> entry) {
+	public static <T> void register(RegistryLoader.Entry<T> entry) {
 		Objects.requireNonNull(entry, "Entry cannot be null!");
 		validateKey(entry.key());
 		MUTABLE_REGISTRIES.add(entry);
 		RegistryLoaderAccessor.setDynamicRegistries(Collections.unmodifiableList(DynamicRegistryRegistry.MUTABLE_REGISTRIES));
-		return entry;
 	}
 
 	/**
@@ -86,11 +85,11 @@ public class DynamicRegistryRegistry {
 	 * @param codec The codec used for deserialization
 	 * @throws IllegalStateException if key path clashes with an already registered entry
 	 */
-	public static <T> RegistryLoader.Entry<T> registerBefore(RegistryKey<? extends Registry<?>> before, RegistryKey<? extends Registry<T>> key, Codec<T> codec) {
+	public static <T> void registerBefore(RegistryKey<? extends Registry<?>> before, RegistryKey<? extends Registry<T>> key, Codec<T> codec) {
 		Objects.requireNonNull(before, "Before key cannot be null!");
 		Objects.requireNonNull(key, "Registry key cannot be null!");
 		Objects.requireNonNull(codec, "Codec cannot be null!");
-		return registerBefore(before, new RegistryLoader.Entry<>(key, codec));
+		registerBefore(before, new RegistryLoader.Entry<>(key, codec));
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


ghost:
By the way `impl.registry.sync.packet.NbtRegistryPacketHandler` seems to be marked for removal (has a "TODO: Remove"), I can remove it as a part of this pr, along with it's uses in test mod, which seems to be the only place it's used.

ghost:
One last thing, subpackages for the api part of the module are inconsistent. It goes like `api.event.registry` and contains everything,  even though some are not event related. I didn't touch anything because it would be a breaking change, and I followed it. Just wanted to point it out.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * Registers a dynamically loaded registry.
+	 *
+	 * @param entry The entry
+	 * @throws IllegalStateException if key path clashes with an already registered entry
+	 */
+	public static <T> RegistryLoader.Entry<T> register(RegistryLoader.Entry<T> entry) {
+		Objects.requireNonNull(entry, "Entry cannot be null!");
+		RegistryLoader.DYNAMIC_REGISTRIES.stream().filter(e -> e.key().getValue().getPath().equals(entry.key().getValue().getPath())).findFirst().ifPresent(e -> {
```
haykam821: This line is a bit unwieldy.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+			throw new IllegalStateException("Dynamic registry path clash between " + e + " and " + entry);
+		});
+
+		List<RegistryLoader.Entry<?>> list = new ArrayList<>(RegistryLoader.DYNAMIC_REGISTRIES);
+		list.add(entry);
+		RegistryLoaderAccessor.setDynamicRegistries(Collections.unmodifiableList(list));
```
haykam821: Is there a simpler way to do this, perhaps?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+import net.fabricmc.fabric.mixin.registry.sync.RegistryLoaderAccessor;
+
+/**
+ * @apiNote The path used for {@code key} is used as path for resource search as well,
+ * so {@code "fabric-api:dynamic_data"} would be searched in {@code "resources/data/modid/dynamic_data/"},
+ * {@code modid} being mod id of any mod, it is recommended to use your mod id in
+ * registry path to avoid path clashes, like {@code "fabric-api:fabric-api/dynamic_data"}.
+ */
+public class DynamicRegistryRegistry {
```
haykam821, line 52: Should we mark this API/class as experimental?

haykam821 (changes requested):
I personally think that a new module should be created for dynamic registries and possibly other registries, as was done in #1031.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
ghost: I can redirect the `List.of` and return a mutable list, though it might be compromising since modders would have full access to the list.

ghost:
> I personally think that a new module should be created for dynamic registries and possibly other registries, as was done in #1031.

I partially agree, it doesn't fit any other module, but I also think it's too small to be a module. Expanding registry-sync module to a registry-related module seems more reasonable to me.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
ghost: Had to check it somehow, most I can do is something like below, which only makes it look slightly cleaner.
```java
String path = entry.key().getValue().getPath();
RegistryLoader.DYNAMIC_REGISTRIES.stream().filter(e -> e.key().getValue().getPath().equals(path)).findFirst().ifPresent(e -> {
```

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
ghost: I decided to keep a private mutable list in the class.

Technici4n:
@haykam821 I disagree. This:
- is fine in the regsync module, and
- should not be done using entrypoints.

haykam821:
I was mainly referring to the module and not the API design. The `fabric-registry-sync-v0` module is getting overstuffed and should have been primarily a helper module rather than an API module.

Technici4n:
It should simply contain all registry-related things, even though this isn't obvious due to the legacy name.

Technici4n:
This module already has the DynRegSetupCallback, and various registry events that work with dynamic registries. Might as well support adding dynregs in there.

haykam821:
The location of `DynamicRegistrySetupCallback` was a mistake when it was first merged, and it remains a mistake today. [My stance on this issue has stayed the same.](https://github.com/FabricMC/fabric/pull/2112#issuecomment-1147041087)

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryRegistry.java
haykam821: I think a stream is the wrong tool for this job, because filtering a stream doesn't short-circuit. If a stream is used, though, its calls should be formatted over several lines.

ghost:
I'm converting this pr to draft because I just realized previously loaded dynamic registries can be accessed from the registry being loaded at the time. So I want to add methods to register before and after specific entries.

ghost:
Okay I updated the pr, and also marked it as experimental like suggested since Mojang can make any changes to these during snapshots and pre-releases. I think it should stay as an experimental feature until full 1.19.3 release.

ghost:
I need some opinions on this. Are return values for the registering methods really necessary? I can't think of any situation where people would need the entry.

ghost:
Closes #1012

ghost:
I don't want to overpopulate this pr but is a `DynamicRegistryFinalizeCallback` suitable for this pr. Or should I open another one after this. The idea is that it will be fired at the end of `RegistryLoader.load` just before DRM becomes immutable.

And which one is the best for the callback parameter:
* Directly pass the mutable DRM and hope no one calls `toImmutable()`.
* Pass a `DynamicRegistryView` and warn modders that they shouldn't use entry added/removed callback helper methods.
* Wrap the mutable DRM with a custom DRM that doesn't implement `toImmutable()`.

ghost:
Hopefully updated this pr for the last time, at least I've got nothing else stuck on my mind.
I'll just be waiting for reviews from now on.

dhyces:
looking forward to this

Issue #871, opened 2020-07-07
Allow modifying the climbing speed for climbable blocks


haykam821:
While blocks can already be added to the `minecraft:climbable` tag, the speed at which a player can climb the block cannot be modified. Being able to adjust the speed of a climbable block, especially based on the climber or [the block's state](https://github.com/FabricMC/fabric/issues/636), would be useful in extending the climbable tag provided by vanilla. While #62 already implemented this, the modifier would only be applicable if the block also has the `minecraft:climbable` tag.

i509VCB:
This is actually something I was covering in an experimental fablabs project I call `block-extensions`. Hopefully in a week or two I'll have something to resolve that.

PR #2721, opened 2022-12-03
Add LivingEntity climbing events

### LINKED ISSUE #636: Expand on the Climbing API
I was talking in the [Fabric Discord](https://discord.com/channels/507304429255393322/507982478276034570/713925476774838312) about what the Fabric equivalent is of Forge's isLadder() hook. I was using the hook to make my block climbable only when it has certain blockstates but since Vanilla uses a climbable tag that is only applied to the block itself, I cannot easily make my block climbable based on blockstates.

If I apply climbable tag on my block, then when my block is facing only downward, the player ends up floating when they walk on the block which doesn't look right. Hence this request.
image

The someone came up with is that Fapi should provide a way to let modders adjust climbing speed based on blockstates or at least a toggle of some kind. I'll be totally find with any way this gets implemented as this can help people make more interesting kinds of blocks and more control over the climbing feature. Maybe even a block that becomes harder to climb when it is raining or become more climbable the more broken it is. :)

Until then, I'll be looking into making a mixin for my use-case but other may really like this feature in Fapi

ghost:
Adds new events to `ServerLivingEntityEvents`.

* `ALLOW_CLIMB` which if any callback returns false, pos is declared unclimbable.
* `MODIFY_CLIMBING_SPEED` which allows players to modify the climbing motion.

Closes #636 and #871

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
@@ -78,34 +78,37 @@ public final class ServerLivingEntityEvents {
 
 	/**
 	 * An event that is called when minecraft checks if an entity is at a climbable block position.
-	 * This is fired before minecraft does its own checks.
+	 * This is fired before minecraft does its own checks, so it's not guaranteed that block at pos
+	 * is in {@link net.minecraft.registry.tag.BlockTags#CLIMBABLE}.
 	 *
-	 * <p><b>Return types:</b>
-	 * <ul>
-	 * 	   <li>{@link ActionResult#SUCCESS} Allows climbing unless another callback returns {@link ActionResult#FAIL}</li>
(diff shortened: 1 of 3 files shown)


ghost:
~~I feel like I should split this event into two events that use boolean instead of ActionResult as return type.
One for mods to check for their own climbable blocks. 
And one for when minecraft tries to return true. (To check for extra stuff)~~
Nevermind this.

Juuxel:
`ActionResult` is not a good return value for events unless you actually specify the behaviour of `CONSUME` and `CONSUME_PARTIAL`.

ghost:
> `ActionResult` is not a good return value for events unless you actually specify the behaviour of `CONSUME` and `CONSUME_PARTIAL`.

They acted same as `SUCCESS`, I checked for `ActionResult#isAccepted` in case someone uses them.
Anyway it doesn't matter anymore, rewrote the pr 🙂

ghost:
~~uhh, checkstyle builds on my computer idk why it gave 3 errors in here.~~ 
Nevermind, I'm stupid I was running checkstyle for data generation api 🤦

PR #2726, opened 2022-12-04
Screen API: Fix MinecraftClientMixin overwriting the vanilla logger

Juuxel:
Without `@Unique`, it's treated as some sort of an overwrite for the `LOGGER` field.

I also `@Unique`d the other field just in case vanilla adds that field in the future :smile:


modmuss50 (approved):
LGTM, would be nice to add some verification around this to prevent it from happing in the future. 👍

PR #2727, opened 2022-12-05
add c:shields common tag

itsmiir:
added a `c:shields` tag and added `minecraft:shield` to it. first FAPI pr so let me know if i messed something up 😆


Juuxel (changes requested):
You need to regenerate the tag files. I think the command for that is `gradlew :fabric-convention-tags-v1:generateResources`

itsmiir:
i thought i did that-- looks like the tag files were gitignored. is this all that's needed?

Juuxel (approved):

modmuss50 (approved):

PR #2740, opened 2022-12-11
Tweak Creative Buttons

matthewperiut:
The creative buttons have been moved to the inside of the creative GUI in the 1.19.3 update, and the sprite and positioning don't appropriately reflect this change. The creative buttons also turn blue when hovered, when this design language has been changed in Minecraft.

Below I'm including photos before and after my changes to these buttons.

Before:
before
After:
after

Hovering:
image

Before video:
https://youtu.be/J3wM2r7FR5M

After video:
https://youtu.be/NoKSa3w1-Og

Methodology:
I enjoyed how the buttons were placed before, but due to the recent changes all available space has been used up, so I agree with moving the buttons into the window. The issue I had was how horrendous the outline looked around the buttons. So, I used the item slots and the menu buttons as references to redesign the buttons.

To combat it looking out of place, I did three things.
1. I adopted the style of an item slot, but inverting it to look as if it is extruded out when it can be pressed.
2. I appropriately spaced it between the elements of the creative menu so that no matter what menu was open, (search or player inventory) it looks great.
3. I remove the blue color while hovering in favor of a white outline as 1.14 was the last version that had used this style of selection highlighting. See below.
Screenshot
Screenshot

Feel free to change anything, but I just want to change from what we currently have in any way, thanks!


haykam821:
Could the Programmer Art version of the textures use the old blue background in addition to the default textures using the white outline to indicate focus?

matthewperiut:
> Could the Programmer Art version of the textures use the old blue background in addition to the default textures using the white outline to indicate focus?

I think that that would be a little out of scope for simply having a neat implementation that looks like it belongs there, I think it'd be more appropriate for an external resource pack or mod to handle programmer art in this case

matthewperiut:
> Could the Programmer Art version of the textures use the old blue background in addition to the default textures using the white outline to indicate focus?

I thought about it and decided that it would be a good feature and implemented it in https://github.com/FabricMC/fabric/pull/2740/commits/3d0697ce223fe563bae42255da27d17aaa044be1

Result shown in video:


https://user-images.githubusercontent.com/18296791/206892570-96254c19-32be-4333-aa03-4a95f601cbd6.mp4

Juuxel (changes requested):
We don't use snake case names in code.

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/impl/client/itemgroup/FabricCreativeGuiComponents.java
```diff
@@ -54,12 +56,12 @@ public void render(MatrixStack matrixStack, int mouseX, int mouseY, float float_
 			this.active = extensions.fabric_isButtonEnabled(type);
 
 			if (this.visible) {
-				int u = active && this.isHovered() ? 22 : 0;
-				int v = active ? 0 : 10;
+				int u = active && this.isHovered() ? 20 : 0;
+				int v = active ? ((this.isHovered() && programmer_art) ? 10 : 0) : 10;
```
modmuss50, line 60: I am not a fan of doing the programmer art support like this. It means other texture packs will also have to create a programmer art version. I think its so minor it may be best to ingore it?

matthewperiut:
Superseded by #2742

PR #2742, opened 2022-12-11
Adapt the existing creative inventory page button texture to new position

Juuxel:
Note: this PR does not supersede #2740 or other future PRs like that, it's just a part of 1.19.3 porting. We've debated this texture a long time in the past, so it's still fine to keep using it for now IMO.


modmuss50 (approved):
Yes, this is much better than what we have. I am happy to contiune to discuss future changes to the texture and possibly location/size of these buttons.

PR #2746, opened 2022-12-11
Proposed Change Creative Buttons Texture

matthewperiut:
Before:
image
After:
image


Technici4n:
Not a fan; this reduces the contrast too much IMO. In vanilla, buttons usually have a black outline that turns white when they're hovered.

matthewperiut:
> Not a fan; this reduces the contrast too much IMO. In vanilla, buttons usually have a black outline that turns white when they're hovered.

well, as is the black outline makes so much contrast that it is the only thing I focus on.
With my current change, it makes it so that it is in tune and not out of place with other elements such as the scroll bar and inventory slots.
It is supposed to be relatively bland and unimportant because you only want to see it when you have to use it.
Since they're small elements, I believe the white outline is the right approach, but the black outline is a lot for the area given. If we can maybe give it more space somehow, I think a black outline would be the right approach.

Although arguably, I understand your perspective and agree ideally with the right circumstances a black outline is ideal, visually here it's too small and too cramped to make it so bold.

I'm open to experimentation and suggestion, ultimately I want the UI element to stop sticking out like a sore thumb, and if you have any suggestion to achieve that, I'm willing to attempt it, There isn't a middle ground to the contrast here in my mind, but let me know if you come up with anything.

Technici4n:
I'm not sure everyone will agree on this contrast thing. I for example prefer the current version with the higher contrast.

I think it might just be a user preference thing, and the answer will probably end up being "use a resource pack if you're not happy with the default".

shartte:
I prefer the current version too.

Technici4n:
Going to close this as you can use a resource pack to adjust the buttons depending on your preferences.

PR #2749, opened 2022-12-12
Fix block api resources path

### LINKED ISSUE #2747: Wrong path used in fabric-block-api resources
I don't even know how this was overlooked but the resources path for fabric block api goes like `resources/fabric-block-api-v1`, there is no `assets` in between. Actually `fabric.mod.json` validation shows that it cannot find the icon file but I guess it doesn't cause any issues during runtime. I can make a pr for this probably tomorrow if it's not fixed by then.

ghost:
I opened an  [issue](https://github.com/FabricMC/fabric/issues/2747) about this yesterday. `resources/fabric-block-api-v1` -> `resources/assets/fabric-block-api-v1`


Juuxel (approved):

Technici4n (approved):

ghost:
What's with the running workflow lol
It's been running for more than 2 hours without actually running anything.

modmuss50:
> What's with the running workflow lol It's been running for more than 2 hours without actually running anything.

Not sure :D I have re-ran it.

modmuss50 (approved):

PR #2759, opened 2022-12-14
Extra data providers for remaining vanilla asset types

62832:
As a companion to #2754, this PR aims to provide some new providers for various remaining asset and data types currently present in vanilla Minecraft, such as sounds, particles, worldgen, dimension types and potentially others. Whichever of these types should end up with a dedicated provider can be left open for discussion.

Current list of provider candidates:
- [x] Sounds
- [x] Particles
- [ ] Worldgen
- [ ] Dimension types
- [ ] Structures _(uses .nbt for some reason)_
- [ ] Atlases _(client-only?)_


62832:
Err... I've just found out MC itself already has what is almost an equivalent set of `Sound` and `SoundEntry` classes of its own, so the codec was more or less already there.

As suggested on Discord, this could in fact be condensed quite nicely into some general "CodecDataProvider" which would also work for things like `DimensionType`.

PR #2773, opened 2022-12-20
Add EventSpy to make it easier to test events within a gametest.

modmuss50:
Adds the following API to ease testing events within a game test:

```java
public class ServerBlockEntityEventsGameTest implements FabricGameTest {
	@GameTest(templateName = EMPTY_STRUCTURE)
	public void loadBlockEntity(TestContext context) {
		var spy = context.eventSpy(ServerBlockEntityEvents.BLOCK_ENTITY_LOAD, spyCtx -> (blockEntity, world) -> {
			// Call invoke to increment the call count
			// Returns true when this test is running, false once this test has completed
			// Passing a blockpos to ensure that we only count events triggered from within the test structure
			if (spyCtx.invoke(blockEntity.getPos())) {
				// Test is running can expect inputs
				if (!(blockEntity instanceof FurnaceBlockEntity)) {
					throw new GameTestException("Expected FurnaceBlockEntity but got %s".formatted(blockEntity.getClass().getName()));
				}
			}
		});

		context.setBlockState(BlockPos.ORIGIN.up(), Blocks.FURNACE);

		context.addFinalTask(() -> {
			// Verify that the event was called once while the test is running.
			spy.verifyCalledOnce();
		});
	}
}
```

`FabricTestContext` is interface injected onto `TestContext` adding two new apis: `eventSpy` and `getGameTestState`.

I am a little unsure if this in scope or not, If we add a client testing module I think it may become more useful, maybe hold off until that happens? Let me know what you think.

This does create a new event listener for each test, technically leaking, it could possibly be reworked to have a static instance of `EventSpy` that can be reused between tests? 🤔 

Another option would be to keep this internal to fabric API (similar to the existing client tests) as I expect it might be seldom used by other mods?

### DIFF
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/EventSpy.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/EventSpy.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


modmuss50:
Going to close this for now, after thinking about it for a bit this doesnt solve the biggest problems with registering an event listener in the test and likely wont be used by other mods outside of fabric api.

PR #2774, opened 2022-12-21
Fix interaction events

### LINKED ISSUE #1260: Possible nullability error in UseEntityCallback
[`UseEntityCallback`](https://github.com/FabricMC/fabric/blob/0d4d7e34589be55835349b26cacc9fdfe05170b6/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/UseEntityCallback.java#L55)'s handler signature has a `@Nullable` parameter, `hitResult`, but this parameter should never be null. The server invoker for this event instantiates the `EntityHitResult` directly, while the client invoker uses `MinecraftClient::crosshairTarget` (from `MinecraftClient::doItemUse`). This field is indeed nullable, but it is only null when not in a world. In a world, this will either be a successful raycast to whatever the client is looking at, or a "miss" result, which is not null.

Alternatively, the client invoker unconditionally uses the variable that gets passed to `hitResult`, which is supposedly nullable (see [here](https://github.com/FabricMC/fabric/blob/0d4d7e34589be55835349b26cacc9fdfe05170b6/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinClientPlayerInteractionManager.java#L135)). Either this is an error, or the event handler parameter should not be marked as `@Nullable`.

### LINKED ISSUE #1870: fabric-events-interaction-v0: UseEntityCallback cancelling not completely cancelling interaction
Hey there!

It looks like subscribing to events using the `UseEntityCallback.EVENT` callback and returning a `ActionResult.FAIL` doesn't actually prevent the entire entity interaction. The mixin seems to only be listening to/affecting the `interact at` interaction type and not the `interact` type, which the Minecraft client will fall back to if the former fails on the client side (see the `ENTITY` handling block in `MinecraftClient#doItemUse` (yarn mappings)) - and cancelling this fallback too works as expected.

Code to reproduce (just right-click a villager and you'll see the UI still opens)
```java
UseEntityCallback.EVENT.register((player, world, hand, entity, hitResult) -> {
    return ActionResult.FAIL;
});
```

I'd love to see this fixed as I'm writing a mod that involves cancelling entity interactions, so I've had to do the double injection to work around this!

----

To fix this, I see two ways of doing it:

1. Adding another injection point at the start of the `interact` method in the same class as the `interactAt` injection point. This has the positive of being simple, but the negative of meaning each handler is called _twice_, which may be a deal-breaker – or will at least need a way of differentiating the two interaction types.

2. Lifting the injection point up one layer so that it covers both `interact` and `interactAt` in one go, so a `FAIL` result skips both `interact at` and `interact` attempts. I believe this would be the preferred option as each listener would be called only once and would be able to truly allow/deny the entire entity interaction.

----

Callback: https://github.com/FabricMC/fabric/blob/1.18/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/UseEntityCallback.java#L41

Mixin: https://github.com/FabricMC/fabric/blob/1.18/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/MixinServerPlayNetworkHandler.java#L45-L57

Technici4n:
I was only looking for a fix for `UseEntityCallback`, but I also noticed other issues. Please look at each commit separately for a better view of what I changed.

- Deprecate `BlockAttackInteractionAware`. The `boolean` return type is simply not enough to describe what should happen (swing hand or not? send packet to server or not?). The event should be used instead.
- Picking fixes.
  - Document `BlockPickInteractionAware` and `EntityPickInteractionAware`.
  - Remove unneeded `@Nullable` annotations.
  - Make sure that `BlockPickInteractionAware` is not called when the hit result missed (i.e. when the block is out of range).
- `UseEntityCallback` fixes.
  - Fix #1870 by moving the injection point on the client, and adding a second injection point on the server. This means that in a specific case the hit result can be `null`. So the `@Nullable` is actually required, which resolves #1260.
  - Add `ActionResult.CONSUME` as a valid return type - mimicking the new behavior of `AttackBlockCallback`.
  - Improve documentation.

### DIFF
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/ClientPlayerInteractionManagerMixin.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/ClientPlayerInteractionManagerMixin.java
@@ -43,16 +43,13 @@ import net.minecraft.util.ActionResult;
 import net.minecraft.util.Hand;
 import net.minecraft.util.TypedActionResult;
 import net.minecraft.util.hit.BlockHitResult;
-import net.minecraft.util.hit.EntityHitResult;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Direction;
-import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.GameMode;
 
 import net.fabricmc.fabric.api.event.player.AttackBlockCallback;
 import net.fabricmc.fabric.api.event.player.AttackEntityCallback;
 import net.fabricmc.fabric.api.event.player.UseBlockCallback;
-import net.fabricmc.fabric.api.event.player.UseEntityCallback;
 import net.fabricmc.fabric.api.event.player.UseItemCallback;
 
 @Mixin(ClientPlayerInteractionManager.class)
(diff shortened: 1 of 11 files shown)


Shnupbups:
Would it be worth simply deprecating the method in `BlockAttackInteractionAware` that returns a boolean in favour of adding a new one that returns an `ActionResult`, instead of deprecating the whole interface? Though I suppose then you'd also have to make sure the new `ActionResult` one had a default impl for backwards compatibility reasons... I dunno, I suppose the event is better anyway.

Technici4n:
We could do that, but honestly I don't think this interface is not _that_ useful. The fact that we didn't receive any issue report despite the limited return value options means that it's not used in practice for what I would expect it to be used...

Technici4n:
Quick rebase to fix merge conflicts.

apple502j (approved):

Juuxel (approved):

modmuss50 (approved):

PR #2775, opened 2022-12-21
Indigo and Renderer API fixes related to fallback consumers

### LINKED ISSUE #1871: Indigo Implementation looses block state data in its fallback consumer.
## How the flaw can be encountered?

In my mod Aurora's Decorations, I tried to implement [sign posts](https://lambdaurora.dev/AurorasDecorations/sign_posts.html).

For reference: sign post

The goal is simple: put a directional sign on a fence block, this can be achieved through the use of entities, but that can create some issues, or by making a custom block, which is what I chose.
The goal of the custom block is simple: mimic a fence block with less properties than the Vanilla block, but renders the same, and as a bonus, every registered fence blocks must have a sign post variant.
To achieve this it must mimics the look too, so I made use of FRAPI and BakedModels, the BakedModel is simple, use the blockstate of the fence it needs to mimic and its baked model, then emits those quads instead.

Now the trouble begins: fence blocks are a multipart model, which led to crashes when trying to emit the quads.

The emitBlockQuads is simple:
```java
	@Override
	public void emitBlockQuads(BlockRenderView blockView, BlockState state, BlockPos pos, Supplier<Random> randomSupplier,
	                           RenderContext context) {
		if (state.getBlock() instanceof SignPostBlock signPostBlock) {
			((FabricBakedModel) this.wrapped).emitBlockQuads(blockView, signPostBlock.getFenceState(state), pos, randomSupplier, context);
		}
	}
```

I tell the model loader to use my baked model and I have the fence block model as a dependency, so I emit it and give it the proper fence block state instead of my block state. And you have to note that a multipart model will do block state comparison.

Due to an oversight in Indigo, it will crash due to missing properties in my block state.

## Cause of the problem

The fence block model being a Vanilla model, it will use the `fallbackConsumer`, and here's the issue with it:
[TerrainFallbackConsumer.java - line 87](https://github.com/FabricMC/fabric/blob/1.18/fabric-renderer-indigo/src/main/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/TerrainFallbackConsumer.java#L87)

As you can see, in the line `final BlockState blockState = blockInfo.blockState;`, it ignores entirely the block state I passed as parameter and use the block state of the block placed in-world!

and that's how the crash happens, making it use the block state passed as parameter could fix this issue entirely.

## Notes

This flaw doesn't only affect this mod, actually it is one of the root cause of another defect in LambdaBetterGrass better snow feature, when using Canvas or Iris, due to materials being tied to block states, the snow model (which is emitted in a similar manner) will go through the fallback consumer, and loose all block state data, which means the renderer will then apply the same shaders as the rest of the block, which will lead to wavy snow. How is this relevant to the issue? Indium and Canvas used Indigo as a base of implementation, so this oversight has been copied over, and led to the same issues.

Fixing this in Indigo would prevent the crash encountered above, and prevent other implementations to copy over the issue.

### LINKED ISSUE #2639: Indigo fallback consumer does not respect BlendMode or emissivity
The parts of code that handled the `BlendMode` and emissivity were likely left out from `TerrainFallbackConsumer` because vanilla models cannot use them, but `QuadTransform`s can still change the quad's material and thus `BlendMode` and emissivity. The fallback consumer will ignore these properties even though it should not. `AbstractMeshConsumer`'s code correctly handles these properties, so this code should likely be abstracted and used by the `TerrainFallbackConsumer` as well to fix this issue.

Technici4n:
- Fix #2639: Rewire fallback consumers a bit so they go through the same logic as the other ways to emit a mesh.
- Fix #1871: Allow passing the block state explicitly to the fallback consumer by adding a `bakedModelConsumer`.
- Expand testmod to test these behaviors.

### DIFF
--- a/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
+++ b/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
@@ -16,7 +16,6 @@
 
 package net.fabricmc.fabric.api.renderer.v1.render;
 
-import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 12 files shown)


Technici4n:
Changed the testmod a bit to test this.

Before the first commit: (tried by reversing the first commit after having written the test)
image

After the first commit:
image

fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
```diff
@@ -42,9 +46,33 @@ public interface RenderContext {
 	 * Fabric causes vanilla baked models to send themselves
 	 * via this interface. Can also be used by compound models that contain a mix
 	 * of vanilla baked models, packaged quads and/or dynamic elements.
+	 *
+	 * <p>For block contexts, this will use the block state of the block being rendered to get the quads.
+	 * For item contexts, this will pass a {@code null} block state to {@link BakedModel#getQuads}.
+	 * {@link #blockFallbackConsumer()} can be used instead to pass the block state explicitly.
```
shartte: What does "will use the block state of the block ... to get the quads"?  It sounds weirdly worded.  It hopefully calls getQuads all the same, the difference is just what it passes to getQuads as the blockstate, right?

modmuss50 (approved):
Seems OK to me, however rendering stuff isnt really forte, reviews from others who are more familar with this would be applicated 👍

hYdos (approved):

Technici4n:
I think it would be nicer to move the `blockFallbackConsumer` to a new `BakedModelConsumer` interface - in case we need more extensions in the future.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/TerrainFallbackConsumer.java
```diff
@@ -102,29 +108,6 @@ private void renderQuad(BakedQuad quad, Direction cullFace, Value defaultMateria
...
-		cullFace = editorQuad.cullFace();
-
-		if (cullFace != null && !blockInfo.shouldDrawFace(cullFace)) {
-			return;
-		}
-
-		if (!editorQuad.material().disableAo(0)) {
-			// needs to happen before offsets are applied
-			aoCalc.compute(editorQuad, true);
-			tessellateSmooth(editorQuad, blockInfo.defaultLayer, editorQuad.colorIndex());
```
Technici4n, line 118: This messes up the render layer I think... Damn it

Technici4n: Nevermind it works fine because the default blendmode defers to `blockInfo.defaultLayer` anyway.

PR #2777, opened 2022-12-22
Allow passing the block state explicitly to the fallback consumer. Fixes #1871

### LINKED ISSUE #1871: Indigo Implementation looses block state data in its fallback consumer.

Technici4n:
Issue reproduced and fixed using the testmod from #2775.

The fix is rather straightforward: we need to pass the `BlockState` explicitly. There is a new API addition, but it is backwards-compatible.

### DIFF
--- a/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
+++ b/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
@@ -16,8 +16,10 @@
 
 package net.fabricmc.fabric.api.renderer.v1.render;
 
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
+import net.minecraft.block.BlockState;
 import net.minecraft.client.render.model.BakedModel;
 
 import net.fabricmc.fabric.api.renderer.v1.mesh.Mesh;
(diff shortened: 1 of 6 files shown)


Technici4n:
Actually I'll put this in #2775 because I found more issues, and they are somewhat related.

PR #2779, opened 2022-12-23
Fix #602: Allow querying registered `DynamicItemRenderer`s

### LINKED ISSUE #602: Multiple builtin item renderers for one item?
The builtin item renderer API (#563) also allows replacing the builtin renderers of vanilla items (intended feature), but there will be a crash if multiple mods want to replace the renderer of one item.

Should we make the implementation use a `Multimap` instead, and render the item with all registered builtin renderers?

Technici4n:
Fixes #602.


Technici4n:
Looks like a spurious build failure.

modmuss50 (approved):

PR #2780, opened 2022-12-23
Add more transitive Access Wideners

### LINKED ISSUE #474: Access to creating StateProviderTypes
StateProviderType.register is private

### LINKED ISSUE #981: Hook to expose DefaultBiomeCreator.getSkyColor
All (AFAIK) vanilla overworld/nether biomes have their sky colour assigned with this private static method, so exposing it would be useful for modded biomes. (It takes the biome temperature and calculates a sky colour from it.)

### LINKED ISSUE #1635: [1.17] Creating a custom RenderLayer is high-friction

Technici4n:
- Expose `StateProviderType` constructor. Fixes #474.
- Expose DefaultBiomeCreator.getSkyColor. Fixes #981.
- Make custom `RenderLayer` registration easier. Fixes #1635.
- Add warning at the beginning of the generated AW to prevent manual editing of the file.

### DIFF
--- a/fabric-data-generation-api-v1/build.gradle
+++ b/fabric-data-generation-api-v1/build.gradle
@@ -63,7 +63,12 @@ import java.util.zip.ZipFile
 task generateAccessWidener() {
 	doLast {
 		File inputJar = loom.namedMinecraftProvider.parentMinecraftProvider.commonJar.toFile()
-		String accessWidener = file("template.accesswidener").text + "\n"
+		String accessWidener = "accessWidener\tv2\tnamed\n"
+		accessWidener += "\n"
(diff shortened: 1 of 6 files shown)


Shnupbups (approved):
Maybe also add the warning to the Data Generation module's access widener, as it also uses a template.

modmuss50 (approved):

PR #2783, opened 2022-12-23
Block Entity Lifecycle fixes

### LINKED ISSUE #1821: The client chunk unload may not fire when the load distance is decreased
Check `ClientChunkManager#updateLoadDistance`. We don't seem to be injecting there to catch unloads. To be confirmed though.

### LINKED ISSUE #2423: BlockEntity Lifecycle Events Don't Trigger Intuitively
BlockEntity UNLOAD don't seem to appear to trigger when a chunk exits render distance, nor do they trigger when the player exits the dimension while they are outside of render distance.

Notably, they do fire the UNLOAD event if the player changes dimensions while they are in render distance.

Secondly, the client LOAD event triggers before the NBT is read from the packet the server sends the client. This can be an issue, if the load handler needs to do something using data that could be in the NBT.

Technici4n:
- Fire client chunk unload event when the load distance is decreased. Fixes #1821.
- Fire client chunk unload event before entities are cleared from the chunk to fix block entity unload event not firing. Clarify that block entity data is unreliable in the block entity load events. Fixes #2423.

### DIFF
--- a/fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java
+++ b/fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java
@@ -55,8 +55,22 @@ public abstract class ClientChunkManagerMixin {
 		}
 	}
 
-	@Inject(method = "unload", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/world/ClientChunkManager$ClientChunkMap;compareAndSet(ILnet/minecraft/world/chunk/WorldChunk;Lnet/minecraft/world/chunk/WorldChunk;)Lnet/minecraft/world/chunk/WorldChunk;", shift = At.Shift.AFTER), locals = LocalCapture.CAPTURE_FAILEXCEPTION)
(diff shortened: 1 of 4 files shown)


fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java
```diff
@@ -55,8 +55,22 @@ private void onChunkUnload(int x, int z, PacketByteBuf buf, NbtCompound tag, Con
 		}
 	}
 
-	@Inject(method = "unload", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/world/ClientChunkManager$ClientChunkMap;compareAndSet(ILnet/minecraft/world/chunk/WorldChunk;Lnet/minecraft/world/chunk/WorldChunk;)Lnet/minecraft/world/chunk/WorldChunk;", shift = At.Shift.AFTER), locals = LocalCapture.CAPTURE_FAILEXCEPTION)
+	@Inject(method = "unload", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/world/ClientChunkManager$ClientChunkMap;compareAndSet(ILnet/minecraft/world/chunk/WorldChunk;Lnet/minecraft/world/chunk/WorldChunk;)Lnet/minecraft/world/chunk/WorldChunk;"), locals = LocalCapture.CAPTURE_FAILEXCEPTION)
```
Technici4n, line 58: Note: this removes the shift because `compareAndSet` (the targeted method) clears the block entities, and we use the block entity list to implement the block entity unload event.

modmuss50 (approved):

PR #2784, opened 2022-12-23
Don't break the ordering of ResourcePackProviders

### LINKED ISSUE #1671: Don't break the ordering of ResourcePackProviders
What:
The mixin that adds datapacks for mods breaks the ordering of resource pack providers for both datapacks and resourcepacks.

https://github.com/FabricMC/fabric/blob/62e33f4127784665f90b9f3ee54a39559ba113a1/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackManagerMixin.java#L48

In the vanilla code, it uses this constructor to make an ImmutableSet from the array:
https://guava.dev/releases/19.0/api/docs/src-html/com/google/common/collect/ImmutableSet.html#line.297
This is documented to maintain the order.

The mixin above copies that set into a HashSet effectively randomising the order.

While the main ordering comes from the first/last ordering specified by the resource pack provider,
this does not help when two different providers use the same value. It is the above set's ordering that determines.

Proposed fix:

Change the mixin to use a LinkedHashSet which will maintain the insertion ordering.

My current workaround is to add a Mixin that runs after the Fabric mixin and does just that.

Off Topic: There are a number of mods that rely on the Fabric API making the providers field mutable.
It might be an idea to make it an official api?
e.g.
https://github.com/architectury/architectury-api/blob/1.17/common/src/main/java/dev/architectury/hooks/PackRepositoryHooks.java

Technici4n:
Fixes #1671.

Honestly I wasn't sure how to verify this one, but the fix seems trivial enough?

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackManagerMixin.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackManagerMixin.java
@@ -16,7 +16,7 @@
 
 package net.fabricmc.fabric.mixin.resource.loader;
 
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 import org.spongepowered.asm.mixin.Final;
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Makes sense 👍

PR #2789, opened 2022-12-24
Biome modification API small enhancements 

### LINKED ISSUE #2079: [1.18.2][FAPI 48.0] BiomeModificationContextImpl.addCarver fails for Biomes without initial carvers
Adding a new Carver assumes, that the `generationSettings` of a Biome have a valid `RegistryEntryList` for the given `GenerationStep`.
https://github.com/FabricMC/fabric/blob/08b1b79b17d800aed105d8eba85a270a0876d355/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/modification/BiomeModificationContextImpl.java#L308

However the vanilla Biome Builder will initialize Biomes with an empty map, which will cause a NullPointerException in the `plus`-Method:
https://github.com/FabricMC/fabric/blob/08b1b79b17d800aed105d8eba85a270a0876d355/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/modification/BiomeModificationContextImpl.java#L325

A possible solution might be to amend `values.stream().toList()` with `values==null?Lists.newArrayList():values.stream().toList()` or something similar.

apple502j:
See each commit's message for description. Resolves #2079

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModificationContext.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModificationContext.java
@@ -17,6 +17,7 @@
 package net.fabricmc.fabric.api.biome.v1;
 
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.function.BiPredicate;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


Technici4n (changes requested):
Should contain a fix for #2079 too if it's still an issue.

Technici4n (approved):
Perfect

modmuss50 (approved):

PR #2795, opened 2022-12-26
Autogenerate Item constructor TAWs

Shnupbups:
`generateAccessWidener` task in TAW module will now automatically generate TAWs for the constructors of non-abstract item classes, similar to how it already did so for blocks. As such, previously manually added item constructor TAWs have been removed from the template. This also means that `HoeItem` is now TAW'd. (I totally overlooked it earlier... whoops!)

### DIFF
--- a/fabric-transitive-access-wideners-v1/build.gradle
+++ b/fabric-transitive-access-wideners-v1/build.gradle
@@ -37,6 +37,8 @@ task generateAccessWidener {
 		try (def fs = FileSystems.newFileSystem(URI.create("jar:${commonJar.toUri()}"), [create: false])) {
 			generateBlockConstructors(lines, fs)
 			lines.add("")
+			generateItemConstructors(lines, fs)
+			lines.add("")
 		}
 
 		Path clientJar = loom.namedMinecraftProvider.parentMinecraftProvider.clientOnlyJar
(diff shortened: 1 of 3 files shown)


Noaaan (approved):
Nice

Juuxel (approved):
LGTM. This is already in use with the exact same logic in [Architectury API](https://github.com/architectury/architectury-api/blob/d979caa4ab7bb4980f673f50b9c00aea2349b738/common/build.gradle#L89) (mojmap warning!) too.

Technici4n (changes requested):
Please also add previously missed TAWs. For example, https://github.com/FabricMC/fabric/blob/1.19.3/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/Screens.java should be deprecated and replaced by TAWs. I believe it would be worth going through all the accessor mixins and checking for similar ones.

Shnupbups:
> Please also add previously missed TAWs. For example, https://github.com/FabricMC/fabric/blob/1.19.3/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/Screens.java should be deprecated and replaced by TAWs. I believe it would be worth going through all the accessor mixins and checking for similar ones.

Eh, I'd say that is out of scope for this simple PR and should be a follow up.

Technici4n:
I think it's reasonable to many missed TAWs in a single pass, see also #2780 where I addressed multiple issues at once.

Noaaan:
Would adding a TAW for `StatusEffect` also be suitable to do in this pass? This constructor is also protected.

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -75,6 +77,32 @@ def generateBlockConstructors(List<String> lines, FileSystem fs) {
 		}
 }
 
+def generateItemConstructors(List<String> lines, FileSystem fs) {
+	lines.add("# Constructors of non-abstract item classes")
+	Files.list(fs.getPath("net/minecraft/item"))
```
modmuss50, line 82: This is fine for now, in the future it may be better to check the class hierarchy similar to how I did it in the data gen module.

modmuss50 (approved):

Issue #1099, opened 2020-09-24
Support KeyBinding with modifier on KeyBindingHelper


MPThLee:
I just working a extension but I want to use with modifier(ex. *CTRL* + C). through `KeyBindingHelper`

Forge supports it as mixed-in method in `net.minecraft.client.options.KeyBinding`

i509VCB:
Edited to remove MCP names.

Issue #58, opened 2019-01-04
Allow multiple keybindings in non-competing contexts


asiekierka:
After https://github.com/FabricMC/fabric/pull/57

haykam821:
I assume the API surface for this feature would be something along the lines of:

```java
public static final KeyBindingGroup SCREEN = new KeyBindingGroup();
public static final KeyBindingGroup WORLD = new KeyBindingGroup();
public static final KeyBindingGroup INVENTORY = new KeyBindingGroup();

public static final KeyBinding ACTIVATE_BATTERY_ITEM_KEY = /* ... */;
public static final KeyBinding SUMMON_MOB_KEY = /* ... */;

static {
	SCREEN.add(ACTIVATE_BATTERY_ITEM_KEY);
	WORLD.add(SUMMON_MOB_KEY);
}
```

If two key bindings are part of different groups, then their key conflicts will not be marked.

Some questions that need answered include:

- How will vanilla key bindings be assigned? Some key bindings in vanilla show as conflicting despite only being usable in different contexts.
- What groups will Fabric API provide by default, if any?
- Can key bindings be part of multiple groups?
- Will groups have inheritance? For example, a more restricted group used when the player is holding a specific item should conflict with key bindings from the general holding-an-item group, but not with key bindings from another holding-a-specific-item group.

apple502j:
Implemented in vanilla.

PR #2799, opened 2022-12-27
Keybindings API: New Years cleanup

apple502j:
Code cleanups and fixes null being thrown.

### DIFF
--- a/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
+++ b/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
@@ -16,6 +16,8 @@
 
 package net.fabricmc.fabric.api.client.keybinding.v1;
 
+import java.util.Objects;
+
 import net.minecraft.client.option.KeyBinding;
 import net.minecraft.client.util.InputUtil;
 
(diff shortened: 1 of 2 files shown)


fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
```diff
@@ -41,8 +44,11 @@ private KeyBindingHelper() {
 	 *
 	 * @param keyBinding the keybinding
 	 * @return the keybinding itself
+	 * @throws NullPointerException when {@code null} is passed
```
Technici4n: Not sure I agree with adding this to almost every API method. I feel like `@throws` should only be documented if it's something the API user should be aware of.

Technici4n:
If we're gonna look into keybindings, it might be nice to evaluate solutions to #58 and #1099.

Technici4n (approved):

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
Juuxel: IMO it doesn't hurt to have, and even the JDK does it. Not suggesting to force it though

modmuss50 (approved):

modmuss50 (approved):

PR #2800, opened 2022-12-27
Replace useless nullability comments with jetbrains annotation

SilverAndro:
This broke my automated tooling which relies on knowing nullability (specifically the block break events fucked it up). I very much hope you aren't planning to add more of these into the codebase because these are tremendously annoying and I assumed it was some "smart view" mode in my IDE hiding the actual annotation.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
@@ -18,12 +18,12 @@ package net.fabricmc.fabric.impl.item;
 
 import java.util.WeakHashMap;
 
+import org.jetbrains.annotations.Nullable;
+
 import net.minecraft.item.Item;
 
 import net.fabricmc.fabric.api.item.v1.CustomDamageHandler;
 import net.fabricmc.fabric.api.item.v1.EquipmentSlotProvider;
-
-import org.jetbrains.annotations.Nullable;
 
 public final class FabricItemInternals {
 	private static final WeakHashMap<Item.Settings, ExtraData> extraData = new WeakHashMap<>();
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
@@ -45,8 +45,8 @@ public final class FabricItemInternals {
 	}
 
 	public static final class ExtraData {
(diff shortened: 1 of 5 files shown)


apple502j:
Are they all nullable, or in fact not nullable and commented out for reasons?

SilverAndro:
I know the block ones are nullable at least for sure, since thats what exploded at runtime

SilverAndro:
Item ones were old comments, I have no clue on the entity builder because theres both a requireNonNull and a "if not null", which seems to imply it can be null but only while building? Theres 0 hints at the intent here
image

SilverAndro:
Going to assume its supposed to be nullable internally for building, since
- The field is private, meaning its not public API
- The setter requires the passed value to not be null (only place its interactable in public API)
- The build method makes a check if its not null, because it can be null if the public api consumer doesn't call the set method

Since reading isnt exposed anywhere, it doesn't really matter since it has 0 impact on public API, so the extra safety for maintainability makes sense

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ItemExtensions.java
```diff
@@ -19,9 +19,11 @@
 import net.fabricmc.fabric.api.item.v1.CustomDamageHandler;
 import net.fabricmc.fabric.api.item.v1.EquipmentSlotProvider;
 
+import org.jetbrains.annotations.Nullable;
+
 public interface ItemExtensions {
-	/* @Nullable */ EquipmentSlotProvider fabric_getEquipmentSlotProvider();
+	EquipmentSlotProvider fabric_getEquipmentSlotProvider();
```
apple502j: Now that I think about this, it is probably nullable. (`ItemMixin` is missing annotation.)

apple502j (changes requested):
Could you run checkstyle and fix the import sorting issues?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemInternals.java
```diff
@@ -43,8 +45,8 @@ public static void onBuild(Item.Settings settings, Item item) {
 	}
 
 	public static final class ExtraData {
-		private /* @Nullable */ EquipmentSlotProvider equipmentSlotProvider;
-		private /* @Nullable */ CustomDamageHandler customDamageHandler;
+		private EquipmentSlotProvider equipmentSlotProvider;
```
apple502j: These too are nullable, I believe this was coded back when nullable annotations were not used?

apple502j (approved):

modmuss50 (approved):

Issue #2812, opened 2023-01-01
Cannot execute command: Cannot invoke CommandDispatcher.execute


apple502j:
This is an issue reported several times by multiple users; see #2631 and #2809.

## Summary
Commands cannot be executed. The error is:`Cannot invoke ''com.mojang.brigadier.CommandDispatcher.execute(String,Object)''because
''net.fabricmc.fabric.impl.command.client.ClientCommandInternals.activeDispatcher'' is null`.

## Cause
Most, if not all, of the reports indicate they are on a server. The cause is currently being investigated.

## Fix
WIP. We haven't been able to reproduce this; I have a speculative fix but I am not sure if folks here are happy with merging it.

haykam821:
This issue can be reproduced by registering a `ClientPlayConnectionEvents.JOIN` listener that throws an exception, then connecting to a server and attempting to run a command:

```java
ClientCommandRegistrationCallback.EVENT.register((dispatcher, registryAccess) -> {
	dispatcher.register(ClientCommandManager.literal("pling").executes(context -> {
		return Command.SINGLE_SUCCESS;
	}));
});

ClientPlayConnectionEvents.JOIN.register((handler, sender, client) -> {
	throw new IllegalStateException();
});
```

Note that this issue cannot be reproduced in singleplayer as the client will just crash. Connecting to a server prints the exception to the log instead, leaving the client command dispatcher in an invalid uninitialized state.

Technici4n:
Fixed by #2813.

PR #2813, opened 2023-01-01
Catch exception inside client-side JOIN event

apple502j:
See #2812
Crashes inside `ClientPlayConnectionEvents.JOIN` previously interrupted the normal join cycle without logging exceptions. This now adds the logging.

Server-side event is not in this PR; I believe it is covered by vanilla logging.

Requesting backport to 1.19.2.

### DIFF
--- a/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java
+++ b/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java
@@ -69,7 +69,7 @@ public final class ClientPlayNetworkAddon extends AbstractChanneledNetworkAddon<
 		try {
 			ClientPlayConnectionEvents.JOIN.invoker().onPlayReady(this.handler, this, this.client);
 		} catch (RuntimeException e) {
-			LOGGER.error("Exception raised while invoking ClientPlayConnectionEvents.JOIN", e);
+			LOGGER.error("Exception thrown while invoking ClientPlayConnectionEvents.JOIN", e);
 		}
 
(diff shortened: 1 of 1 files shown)


Technici4n (approved):

Player3324 (approved):

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java
```diff
@@ -39,6 +42,8 @@ public final class ClientPlayNetworkAddon extends AbstractChanneledNetworkAddon<
 	private final MinecraftClient client;
 	private boolean sentInitialRegisterPacket;
 
+	private static final Logger LOGGER = LogUtils.getLogger();
```
haykam821, line 45: This logger could use `LoggerFactory.getLogger`, with reasoning in #2814.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java
```diff
@@ -61,7 +66,11 @@ public void lateInit() {
 	}
 
 	public void onServerReady() {
-		ClientPlayConnectionEvents.JOIN.invoker().onPlayReady(this.handler, this, this.client);
+		try {
+			ClientPlayConnectionEvents.JOIN.invoker().onPlayReady(this.handler, this, this.client);
+		} catch (RuntimeException e) {
+			LOGGER.error("Exception raised while invoking ClientPlayConnectionEvents.JOIN", e);
```
haykam821:
```suggestion
			LOGGER.error("Exception thrown while invoking ClientPlayConnectionEvents.JOIN", e);
```

modmuss50 (approved):

PR #2818, opened 2023-01-02
Two Transfer API fixes

### LINKED ISSUE #2522: (Request) Behavior change to FluidStorageUtil#interactWithFluidStorage
The javadocs say `In creative mode, the player's inventory will not be modified` and yes, that is true, but it works a bit too well. I believe the inventory shouldn't be modified when filling, but should when emptying a storage. 

If you want to fill a tank on creative, you want it to be quick and easy, so it makes sense to not remove the fluid from your bucket. 
But when emptying your tank, it might confuse the player as to why they are not getting the fluid on their bucket, despite the tank's content decreasing. And even if the player don't become confused by this, they might want to quickly obtain the tank's fluid on a bucket, similar to how a middle click works on creative.

### LINKED ISSUE #2810: Double chest TRAPI wrapper does not always update both blocks
We only call `markDirty` on the changed half. See also https://github.com/Fabricators-of-Create/Create/issues/724.

Technici4n:
Fixes #2810 by updating the second half of chests. Comes with a gametest structure. Whether the fix will cause performance issues is yet TBD - if that happens I have some optimization ideas.

Fixes #2522 by making the behavior of `ContainerItemContext.forPlayerInteraction` match that of `ItemUsage.exchangeStack`, more or less.


modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/InitialContentsContainerItemContext.java
```diff
@@ -37,6 +37,22 @@ protected ItemVariant getBlankVariant() {
...
+			return 0;
+		}
+
+		@Override
+		public long extract(ItemVariant extractedVariant, long maxAmount, TransactionContext transaction) {
+			StoragePreconditions.notBlankNotNegative(extractedVariant, maxAmount);
+
+			// Pretend we can extract anything, but never actually do it.
+			return maxAmount;
+		}
```
Juuxel: Hmm, isn't this a change to `ContainerItemContext.withInitial` behaviour as well?

Technici4n: That's true, but we're still within its (somewhat vague) contract:

```java
 	/**
	 * Return a context that can accept anything, and will accept (and destroy) any overflow items, with some initial variant and amount.
	 * This can typically be used to check if a variant provides an API, or simulate operations on the returned API,
	 * for example to simulate how much fluid could be extracted from the variant and amount.
	 */
```

It depends whether you read `with some initial variant and amount` as "initially the context has this content, and whatever happens after that is undefined", or as "the context always has the initial content whatever happens". I think the latter makes more sense.

Juuxel: To me "initial" implies that there can be other content that isn't "initial", otherwise it's pointless to have that word there 😛

IMO the new behaviour would be something like `withConstant`

Technici4n: Hmmm, what would happen to `withInitial` after mutating the stack would still not be defined. How about deprecating `withInitial` in favor of `withConstant` ? (Both with the new behavior)

Technici4n: Went for it, but keeping the old behavior for `withConstant` just in case.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
```diff
@@ -24,6 +24,7 @@
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.player.PlayerInventory;
 import net.minecraft.item.ItemStack;
+import net.minecraft.item.ItemUsage;
```
Juuxel, line 27: Hmm, do we generally do javadoc-only imports? Feels a bit weird to me to only reference imported classes in comments

Juuxel (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/context/ContainerItemContext.java
Technici4n: I do it to avoid rendering the fully qualified name: (and it's done throughout the transfer API already)
image

PR #2820, opened 2023-01-03
Networking: introduce packet-object based API

apple502j:
See the javadoc for designs. Huge thanks to @Technici4n and @Sollace.

## FAQ
- **Is this a breaking change?** Nope.
- **Is the old API deprecated?** Nope.


Rasmion:
Perhaps this should be extended not only to play networking, but also to the login?

apple502j:
@RasmusHD Nope, in fact these APIs need a huge rework. I don't think login stuff would live very long.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PacketType.java
```diff
@@ -0,0 +1,73 @@
...
+	/**
+	 * Returns the identifier of the channel used to send the packet.
+	 * @return the identifier of the associated channel.
+	 */
+	public Identifier getId() {
+		return id;
+	}
+
+	@ApiStatus.Internal
+	public T read(PacketByteBuf buf) {
```
Juuxel, line 64: Why is this internal? I don't see any harm in exposing this.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +403,44 @@ public interface PlayChannelHandler {
 		 */
 		void receive(MinecraftClient client, ClientPlayNetworkHandler handler, PacketByteBuf buf, PacketSender responseSender);
 	}
+
+	/**
+	 * An internal packet handler that works as a proxy between old and new API.
+	 * @param <T> the type of the packet
+	 */
+	interface PlayChannelHandlerProxy<T extends FabricPacket> extends PlayChannelHandler {
```
Juuxel:
```suggestion
	private interface PlayChannelHandlerProxy<T extends FabricPacket> extends PlayChannelHandler {
```
I think this can be private?

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +403,44 @@ public interface PlayChannelHandler {
...
+	@FunctionalInterface
+	public interface PlayPacketHandler<T extends FabricPacket> {
+		/**
+		 * Handles the incoming packet. This is called on the render thread, and can safely
+		 * call rendering-related methods.
+		 *
+		 * <p>An example usage of this is to display an overlay message:
+		 * <pre>{@code
+		 * // See FabricPacket for creating the packet
+		 * ClientPlayNetworking.registerReceiver(OVERLAY_PACKET_TYPE, (player, packet, responseSender) -&rt; {
```
Juuxel:
```suggestion
		 * ClientPlayNetworking.registerReceiver(OVERLAY_PACKET_TYPE, (player, packet, responseSender) -> {
```
1. I think it should be `&gt;` in the first place
2. `{@code}` escapes lt and gt anyway

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +403,44 @@ public interface PlayChannelHandler {
...
+		 * <p>The network handler can be accessed via {@link ClientPlayerEntity#networkHandler}.
+		 *
+		 * @param player the player that received the packet
+		 * @param packet the packet
+		 * @param responseSender the packet sender
+		 * @see FabricPacket
+		 */
+		void receive(MinecraftClient client, ClientPlayerEntity player, T packet, PacketSender responseSender);
+
+		default boolean processOnNetworkThread(MinecraftClient client, ClientPlayerEntity player, T packet, PacketSender responseSender) {
```
Juuxel: Should have jd

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -197,6 +348,20 @@ public static boolean canSend(ServerPlayNetworkHandler handler, Identifier chann
 		return ServerNetworkingImpl.getAddon(handler).getSendableChannels().contains(channelName);
 	}
 
+	/**
+	 * Checks if the connected client declared the ability to receive a specific type of packet.
+	 *
+	 * @param handler the network handler
+	 * @param type the packet type
+	 * @return True if the connected client has declared the ability to receive a specific type of packet
```
Juuxel:
```suggestion
	 * @return {@code true} if the connected client has declared the ability to receive a specific type of packet
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -293,4 +475,56 @@ public interface PlayChannelHandler {
...
+	@FunctionalInterface
+	public interface PlayPacketHandler<T extends FabricPacket> {
+		/**
+		 * Handles the incoming packet. This is called on the server thread, and can safely
+		 * manipulate the world.
+		 *
+		 * <p>An example usage of this is to create an explosion where the player is looking:
+		 * <pre>{@code
+		 * // See FabricPacket for creating the packet
+		 * ServerPlayNetworking.registerReceiver(BOOM_PACKET_TYPE, (player, packet, responseSender) -&rt; {
```
Juuxel:
```suggestion
		 * ServerPlayNetworking.registerReceiver(BOOM_PACKET_TYPE, (player, packet, responseSender) -> {
```

Juuxel (changes requested):

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -41,6 +46,9 @@
  *
  * <p>This class should be only used on the physical client and for the logical client.
  *
+ * <p>See {@link ServerPlayNetworking} for information on how to use the new packet-object
+ * based API.
```
Juuxel:
```suggestion
 * <p>See {@link ServerPlayNetworking} for information on how to use the packet object-based API.
```
It won't be new in the future, and I think the hyphen needs to be on "...-based"

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PacketType.java
```diff
@@ -0,0 +1,75 @@
...
+ * This holds the channel ID used for the packet.
+ *
+ * <p>Example of creating a packet type:
+ * <pre>{@code
+ * public static final PacketType<BoomPacket> TYPE = PacketType.create(new Identifier("example:boom"), BoomPacket::new);
+ * }</pre>
+ *
+ * @param <T> the type of the packet
+ */
+public final class PacketType<T extends FabricPacket> {
```
liach, line 31: I want to move the packet write method to a function in the packet type as well, so we can use third-party java classes as packet types.

Technici4n: That doesn't really make sense to me. And it's pretty easy to make a record to wrap a java object as a fabric packet.

Technici4n (commented):
I wonder if we shouldn't split this stuff to a separate class? 500 LOC for `ServerPlayNetworking` is a lot.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PacketType.java
apple502j: Yeah, the vanilla-like system is more intuitive. (See also BlockEntity, Entity etc, where subclass does serialization)

apple502j:
@Technici4n Not sure how to split those - the file is big, yes, but there isn't a clear way of splitting them.

If we're doing a v2, we can drop some methods that don't make sense in FAPI, but not right now.

apple502j:
Question: should there be Packet#getType? Could remove the packet type from send methods then.

Technici4n:
Sure, no split is probably fine.

> Question: should there be Packet#getType? Could remove the packet type from send methods then.

Not sure, either is probably ok.

Juuxel:
Having `getType` would match vanilla code where most things have that method, so it'd be a good change imo. I don't really see multiple types for one packet instance happening

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/play/NetworkingPlayPacketTest.java
```diff
@@ -32,17 +32,16 @@
...
 import net.fabricmc.fabric.api.networking.v1.ServerPlayNetworking;
 import net.fabricmc.fabric.test.networking.NetworkingTestmods;
 
 public final class NetworkingPlayPacketTest implements ModInitializer {
 	public static final Identifier TEST_CHANNEL = NetworkingTestmods.id("test_channel");
 
 	public static void sendToTestChannel(ServerPlayerEntity player, String stuff) {
-		PacketByteBuf buf = PacketByteBufs.create();
-		buf.writeText(Text.literal(stuff));
-		ServerPlayNetworking.send(player, TEST_CHANNEL, buf);
```
modmuss50, line 45: We should contiune to test the old API, its going no where.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +402,57 @@ public interface PlayChannelHandler {
...
+		 * This is for advanced users only; for normal uses, {@link #receive} should be
+		 * implemented instead.
+		 *
+		 * @param client the client
+		 * @param player the player that received the packet
+		 * @param packet the packet
+		 * @param responseSender the packet sender
+		 * @return whether to proceed to {@link #receive}
+		 */
+		default boolean processOnNetworkThread(MinecraftClient client, ClientPlayerEntity player, T packet, PacketSender responseSender) {
```
modmuss50: Maybe put the packet first?

modmuss50: Question: Do we want to pass some sort of context object instad of the client and player? Would make it easier to expand in the future without breaking?

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +402,57 @@ public interface PlayChannelHandler {
...
+		 * @param responseSender the packet sender
+		 * @see FabricPacket
+		 */
+		void receive(MinecraftClient client, ClientPlayerEntity player, T packet, PacketSender responseSender);
+
+		/**
+		 * Processes an incoming packet. This is called on the network thread.
+		 * If this method returns {@code false}, {@link #receive} is not called.
+		 * This is for advanced users only; for normal uses, {@link #receive} should be
+		 * implemented instead.
```
modmuss50: Add a test for this behaviour please.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +402,57 @@ public interface PlayChannelHandler {
...
+	 */
+	@FunctionalInterface
+	public interface PlayPacketHandler<T extends FabricPacket> {
+		/**
+		 * Handles the incoming packet. This is called on the render thread, and can safely
+		 * call rendering-related methods.
+		 *
+		 * <p>An example usage of this is to display an overlay message:
+		 * <pre>{@code
+		 * // See FabricPacket for creating the packet
```
modmuss50, line 438: Nit: is this code comment really needed?

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -62,6 +76,39 @@ public static boolean registerGlobalReceiver(Identifier channelName, PlayChannel
...
+			public void receive(MinecraftClient client, ClientPlayNetworkHandler networkHandler, PacketByteBuf buf, PacketSender sender) {
+				T packet = type.read(buf);
+
+				if (handler.processOnNetworkThread(client, client.player, packet, sender)) {
+					client.execute(() -> {
+						if (networkHandler.getConnection().isOpen()) handler.receive(client, client.player, packet, sender);
+					});
+				}
+			}
+		});
```
modmuss50: Nit: maybe move this to an impl class, and possibly remove duplicatated logic with registerReceiver?

modmuss50 (changes requested):
This is looking great, just a few small things 👍

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Juuxel: Also, the player is available from the client instance

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/play/NetworkingPlayPacketTest.java
apple502j: Ok, reverting.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
apple502j: The signature was from the classic API, which had the handler first. That said yeah, packet first makes more sense.

apple502j: I believe our convention is not to pass context object unless it makes more sense (see DynamicRegistrySetupCallback).

The reason this passes player is simple - it passed `player` only first, realized `player.client` is protected, so I passed client as well; `client.player` does exist, but it is marked `@Nullable` (when, here, it is never null). I believe the code here is fine.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
apple502j: Well, it's never a bad idea to do over-documenting 😛

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
apple502j: Maybe, want more opinions on this

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Sollace: Client is available from a static reference at `MinecraftClient.getInstance()`. I don't see any need for it to be passed at all.

apple502j: Contributing guideline specifies, "Consider passing the MinecraftClient instance as a parameter if it makes sense."

Sollace: Operative words being "if it makes sense". I explicitly ignore it in all of my networking code because it's just another client-exclusive class that I'd have to perform workarounds to reference in a way that won't break on the server.

Passing player makes sense, because I assume it's possible to receive packets for an instance of ClientOtherPlayerEntity as well (I hope), whereas passing client is never more than a convenience since it's always going to be accessible via the public non-nullable, static singleton reference (and is never not that).

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -49,9 +57,15 @@ public final class ClientPlayNetworking {
...
 	 * A global receiver is registered to all connections, in the present and future.
 	 *
+	 * <p>The handler runs on the network thread. After reading the buffer there, rendering
+	 * must be performed in the render thread by calling {@link ThreadExecutor#execute(Runnable)}.
+	 *
 	 * <p>If a handler is already registered to the {@code channel}, this method will return {@code false}, and no change will be made.
 	 * Use {@link #unregisterGlobalReceiver(Identifier)} to unregister the existing handler.
 	 *
+	 * <p>For new code, {@link #registerGlobalReceiver(PacketType, PlayPacketHandler)}
+	 * should be used instead, as it is designed in a way that prevents thread safety issues.
```
Technici4n:
```suggestion
	 * should be preferred, as it is designed in a way that prevents thread safety issues.
```

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -49,9 +57,15 @@ public final class ClientPlayNetworking {
 	 * Registers a handler to a channel.
 	 * A global receiver is registered to all connections, in the present and future.
 	 *
+	 * <p>The handler runs on the network thread. After reading the buffer there, rendering
+	 * must be performed in the render thread by calling {@link ThreadExecutor#execute(Runnable)}.
```
Technici4n:
```suggestion
	 * <p>The handler runs on the network thread. After reading the buffer there, access to
	 * game state must be performed in the render thread by calling {@link ThreadExecutor#execute(Runnable)}.
```

Technically it's not just rendering.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -97,6 +163,9 @@ public static Set<Identifier> getGlobalReceivers() {
 	 * <p>For example, if you only register a receiver using this method when a {@linkplain ClientLoginNetworking#registerGlobalReceiver(Identifier, ClientLoginNetworking.LoginQueryRequestHandler)}
 	 * login query has been received, you should use {@link ClientPlayConnectionEvents#INIT} to register the channel handler.
 	 *
+	 * <p>For new code, {@link #registerReceiver(PacketType, PlayPacketHandler)}
+	 * should be used instead, as it is designed in a way that prevents thread safety issues.
```
Technici4n:
```suggestion
	 * should be preferred, as it is designed in a way that prevents thread safety issues.
```

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +402,57 @@ public interface PlayChannelHandler {
...
+		 * Processes an incoming packet. This is called on the network thread.
+		 * If this method returns {@code false}, {@link #receive} is not called.
+		 * This is for advanced users only; for normal uses, {@link #receive} should be
+		 * implemented instead.
+		 *
+		 * @param client the client
+		 * @param player the player that received the packet
+		 * @param packet the packet
+		 * @param responseSender the packet sender
+		 * @return whether to proceed to {@link #receive}
```
Technici4n: This is unclear and needs to specify explicitly what happens with `{@code true}` / `{@code false}`. Even then, the name could be misleading (one could understand it as "should processing happen on network thread"). A less ambiguous name would be nice.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Technici4n: Packet first makes more sense indeed.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Technici4n: Unless the network thread processing can do everything that the previous API can do? (In which case it's fine to deprecate the old API).

Technici4n (changes requested):

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
apple502j: We prefer, not the readers; changed to "is preferred"

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/channeltest/NetworkingChannelTest.java
```diff
@@ -94,6 +94,7 @@ private static CompletableFuture<Suggestions> suggestReceivableChannels(CommandC
 		return CommandSource.suggestIdentifiers(ServerPlayNetworking.getReceived(player), builder);
 	}
 
+	@SuppressWarnings("deprecation")
```
Juuxel: Was anything here actually deprecated?

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/channeltest/NetworkingChannelTest.java
```diff
@@ -110,6 +111,7 @@ private static int registerChannel(CommandContext<ServerCommandSource> context,
 		return 1;
 	}
 
+	@SuppressWarnings("deprecation")
```
Juuxel: same here

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/channeltest/NetworkingChannelTest.java
apple502j: Nope, removing

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
```diff
@@ -255,4 +402,56 @@ public interface PlayChannelHandler {
...
+		 * If this method returns {@code false}, {@link #receive} is not called.
+		 * This is for advanced users only; for normal uses, {@link #receive} should be
+		 * implemented instead.
+		 *
+		 * @param packet the packet
+		 * @param player the player that received the packet
+		 * @param responseSender the packet sender
+		 * @return {@code true} to proceed to {@link #receive}
+		 */
+		default boolean processOnNetworkThread(T packet, ClientPlayerEntity player, PacketSender responseSender) {
```
modmuss50: Im a little unsure about the naming for this? Its fine unless someone can come up with something better.

modmuss50 (approved):

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Technici4n: I agree it should be changed, the name is the only blocker for me.

apple502j: Hmm, yeah, we need a better name but I can't really come up with one

Juuxel (approved):

apple502j:
Removing from last call now, we might want to remove network thread thing after all

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/api/client/networking/v1/ClientPlayNetworking.java
Sollace: +1 on removing duplicated logic

apple502j:
This is last-call ready.

hYdos (approved):

modmuss50:
Manually merged into 1.19.4, Many Thanks 👍

PR #2821, opened 2023-01-03
Resource Conditions Additions

### LINKED ISSUE #2548: Add `item_registered` Resource Condition
Would also be nice to support any tag key (instead of just blocks, fluids, items) now that they have the registry key too.

Technici4n:
- Add `registry_contains` condition. Closes #2548.
- Make `fabric:load_conditions` appear first in generated JSON objects.
- Uniformize implementation a bit:
  - Move the features enabled `ConditionJsonProvider` implementation up next to the other provider implementations.
  - Rename and finalize `ThreadLocal<FeatureSet> currentFeature` to `final ThreadLocal<FeatureSet> CURRENT_FEATURES`.

### DIFF
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
@@ -16,12 +16,20 @@
 
 package net.fabricmc.fabric.api.resource.conditions.v1;
 
+import java.util.Arrays;
+import java.util.function.Function;
+
+import com.google.common.base.Preconditions;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 
 import net.minecraft.block.Block;
 import net.minecraft.fluid.Fluid;
 import net.minecraft.item.Item;
+import net.minecraft.item.ItemConvertible;
+import net.minecraft.registry.Registries;
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
 import net.minecraft.registry.RegistryKeys;
 import net.minecraft.registry.tag.TagKey;
 import net.minecraft.resource.featuretoggle.FeatureFlag;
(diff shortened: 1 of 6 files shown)


fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -183,6 +240,19 @@ static void init() {
...
+	@SafeVarargs
+	private static <A, B> B[] transform(A[] input, Function<A, B> mapper, B... outputType) {
+		B[] output = Arrays.copyOf(outputType, input.length);
+
+		for (int i = 0; i < input.length; i++) {
+			output[i] = mapper.apply(input[i]);
+		}
+
+		return output;
 	}
```
Technici4n, line 256: I know this is ugly, but it makes the vararg manipulation a lot shorter, which makes the behavior of the various factory methods much easier to follow for someone reading through them the first time. Worth it imo. :wink:

modmuss50 (approved):

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/mixin/resource/conditions/DataProviderMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.data.DataProvider;
+
+import net.fabricmc.fabric.api.resource.conditions.v1.ResourceConditions;
+
+/**
+ * Make the {@value ResourceConditions#CONDITIONS_KEY} appear first in generated JSON objects.
+ */
+@Mixin(DataProvider.class)
+public interface DataProviderMixin {
+	@Dynamic // lambda method passed to Util.make
```
Juuxel:
```suggestion
	@Dynamic("lambda method passed to Util.make")
```

Dynamic can also take the description as a parameter which is used elsewhere in fabric api:
https://github.com/FabricMC/fabric/blob/0780111d1981ebd00ff5c399c2bc8131f866f26e/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java#L99

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -159,6 +168,54 @@ public static ConditionJsonProvider featuresEnabled(FeatureFlag... features) {
 		return ResourceConditionsImpl.featuresEnabled(FEATURES_ENABLED, features);
 	}
 
+	/**
+	 * Creates a condition that returns true if all the passed items are loaded.
+	 *
+	 * @see #registryContains(RegistryKey, Identifier...)
+	 */
+	public static ConditionJsonProvider itemsLoaded(ItemConvertible... items) {
```
Juuxel: Maybe `itemsRegistered` since it wraps `registry_contains`? I don't think we usually talk about loading items, except item models and such

Juuxel (changes requested):

PR #2823, opened 2023-01-03
Fix #2552: Add way to use resource conditions for block loot table datagen

### LINKED ISSUE #2552: Resource conditions can't be used for block loot table datagen
The `FabricLootTableProvider` has a `withConditions` method that can be used with `SimpleFabricLootTableProvider`s to create a `BiConsumer` that applies the condition, but for `FabricBlockLootTableProvider` that is unreasonable as the base class implements `accept` to run the `generateBlockLootTables` method in which is used to generate the loot tables. Thus the only way to apply a condition to block loot tables is by overriding `accept` and calling super with a modified `BiConsumer`, but that applies to all loot tables and doesn't play nicely with strict validation (if split into multiple providers).

Looking at the test mod for datagen [the block loot table provider](https://github.com/FabricMC/fabric/blob/1.19.2/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java#L275) doesn't use any conditions (because it resonably can't) while [the barter loot provider](https://github.com/FabricMC/fabric/blob/1.19.2/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java#L289) does.

Technici4n:
Fixes #2552.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.mixins.json
+++ b/fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.mixins.json
@@ -3,6 +3,7 @@
   "package": "net.fabricmc.fabric.mixin.datagen",
   "compatibilityLevel": "JAVA_16",
   "mixins": [
+    "loot.BlockLootTableGeneratorMixin",
     "AbstractTagProviderMixin",
     "ModelProviderMixin",
     "TagBuilderMixin"
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricBlockLootTableGenerator.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricBlockLootTableGenerator.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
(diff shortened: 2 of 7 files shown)


fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.mixins.json
```diff
@@ -3,6 +3,7 @@
   "package": "net.fabricmc.fabric.mixin.datagen",
   "compatibilityLevel": "JAVA_16",
   "mixins": [
+    "loot.BlockLootTableGeneratorMixin",
```
modmuss50, line 6: Is the package really neeeded for one mixin?

Technici4n: Not sure. Since we're gonna expand this module in the future I thought it might be nicer to keep everything organized? The new API file is in a `loot` subpackage (#2754 will also add subpackages), and that should stay like this IMO, but I'm happy to remove the mixin subpackage if you want.

modmuss50: Thats fine, if it did remove it its likely it would never be added back. 👍

modmuss50 (approved):

PR #2825, opened 2023-01-03
Remove leftover `@ApiStatus.Internal` annotations and add annotation validation

### LINKED ISSUE #2822: Remove leftover `@ApiStatus.Internal` annotations
Since the annotations are now autogenerated for impl classes, most of the remaining annotations can be removed.

Technici4n:
Fixes #2822.


gradle/validate-annotations.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+import static groovy.io.FileType.FILES
+
+class ValidateAnnotations extends SourceTask {
+	private static final def API_STATUS_INTERNAL = ~/@ApiStatus\.Internal/
+	private static final def ENVIRONMENT = ~/@Environment/
+
+	@SkipWhenEmpty
+	@InputDirectory
+	final DirectoryProperty sourceRoot = project.objects.directoryProperty()
```
modmuss50: This isnt needed when using `SourceTask`. 

You can use getSource https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/api/tasks/SourceTask.java#L58

Technici4n: Thanks. How about now?

gradle/validate-annotations.gradle
```diff
@@ -0,0 +1,44 @@
...
+	// We have to capture the source set name for the lazy string literals,
+	// otherwise it'll just be whatever the last source set is in the list.
+	def sourceSetName = sourceSet.name
+	def taskName = sourceSet.getTaskName('validate', 'Annotations')
+	def task = tasks.register(taskName, ValidateAnnotations) {
+		group = 'fabric'
+		description = "Validate annotations for $sourceSetName packages."
+
+		// Only apply to default source directory since there's also generated package-info files.
+		setSource file("src/$sourceSetName/java")
```
modmuss50: Looking better, can this become a single task with the following as the input?:

```gradle
source sourceSets.main.allSource
source sourceSets.client.allSource
```

Technici4n: Wouldn't that include the generated package-info files?

modmuss50: Ah yeah, it would do 👍 

I still think you should be able to make this one task though.

Technici4n: How about now? :smile:

modmuss50: Looks great, thanks 👍

gradle/validate-annotations.gradle
```diff
@@ -0,0 +1,40 @@
...
+				}
+
+				def contents = it.text
+
+				// @Environment is never allowed
+				if (ENVIRONMENT.matcher(contents).find()) {
+					throw new RuntimeException("Found @Environment annotation in file: $it")
+				}
+
+				// @ApiStatus.Internal is only allowed in api packages (it's auto-generated for impl packages)
```
Juuxel:
```suggestion
				// @ApiStatus.Internal is only allowed in api packages (it's auto-generated for impl and mixin packages)
```

Juuxel (approved):

modmuss50 (approved):

PR #2842, opened 2023-01-09
Fix #2814

### LINKED ISSUE #2814: Standardize using one logger factory method
Currently, a mix of SLF4J's `LoggerFactory.getLogger` and Mojang's `LogUtils.getLogger` are used within Fabric API to construct loggers. While most loggers are constructed using the former method, the latter method seems to be used only three times:

https://github.com/FabricMC/fabric/blob/2dae9cfaf02178bc9444422e65fd9681862d2da6/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/impl/biome/NetherBiomeData.java#L51 https://github.com/FabricMC/fabric/blob/2dae9cfaf02178bc9444422e65fd9681862d2da6/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java#L45 https://github.com/FabricMC/fabric/blob/2dae9cfaf02178bc9444422e65fd9681862d2da6/fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/RegistrySyncTest.java#L52

Ideally, Fabric API should stick to using one of these methods. The least impactful refactor would to replace `LogUtils.getLogger` usage with `LoggerFactory.getLogger`.

Domaman202:


fabric-convention-tags-v1/src/generated/resources/data/c/tags/worldgen/biome/climate_cold.json
```diff
@@ -7,9 +7,11 @@
     "minecraft:snowy_taiga",
     "minecraft:old_growth_spruce_taiga",
     "minecraft:old_growth_pine_taiga",
+    "minecraft:snowy_plains",
+    "minecraft:ice_spikes",
```
Technici4n: Don't edit this file manually. It's generated automatically (see build.gradle file of the module). Please update the code and re-run datagen instead.

fabric-convention-tags-v1/src/generated/resources/data/c/tags/worldgen/biome/climate_cold.json
```diff
@@ -12,4 +12,4 @@
       "required": false
     }
   ]
-}
\ No newline at end of file
+}
```
Technici4n, line 15: Why?

Domaman202: Idk lol

PR #2848, opened 2023-01-15
Fix memory leak by using weak set for ArrayBackedEvent list

BasiqueEvangelist:
~~Not completely tested right now, pinging @gliscowo for it™~~ Tested

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/EventFactoryImpl.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/EventFactoryImpl.java
@@ -25,15 +25,17 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Proxy;
 import java.util.Collections;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


BasiqueEvangelist:
I've tested this with owo-lib, and this works: events aren't leaked anymore, even if I open screens
(top is before, bottom is after)
image image

apple502j (approved):
LGTM. We should probably deprecate-for-removal `EventFactory#invalidate`, the only code using this field and is not used by any mods.

Technici4n (approved):
This will do for now. Let's also note that this is completely not thread-safe even though the rest of event creation is.

SquidDev:
You could use `new MapMaker().weakKeys().makeMap()` instead of `WeakHashMap` to make it thread safe, but it may not be worth the bother.

Technici4n:
That's a good idea and it should probably be done just in case.

Player3324 (approved):

modmuss50 (approved):

Issue #2805, opened 2022-12-29
Fabric API breaks Custom Dimensions


Philainel:
I'm building a datapack with custom dimensions (including custom biome & noise settings). Once i noticed that if i rejoin world, the custom dimensions disappear. I've investigated that dimensions disappear from level.dat file as soon as I open world from game or starting the server.
This behavior doesn't applies for vanilla / paper / fabric without fabric-api.
I know that this features are experimental, but I hope it will be fixed later.

Technici4n:
Please link a datapack that we can use to reproduce. Do also mention your MC version.

warjort:
AFAIK this is deliberate to workaround this bug: https://bugs.mojang.com/browse/MC-197860

See:
https://github.com/FabricMC/fabric/blob/1.19.3/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/LevelStorageMixin.java

Although the dimension is not mentioned in level.dat it should still be usable.
Provided there is a datapack/mod loaded defining the dimension.

Philainel:
@Technici4n yep, Minecraft version 1.19.3, fabric-api version 0.70.0
[DiPixDatapack.1672251286.zip](https://github.com/FabricMC/fabric/files/10324117/DiPixDatapack.1672251286.zip)

Philainel:
@warjort I am not deleting any dimension, so i don't think that bug refers to this issue.
Btw dimension files stay in world folder but `/execute in namespace:dimension` can't find them

qouteall:
When I tried to fix the issue of cannot enter world after removing dimension mod, I provided two solutions.
The first one is simple, remove non-vanilla dimensions when reading level data. Because in 1.18.x the custom dimensions are added once the datapack loads, so the dimensions will be re-added and that won't cause issue. However there is now world preset and its dimension is not added every time when loading the world.

My original second solution is more complex: it makes deserialization fault-tolerant. Vanilla has a system of checking "dangling reference" when freezing registry which should also be considered. It's also Forge's solution to this issue.

I am going to remove the original simple solution and use the second solution.

Technici4n:
Fixed by #2856.

PR #2856, opened 2023-01-20
Fix cannot load world after uninstalling dimension mod/datapack without breaking world presets.

qouteall:
In 1.18.x, I had two solutions:
1. Remove non-vanilla dimensions when deserializing as they will be added back later. Was working in 1.18.x. Not working with 1.19 custom presets. https://github.com/FabricMC/fabric/pull/2078
2. Make deserialization fail-soft and remove the invalid entries to avoid having issue in the dangling reference check when freezing registry. https://github.com/FabricMC/fabric/pull/2088

The first solution was chosen because of simplicity.

In 1.19.3, as I tested, there is no need to manually remove the dangling references to avoid error in freezing process. So the second solution become simpler. This PR removes the first solution and uses the second solution.


qouteall:
I tested 
* Loading a world that has bumblezone dimension without bumblezone mod. tested on both client side and server side
* Loading a world that has lost world datapack dimension(https://www.planetminecraft.com/data-pack/the-lost-world-a-liminal-dimension/) after removing the datapack. tested on client side
* That world preset datapack https://github.com/FabricMC/fabric/issues/2805#issuecomment-1367838063 works after relogging. tested on client side

apple502j (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,129 @@
...
+import com.mojang.serialization.codecs.BaseMapCodec;
+import com.mojang.serialization.codecs.UnboundedMapCodec;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Has the same functionality as {@link UnboundedMapCodec}.
+ * But it will fail-soft when an entry cannot be deserialized.
+ */
+public class FailSoftMapCodec<K, V> implements BaseMapCodec<K, V>, Codec<Map<K, V>> {
```
modmuss50: Nit: Make final

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,129 @@
...
+	public static final Logger LOGGER = LoggerFactory.getLogger("FailSoftMapCodec");
```
modmuss50: Nit: can be private

TelepathicGrunt:
And what about dimension datapacks using vanilla chunk generators/biome sources that users remove and fully expected the dimension to be removed? 

Vanilla behavior is to continue to try and create the dimension despite the datapack not on. There won’t be any parsing error since it has no modded entries. Yet the vanilla behavior is a bug and unexpected and against user’s wishes. 

The nuking behavior at least let users know the dimension is gone when datapack is removed.

I don’t see any way of properly handling dimensions in level.dat file. It’s a painful and waste of time when the real solution is to make bug reports to Mojang and push Mojang to advance their plans of overhauling how dimension data is stored

Situation: 

If a user select a preset and tries to load the world afterwards, then that means the level.dat’s dimension data needs to be trusted and read in order to keep the preset alive.

But if the user adds a dimension mod and then later removes it, the level.dat would blow up due to unknown chunk generator types etc. The user expectation is the dimension just gets removed. Not blow up. So the level.dat data can’t be trusted.

And if the user adds a dimension datapack and later removes it, the game still can parse the dimension data and creates it and keeps it persisted against user wishes. No way to know if the dimension datapack is on or not because level.dat is parsed so early before the datapack’s dimensions are read and no way to know if the dimension is removed because it uses valid existing vanilla chunk generators etc. So level.dat data here cannot be trusted either 

Conclusion, to fit all use cases, the level.dat file needs to be simultaneously trusted and not trusted at the same time. A paradox with no solution.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,129 @@
...
+
+	@Override
+	public Codec<K> keyCodec() {
+		return keyCodec;
+	}
+
+	@Override
+	public Codec<V> elementCodec() {
+		return elementCodec;
+	}
```
modmuss50: Nit: Looks like this can cleanly become a record.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,129 @@
...
+					LOGGER.error("Failed to decode key {} from {}  {}", k, pair, kPartialResult);
+				});
+				v.get().ifRight(vPartialResult -> {
+					LOGGER.error("Failed to decode value {} from {}  {}", v, pair, vPartialResult);
+				});
+
+				if (k.get().left().isPresent() && v.get().left().isPresent()) {
+					builder.put(k.get().left().get(), v.get().left().get());
+				} else {
+					// ignore failure
```
modmuss50, line 78: Its a little unclear to me (with only basic knowlage of dfu) what failuire path is taken when a world is missing. May be a good idea to log out when this happens, or this is already handled by one of the other cases?

qouteall: The failure path is normally that, when it tries to parse a generator, it cannot recognize the generator type, then the failure propagates to the whole dimension information and the whole deserialization fails. In old versions it will still load the world but have the non-overworld dimensions removed. In newer versions it will show a screen that it fails to load and the user can click "safe mode", althought clicking "safe mode" doesn't do anything.
With this PR, when the generator fails to deserialize, it will be ignored.

qouteall:
As @TelepathicGrunt said, with this PR, removing a dimension datapack that only uses vanilla things will not have that dimension removed. This is a rare case (most dimension datapacks use its own things).

It's possible to fix both issues by adding extra data to tell that some dimensions comes from preset and should not be removed before reading `level.dat`. This data can be stored in an extra file. It's more complex than the current solution (I may make a PR when I have time). Another edge case is that the preset dimension depends on some things that a datapack provides and that datapack gets removed.

qouteall:
I think that Minecraft's world preset design is not good. The world preset can add dimensions but there was already the functionality of adding dimensions. So there is a duplication of functionality. The world preset should only be a list of dimension ids and the datapack should add the relevant dimensions in the old way, then in `level.dat` it should reference the world preset id.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
qouteall: Changed to record. Checkstyle does not allow a record to be final.

modmuss50 (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/**
+	 * In {@link BaseMapCodec#decode(DynamicOps, MapLike)},
+	 * the whole deserialization will fail if one element fails.
+	 * `apply2stable` will return fail when any of the two elements is failed.
+	 * In this implementation, if one deserialization fails, it will log and ignore.
+	 */
+	@Override
+	public <T> DataResult<Map<K, V>> decode(final DynamicOps<T> ops, final MapLike<T> input) {
```
Technici4n, line 60: How and why is this different from https://github.com/Mojang/DataFixerUpper/pull/55?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/RegistryCodecsMixin.java
```diff
@@ -0,0 +1,63 @@
...
+	 * It doesn't redirect `Codec.unboundedMap` because `FailSoftMapCodec` does not inherit `UnboundedMapCodec`.
+	 * Currently, `createKeyedRegistryCodec` is only used in dimension codec.
+	 */
+	@SuppressWarnings("UnstableApiUsage")
+	@Inject(
+			method = "createKeyedRegistryCodec",
+			at = @At("HEAD"),
+			cancellable = true
+	)
+	private static <E> void injectCreateKeyedRegistryCodec(
```
Technici4n: Can you do this with a `@ModifyVariable` on the `codec` variable instead?

Technici4n (changes requested):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/RegistryCodecsMixin.java
qouteall: Ok I will try that.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
qouteall: I didn't see that PR. Yes this PR probably does similar thing. (The relevant datafixerupper code contains a lot functional abstractions so I am not sure whether that PR works the same as this PR)

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/RegistryCodecsMixin.java
qouteall: By viewing the bytecode, there seem to be two local variables of type `Codec`. The temporary value is also a variable. So `ModifyVariable` may be a little big error-prone. I decided to redirect the call of `xmap`.

qouteall: I forgot that redirect cannot get outer parameter.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
qouteall: I see the difference. In that PR, when one deserialization fails, the result will be in "partial result". In this PR, the result will always be normal result.
The `DataResult` can be in 3 states:
* successful with one result
* failed with one partial result
* failed with no partial result
I am not sure how vanilla handles the partial result (DFU is very convoluted). As I tested, always yielding successful result works.

qouteall: Update: In Minecraft 1.19.3, it rejects partial result of dimension data. (`LevelStorage.createLevelDataParser` uses `getOrThrow` with `allowPartial` argument false.) So that PR won't directly solve the issue without extra changes.

Technici4n (approved):
Looks great now, good call on adding the instanceof check.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FailSoftMapCodec.java
```diff
@@ -0,0 +1,109 @@
...
+			return true;
+		}
+
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+
+		final FailSoftMapCodec<?, ?> that = (FailSoftMapCodec<?, ?>) o;
+		return Objects.equals(keyCodec, that.keyCodec) && Objects.equals(elementCodec, that.elementCodec);
+	}
```
Juuxel: Is this really needed? Records generate an equals method already.

PR #2878, opened 2023-02-02
Add 'Windswept Hills' biome convention tag

TheDeathlyCow:
Adds a 'windswept' tag for biomes that are windswept. I.e., biomes that are hilly with shattered terrain. Similar to the existing `#c:extreme_hills` tag, but also includes the Windswept Forest and Windswept Savanna.


Technici4n:
I don't think this is a good tag name: windswept seems independent from hills

TheDeathlyCow:
Also somewhat related to this tag, the biome `minecraft:windswept_forest` is *not* part of `#c:extreme_hills`. However, this biome was previously called "Extreme Hills+" when it was added in 13w36a ([source](https://minecraft.fandom.com/wiki/Java_Edition_13w36a#World_generation)). I can create a separate issue for this if needed.

Technici4n:
Aren't windswept hills and extreme hills the same thing?

TheDeathlyCow:
I'd argue they arent; mainly due to the existence of the windswept savanna. 'Windswept' biomes are just biomes that have a 'shattered' appearance in their terrain. Extreme hills are a lower temperature and more specialized subset of windswept biomes that is mostly just a holdout of older biomes. Windswept hills, forest, and gravelly hills are the new names for the pre-1.18 mountains (which was a new name for the same biomes that were called 'extreme hills' prior to 1.13). Before 1.18, the windswept savanna was the 'Shattered Savanna Plateau', and generated as a variant of the Savanna rather than as a variant of mountains/extreme hills. Windswept savannas also generally have a bit more magnified look compared to extreme hills.

Technici4n:
Looks like extreme hills to me:
image

(https://minecraft.fandom.com/wiki/Savanna#Windswept_Savanna)

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BiomeTagGenerator.java
```diff
@@ -90,6 +90,10 @@ private void generateCategoryTags() {
 		getOrCreateTagBuilder(ConventionalBiomeTags.EXTREME_HILLS)
 				.add(BiomeKeys.WINDSWEPT_GRAVELLY_HILLS)
 				.add(BiomeKeys.WINDSWEPT_HILLS);
+		getOrCreateTagBuilder(ConventionalBiomeTags.WINDSWEPT_HILLS)
+				.addOptionalTag(ConventionalBiomeTags.EXTREME_HILLS)
```
Technici4n: If we go for this tag, I would rather list them than include the tag - mods might add extreme hills that are not windswept?

Technici4n:
Hmmm, according to MC wiki we have:
- windswept hills, windswept forest, windswept gravelly hills as variants of windswept hills
- windswept savanna as a variant of savannas.

So using `#c:windswept_hills` is probably not accurate? Maybe just `#c:windswept` then tbh... I'd like some input from @FabricMC/fabric-triage.

Technici4n (approved):

Juuxel (approved):

modmuss50 (approved):

PR #2884, opened 2023-02-03
Add FabricStructureProcessor to allow processing of the StructureEntityInfo 

modmuss50:
Javadoc's likely needs improving.

Most of the changes in this PR are just adding a structure to the testmods.


Technici4n (changes requested):
This should really go into a new module (`fabric-structure-api-v1`?) to avoid adding more structure extensions to the biome API in the future. Remember the registry sync mess...

fabric-biome-api-v1/src/testmod/java/net/fabricmc/fabric/test/biome/TestStructureProcessor.java
```diff
@@ -0,0 +1,76 @@
...
+		final String entityId = entityInfo.nbt.getString("id");
+
+		if (entityId.equals("minecraft:item_frame")) {
+			final NbtCompound nbt = entityInfo.nbt.copy();
+			nbt.getCompound("Item").putString("id", item.getKey().get().getValue().toString());
+			return new StructureTemplate.StructureEntityInfo(entityInfo.pos, entityInfo.blockPos, nbt);
+		} else if (entityId.equals("minecraft:armor_stand")) {
+			// Don't spawn armor stands
+			return null;
+		}
```
Technici4n: Couldn't we possibly validate this behavior using a game test? :wink:

Technici4n (changes requested):

fabric-biome-api-v1/src/testmod/java/net/fabricmc/fabric/test/biome/TestStructureProcessor.java
modmuss50: Not easily, game test structure loading is seprate from this. and then they can spawn basically anywhere when placed using the command.

modmuss50:
> This should really go into a new module (`fabric-structure-api-v1`?) to avoid adding more structure extensions to the biome API in the future.

Done 👍 I went with -v1 but started the version at 4.0.0 as we previously had a module with this name released as 3.x.x

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/StructureTemplateMixin.java
```diff
@@ -0,0 +1,81 @@
...
+		contextThreadLocal.set(null);
+	}
+
+	@Redirect(method = "spawnEntities", at = @At(value = "INVOKE", target = "Ljava/util/Iterator;next()Ljava/lang/Object;"))
+	private Object applyEntityProcessor(Iterator<StructureTemplate.StructureEntityInfo> instance) {
+		final StructurePlacementContext context = contextThreadLocal.get();
+		StructureTemplate.StructureEntityInfo entityInfo = instance.next();
+
+		for (final StructureProcessor processor : context.placementData().getProcessors()) {
+			entityInfo = processor.process(
```
AlphaMode: https://github.com/Fabricators-of-Create/Porting-Lib/blob/f86f1ac1f5d7407dafb6eaadc7934da566cf29a2/extensions/src/main/java/io/github/fabricators_of_create/porting_lib/extensions/extensions/StructureProcessorExtensions.java#L9 (mojmap) This doesn't pass the StructureTemplate like StructureProcessorExtensions does meaning I can't even use this

AlphaMode (changes requested):

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/StructureTemplateMixin.java
modmuss50: Hi, thanks for taking a look. What is the use case for this? The vanilla StructureProcessor does not pass it, hence why I didnt. Im quite happy to add it, just a little curious about the use case.

fabric-biome-api-v1/src/testmod/java/net/fabricmc/fabric/test/biome/TestStructureProcessor.java
yungnickyoung: Maybe it's useful to be able to access the StructureTemplate's `entityInfoList`? E.g. to constrain the number of a specific entity that spawns within a structure. Or perhaps the its `size` field could be useful for conditionally processing entities.

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/StructureTemplateMixin.java
yungnickyoung: Maybe it's useful for accessing the StructureTemplate's `entityInfoList`? E.g. to constrain the number of a specific entity that spawns within a structure. Or perhaps its `size` field could be useful for conditionally processing entities.

Technici4n: @AlphaMode so, why is this useful?

Technici4n: @AlphaMode up?

Faboslav: Hello there, is there anything that can be done? This is such a good feature to have for mod compatibility.

yungnickyoung: @Technici4n Btw AlphaMode said on Discord they are actually not sure about use cases for the StructureTemplate. And considering no one else has chimed in about this, I think it's safe to say we don't need to include the StructureTemplate as a parameter. That, or we could just include it anyway in case someone needs it. Regardless, it would be nice to see some activity on this PR, as the feature would be really useful!

modmuss50: Ill get this PR updated, worse case we can always add an overload passing it later.

PR #2896, opened 2023-02-07
Add a localization string to default Fabric API resource pack

### LINKED ISSUE #2895: Fabric API resource pack description should be translatable
https://github.com/FabricMC/fabric/blob/1.19.3/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java#L50 should have a translation string on https://github.com/FabricMC/fabric/blob/1.19.3/fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json

Madis0:
Fixes #2895
Currently untested

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
@@ -47,8 +47,8 @@ public class FabricModResourcePack extends GroupResourcePack {
 		String fileName = String.join("/", pathSegments);
 
 		if ("pack.mcmeta".equals(fileName)) {
-			String description = "pack.description.fabricApi";
-			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":{\"translate\":\"%s\"}}}", description);
+			String description = "pack.description.modResources";
+			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":{\"translate\":\"%s\",\"fallback\":\"Mod resources.\"}}}", description);
 			return () -> IOUtils.toInputStream(pack, Charsets.UTF_8);
 		} else if ("pack.png".equals(fileName)) {
 			return FabricLoader.getInstance().getModContainer("fabric-resource-loader-v0")
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
@@ -47,7 +47,7 @@ public class FabricModResourcePack extends GroupResourcePack {
 		String fileName = String.join("/", pathSegments);
 
 		if ("pack.mcmeta".equals(fileName)) {
-			String description = "Mod resources.";
+			String description = Text.translatable("pack.fabricapi.description");
 			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":\"%s\"}}", description);
 			return () -> IOUtils.toInputStream(pack, Charsets.UTF_8);
 		} else if ("pack.png".equals(fileName)) {
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
@@ -47,8 +47,8 @@ public class FabricModResourcePack extends GroupResourcePack {
 		String fileName = String.join("/", pathSegments);
 
 		if ("pack.mcmeta".equals(fileName)) {
-			String description = Text.translatable("pack.fabricapi.description");
-			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":\"%s\"}}", description);
+			String description = "pack.description.fabricApi";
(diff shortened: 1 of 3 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
```diff
@@ -47,7 +47,7 @@ public InputSupplier<InputStream> openRoot(String... pathSegments) {
 		String fileName = String.join("/", pathSegments);
 
 		if ("pack.mcmeta".equals(fileName)) {
-			String description = "Mod resources.";
+			String description = Text.translatable("pack.fabricapi.description");
```
haykam821: This does not compile as `Text` is not compatible with `String`.

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json
```diff
@@ -1,4 +1,5 @@
 {
+  "pack.fabricapi.description": "Mod resources.",
```
haykam821: Shouldn't this translation key be `pack.description.fabricMod` or similar to better match the other translation keys?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
Juuxel: Yep, the proper replacement is to use `{"translate":"the tl key here"}` instead of `"description text"` in the json

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json
Madis0: Perhaps this would make more sense

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
Madis0: This should work:
```suggestion
			String description = Text.translatable("pack.fabricapi.description").getString();
```
I also understand what @Juuxel meant, but not _where_ would that apply? In other words, I don't see a relevant pack.mcmeta in the repo where I could directly apply it to.

haykam821: @Madis0 Make sure that works for data packs on a dedicated server. It should be fine in theory, especially with dedicated servers loading modded translations, but ideally we should use @Juuxel's strategy if that works.

Juuxel: > but not where would that apply

The very next line in that file

altrisi:
Note that you also need to import the Text class.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricModResourcePack.java
```diff
@@ -47,7 +47,7 @@ public InputSupplier<InputStream> openRoot(String... pathSegments) {
...
 			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":\"%s\"}}", description);
```
Juuxel:
```suggestion
			String description = "pack.description.fabricApi";
			String pack = String.format("{\"pack\":{\"pack_format\":" + type.getPackVersion(SharedConstants.getGameVersion()) + ",\"description\":{\"translate\":\"%s\"}}}", description);
```
This change also fixes the now-incorrect translation key

Technici4n:
Tested by adding `"pack.description.fabricApi": "Ressources des mods.",` to the `fr_fr.json` file, works great!
image

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json
```diff
@@ -1,4 +1,5 @@
 {
+  "pack.description.fabricApi": "Mod resources.",
```
Technici4n: Can we use another id maybe? How about `pack.description.modResources`?

Technici4n (changes requested):

haykam821:
The description will be untranslated when reading the pack information from a vanilla client, such as through the `/datapack list` command:

image

Keep in mind that Minecraft 1.19.4 has a `fallback` field for translatable text that can solve this issue. For Minecraft 1.19.3 and below, the issue is minor enough and pre-existing so it isn't worth figuring out a hack for this case.

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json
Madis0: I don't know, I shall let other FAPI developers choose that.

Madis0:
> The description will be untranslated when reading the pack information from a vanilla client, such as through the `/datapack list` command:
> 
> image
> 
> Keep in mind that Minecraft 1.19.4 has a `fallback` field for translatable text that can solve this issue. For Minecraft 1.19.3 and below, the issue is minor enough and pre-existing so it isn't worth figuring out a hack for this case.

Ahh, I thought one can assume the language files are always loaded. But no problem, I can add that as well and target 1.19.4 branch instead.

fabric-resource-loader-v0/src/main/resources/assets/fabric-resource-loader-v0/lang/en_us.json
Technici4n: I am one of them. :smile:

Madis0:
Rebased to https://github.com/FabricMC/fabric/pull/2906

PR #2900, opened 2023-02-11
Add getters for equipment slot providers and damage handlers

Juuxel:
Some mods need to access these and there's also no reason to "hide" them from mod devs.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomDamageHandler.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomDamageHandler.java
@@ -46,7 +46,7 @@ public interface CustomDamageHandler {
 	int damage(ItemStack stack, int amount, LivingEntity entity, Consumer<LivingEntity> breakCallback);
 
 	/**
-	 * {@return the custom damage handler of the specified item, or {@code null} if it doesn't have one}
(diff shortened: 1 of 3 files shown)


AlphaMode (commented):
Can methods for getting the slot provider be added to FabricItem and FabricItemStack?

Juuxel:
Not sure about FabricItem since it'd mean that it could be overridden which adds another way of registering the same thing.

As for FabricItemStack, it could actually have a method for getting the preferred equipment slot directly I think since you can apply the provider to `this` stack.

Technici4n (changes requested):
May I suggest `ItemConvertible`?

Technici4n:
> Not sure about FabricItem since it'd mean that it could be overridden which adds another way of registering the same thing.

Agreed, let's keep the APIs separate.

> As for FabricItemStack, it could actually have a method for getting the preferred equipment slot directly I think since you can apply the provider to `this` stack.

Why not just use the following `LivingEntity` method?
```java
    public static EquipmentSlot getPreferredEquipmentSlot(ItemStack stack) {
```

May I also ask what the point of exposing these is?

Juuxel:
This was @AlphaMode's request, but I agree the LivingEntity method is probably better. The damage handler is just for the sake of completeness since it uses the same system to attach it to an item.

Juuxel:
Closed in favour of a different approach (PR coming shortly).

PR #2901, opened 2023-02-11
Add registry for core shaders in custom namespaces

Juuxel:
- `CoreShaderRegistrationCallback` lets mods register core shader programs (like `POSITION_TEX`, stored in `assets/namespace/shaders/core`) in any namespace.
- Likewise, the individual shader stages (vertex/fragment) can also be in any namespace.
- `#moj_import` in core shaders lets you import files from custom namespaces too. For example, `#moj_import <my_mod:example.glsl>` imports `assets/my_mod/shaders/include/example.glsl`.

This functionality is useful especially for GUI rendering where code that is slow on the CPU side can be moved to a simple shader, such as [simple tiled textures](https://github.com/CottonMC/LibGui/pull/185).

This PR doesn't touch the post processing shaders (`EffectShaderProgram`) nor their imports so they're still only loadable from the `minecraft` namespace.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/FabricShaderProgram.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/FabricShaderProgram.java
@@ -40,6 +40,6 @@ public final class FabricShaderProgram extends ShaderProgram {
 	 */
 	public static String rewriteAsId(String input, String containedId) {
 		Identifier contained = new Identifier(containedId);
-		return contained.getNamespace() + ':' + input.replace(containedId, contained.getPath());
+		return contained.getNamespace() + Identifier.NAMESPACE_SEPARATOR + input.replace(containedId, contained.getPath());
 	}
 }
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/FabricShaderProgram.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/FabricShaderProgram.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.client.rendering;
+
+import java.io.IOException;
+
+import net.minecraft.client.gl.ShaderProgram;
+import net.minecraft.client.render.VertexFormat;
+import net.minecraft.resource.ResourceFactory;
+import net.minecraft.util.Identifier;
+
+public final class FabricShaderProgram extends ShaderProgram {
+	public FabricShaderProgram(ResourceFactory factory, Identifier name, VertexFormat format) throws IOException {
+		super(factory, name.toString(), format);
+	}
+
+	/**
+	 * Rewrites the input string containing an identifier
+	 * with the namespace of the id in the front instead of in the middle.
+	 *
+	 * <p>Example: {@code shaders/core/my_mod:xyz} -> {@code my_mod:shaders/core/xyz}
+	 *
+	 * @param input       the raw input string
+	 * @param containedId the ID contained within the input string
+	 * @return the corrected full ID string
+	 */
+	public static String rewriteAsId(String input, String containedId) {
+		Identifier contained = new Identifier(containedId);
+		return contained.getNamespace() + ':' + input.replace(containedId, contained.getPath());
+	}
+}
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.client.rendering.shader;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.fabricmc.fabric.impl.client.rendering.FabricShaderProgram;
+
+/**
+ * Lets modded shaders {@code #moj_import} shaders from any namespace with the
+ * {@code <>} syntax.
+ */
+@Mixin(targets = "net.minecraft.client.gl.ShaderProgram$1")
+abstract class ShaderProgramImportProcessorMixin {
+	@Unique
+	private String capturedImport;
+
+	@Inject(method = "loadImport", at = @At("HEAD"))
+	private void fabric_captureImport(boolean inline, String name, CallbackInfoReturnable<String> info) {
+		capturedImport = name;
+	}
+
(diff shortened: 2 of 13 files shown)


AlphaMode (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
```diff
@@ -0,0 +1,55 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.fabricmc.fabric.impl.client.rendering.FabricShaderProgram;
+
+/**
+ * Lets modded shaders {@code #moj_import} shaders from any namespace with the
+ * {@code <>} syntax.
+ */
+@Mixin(targets = "net.minecraft.client.gl.ShaderProgram$1")
+abstract class ShaderProgramImportProcessorMixin {
```
Technici4n, line 35: How is this mixin supposed to work? Does it correctly take the path normalization into account?

image

Technici4n (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Juuxel: Yes, it should work with normalisation as it modifies the normalised path. The original name without the directory `shaders/include/` is included for the second param of `FabricShaderProgram.rewriteAsId` (param 1 is the normalised path, i.e. the current value of `name`).

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
```diff
@@ -0,0 +1,55 @@
...
+	private String capturedImport;
+
+	@Inject(method = "loadImport", at = @At("HEAD"))
+	private void fabric_captureImport(boolean inline, String name, CallbackInfoReturnable<String> info) {
+		capturedImport = name;
+	}
+
+	@ModifyVariable(method = "loadImport", at = @At("STORE"), ordinal = 0, argsOnly = true)
+	private String fabric_modifyImportId(String id, boolean inline) {
+		if (!inline && capturedImport.contains(":")) {
```
modmuss50: Nit:
```suggestion
		if (!inline && capturedImport.contains(Identifier.NAMESPACE_SEPARATOR)) {
```
(Will also need the import)

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/FabricShaderProgram.java
```diff
@@ -0,0 +1,45 @@
...
+	 *
+	 * <p>Example: {@code shaders/core/my_mod:xyz} -> {@code my_mod:shaders/core/xyz}
+	 *
+	 * @param input       the raw input string
+	 * @param containedId the ID contained within the input string
+	 * @return the corrected full ID string
+	 */
+	public static String rewriteAsId(String input, String containedId) {
+		Identifier contained = new Identifier(containedId);
+		return contained.getNamespace() + ':' + input.replace(containedId, contained.getPath());
```
modmuss50: Nit:
```suggestion
		return contained.getNamespace() + Identifier.NAMESPACE_SEPARATOR + input.replace(containedId, contained.getPath());
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
```diff
@@ -0,0 +1,55 @@
...
+ * Lets modded shaders {@code #moj_import} shaders from any namespace with the
+ * {@code <>} syntax.
+ */
+@Mixin(targets = "net.minecraft.client.gl.ShaderProgram$1")
+abstract class ShaderProgramImportProcessorMixin {
+	@Unique
+	private String capturedImport;
+
+	@Inject(method = "loadImport", at = @At("HEAD"))
+	private void fabric_captureImport(boolean inline, String name, CallbackInfoReturnable<String> info) {
```
modmuss50: Nit: No need for the `fabric_` prefix. See other methods as well.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramMixin.java
```diff
@@ -0,0 +1,57 @@
...
+			return FabricShaderProgram.rewriteAsId(id, name);
+		}
+
+		return id;
+	}
+
+	// Allow loading shader stages from arbitrary namespaces.
+	@ModifyVariable(method = "loadShader", at = @At("STORE"), ordinal = 1)
+	private static String fabric_modifyStageId(String id, ResourceFactory factory, ShaderStage.Type type, String name) {
+		if (name.contains(":")) {
```
modmuss50: Nit:

```suggestion
		if (name.contains(Identifier.NAMESPACE_SEPARATOR)) {
```

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java
```diff
@@ -0,0 +1,70 @@
...
+import net.minecraft.client.util.Window;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.rendering.v1.CoreShaderRegistrationCallback;
+import net.fabricmc.fabric.api.client.rendering.v1.HudRenderCallback;
+
+/**
+ * Tests {@link HudRenderCallback} and {@link CoreShaderRegistrationCallback} by drawing a green rectangle
+ * in the upper-right corner of the screen.
```
modmuss50: This test looks great, would be cool if it would show in the auto client test screenshots, right now its hidden by the tips toast. Bottom right should solve this. Im not bothered either way.

modmuss50 (approved):
Looks great, a few minor nit picks. Im far from an expert on rendering stuff like this, so any reviews from people who are would be appreciated. 👍

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Technici4n: Is the `fabric_` prefix (or equivalent) added automatically by mixin now?

modmuss50: Yes.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Technici4n: We don't need to consider normalization since `\` is not a valid ID character. So this will be fine. :+1:

Technici4n (approved):
Looks good to me, the injects make sense and the testmod is great.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Technici4n: All good then :+1:

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java
Juuxel: I'm not very familiar with how the system works, so maybe better left for a follow-up PR?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Juuxel: Oh nice, that was finally merged. Will remove the prefixes.

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java
modmuss50: Should just be a matter of moving it to another corner tbh as its hidden. Have a look at the Test Screenshots from a GHA build: https://github.com/Juuxel/fabric/actions/runs/4151605994

Like I said its not a big issue at all, and we could likely do something a lot better in another PR.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramMixin.java
Juuxel: Needs the import, so merged in manually (also needs a conversion to String)

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramImportProcessorMixin.java
Juuxel: Merged in manually (see the other one)

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java
Juuxel: Alright, I moved it there now

PR #2905, opened 2023-02-13
Support creating FabricRegistryBuilder with a RegistryKey, deprecate Class based APIs.

### LINKED ISSUE #846: Fabric Registry Builder should not take Class<T> argument for type
E.g.
https://github.com/FabricMC/fabric/blob/dfd0b839f4ea14da86c0ac92db3f60a6128d1a30/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryBuilder.java#L66-L68
can just become
```java
 public static <T> FabricRegistryBuilder<T, SimpleRegistry<T>> createSimple(Identifier registryId) { 
 	return from(new SimpleRegistry<T>(RegistryKey.ofRegistry(registryId), Lifecycle.stable())); 
 } 
```
Vanilla's generic type on the stupid registry key thing is just a compile-time favor. It is totally a sugar.
Also what we require here for `Class<T>` would be problematic if you have a registry like `Registry<Predicate<String>>`, where it's a trouble to create a `Class<Predicate<String>>` without extensive casting.

modmuss50:
Fixes #846

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryBuilder.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryBuilder.java
@@ -20,12 +20,13 @@ import java.util.EnumSet;
 
 import com.mojang.serialization.Lifecycle;
 
-import net.minecraft.util.Identifier;
 import net.minecraft.registry.DefaultedRegistry;
 import net.minecraft.registry.MutableRegistry;
+import net.minecraft.registry.Registry;
 import net.minecraft.registry.RegistryKey;
 import net.minecraft.registry.SimpleDefaultedRegistry;
 import net.minecraft.registry.SimpleRegistry;
+import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.mixin.registry.sync.RegistriesAccessor;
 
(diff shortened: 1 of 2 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistryBuilder.java
```diff
@@ -34,7 +35,8 @@
  *
  * <pre>
  * {@code
- *  SimpleRegistry<String> registry = FabricRegistryBuilder.createSimple(String.class, new Identifier("registry_sync", "fabric_registry"))
+ *  RegistryKey<Registry<String>> registryKey = RegistryKey.ofRegistry(new Identifier("registry_sync", "fabric_registry"));
```
Technici4n: Could probably use a description (maybe "Here is an example of how to create a registry of `String` objects"), and another id such as `new Identifier("mod_id", "registry_id")`.

modmuss50: Hopefully thats better now?

Technici4n (approved):

Juuxel (approved):

PR #2908, opened 2023-02-14
Transfer API: Add slotted storage and non-empty iterator

Technici4n:
- Add non-empty iterator system. This allows many slots to be skipped in large inventories provided a suitable tracking structure. Adds `StorageUtil.extractAny`.
- Add `SlottedStorage` that allows storages to optionally expose a slot list. This is optional, and is meant for usage by chest-like inventories.
- Clarify edge case in `SingleVariantStorage#getBlankVariant`.
- Remove unused field from `TransferApiImpl`.

None of these should be breaking changes.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
@@ -74,17 +74,11 @@ public interface InventoryStorage extends SlottedStorage<ItemVariant> {
 	@UnmodifiableView
 	List<SingleSlotStorage<ItemVariant>> getSlots();
 
-	/**
-	 * {@inheritDoc}
-	 */
 	@Override
 	default int getSlotCount() {
 		return getSlots().size();
 	}
 
-	/**
-	 * {@inheritDoc}
-	 */
 	@Override
 	default SingleSlotStorage<ItemVariant> getSlot(int slot) {
 		return getSlots().get(slot);
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
@@ -28,7 +28,7 @@ import net.minecraft.inventory.SidedInventory;
 import net.minecraft.inventory.SimpleInventory;
 import net.minecraft.util.math.Direction;
 
-import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
+import net.fabricmc.fabric.api.transfer.v1.storage.SlottedStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.CombinedStorage;
 import net.fabricmc.fabric.api.transfer.v1.storage.base.SingleSlotStorage;
 import net.fabricmc.fabric.impl.transfer.item.InventoryStorageImpl;
@@ -50,7 +50,7 @@ import net.fabricmc.fabric.impl.transfer.item.InventoryStorageImpl;
  */
 @ApiStatus.Experimental
 @ApiStatus.NonExtendable
-public interface InventoryStorage extends Storage<ItemVariant> {
+public interface InventoryStorage extends SlottedStorage<ItemVariant> {
 	/**
 	 * Return a wrapper around an {@link Inventory}.
 	 *
@@ -69,11 +69,21 @@ public interface InventoryStorage extends Storage<ItemVariant> {
 	 * Retrieve an unmodifiable list of the wrappers for the slots in this inventory.
 	 * Each wrapper corresponds to a single slot in the inventory.
 	 */
+	@Override
 	List<SingleSlotStorage<ItemVariant>> getSlots();
 
 	/**
-	 * Retrieve a wrapper around a specific slot of the inventory.
+	 * {@inheritDoc}
 	 */
+	@Override
+	default int getSlotCount() {
+		return getSlots().size();
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
 	default SingleSlotStorage<ItemVariant> getSlot(int slot) {
 		return getSlots().get(slot);
 	}
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.UnmodifiableView;
 
 import net.minecraft.entity.player.PlayerInventory;
 import net.minecraft.inventory.Inventory;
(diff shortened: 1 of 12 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
```diff
@@ -69,11 +69,21 @@ static InventoryStorage of(Inventory inventory, @Nullable Direction direction) {
...
+	@Override
 	List<SingleSlotStorage<ItemVariant>> getSlots();
 
 	/**
-	 * Retrieve a wrapper around a specific slot of the inventory.
+	 * {@inheritDoc}
 	 */
+	@Override
+	default int getSlotCount() {
+		return getSlots().size();
```
TiphaineLAURENT, line 79: Isn't that a problematic recursion since `getSlots` uses `getSlotCount` ?

Technici4n: `InventoryStorage` reabstracts `getSlots`, and the two implementations override it anyway.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/SlottedStorage.java
```diff
@@ -0,0 +1,64 @@
...
+	 * Retrieve a specific slot of this storage.
+	 *
+	 * @throws IndexOutOfBoundsException If the slot index is out of bounds.
+	 */
+	SingleSlotStorage<T> getSlot(int slot);
+
+	/**
+	 * Retrieve all the slots of this storage. <b>The list must not be modified.</b>
+	 */
+	default List<SingleSlotStorage<T>> getSlots() {
```
SquidDev, line 64: I know I said on Discord I didn't have anything to add, but apparently that's not true. Sorry!

I'm not quite sure when it'd be appropriate to call this function. I feel ideally one of the two would be true:

 - This function is fast to run (like the `InventoryStorage` implementation), in which case it's safe to call relatively often.
 - The return value of this is constant, in which case it can potentially be cached by consumers.

I wonder if it'd be better to either:
 - Drop the default implementation and require implementations to provide an efficient implementation.
 - Just return an `Iterable` here.
 - Return a  `List` which acts as a view over the current `SlottedStorage` rather than creating a new array.

Phoupraw: Maybe `org.jetbrains.annotations.UnmodifiableView` is helpful.

Technici4n: I don't want to force implementors to return an immutable list, it's up to users not to modify the list.

For your comment @SquidDev: the return value cannot really be cached in the sense that the inventory can change its number of slots at any time in theory (and the inventory itself should not be cached anyway). I like the idea of returning a list that acts as a view over the storage.

Phoupraw: > I don't want to force implementors to return an immutable list, it's up to users not to modify the list.
> 
> For your comment @SquidDev: the return value cannot really be cached in the sense that the inventory can change its number of slots at any time in theory (and the inventory itself should not be cached anyway). I like the idea of returning a list that acts as a view over the storage.

Implementors don's need to return a actual immutable list. The annotation is just to warn users not to modify the list. After all, annotations are stronger than doc.

Technici4n: Both suggestions have been implemented, thank you both.

SquidDev (approved):
Thanks!

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
```diff
@@ -69,11 +70,22 @@ static InventoryStorage of(Inventory inventory, @Nullable Direction direction) {
...
 	 * Each wrapper corresponds to a single slot in the inventory.
 	 */
+	@Override
+	@UnmodifiableView
 	List<SingleSlotStorage<ItemVariant>> getSlots();
 
 	/**
-	 * Retrieve a wrapper around a specific slot of the inventory.
+	 * {@inheritDoc}
 	 */
```
Juuxel: This isn't needed at all if the javadoc is otherwise empty.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/item/InventoryStorage.java
```diff
@@ -69,11 +70,22 @@ static InventoryStorage of(Inventory inventory, @Nullable Direction direction) {
...
+	@Override
+	default int getSlotCount() {
+		return getSlots().size();
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
```
Juuxel: Same here

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/SlottedStorage.java
```diff
@@ -0,0 +1,67 @@
...
+	SingleSlotStorage<T> getSlot(int slot);
+
+	/**
+	 * Retrieve a list containing all the slots of this storage. <b>The list must not be modified.</b>
+	 *
+	 * <p>This function can be used to interface with code that requires a slot list,
+	 * for example {@link StorageUtil#insertStacking} or {@link ContainerItemContext#getAdditionalSlots()}.
+	 *
+	 * <p>It is guaranteed that calling this function is fast.
+	 * The default implementation returns a view over the storage that delegates to {@link #getSlotCount} and {@link #getSlot}.
```
Juuxel, line 61: Do we have `@implSpec` here for specifying the default impl? (Yarn adds it to its list of javadoc tags at least)

Juuxel (changes requested):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/SlottedStorage.java
Technici4n: Sadly we don't.

Technici4n: (They just vanish if you run `gradlew javadoc`...)

Juuxel: Probably a good improvement for a future PR, they're really easy to add:
https://github.com/FabricMC/yarn/blob/4eed62e0827b13c25af0a5b1262fb204c9ddbf52/build.gradle#L554-L558

Juuxel (approved):

modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/TransferApiImpl.java
```diff
@@ -89,4 +88,54 @@ public T next() {
...
+			}
+
+			@Override
+			public StorageView<T> next() {
+				if (!hasNext()) {
+					throw new NoSuchElementException();
+				}
+
+				return next;
+			}
```
modmuss50: This never calls `findNext()`, so next is never null or advanced, causing an infinite loop.

PR #2898, opened 2023-02-09
Indigo shade related fixes and other changes

PepperCode1:
- Apply disabled shade from vanilla quads directly to material. `MutableQuadViewImpl#fromVanilla` now modifies the passed material to force disable diffuse if the vanilla quad has shade disabled, which allows the material to accurately represent the diffuse state of the quad. `QuadViewImpl` no longer stores a hidden `shade` boolean that cannot be copied by a FRAPI user. These changes do not change the behavior of any Indigo methods.
- Fix `AoCalculator` using AO face data computed with a potentially different shade state. With enhanced AO mode, after a non-axis-aligned quad had AO data calculated and cached, any future quads on the same block with a different shade state would be rendered as though they had the shade state of the first quad. The AO data cache now separates data by shade state to fix this issue.
- Respect cull parameter passed to non-terrain rendering.
- Other minor cleanup and performance improvements

PR #2932, opened 2023-02-23
Fix #2898 backport

PepperCode1:
- Restore BlockRenderContext.didOutput
- Restore null checks for AoCalculator.vanillaCalc


Technici4n (approved):
Thanks for the fix, the AO calc stuff would likely have went unnoticed. Weird that the `didOutput` stuff didn't fail.

modmuss50 (approved):
Thanks 👍

PR #2940, opened 2023-02-26
Fixed end biome source injection

TelepathicGrunt:
While I know of no mod that actually uses the Fabric API system to inject biomes into the end, I am fairly certain this code is bugged. Not sure how this was missed

biomeSetModified is set to false initially. When a locate command or anything calls getBiomes(), biomeSetModified is set to true and the modded biomes added to output of getBiomes(). Afterwards, all future calls to getBiomes() will never have the modded biomes because biomeSetModified is now true which makes the if statement fail.

You can change the implementation if you wish but this PR is just a quick fix alternative by introducing another bool to use to properly set biomeSetModified as true or false based on if there are modded biomes or not. And then always add the modded biomes to getBiomes() output if biomeSetModified is true.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/TheEndBiomeSourceMixin.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/TheEndBiomeSourceMixin.java
@@ -55,6 +55,9 @@ public class TheEndBiomeSourceMixin extends BiomeSourceMixin {
 	@Unique
 	private boolean biomeSetModified = false;
 
+	@Unique
+	private boolean hasCheckedForModifiedSet = false;
+
 	/**
 	 * Modifies the codec, so it calls the static factory method that gives us access to the
 	 * full biome registry instead of just the pre-defined biomes that vanilla uses.
@@ -106,8 +109,12 @@ public class TheEndBiomeSourceMixin extends BiomeSourceMixin {
 
 	@Override
 	protected void fabric_modifyBiomeSet(Set<RegistryEntry<Biome>> biomes) {
-		if (!biomeSetModified) {
-			biomeSetModified = true;
+		if (!hasCheckedForModifiedSet) {
+			hasCheckedForModifiedSet = true;
+			biomeSetModified = !overrides.get().customBiomes.isEmpty();
+		}
+
+		if (biomeSetModified) {
 			biomes.addAll(overrides.get().customBiomes);
 		}
 	}


apple502j:
@TelepathicGrunt Does this need backport, or is this 1.19.4 only? (the relevant code has changed in 1.19.4-pre1)

TelepathicGrunt:
Looks the same in 1.19.3 and 1.19.2. Thus they also have the bug
https://github.com/FabricMC/fabric/blob/4c3c5bfdf5426ae9acfef49f36ac2b528542cd26/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/TheEndBiomeSourceMixin.java#L109

apple502j:
@TelepathicGrunt No, does this bug actually occur? Or is it theoretical?

TelepathicGrunt:
Again, I don’t know if any more actually using the biome api to add end biomes. And I can’t run Fabric API in dev as it always breaks for me. 

But this came up when QSL made a test mod that runs a locate for the modded end biome in the actual game when the world starts up to verify for sure if the biome was injected. It was failing.

just looking at the code points to why. That biomeSetModified field is false initially for the end biome source. After you call getBiome once, it is set to true and nothing else ever modifies that field again. Thus future getBiomes will not have the modded biomes. 

This means if you test in game, you may find that /locate biome works once to find the end biome. But I guarantee if you run the locate command a second time, it will fail. This is because locate biome doesn’t start the search for the biome right away. It goes to the biomeSource first, calls getBiomes to get all of the biomes, and then checks if the wanted biome is in that set first before it actually starts checking positions in the world

See this for fabric API’s change to getBiomes method itself
https://github.com/FabricMC/fabric/blob/153d9bf1aa78eb569ff6e975c4b03c5c1c942a36/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/BiomeSourceMixin.java#L37

so if that biomeSetModified is set to true and never false again, that original code is never going to go inside the if statement to add the modified biome again to the new hash set instance that is created every time with the default vanilla biomes when getBiomes is called

apple502j (approved):

Technici4n:
I think it would be useful to have a similar test ourselves.

modmuss50 (changes requested):
I agree, this needs a test. No point fixing it just for it to break again during a port.

> And I can’t run Fabric API in dev as it always breaks for me.

Feel free to reach out, you know where to find me. Ideally it should just be the same as a normal mod, but in practice I know its not always the case. Its possible we need some better setup docs?

apple502j:
Automated test would be either too fragile or RAM-eating since we don't fix the seed. The most obvious "solution" I won't discuss here.

modmuss50 (approved):

LambdAurora:
> Automated test would be either too fragile or RAM-eating since we don't fix the seed. The most obvious "solution" I won't discuss here.

I'm not sure how automatically executing the code of a `/locate biome` command has such issues. The issue here has been caught thanks to such test.

apple502j:
@LambdAurora Since we generate a random world, there is a possibility that the code works, but RNG just disagrees. We can repeat locatebiome command to generate and check for more chunks, but that slows things down especially when running ourselves. The solution here is to use a set seed, but that is a topic for another PR.

LambdAurora:
Sure there's RNG, but in the case of the test biomes, for the locate code to fail would suggest a more likely broken code than bad luck.

modmuss50:
I dont think RAM is an issue. However I can see there being a rare chance that the test fails due to the lack of a hard coded seed. With this one test its unlikely, but if we started locating a lot more of the testmod world gen content it can become quite likely. Having stable tests is critical, when tests become too unstable they get easily ignored/overlooked.

I agree that this can be resolved at in another PR, as the fix presented here is trivial.

PR #2947, opened 2023-03-03
Fix custom packet inside of bundled packet not being processed

Technici4n:
The contents of a bundled packet are processed on the main thread, so we shouldn't ignore custom payload packets on the main thread.

This fix has the side-effect of going through the ClientPlayNetworkAddon interception logic twice if the custom packet is not registered, but that should not cause problems (it's just a map lookup).

PS: Trying to flag the packet as "processed" could technically be incorrect in case the packet is sent multiple times in an integration connection, so I think it's best avoided.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerPlayNetworkAddon.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerPlayNetworkAddon.java
@@ -76,11 +76,6 @@ public final class ServerPlayNetworkAddon extends AbstractChanneledNetworkAddon<
 	 * @return true if the packet has been handled
 	 */
 	public boolean handle(CustomPayloadC2SPacket packet) {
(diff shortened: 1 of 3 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerPlayNetworkAddon.java
```diff
@@ -76,11 +76,6 @@ public void onClientReady() {
 	 * @return true if the packet has been handled
 	 */
 	public boolean handle(CustomPayloadC2SPacket packet) {
-		// Do not handle the packet on game thread
-		if (this.server.isOnThread()) {
-			return false;
-		}
-
```
Technici4n, line 83: This is not needed to fix the issue at hand, but I thought I might as well do it for consistency and in case bundled C2S packets are ever introduced. Can be reverted if desired.

apple502j (approved):

modmuss50 (approved):

Issue #2649, opened 2022-11-10
The climate_cold tag does not contain the snowy plains biome


maityyy:
[climate_cold.json](https://github.com/FabricMC/fabric/blob/1.19.2/fabric-convention-tags-v1/src/generated/resources/data/c/tags/worldgen/biome/climate_cold.json)
[Snowy Plains Biome](https://minecraft.fandom.com/wiki/Snowy_Plains#ID)

PR #2955, opened 2023-03-08
Conventional Biome Tags tweaks

### LINKED ISSUE #2330: More convention tags
So i think about adding tags like `birch_forest`,`spruce_forest` etc. It would be useful for mods that adds structures wood type-specific.

maityyy:
See #2330 & #2649


apple502j (changes requested):
Please run datagen and generate files

Juuxel (approved):

modmuss50 (approved):

Issue #2963, opened 2023-03-13
ExtendedScreenHandlerFactory is not passed ScreenHandler for context


Ueaj-Kerman:
`ExtendedScreenHandlerFactory` has method `void writeScreenOpeningData(ServerPlayerEntity player, PacketByteBuf buf)`, however this method is not passed the ScreenHandler that is being written. This makes it rather annoying to take data straight from the ScreenHandler and write it on open. As a result the same code can't be used for sync-on-change and sync-on-open.

<img width="884" alt="image" src="https://user-images.githubusercontent.com/28693059/224834662-9f92172b-7107-4515-b7b9-073a59814e51.png">

This can be fixed by either changing the signature of the method or creating a `ContextAwareExtendedScreenHandlerFactory` of sorts to retain compatibility.

Technici4n:
With #3077, the screen handler can be accessed via the passed player object.

PR #2964, opened 2023-03-13
Pass ScreenHandler to ExtendedScreenHandlerFactory

Ueaj-Kerman:
Addresses #2963

In order to retain backwards compatibility with the old API, a second interface (WithScreenHandler) was created, that throws if you use the old method to initialize it.

### DIFF
--- a/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java
+++ b/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java
@@ -65,7 +65,11 @@ public final class Networking {
 		buf.writeIdentifier(typeId);
 		buf.writeVarInt(syncId);
 		buf.writeText(factory.getDisplayName());
-		factory.writeScreenOpeningData(player, buf);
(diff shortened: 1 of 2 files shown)


fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java
```diff
@@ -65,7 +65,11 @@ public static void sendOpenPacket(ServerPlayerEntity player, ExtendedScreenHandl
 		buf.writeIdentifier(typeId);
 		buf.writeVarInt(syncId);
 		buf.writeText(factory.getDisplayName());
-		factory.writeScreenOpeningData(player, buf);
+		if(factory instanceof ExtendedScreenHandlerFactory.WithScreenHandler withHandlerFactory) {
```
modmuss50, line 68: Please fix checkstyle

modmuss50 (changes requested):

Technici4n:
This approach compromises the quality of the API. A workaround is to use a different screen handler factory every time and store the screen handler in it.

modmuss50:
I have created a new PR with a much cleaner none breaking fix: https://github.com/FabricMC/fabric/pull/3077 please let me know what you think.

PR #2975, opened 2023-03-16
Add `BEFORE_SPAWN` event

Digifox03:
This pull request adds a new event invoked before a player is spawned (i.e. when opening a single-player world or joining a server).
Some use cases are:
- Detecting when a player is about to spawn.
- Overriding where a player initially spawns.
- Changing a player's initial data before the server sends it to the client.

I have tested it on `1.19.4` and `1.19.3`, but it should work with most previous versions.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -27,6 +27,7 @@ public final class ServerPlayerEvents {
 	 * An event that is called before a player is spawned when opening a world or joining a server.
 	 *
 	 * <p>Mods may use this event to set the player's world and position before it is completely spawned.
+	 *
 	 * <p>Note: this event is not called when a player is respawned.
 	 * For that purpose, {@link ServerPlayerEvents#COPY_FROM} should be used.
 	 */
@@ -41,9 +42,12 @@ public final class ServerPlayerEvents {
 	 *
 	 * <p>This event is typically called before a player is completely respawned.
 	 * Mods may use this event to copy old player data to a new player.
+	 *
 	 * <p>Notes:
-	 * <ul><li>Unlike event {@link ServerPlayerEvents#BEFORE_SPAWN}, changing the player's world might cause undefined behaviour and crash the game.
-	 * <li>Changing the player's position may have undesired effects because respawn side effects (such as the respawn anchor losing a charge) will happen anyway.
+	 * <ul>
+	 * <li>Unlike event {@link ServerPlayerEvents#BEFORE_SPAWN}, changing the player's world might cause undefined behaviour and crash the game.</li>
+	 * <li>Changing the player's position may have undesired effects because respawn side effects (such as the respawn anchor losing a charge) will happen anyway.</li>
+	 * </ul>
 	 */
 	public static final Event<ServerPlayerEvents.CopyFrom> COPY_FROM = EventFactory.createArrayBacked(ServerPlayerEvents.CopyFrom.class, callbacks -> (oldPlayer, newPlayer, alive) -> {
 		for (CopyFrom callback : callbacks) {
@@ -97,8 +101,10 @@ public final class ServerPlayerEvents {
 		 * Called when player data is copied to a new player.
 		 *
 		 * <p>Notes:
-		 * <ul><li>Unlike event {@link ServerPlayerEvents#BEFORE_SPAWN}, changing the player's world might cause undefined behaviour and crash the game.
-		 * <li>Changing the player's position may have undesired effects because respawn side effects (such as the respawn anchor losing a charge) will happen anyway.
+		 * <ul>
+		 * <li>Unlike event {@link ServerPlayerEvents#BEFORE_SPAWN}, changing the player's world might cause undefined behaviour and crash the game.</li>
+		 * <li>Changing the player's position may have undesired effects because respawn side effects (such as the respawn anchor losing a charge) will happen anyway.</li>
+		 * </ul>
 		 *
 		 * @param oldPlayer the old player
 		 * @param newPlayer the new player
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -62,6 +62,17 @@ public final class ServerPlayerEvents {
 		return true;
 	});
 
+	/**
+	 * An event that is called before a player is initially spawned.
+	 *
+	 * <p>Mods may use this event to set the player's world and position before it is completely spawned.
+	 */
+	public static final Event<BeforeSpawn> BEFORE_SPAWN = EventFactory.createArrayBacked(BeforeSpawn.class, callbacks -> (player) -> {
+		for (BeforeSpawn callback : callbacks) {
+			callback.beforeSpawn(player);
+		}
+	});
+
 	@FunctionalInterface
 	public interface CopyFrom {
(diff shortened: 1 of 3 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -62,6 +62,17 @@ public final class ServerPlayerEvents {
 		return true;
 	});
 
+	/**
+	 * An event that is called before a player is initially spawned.
```
kevinthegreat1: I suggest mentioning that this is only called when the player first connects to the server. Also, I think it's better to put this above `COPY_FROM`.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -103,6 +114,16 @@ public interface AllowDeath {
 		boolean allowDeath(ServerPlayerEntity player, DamageSource damageSource, float damageAmount);
 	}
 
+	@FunctionalInterface
+	public interface BeforeSpawn {
+		/**
+		 * Called before a player is first spawned when opening a world or joining a server.
```
kevinthegreat1: Same as above.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/PlayerManagerMixin.java
```diff
@@ -32,4 +38,19 @@ abstract class PlayerManagerMixin {
 	private void afterRespawn(ServerPlayerEntity oldPlayer, boolean alive, CallbackInfoReturnable<ServerPlayerEntity> cir) {
 		ServerPlayerEvents.AFTER_RESPAWN.invoker().afterRespawn(oldPlayer, cir.getReturnValue(), alive);
 	}
+
+	@Inject(method = "onPlayerConnect", at = @At(value = "INVOKE", target = "net/minecraft/server/network/ServerPlayerEntity.setWorld(Lnet/minecraft/server/world/ServerWorld;)V", shift = At.Shift.AFTER))
+	private void afterWorldSet(ClientConnection connection, ServerPlayerEntity player, CallbackInfo ci) {
+		ServerPlayerEvents.BEFORE_SPAWN.invoker().beforeSpawn(player);
+	}
+
+	@ModifyVariable(method = "onPlayerConnect", name = "serverWorld2", at = @At(value = "INVOKE", target = "net/minecraft/server/network/ServerPlayerEntity.setWorld(Lnet/minecraft/server/world/ServerWorld;)V", shift = At.Shift.AFTER))
```
kevinthegreat1: Using `name` is not good practice. Can you use ordinal instead?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/PlayerManagerMixin.java
```diff
@@ -32,4 +38,19 @@ abstract class PlayerManagerMixin {
...
+	private ServerWorld fixServerWorld(ServerWorld world, ClientConnection connection, ServerPlayerEntity player) {
+		return player.getWorld();
+	}
+
+	@ModifyVariable(method = "onPlayerConnect", name = "registryKey", at = @At(value = "INVOKE", target = "net/minecraft/server/network/ServerPlayerEntity.setWorld(Lnet/minecraft/server/world/ServerWorld;)V", shift = At.Shift.AFTER))
```
kevinthegreat1: Same as above. Also `registryKey` isn't used after where you're injecting (`setWorld`), but I guess leaving it here is fine.

kevinthegreat1 (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
Digifox03: I'm not sure what you mean by that.
To avoid misunderstandings, the event is invoked:
- in single-player: every time you load the world.
- in multi-player: every time you join the server.

The event is not restricted to when a player joins a server (or a single-player world) for the first time.

kevinthegreat1: I meant that this event is not invoked on respawn. Maybe direct people to use `COPY_FROM` if they want to be invoked on respawn.

Digifox03: That is possible, but it is important to note that the `COPY_FROM` event does not offer the same semantics.
While the `BEFORE_SPAWN` event accounts for the handlers setting the player's dimension, `COPY_FROM` does not, resulting in a corrupted state.

kevinthegreat1: Yeah ok, I think its good to have all that info in the docs.

kevinthegreat1 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/PlayerManagerMixin.java
```diff
@@ -32,4 +38,19 @@ abstract class PlayerManagerMixin {
...
+
+	@ModifyVariable(method = "onPlayerConnect", ordinal = 1, at = @At(value = "INVOKE", target = "net/minecraft/server/network/ServerPlayerEntity.setWorld(Lnet/minecraft/server/world/ServerWorld;)V", shift = At.Shift.AFTER))
+	private ServerWorld fixServerWorld(ServerWorld world, ClientConnection connection, ServerPlayerEntity player) {
+		return player.getWorld();
+	}
+
+	@ModifyVariable(method = "onPlayerConnect", ordinal = 0, at = @At(value = "INVOKE", target = "net/minecraft/server/network/ServerPlayerEntity.setWorld(Lnet/minecraft/server/world/ServerWorld;)V", shift = At.Shift.AFTER))
+	private RegistryKey<World> fixRegistryKey(RegistryKey<World> world, ClientConnection connection, ServerPlayerEntity player) {
+		return player.getWorld().getRegistryKey();
+	}
```
modmuss50, line 55: Why is this needed? Some comments might be useful.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -23,11 +23,31 @@
...
 public final class ServerPlayerEvents {
+	/**
+	 * An event that is called before a player is spawned when opening a world or joining a server.
+	 *
+	 * <p>Mods may use this event to set the player's world and position before it is completely spawned.
+	 *
+	 * <p>Note: this event is not called when a player is respawned.
+	 * For that purpose, {@link ServerPlayerEvents#COPY_FROM} should be used.
+	 */
+	public static final Event<BeforeSpawn> BEFORE_SPAWN = EventFactory.createArrayBacked(BeforeSpawn.class, callbacks -> (player) -> {
```
modmuss50, line 34: BEFORE_JOIN_SERVER maybe a better name? Also dont want to confuse it with ServerPlayConnectionEvents.JOIN

If anyone has any suggestions please let me know.

frqnny: BEFORE_FIRST_SPAWN is also an option because I think it's about the first spawn in a World?

PR #2979, opened 2023-03-20
Add codec data provider

ErrorCraft:
This pull request adds a data provider for atlas sources. It also includes a generic class (`FabricDataProvider<T>`) for potential other data generators. (See #2759)

\
I do have a few questions myself open for discussion.

Should `FabricDataProvider<T>` be moved to the `impl` package? It *could* be used by a user to implement their own provider if they decided to add a generator specific for their own mod, which is why I placed it under the `api` package instead.

Should `FabricAtlasSourceProvider` be moved under a `provider.client` package? Other ones like `FabricModelProvider` aren't under a `client`-specific package either, while the the existing `ModelProvider` is.

Is it recommended to create a special functional interface in the `FabricDataProvider<T>` class or to use the built-in consumers? In other words, would a `Provider<T>` or a `BiConsumer<Identifier, T>` be more ideal?
```java
@FunctionalInterface
public interface Provider<T> {
	void accept(Identifier id, T value);
}
```


modmuss50:
Nice, this is a good start however I think this would be best just as a simple codec provider, I dont think there is a need to have a specific `FabricAtlasSourceProvider`. It could also be used for some of the dimension jsons for example.

ErrorCraft:
Alright, I renamed `FabricDataProvider` to `FabricCodecDataProvider` as suggested in Discord. I also removed the `FabricAtlasSourceProvider` as it no longer really added anything. I did, however, keep the atlas sources provider in the test mod, mostly to test a specific implementation (as this was the original purpose) and as an example. Not sure if you want it removed as well, or if other ones like one for the dimensions should be included? I'll update the pull request name too to reflect its more general purpose now.

modmuss50 (approved):

Technici4n (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -0,0 +1,81 @@
...
+
+		this.configure(provider);
+		return this.write(writer, entries);
+	}
+
+	protected abstract void configure(BiConsumer<Identifier, T> provider);
+
+	private JsonElement convert(Identifier id, T value) {
+		DataResult<JsonElement> dataResult = this.codec.encodeStart(JsonOps.INSTANCE, value);
+		return dataResult.result().orElseThrow(() -> new IllegalStateException("Invalid entry " + id));
```
Juuxel: Might be good to not swallow the original error message by using `DataResult.get` instead:
```suggestion
		return dataResult.get()
				.mapRight(partial -> "Invalid entry %s: %s".formatted(id, partial.message()))
				.orThrow();
```
(Written in github comments so not 100% sure whether it compiles; also `RuntimeException` is better here since `IllegalStateException` has a different purpose that doesn't apply here)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -0,0 +1,81 @@
...
+			if (existingJson != null) {
+				throw new IllegalArgumentException("Duplicate entry " + id);
+			}
+		};
+
+		this.configure(provider);
+		return this.write(writer, entries);
+	}
+
+	protected abstract void configure(BiConsumer<Identifier, T> provider);
```
Juuxel, line 73: This method could use a short javadoc comment like in the related methods in other providers

Juuxel (changes requested):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
ErrorCraft: Finally had the time to look at the review. I thought I saw this exception being thrown in other places too, hence why I used it for consistency, but I must've misread it since I can't seem to find it anymore.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
ErrorCraft: True, then it would probably be useful to create another PR that adds javadoc to some of the other methods like [`FabricDynamicRegistryProvider::configure`](https://github.com/FabricMC/fabric/blob/1.19.4/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java#L71)?

PR #2995, opened 2023-04-05
Add an event that fires when client player left-clicks

qouteall:
Currently there is no event that handles the case where the player left-clicks air (not pointing to any block or entity). [The portal gun mod needs this event](https://github.com/iPortalTeam/PortalGun/blob/1.19/src/main/java/tk/meowmc/portalgun/mixin/MixinMinecraft.java).

The `ClientPreAttackCallback` will fire when the client player left-clicks and is not in cooldown or in hand-busy state (e.g. pressing W in a boat). As it's a client event, common usages of it requires sending packets to the server.


fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,41 @@
...
+
+package net.fabricmc.fabric.api.event.client.player;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+import net.minecraft.util.ActionResult;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public interface ClientPreAttackCallback {
```
modmuss50, line 46: Some basic docs would be helpful

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,41 @@
...
+	Event<ClientPreAttackCallback> EVENT = EventFactory.createArrayBacked(
+			ClientPreAttackCallback.class,
+			(listeners) -> (player) -> {
+				for (ClientPreAttackCallback event : listeners) {
+					ActionResult result = event.onClientPlayerPreAttack(player);
+
+					if (result != ActionResult.PASS) {
+						return result;
+					}
+				}
```
modmuss50, line 56: Checkstyle

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
```diff
@@ -157,4 +162,25 @@ private void injectUseEntityCallback(CallbackInfo ci, Hand[] hands, int i1, int
...
+					target = "Lnet/minecraft/client/network/ClientPlayerEntity;getStackInHand(Lnet/minecraft/util/Hand;)Lnet/minecraft/item/ItemStack;"
+			),
+			cancellable = true
+	)
+	private void onDoAttack(CallbackInfoReturnable<Boolean> cir) {
+		ActionResult actionResult = ClientPreAttackCallback.EVENT.invoker().onClientPlayerPreAttack(MinecraftClient.getInstance().player);
+
+		if (actionResult != ActionResult.PASS) {
+			if (actionResult.isAccepted()) {
+				attackCooldown = 10;
```
modmuss50: Not a massive fan of the magic numbers, MinecraftClient.field_32145 is likely this constant? Might be nice to get it mapped in yarn?

modmuss50 (changes requested):

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,60 @@
...
+ * <ul>
+ *    <li>PASS : still do vanilla handling (break block, attack entity, swing hand).</li>
+ *    <li>FAIL : it will not do vanilla handling. </li>
+ *    <li>SUCCESS : it will not do vanilla handling and the attack cooldown will be reset.</li>
+ * </ul>
+ * <p>
+ * This event will not fire when in attack cooldown or when the player hand is busy riding.
+ * </p>
+ * <p>
+ * If the player click multiple times in one tick, this event may be fired multiple times in one tick.
```
modmuss50:
```suggestion
 * If the player clicks multiple times in one tick, this event may be fired multiple times in one tick.
```

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,60 @@
...
+ * <p>
+ * Callback for when the client player left-clicks.
+ * This event is client-only. Handling this event probably involves sending custom packets.
+ * </p>
+ * The meaning of return value:
+ * <ul>
+ *    <li>PASS : still do vanilla handling (break block, attack entity, swing hand).</li>
+ *    <li>FAIL : it will not do vanilla handling. </li>
+ *    <li>SUCCESS : it will not do vanilla handling and the attack cooldown will be reset.</li>
+ * </ul>
```
Juuxel: New APIs should avoid using `ActionResult` return values since it has other values that especially interaction events tend to leave unhandled or unspecified, such as `CONSUME` and `CONSUME_PARTIAL` in your case. Fabric's own `TriState` is probably a better fit here.

qouteall: @Juuxel I am going to make it to just return a boolean for whether to intercept, and add extra methods for getting and setting attack cooldown. Do you think it's ok?

qouteall:
I changed the callback to return a boolean indicating whether to intercept. The event does not touch cooldown now, the developer may need to manually set cooldown.

The `MinecraftClient#attackCooldown` is protected, should Fabric API provide an accessor of it?

Technici4n:
Thanks for the PR! A few questions:

- Does this fire even when the client will hit something such as a block or an entity? Might be useful to clarify in the javadoc.
- Is there a way to intercept the action while choosing whether to keep or remove the hand swing, similarly to what other interaction events do? (I believe that's why Juuz suggested a `TriState`).

qouteall:
> Does this fire even when the client will hit something such as a block or an entity?

Yes. Will update doc.

> Is there a way to intercept the action while choosing whether to keep or remove the hand swing, similarly to what other interaction events do?

If it intercepts, it will not do hand swing. However block breaking can still make hand swing. Using `AttackBlockCallback` can disable hand swing caused by block breaking.

qouteall:
Only this event cannot stop block breaking, because vanilla runs block breaking logic when the key is in pressed state every tick, but that event only fires when clicking.

If the mod wants to fully disable block breaking when the player is holding some item, it needs to also use `AttackBlockCallback`

qouteall:
git messed up, going to create another PR

PR #2997, opened 2023-04-05
Fix `ProgrammerArtResourcePack::getName`

NebelNidas:
`getName` is a misleading Yarn name and should instead be called `getId`. Because of that misleading name I made the method return `Programmer Art` instead of the correct `programmer_art` in #2957. See the initial issue report here:  https://discord.com/channels/507304429255393322/566276937035546624/1090437627108413511.

This PR restores the correct Vanilla behavior.


Juuxel (approved):
Probably worth making a Yarn issue as well if it's a yarnbug

modmuss50 (approved):

PR #3005, opened 2023-04-10
Add a Fake Player API

### LINKED ISSUE #1980: Fake Player API
It seems there are a few mods that use some sort of 'fake player' system in order to allow non-player actors to interact with the world. However, I've seen several implementations of this system that fail to replicate vanilla contracts of real players, causing conflicts with one of my mods.

By having one single fake player implementation in Fabric API, there would be only one set of quirks that mod developers would have to be aware of. Fabric API could also reasonably choose ways to alter the vanilla contracts if necessary to support a fake player system, in ways that other mods would have trouble getting away with.

*Original issue:* https://github.com/gnembon/fabric-carpet/issues/1142
*Another issue at the same point, caused by another incomplete fake player implementation:* https://github.com/KyoriPowered/adventure-platform-fabric/issues/14

@altrisi @gnembon @TropheusJ what do you think about a Fabric API-based solution for your fake player systems?

Technici4n:
Adds a single API class to the interaction events module: `FakePlayer` extending `ServerPlayerEntity`. A fake player can either be obtained directly using the static `get` method, or the class can be further subclassed if required by a mod for some reason. This makes it easier to make non-players perform player actions (e.g. block break spell, block placing machine, etc...).

- **Q: Why do we need this API?**
  - A: Each fake player API introduces some edge cases. It's better to have a single implementation for those mods that need to handle the edge cases. It also makes fake player usage easier in general.
- **Q: Why does this not mock more methods?**
  - A: The goal is to stay rather lean. Compatibility issues with vanilla code should be addressed by mocking more methods. Compatibility issues with modded code can be dealt with in the other mod by checking `instanceof FakePlayer`.
- **Q: Why in the interaction events module?**
  - A: This is related to interactions, and I found no better place. Note that in some cases a fake player might be used to fire some event with a player parameter.

Closes #1980.


shartte:
Should the fake player implementation itself be hidden, and instead be exposed via a factory class, which also exposes an `is` method to check if a player is a fake player?

also: Should we allow setting the profile? (See AE2 issue https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/6992)

Technici4n:
> Should the fake player implementation itself be hidden, and instead be exposed via a factory class, which also exposes an `is` method to check if a player is a fake player?

I considered it, but I think it's useful to be able to subclass the fake player if needed. That's why it's not `final`.

> also: Should we allow setting the profile? (See AE2 issue [AppliedEnergistics/Applied-Energistics-2#6992](https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/6992))

To set the profile, you would query another fake player with a different `GameProfile`. I will add a few helper methods to streamline things a bit.

altrisi:
Would it be possible to recommend a different way to check whether a player is a fake player? For example `getClass() != ServerPlayerEntity.class` (from what I know this is what multiple mods do atm), or a method that can be soft-overridden?

For example Carpet fake players extend directly `ServerPlayerEntity`, and can't extend this new class given we don't use Fabric API (and I don't think you can conditionally change the base class at runtime easily), so that'd make mods relying on this `instanceof` check to not detect these players as fake.

Technici4n:
One goal of this API is to standardize what is definitely supported by fake player implementations, which is why subclassing `FakePlayer` is mandatory. Can you link your fake player implementation?

altrisi:
It's a bit messy as it hasn't received much love in a long time, but it's this: (Mojmap) https://github.com/gnembon/fabric-carpet/blob/51f213fa15cbbd70c5a73f4d5fecb280ac33f846/src/main/java/carpet/patches/EntityPlayerMPFake.java

In our case even if it's standardized and it could be used for this use case, that still doesn't solve the issue of what to do when the mod doesn't use Fabric API though. It's not that extending this class would not allow it to do something ~(I haven't checked, probably would be possible)~, it's that we can't extend the class without the dependency.

altrisi:
Actually (from a quick look) these not ticking and being invulnerable is an issue given we want these players to behave mostly like regular players, for their use on testing farms while allowing to automate some actions. Plus other modifications like removing stats may not always be desired.

This API may just not be suited for our use case, though I still think mods should be able to find that they're fake players. Though then I'm not understanding what the use case is.

Juuxel:
Some prior work for a more flexible fake player check can be found in Architectury API's Fabric code which has an event a bit like this:
```java
interface FakePlayerCallback {
  Event<FakePlayerCallback> EVENT = ...;

  TriState isFake(PlayerEntity player);
}
```

(with the fallback impl doing the check mentioned earlier: `instanceof ServerPlayerEntity && getClass() != ServerPlayerEntity.class`)

Actual code (mojmap warning!!): [FakePlayers.java](https://github.com/architectury/architectury-api/blob/7a715af053ba794d3d4b353c8b201ee5242bdf6a/fabric/src/main/java/dev/architectury/hooks/level/entity/fabric/FakePlayers.java)

Technici4n:
@altrisi I'm not sure how exactly you're using fake players, but it's pretty clear that you don't benefit from the provided base implementation.

We can move the check to an interface-injected method, but I am worried about people using an instanceof check by mistake

apple502j:
@Technici4n You clearly haven't read the Discord discussion concerning exactly that point regarding Carpet, which is that mocking a player requires mocking everything from clientside interactions, packet handling, etc..

Carpet bot is not "fake", but more like "client-integrated". And I believe most uses, outside of the simple "just let me run this code without NPEing, I do not mind if it deviates a ton from real player" require such players. This API's scope is still unclear, and I believe this should not be included in the main API.

apple502j:
If the only rationale is that a mod might want to determine if someone is a fake player, injected method or event is enough.

shartte:
> Carpet bot is not "fake", but more like "client-integrated".

What does that even mean?

SquidDev:
Really exciting to see this! Fake players are incredibly nasty to get right, so one standardised library for it is definitely nice to see :).

A couple of other vanilla methods it may be worth stubbing:

 - Override `LivingEntity.sleep` to be a no-op. Otherwise fake players right clicking a bed will permanently occupy it.
 - Similarly, `Entity.startRiding(Entity, bool)` should just be a no-op and return `false`.
 - I'd argue `shouldDamagePlayer` should always return `true`, but that's probably more open  to debate. Thank you for making this a user-extensible class though, as it means this behaviour can be customised!
 - `PlayerAdvancementTracker.setOwner` should be a no-op when passed a `FakePlayer`. Otherwise creating a fake player blocks a normal player from obtaining advancements. Probably should also no-op when awarding advancements to a fake player.

Technici4n:
Thank you Squid for these suggestions, these have been implemented. I chose to override `getScoreboardTeam` which should take care of `shouldDamagePlayer`.

I also added a default UUID and a helper method to use it. The default UUID matches the one used by Forge to hopefully make some people's lives easier.

Regarding carpet, since its fake player should behave as much as possible like a real player, it has been decided that it should not be considered a fake player at all, and thus not considered for this PR. Hence, I am keeping the current design as it makes the API much cleaner.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+import net.minecraft.screen.NamedScreenHandlerFactory;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.stat.Stat;
+import net.minecraft.util.math.BlockPos;
+
+import net.fabricmc.fabric.impl.event.interaction.FakePlayerNetworkHandler;
+
+/**
+ * A "fake player", i.e. a {@link ServerPlayerEntity} that is not a human player.
```
apple502j: I think this would make the javadoc class description `A "fake player", i.`

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+ * They are typically used to automatically perform player actions such as placing blocks.
+ *
+ * <p>The easiest way to obtain a fake player is with {@link FakePlayer#get(ServerWorld, GameProfile)}.
+ * It is also possible to create a subclass for more control over the fake player's behavior.
+ * For good inter-mod compatibility, fake players should have the UUID of their owning (human) player.
```
apple502j: Hmm, how does equals work? Fake players shouldn't equal the real ones.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * <p>Fake players try to behave like regular {@link ServerPlayerEntity} objects to a reasonable extent.
+ * In some edge cases, or for gameplay considerations, it might be necessary to check whether a {@link ServerPlayerEntity} is a fake player.
+ * This can be done with an {@code instanceof} check: {@code player instanceof FakePlayer}.
+ */
+public class FakePlayer extends ServerPlayerEntity {
+	/**
+	 * Default UUID, for fake players not associated with a specific (human) player.
+	 */
+	public static final UUID DEFAULT_UUID = UUID.fromString("41C82C87-7AfB-4024-BA57-13D2C99CAE77");
```
apple502j, line 68: Add a comment explaining why this specific UUID is chosen?

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+	private static final GameProfile DEFAULT_PROFILE = new GameProfile(DEFAULT_UUID, "[Minecraft]");
+
+	/**
+	 * Retrieve a fake player for the specified world, using the {@link #DEFAULT_UUID default UUID}.
```
apple502j: Retrieves.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+	 * This is suitable when the fake player is not associated with a specific (human) player.
```
apple502j, line 73: Should this have a note about reuse?

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,141 @@
...
+	 * Otherwise, the UUID of the owning (human) player should be used.
+	 *
+	 * @see #get(ServerWorld, GameProfile)
+	 */
+	public static FakePlayer get(ServerWorld world) {
+		return get(world, DEFAULT_PROFILE);
+	}
+
+	/**
+	 * Retrieve a fake player for the specified world and game profile.
```
apple502j: Retrieves

apple502j (changes requested):

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
haykam821:
```suggestion
	public static final UUID DEFAULT_UUID = UUID.fromString("FAB41CF4-B41C-FAB1-41C1-FAB41CFAB41C");
```

Technici4n: I said it's the one used by Forge, but it doesn't really need a comment, could very well have been any one UUID.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
Technici4n: Interestingly, it makes it `A "fake player", i.e.`

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
Technici4n: What kind of `equals`? If you need to differentiate you can generally use `instanceof FakePlayer`.

apple502j: What happens if a hash map contains a real player, and it is queried using fake player with the same profile, for example.

Technici4n: Better now?

modmuss50 (approved):

apple502j (approved):

Juuxel (approved):

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/entity/FakePlayer.java
```diff
@@ -0,0 +1,154 @@
...
+	 */
+	public static FakePlayer get(ServerWorld world, GameProfile profile) {
+		Objects.requireNonNull(world, "World may not be null.");
+		Objects.requireNonNull(profile, "Game profile may not be null.");
+
+		return FAKE_PLAYER_MAP.computeIfAbsent(new FakePlayerKey(world, profile), key -> new FakePlayer(key.world, key.profile));
+	}
+
+	private record FakePlayerKey(ServerWorld world, GameProfile profile) { }
+	private static final Map<FakePlayerKey, FakePlayer> FAKE_PLAYER_MAP = new MapMaker().weakValues().makeMap();
```
altrisi, line 102: Won't this leak the worlds anyway because the values are what are made weak while `FakePlayerKey`s aren't and they hold a reference to the `ServerWorld`?

It should be cleared at some point.

altrisi: Or does the value being GC'd somehow trigger the key to be to?

Technici4n: > Or does the value being GC'd somehow trigger the key to be to?

Indeed, that is the whole point. 😉

PR #3033, opened 2023-04-26
Split test mods into client and common.

modmuss50:
The existing testmod run config will need editing or regenerating to point to the new `testmodClient` sourceset.


Technici4n:
Too much to review, but we have tests and we can easily fix things later if they break. :+1:

PR #3043, opened 2023-05-03
Add an event that fires when client player left-clicks

qouteall:
[Original PR](https://github.com/FabricMC/fabric/pull/2995)

Currently there is no event that handles the case where the player left-clicks air (not pointing to any block or entity). The portal gun mod needs this.

The `ClientPreAttackCallback` will fire every tick when the attack key is pressed, before vanilla attack handling. If the callback returns true, then the vanilla handling (breaking block, attacking entity, swining hand) will be cancelled. For multiple callbacks, if the former callback returns true, the later callback won't execute.

This event does not consider attack cooldown.

### DIFF
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
@@ -23,7 +23,8 @@ import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 /**
- * This event fires every tick when the attack key (left mouse button by default) is in pressed state.
+ * This event fires every tick when the attack key (left mouse button by default) is pressed
+ * (including clicking and holding the attack key).
  * If the callback returns true,
  * the vanilla handling (block breaking, entity attacking, hand swing) will be cancelled,
  * and the later callbacks of this event are also cancelled.
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
@@ -16,41 +16,33 @@
 
 package net.fabricmc.fabric.api.event.client.player;
 
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.network.ClientPlayerEntity;
 
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 /**
- * <p>
- * This event fires every tick when the the attack key (left mouse button by default) is in pressed state.
+ * This event fires every tick when the attack key (left mouse button by default) is in pressed state.
  * If the callback returns true,
  * the vanilla handling (block breaking, entity attacking, hand swing) will be cancelled,
  * and the later callbacks of this event are also cancelled.
- * </p>
- * <p>
- * This event is client-only, which means handling it may require sending custom packets.
- * </p>
- * <p>
- * To check whether the attack key is just pressed, use
- * <code>
- * MinecraftClient.getInstance().options.attackKey.wasPressed()
- * </code>
- * inside the callback.
- * </p>
- * <p>
- * The vanilla attack cooldown does not affect this event.
+ *
+ * <p>This event is client-only, which means handling it may require sending custom packets.
+ *
+ * <p>The event fires both when clicking and holding attack key.
+ * To check whether the attack key is just clicked, use {@code clickCount != 0}
+ *
+ * <p>The vanilla attack cooldown and player game mode does not affect this event.
+ * The mod probably needs to check {@link net.minecraft.client.MinecraftClient#attackCooldown} and the game mode.
  * {@link net.minecraft.entity.player.ItemCooldownManager} can be used for custom item cooldown handling.
- * </p>
  */
 public interface ClientPreAttackCallback {
 	Event<ClientPreAttackCallback> EVENT = EventFactory.createArrayBacked(
 			ClientPreAttackCallback.class,
-			(listeners) -> (player) -> {
+			(listeners) -> (client, player, clickCount) -> {
 				for (ClientPreAttackCallback event : listeners) {
-					boolean intercepts = event.onClientPlayerPreAttack(player);
-
-					if (intercepts) {
+					if (event.onClientPlayerPreAttack(client, player, clickCount)) {
 						return true;
 					}
 				}
@@ -61,7 +53,8 @@ public interface ClientPreAttackCallback {
 
 	/**
 	 * @param player the client player
+	 * @param clickCount the click count of the attack key in this tick.
 	 * @return whether to intercept attack handling
 	 */
-	boolean onClientPlayerPreAttack(ClientPlayerEntity player);
+	boolean onClientPlayerPreAttack(MinecraftClient client, ClientPlayerEntity player, int clickCount);
 }
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.client.player;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * <p>
+ * This event fires every tick when the the attack key (left mouse button by default) is in pressed state.
+ * If the callback returns true,
+ * the vanilla handling (block breaking, entity attacking, hand swing) will be cancelled,
+ * and the later callbacks of this event are also cancelled.
+ * </p>
+ * <p>
+ * This event is client-only, which means handling it may require sending custom packets.
+ * </p>
+ * <p>
+ * To check whether the attack key is just pressed, use
+ * <code>
+ * MinecraftClient.getInstance().options.attackKey.wasPressed()
+ * </code>
+ * inside the callback.
+ * </p>
+ * <p>
+ * The vanilla attack cooldown does not affect this event.
+ * {@link net.minecraft.entity.player.ItemCooldownManager} can be used for custom item cooldown handling.
+ * </p>
+ */
+public interface ClientPreAttackCallback {
+	Event<ClientPreAttackCallback> EVENT = EventFactory.createArrayBacked(
+			ClientPreAttackCallback.class,
+			(listeners) -> (player) -> {
+				for (ClientPreAttackCallback event : listeners) {
+					boolean intercepts = event.onClientPlayerPreAttack(player);
+
+					if (intercepts) {
+						return true;
+					}
+				}
+
+				return false;
+			}
+	);
+
+	/**
+	 * @param player the client player
+	 * @return whether to intercept attack handling
+	 */
+	boolean onClientPlayerPreAttack(ClientPlayerEntity player);
+}
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
@@ -188,8 +188,7 @@ public abstract class MinecraftClientMixin {
 			fabric_attackCancelled = ClientPreAttackCallback.EVENT.invoker().onClientPlayerPreAttack(
 					(MinecraftClient) (Object) this, player, attackKeyPressCount
 			);
-		}
-		else {
+		} else {
 			fabric_attackCancelled = false;
 		}
 	}
--- a/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
+++ b/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
@@ -142,11 +142,11 @@ public abstract class MinecraftClientMixin {
 	@Shadow
 	@Final
 	public GameOptions options;
-	
+
 	@Shadow
 	@Nullable
 	public ClientPlayerInteractionManager interactionManager;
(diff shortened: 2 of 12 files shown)


fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package net.fabricmc.fabric.api.event.client.player;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
```
Technici4n, line 25: For javadoc formatting, we:
- do not open the first paragraph with `<p>`,
- leave a blank line before each `<p`>
- put `<p>` on the same line as the opening paragraph,
- don't use `</p>` at all.

We also use `{@code xxx}` instead of `<code>` tags.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,67 @@
...
+					boolean intercepts = event.onClientPlayerPreAttack(player);
+
+					if (intercepts) {
+						return true;
+					}
+				}
+
+				return false;
+			}
+	);
```
Technici4n: Nitpick: can be written in a shorter way

```suggestion
	Event<ClientPreAttackCallback> EVENT = EventFactory.createArrayBacked(ClientPreAttackCallback.class,
			(listeners) -> (player) -> {
				for (ClientPreAttackCallback event : listeners) {
					if (event.onClientPlayerPreAttack(player)) {
						return true;
					}
				}

				return false;
			}
	);
```

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * @param player the client player
+	 * @return whether to intercept attack handling
+	 */
+	boolean onClientPlayerPreAttack(ClientPlayerEntity player);
```
Technici4n: `boolean` parameter is not good enough as I can see 3 use cases:
- cancel vanilla logic, return `true` from `doAttack` to spawn particles ("success")
- cancel vanilla logic, return `false` from `doAttack` to not spawn particles ("failure")
- continue to next event listener / vanilla logic ("pass")

I would suggest using `ActionResult` and documenting the possible cases (SUCCESS, PASS, FAIL at least). See how it's done in `AttackBlockCallback` for example.

Technici4n: It would be nice to have the `crosshair target` as a parameter too. Might also be nice to have the `MinecraftClient`.

Technici4n (changes requested):
I just looked at the code, and I think the old event made more sense? This one won't fire if the key was just pressed and the released before that could be processed by the client, among other things. The rest of this review is written with the assumption that this event should fire from `doAttack` right before the `switch`, let me know if you think that doesn't make sense!

The documentation problem with the old event (I assume that was the problem) can be solved by writing extensive javadoc. A list of things to keep in mind can be appropriate. Having to set the `attackCooldown` to stop further invocations, and the fact that this event fires before the block and entity ones would make sense to have in the javadoc imo. The spectator check mentioned in `AttackBlockCallback` might also be worth including.

For the `attackCooldown`, we could add a transitive access widener as it is quite useful (I use it in AE2 and MI for the attack block callback).

I will also include some review comments about code style so that everything _can_ be done in one update pass from you, but of course the most important things is getting the injection point right.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
Juuxel: TriState > ActionResult here IMO since ActionResult has those unnecessary `CONSUME(_PARTIAL)` values that typically have undefined behaviour in FAPI events.

Technici4n: Imo the `TriState` values are still less clear. Consider `TRUE` vs `SUCCESS`, `FALSE` vs `FAIL`, or `DEFAULT` vs `PASS`...

qouteall:
@Technici4n The reason that I change the old event into the new event is that, the new event is more general and more capable:
* The old event can intercept block breaking if the player clicks very fast (press and then release in one tick), but cannot intercept block breaking if the player holds attack key
* The old event does not fire when the player holds attack key
* The new event runs when the attack key is in pressed state, and the mod can check whether it's clicked and consume the click.
* The new event allows fully disabling block breaking. (In the old event, using both that event and AttackBlockCallback can disable block breaking, but is more fragile and verbose.)
* The new event allows operations that does not consier cooldown. If the mod wants cooldown, it can manually check the cooldown or use `ItemCooldownManager`. The old event does not allow operations that ignore cooldown.

Making the attackCooldown accessible would be good.

Technici4n:
Hmmm, I am not too sure. The idea is to provide an event that fills some of the gaps while still being useful to a larger set of mods, and limiting the amount of misuse.

If the idea is to provide a "raw input" event (which seems to be what your vision is), wouldn't that be better suited as a more generic keyboard and mouse input handling API?

>     * The old event can intercept block breaking if the player clicks very fast (press and then release in one tick), but cannot intercept block breaking if the player holds attack key
> 
>     * The old event does not fire when the player holds attack key

Would it not be confusing to have the event fire every tick instead of just when the attack key is pressed? I am not sure. How do you prevent the event from firing every tick in your portal gun code?

qouteall:
@Technici4n This event is not a raw input event because it has the functionality of disabling vanilla attack action (breaking block, attacking entity). 

The testmod already shows how to run only when clicking by `attackKey.wasPressed()`  https://github.com/FabricMC/fabric/pull/3043/files#diff-a4414bc28298c7766eb961096e202737b149cbfc18487ee95ea882c2b8ed318fR37

(`wasPressed` will consume the click. For item-specific operations consuming the click is fine. But if the mod don't want to consume the click it can directly read the field of `attackKey`)

I am going to make this event to pass `clicked` boolean indicating whether it clicks, without consuming the click.

What if a mod wants to do somthing when the player holds attack key and not pointing to any block? That's why I made it fire every tick.

For a raw input event, to cancel attack, it may need to modify key state which may cause unintended side effects. However a raw input event would be useful in other cases.

Technici4n:
My main motivation is trying to avoid having ending up with 10 events that do slightly different things as this can be very hard to reason about from a maintenance POV, especially if each event ends up being a slight mixin variation.

I think having a clicked parameter is a good idea as it makes the event quite versatile, that makes the design fine by me then!

I wonder if clicked should even be an integer to fully replace checking the key binding. Not sure, your call.

Technici4n:
PS: Your example should also include a check for the gamemode and the attack cooldown in a "real world" use case I assume.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
qouteall: Which particle do you mean by that? If block breaking is cancelled, there will be no block breaking particle.

qouteall: Added MinecraftClient. The crosshair target can be easily get

qouteall:
Changes pushed

fabric-events-interaction-v0/src/main/resources/fabric-events-interaction-v0.accesswidener
```diff
@@ -0,0 +1,7 @@
+accessWidener v2 named
+
+# fields for mods to use
+transitive-accessible field net/minecraft/client/MinecraftClient attackCooldown I
```
Technici4n: These should be moved to the `fabric-transitive-access-wideners-v1` module.

Technici4n (changes requested):
Small comment. Looks very good now! I should have time for a full review at the end of the week.

fabric-events-interaction-v0/src/main/resources/fabric-events-interaction-v0.accesswidener
Technici4n: (The reasoning is that we need to minimize the amount of modules with AWs to keep the build times sane, as each module with an AW has its own MC jar. And it makes sense to group TAWs together.)

fabric-events-interaction-v0/src/main/resources/fabric-events-interaction-v0.accesswidener
```diff
@@ -0,0 +1,7 @@
...
+
+transitive-accessible field net/minecraft/client/option/KeyBinding timesPressed I
```
Technici4n: This one will need an accessor mixin (TAWs from other modules don't work).

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
```diff
@@ -157,4 +166,36 @@ private void injectUseEntityCallback(CallbackInfo ci, Hand[] hands, int i1, int
...
+	@Inject(
+			method = "handleInputEvents",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/client/network/ClientPlayerEntity;isUsingItem()Z",
+					ordinal = 0
+			)
+	)
+	private void injectHandleInputEventsForPreAttackCallback(CallbackInfo ci) {
+		if (options.attackKey.isPressed()) {
```
Technici4n: Would it make sense to also fire the event if `timesPressed > 0`? Then listeners could check `if (key.isPressed())` if they needed to.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
```diff
@@ -157,4 +166,36 @@ private void injectUseEntityCallback(CallbackInfo ci, Hand[] hands, int i1, int
...
+
+	@Inject(method = "doAttack", at = @At("HEAD"), cancellable = true)
+	private void injectDoAttackForCancelling(CallbackInfoReturnable<Boolean> cir) {
+		if (fabric_attackCancelled) {
+			cir.setReturnValue(false);
+		}
+	}
+
+	@Inject(method = "handleBlockBreaking", at = @At("HEAD"), cancellable = true)
+	private void injectHandleBlockBreakingForCancelling(boolean breaking, CallbackInfo ci) {
```
Technici4n, line 204: Should we let the `attackCooldown` be set to `0` or not? Not sure.

Technici4n (changes requested):
PR looks solid, just a few questions.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java
Technici4n: Yes you are right, this fires way before block breaking particles are even considered.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
Technici4n: Should we also call `cancelBlockBreaking`?

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
qouteall: If timesPressed > 0 then isPressed() must be true.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
qouteall: The attackCooldown should be set by mod if necessary

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
Technici4n: Not necessarily if the click was very fast, right?

qouteall: In MC code, it always call `onKeyPressed` after `setKeyPressed` regardless how fast it clicks

qouteall:
I originally want to make two accessors of KeyBinding to have different names to reduce confusion but the github action checks mixin name so two accessors can only have the same name. Now there are two `KeyBindingAccessor`s

qouteall:
Can anyone explain why check_resources failed

modmuss50:
You didn’t update the transitive access winders template and then run the gradle task to generate the final file.

qouteall:
After rebasing and push it will show unrelated changes, but changing base to 1.20 and then change back to 1.19.4 makes them disappear. Weird.

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
Technici4n: I still believe that's incorrect if the following code runs:
```java
// Click first
KeyBinding.setKeyPressed(key2, true);
KeyBinding.onKeyPressed(key2);
// Immediately release (same "input tick")
KeyBinding.setKeyPressed(key2, false);
```
In that case, we will have `timesPressed > 0`, but `isPressed()` will still be false.

qouteall: Multiple callbacks can execute in same tick. I see.

qouteall: Fixed

Technici4n:
(FYI you can run `gradlew checkstyleMain checkstyleTestmod` to run checkstyle locally.)

Technici4n (approved):
Looks great now, thanks a lot!

modmuss50 (approved):

Benonardo (approved):

fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/mixin/event/interaction/client/MinecraftClientMixin.java
```diff
@@ -157,4 +172,43 @@ private void injectUseEntityCallback(CallbackInfo ci, Hand[] hands, int i1, int
...
+	)
+	private void injectHandleInputEventsForPreAttackCallback(CallbackInfo ci) {
+		int attackKeyPressCount = ((KeyBindingAccessor) options.attackKey).fabric_getTimesPressed();
+
+		if (options.attackKey.isPressed() || attackKeyPressCount != 0) {
+			fabric_attackCancelled = ClientPreAttackCallback.EVENT.invoker().onClientPlayerPreAttack(
+					(MinecraftClient) (Object) this, player, attackKeyPressCount
+			);
+		}
+		else {
```
modmuss50: Still a checkstyle issue to fix here.

PR #3047, opened 2023-05-06
Fix `getName` for builtin packs not returning an internal name

Technici4n:
`getName` should return an identifier rather than a user-facing name. In particular, the stringified version of the `Text` used for display is bad because lang resources may not be loaded yet.

#### List of changes
- Change `getName` of the "Fabric Mods" pack to `fabric`. This matches what appears in `options.txt` (the id of the `ResourcePackProfile`).
- Replace the `Identifier id` in `ModNioResourcePack` by a `String`, since it is only used as a string anyway. This makes the next change possible.
- Remove `name` from `ModNioResourcePack`. Change `getName` of the mod resource packs to be exactly the modid. This changes the log entry to the modid instead of the mod name.

#### With these changes
- `options.txt`:
```resourcePacks:["vanilla","fabric","fabric-resource-loader-v0-testmod:test"]```
- log line for the reload:
```[10:47:57] [Render thread/INFO] (Minecraft) Reloading ResourceManager: vanilla, fabric (fabric-renderer-api-v1-testmod, fabric-entity-events-v1, fabric-particles-v1, fabric-crash-report-info-v1, fabric-game-rule-api-v1, fabric-networking-v0, fabric-recipe-api-v1, fabric-blockrenderlayer-v1, fabric-command-api-v1, fabric-networking-api-v1, fabric-rendering-data-attachment-v1, fabric-loot-tables-v1, fabric-renderer-indigo, fabric-lifecycle-events-v1, fabric-transfer-api-v1, fabric-data-generation-api-v1, fabric-convention-tags-v1, fabric-resource-loader-v0, fabric-object-builder-api-v1-testmod, fabric-renderer-api-v1, fabric-message-api-v1, fabric-screen-handler-api-v1-testmod, fabric-screen-handler-api-v1, fabric-models-v0, fabric-key-binding-api-v1, fabric-sound-api-v1-testmod, fabric-rendering-v1-testmod, fabric-keybindings-v0, fabric-command-api-v2, fabric-data-gen-api-v1-testmod, fabric-object-builder-api-v1, fabric-renderer-registries-v1, fabric-biome-api-v1, fabric-rendering-fluids-v1, fabric-item-api-v1-testmod, fabric-resource-loader-v0-testmod, fabric-rendering-v1, fabric-registry-sync-v0, fabric-containers-v0, fabric-content-registries-v0, fabric-mining-level-api-v1, fabric-networking-api-v1-testmod, fabric-rendering-fluids-v1-testmod, fabric-events-lifecycle-v0, fabricloader, fabric-gametest-api-v1, fabric-api-base, fabric-game-rule-api-v1-testmod, fabric-item-api-v1, fabric-screen-api-v1, fabric-commands-v0, fabric-models-v0-testmod, fabric-transitive-access-wideners-v1, fabric-events-interaction-v0, fabric-resource-conditions-api-v1, fabric-rendering-v0, fabric-api, fabric-client-tags-api-v1, fabric-sound-api-v1, fabric-dimensions-v1, fabric-key-binding-api-v1-testmod, fabric-block-api-v1, fabric-item-group-api-v1, fabric-api-lookup-api-v1, fabric-loot-api-v2), fabric-resource-loader-v0-testmod:test```

The old log line would have been `vanilla, Fabric Mods (mod name 1, mod name 2, etc...), <stringified version of the display name text used by the test builtin pack>`

### DIFF
--- a/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
+++ b/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
@@ -40,7 +40,6 @@ import net.minecraft.nbt.NbtList;
 import net.minecraft.nbt.NbtString;
 import net.minecraft.resource.ResourcePack;
 import net.minecraft.resource.ResourcePackProfile;
-import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.impl.resource.loader.ModNioResourcePack;
 import net.fabricmc.fabric.impl.resource.loader.ModResourcePackCreator;
@@ -71,7 +70,7 @@ public class GameOptionsMixin {
 		}
 
 		File trackerFile = new File(dataDir, "fabricDefaultResourcePacks.dat");
-		Set<Identifier> trackedPacks = new HashSet<>();
+		Set<String> trackedPacks = new HashSet<>();
 
 		if (trackerFile.exists()) {
 			try {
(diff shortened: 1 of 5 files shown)


PepperCode1 (approved):

Juuxel:
Will the position of `Fabric Mods` in the resource pack list change now if a player has previously moved it?

LambdAurora:
> Will the position of `Fabric Mods` in the resource pack list change now if a player has previously moved it?

~~Yes it will. But~~ tbh, that pack shouldn't even exist to users and be hidden inside the Default pack.

Technici4n:
The id of the `ResourcePackProfile` is what's stored inside `options.txt`, and this PR doesn't touch that. So not this doesn't change the position (I just tested this).

LambdAurora:
Ah, how is the ID of the `ResourcePackProfile` determined again?
Memory is not happy

Ok, looked at the code, Fabric Mods is hardcoded to `fabric` and other built-in packs are the `identifier` given when registering them.

Juuxel (approved):

modmuss50 (approved):

Issue #2923, opened 2023-02-22
Transfer API crashes are hard to follow


MissPotato:
Preface: I'm not a coder. Just a pack dev. In the attached crash log a hopper at 610,66,32 is trying to push items into a container.  The container is claiming to have -1024 items in it. This results in the game seemingly blaming the hopper as the source of the crash, as it is the object ticking.  While this may technically be true, the container is the one with the overflow. The actual overflow was caused by a rest config in Extended Drawers mod. Once I was able to figure out what block was being pushed into the fix was really easy.

That's the part I'm finding hard to follow. This log doesn't indicate at seemingly any point the container with the invalid size. I'm not even sure if it's possible, as again, I'm not a coder, but it would be very useful for us players to know what block is being pushed into / pulled from on top of the actual ticking block.

Specifically, I'm talking about the Details section, where it lists the same hopper twice. (I assume one is the block, and one is the entity.)  The block entity being interacted with being printed here as well would make this crash far easier to understand, as it seems like everything in the stacktrace is Fabric API level. 

This https://github.com/FabricMC/fabric/issues/2201 inspired me to look at the crash from a different light, and instead of figuring out why the hopper was pushing a negative stack, figure out what it was pushing into. 

https://pastebin.com/mPL3c4A8

Technici4n:
This was partially addressed in #3049. Mods can also add relevant logging code following a similar pattern. :)

MissPotato:
Thank you so much!

PR #3049, opened 2023-05-06
Improve transfer debug messages

Juuxel:
## Changes

- Added `toString` to all default `Storage`, `ContainerItemContext` and `Transaction` implementations
- Removed `Impl` suffix from `ItemVariantImpl.toString` and `FluidVariantImpl.toString`. Since they are the only implementations, the suffix offers no meaningful info for callers.
- In `StorageUtil` and `FluidStorageUtil` methods for moving resources, exceptions are now caught are more context is added to them with `CrashReport`.
  - Open question: Would it make sense to catch the precondition exceptions thrown by the methods themselves?

## Example
Helps a bit with #2923, but without adding context like dimensions and positions everywhere, it's hard to fix completely.

With this PR, the error message would've looked somewhat like:

```
java.lang.IllegalArgumentException: Amount may not be negative, but it is: -1024
     ....stacktrace...

A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Move details --
Details:
    Input storage: SingleStackStorage[10 minecraft:diamond]
    Output storage: com.example.badmod.BrokenStorage@12345678
    Filter: some lambda
    Max amount: 64
    Transaction: Transaction[depth=0, lifecycle=NONE, thread=Server thread]
```

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/PlayerContainerItemContext.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/PlayerContainerItemContext.java
@@ -59,6 +59,7 @@ public class PlayerContainerItemContext implements ContainerItemContext {
 
 	@Override
 	public String toString() {
-		return "PlayerContainerItemContext[" + playerWrapper + "]";
+		return "PlayerContainerItemContext[%d %s %s/%s]"
+				.formatted(slot.getAmount(), slot.getResource(), playerWrapper, slot);
 	}
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/PlayerContainerItemContext.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/PlayerContainerItemContext.java
@@ -56,4 +56,9 @@ public class PlayerContainerItemContext implements ContainerItemContext {
 	public List<SingleSlotStorage<ItemVariant>> getAdditionalSlots() {
 		return playerWrapper.getSlots();
 	}
+
+	@Override
+	public String toString() {
+		return "PlayerContainerItemContext[" + playerWrapper + "]";
+	}
 }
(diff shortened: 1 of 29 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/fluid/CauldronStorage.java
```diff
@@ -47,6 +47,10 @@
 public class CauldronStorage extends SnapshotParticipant<BlockState> implements SingleSlotStorage<FluidVariant> {
 	// Record is used for convenient constructor, hashcode and equals implementations.
 	private record WorldLocation(World world, BlockPos pos) {
```
apple502j, line 50: (hmm, why are we using this and not vanilla `GlobalPos`?)

Technici4n: This was introduced in 1.16 when there was no `GlobalPos`. Don't think it would be shorter with `GlobalPos` anyway as we'd need more code to access the world...

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/PlayerContainerItemContext.java
```diff
@@ -56,4 +56,9 @@ public long insertOverflow(ItemVariant itemVariant, long maxAmount, TransactionC
 	public List<SingleSlotStorage<ItemVariant>> getAdditionalSlots() {
 		return playerWrapper.getSlots();
 	}
+
+	@Override
+	public String toString() {
+		return "PlayerContainerItemContext[" + playerWrapper + "]";
```
Technici4n: Would be nice to have the amount and resource to match the other contexts.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/context/SingleSlotContainerItemContext.java
```diff
@@ -46,4 +46,9 @@ public long insertOverflow(ItemVariant itemVariant, long maxAmount, TransactionC
 	public List<SingleSlotStorage<ItemVariant>> getAdditionalSlots() {
 		return Collections.emptyList();
 	}
+
+	@Override
+	public String toString() {
+		return "SingleSlotContainerItemContext[" + slot + "]";
```
Technici4n: Would be nice to have the amount and resource to match the other contexts.

Technici4n (commented):
I wonder if we should print the positions for the vanilla `Inventory` instances. That's probably not included in `toString()`

Technici4n (commented):
The non-trivial items in the crash report should ideally use `.add("xxx", () -> xxx)` in case they crash. (The `CrashCallable` override)

Technici4n (approved):

modmuss50 (approved):

PR #3050, opened 2023-05-06
Terminally deprecate `EventFactory` `isProfilingEnabled` and `invalidate`

### LINKED ISSUE #2849: Deprecate and remove EventFactory#invalidate
`EventFactory#invalidate` is used to enable profiling for event handlers added before the profiling is enabled. However, this requires the events to be stored globally, causing memory leaks. While the leak can be mitigated using weakref, it's still a bad practice, given events are created thousands of times during the lifetime.

This method is not used by any mods. Therefore I suggest `forRemoval=true` deprecation of this method, and dropping it in the next big Minecraft version (perhaps along with `NbtType` const).

See #2848

Technici4n:
This functionality is useless clutter, using a proper profiler is much better.

Fixes #2849.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
@@ -110,12 +110,9 @@ public final class EventFactory {
 	}
 
 	/**
-	 * Get the listener object name. This can be used in debugging/profiling
-	 * scenarios.
-	 *
-	 * @param handler The listener object.
-	 * @return The listener name.
+	 * @deprecated This is not to be used in events anymore.
 	 */
+	@Deprecated
 	public static String getHandlerName(Object handler) {
 		return handler.getClass().getName();
 	}
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
@@ -26,26 +26,7 @@ import net.fabricmc.fabric.impl.base.event.EventFactoryImpl;
  * Helper for creating {@link Event} classes.
  */
 public final class EventFactory {
-	private static boolean profilingEnabled = true;
-
 	private EventFactory() { }
-
-	/**
-	 * @return True if events are supposed to be profiled.
-	 */
-	public static boolean isProfilingEnabled() {
-		return profilingEnabled;
-	}
-
-	/**
-	 * Invalidate and re-create all existing "invoker" instances across
-	 * events created by this EventFactory. Use this if, for instance,
-	 * the profilingEnabled field changes.
-	 */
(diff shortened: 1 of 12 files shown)


fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -138,4 +119,24 @@ public static <T> Event<T> createWithPhases(Class<? super T> type, Function<T[],
 	public static String getHandlerName(Object handler) {
 		return handler.getClass().getName();
 	}
+
+	/**
+	 * @deprecated Always returns {@code false}, do not use.
+	 */
+	@Deprecated(forRemoval = true)
```
modmuss50: I dont think we should set `forRemoval`

Technici4n: I made it `forRemoval` to give a strong warning (especially since `invalidate` should be removed, imo), but maybe this one can stay forever (it's a trivial method anyway).

modmuss50 (approved):

SquidDev:
> This functionality is useless clutter, using a proper profiler is much better.

One of the nice things about the vanilla profile, that Spark/async-profiler don't make as easy, is that it can provides per-tick profiling data. This is very useful when trying to track down stutter/lag spikes. I've definitely benefitted from this feature in the past.

That said, realise this does remove a lot of code, so it may not be worth keeping around just for that.

modmuss50:
Im not sure if anyone actully ever used this feature, there was no public API to enable it. I do think the vanilla profiling tools are under used (By players and modders), maybe there are things we can do to help increase its useage.

Sampler can provider per tick profiles, and can also only trigger on ticks that take longer than a given time.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
```diff
@@ -138,4 +119,24 @@ public static <T> Event<T> createWithPhases(Class<? super T> type, Function<T[],
 	public static String getHandlerName(Object handler) {
```
modmuss50, line 116: I think we should also deprecate `getHandlerName`

SquidDev:
> Im not sure if anyone actully ever used this feature, there was no public API to enable it. 

Was it not enabled by default?

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/EventFactory.java
Technici4n: I didn't even know this was a thing...

Technici4n:
> Was it not enabled by default?

Ah, I just realized... Then it should maybe be set to `true` in the PR? Not that it's really useful anyway...

modmuss50:
Oh, apparenly it was enabled by default, my bad. I guess there is a small pref gain from this change then 👍

PR #3051, opened 2023-05-07
Convention Tags Improvements

### LINKED ISSUE #2831: In convention tags, the tag c:water_buckets should be separated into c:water_buckets and c:entity_water_buckets.
The entity buckets are functionally different and don't have a recipe remainder set, this makes the tag useless for recipes and other things.

If you want to argue that the entity buckets are a water_bucket plus an entity why doesn't c:buckets have all the buckets in it? A water_bucket or lava_bucket is just a bucket plus water or lava.

Technici4n:
Various convention tag improvements.

- Cleanup code:
  - Fix `.gitignore` not properly including the generated file.
  - Use `copy` instead of rebuilding tags entirely when they exist both for blocks and for items.
  - Reorder tags a bit, especially block tags so they match item tags.
- Add item version of the block tags where it made sense.
- Add tags for budding blocks: `c:budding_blocks`, `c:buds`, `c:clusters`.
- Add resource category tags: `c:dusts`, `c:gems`, etc...
- Add `c:wooden_barrels`.
- Fix https://github.com/FabricMC/fabric/issues/2831: Move entity buckets to `c:entity_water_buckets`.
- Add `c:villager_job_sites` and the various sandstone tags from #2268.

**Since these tags are mostly based on community usage, I would appreciate that casual readers have a look and verify that everything makes sense. If you disagree with anything, don't hesitate to say so.**


modmuss50 (approved):

oliviathevampire:
Idk if we already have this, but tags for terracotta, concrete, concrete dust, glazed terracotta and mostly all the colored stuff, would be super nice

Technici4n:
`#minecraft:terracotta` already exists for example. I am not sure about the others.

fabric-convention-tags-v1/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -55,11 +54,76 @@ protected void configure(RegistryWrapper.WrapperLookup registries) {
...
+				.add(Blocks.BARREL);
+
+		generateBuddingTags();
+
+		getOrCreateTagBuilder(ConventionalBlockTags.VILLAGER_JOB_SITES)
+				.add(Blocks.BARREL)
+				.add(Blocks.BLAST_FURNACE)
+				.add(Blocks.BREWING_STAND)
+				.add(Blocks.CARTOGRAPHY_TABLE)
+				.add(Blocks.CAULDRON)
```
apple502j: Needs cauldron variants (though not the tag, as the game does not use it for this purpose)

apple502j (changes requested):

PR #3056, opened 2023-05-10
Add and test for missing method overrides in FabricBlockSettings

modmuss50:
5 mins into porting my mod to 1.20, I found this problem. Its amazing what you find when you actually use something 😆 

This should be a none breaking change, also added a basic test to ensure it doesnt get like this again.

### DIFF
--- a/fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/FabricBlockSettingsTest.java
+++ b/fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/FabricBlockSettingsTest.java
@@ -17,33 +17,34 @@
 package net.fabricmc.fabric.test.object.builder;
 
 import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+
+import org.objectweb.asm.Opcodes;
 
 import net.minecraft.block.AbstractBlock;
 
 import net.fabricmc.api.ModInitializer;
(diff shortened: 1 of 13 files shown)


haykam821:
Could these tests be expanded to `FabricItemSettings` as well? There might be other classes that could benefit from these changes but item settings seem like the most prone to changes.

modmuss50:
> Could these tests be expanded to FabricItemSettings as well?

Yes, good idea. Its kinda a pain as its in a diffrent module. I may just copy/paste the tests over to the other module 🤔

Technici4n:
Ideally the whole thing would be replaced by an injected interface

fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/FabricBlockSettingsTest.java
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		throw new IllegalStateException("Missing method overrides in FabricBlockSettings: " + String.join(", ", missingMethods));
+	}
+
+	private List<String> getMethods(Class<?> clazz) {
+		List<String> methods = new ArrayList<>();
+
+		for (final Method method : clazz.getDeclaredMethods()) {
+			if (method.getReturnType() != clazz) {
```
liach: I would recommend just checking `getMethods()`, and report any non-bridge method that returns a vanilla `Settings`.

modmuss50: Good idea, its much cleaner now 👍

PR #3061, opened 2023-05-13
Fix screen api passing wrong MatrixStack to render events

isXander:


Technici4n (approved):
Looks plausible

isXander:
Obviously tested. Non-working before. Working now.
The issue was the first matrix stack (that was previously provided) had a translation of z -2000, which seems to prevent rendering (near plane?). A second was created without this that is given to screens and toasts (and I guess `InGameHud`)

modmuss50 (approved):
LGTM, thanks for fixing 👍

PR #3060, opened 2023-05-13
Some TAWs for worldgen

maityyy:
Many worldgen mods use this

PR #3064, opened 2023-05-15
Fix last few transitive access wideners missing the `transitive-` prefix

maityyy:
Also ~~checkstyle and~~ fix my wrong in #3060

### DIFF
--- a/fabric-transitive-access-wideners-v1/build.gradle
+++ b/fabric-transitive-access-wideners-v1/build.gradle
@@ -45,6 +45,8 @@ task generateAccessWidener {
 		}
 
 		file('src/main/resources/fabric-transitive-access-wideners-v1.accesswidener').text = String.join('\n', lines) + '\n'
+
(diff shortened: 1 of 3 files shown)


maityyy:
example
image

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -124,4 +123,22 @@ ClassNode loadClass(Path path) {
 	return node
 }
 
+task validateTemplateAccessWidener {
```
Technici4n: I think it would be better to check the generated file (using the `lines` variable) at the end of `generateAccessWidener` in case we make a mistake there too.

Technici4n (changes requested):
IMO the reformatting lowers the quality of the git history for no real benefit, and should be undone.

Technici4n (approved):

modmuss50 (approved):

PR #3070, opened 2023-05-21
Fix FabricBlockLootTableProvider breaking depending on the mapping set

### LINKED ISSUE #3069: FabricBlockLootTableProvider breaks depending on the mapping set
This affects all Fabric APIs from Minecraft 1.19.3 onwards.

## A Brief Background

`FabricBlockLootTableProvider` implements `FabricLootTableProvider`. `FabricLootTableProvider` extends a `Consumer` which is used to register the loot table entries to be written to the disk. The consumer itself is implemented within `FabricBlockLootTableProvider#accept` which implements the registration method.

https://github.com/FabricMC/fabric/blob/926dc7fcc3b8096d7c395ee351d23fbe6193839c/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java#L79-L114

## What is `#accept`?

It would be fine if `#accept` was just the name of the method passed into the `Consumer`. However, it is also the yarn mapping name for `BlockLootTableGenerator#accept`. In Yarn mappings, this is fine as `#accept` would then override `Consumer#accept` and implement perfectly. However, in other mapping sets, such as Mojmaps, the name is `BlockLootSubProvider#generate`. This requires a modder using Mojmaps to implement the `#accept` method from the Consumer and delegate to `BlockLootSubProvider#generate`.

This is an issue since it means that the code is reliant upon the mapping set. You need to add the method in non-Yarn mappings sets and remove the method in Yarn mapping sets, otherwise the call would infinitely recurse.

## Possible Solution

Because of how Fabric implements loot table providers, the main solution would be to change `FabricLootTableProvider` to extend `LootTableGenerator` and call the associated method. This would prevent any disambiguation between mapping sets as the method's name would then change according to which mapping set it was apart of, rather than be static from the `Consumer`.

Technici4n:
Fixes #3069.

Also removes unnecessary internal methods from `FabricLootTableProvider` by moving the `run` logic to an impl method.

This is not a breaking change assuming modders used the API as intended.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java
@@ -20,22 +20,24 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.BiConsumer;
 
 import com.google.common.collect.Sets;
 
 import net.minecraft.block.Block;
+import net.minecraft.data.DataWriter;
 import net.minecraft.data.server.loottable.BlockLootTableGenerator;
 import net.minecraft.loot.LootTable;
 import net.minecraft.loot.LootTables;
-import net.minecraft.loot.context.LootContextType;
 import net.minecraft.loot.context.LootContextTypes;
+import net.minecraft.registry.Registries;
 import net.minecraft.resource.featuretoggle.FeatureFlags;
 import net.minecraft.util.Identifier;
-import net.minecraft.registry.Registries;
 
 import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
 import net.fabricmc.fabric.api.datagen.v1.FabricDataOutput;
+import net.fabricmc.fabric.impl.datagen.loot.FabricLootTableProviderImpl;
 
 /**
  * Extend this class and implement {@link FabricBlockLootTableProvider#generate}.
@@ -64,16 +66,6 @@ public abstract class FabricBlockLootTableProvider extends BlockLootTableGenerat
 	 */
 	public void excludeFromStrictValidation(Block block) {
 		excludedFromStrictValidation.add(Registries.BLOCK.getId(block));
-	}
-
-	@Override
-	public LootContextType getLootContextType() {
-		return LootContextTypes.BLOCK;
-	}
-
-	@Override
-	public FabricDataOutput getFabricDataOutput() {
-		return output;
 	}
 
 	@Override
@@ -111,6 +103,11 @@ public abstract class FabricBlockLootTableProvider extends BlockLootTableGenerat
 				throw new IllegalStateException("Missing loot table(s) for %s".formatted(missing));
 			}
 		}
+	}
+
+	@Override
+	public CompletableFuture<?> run(DataWriter writer) {
+		return FabricLootTableProviderImpl.run(writer, this, LootContextTypes.BLOCK, output);
 	}
 
 	@Override
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLootTableProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLootTableProvider.java
@@ -16,29 +16,16 @@
 
 package net.fabricmc.fabric.api.datagen.v1.provider;
 
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.CompletableFuture;
 import java.util.function.BiConsumer;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Maps;
-import com.google.gson.JsonObject;
 import org.jetbrains.annotations.ApiStatus;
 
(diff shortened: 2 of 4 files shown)


ChampionAsh5357:
So, I don't understand how this fixes the issue, since it still persists. You still need to implement `#accept` to delegate to `#generate` in Mojang's mapping set. Once again, I will suggest just simply changing the `Consumer` to `LootTableGenerator` so that the generator is not dependent on the mapping set to determine whether the method is implemented or not.

modmuss50 (changes requested):
~~Yes, I fail to see how this resolves the issue.~~

Technici4n:
It should fix the issue by removing the requirement for an implementation of `Consumer<BiConsumer<...>>`. So `SimpleFabricLootTableProvider#accept` will remain `#accept` when mapped to intermediary as it does not override a Minecraft method, whereas `FabricBlockLootTableProvider#accept` will correctly get remapped to `method_<...>` then back to `generate` in mojmap environments. Since the `Consumer` superinterface is gone, there will not be a compile error.

I will investigate the alternative approach, however I should say that removing `getLootContextType`, `getFabricDataOutput`, and `run` from `FabricLootTableProvider` makes the code flow much easier to understand IMO, and should be done either way.

modmuss50:
Maybe im wrong, I did only have a quick look at it. I will need to get setup to actually test this.

ChampionAsh5357:
> It should fix the issue by removing the requirement for an implementation of Consumer<BiConsumer<...>>. So SimpleFabricLootTableProvider#accept will remain #accept when mapped to intermediary as it does not override a Minecraft method, whereas FabricBlockLootTableProvider#accept will correctly get remapped to method_<...> then back to generate in mojmap environments. Since the Consumer superinterface is gone, there will not be a compile error.

I think this is my bad as well. I set up Fabric for the first time on my machine to test it, and it seems I forgot to switch to the PR branch. Apologies, I believe the solution should work fine.

Technici4n:
I confirmed that this works by cherry-picking to 1.19.4, publishing to maven local via `gradlew clean build publishToMavenLocal` and then depending on it in AE2 (needed a 1.19.4 mojmap project). If you want to do that too, make sure to uncomment `repositories.mavenLocal()` in the Fabric API `build.gradle` and add maven local to your project's repositories. If you want to test in 1.19.4 you can use that branch: https://github.com/Technici4n/fabric/tree/fix-3069.

Could probably use extra testing by someone else, but now I'm fully certain that this fix works. Whether it should be backported to 1.19.4 is up for debate... :smile:

Technici4n:
Using `LootTableGenerator` would also work, indeed. It only has a single `accept` method.

Is it useful to have an `accept(BiConsumer<Identifier, LootTable.Builder> exporter)` method in `FabricLootTableProvider`, or not? I don't think one would ever use it. If not we can keep the current PR, if yes I will add it back.

For 1.19.4 I think replacing the `Consumer` superinterface by a `LootTableGenerator` is a good fix, however I'd like to clean up the code for 1.20.

EDIT: opened 1.19.4 version: #3071.

ChampionAsh5357:
> Is it useful to have an accept(BiConsumer<Identifier, LootTable.Builder> exporter) method in FabricLootTableProvider, or not? I don't think one would ever use it. If not we can keep the current PR, if yes I will add it back.

Since Fabric doesn't use Mojang's method of generator entries in a single loot table provider, I would say it really just depends on how much logic you want to delegate to the already existing implementations. The Fabric implementation only provides separation of loot table providers, so it's just purely a matter of compatibility if somebody wanted to use Mojang's method of registering a loot table provider instead.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/loot/LootTableProviderImpl.java
```diff
@@ -0,0 +1,81 @@
...
+import net.minecraft.loot.context.LootContextType;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.datagen.v1.FabricDataOutput;
+import net.fabricmc.fabric.api.datagen.v1.provider.FabricBlockLootTableProvider;
+import net.fabricmc.fabric.api.datagen.v1.provider.SimpleFabricLootTableProvider;
+import net.fabricmc.fabric.api.resource.conditions.v1.ConditionJsonProvider;
+import net.fabricmc.fabric.impl.datagen.FabricDataGenHelper;
+
+public class LootTableProviderImpl {
```
modmuss50: Nit: final + private constructor

Technici4n:
> > Is it useful to have an accept(BiConsumer<Identifier, LootTable.Builder> exporter) method in FabricLootTableProvider, or not? I don't think one would ever use it. If not we can keep the current PR, if yes I will add it back.
> 
> Since Fabric doesn't use Mojang's method of generator entries in a single loot table provider, I would say it really just depends on how much logic you want to delegate to the already existing implementations. The Fabric implementation only provides separation of loot table providers, so it's just purely a matter of compatibility if somebody wanted to use Mojang's method of registering a loot table provider instead.

Then I will add `LootTableGenerator` too, it doesn't hurt.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/loot/LootTableProviderImpl.java
Technici4n: It's an impl class so I didn't bother, why not though.

modmuss50: I guess, I think generally we try to, but its not a problem if its forgotten.

PR #3071, opened 2023-05-22
Fix FabricBlockLootTableProvider breaking depending on the mapping set (1.19.4 simplified edition)

Technici4n:
1.19.4 fix for #3069, as suggested by @ChampionAsh5357.


Technici4n:
We will go for the cleaner fix in 1.19.4 too: #3070.

PR #3074, opened 2023-05-24
Add metadata information to POMs

lineargraph:


build.gradle
```diff
@@ -509,6 +509,23 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
...
+			url = "https://github.com/FabricMC/fabric/blob/HEAD/LICENSE"
+		}
+	}
+	pom.developers {
+		developer {
+			name = "FabricMC"
+		}
+	}
+	pom.scm {
+		url = "https://github.com/FabricMC/fabric"
```
modmuss50, line 536: A few more things can be added to the scm as seen here: https://github.com/FabricMC/mapping-io/blob/master/build.gradle#L104

build.gradle
```diff
@@ -509,6 +509,23 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
 	}
 }
 
+def addPomMetadataInformation(MavenPom pom) {
```
modmuss50: Would be nice to also set name and url. description could possibly be driven by the fabric.mod.json file, but thats a little more effort.

build.gradle
```diff
@@ -509,6 +509,23 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
...
+	pom.licenses {
+		license {
+			name = "Apache 2.0"
```
modmuss50:
```suggestion
			name = "Apache-2.0"
```

According to the [docs](https://maven.apache.org/pom.html#Licenses) Using an [SPDX identifier](https://spdx.org/licenses/) as the license name is recommended.

build.gradle
modmuss50: issueManagement might be another useful one to add.

modmuss50 (changes requested):
If we are going to do this, I think we should make sure to populate it fully. See my comments and https://github.com/FabricMC/mapping-io/blob/master/build.gradle#L99 for some ideas.

build.gradle
```diff
@@ -442,7 +445,8 @@ dependencies {
 	productionRuntimeServer "net.fabricmc:fabric-installer:${project.installer_version}:server"
 }
 
-import net.fabricmc.loom.util.Platform
+
+import org.kohsuke.github.GHReleaseBuilder
```
modmuss50: No need to move these imports, I think its generally better to keep the imports close to where they are used.

build.gradle
```diff
@@ -509,6 +513,37 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
 	}
 }
 
+def addPomMetadataInformation(Project project, MavenPom pom) {
+	def modJsonFile = project.file("src/main/resources/fabric.mod.json")
+	if (!modJsonFile.exists())
+		modJsonFile = project.file("src/client/resources/fabric.mod.json")
+	def modJson = new JsonSlurper().parse(modJsonFile)
+	pom.name = modJson.name
```
modmuss50, line 519: Great idea 👍

build.gradle
```diff
@@ -509,6 +513,37 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
...
+	pom.url = "https://github.com/FabricMC/fabric/tree/${grgit.head().abbreviatedId}/${project.rootDir.relativePath(project.projectDir)}"
```
modmuss50: I would just hard code the URL to `https://github.com/FabricMC/fabric`, as this wont work if someone tries to build it outside of a github repo. (Such as when you just download the zip from github)

build.gradle
```diff
@@ -509,6 +513,37 @@ task runProductionAutoTestServer(type: JavaExec, dependsOn: [remapJar, remapTest
...
+	pom.description = modJson.description
+	pom.licenses {
+		license {
+			name = "Apache-2.0"
+			url = "https://github.com/FabricMC/fabric/blob/${grgit.head().abbreviatedId}/LICENSE"
```
modmuss50: Again, dont use git here.

build.gradle
```diff
@@ -668,7 +710,7 @@ if (signingEnabled) {
 	build.dependsOn signRemapJar
 }
 
-import org.kohsuke.github.GHReleaseBuilder
+
```
modmuss50: Revert these import changes.

modmuss50 (changes requested):
This is looking good, nice work reading it from the FMJ. Im not too keen on using grgit like this as people can and do build it without being in a git repo.

lineargraph:
@modmuss50 I now made grgit optional. I can also entirely remove it, if you want that.

modmuss50:
> @modmuss50 I now made grgit optional. I can also entirely remove it, if you want that.

I think it would be best, helps enusre that its consitant between runs.

modmuss50 (approved):

PR #3077, opened 2023-05-24
Set player.currentScreenHandler before invoking ExtendedScreenHandlerFactory

### LINKED ISSUE #2963: ExtendedScreenHandlerFactory is not passed ScreenHandler for context

modmuss50:
Fixes #2963

Alternative fix from https://github.com/FabricMC/fabric/pull/2964

### DIFF
--- a/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java
+++ b/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java
@@ -61,6 +61,8 @@ public final class Networking {
 			return;
 		}
 
+		player.currentScreenHandler = handler;
+
 		PacketByteBuf buf = new PacketByteBuf(Unpooled.buffer());
 		buf.writeIdentifier(typeId);
 		buf.writeVarInt(syncId);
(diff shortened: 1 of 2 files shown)


Technici4n:
Could use a comment, maybe something like `// Set the screen handler, so the packet writing method can access it through the player. Vanilla sets it too but a bit too late.`. Could be nice in the future to know why this exists.

Technici4n:
We could even set it in `fabric_storeOpenedScreenHandler` directly in the mixin, and get rid of the `ThreadLocal` usage and of the cleanup mixin. Otherwise this PR is also fine. Your choice. :)

modmuss50:
> We could even set it in `fabric_storeOpenedScreenHandler` directly in the mixin, and get rid of the `ThreadLocal` usage and of the cleanup mixin. Otherwise this PR is also fine. Your choice. :)

Done, good idea 👍

Technici4n (approved):

Juuxel (approved):

PR #3078, opened 2023-05-25
fix: floral tag missing meadow biome

maityyy:
https://minecraft.fandom.com/wiki/Meadow
> filled with patches of flowers


Technici4n (approved):

Juuxel (approved):

modmuss50 (approved):

Technici4n:
So, do we add lush caves to this tag or not?

maityyy:
> So, do we add lush caves to this tag or not?

I am not sure, but of all the blocks in this biome, only `flowering_azalea_leaves` is a flower, since it is in the `flowers` tag

maityyy:
More likely no than yes

Technici4n:
Alright then this is fine

PR #3083, opened 2023-05-27
Resource Loader v1

Technici4n:
A much needed update to resource loader.

## Resource Reload Events
- `ClientResourceReloadEvents` and `ServerResourceReloadEvents` are added to hold new events.
- `START_RELOAD` fires right before resource reloaders start running, and can for example be used to invalidate a cache before the resource reloader recomputes it.
- `END_RELOAD` fires right when the reload completes (successfully or with a failure).
- The exact place where the start event should be invoked is up for debate. Currently it is invoked after the resource reloaders are collected on the server side. Note that it's technically possible for `END_RELOAD` to fire without `START_RELOAD` having fired, if case something fails before `START_RELOAD`.
- `REGISTER_RELOADERS`: see below.

## Resource Reloaders
- Resource reloader registration now happens in the `REGISTER_RELOADERS` events.
- They still have an ID, however the ID is passed alongside the resource reloader at registration time. (Previously `getFabricId` had to be overridden).
- Resource reloaders are instantiated once per reload on the server and stored inside `DataPackContents`; on the client they are instantiated once and then stored inside `MinecraftClient`.
- Resource reloaders can be accessed from the data pack contents, the client and the server thanks to interface injection of `ResourceReloaderHolder`.
- The event allows `DataPackContents` and `MinecraftClient` to be captured for cross-reloader talk during the apply phase.
- Arbitrary ordering of reloaders was implemented (will use the same backend as event phases). Previously it was not possible to order modded reloaders before vanilla's. This was added in a fully backwards-compatible way.
- Note 1: the client-side `REGISTER_RELOADERS` event only fires exactly once at the start of the game, but I figured it would still be nicer for consistency to have an event.
- Note 2: the v0 API is bridged via `LegacyReloaderHolder` because the new client event is client-only, whereas the old API was side agnostic (since it did not provide any context whatsoever).

Here is an example of using this new API for a client reloader. I am not fully convinced by this usage pattern yet:
```java
ClientResourceReloadEvents.REGISTER_RELOADERS.register(context -> {
	context.addReloader(SpecificModelReloadListener.ID, new SpecificModelReloadListener());
	context.addReloaderOrdering(ResourceReloaderKeys.BAKED_MODELS, SpecificModelReloadListener.ID);
	context.addReloaderOrdering(SpecificModelReloadListener.ID, ResourceReloaderKeys.ENTITY_RENDER_DISPATCHER);
});

// Later on:
SpecificModelReloadListener myListener = (SpecificModelReloadListener) MinecraftClient.getInstance().getResourceReloader(SpecificModelReloadListener.ID);
```

For server reloaders however, this matches what vanilla does, and is much nicer than what the v0 API provides today.

## Resource Packs
- Builtin packs: functionality is kept unchanged, however providing a display name is now required.
- Currently they are registered by calling `ResourcePackHelper`, perhaps a better name should be found?
- Add new events to directly register `ResourcePack` instances. See `ResourcePackRegistrationEvents#afterMods(ResourceType)` and `afterAll(ResourceType)`.
 - These events are side-agnostic. Should we provide more (side-dependent) context to them? I am not convinced that this would be necessary.
 - These events provide a `ResourceManager` that allows inspecting the previous packs. This can be useful for virtual resource packs, among other things, however note that this API does **not** provide the virtual resource pack implementation itself.
- Add `ResourcePackHelper#createModResourcePack` to directly instantiate mod nio resource packs. This can be useful to access resources in a mod JAR outside of the normal resource loading lifecycle.

## Further possibilities
- We could add an API to register custom `ResourcePackProvider`s or `ResourcePackProfile`s. This can be left for a future PR imo. This would either be an event (maybe in `ResourcePackRegistrationEvents`), or a static registration in `ResourcePackHelper`.

## TODO items
- [ ] Does v1 need to implement `ModResourcePack` on its builtin packs too?
- [ ] Possibly find a better name for `ResourcePackHelper`.
- [ ] Possibly eliminate the action type from `ModNioResourcePack` entirely. Or can also be left for a future PR.
- [ ] Add tests for `ResourcePackRegistrationEvents`
- [ ] Consider adding `isFirst` to the context of `ResourcePackRegistrationEvents`. I think it can be left for later, it's a bit annoying.
- [ ] Should we add `ResourceReloaderKeys` for all the vanilla reloaders? Many are missing on the client side.
- [ ] The old resource reloader API broke ties by registration order, whereas the new APIs break ties by identifier. This is a potential breaking changes for mods. However they were not correctly specifying dependencies as required by the API, so this might be acceptable. A workaround might be to break ties by registration order when using the legacy bridge, and by identifier when using the new API.
- [ ] Add javadoc note saying that END_RELOAD can technically fire even if START_RELOAD did not fire.


fabric-models-v0/src/testmodClient/java/net/fabricmc/fabric/test/model/SpecificModelReloadListener.java
```diff
@@ -16,23 +16,15 @@
...
-import net.minecraft.util.Unit;
-import net.minecraft.util.profiler.Profiler;
 
 import net.fabricmc.fabric.api.client.model.BakedModelManagerHelper;
-import net.fabricmc.fabric.api.resource.IdentifiableResourceReloadListener;
-import net.fabricmc.fabric.api.resource.ResourceReloadListenerKeys;
 
-public class SpecificModelReloadListener extends SinglePreparationResourceReloader<Unit> implements IdentifiableResourceReloadListener {
-	public static final SpecificModelReloadListener INSTANCE = new SpecificModelReloadListener();
+public class SpecificModelReloadListener implements SynchronousResourceReloader {
```
Technici4n, line 27: `getSpecificModel` needs a null-check on the model. This will make the error happen at load time rather than at render time.

fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/event/SortableNode.java
```diff
@@ -14,10 +14,15 @@
...
+import java.util.List;
 
-public interface FabricLifecycledResourceManager {
-	ResourceType fabric_getResourceType();
+public abstract class SortableNode<N extends SortableNode<N>> implements Comparable<N> {
+	public final List<N> subsequentNodes = new ArrayList<>();
+	public final List<N> previousNodes = new ArrayList<>();
+	int visitStatus = 0; // 0: not visited, 1: visiting, 2: visited
+
+	protected abstract void printCycleWarning(N otherNode);
```
Technici4n: Actually it would be better to print all the SCCs with size > 1 in the `NodeSorting` method directly. Maybe it can have what it is sorting as a parameter (e.g. `"Found cycle(s) when sorting %s:"` where `%s` could be either `"Event phases"` or `"Resource reloaders"`).

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+import net.minecraft.resource.ResourceManager;
+import net.minecraft.resource.ResourceReloader;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.api.resource.loader.v1.ResourceReloaderHolder;
+
+public final class ClientResourceReloadEvents {
```
apple502j, line 31: All classes should have a javadoc, IMO.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	/**
+	 * Event to register resource reloaders for client resource packs (i.e. {@link ResourceType#CLIENT_RESOURCES}).
```
apple502j, line 33: I think this gets cut at `i.e.` when in summary.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	 * Event that fires at the start of each resource reload, after the resource manager is updated, but before resource reloaders run.
+	 */
+	public static final Event<StartReload> START_RELOAD = EventFactory.createArrayBacked(StartReload.class, callbacks -> context -> {
+		for (StartReload callback : callbacks) {
+			callback.onStartResourceReload(context);
+		}
+	});
+
+	/**
+	 * Event that fires right when the resource reloaders finish running, but before subsequent cleanup has been performed.
```
apple502j, line 57: I don't think "has been performed" is appropriate here, you can just use "occurs"

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	 * For example, the world renderer was not refreshed yet when this event fires.
+	 */
+	public static final Event<EndReload> END_RELOAD = EventFactory.createArrayBacked(EndReload.class, callbacks -> (context, success) -> {
+		for (EndReload callback : callbacks) {
+			callback.onEndResourceReload(context, success);
+		}
+	});
+
+	@FunctionalInterface
+	public interface RegisterReloaders {
```
apple502j, line 67: Needs javadoc.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+		void registerResourceReloaders(RegisterContext context);
+	}
+
+	@FunctionalInterface
+	public interface StartReload {
```
apple502j, line 72: Needs javadoc.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	public interface Context {
+		/**
+		 * Provides access to the new resource manager.
+		 */
+		ResourceManager getResourceManager();
+
+		/**
+		 * Provides access to the minecraft client.
+		 */
+		MinecraftClient getMinecraftClient();
```
apple502j, line 97: `getClient` is probably enough. There's no other client applicable here.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/reloader/MinecraftClientMixin.java
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * Fire end reload event for the first reload.
+	 */
+	@Inject(
+			method = "method_24040",
+			at = @At("HEAD")
+	)
+	private void onFirstReloadEnd(Optional<Throwable> error, CallbackInfo ci) {
+		ClientResourceReloadEvents.END_RELOAD.invoker().onEndResourceReload(new ClientReloadContext(), error.isEmpty());
```
apple502j, line 88: I wonder if ClientReloadContext should be singletoned, since it holds no fields.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackHelper.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.resource.ResourcePack;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.text.Text;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.resource.loader.BuiltinResourcePackHolder;
+import net.fabricmc.fabric.impl.resource.loader.ModNioResourcePack;
+import net.fabricmc.loader.api.ModContainer;
+
+// TODO: find better and more future-proof name.
```
apple502j, line 30: Future proof in what way? In case it gets non-packy stuff?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackHelper.java
```diff
@@ -0,0 +1,61 @@
...
+public final class ResourcePackHelper {
+	/**
+	 * Registers a built-in resource pack.
+	 *
+	 * <p>A built-in resource pack is an extra resource pack provided by your mod which is not always active, it's similar to the "Programmer Art" resource pack.
+	 *
+	 * <p>Why and when to use it? A built-in resource pack should be used to provide extra assets/data that should be optional with your mod but still directly provided by it.
+	 * For example, it could provide textures of your mod in another resolution, or could allow to provide different styles of your assets.
+	 *
+	 * <p>The path in which the resource pack is located is in the mod JAR file under the {@code "resourcepacks/<id path>"} directory. {@code id path} being the path specified
```
apple502j, line 40: Should document where this is in a source

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackHelper.java
```diff
@@ -0,0 +1,61 @@
...
+	 * @param displayName    the display name of the resource pack, should include mod name for clarity
+	 * @param activationType the activation type of the resource pack
+	 * @return {@code true} if successfully registered the resource pack, else {@code false}
+	 */
+	public static boolean registerBuiltinResourcePack(Identifier id, ModContainer container, Text displayName, ResourcePackActivationType activationType) {
+		return BuiltinResourcePackHolder.registerBuiltinResourcePack(id, "resourcepacks/" + id.getPath(), container, displayName, activationType);
+	}
+
+	@Nullable
+	public static ResourcePack createModResourcePack(ModContainer container, ResourceType type) {
```
apple502j, line 54: Javadoc?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackRegistrationEvents.java
```diff
@@ -0,0 +1,89 @@
...
+	 */
+	public static Event<AfterAll> afterAll(ResourceType resourceType) {
+		return afterAllEvents.computeIfAbsent(resourceType, k -> EventFactory.createArrayBacked(AfterAll.class, listeners -> context -> {
+			for (AfterAll listener : listeners) {
+				listener.registerPacksAfterAll(context);
+			}
+		}));
+	}
+
+	public interface AfterMods {
```
apple502j, line 64: Javadoc.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackRegistrationEvents.java
```diff
@@ -0,0 +1,89 @@
...
+		void registerPacksAfterMods(Context context);
+	}
+
+	public interface AfterAll {
+		void registerPacksAfterAll(Context context);
+	}
+
+	public interface Context {
```
apple502j, line 72: Needs NonExtendable annotation.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourcePackRegistrationEvents.java
```diff
@@ -0,0 +1,89 @@
...
+		/**
+		 * Add a pack to the resource manager.
```
apple502j, line 74: "Adds"?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ResourceReloaderHolder.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.util.Identifier;
+
+/**
+ * Injected interface to grant access to modded resource reloaders by id.
+ *
+ * <p>When implemented on {@link DataPackContents} or {@link MinecraftServer},
+ * provides access to the reloaders registered through {@link ServerResourceReloadEvents.RegisterReloaders}.
+ *
+ * <p>When implemented on {@code MinecraftClient},
+ * provides access to the reloaders registered through {@code ServerResourceReloadEvents.RegisterReloaders}.
```
apple502j, line 33: copypaste error

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ServerResourceReloadEvents.java
```diff
@@ -0,0 +1,144 @@
...
+		 * <b>only for use in the apply phase of resource reloading or once the reload succeeded</b>.
+		 */
+		DataPackContents getDataPackContents();
+
+		/**
+		 * Provides access to the minecraft server. {@code null} for the initial resource loading,
+		 * and available for data pack reloads (using the {@code /reload} command).
+		 */
+		@Nullable
+		MinecraftServer getMinecraftServer();
```
apple502j, line 113: Just "getServer"?

deprecated/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/v0/ResourceManagerHelperImpl.java
```diff
@@ -0,0 +1,46 @@
...
+
+package net.fabricmc.fabric.impl.resource.loader.v0;
+
+import net.minecraft.resource.ResourceType;
+
+import net.fabricmc.fabric.api.resource.IdentifiableResourceReloadListener;
+import net.fabricmc.fabric.api.resource.ResourceManagerHelper;
+import net.fabricmc.fabric.impl.resource.loader.reloader.LegacyReloaderHolder;
+
+public class ResourceManagerHelperImpl implements ResourceManagerHelper {
```
apple502j, line 25: If this has a private constructor, it should be marked as final, just for consistency.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	public interface StartReload {
+		void onStartResourceReload(Context context);
+	}
+
+	@FunctionalInterface
+	public interface EndReload {
+		/**
+		 * @param success {@code true} if the reload was successful, {@code false} otherwise
+		 */
+		void onEndResourceReload(Context context, boolean success);
```
apple502j, line 81: Needs javadoc.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+		MinecraftClient getMinecraftClient();
+	}
+
+	/**
+	 * Extended context available to {@link RegisterReloaders} that allows registering reloaders.
+	 */
+	@ApiStatus.NonExtendable
+	public interface RegisterContext extends Context {
+		/**
+		 * Register a new resource reloader.
```
apple502j, line 106: Registers.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+		 */
+		void addReloader(Identifier identifier, ResourceReloader reloader);
+
+		/**
+		 * Request that <b>the apply phase</b> of one reloader be executed before <b>the apply phase</b> of another reloader.
```
apple502j, line 111: Requests.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+		 * Prepare phases happen in parallel, and cannot be ordered.
+		 *
+		 * <p>The identifiers in {@link net.fabricmc.fabric.api.resource.loader.v1.ResourceReloaderKeys} can be used to refer to vanilla resource reloaders.
+		 * Unless otherwise requested, reloaders will be ordered after vanilla resource reloaders.
+		 *
+		 * <p>Incompatible ordering constraints such as cycles will lead to inconsistent behavior:
+		 * some constraints will be respected and some will be ignored. If this happens, a warning will be logged.
+		 *
+		 * @param firstReloader The identifier of the reloader that should run before the other.
+		 * @param secondReloader The identifier of the reloader that should run after the other.
```
apple502j, line 121: The convention, I believe, is to start with lowercase and omit the period.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/api/resource/loader/v1/client/ClientResourceReloadEvents.java
```diff
@@ -0,0 +1,128 @@
...
+	 */
+	@ApiStatus.NonExtendable
+	public interface Context {
+		/**
+		 * Provides access to the new resource manager.
+		 */
+		ResourceManager getResourceManager();
+
+		/**
+		 * Provides access to the minecraft client.
```
apple502j, line 95: M should be capitalized.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ServerResourceReloadEvents.java
```diff
@@ -0,0 +1,144 @@
...
+import net.minecraft.resource.ResourceReloader;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.server.DataPackContents;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ServerResourceReloadEvents {
```
apple502j, line 33: Classes should have javadocs.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ServerResourceReloadEvents.java
```diff
@@ -0,0 +1,144 @@
...
+		MinecraftServer getMinecraftServer();
+	}
+
+	/**
+	 * Extended context available to {@link RegisterReloaders} that allows registering reloaders.
+	 */
+	@ApiStatus.NonExtendable
+	public interface RegisterContext extends Context {
+		/**
+		 * Register a new resource reloader.
```
apple502j, line 122: Registers.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ServerResourceReloadEvents.java
```diff
@@ -0,0 +1,144 @@
...
+		 */
+		void addReloader(Identifier identifier, ResourceReloader reloader);
+
+		/**
+		 * Request that <b>the apply phase</b> of one reloader be executed before <b>the apply phase</b> of another reloader.
```
apple502j, line 127: Requests.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/loader/v1/ServerResourceReloadEvents.java
```diff
@@ -0,0 +1,144 @@
...
+		 * Prepare phases happen in parallel, and cannot be ordered.
+		 *
+		 * <p>The identifiers in {@link ResourceReloaderKeys} can be used to refer to vanilla resource reloaders.
+		 * Unless otherwise requested, reloaders will be ordered after vanilla resource reloaders.
+		 *
+		 * <p>Incompatible ordering constraints such as cycles will lead to inconsistent behavior:
+		 * some constraints will be respected and some will be ignored. If this happens, a warning will be logged.
+		 *
+		 * @param firstReloader The identifier of the reloader that should run before the other.
+		 * @param secondReloader The identifier of the reloader that should run after the other.
```
apple502j, line 137: Param javadoc should be lowercase, no period

fabric-resource-loader-v1/src/main/resources/fabric-resource-loader-v1.accesswidener
```diff
@@ -4,3 +4,7 @@ accessible method net/minecraft/resource/NamespaceResourceManager getMetadataPat
 accessible method net/minecraft/resource/NamespaceResourceManager loadMetadata (Lnet/minecraft/resource/InputSupplier;)Lnet/minecraft/resource/metadata/ResourceMetadata;
 accessible field net/minecraft/resource/FileResourcePackProvider source Lnet/minecraft/resource/ResourcePackSource;
 accessible field net/minecraft/resource/ResourcePackManager providers Ljava/util/Set;
+accessible class net/minecraft/server/MinecraftServer$ResourceManagerHolder
+accessible field net/minecraft/resource/ReloadableResourceManagerImpl reloaders Ljava/util/List;
+
```
apple502j, line 9: Extraneous line?

apple502j (changes requested):

dima-dencep:
@Technici4n any progress?

Technici4n:
No, I don't think I will be finishing this.

PR #3087, opened 2023-05-29
Add a method to data generator entrypoints to register custom keys with priorities

ErrorCraft:
Adds an event to register custom keys with priorities for data generation.

The name `JsonKeySortOrderCallback` comes from the yarn field name `JSON_KEY_SORT_ORDER` in the `net.minecraft.data.DataProvider` interface.

The interface `JsonKeySortOrderCallback.Adder` exists solely to explain what the parameters do. An `ObjIntConsumer<String>` would suffice, but it seems to be more common to use a custom functional interface for events in order to explain what it does, so I went with that instead. However, I'm not really sure about the name `Adder`, so if anyone has a better name, please suggest it. (Though this might be me bikeshedding again)

The test mod adds the `trigger` key with a priority value of `0`, since that is a key used in advancements that are also generated in the same test mod. This should put `trigger` above `conditions`, despite `conditions` coming before `trigger` lexicographically.


Technici4n (commented):
If there are two datagen entrypoints, will events registered in the second entrypoint be registered in time? Maybe it should be some function in `DataGeneratorEntrypoint` instead? (This can be implemented by capturing the map inside the `Util.make` lambda, storing it in a static field somewhere, and adding to it later).

ErrorCraft:
Wouldn't that require it to load `DataProvider` in order to store it to a field, which only happens when running the first data provider, which means it should have already started? (Forcing a class to load for the sake of capturing a field doesn't really jive with me either.) I see what you mean though, since the second entrypoint might only start when another entrypoint has already started a data provider, which will no longer modify the map as it's already been initialised.

Maybe storing another map in a static field works. Then inject everything at once as soon as the first data provider is called. Something like this:
- `DataGeneratorEntrypointA` stores its keys and priorities in a static field.
- `DataGeneratorEntrypointB` stores its keys and priorities in a static field.
- The data providers from `DataGeneratorEntrypointA` start, thus loading `DataProvider` and injecting all keys and priorities.
- The data providers from `DataGeneratorEntrypointB` start.

However, won't you face the same problem this way? When `DataGeneratorEntrypointA` stores the keys and priorities and starts, `DataProvider` may have been initialised already. This could be solved by looping over the entrypoints twice; Once for initialising the keys and priorities, and a second time actually running all the providers. However, I'm not sure if that's the preferred way of doing this...

Technici4n:
It could be done in a way that doesn't require DataProvider to be loaded early. For example if the stored map is null, we know that it hasn't been loaded yet and can add the entries when it does load.

ErrorCraft:
In the case where it isn't `null`, which means at least one generator is running already, wouldn't that lead to potential inconsistent results, such as one generator having a different sort priority for a key than another?

Technici4n:
Ah, I was thinking that fabric could add to it right before starting the datagen but after the providers have been collected. Just some ideas ^^

ErrorCraft:
Then I think the structure should be changed to accommodate for this accordingly, because currently each generator is started right after creating it, as seen in [this bit here](https://github.com/ErrorCraft/fabric/blob/key-priority-datagen/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java#L130-L132). Which, now that I think about it, won't this overwrite files that were just generated and written to if you have multiple entry points, since it uses the same output directory and removes stale entries?

What if all generators are initialised, collected and then ran together, rather than initialising and running them immediately?

Technici4n:
Oh, that's my bad, I thought they were all collected and run at once. I don't think we should change this.

What if we "reset" the map between runs? I think force-loading the class is not a problem btw.

I can imagine something like (pseudocode)
```java
DataGeneratorEntrypoint entrypoint = ...;
// Force load class to capture map in a static variable (say MAP in this pseudocode)
new DataProvider() { ... }
// Collect priorities from the entrypoint
Set<Key> addedKeys = new HashSet<>();
for key, prio in entrypoint.priorities() {
    MAP.put(key, prio);
    addedKeys.add(key);
}
// Run generator
dataGenerator.run();
// Clear added keys
MAP.keySet().removeAll(addedKeys);
```

Technici4n:
Up?

ErrorCraft:
Finally got back to this...
I like that idea, but what about the case where someone overrides an existing/default key, like `type`? This would be a poor choice, but someone might still do it. It would remove the entry after the generator finished, leading to different results for the next generator. We could copy the original map contents to another variable and replace them again after we removed all the keys, but that would create a lot of copies.
Also, would it be possible to just access `DataProvider.JSON_KEY_SORT_ORDER` directly and cast it back to `Object2IntOpenHashMap<String>`, rather than creating an instance to load the class in order to store the field using a mixin? It wouldn't be future proof if they ever decide to change the type, but at that point we would have to update it regardless if we used a mixin or not, so I doubt that's really a problem.

Technici4n:
> We could copy the original map contents to another variable and replace them again after we removed all the keys, but that would create a lot of copies.

That's only one copy per data generation entrypoint, which is acceptable.

> Also, would it be possible to just access `DataProvider.JSON_KEY_SORT_ORDER` directly and cast it back to `Object2IntOpenHashMap<String>`

Great idea, I didn't think of that. Much simpler indeed... :)

ErrorCraft:
> That's only one copy per data generation entrypoint, which is acceptable.

I realised I could place the copy outside of the loop, so there is only one copy after all.

Note that the sort keys currently do not get applied in the client test mod, because `DataGeneratorClientTestEntrypoint::onInitializeDataGenerator` is called immediately in `DataGeneratorTestEntrypoint`. I didn't do anything with this as it's meant to be replaced at some point anyways, as per the TODO in `DataGeneratorTestEntrypoint`.

Also, the nice thing about the way it works now is that each entry point has its own set of keys and values, so keys clashing should realistically never happen if people were to use multiple entry points. That's also why I could force `type` to be at the end in the client test mod, since only atlas sources are present and there aren't any other common keys.

Technici4n (approved):
Nice and clean, I like it.

liach:
Should we use a topological sort instead of int-based? int-based may be problematic if we try to insert a new key between two keys.

Technici4n:
I think int is fine here. This is per-entrypoint, so mods can just change their other keys if they want to.

ErrorCraft:
There's also the lexicographical sort that gets applied afterwards, so keys with the same priority already get sorted again.

ErrorCraft:
I also updated the title to better reflect the changes.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -60,4 +60,7 @@ default String getEffectiveModId() {
 	 */
 	default void buildRegistry(RegistryBuilder registryBuilder) {
 	}
+
+	default void addJsonKeySortOrders(JsonKeySortOrderAdder adder) {
```
Juuxel: Should have some documentation about what it does.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.datagen.v1;
+
+@FunctionalInterface
+public interface JsonKeySortOrderAdder {
```
Juuxel: This interface should likewise have a bit of javadoc, it can be a bit unclear if you only look at the method docs

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
```diff
@@ -123,13 +129,22 @@ private static void runInternal() {
 				final String effectiveModId = entrypoint.getEffectiveModId();
 				ModContainer modContainer = entrypointContainer.getProvider();
 
+				HashSet<String> keys = new HashSet<>();
+				entrypoint.addJsonKeySortOrders((key, value) -> {
+					jsonKeySortOrders.put(key, value);
```
Juuxel, line 135: Might want to check for null keys here to prevent accidental bugs

Juuxel (changes requested):
The change is nice, only needs some minor tweaks

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/impl/datagen/FabricDataGenHelper.java
ErrorCraft: What would be more preferred here? Throwing an exception or ignoring the value and logging a warning?

Technici4n: User error so exception. You can just add `Objects.requireNonNull` at the top of the lambda. (See other modules it's used a lot)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/DataGeneratorEntrypoint.java
```diff
@@ -60,4 +60,11 @@ default String getEffectiveModId() {
 	 */
 	default void buildRegistry(RegistryBuilder registryBuilder) {
 	}
+
+	/**
+	 * Adds json keys and priorities to sort them in the generated files.
+	 * @param adder A json key sort order adder callback.
```
haykam821:
```suggestion
	 * Provides a callback for setting the sort priority of object keys in generated JSON files.
	 * @param adder a callback for setting the sort priority for a given key
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Provides a way to add keys and priorities for sorting them in generated files.
+ */
+@FunctionalInterface
+public interface JsonKeySortOrderAdder {
+	/**
+	 * Adds a key with a priority for sorting.
+	 * @param key The key to compare.
+	 * @param priority The priority for this key. Lower numbers appear before ones with higher numbers. The default priority is 2.
```
haykam821:
```suggestion
	 * Sets the sort priority for a given object key within generated JSON files.
	 * @param key the key to set priority for
	 * @param priority the priority for the key, where keys with lower priority are sorted before keys with higher priority
	 * @implNote The default priority is 2.
```

haykam821: The default priority here doesn't apply to `type` (`0`) or `parent` (`1`). Should these exceptions that are default in vanilla be documented as well?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.datagen.v1;
+
+/**
+ * Provides a way to add keys and priorities for sorting them in generated files.
```
haykam821:
```suggestion
 * Provides a callback for setting the sort priority of object keys in generated JSON files.
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
ErrorCraft: That would be useful for an additional comparison. Should the two keys provided by fabric (`fabric:load_conditions` and `fabric:type`) also be included then?

Technici4n: I'd keep them as implementation details.

ErrorCraft: So do I not add them at all to the `@implNote` or just the vanilla ones?

Technici4n: Vanilla ones are fine but could also get outdated. Maybe we just add `@see <field reference>`?

ErrorCraft: Alright, I added a reference to `net.minecraft.data.DataProvider#JSON_KEY_SORT_ORDER`.

ErrorCraft:
Just saw that the target branch was still 1.20, rather than 1.20.1.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.datagen.v1;
+
+/**
+ * Provides a callback for setting the sort priority of object keys in generated JSON files.
+ */
+@FunctionalInterface
+public interface JsonKeySortOrderAdder {
```
modmuss50: `Consumer` might be a bit of a better name, or `Callback`? We use `Builder` above, but I dont think thats correct in this case. `Adder` isnt really something we have used before anywhere.

modmuss50 (approved):
Looks good, maybe change the interface name? Other than that looks good.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/JsonKeySortOrderAdder.java
ErrorCraft: `Adder` is used in a couple of yarn class names for UI and was really the first thing I thought of, presumably because I was working with UI around that time. Anyhow, I'll change the name to `Callback`, as it's used in the javadoc anyways, so that's an additional consistency too.

haykam821 (approved):

PR #3099, opened 2023-06-05
Interface inject FabricBakedModel

PepperCode1:
### API Changes
- Interface inject `FabricBakedModel`
This change was done to eliminate the constant need to cast `BakedModel` to `FabricBakedModel`, especially when rendering sub-models. The implementations for the methods of `FabricBakedModel` were moved from `BakedModelMixin` to `FabricBakedModel` as default implementations. Some javadoc was updated to reflect this change.
- Deprecate the mesh consumer (`Consumer<Mesh>` retrieved from `RenderContext`)
This change was done to ensure consistency across all current and future contexts in which a mesh may need to be output to a quad emitter. The preferred direct replacement is the new method `Mesh#outputTo(QuadEmitter)`. Some javadoc was updated to reflect this change.
- Deprecate the baked model consumer (`BakedModelConsumer` retrieved from `RenderContext`)
This change was done to ensure consistent rendering of sub-models and to eliminate assumptions about what features a model uses. The preferred direct replacement is to use the appropriate `emit` method in `FabricBakedModel`. Some javadoc was updated to reflect this change.
Even though the consumer is now deprecated, the default `FabricBakedModel` method implementations still use it and renderers must still implement the getter. This is because Indigo and Indium sometimes apply smooth lighting differently based on if the quad came from a vanilla model or not. The eventual solution to allow all baked model consumer usage to be removed is to allow renderers to override the default `BakedModelMixin` and handle default vanilla models directly.
- Fix `QuadView#toVanilla`'s javadoc reporting the wrong minimum array size

### Implementation Changes
- Restructure `RenderContext` implementations
This change was done to reduce code duplication, improve standardization, and improve readability. Most code of `AbstractQuadRenderer` was moved into a new class called `AbstractBlockRenderContext` (which `BlockRenderContext` and `TerrainRenderContext` now extend), with the main buffering method being moved to `AbstractRenderContext`.
- Remove red blue color swap
It is unclear why this code was necessary in the first place. Indigo stores vertex color in ARGB format, then converts it to ABGR format only if the native byte order is little endian, and then reads the vertex color in ABGR format when buffering vertex data. This would mean that on big endian systems, all red and blue color components would be swapped.
Now, color is read in ARGB format when buffering and no color format conversion or swapping is done.
- Encode face normal
This change was done to improve the performance of meshes. Meshes already encoded all other geometry data, but the face normal would still have to be recomputed every time a mesh quad was rendered. Now, it is stored in the quad data header, so it is simply decoded, which is significantly faster.
- Fix some bugs
  - Outputting a mesh via the mesh consumer would result in the emitter retaining the data of the last quad in the mesh.
  - In non-terrain block rendering, the baked model consumer would incorrectly use the random seed provided by the block state instead of the passed random seed argument.
  - When converting to or from vanilla quad data, the color would not be converted. Indigo uses ARGB format to store color, while vanilla quads use ABGR or RGBA. See the comment near the bottom of `ColorHelper` for more information.
  - Fix TerrainRenderContext using incorrect overlay value of `0` instead of `OverlayTexture.DEFAULT_UV`.
- Improve performance of some code
  - `QuadViewImpl#computeGeometry()` was being called before transforms and the cull test were applied. These operations can change the geometry and invalidate computed geometry data or cancel the quad outright. Calling the method explicitly in those contexts was also not necessary in the first place.
  - A single `Vector4f` instance is now reused to transform the vertex position instead of allocating a new `Vector4f` for each vertex.
  - The random seed lazy computation now uses a separate boolean instead of checking if the seed is `-1`.


Technici4n (changes requested):
You're probably aware but injected interfaces must have default method bodies (otherwise it leads to weird compile errors) and a class javadoc saying that the interface is injected (by convention, see e.g. FabricItem).

PepperCode1:
Reviewers should consider the following additional refactors related to `RenderContext` implementation code.
- Should `ChunkRenderInfo` be merged into `TerrainRenderContext`? This class is not used or accessed outside of `TerrainRenderContext`, and the context always uses the same `final` instance of the `ChunkRenderInfo`. Merging them could improve readability since there are less classes overall and it would very slightly improve performance since some operations would no longer be deferred to a separate object.
- Should the fields that store the parameters that `ItemRenderContext` receives be separated into a separate `ItemRenderInfo` class, much like `BlockRenderInfo`? This would decrease the size of the `ItemRenderContext` class. If these two render info classes were to share a super-class, it would also allow the `colorizeQuad` method from `AbstractBlockRenderContext` and `ItemRenderContext` to be de-duplicated and moved to `AbstractRenderContext`.

Additionally, I may add the face normal encoding change to this pull request within a few days. This change is simple and unrelated to any existing changes, however, so reviews are still useful at this point.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/helper/ColorHelper.java
```diff
@@ -28,13 +26,13 @@
 public abstract class ColorHelper {
 	private ColorHelper() { }
 
-	private static final Int2IntFunction colorSwapper = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN ? color -> ((color & 0xFF00FF00) | ((color & 0x00FF0000) >> 16) | ((color & 0xFF) << 16)) : color -> color;
```
Technici4n, line 31: These constant lambdas are very fast, likely faster than an `if` check every time, and they should get inlined by hotspot. I suspect it was done this way for performance reasons in the first place.

PepperCode1: FREX base switched to a static final boolean as well. There is a comment saying that if checks with static final booleans are inlined, which if true, would be faster than using a function. I had not ever heard that functions like this can be inlined. I will ask a few community members that know more about how the JIT works.

LlamaLad7: Both get inlined, the resulting assembly is identical: https://gist.github.com/LlamaLad7/0c4c39b1e86f4ebb8bd01bd4c6f9f2d2.
The JIT is very good at its job and there is normally no need to make such micro-optimizations manually.

Technici4n: Ah, that's great. Thank you Llama. In that case we should probably keep the boolean check in `ColorHelper` to keep the quad emitting code more readable.

PepperCode1: The red blue color swap has been removed completely in the latest commit.

fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/mesh/QuadView.java
```diff
@@ -201,7 +201,7 @@ default Vector2f copyUv(int vertexIndex, @Nullable Vector2f target) {
 	 * @param target Target array for the baked quad data.
 	 *
 	 * @param targetIndex Starting position in target array - array must have
-	 * at least 28 elements available at this index.
+	 * at least {@link #VANILLA_QUAD_STRIDE} elements available at this index.
```
Technici4n, line 204: You can even use `@value` here

PepperCode1: IntelliJ says `@value tag must reference a field with a constant initializer`, so `@link` is a better option.

Technici4n: Right...

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/MutableQuadViewImpl.java
```diff
@@ -174,8 +174,16 @@ public MutableQuadViewImpl copyFrom(QuadView quad) {
 
 	@Override
 	public final MutableQuadViewImpl fromVanilla(int[] quadData, int startIndex) {
-		System.arraycopy(quadData, startIndex, data, baseIndex + HEADER_STRIDE, QUAD_STRIDE);
+		System.arraycopy(quadData, startIndex, data, baseIndex + HEADER_STRIDE, VANILLA_QUAD_STRIDE);
```
Technici4n, line 183: Doesn't this just skip the end of the array?

PepperCode1: `QUAD_STRIDE` and `VANILLA_QUAD_STRIDE` are exactly the same value (this is verified in `EncodingFormat`), but I thought switching it might make this slightly easier to understand.

Technici4n: Should also be changed in `toVanilla` then

PepperCode1: I don't think so. Whether the stride is vanilla should match whether the source array is vanilla, and with this change, it does.

Technici4n: Idk, it was confusing for me because I didn't know that the values were equal.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/mesh/MutableQuadViewImpl.java
```diff
@@ -200,4 +194,17 @@ public final MutableQuadViewImpl fromVanilla(BakedQuad quad, RenderMaterial mate
...
 	}
+
+	/**
+	 * Emit the quad without clearing the underlying data.
+	 * Geometry is not guaranteed to be valid when called, but can be computed by calling {@link #computeGeometry()}.
+	 */
+	public abstract void emitDirectly();
+
+	@Override
+	public MutableQuadViewImpl emit() {
```
Technici4n: Should probably be final.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/TerrainFallbackConsumer.java
```diff
@@ -1,114 +0,0 @@
...
-/**
- * Consumer for vanilla baked models. Generally intended to give visual results matching a vanilla render,
- * however there could be subtle (and desirable) lighting variations so is good to be able to render
- * everything consistently.
- *
- * <p>Also, the API allows multi-part models that hold multiple vanilla models to render them without
- * combining quad lists, but the vanilla logic only handles one model per block. To route all of
- * them through vanilla logic would require additional hooks.
- *
- *  <p>Works by copying the quad data to an "editor" quad held in the instance,
```
Technici4n, line 53: Would be nice to keep parts of this comment, moved to the relevant locations.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
```diff
@@ -317,40 +288,33 @@ public void accept(BakedModel model) {
...
-						}
+					for (int j = 0; j < count; j++) {
+						final BakedQuad q = quads.get(j);
+						editorQuad.fromVanilla(q, IndigoRenderer.MATERIAL_STANDARD, cullFace);
+						// Call renderQuad directly instead of emit for efficiency
+						renderQuad(editorQuad);
 					}
 				}
+
+				editorQuad.clear();
```
Technici4n, line 296: We don't need to clear here, right?

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
```diff
@@ -317,40 +288,33 @@ public void accept(BakedModel model) {
 		@Override
 		public void accept(BakedModel model, @Nullable BlockState state) {
 			if (hasTransform()) {
```
Technici4n, line 278: Note: once we move away from the fallback consumer in the FBM default implementation, won't we be routing everything through FRAPI? Isn't that inefficiencient?

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
```diff
@@ -317,40 +288,33 @@ public void accept(BakedModel model) {
...
-					if (count != 0) {
-						for (int j = 0; j < count; j++) {
-							final BakedQuad q = quads.get(j);
-							editorQuad.fromVanilla(q, IndigoRenderer.MATERIAL_STANDARD, cullFace);
-							renderMeshQuad(editorQuad);
-						}
+					for (int j = 0; j < count; j++) {
+						final BakedQuad q = quads.get(j);
+						editorQuad.fromVanilla(q, IndigoRenderer.MATERIAL_STANDARD, cullFace);
+						// Call renderQuad directly instead of emit for efficiency
```
Technici4n, line 291: Won't this be slower again when we will route the model through `QuadEmitter#emit`?

Technici4n (commented):
Looks good, nice cleanup! Few remarks

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
PepperCode1: We do, because the `ItemRenderContext` has only one editor quad; it does not need to distinguish between vanilla quads like the `AbstractBlockRenderContext` does. The editor quad that the baked model consumer here uses is the same one that is exposed through `getEmitter`, so it needs to be cleared.

Technici4n:
>     * Should `ChunkRenderInfo` be merged into `TerrainRenderContext`? This class is not used or accessed outside of `TerrainRenderContext`, and the context always uses the same `final` instance of the `ChunkRenderInfo`. Merging them could improve readability since there are less classes overall and it would very slightly improve performance since some operations would no longer be deferred to a separate object.
> 
>     * Should the fields that store the parameters that `ItemRenderContext` receives be separated into a separate `ItemRenderInfo` class, much like `BlockRenderInfo`? This would decrease the size of the `ItemRenderContext` class. If these two render info classes were to share a super-class, it would also allow the `colorizeQuad` method from `AbstractBlockRenderContext` and `ItemRenderContext` to be de-duplicated and moved to `AbstractRenderContext`.

For 1), I think the current `ChunkRenderInfo` separation is nice. For 2) I don't particularly care either way about putting the state in `ItemRenderInfo` vs keeping it in the context, so that's up to you. I will say that abstracting `colorizeQuad` would barely reduce the code duplication (it's a simple method after all), but would make understanding the implementation harder, so I wouldn't do that.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
PepperCode1: If it goes through the context's emitter, then yes. However, if the renderer is able to override the default `emitItemQuads` method (as planned), then it can cast the context or emitter to some internal type and then use intrinsics to render the quads and model, making it as efficient as it is now.

I am not sure that the vanilla path is really that fast though; it allocates quite a few different objects while Indigo is able to avoid all allocations. The emissive check and per-quad vertex consumers could provide some slight slowdown, but again this can be avoided using the solution described earlier.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
Technici4n: Ah yes I see. I was confused by the two different variables both named `editorQuad`.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
Technici4n: Are all renderers expected to override emitBlockQuads / emitItemQuads to provide a faster path? or just Indigo?

The idea was that only Indigo would override `emitBlockQuads`, and just to capture whether a quad is vanilla or not.

PepperCode1:
> > ```
> > * Should `ChunkRenderInfo` be merged into `TerrainRenderContext`? This class is not used or accessed outside of `TerrainRenderContext`, and the context always uses the same `final` instance of the `ChunkRenderInfo`. Merging them could improve readability since there are less classes overall and it would very slightly improve performance since some operations would no longer be deferred to a separate object.
> > 
> > * Should the fields that store the parameters that `ItemRenderContext` receives be separated into a separate `ItemRenderInfo` class, much like `BlockRenderInfo`? This would decrease the size of the `ItemRenderContext` class. If these two render info classes were to share a super-class, it would also allow the `colorizeQuad` method from `AbstractBlockRenderContext` and `ItemRenderContext` to be de-duplicated and moved to `AbstractRenderContext`.
> > ```
> 
> For 1), I think the current `ChunkRenderInfo` separation is nice. For 2) I don't particularly care either way about putting the state in `ItemRenderInfo` vs keeping it in the context, so that's up to you. I will say that abstracting `colorizeQuad` would barely reduce the code duplication (it's a simple method after all), but would make understanding the implementation harder, so I wouldn't do that.

At the time that I asked the question, `colorizeQuad` was more complicated because of the color swap, but that has been removed now. I do not think that it is worth abstracting it anymore either.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/render/ItemRenderContext.java
PepperCode1: My impression was that `FabricBakedModel`'s default implementations would route the model's quads only through the quad emitter as exposed through the API, but renderers can override that behavior for either `emit` method to provide extra functionality (like vanilla lighting in Indigo) or efficiency (like with some vanilla baked models).

> Are all renderers expected to override emitBlockQuads / emitItemQuads to provide a faster path? or just Indigo?

If they want to provide a faster path, they can do that, but it is possible that it is unnecessary or not worth it. FREX base and the FREX FRAPI implementation does not choose do this, for example.

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/helper/GeometryHelper.java
```diff
@@ -82,11 +81,7 @@ public static int computeShapeFlags(QuadView quad) {
 	 * Does not validate quad winding order.
 	 * Expects convex quads with all points co-planar.
 	 */
-	public static boolean isQuadParallelToFace(@Nullable Direction face, QuadView quad) {
-		if (face == null) {
-			return false;
-		}
-
+	public static boolean isQuadParallelToFace(@NotNull Direction face, QuadView quad) {
```
Technici4n: No need for NotNull

Technici4n:
All commits look good so far.

Technici4n (approved):
Please remember to remove that `@NotNull` (see comment above).

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/renderer/helper/ColorHelper.java
```diff
@@ -65,9 +65,9 @@ public static int maxBrightness(int b0, int b1) {
 	}
 
 	/*
-	Renderer color format: ARGB
-	Vanilla color format (little endian): ABGR
-	Vanilla color format (big endian): RGBA
+	Renderer color format: ARGB (0xAARRGGBB)
+	Vanilla color format (little endian): ABGR (0xAABBGGRR)
+	Vanilla color format (big endian): RGBA (0xRRGGBBAA)
```
shartte, line 70: I find the endian-ness very confusing since that only applies to byte-representation, not the integers. 
Also: An integer 0xAARRGGBB is laid out as (0xBB 0xGG 0xRR 0xAA) in little endian memory, isn't it?

shartte: p.s.: Is there *ANY* big-endian platform that LWJGL even runs on?

Technici4n (approved):
Went through the full diff once again, still looks good. :smile:

modmuss50 (approved):
Awesome job, Many Thanks 👍

PR #3115, opened 2023-06-11
Add builders for BlockSetType and WoodType to replace registries

Shnupbups:
Adds `BlockSetTypeBuilder` and `WoodTypeBuilder` to replace `BlockSetTypeRegistry` and `WoodTypeRegistry` respectively.

The builder format is more suited to these as they have an large amount of values which aren't often changed. `BlockSetTypeRegistry`'s register method in particular has a whopping 11 parameters, most of which are sound events that you'd have to try not to mess up the order on.

Previously, if you wanted a `BlockSetType` that was identical to `BlockSetType.OAK` except with custom button sounds and doors that require redstone to open, you'd have to write:
```java
public static final BlockSetType MAPLE = BlockSetTypeRegistry.register(
	 new Identifier("mymod","maple"),
	 false,
	 BlockSoundGroup.WOOD,
	 SoundEvents.BLOCK_WOODEN_DOOR_CLOSE,
	 SoundEvents.BLOCK_WOODEN_DOOR_OPEN,
	 SoundEvents.BLOCK_WOODEN_TRAPDOOR_CLOSE,
	 SoundEvents.BLOCK_WOODEN_TRAPDOOR_OPEN,
	 SoundEvents.BLOCK_WOODEN_PRESSURE_PLATE_CLICK_OFF,
	 SoundEvents.BLOCK_WOODEN_PRESSURE_PLATE_CLICK_ON,
	 MyModSoundEvents.BLOCK_MAPLE_BUTTON_CLICK_OFF,
	 MyModSoundEvents.BLOCK_MAPLE_BUTTON_CLICK_ON
);
```
Now you can write simply:
```java
public static final BlockSetType MAPLE = BlockSetTypeBuilder.copyOf(BlockSetType.OAK)
	 .openableByHand(false)
	 .buttonClickOffSound(MyModSoundEvents.BLOCK_MAPLE_BUTTON_CLICK_OFF)
	 .buttonClickOnSound(MyModSoundEvents.BLOCK_MAPLE_BUTTON_CLICK_ON)
	 .register(new Identifier("mymod","maple"));
```
Much clearer what each value does (particularly the `openableByHand` boolean value!), no need to worry about getting the order wrong for the many sound events, and the `copyOf` method allows to skip things that are already set just how you want them.

An added bonus is, that if Mojang decides to add new values to block set types or wood types (as they did in 1.20 by adding the openableByHand boolean!), where previously the already unwieldy `register` methods would need those new values added, changing the method signatures and making them even longer, now with the builder format we can simply add a new method.

All the values default to those used by the shortform constructors of `BlockSetType` and `WoodType`.

The old registry classes are deprecated in favour of these builders.

This can be backported to 1.19.4 (would need references to the `openableByHand` boolean removed).


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/type/BlockSetTypeBuilder.java
```diff
@@ -0,0 +1,238 @@
...
+
+	/**
+	 * Builds and registers a {@link BlockSetType} from this builder's values.
+	 *
+	 * @param id the id for the built {@link BlockSetType}
+	 *
+	 * @return the built and registered {@link BlockSetType}
+	 */
+	public BlockSetType build(Identifier id) {
+		return BlockSetType.register(new BlockSetType(id.toString(), openableByHand, soundGroup, doorCloseSound, doorOpenSound, trapdoorCloseSound, trapdoorOpenSound, pressurePlateClickOffSound, pressurePlateClickOnSound, buttonClickOffSound, buttonClickOnSound));
```
Technici4n: Would be nice to have a way to build it without registering it, maybe?

Shnupbups: What purpose would that serve? I only forsee that causing problems where people forget to register it and it causes issues. If you can think of a good usecase, I guess I can add a `buildWithoutRegistering` method or something?

Technici4n: If anything, a method named `build` should not `register`

Shnupbups: I just thought the name `buildAndRegister` was too verbose. There's no sense having an unregistered type, and being able to build without registering I feel would lead to too many people forgetting to register.
Keep in mind it's not an actual Registry, it's just adding to the internal `private static final Set<BlockSetType> VALUES` (or the WoodType equivalent).
I suppose I could remove the call to `register` and just return the built type, and rely on modders to call `BlockSetType.register` themselves... would that be preferred?

Juuxel: Or the method could be called `register`? IMO it's clear from the return type that it's built.

Shnupbups: Yeah I suppose that works too. I guess I just went "it's a builder -> it ends with `build`" but `register` would be a fine name. Is that good with you @Technici4n ?

Technici4n: could be `buildAndRegister` if it does both. :P

Technici4n: or just `register` sure

Shnupbups: Decided to make `build` and `register` two different methods - `register` calling `build` and registering its returned type, and `build` simply making the type without registering. Added javadoc to (hopefully!) make this clear.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/type/BlockSetTypeBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+
+	/**
+	 * Sets this block set type's door and trapdoor to be unopenable by hand.
+	 *
+	 * @return this builder for chaining
+	 */
+	public BlockSetTypeBuilder unopenableByHand() {
+		this.openableByHand = false;
+		return this;
+	}
```
Technici4n: Not too sure about these two, especially the last, but I guess they don't hurt.

Technici4n (approved):
Looks good, though someone else should check in an IDE that the names make sense.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/type/BlockSetTypeBuilder.java
modmuss50: I agree, I think generally id just go with the `openableByHand(boolean openableByHand)` 👍

modmuss50 (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/type/BlockSetTypeBuilder.java
Shnupbups: Removed these shortcut methods in latest commit

PR #3117, opened 2023-06-12
Fix resource conditions for loot tables

### LINKED ISSUE #3116: Resource conditions API seem to be broken with loot tables in 1.20
```json
{
	"type": "minecraft:block",
	"pools": [
		{
			"rolls": 1,
			"entries": [
				{
					"type": "minecraft:item",
					"name": "promenade:carbonite_column"
				}
			],
			"conditions": [
				{
					"condition": "minecraft:survives_explosion"
				}
			]
		}
	],
	"fabric:load_conditions": [
		{
			"values": [
				"columns"
			]
		}
	]
}
```

Error: `Expected name to be an item, was unknown string 'promenade:carbonite_column'`
It tries to load that table even if the `columns` mod is absent

apple502j:
Resolves #3116

And predicates (loot conditions), item modifiers (loot functions). They are handled by the same method.

`LootManager` is no longer a `JsonDataLoader` - in fact it is not even a `SinglePreparationResourceReloader` anymore. Therefore it needs a new Mixin.

The new mixin, instead of manipulating a map like the old Mixin, early-returns a `forEach` call that parses JSONs.

Added two tests using predicates.

### DIFF
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/mixin/resource/conditions/LootManagerMixin.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/mixin/resource/conditions/LootManagerMixin.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


apple502j:
While tag conditions do not currently work, I still think we should merge it now and fix later. Kinda complex, I think.

Technici4n:
I agree, will have a look in a bit.

Technici4n (approved):

modmuss50 (approved):

PR #3118, opened 2023-06-12
Add support for partially synced client tags

dexman545:
Allows for client tags to reference tags that may be synced from a server, allowing for them to be used.

An example of this being useful is in AutoSwitch. To address `mc:sword_efficient` not containing bamboo, a wrapper tag is needed `as:sword_efficient`. With partially synced tags it would allow for server-side changes to `mc:sword_efficient` to be seen in `as:sword_efficient`, which better reflects the intent of the tag.

This is a draft as implementation is worked out, some todos are left in with some questions.

### DIFF
--- a/fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
+++ b/fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
@@ -62,6 +62,24 @@ public final class ClientTags {
 	}
 
 	/**
+	 * Checks if an entry is in a tag.
+	 *
+	 * <p>If the synced tag does exist, it is queried. If it does not exist,
+	 * the tag populated from the available mods is checked, recursively checking the
+	 * synced tags and entries contained within.
+	 *
+	 * @param tagKey the {@code TagKey} to being checked
+	 * @param entry  the entry to check
+	 * @return if the entry is in the given tag
+	 */
+	public static <T> boolean isInWithLocalFallback(TagKey<T> tagKey, T entry) {
+		Objects.requireNonNull(tagKey);
+		Objects.requireNonNull(entry);
+
+		return getRegistryEntry(tagKey, entry).map(re -> isInWithLocalFallback(tagKey, re)).orElse(false);
+	}
+
+	/**
 	 * Checks if an entry is in a tag, for use with entries from a dynamic registry,
 	 * such as {@link net.minecraft.world.biome.Biome}s.
 	 *
@@ -73,6 +91,7 @@ public final class ClientTags {
 	 * @param registryEntry the entry to check
 	 * @return if the entry is in the given tag
 	 */
+	@SuppressWarnings("unchecked")
 	public static <T> boolean isInWithLocalFallback(TagKey<T> tagKey, RegistryEntry<T> registryEntry) {
 		Objects.requireNonNull(tagKey);
 		Objects.requireNonNull(registryEntry);
@@ -102,24 +121,6 @@ public final class ClientTags {
 		}
 
 		return isIn;
-	}
-
-	/**
-	 * Checks if an entry is in a tag.
-	 *
-	 * <p>If the synced tag does exist, it is queried. If it does not exist,
-	 * the tag populated from the available mods is checked, recursively checking the
-	 * synced tags and entries contained within.
-	 *
-	 * @param tagKey the {@code TagKey} to being checked
-	 * @param entry  the entry to check
-	 * @return if the entry is in the given tag
-	 */
-	public static <T> boolean isInWithLocalFallback(TagKey<T> tagKey, T entry) {
-		Objects.requireNonNull(tagKey);
-		Objects.requireNonNull(entry);
-
-		return getRegistryEntry(tagKey, entry).map(re -> isInWithLocalFallback(tagKey, re)).orElse(false);
 	}
 
 	/**
--- a/fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
+++ b/fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.api.tag.client.v1;
 
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -23,12 +24,12 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import net.minecraft.client.MinecraftClient;
-import net.minecraft.registry.tag.TagKey;
-import net.minecraft.util.Identifier;
 import net.minecraft.registry.Registries;
 import net.minecraft.registry.Registry;
-import net.minecraft.registry.entry.RegistryEntry;
 import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.entry.RegistryEntry;
+import net.minecraft.registry.tag.TagKey;
+import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.impl.tag.client.ClientTagsLoader;
 
@@ -46,6 +47,7 @@ import net.fabricmc.fabric.impl.tag.client.ClientTagsLoader;
  */
 public final class ClientTags {
 	private static final Map<TagKey<?>, Set<Identifier>> LOCAL_TAG_CACHE = new ConcurrentHashMap<>();
+	private static final Map<TagKey<?>, ClientTagsLoader.ChildHolder> LOCAL_TAG_HIERARCHY = new ConcurrentHashMap<>(); //todo use LoadedTag, merge caches?
 
 	private ClientTags() {
 	}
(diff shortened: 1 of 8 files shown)


fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -60,13 +62,84 @@ public static Set<Identifier> getOrCreateLocalTag(TagKey<?> tagKey) {
...
+			}
+		}
+
+		if (registryEntry.getKey().isPresent()) {
+			ClientTagsLoader.ChildHolder wt = getOrCreatePartiallySyncedTag(tagKey);
+			boolean isIn = wt.immediateChildIds().contains(registryEntry.getKey().get().getValue());
+			Iterator<TagKey<?>> it = wt.immediateChildTags().iterator();
+
+			while (!isIn && it.hasNext()) {
+				isIn = isInPartiallySyncedTag((TagKey<T>) it.next(), registryEntry);
```
Technici4n: Do we want to recursively check the full hierarchy every time, or cache the result per tag? (And invalidate when updated tags are received)

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -60,13 +62,84 @@ public static Set<Identifier> getOrCreateLocalTag(TagKey<?> tagKey) {
...
+		// Check if the tag exists in the dynamic registry first
+		Optional<? extends Registry<T>> maybeRegistry = getRegistry(tagKey);
+
+		if (maybeRegistry.isPresent()) {
+			if (maybeRegistry.get().getEntryList(tagKey).isPresent()) {
+				return registryEntry.isIn(tagKey);
+			}
+		}
+
+		if (registryEntry.getKey().isPresent()) {
```
Technici4n: The check could be inverted to make control flow a bit more linear:

```java
if (registryEntry.getKey().isEmpty()) {
    // No key?
    return false;
}
```

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -60,13 +62,84 @@ public static Set<Identifier> getOrCreateLocalTag(TagKey<?> tagKey) {
...
+				isIn = isInPartiallySyncedTag((TagKey<T>) it.next(), registryEntry);
+			}
+
+			return isIn;
+		}
+
+		return false;
+	}
+
+	public static <T> boolean isInPartiallySyncedTag(TagKey<T> tagKey, T entry) {
```
Technici4n: Can't we put the logic to retrieve `RegistryEntry<T>` from `TagKey` and `T` in a shared private helper to avoid repeating it?

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -46,6 +47,7 @@
  */
 public final class ClientTags {
 	private static final Map<TagKey<?>, Set<Identifier>> LOCAL_TAG_CACHE = new ConcurrentHashMap<>();
+	private static final Map<TagKey<?>, ClientTagsLoader.ChildHolder> LOCAL_TAG_HIERARCHY = new ConcurrentHashMap<>(); //todo use LoadedTag, merge caches?
```
Technici4n: `LOCAL_TAG_CACHE` is the `completeIds`, right?

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
Technici4n: I think recursively checking the hierarchy as you're doing here is a good approach for now, and we can reconsider it if there's performance issue reports. Good idea :)

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/impl/tag/client/ClientTagsLoader.java
```diff
@@ -74,25 +74,37 @@ public static Set<Identifier> loadTag(TagKey<?> tagKey) {
...
 				}
 			}, ids::add);
 		}
 
-		return Collections.unmodifiableSet(ids);
+		return new LoadedTag(Collections.unmodifiableSet(ids), new ChildHolder(Collections.unmodifiableSet(immediateChildTags),
+				Collections.unmodifiableSet(immediateChildIds)));
+	}
+
+	//todo merge these?
```
Technici4n: yes probably ^^

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
dexman545: Yes, `LOCAL_TAG_CACHE` is the same as `completeIds`

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -126,16 +100,27 @@ public static <T> boolean isInWithLocalFallback(TagKey<T> tagKey, RegistryEntry<
...
-			return isInLocal(tagKey, registryEntry.getKey().get());
+		if (registryEntry.getKey().isEmpty()) {
+			// No key?
+			return false;
 		}
 
-		return false;
+		// Recursively search the entries contained with the tag
+		ClientTagsLoader.LoadedTag wt = getOrCreatePartiallySyncedTag(tagKey);
+		boolean isIn = wt.immediateChildIds().contains(registryEntry.getKey().get().getValue());
```
Technici4n: Nitpick: this variable isn't useful, it could be replaced by returning true and false directly in the right places

Technici4n (approved):

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -175,4 +163,30 @@ private static <T> Optional<? extends Registry<T>> getRegistry(TagKey<T> tagKey)
...
+	private static ClientTagsLoader.LoadedTag getOrCreatePartiallySyncedTag(TagKey<?> tagKey) {
+		ClientTagsLoader.LoadedTag loadedTag = LOCAL_TAG_HIERARCHY.get(tagKey);
+
+		if (loadedTag == null) {
+			loadedTag = ClientTagsLoader.loadTag(tagKey);
+			LOCAL_TAG_HIERARCHY.put(tagKey, loadedTag);
+		}
+
+		return loadedTag;
+	}
```
modmuss50: Could maybe move these methods and `LOCAL_TAG_HIERARCHY` to the impl class? Generally I think its best to keep the public API classes free of a lot of logic, it does make it a little easier to se what is/isnt api. Im not too bothered either way though as this class already contained quite a bit of logic anyway 👍

fabric-client-tags-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/tag/client/v1/ClientTagTest.java
```diff
@@ -20,20 +20,33 @@
...
 public class ClientTagTest implements ClientModInitializer {
 	private static final Logger LOGGER = LoggerFactory.getLogger(ClientTagTest.class);
+	private static final String MODID = "fabric-clients-tags-api-v1-testmod";
 
 	@Override
 	public void onInitializeClient() {
+		FabricLoader.getInstance().getModContainer(MODID)
+				.map(container -> ResourceManagerHelper.registerBuiltinResourcePack(new Identifier(MODID, "test2"),
+						container, ResourcePackActivationType.ALWAYS_ENABLED))
+				.filter(success -> !success).ifPresent(success -> LOGGER.warn("Could not register built-in resource pack."));
```
modmuss50:
```suggestion
		final ModContainer container = FabricLoader.getInstance().getModContainer(MODID).get();

		if (!ResourceManagerHelper.registerBuiltinResourcePack(new Identifier(MODID, "test2"),
				container, ResourcePackActivationType.ALWAYS_ENABLED)) {
			throw new IllegalStateException("Could not register built-in resource pack.");
		}
```

Id perfer to fail hard if something hasnt gone quite right, especially in a test.

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
```diff
@@ -126,13 +99,28 @@ public static <T> boolean isInWithLocalFallback(TagKey<T> tagKey, RegistryEntry<
...
 		}
 
-		if (registryEntry.getKey().isPresent()) {
-			return isInLocal(tagKey, registryEntry.getKey().get());
+		if (registryEntry.getKey().isEmpty()) {
+			// No key?
+			return false;
+		}
+
+		// Recursively search the entries contained with the tag
```
modmuss50: Question: Do we need to worry about getting into an infine loop here?

modmuss50 (commented):
Looks great 👍 I just have a few questions/nit picks.

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
dexman545: I'll leave it for now.

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/api/tag/client/v1/ClientTags.java
dexman545: I don't think so as tags can't be self-referential, but we could change it.

Technici4n:
Your change to the recursion handling means that nested tags won't be correctly overridden anymore, as far as I can tell.

dexman545:
> Your change to the recursion handling means that nested tags won't be correctly overridden anymore, as far as I can tell.

Good catch. I've restored the recursive search, but added tracking of already tested tags so we should now always terminate.

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/impl/tag/client/ClientTagsImpl.java
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		checked.add(tagKey);
+
+		for (TagKey<?> key : wt.immediateChildTags()) {
+			if (isInWithLocalFallback((TagKey<T>) key, registryEntry, checked)) {
+				return true;
+			}
+
+			checked.add((TagKey<T>) key);
```
Technici4n, line 75: I would move `checked.add(tagKey)` to the top of the method right after the `if` check and remove this one.

fabric-client-tags-api-v1/src/client/java/net/fabricmc/fabric/impl/tag/client/ClientTagsLoader.java
```diff
@@ -73,26 +72,42 @@ public static Set<Identifier> loadTag(TagKey<?> tagKey) {
...
-		return Collections.unmodifiableSet(ids);
+		// Ensure that the tag does not refer to itself
+		completeTags.remove(tagKey);
+		immediateChildTags.remove(tagKey);
+
+		return new LoadedTag(Collections.unmodifiableSet(completeIds), Collections.unmodifiableSet(completeTags),
+				Collections.unmodifiableSet(immediateChildTags), Collections.unmodifiableSet(immediateChildIds));
+	}
+
+	public record LoadedTag(Set<Identifier> completeIds, Set<TagKey<?>> completeChildTags, Set<TagKey<?>> immediateChildTags, Set<Identifier> immediateChildIds) {
```
Technici4n: The `completeChildTags` are now unnecessary again. I think you can remove `immediate` in front of the other two fields. :wink:

Technici4n (approved):

modmuss50 (approved):

PR #3140, opened 2023-06-24
Render Data API v2

PepperCode1:
This pull requests adds the Render Data v2 module, which directly supersedes the existing Rendering Data Attachment v1 module.

The original goal was to interface inject the two interfaces provided by Rendering Data Attachment v1, `RenderAttachmentBlockEntity` and `RenderAttachedBlockView`, to allow for an easier to use API. It proved to be impossible to achieve this and still provide a clean API without making a new module, as explained below.

The documentation of `RenderAttachedBlockView` states that it is guaranteed to be implemented on all instances of `BlockRenderView`, but this is not the case. The module implementation only guarantees that all `WorldView` instances implement the interface. `RenderAttachedBlockView` extends `BlockRenderView`, meaning that trying to change the mixin to satisfy the stated contract is not possible; thus, it can be deduced that the mixin is correct but the stated contract is not, though such a contract does not lead to an easy to use API. Such an API design also makes interface injection of `RenderAttachedBlockView` impossible.

To allow for the currently stated contract to be satisfied and to allow for interface injection to happen, a new module was made. It uses the same design as the existing module, with two interfaces that each have one method. Of course, the replacement for `RenderAttachedBlockView` no longer extends `BlockRenderView`. Some terminology and class member names were changed when creating the new module. The object that used to be termed "Render Data Attachment" or "Render Attachment" has been renamed to "Block Entity Render Data" or simply "Render Data". All methods, documentation, and other mentions of the old terms, apart from within the existing module, have been changed to use the new terms. Additionally, the documentation was rewritten to be more concise, provide more examples, remove mentions of implementation details, and add concrete requirements for use of certain methods.

The existing module has been completely deprecated and its documentation has been updated to state the contract that is actually satisfied and link to the v2 replacements. It was not possible to achieve perfect backward and forward compatibility since this would result in infinite recursion at first thought and degraded performance with an actual implementation. To compromise, it was observed that the `BlockRenderView` extension interface is invoked much more frequently than it is implemented, and that the `BlockEntity` extension interface is implemented much more frequently than it is invoked. As such, it was decided that the v1 `BlockRenderView` extension interface should proxy to v2 and the v2 `BlockEntity` extension interface should proxy to v1.


fabric-render-data-v2/src/client/java/net/fabricmc/fabric/impl/renderdata/client/RenderDataMapConsumer.java
```diff
@@ -14,10 +14,10 @@
...
-package net.fabricmc.fabric.impl.rendering.data.attachment;
+package net.fabricmc.fabric.impl.renderdata.client;
 
-import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 
-public interface RenderDataObjectConsumer {
-	void fabric_acceptRenderDataObjects(Long2ObjectOpenHashMap<Object> renderDataObjects);
+public interface RenderDataMapConsumer {
+	void fabric$acceptRenderDataMap(Long2ObjectMap<Object> renderDataMap);
```
Technici4n: nit: we use `fabric_` more or less everywhere

fabric-render-data-v2/src/client/java/net/fabricmc/fabric/mixin/renderdata/client/ChunkRendererRegionBuilderMixin.java
```diff
@@ -14,52 +14,58 @@
...
 				// Hash maps in chunks should generally not be modified outside of client thread
 				// but does happen in practice, due to mods or inconsistent vanilla behaviors, causing
-				// CMEs when we iterate the map.  (Vanilla does not iterate these maps when it builds
+				// CMEs when we iterate the map. (Vanilla does not iterate these maps when it builds
 				// the chunk cache and does not suffer from this problem.)
 				//
-				// We handle this simply by retrying until it works.  Ugly but effective.
-				for (;;) {
+				// We handle this simply by retrying until it works. Ugly but effective.
+				while (true) {
```
Technici4n: Sad that you remove the amazing `for (;;)` :smile:

fabric-render-data-v2/src/client/java/net/fabricmc/fabric/mixin/renderdata/client/ChunkRendererRegionBuilderMixin.java
```diff
@@ -14,52 +14,58 @@
...
 
 @Mixin(ChunkRendererRegionBuilder.class)
 public abstract class ChunkRendererRegionBuilderMixin {
 	private static final AtomicInteger ERROR_COUNTER = new AtomicInteger();
 	private static final Logger LOGGER = LoggerFactory.getLogger(ChunkRendererRegionBuilderMixin.class);
 
-	@Inject(at = @At("RETURN"), method = "build", locals = LocalCapture.CAPTURE_FAILHARD)
-	private void create(World world, BlockPos startPos, BlockPos endPos, int chunkRadius, CallbackInfoReturnable<ChunkRendererRegion> info, int i, int j, int k, int l, ChunkRendererRegionBuilder.ClientChunk[][] chunkData) {
+	@Inject(method = "build(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/util/math/BlockPos;I)Lnet/minecraft/client/render/chunk/ChunkRendererRegion;", at = @At("RETURN"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private void fabric$createDataMap(World world, BlockPos startPos, BlockPos endPos, int offset, CallbackInfoReturnable<ChunkRendererRegion> cir, int startX, int startZ, int endX, int endZ, ChunkRendererRegionBuilder.ClientChunk[][] chunksXZ) {
```
Technici4n: You don't need to specify the parameter types if there's a single method named `build`.

fabric-render-data-v2/src/client/java/net/fabricmc/fabric/mixin/renderdata/client/ChunkRendererRegionBuilderMixin.java
```diff
@@ -76,35 +82,34 @@ private void create(World world, BlockPos startPos, BlockPos endPos, int chunkRa
...
-		if (map != null && rendererRegion != null) {
-			((RenderDataObjectConsumer) rendererRegion).fabric_acceptRenderDataObjects(map);
+		if (map != null) {
+			((RenderDataMapConsumer) rendererRegion).fabric$acceptRenderDataMap(map);
 		}
 	}
 
-	private static Long2ObjectOpenHashMap<Object> mapChunk(WorldChunk chunk, BlockPos posFrom, BlockPos posTo, Long2ObjectOpenHashMap<Object> map) {
+	@Unique
+	private static Long2ObjectOpenHashMap<Object> fabric$mapChunk(WorldChunk chunk, BlockPos posFrom, BlockPos posTo, Long2ObjectOpenHashMap<Object> map) {
```
Technici4n: `fabric$` prefix is unnecessary here, mixin will add it (at last Fabric Mixin).

fabric-render-data-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.util.math.BlockPos;
+
+/**
+ * Extensions that allow {@link BlockEntity} subclasses to provide render data.
+ *
+ * <p>Block entity render data is arbitrary data that is safe to access and use in a client-side
```
Technici4n: I think this should explain that it is retrieved from the BlockRenderView, and why the BE shouldn't be used directly (it might be modified concurrently on the client side).

Technici4n: Maybe something like `Directly accessing the fields in a block entity is not safe in a multi-threaded environment, such as chunk building, as the block entity might be modified concurrently on the main thread. Instead, the block entity should first provide the relevant data in getRenderData, and the code that needs this data can then query it from the BlockRenderView.` in the very first paragraph.

fabric-render-data-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity.java
```diff
@@ -0,0 +1,61 @@
...
+ * @see RenderDataBlockView
+ */
+public interface RenderDataBlockEntity {
+	/**
+	 * Gets the render data provided by this block entity. The returned object must be safe to
+	 * use in a multithreaded environment.
+	 *
+	 * <p>Note: This method should not be called directly, unless the result is used to implement
+	 * {@link RenderDataBlockView#getBlockEntityRenderData(BlockPos)}. Otherwise, use
+	 * {@link RenderDataBlockView#getBlockEntityRenderData(BlockPos)} to access render data instead.
```
Technici4n: nit: otherwise and instead in the same sentence

fabric-render-data-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockView.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @see RenderDataBlockEntity
+ */
+public interface RenderDataBlockView {
+	/**
+	 * Retrieves block entity render data for a given block position. Always use this method
+	 * instead of {@link RenderDataBlockEntity#getRenderData()} to ensure thread safety and
+	 * data consistency.
+	 *
+	 * <p>Callers of this method should always check the returned object's class (for example,
```
Technici4n: Should they, actually? Are there cases where the block entity data does not match the block state?

fabric-render-data-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity.java
Technici4n: Ah I see that there's some explanation in `RenderDataBlockView` already. Still, it would be good to have something here too.

fabric-render-data-v2/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,38 @@
...
+    }
+  ],
+  "custom": {
+    "fabric-api:module-lifecycle": "stable",
+    "loom:injected_interfaces": {
+      "net/minecraft/class_2586": ["net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity"],
+      "net/minecraft/class_1920": ["net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockView"]
+    }
+  },
+  "accessWidener": "fabric-render-data-v2.accesswidener"
```
Technici4n: there's no AW, right?

fabric-render-data-v2/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,38 @@
...
+    "homepage": "https://fabricmc.net",
+    "irc": "irc://irc.esper.net:6667/fabric",
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.4.0",
```
Technici4n: Should be whatever is used at the current in Fabric API.

fabric-render-data-v2/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,38 @@
...
+    "fabric-api-base": "*"
```
Technici4n: Are we using it? If not we can leave it out

fabric-render-data-v2/src/main/resources/fabric-render-data-v2.mixins.json
```diff
@@ -0,0 +1,12 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.renderdata",
+  "compatibilityLevel": "JAVA_16",
```
Technici4n: I'm not sure this is correct, someone would have to check (the compatibility level setting is a mess in any case).

Technici4n (changes requested):

fabric-render-data-v2/src/main/resources/fabric.mod.json
PepperCode1: There is. It was renamed but not changed.

fabric-render-data-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockView.java
PepperCode1: Discussed on Discord; it was decided that the data should always be checked for null and incorrect types. The javadoc was adjusted in the latest commit to reflect and clarify what is expected of users of this method.

fabric-render-data-api-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity.java
```diff
@@ -0,0 +1,69 @@
...
+ * <p>Note: This interface is automatically implemented on all {@link BlockEntity} instances via Mixin and interface injection.
+ *
+ * @see RenderDataBlockView
+ */
+public interface RenderDataBlockEntity {
+	/**
+	 * Gets the render data provided by this block entity. The returned object must be safe to
+	 * use in a multithreaded environment.
+	 *
+	 * <p>Note: <b>This method should not be called directly</b>; use
```
Technici4n, line 58: Should we add `@ApiStatus.OverrideOnly` to the method?

Technici4n (approved):
All good now. It would be great to test backwards compatibility in prod, just in case.

modmuss50:
Might be a nice idea to edit the PR description to include a short rundown of what is diffrent with this new API as its not that obvious, and possibly a migration guide. I usually copy these into the commit when suqishing+merging.

Technici4n:
Yes this would be nice. We could then reference the PR in the update announcement.

Something else we need at the ready for the announcement is the gradle invocation to include the old API back in dev:

"The old v0 rendering data attachment module is deprecated, and therefore not in the default dev jar anymore. Here is how to add it back if you don't want to update immediately:

<gradle diff>"

fabric-render-data-api-v2/src/main/java/net/fabricmc/fabric/api/renderdata/v2/RenderDataBlockEntity.java
PepperCode1: I do not think so, because there is no distinction between API users and implementors; both are still users. This even applies to FRAPI, where interfaces that are to be implemented by the renderer are not marked with this annotation. This annotation should be reserved exclusively for methods that are implemented only by the module implementation, which is not the case here.

modmuss50 (approved):

PepperCode1:
Superseded by #3268.

PR #3141, opened 2023-06-24
Fix default fluid names for non-placeable fluids, clarify Storage iterator lifecycle with modification

### LINKED ISSUE #3114: Injected getFluidName method for fluids
Currently, if you only want to provide a name for a fluid that has no block, it's pretty cumbersome and needs you register an additional object like this:

```java
FluidVariantAttributes.register(MY_FLUID, new FluidVariantAttributeHandler() {
    @Override
    public Text getName(FluidVariant variant) {
        return Text.translatable("fluid.my_mod.my_fluid");
    }
});
```
I'm proposing a `FabricFluid` injected interface with these methods:
```java
public interface FabricFluid {
    default Text getFluidName() {
        // the current default in FluidVariantAttributeHandler
        return ((Fluid) this).getDefaultState().getBlockState().getBlock().getName();
    }

    // If this interface is part of the transfer API:
    default Text getFluidName(FluidVariant variant) {
        return getFluidName();
    }
}
```

The default fluid variant attribute handler would then delegate to these methods.

(`getFluidName` so that a fluid class doesn't end up inheriting both this unobfuscated method and `method_5477` - present in many classes such as [`Nameable`](https://github.com/FabricMC/yarn/blob/de6b90b586e5fe8cb814a134cb3bb9a307acd8a6/mappings/net/minecraft/util/Nameable.mapping#L10) - which would cause a remapping conflict)

### LINKED ISSUE #3123: Define that `Storage#iterator` must be inside of a transaction if modification is expected.
See discussion at https://discord.com/channels/507304429255393322/566417932503089232/1118520686625968209.

Technici4n:
Fixes #3114. Fixes #3123.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
@@ -47,7 +47,7 @@ public interface FluidVariantAttributeHandler {
 	default Text getName(FluidVariant fluidVariant) {
 		Block fluidBlock = fluidVariant.getFluid().getDefaultState().getBlockState().getBlock();
 
-		if (fluidBlock == Blocks.AIR) {
+		if (!fluidVariant.isBlank() && fluidBlock == Blocks.AIR) {
 			// Some non-placeable fluids use air as their fluid block, in that case infer translation key from the fluid id.
 			return Text.translatable(Util.createTranslationKey("block", Registries.FLUID.getId(fluidVariant.getFluid())));
 		} else {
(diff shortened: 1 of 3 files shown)


modmuss50 (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
```diff
@@ -41,7 +45,14 @@ public interface FluidVariantAttributeHandler {
 	 * Return the name that should be used for the passed fluid variant.
 	 */
 	default Text getName(FluidVariant fluidVariant) {
-		return fluidVariant.getFluid().getDefaultState().getBlockState().getBlock().getName();
+		Block fluidBlock = fluidVariant.getFluid().getDefaultState().getBlockState().getBlock();
+
+		if (fluidBlock == Blocks.AIR) {
```
Juuxel: This check for `fluidVariant.getFluid() != Fluids.EMPTY` as the translation key would be different from the current one and would have no translation.

Juuxel (changes requested):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariantAttributeHandler.java
Technici4n: Whoops thank you so much, this is embarrassing. 😄

Juuxel (approved):

PR #3142, opened 2023-06-24
Model load observation API

embeddedt:
The following PR implements a standardized API mods such as Continuity can use to have low-level control over what model instances are loaded and baked within the `ModelLoader`. One benefit of standardizing this system is that it will allow optimization mods such as my own to safely implement dynamic model loading/unloading at runtime without running into compatibility problems when mods that inject into `ModelLoader` directly. It also makes it easy for modders to replace models without needing to learn exactly how the model loader works.

EDIT: Will run checkstyle after other concerns are addressed.


Technici4n:
This now lives inside of #3145.

Issue #571, opened 2020-04-12
ModelVariantProviders not called for every state if one state loaded by vanilla


TwilightFlower:
If a ModelVariantProvider chooses to not override one state and lets it load the vanilla way, it will not be called for any further states of that block. This is a rather niche bug but I'm having trouble finding a workaround.

Jkmcameron:
Bump, still an issue. Not much else to be said other than what is said here. Not really niche at this point, seems more like an oversight

Technici4n:
This will be addressed by the Model Loading API v1: #3145.

Technici4n:
Fixed in #3145.

PR #3145, opened 2023-06-25
Model Loading API v1

Technici4n:
This PR deprecates and supersedes the old `fabric-models-v0` module.

### Model loading hooks
Mods now register a single `ModelLoadingPlugin` for any change they want to make to the model loading process. (Or multiple plugins if they want to). This replaces the old `ModelLoadingRegistry`.

Here is an example:

```java
ModelLoadingPlugin.register(pluginContext -> {
    // ResourceManager access is provided like in the v0 API, but in a central place, and can be used for shared processing in the plugin.
    ResourceManager manager = pluginContext.resourceManager();

    // Model resource providers still exist, but they are called model resolvers now
    pluginContext.resolveModel().register(context -> {
        // ...
    });
});
```

#### `ModelVariantProvider` -> `BlockStateResolver`
`ModelVariantProvider` was heavily reworked, and is now replaced by the `BlockStateResolver`, with a much better defined contract.

#### `ModelResourceProvider` -> `ModelResolver`
The resource provider is mostly unchanged. The biggest difference is that it is now registered as an event listener. This allows mods to use event phases for ordering between conflicting ~~providers~~ resolvers.

#### Removed custom exception
Additionally, `ModelProviderException` could be thrown by a variant or resource provider in the v0 API. This was not explained in the documentation, and would according to the code stop further processing of the providers and log an error.

In the new API, any `Exception` is caught and logged. If that happens, the other resolvers are still processed. There is no custom `Exception` subclass anymore.

### Helper method to get a `BakedModel` by `Identifier` from the manager
The v0 had such a method in a helper class: `BakedModelManagerHelper#getBakedModel`. It is now interface-injected instead. See `FabricBakedModelManager`.

## New model wrapping hooks
New hooks are added for the various needs of mods that want to override or wrap specific models. Thanks to @embeddedt for contributing an initial version of them!

Here is an example of wrapping the gold model to remove the bottom quads, for example:
```java
ModelLoadingPlugin.register(pluginContext -> {
	// remove bottom face of gold blocks
	pluginContext.modifyModelAfterBake().register(ModelModifier.WRAP_PHASE, (model, context) -> {
		if (context.identifier().getPath().equals("block/gold_block")) {
			return new DownQuadRemovingModel(model);
		} else {
			return model;
		}
	});
});
```

There are 3 events, for the following use cases:
- Wrapping `UnbakedModel`s right when they are loaded. This allows replacing them entirely in dependent models too.
- Wrapping `UnbakedModel`s right before they are baked. This allows replacing them without affecting dependent models (which might not be expecting a model class change).
- Wrapping `BakedModel`s right when they are baked.

Multiple mods have implemented their own version of them. Providing them in Fabric API will make it easier on these mods, and will additionally allow optimization mods that perform on-demand model loading to simply fire the hooks themselves instead of working around injections performed by other mods.


fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelVariantProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ * also allow you to add your own block- or item-state formats. To trigger the loading
+ * of another model, use the passed {@link ModelProviderContext}.
+ *
+ * <p>As every model loading is instantiated with a new provider, it is safe
+ * (and recommended!) to cache information.
+ *
+ * <p>Keep in mind that only *one* ModelVariantProvider may respond to a given model
+ * at any time.
+ */
+// TODO: add warning that vanilla will load all block variants at once (can't intercept only a single one)
```
Technici4n: This refers to issue #571. I believe this can be solved by replacing models using embeddedt's extra hooks (coming soon).

Technici4n: Now addressed.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoaderInstance.java
```diff
@@ -0,0 +1,124 @@
...
+				plugin.onInitializeModelLoader(context);
+			} catch (Exception exception) {
+				LOGGER.error("Failed to run model loading plugin {}", plugin.getClass().getName(), exception);
+			}
+		}
+	}
+
+	@Override
+	public UnbakedModel loadModel(Identifier id) {
+		return ((ModelLoaderHooks) loader).fabric_loadModel(id);
```
Technici4n: This is suspicious... If the model is already loaded, it will be loaded again and overridden?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelResourceProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * <p>Keep in mind that only *one* ModelResourceProvider may respond to a given model
+ * at any time. If you're writing, say, an OBJ loader, this means you could
+ * easily conflict with another OBJ loader unless you take some precautions,
+ * for example:
+ *
+ * <ul><li>Only load files with a mod-suffixed name, such as .architect.obj,
+ * <li>Only load files from an explicit list of namespaces, registered elsewhere.</ul>
+ */
+@FunctionalInterface
+public interface ModelResourceProvider {
```
Technici4n: Should both `ModelResourceProvider` and `ModelVariantProvider` be refactored to the same outer class? (`ModelProvider.Resource` and `ModelProvider.Variant`) This would align their design with the `ModelModifier`s.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelLoadingPlugin.java
```diff
@@ -0,0 +1,86 @@
...
+
+		/**
+		 * Event access to register model resource providers.
+		 */
+		Event<ModelResourceProvider> resourceProviders();
+
+		/**
+		 * Event access to monitor unbaked model loads and replace the loaded model.
+		 */
+		Event<ModelModifier.Unbaked> onUnbakedModelLoad();
```
Technici4n: Will need some triage help to decide the naming of the event getters.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
```diff
@@ -0,0 +1,127 @@
...
+		 * Context for a baked model load event.
+		 *
+		 * @param identifier the identifier of this model (may be a {@link ModelIdentifier})
+		 * @param sourceModel the unbaked model that is being baked
+		 * @param textureGetter function that can be used to retrieve sprites
+		 * @param settings the settings this model is being baked with
+		 * @param baker the baker being used to bake this model
+		 * @param loader the current model loader instance (changes when resource packs reload)
+		 */
+		record Context(
```
modmuss50: I think an interface would be better (can still be implimented by an impl record). Its easier to maintain and will likely make the documentation better exposed?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelResourceProvider.java
Technici4n: Done.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelLoadingPlugin.java
Technici4n: Naming should have improved already.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
Technici4n: Moved to interfaces.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
```diff
@@ -0,0 +1,152 @@
...
+	 */
+	public static final Identifier WRAP_PHASE = new Identifier("fabric", "wrap");
+	/**
+	 * Recommended phase to use when wrapping models with transformations that want to happen last,
+	 * e.g. for connected textures or other similar visual effects that should be the final processing step.
+	 */
+	public static final Identifier WRAP_LAST_PHASE = new Identifier("fabric", "wrap_last");
+
+	@FunctionalInterface
+	public interface Unbaked {
```
Technici4n: This needs to be split up across two events because one of them can receive more context.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
```diff
@@ -0,0 +1,152 @@
...
+		 * @param model the current unbaked model instance
+		 * @param context context with additional information about the model/loader
+		 * @return the model that should be used in this scenario. If no changes are needed, just return {@code model} as-is.
+		 */
+		UnbakedModel modifyUnbakedModel(UnbakedModel model, Context context);
+
+		/**
+		 * Context for an unbaked model load/pre-bake event.
+		 */
+		interface Context {
```
Technici4n, line 135: It might be useful to provide a hook to load a model here.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelResolver.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.client.util.ModelIdentifier;
+import net.minecraft.util.Identifier;
+
+/**
+ * Contains interfaces for the events mods can use to load their own model formats.
+ *
+ * <p>These hooks fire for <b>every single model that's loaded</b>, so the code written here should be as simple/performant
+ * as possible.
+ */
+public final class ModelResolver {
```
Technici4n: Needs a second pass to eliminate all references to "model <thing> providers"

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoaderInstance.java
Technici4n: Replaced by `getOrLoad` model calls.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/ModelLoaderMixin.java
```diff
@@ -0,0 +1,201 @@
...
+
+	@Override
+	public void fabric_queueModelDependencies(UnbakedModel model) {
+		modelsToLoad.addAll(model.getModelDependencies());
+	}
+
+	@Override
+	public JsonUnbakedModel fabric_loadModelFromJson(Identifier id) {
+		return loadModelFromJson(id);
+	}
```
Juuxel, line 200: Is this needed? The shadowed method seems to be public.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/ModelLoaderMixin.java
```diff
@@ -0,0 +1,201 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.profiler.Profiler;
+
+import net.fabricmc.fabric.impl.client.model.loading.ModelLoaderHooks;
+import net.fabricmc.fabric.impl.client.model.loading.ModelLoadingEventDispatcher;
+import net.fabricmc.fabric.impl.client.model.loading.ModelLoadingPluginManager;
+
+@Mixin(ModelLoader.class)
+public abstract class ModelLoaderMixin implements ModelLoaderHooks {
+	// this is the first one
```
Juuxel: The first one?

fabric-model-loading-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/model/loading/ModelTestModClient.java
```diff
@@ -0,0 +1,122 @@
...
+
+				context.setModel(state.with(CropBlock.AGE, 7), context.getOrLoadModel(new Identifier("block/wheat_stage7")));
+			});
+		});
+
+		ResourceManagerHelper.get(ResourceType.CLIENT_RESOURCES).registerReloadListener(SpecificModelReloadListener.INSTANCE);
+
+		LivingEntityFeatureRendererRegistrationCallback.EVENT.register((entityType, entityRenderer, registrationHelper, context) -> {
+			if (entityRenderer instanceof PlayerEntityRenderer) {
+				registrationHelper.register(new BakedModelFeatureRenderer<>((PlayerEntityRenderer) entityRenderer, SpecificModelReloadListener.INSTANCE::getSpecificModel));
```
Juuxel:
```suggestion
			if (entityRenderer instanceof PlayerEntityRenderer playerRenderer) {
				registrationHelper.register(new BakedModelFeatureRenderer<>(playerRenderer, SpecificModelReloadListener.INSTANCE::getSpecificModel));
```
We are not on Java 8 anymore

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
```diff
@@ -0,0 +1,220 @@
...
+			/**
+			 * The identifier of this model (may be a {@link ModelIdentifier}).
+			 */
+			Identifier id();
+
+			/**
+			 * Function that can be used to retrieve sprites.
+			 */
+			Function<SpriteIdentifier, Sprite> textureGetter();
```
Juuxel, line 144: Should this be `Sprite getTexture(SpriteIdentifier)` instead? Exposing functional interfaces as method return values is a bit strange.

deprecated/fabric-models-v0/src/client/java/net/fabricmc/fabric/api/client/model/ModelVariantProvider.java
Juuxel, line 1: If a mod uses `ModelIdentifier`-based custom models but not for blocks, this PR seems to leave them in a bit of trouble since there is tno moreModelIdentifier ⇒ UnbakedModel hook available for those uses.

This could apply to e.g. multipart models (the microblock kind, not the `blockstates` json type) in mods like mine that add stateful multiparts with block state-like property definitions.

**Edit:** See Discord, this can be marked as resolved

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BakedModelManagerMixin.java
```diff
@@ -0,0 +1,81 @@
...
+	public BakedModel getModel(Identifier id) {
+		return models.get(id);
+	}
+
+	@Redirect(
+			method = "reload",
+			at = @At(
+					value = "INVOKE",
+					target = "java/util/concurrent/CompletableFuture.thenCombineAsync(Ljava/util/concurrent/CompletionStage;Ljava/util/function/BiFunction;Ljava/util/concurrent/Executor;)Ljava/util/concurrent/CompletableFuture;"
+			),
```
Juuxel: I think this should be `remap = false`?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoadingPluginManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+	public static final ThreadLocal<List<ModelLoadingPlugin>> CURRENT_PLUGINS = new ThreadLocal<>();
+
+	public static void registerPlugin(ModelLoadingPlugin plugin) {
+		PLUGINS.add(plugin);
+	}
+
+	public static <T> void registerPlugin(PreparableModelLoadingPlugin.DataLoader<T> loader, PreparableModelLoadingPlugin<T> plugin) {
+		PREPARABLE_PLUGINS.add(new PreparablePluginHolder<>(loader, plugin));
+	}
```
Juuxel: Should these be null-checked to prevent later crashes?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelResolver.java
```diff
@@ -0,0 +1,86 @@
...
+
+import net.minecraft.client.render.model.ModelLoader;
+import net.minecraft.client.render.model.UnbakedModel;
+import net.minecraft.client.util.ModelIdentifier;
+import net.minecraft.util.Identifier;
+
+/**
+ * Interface for model resolvers.
+ *
+ * <p>Model resolvers hook the loading of model *files* from the resource tree;
```
Juuxel:
```suggestion
 * <p>Model resolvers hook the loading of model <em>files</em> from the resource tree;
```

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoadingEventDispatcher.java
```diff
@@ -0,0 +1,465 @@
...
+			extraModelConsumer.accept(id);
+		}
+	}
+
+	/**
+	 * @return {@code true} to cancel the vanilla method
+	 */
+	public boolean loadModel(Identifier id) {
+		if (id instanceof ModelIdentifier modelId) {
+			if (Objects.equals(modelId.getVariant(), "inventory")) {
```
Juuxel:
```suggestion
			if ("inventory".equals(modelId.getVariant())) {
```

Juuxel (changes requested):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/ModelLoaderMixin.java
Technici4n: image

It's private.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelResolver.java
Technici4n: Not relevant anymore after Pepper's javadoc changes.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
Technici4n: This mimics what `UnbakedModel` receives - as discussed on Discord it's fine either way.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java
```diff
@@ -0,0 +1,220 @@
...
+import net.fabricmc.fabric.api.event.Event;
+
+/**
+ * Contains interfaces for the events that can be used to modify models at different points in the loading and baking
+ * process.
+ *
+ * <p>Example use cases:
+ * <ul>
+ *     <li>Overriding a model for a particular blockstate - check if the given identifier is a {@link ModelIdentifier},
+ *     and then check if it has the appropriate variant for that blockstate. If so, return your desired model,
```
Juuxel:
```suggestion
 *     <li>Overriding a model for a particular block state - check if the given identifier is a {@link ModelIdentifier},
 *     and then check if it has the appropriate variant for that block state. If so, return your desired model,
```

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoadingEventDispatcher.java
```diff
@@ -0,0 +1,465 @@
...
+		@Override
+		public void setModel(BlockState state, UnbakedModel model) {
+			Objects.requireNonNull(model, "state cannot be null");
+			Objects.requireNonNull(model, "model cannot be null");
+
+			if (!state.isOf(block)) {
+				throw new IllegalArgumentException("Attempted to set model for state " + state + " on block " + block);
+			}
+
+			if (models.put(state, model) != null) {
```
Juuxel:
```suggestion
			if (models.putIfAbsent(state, model) != null) {
```
We use `putIfAbsent` for similar code in other modules

Juuxel (approved):

modmuss50 (approved):

PR #3155, opened 2023-06-29
Camera Update Callback

dhyces:
This callback is a simple implementation for a callback after the `Camera#update` method is called from the `WorldRenderer`. The mixin itself is rather trivial, though due to the nature of rendering code (changing values to get it just right and hotswapping), it requires a public static method outside of the mixin class when not using mixin hotswap. The included test slowly rotates the camera upward when the player sleeps in a bed.


fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/CameraUpdateCallback.java
```diff
@@ -0,0 +1,47 @@
...
+
+import net.minecraft.entity.Entity;
+import net.minecraft.world.BlockView;
+import net.minecraft.client.render.Camera;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Callback for after the game camera is updated.
```
Technici4n: The javadoc should describe what the event can do. I assume the goal is to modify the camera to implement various rendering effects?

dhyces: Copied from my response on discord:

> ah, right, sorry, meant to respond yesterday and forgot. Yeah, so the event is for modifying the camera after it's been updated for third person and for the sleep angle and position if the entity is sleeping. vanilla calls this in WorldRenderer#render to reset or setup the camera, so yes, it's reset to vanilla's handling if a mod stops modifying it

fabric-rendering-v1/src/client/resources/fabric-rendering-v1.accesswidener
```diff
@@ -0,0 +1,6 @@
+accessWidener	v2	named
+
+accessible method net/minecraft/client/render/Camera moveBy (DDD)V
```
Technici4n: Please use an accessor mixin in the testmod instead - creating a new AW adds a MC jar to the gradle project, so it should be avoided when possible. ;)

dhyces: That doesn't sound very user friendly, since it would require every user of the api to create a camera accessor to modify its transformation. Should a public facing accessor mixin be preferred over an AW?

Juuxel: This access widener would not be exposed to users of this API anyway as it's not transitive.

dhyces: fair point, I forgot I moved it

Technici4n (approved):

modmuss50 (commented):
Im not 100% convinced that this fits into fabric API's scope TBH.

Im not sure on the number of mods that are actually going to use this? Its a trivial mixin that can be added to your own mods. I dont want to end up with too many of these rendering related events as they will have some perf overhead, even if very small.

The actual change here looks fine, but I think we need to discuss if we actually need this.

PR #3163, opened 2023-07-02
Add dynamic registry API

### LINKED ISSUE #1012: Hooks for adding custom registries into DynamicRegistryManager

Juuxel:
Closes #1012, supersedes #1031 and #2719.

This PR also fixes the tag paths for modded dynamic registries by adding the registry namespace to them:
```
data/<tag namespace>/tags/<registry namespace>/<registry path>/<tag path>.json
                          ^~~~~~~~~~~~~~~~~~~~^
                                 added
```


fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/TestDynamicObject.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.fabric.test.registry.sync;
+
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+
+public record TestDynamicObject(String name) {
+	public static final Codec<TestDynamicObject> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+			Codec.STRING.fieldOf("name").forGetter(TestDynamicObject::name)
+	).apply(instance, TestDynamicObject::new));
```
maityyy: Perhaps that's better?
```java
public static final Codec<TestDynamicObject> CODEC = Codec.STRING
        .fieldOf("name")
        .xmap(TestDynamicObject::new, TestDynamicObject::name)
        .codec();
```

apple502j: Test code should be something modders can copy and adapt. Since one-field data is rare I think the current code is better for that purpose.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+
+	private DynamicRegistryEvents() {
+	}
+
+	/**
+	 * Collects all entries for dynamic registries.
+	 *
+	 * @return the entries
+	 */
+	public static List<RegistryLoader.Entry<?>> collectDynamicRegistries() {
```
Technici4n: Is this exposed in case mods need it?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	public static final Event<DynamicRegistrySetupCallback> BEFORE_LOAD = DynamicRegistrySetupCallback.EVENT;
+
+	// Phases
+
+	/**
+	 * A phase of {@link #REGISTER_REGISTRIES} when vanilla dynamic registries are registered.
+	 * This phase executes before {@linkplain Event#DEFAULT_PHASE the default phase}.
+	 */
+	public static final Identifier REGISTER_VANILLA_REGISTRIES_PHASE = new Identifier("fabric-api", "register_vanilla_registries");
```
Technici4n: IMO this is insufficient and needs to use a topological sort between registries, in case mods need to load some registry entries before a specific vanilla registry. Should I send you a PR with the relevant event phase rework to allow the system to be used here?

Technici4n (changes requested):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Technici4n: Does this need to be integrated with the datagen API?

Juuxel: Wdym integrated? This PR makes the data generation API use this method instead of the vanilla field.

Juuxel: Yep, custom data generators and the like might want this. It's also used to share the implementation between its usages in for actual registry loading in game and for loading regs for data generation.

fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/TestDynamicObject.java
Juuxel: I don't think it really matters which variant we use here. Using `RecordCodecBuilder` might be more recognisable to people reading this code since it's used way more often in vanilla and mods than converting a single codec `fieldOf` to a full codec.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Technici4n: Ah right, missed that.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+	/**
+	 * An event for registering new dynamic registries.
+	 */
+	public static final Event<RegisterRegistries> REGISTER_REGISTRIES = EventFactory.createArrayBacked(RegisterRegistries.class, callbacks -> context -> {
+		for (RegisterRegistries callback : callbacks) {
+			callback.registerDynamicRegistries(context);
+		}
+	});
+
+	// TODO: is this good to have?
```
Technici4n: IMO no, this still depends on the outside functional interface existing.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Events relating to dynamic registries.
+ */
+public final class DynamicRegistryEvents {
+	// Events
+
+	/**
+	 * An event for registering new dynamic registries.
+	 */
+	public static final Event<RegisterRegistries> REGISTER_REGISTRIES = EventFactory.createArrayBacked(RegisterRegistries.class, callbacks -> context -> {
```
Technici4n: What about just `REGISTER`?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Collects all entries for dynamic registries.
+	 *
+	 * @return the entries
+	 */
+	public static List<RegistryLoader.Entry<?>> collectDynamicRegistries() {
+		var context = new DynamicRegistryRegistrationContextImpl();
+		REGISTER_REGISTRIES.invoker().registerDynamicRegistries(context);
+		return context.getEntries();
```
Technici4n: Should the list be made unmodifiable?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
```diff
@@ -0,0 +1,135 @@
...
+		void add(RegistryLoader.Entry<?>... entries);
+
+		/**
+		 * Registers dynamic registries.
+		 *
+		 * <p>The collection must not contain duplicates.
+		 *
+		 * @param entries the registry loader entries representing the dynamic registries
+		 */
+		void add(Collection<? extends RegistryLoader.Entry<?>> entries);
```
Technici4n: I don't really see the point of these methods - if we remove them, then we can also get rid of the context object and use static registration instead.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Juuxel: Yeah, I thought it could be nice if we also had #2111 here but I think I'll make it its own PR.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Juuxel: I thought about that too, but `REGISTER` also sounds like an event that registers *entries* in dynamic registries. This is a bit more clear.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Juuxel: True, I think my design with an event was influenced by the earlier PRs. I think I can change it to something like this:

```java
class DynamicRegistries {
  static <T> void register(RegistryKey<? extends Registry<T>> key, Codec<T> codec);
}
```

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Juuxel: Sure.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistryEvents.java
Technici4n: Yes I think that makes the most sense - with some ordering too of course. :)

fabric-registry-sync-v0/build.gradle
```diff
@@ -13,4 +13,5 @@ moduleDependencies(project, [
 testDependencies(project, [
 	':fabric-lifecycle-events-v1',
 	':fabric-command-api-v2',
+	':fabric-resource-loader-v0',
```
Technici4n: I think it's already included by default for all modules

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java
```diff
@@ -0,0 +1,104 @@
...
+	 * @return an unmodifiable list of all dynamic registries
+	 */
+	public static @Unmodifiable List<RegistryLoader.Entry<?>> getDynamicRegistries() {
+		return DynamicRegistriesImpl.getDynamicRegistries();
+	}
+
+	/**
+	 * Registers a non-synced dynamic registry.
+	 *
+	 * <p>The entries of the registry will be loaded from data packs at the file path
```
Technici4n, line 91: Question: what is the path for tags?

fabric-registry-sync-v0/build.gradle
Juuxel: TIL, I'll remove it from here then.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java
Juuxel: Currently unadjusted from vanilla's unprefixed directory - I meant to make a PR to fix it separately (as it also affects static registries), but forgot... I can add it here too.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistrySyncOption.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.event.registry;
+
+/**
+ * Flags for configuring dynamic registry syncing.
+ */
+public enum DynamicRegistrySyncOption {
```
Technici4n: Can we put that in the main class?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistrySyncOption.java
```diff
@@ -0,0 +1,29 @@
...
+	/**
+	 * Only synchronizes the dynamic registry if it's not empty.
+	 * This is useful for compatibility with vanilla clients,
+	 * or other clients that do not have the registry.
```
Technici4n:
```suggestion
	 * or other clients that might not have the registry.
```

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java
```diff
@@ -0,0 +1,154 @@
...
+import org.jetbrains.annotations.Unmodifiable;
+
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.RegistryLoader;
+
+import net.fabricmc.fabric.impl.registry.sync.DynamicRegistriesImpl;
+
+/**
+ * Contai§ns methods for registering and accessing dynamic {@linkplain Registry registries}.
```
Technici4n:
```suggestion
 * Contains methods for registering and accessing dynamic {@linkplain Registry registries}.
```

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	/**
+	 * Returns an unmodifiable list of all dynamic registries, including modded ones.
+	 *
+	 * <p>The list will not reflect any changes caused by later registrations.
+	 *
+	 * @return an unmodifiable list of all dynamic registries
+	 */
+	public static @Unmodifiable List<RegistryLoader.Entry<?>> getDynamicRegistries() {
```
Technici4n, line 84: Annotation on a separate line?

Technici4n (approved):
LGTM

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java
Juuxel: It's not an annotation on the method, it's an annotation on the type. `@Unmodifiable` only works on `TYPE_USE`.

Technici4n: Same as `@Nullable`. They're equivalent in bytecode I believe.

Juuxel: No, it's different since `@Nullable` is also a method annotation. When you use `@Nullable` on a method return type, it makes it annotate both the return type and the method since it can be applied to either one. That's why it shows up twice for all usages in the Yarn javadoc, for example.

`@Unmodifiable` differs in that it's only a type use annotation, so it should go with the type IMO.

Technici4n: Sure

apple502j (approved):

modmuss50 (approved):

PR #3164, opened 2023-07-02
Fix injected resources not overriding certain vanilla resources (1.20)

NebelNidas:
Mods can override vanilla textures just fine by placing files into `resources/assets/minecraft`:
Default

However, the same doesn't work (at least not consistently) for injected resources (via `resources/programmer_art/assets/minecraft` for example):
Injected Current

This PR fixes the inconsistent behavior:
Injected Fixed


Technici4n (changes requested):
Looks good indeed.

There's a few related problems that should be fixed too:
- `GroupResourcePack#findResources` should iterate first-to-last.
- `GroupResourcePack#appendResources` should iterate last-to-first.
Can you add them to this PR?

NebelNidas:
Done! I've also removed some redundant code while being at it

modmuss50:
The build failure is due to checkstyle 👍

NebelNidas:
Oops, should be fixed now

fabric-resource-loader-v0/src/testmod/resources/assets/minecraft/textures/block/dirt.png
modmuss50, line 1: I dont think we want to override dirt and gravel in the test mod, its a little bit invasive. Maybe a diamond block or something that has to be placed?

NebelNidas: Dirt seemed like a good candidate because the [built-in resourcepack test overrides it as well](https://github.com/FabricMC/fabric/blob/28c34730642b48d9d4de2f1d4c306204520c1599/fabric-resource-loader-v0/src/testmod/resources/resourcepacks/test/assets/minecraft/textures/block/dirt.png), so overriding dirt via an injected resourcepack simultaneously allows us to check if prioritization between injected resources and built-in pack resources works too.
As for gravel, I just chose a random non-dirt vanilla block, since dirt already had some weird behavior with injected resources (see my second screenshot above).

Technici4n (approved):
The code changes are good, thanks a lot.

fabric-resource-loader-v0/src/testmod/resources/assets/minecraft/textures/block/dirt.png
modmuss50: I know the test mods do make a few werid changes to the world, but I think this is a little extreme. Id prefer to know that I need to place a block to test this feature. In the future we can possibly look into adding some automated tests for this. it should be trivial to change this PR to override another vanilla block.

modmuss50 (approved):

PR #3168, opened 2023-07-03
Expand upon the internal client test framework.

modmuss50:
All of this code is internal to Fabric API and is not exposed for mods to use. (Yet?)

This PR creates a very primitive API for other modules to impliment their own automated client tests, I started by adding some tests for the issues fixed in: https://github.com/FabricMC/fabric/pull/3164

Things I want to add:

- A command to run the in world tests without needing to restart the game, should make developing and debugging these tests much easier.
- Golden tests where part of the UI, or a custom exported image are saved into the github repo, thus any change to the image can fail the CI. In this example we could render the block to ensure that the texture changes. 
- Some documentation/guides to help other Fabric API developers make use of this
- Maybe: record a video of the tests running in CI, just because its cool to see it click around the UI like a super human.

See the test screenshots on github actions or the following examples:

injected_resources_default
injected_resources_high_contrast
injected_resources_programmer


fabric-resource-loader-v0/src/testmodClient/java/net/fabricmc/fabric/test/resource/loader/client/ResourceLoaderClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+	@ClientTest
+	public void injectedResources() {
+		submitServer((world, player) ->
+				world.setBlockState(
+						player.getBlockPos().up().offset(player.getHorizontalFacing(), 1),
+						Blocks.DIAMOND_BLOCK.getDefaultState()
+				)
+		);
+		// Takes a little bit for our block to appear :D
+		waitForClientTicks(5);
```
modmuss50: Looks like Github actions is even slower, need to find a good way to know when the chunk has been fully re-rendered, it is also needed after changing the resource packs 🤔

modmuss50:
This PR is old, im going to work on this slowly, one bit at a time. Starting with https://github.com/FabricMC/fabric/pull/4057

PR #3181, opened 2023-07-09
[Lifecycle Events] Start and end task execution

isXander:
New events to hook into before and after `Minecraft#render.runTasks()` is ran. This is where all scheduled tasks are ran, including keyboard and mouse handling.

Use cases for this are unlimited, and guarantees that the event is ran after any scheduled tasks. It also proves a great place to hook to add custom keypresses for KeyBinding. Such as running `KeyBinding.onKeyPress(Key)` manually to press a key automatically.


isXander:
Task execution happens a lot more frequently than the normal client tick.

Technici4n (changes requested):
As I said on Discord, this is not suitable for Fabric API. I'd prefer to see proper input events.

PR #3182, opened 2023-07-09
Fix #3170

### LINKED ISSUE #3170: Incorrect behavior with ClientMessageEvents and CommandApi
Two separate mixins are used for commands in `ClientSendMessageEvents`. The first one implements `ALLOW_COMMAND` and `COMMAND_CANCELED` and injects to `HEAD`, which executes before Client Command Api. The second one implements `MODIFY_COMMAND` and `COMMAND` and injects into the first variable load, which is after Client Command Api. The documentation however states that `MODIFY_COMMAND` and `COMMAND` will work with client commands, which it currently does not, since Client Command Api cancels the method before those invokers are called.  

My specific use case is replacing a command with another before sending it to the server. I also registered my custom command to client commands api so command suggestions will work. But Client Command Api cancels the method and never gets to the `MODIFY_COMMAND` and `COMMAND` events.  

I think it would make sense for `MODIFY_COMMAND` and `COMMAND` moved to inject at `HEAD`. Below is the mixin output.
Mixin output for ClientPlayNetworkHandler#sendChatCommand

kevinthegreat1:
Fix #3170 by moving injection point of `MODIFY_COMMAND` to `HEAD`.  
Add test for `MODIFY_COMMAND` and the client command api.  
Also moves `MODIFY_MESSAGE` to `HEAD` for consistency.

### DIFF
--- a/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
+++ b/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/message/ClientPlayNetworkHandlerMixin.java
@@ -40,7 +40,7 @@ public abstract class ClientPlayNetworkHandlerMixin {
 		}
 	}
 
-	@ModifyVariable(method = "sendChatMessage", at = @At(value = "LOAD", ordinal = 0), ordinal = 0, argsOnly = true)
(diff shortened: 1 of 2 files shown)


Technici4n (commented):
Should we add to the javadoc of `MODIFY_COMMAND` that it can also modify client commands?

Technici4n (approved):
Nevermind, it's already in the javadoc. It turns out that #3170 was introduced by the port to 22w42a!

modmuss50 (approved):

PR #3190, opened 2023-07-14
Fix crash when beehive is broken by fake player

SquidDev:
When a beehive is broken, every nearby bee targets a random player. However, if there are no nearby players, the game crashes.

This should not occur under normal (vanilla) conditions. However, if a beehive is broken by a fake player there are no players in range, and so we see a crash.

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BeehiveBlockMixin.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BeehiveBlockMixin.java
@@ -16,11 +16,7 @@
 
 package net.fabricmc.fabric.mixin.event.interaction;
 
-import net.minecraft.block.BeehiveBlock;
-import net.minecraft.entity.passive.BeeEntity;
-import net.minecraft.entity.player.PlayerEntity;
-import net.minecraft.util.math.BlockPos;
-import net.minecraft.world.World;
+import java.util.List;
 
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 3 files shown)


apple502j (approved):
I actually have (unsuccessfully) attempted to crash the game in vanilla; yeah, this seems good.

Technici4n (approved):

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BeehiveBlockMixin.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.block.BeehiveBlock;
+import net.minecraft.entity.passive.BeeEntity;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@Mixin(BeehiveBlock.class)
+public class BeehiveBlockMixin {
```
Juuxel:
```suggestion
class BeehiveBlockMixin {
```
Should be package-private according to CONTRIBUTING.md

Juuxel (approved):

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/mixin/event/interaction/BeehiveBlockMixin.java
SquidDev: :+1: Someone may want to do a cleanup pass in the rest of this module - I believe the other mixins are also public there.

Juuxel: Yeah, I believe old mixins (and most other code) weren't adjusted when the current contributing guidelines were added.

modmuss50 (approved):
Thanks for adding a test 👍

SquidDev:
Thank you!

PR #3204, opened 2023-07-17
A few transfer API improvements and deprecations

Technici4n:
- Deprecate `exactView` with no replacement. It is just clutter that nobody uses as far as I know.
- Move `simulateInsert` and `simulateExtract` methods to `StorageUtil`. This removes these helper methods from the core `Storage` interface, and places them alongside other similar helpers. As a side effect, this allows `simulateExtract` to work on `StorageView`s which also have an `extract` method.
- Link to wiki tutorial in `Storage` javadoc. We don't usually do that, however this API is too hard to comprehend without an entry-level tutorial. The tutorial is not complete yet, but I will be able to expand it as time passes.
- Clarify what the estimate means in `StorageView#getCapacity`.

Thanks to @agnor99 who suggested these changes!

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
@@ -221,9 +221,7 @@ public interface Storage<T> extends Iterable<StorageView<T>> {
 	 */
 	@Deprecated(forRemoval = true)
 	default long simulateInsert(T resource, long maxAmount, @Nullable TransactionContext transaction) {
-		try (Transaction simulateTransaction = Transaction.openNested(transaction)) {
-			return insert(resource, maxAmount, simulateTransaction);
-		}
+		return StorageUtil.simulateInsert(this, resource, maxAmount, transaction);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/Storage.java
```diff
@@ -247,4 +212,48 @@ default long getVersion() {
...
+
+	/**
+	 * Convenient helper to simulate an insertion, i.e. get the result of insert without modifying any state.
+	 * The passed transaction may be null if a new transaction should be opened for the simulation.
+	 * @see #insert
+	 * @deprecated Either use transactions directly, or use {@link StorageUtil#simulateInsert}.
+	 */
+	@Deprecated(forRemoval = true)
+	default long simulateInsert(T resource, long maxAmount, @Nullable TransactionContext transaction) {
+		try (Transaction simulateTransaction = Transaction.openNested(transaction)) {
```
liach: The implementation should be switched to call the StorageUtil methods directly, so users know what they should call instead even if the javadoc is not available and reduces changes if the simulate implementation needs an update.

liach (approved):

modmuss50 (approved):

PR #3208, opened 2023-07-20
Improve Indigo and FRAPI Test Mod

PepperCode1:
### Improve flat shade

- Use AO mode to make flat shade calculation consistent with shade applied by smooth lighting
- Use face normal to calculate shade if necessary
- Use normal shade even if no custom normals are set

 ### Improve FRAPI test mod

- Add octagonal column to test irregular face lighting
- Improve `PillarBakedModel` to fully support custom block appearance 
- Use obsidian sprite instead of missing sprite for frame mesh
- Simplify and organize registration
- Inline `simple` package


fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/mixin/renderer/client/OverlayVertexConsumerMixin.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraft.util.math.Direction;
+import net.minecraft.util.math.MathConstants;
+
+@Mixin(OverlayVertexConsumer.class)
+public class OverlayVertexConsumerMixin {
+	@Unique
+	private static final Direction[] DIRECTIONS = Direction.values();
+
+	@Redirect(method = "next()V", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/math/Direction;getFacing(FFF)Lnet/minecraft/util/math/Direction;"))
+	private Direction redirectGetFacing(float x, float y, float z) {
```
Technici4n, line 45: Would be nice to add a comment explaining the issue.

Technici4n (changes requested):

Technici4n (approved):

modmuss50 (approved):

PR #3213, opened 2023-07-22
Make DFU error-tolerant to mod custom generator types

qouteall:
The PR fixes an issue of DFU that, when upgrading a world, it fails to deserialize mod custom generator type when upgrading, then the vanilla dimensions may get lost.

[My previous PR about dimension deserialization](https://github.com/FabricMC/fabric/pull/2856) is different. That PR fixes the issue after uninstalling dimension mod/dimension datapack by making the deserialization fail-soft. That deserialization is for reading data in `level.dat`. This PR fixes deserialization in DFU, triggered when upgrading world.

In MC, it firstly upgrade the data then deserialize the data. It also does deserialization and serialization during upgrading. The deserialization here means the deserialization during upgrading.

The fix is to change `Schema2823` and make its `TaggedChoice` to ignore unknown chunk generator types.

When testing it, I found that DFU is sensitive to dimension id's hashcode, that's why I added another dimension `fabric_dimension:void_earilier_in_hash_map` into the test mod. In `NbtCompound` it uses `HashMap` so the order in the map is correlated with hashcode. In `CompoundListCodec.decode`, it deserializes according to the order in the map. And when one entry's deserialization fails, the thing in `result` will be a failed result. In `apply2stable`, if one fails, the failure will propagate the the remaining elements (it's how applicative works). If the result is in failed state, the lambda in `apply2stable` will not execute, so the remaining things will not go to the result even if deserialziation succeededs. The dimension `fabric_dimensions:void` is after all vanilla dimensions in hash map so it will not trigger this issue. So I added the dimension `fabric_dimension:void_earilier_in_hash_map` to make this issue manifest.

DFU is so "GREAT".

This only happens with custom chunk generator types (not happening with dimension datapacks). To test this, I added the dimension `fabric_dimension:void_earilier_in_hash_map` in 1.19.4 branch of fabric api and created a world that contains that dimension in 1.19.4. 

[TestDimUpgrade.zip](https://github.com/FabricMC/fabric/files/12135956/TestDimUpgrade.zip)

Without this PR, opening that world then it will log `Unsupported key: fabric_dimension:void; Unsupported key: fabric_dimension:void` and the nether and end dimensions will vanish.

This makes the DFU to not fail when encountering custom generator type. (The PR also make it not fail when encountering custom biome source type in vanilla chunk generator types). It will possibly be compatible with mod custom datafixers for mod custom chunk generator in `level.dat` (I haven't seen such mod).

About the usage of `@Redirect`: MixinExtra allows WrapOperation to a call that has been redirected, so there is not much need to replace redirect to other injections (I think other injections are more error-prone than redirect in this case).

### DIFF
--- a/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/Schema2832Mixin.java
+++ b/fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/Schema2832Mixin.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.fabricmc.fabric.mixin.dimension;
 
 import java.util.Map;
@@ -16,7 +32,7 @@ import net.minecraft.datafixer.schema.Schema2832;
 import net.fabricmc.fabric.impl.dimension.TaggedChoiceExtension;
 
 @Mixin(value = Schema2832.class, remap = false)
-public class MixinSchema2832 {
+public class Schema2832Mixin {
 	/**
 	 * Make the DSL.taggedChoiceLazy to ignore mod custom generator types and not cause deserialization failure.
 	 */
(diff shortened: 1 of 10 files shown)


fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/Schema2832Mixin.java
```diff
@@ -0,0 +1,57 @@
...
+import com.mojang.datafixers.types.templates.TypeTemplate;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import net.minecraft.datafixer.schema.Schema2832;
+
+import net.fabricmc.fabric.impl.dimension.TaggedChoiceExtension;
+
+@Mixin(value = Schema2832.class, remap = false)
```
liach: The schema class is obfuscated in jar, and it should be remapped.

qouteall: It coincidentally didn't crash with remap disabled. Now fixed.

qouteall:
The core thing is in `TaggedChoiceTypeMixin`. `TaggedChoiceType` contains a map from string to codec. It simply dispatches the codec based on the type field.

All other mixins are for only enabling fail-soft in the specified places. If making all `TaggedChoiceType` fail-soft is appropriate, this PR can be reduced to only one Mixin.

liach (approved):
I think minimizing behavioral impact is more important, so having more mixins is ok.

modmuss50 (approved):

PR #3216, opened 2023-07-23
Fix data generation for custom dynamic registry

### LINKED ISSUE #3215: [FAPI 0.86] RegistryLoaderMixin is overwriting Registry Path
This mixing is changing the path for every, non vanilla Registry. Even if a Registry is not managed  with `DynamicRegistries`:

https://github.com/FabricMC/fabric/blob/e3d1b5f3d89f780b644f6e967963224c2cdef194/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/RegistryLoaderMixin.java#L65-L71

The results in registries not getting loaded (quiqueck/BetterEnd#273)

apple502j:
Fixes #3215

This also adds a test for the dynamic registry provider.

### DIFF
--- a/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestContent.java
+++ b/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestContent.java
@@ -56,7 +56,6 @@ public class DataGeneratorTestContent implements ModInitializer {
 			TEST_DATAGEN_DYNAMIC_REGISTRY_KEY,
 			new Identifier(MOD_ID, "tiny_potato")
 	);
-	public static final TestDatagenObject TEST_DYNAMIC_REGISTRY_ITEM = new TestDatagenObject(":tiny_potato:");
 
 	@Override
 	public void onInitialize() {
--- a/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestContent.java
+++ b/fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestContent.java
@@ -16,6 +16,9 @@
 
 package net.fabricmc.fabric.test.datagen;
 
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+
 import net.minecraft.block.AbstractBlock;
 import net.minecraft.block.Block;
 import net.minecraft.block.Blocks;
@@ -32,6 +35,7 @@ import net.minecraft.text.Text;
 import net.minecraft.util.Identifier;
 
 import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.registry.DynamicRegistries;
 import net.fabricmc.fabric.api.itemgroup.v1.FabricItemGroup;
 import net.fabricmc.fabric.api.itemgroup.v1.ItemGroupEvents;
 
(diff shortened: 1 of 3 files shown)


Technici4n:
Should this only be done for registries registered via the API?

apple502j:
@Technici4n Perhaps. Another solution is to make a new method (with the check you mentioned in place), which providers can override.

fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestContent.java
```diff
@@ -46,6 +50,14 @@ public class DataGeneratorTestContent implements ModInitializer {
 
 	public static final RegistryKey<ItemGroup> SIMPLE_ITEM_GROUP = RegistryKey.of(RegistryKeys.ITEM_GROUP, new Identifier(MOD_ID, "simple"));
 
+	public static final RegistryKey<Registry<TestDatagenObject>> TEST_DATAGEN_DYNAMIC_REGISTRY_KEY =
+			RegistryKey.ofRegistry(new Identifier("fabric", "test_datagen_dynamic"));
+	public static final RegistryKey<TestDatagenObject> TEST_DYNAMIC_REGISTRY_ITEM_KEY = RegistryKey.of(
+			TEST_DATAGEN_DYNAMIC_REGISTRY_KEY,
+			new Identifier(MOD_ID, "tiny_potato")
+	);
+	public static final TestDatagenObject TEST_DYNAMIC_REGISTRY_ITEM = new TestDatagenObject(":tiny_potato:");
```
modmuss50: I dont think this follows the correct pattern (Fabric API should set a good example), this should only be created in the datagenerator's buildRegistry.

fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java
```diff
@@ -114,6 +119,14 @@ public void onInitializeDataGenerator(FabricDataGenerator dataGenerator) {
 		}
 	}
 
+	@Override
+	public void buildRegistry(RegistryBuilder registryBuilder) {
+		registryBuilder.addRegistry(
+				DataGeneratorTestContent.TEST_DATAGEN_DYNAMIC_REGISTRY_KEY,
+				(registerable) -> { }
```
modmuss50: I would create a full method for this and register the test entry to this `Registerable` see TestBiomes.bootstrap()

fabric-data-generation-api-v1/src/testmod/java/net/fabricmc/fabric/test/datagen/DataGeneratorTestEntrypoint.java
```diff
@@ -364,4 +377,24 @@ public void accept(BiConsumer<Identifier, LootTable.Builder> consumer) {
...
+	 * Note that Biome API testmod provides the test for vanilla dynamic registries.
+	 */
+	private static class TestDynamicRegistryProvider extends FabricDynamicRegistryProvider {
+		TestDynamicRegistryProvider(FabricDataOutput output, CompletableFuture<RegistryWrapper.WrapperLookup> registriesFuture) {
+			super(output, registriesFuture);
+		}
+
+		@Override
+		protected void configure(RegistryWrapper.WrapperLookup registries, Entries entries) {
+			entries.add(TEST_DYNAMIC_REGISTRY_ITEM_KEY, TEST_DYNAMIC_REGISTRY_ITEM);
```
modmuss50: `TEST_DYNAMIC_REGISTRY_ITEM` should be retrived via the WrapperLookup and not have a constant.

modmuss50 (changes requested):
The code changes are fine, however I think we should really demonstate the correct pattern for the testmod.

modmuss50 (approved):

modmuss50:
> Should this only be done for registries registered via the API?

Yeah, I think this would be wise, and should match what is done in https://github.com/FabricMC/fabric/pull/3219

I will merge 3219 first, and then update this to use the stuff added in this PR.

PR #3223, opened 2023-07-25
Fixed NBT ingredient json (1.19.2)

LatvianModder:
Fixed number type erasure bug when serializing custom NBT ingredient. This should be fixed in future versions too.

### DIFF
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/builtin/NbtIngredient.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/builtin/NbtIngredient.java
@@ -143,7 +143,7 @@ public class NbtIngredient implements CustomIngredient {
 			json.addProperty("strict", ingredient.strict);
 
 			if (ingredient.nbt != null) {
-				json.add("nbt", ingredient.nbt.toString());
+				json.addProperty("nbt", ingredient.nbt.toString());
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/builtin/NbtIngredient.java
```diff
@@ -145,7 +145,7 @@ public void write(JsonObject json, NbtIngredient ingredient) {
 			json.addProperty("strict", ingredient.strict);
 
 			if (ingredient.nbt != null) {
-				json.add("nbt", NbtOps.INSTANCE.convertTo(JsonOps.INSTANCE, ingredient.nbt));
```
modmuss50, line 148: This wont pass checkstyle, due to the unused import.

modmuss50, line 148: Can a test for this be added?

modmuss50 (changes requested):
Please target the default branch (1.20.1) first, atm we aren't actively updating 1.19.2 but might look into doing a bulk backport.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/builtin/NbtIngredient.java
LatvianModder: Is it needed? The read/parse function in the same class reads the string version of Json already, (`TagParser.parseTag`) so it's backwards compatible. Not sure how can you test that any further

LatvianModder: That I can fix, I did the PR from web UI 😄

modmuss50: I think so, to prove what the problem is and how it was solved. Would fit nicely into SerializationTests.

PR #3237, opened 2023-08-04
23w31a port 

modmuss50:
- Removed fabric-networking-v0
- Major networking refactor

Things are broken, but I think this is at a point where we can release an inital beta to let people provide feedback and find whats broken. I fully expect we may need to make some breaking changes to the the network configuration APIs between now and a stable version, so I have marked those APIs as experimental.

TODO:
- [x] Networking
- [x] Registry sync move to configure networking phase
- [x] ResourcePackProfile.PackFactory.method_52425, maybe look at ZipResourcePack.class_8615 (needs mappings ideally)
- [x] Mouse scroll events
- [ ] Nested dynamic registry test
- [x] Client tag tests


modmuss50:
Things are broken, but I think this is at a point where we can release an inital beta to let people provide feedback and find whats broken. I fully expect we may need to make some breaking changes to the the network configuration APIs between now and a stable version, so I have marked those APIs as experimental.

PR #3246, opened 2023-08-09
Make deprecated modules opt-out instead of opt-in

modmuss50:
This adds back the deprecated modules into dev envs by default, its causing too much confusing. Mod devs can opt out of having these modules like so:

```kotlin
modImplementation("net.fabricmc.fabric-api:fabric-api:0.86.1+local-1.20.1") {
   exclude(module = "fabric-api-deprecated")
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -657,7 +657,7 @@ publishing {
 			pom.withXml {
 				def depsNode = asNode().appendNode("dependencies")
 				subprojects.each {
-					// The maven BOM containing all of the deprecated modules is added manually bellow.
+					// The maven BOM containing all of the deprecated modules is added manually below.
 					if (it.path.startsWith(":deprecated")) {
 						return
 					}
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -657,7 +657,6 @@ publishing {
 			pom.withXml {
 				def depsNode = asNode().appendNode("dependencies")
 				subprojects.each {
-					// Dont depend on the deprecated modules in the main artifact.
```
Juuxel, line 660: Maybe add a note about it being handled separately below?

build.gradle
```diff
@@ -657,7 +657,7 @@ publishing {
...
+					// The maven BOM containing all of the deprecated modules is added manually bellow.
```
Juuxel:
```suggestion
					// The maven BOM containing all of the deprecated modules is added manually below.
```

Juuxel (approved):

PR #3264, opened 2023-08-14
[1.20.2] Remove login networking APIs

modmuss50:
The hope is all mods that use these APIs can move over to using the configuration networking APIs. Please do let me know if you have a usecase for these APIs that cannot be done during the configuration phase. I look forward to discussion about this idea.


apple502j (approved):

liach:
I strongly disagree. This API is useful for communication with proxy servers during authentication, such as velocity; for instance, see https://github.com/OKTW-Network/FabricProxy-Lite/blob/00bcc13ca17093845a77ce5fbf18e646cee28f60/src/main/java/one/oktw/FabricProxyLite.java#L28-L30

apple502j:
Good point. Guess we should maybe put Deprecated instead temporarily? The majority of uses would be replaceable with config phases, but the minority we cannot ignore.

apple502j (changes requested):
Retracting approval.

liach:
A way to manually introduce an incompatibility is to change the signature of the packet handlers, such as using the new CustomPayload or dropping MinecraftServer instance when ServerLoginNetworkHandler is available (ask to use ServerLoginNetworking.getMinecraftServer instead)

apple502j:
Tbh, getMinecraftServer should also be removed. We have TAW now :)

modmuss50:
I think its fine to keep it as-is, no point in breaking it for the point of it. Thanks @liach for pointing out that usecase 👍

PR #3265, opened 2023-08-14
[1.20.2] Misc networking API cleanup + fixes

modmuss50:
Just a handful of cleanup and fixes to the networking API. Please do let me know if you know of anything else that can be improved.


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
```diff
@@ -25,12 +25,6 @@ public final class NetworkingImpl {
 	public static final String MOD_ID = "fabric-networking-api-v1";
 	public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
 
-	/**
-	 * When enabled the fabric packet is written to the {@link net.minecraft.network.PacketByteBuf} on the calling thread.
-	 * This is not enabled by default as it currently causes issues in single player.
-	 */
-	public static final boolean WRITE_FABRIC_PACKET_CALLING_THREAD = Boolean.parseBoolean(System.getProperty("fabric-api.networking.write-fabric-packet-calling-thread", "true"));
-
```
modmuss50, line 33: Turns out this is kinda of a mess to support, and would require having 2 seprate paths for the "legacy" API and the FabricPacket based APIs.

Technici4n: Couldn't we instead force all packets (vanilla too) to be serialized? There is an option for that in the constants class, but javac probably removed some inaccessible code. In principle it should just be a few tweaks to the netty pipeline construction...

SquidDev: > Turns out this is kinda of a mess to support

I'm assuming there's some holes in the current implementation? Aside from some duplication between `createC2SPacket`/`createS2CPacket` (which probably could be reduced), it doesn't seem that bad. I assume I'm missing something?

I realise this is a bit of a weird thing to focus on, but I have had several times where a non-trivial amount of tick time has been spent in packet serialisation (often when the packet also performs compression). I do think moving packet serialisation onto the network thread (where possible) should be something Fabric tries to do.

modmuss50: > I'm assuming there's some holes in the current implementation? Aside from some duplication between createC2SPacket/createS2CPacket (which probably could be reduced), it doesn't seem that bad. I assume I'm missing something?

Yeah, when we last spoke I had hoped this would be all thats needed, and it is when playing on a multiplayer server. (You can try this out by enabling it with system prop in the current snapshot release)

However the issue comes when playing in singleplayer and the payload object is re-used. Currently the object based API is implemented ontop of the PacketByteBuf (PBB) based api, thus it expects all custom packets will be using a `CustomPayloadC2SPacket` and not the `FabricPacketPayload` that I added. You can see where the packet comes into our network addons here: https://github.com/FabricMC/fabric/blob/1.20.1/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerPlayNetworkAddon.java#L78

Without duplicating a lot of this logic and having two seperate paths for incoming packets in single player I cannot see an obvious solution to this. To do it cleanly would be quite a large refactor. Maybe less so if we only did this for the play phase?

>  but I have had several times where a non-trivial amount of tick time has been spent in packet serialisation 

Thats good to know, I was unaware of how much of an issue it was, it may be worth spending more effort trying to resolve this.

If we could get this to work there would be no packet de/serialisation at all in singleplayer.

SquidDev: > I cannot see an obvious solution to this

Yeah, this is made significantly more gnarly by the fact that channels (and thus packet types) are per-channel rather than global.

Ideally we'd just match the vanilla code as much as possible: registering a new receiver  would just add an entry into the custom packet's ID -> deserialiser map (so reusing vanilla's deserialisation code) and handling would just do a `Map<Identifier, XyzNetworking.XyzPacketHandler<?>>` lookup (effectively extending vanilla's existing `if`/`elseif` chain statement).

I guess we could mandate that `PacketType<T>`s are globally unique and registered at mod init, which would mean creating a packet type could also register its deserialiser. I'd argue that this is how mods should be using `PacketType` anyway, but not sure it's worth imposing this constraint for this small gain.

modmuss50:
Going to merge this, the Fabric packet stuff is going to be a much bigger job for another day.

PR #3268, opened 2023-08-16
BlockView API v2

PepperCode1:
This pull requests adds the BlockView API v2 module, which solves the issues with and supersedes the existing Rendering Data Attachment v1 module and adds additional features.

The original goal was to interface inject the two interfaces provided by Rendering Data Attachment v1, `RenderAttachmentBlockEntity` and `RenderAttachedBlockView`, to allow for an easier to use API. It proved to be impossible to achieve this and still provide a clean API without making a new module, as explained below.

The documentation of `RenderAttachedBlockView` states that it is guaranteed to be implemented on all instances of `BlockRenderView`, but this is not the case. The module implementation only guarantees that all `WorldView` instances implement the interface. `RenderAttachedBlockView` extends `BlockRenderView`, meaning that changing the mixin to satisfy the stated contract is not possible; thus, it can be deduced that the mixin is correct but the stated contract is not. However, this does not lead to an easy to use API and makes interface injection of `RenderAttachedBlockView` impossible.

The BlockView API v2 module was made to solve these issues, solve other issues with Rendering Data Attachment v1, and provide additional features.

- The object that used to be inconsistently termed "Render Data Attachment" or "Render Attachment" has been renamed to "Block Entity Render Data" or simply "Render Data" where sufficient. Only the new terms are used in the new module.
- The documentation was rewritten to be more concise, provide more examples, remove mentions of implementation details, and add concrete requirements for use of certain methods.
- The new `FabricBlockView` interface supersedes the `RenderAttachedBlockView` interface.
  - Instead of being implemented on all `BlockRenderView` instances, as the current contract of `RenderAttachedBlockView` states, it is implemented on all `BlockView` instances to provide more flexibility.
  - It contains two new methods, `hasBiomes` and `getBiomeFabric`, to allow biome retrieval. This feature is the only way to create biome-dependent block models, as the `BlockRenderView` provided normally does not have this functionality.
- The new `RenderDataBlockEntity` interface supersedes the `RenderAttachmentBlockEntity` interface.

The Rendering Data Attachment v1 module has been deprecated and its documentation has been updated to state the contract that is actually satisfied and link to the v2 replacements. Perfect backward and forward compatibility would result in degraded performance. To compromise, it was observed that the `BlockRenderView` extension interface is invoked much more frequently than it is implemented, and that the `BlockEntity` extension interface is implemented much more frequently than it is invoked. As such, it was decided that the v1 `RenderAttachedBlockView#getBlockEntityRenderAttachment` should proxy to the v2 `FabricBlockView#getBlockEntityRenderData` and the v2 `RenderDataBlockEntity#getRenderData` should proxy to the v1 `RenderAttachmentBlockEntity#getRenderAttachmentData`.


fabric-block-view-api-v2/src/main/java/net/fabricmc/fabric/api/blockview/v2/FabricBlockView.java
```diff
@@ -0,0 +1,98 @@
...
+	 * {@link #hasBiomes()} returns {@code false}, this method will always return {@code null}.
+	 *
+	 * @implNote Implementations which do not return null are encouraged to use the plains biome as
+	 * the default value, for example when the biome at the given position is unknown.
+	 *
+	 * @param pos the position for which to retrieve the biome
+	 * @return the biome, or null if biome retrieval is not supported
+	 * @see #hasBiomes()
+	 */
+	default RegistryEntry<Biome> getBiomeFabric(BlockPos pos) {
```
Technici4n, line 101: Should we add `@UnspecifiedNullability`?

Technici4n, line 101: I am not sure of the name, but I don't have a better suggestion. Should we also explain the difference with `WorldView#getBiome`? (i.e. that it is supported in more contexts)

Technici4n (commented):
Looks amazing! Left a few comments.

fabric-block-view-api-v2/src/main/java/net/fabricmc/fabric/api/blockview/v2/FabricBlockView.java
PepperCode1: Yes, this has been added in the latest commit. I did not know this annotation existed.

PepperCode1: A section about `WorldView#getBiome` has been added to the javadoc in the latest commit.

fabric-block-view-api-v2/src/main/java/net/fabricmc/fabric/api/blockview/v2/FabricBlockView.java
```diff
@@ -0,0 +1,106 @@
...
+		return false;
+	}
+
+	/**
+	 * Retrieves the biome at the given position if biome retrieval is supported. If
+	 * {@link #hasBiomes()} returns {@code true}, this method will always return a non-null
+	 * {@link RegistryEntry} whose {@link RegistryEntry#value() value} is non-null. If
+	 * {@link #hasBiomes()} returns {@code false}, this method will always return {@code null}.
+	 *
+	 * <p>Do not use this method if the available type of this instance is {@link WorldView};
```
Technici4n: Maybe we can just say that `This method is always supported on {@link WorldView} and gives the same result as {@link WorldView#getBiome}.` and leave out all the rest - it's not perfect but ultimately it is less confusing.

modmuss50 (approved):

Technici4n (approved):

PR #3271, opened 2023-08-18
Improve AttackEntityCallback and ClientPlayerInteractionManagerMixin.

Dev0Louis:
And make it consistent with yarn.
This shouldn't break anything it is just to make the Names better c:


apple502j:
See the introduction to the Development Guidelines:

> Old code or parts thereof might not yet be up to the standards defined by these guidelines. When working with old code, try to adhere to these guidelines, but don't bulk update legacy code to match them. The team will handle updating older code to match the newer standards when appropriate.

- https://github.com/FabricMC/fabric/blob/1.20.1/CONTRIBUTING.md

haykam821:
I think renaming the argument is fine. The event isn't deprecated, and the change only serves to improve the documentation.

PR #3282, opened 2023-08-27
Fix Random mismatch between vanilla and FRAPI baked models (weighted and multipart)

### LINKED ISSUE #3278: Chorus Plant is not rendered correctly, it seems to be rendered with a rotation.
Fabric API:
image
Vanilla:
image

Technici4n:
Fixes #3278.


modmuss50 (approved):
Nice fix, awesome test 👍

PepperCode1 (approved):

PR #3286, opened 2023-08-28
Key binding context

deirn:
Add `KeyBindingContext` to register multiple key binds that won't conflict with others.

This PR doesn't change vanilla behavior in which only one conflicting key (same context and key) will be registered as pressed. 
The order in which what key to functions depends on when the `KeyBinding` instance is created.

~~Do we want `KeyBinding#isPressed` and `wasPressed` to work when a screen is open?~~ Now it also makes it usable on screen.

### DIFF
--- a/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
+++ b/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
@@ -58,7 +58,7 @@ public interface KeyBindingContext {
 	 * Returns whether one context conflicts with the other.
 	 */
 	static boolean conflicts(KeyBindingContext left, KeyBindingContext right) {
-		return left.conflictsWith(right) || right.conflictsWith(left);
+		return left.conflicts(right) || right.conflicts(left);
 	}
 
 	/**
@@ -77,5 +77,5 @@ public interface KeyBindingContext {
 	 * <pre>return this == other || other == IN_GAME;</pre>
 	 */
 	@ApiStatus.OverrideOnly
-	boolean conflictsWith(KeyBindingContext other);
+	boolean conflicts(KeyBindingContext other);
 }
--- a/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
+++ b/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
@@ -25,8 +25,8 @@ import net.fabricmc.fabric.impl.client.keybinding.KeyBindingExtensions;
 
 /**
  * {@link KeyBindingContext} decides how {@link KeyBinding} with same bounded key behaves in regard to each other.
- * <p>
- * Bindings with different context will not conflict with each other even if they have the same bounded key.
+ *
+ * <p>Bindings with different context will not conflict with each other even if they have the same bounded key.
  */
 public interface KeyBindingContext {
 	/**
--- a/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
+++ b/fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
@@ -21,54 +21,61 @@ import org.jetbrains.annotations.ApiStatus;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.option.KeyBinding;
 
+import net.fabricmc.fabric.impl.client.keybinding.KeyBindingContextImpl;
 import net.fabricmc.fabric.impl.client.keybinding.KeyBindingExtensions;
 
 /**
  * {@link KeyBindingContext} decides how {@link KeyBinding} with same bounded key behaves in regard to each other.
  *
  * <p>Bindings with different context will not conflict with each other even if they have the same bounded key.
+ *
+ * <p>Along with the provided generic contexts, mods can also create its own context by implementing this interface.
  */
 public interface KeyBindingContext {
 	/**
 	 * {@link KeyBinding} that used in-game. All vanilla key binds have this context.
 	 */
-	KeyBindingContext IN_GAME = () -> MinecraftClient.getInstance().currentScreen == null;
+	KeyBindingContext IN_GAME = KeyBindingContextImpl.IN_GAME;
 
 	/**
 	 * {@link KeyBinding} that used when a screen is open.
 	 */
-	KeyBindingContext IN_SCREEN = () -> MinecraftClient.getInstance().currentScreen != null;
+	KeyBindingContext IN_SCREEN = KeyBindingContextImpl.IN_SCREEN;
 
 	/**
 	 * {@link KeyBinding} that might be used in any context. This context conflicts with any other context.
 	 */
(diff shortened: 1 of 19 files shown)


fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
```diff
@@ -47,8 +47,20 @@ private KeyBindingHelper() {
 	 * @throws IllegalArgumentException when a key binding with the same ID is already registered
 	 */
 	public static KeyBinding registerKeyBinding(KeyBinding keyBinding) {
+		return registerKeyBinding(keyBinding, KeyBindingContext.IN_GAME);
```
modmuss50, line 50: Should this not be all for existing keybinds? Correct me if im wrong but can exitsing keybinds be used when in a screen?

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/keybinding/KeyBindingEntryMixin.java
```diff
@@ -0,0 +1,36 @@
...
+@Mixin(ControlsListWidget.KeyBindingEntry.class)
+public class KeyBindingEntryMixin {
+	@ModifyConstant(method = "update", constant = @Constant(stringValue = ", "))
+	private String makeConflictTextMultiline(String constant) {
+		return "\n";
+	}
+
+	@ModifyConstant(method = "update", constant = @Constant(stringValue = "controls.keybinds.duplicateKeybinds"))
+	private String replaceConflictText(String constant) {
+		return "fabric.keybinding.conflicts";
```
modmuss50: We shouldnt always be replacing this, it will make this text no longer translated for anyone using fabric API even if there are no mods using this new API.

fabric-key-binding-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/client/keybinding/CustomKeyBindingContext.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.client.MinecraftClient;
+
+import net.fabricmc.fabric.api.client.keybinding.v1.KeyBindingContext;
+
+public class CustomKeyBindingContext implements KeyBindingContext {
+	@Override
+	public boolean isActive() {
+		return MinecraftClient.getInstance().currentScreen == null;
+	}
+}
```
modmuss50: ~~Is this not the same as IN_GAME? Might be better to have a test that does something slightly different.~~ Just this is is `isActive` and not `conflictsWith`. I think the test mod needs expanding to cover more cases.

Maybe also look into automated testing, but that likely requires me to finish: <https://github.com/FabricMC/fabric/pull/3168> Unit tests might be possible.

fabric-key-binding-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/client/keybinding/KeyBindingsTest.java
```diff
@@ -51,6 +66,30 @@ public void onInitializeClient() {
 			while (duplicateBinding.wasPressed()) {
 				client.player.sendMessage(Text.literal("Duplicate Key was pressed!"), false);
 			}
+
+			while (inGameBinding.wasPressed()) {
```
modmuss50: A small comment saying how to test this might be helpful. Just list out a few situations and the expected outcome.

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	boolean isActive();
+
+	/**
+	 * Use {@link #conflicts(KeyBindingContext, KeyBindingContext)} for checking if two context conflicts.
+	 */
+	@ApiStatus.OverrideOnly
+	default boolean conflictsWith(KeyBindingContext other) {
+		return this == other;
```
modmuss50: This doesnt feel right for some reason. If I understand this correctly both keybinds must not conflict with each other. So evem if you did return false to have a keybind that could run everywhere it still wouldnt work.

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingContext.java
```diff
@@ -0,0 +1,74 @@
...
+
+	static KeyBindingContext of(KeyBinding binding) {
+		return ((KeyBindingExtensions) binding).fabric_getContext();
+	}
+
+	static boolean conflicts(KeyBindingContext left, KeyBindingContext right) {
+		return left.conflictsWith(right) || right.conflictsWith(left);
+	}
+
+	boolean isActive();
```
modmuss50: Can you document this?

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
```diff
@@ -47,8 +47,20 @@ private KeyBindingHelper() {
...
+	/**
+	 * Registers the keybinding and add the keybinding category if required.
+	 *
+	 * @param keyBinding the keybinding
+	 * @param context    the keybinding context
+	 * @return the keybinding itself
+	 * @throws IllegalArgumentException when a key binding with the same ID is already registered
+	 */
+	public static KeyBinding registerKeyBinding(KeyBinding keyBinding, KeyBindingContext context) {
 		Objects.requireNonNull(keyBinding, "key binding cannot be null");
```
modmuss50, line 62: Also null check the context

modmuss50 (commented):
Im not fully convinced by the API design here, it feels a little brittle. I dont really have a good idea for something better atm.

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/api/client/keybinding/v1/KeyBindingHelper.java
deirn: > Should this not be all for existing keybinds? 

Yes, that is the default context for all keybinds.

> Can exitsing keybinds be used when in a screen?

No, you need to check the bounded key if it's pressed manually. (eg. via `InputUtil.isKeyPressed` or on screen's key methods)

fabric-key-binding-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/client/keybinding/CustomKeyBindingContext.java
deirn: This is to check whether 2 keys with different context _instance_ work simultaneously.

fabric-key-binding-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/keybinding/KeyBindingEntryMixin.java
deirn: Done, it's now only changed if the conflicting key has a different context.

modmuss50 (approved):
API looks good to me, I think it makes more sense than when I last looked at it. The mixin looks a little friagile, but i dont think there is a lot that can be done about that.

PR #3287, opened 2023-08-28
FRAPI improvements: context getters, full removal of fallback consumers, small enhancements

Technici4n:
### Add cull check and item transformation mode getter to FRAPI
New methods in `RenderContext`:
- `isFaceCulled` to perform a cull check that can be optimized and cached by the renderer.
- `hasTransform`, self-explanatory. Must be checked before using `isFaceCulled` to skip some geometry.
- `itemTransformationMode` to get the current item transformation mode in item contexts.

### Terminally deprecate `fallbackConsumer` and `bakedModelConsumer`
The primary goal of this change is to make renderers not need to implement these fallback consumers themselves. However, their design relies on hidden context state that cannot be queried via the API. It is thus impossible to implement a fallback layer for them, and they must be **deprecated, with plans for removal in ~~1.20.2~~ (EDIT: too late for that)** in a follow-up PR.

Note that `BakedQuadFallback.emitBlockQuads` uses the newly introduced `isFaceCulled` and `hasTransform` methods to shortcut some geometric analysis.

I also removed the "vanilla quad handler" from Indigo.

### Fix uvs in octagonal column test mod
Fix the uvs in the octagonal column test mod by:
- setting the cull face and thus the nominal face to leverage `spriteBake` for the axis-aligned quads;
- manually setting the UVs for the corner quads.


fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/render/RenderContext.java
```diff
@@ -76,6 +88,34 @@ public interface RenderContext {
 	 */
 	void popTransform();
 
+	/**
+	 * Returns {@code true} if the given face will be culled away.
+	 *
+	 * <p>This function can be used to avoid complex transformations of quads that will be culled anyway.
```
Technici4n: I should probably mention that the user should check `hasTransform` too.

PepperCode1 (approved):

PR #3292, opened 2023-08-30
23w35a Port

modmuss50:
Not complete yet, lacking any testing at all.


fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -114,11 +133,10 @@ public void write(PacketByteBuf buf) {
 	}
 
 	@Override
-	public JsonElement toJson() {
-		JsonObject json = new JsonObject();
-		json.addProperty(TYPE_KEY, customIngredient.getSerializer().getIdentifier().toString());
-		customIngredient.getSerializer().write(json, coerceIngredient());
-		return json;
+	public JsonElement toJson(boolean bl) {
+		// TODO do we need to care about bl? I think we do.
```
modmuss50: Note: I think this boolean has something to do with allowing empty stacks. Ingridents have 2 codecs, I think we need to also have 2?

Shnupbups: Correct, the boolean decides whether `minecraft:air` is allowed

ErrorCraft: They changed it to allow an empty array instead, rather than `minecraft:air` in [1.20 pre-release 6](https://www.minecraft.net/en-us/article/minecraft-1-20-pre-release-5). `minecraft:air` is no longer allowed.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -150,4 +137,33 @@ public boolean isEmpty() {
 	private <T> T coerceIngredient() {
 		return (T) customIngredient;
 	}
+
+	public static <T> Codec<T> first(Codec<T> first, Codec<T> second) {
+		return new First<>(first, second);
+	}
+
+	// Decode/encode the first codec, if that fails return the result of the second.
```
maityyy, line 145: Why not `ExtraCodecs::withAlternative` (mojmap)?

modmuss50: I struggled to get this to work, the encode is slightly different I think. We can always improve this later, I want to get this out now 👍

Alexander01998:
Hey @modmuss50, it seems that this version was not published on CurseForge. It only shows up on GitHub and Modrinth.
Not sure if it's a glitch or something, but thought you should know.
Thanks for all the work you put into these ports!

modmuss50:
> Hey @modmuss50, it seems that this version was not published on CurseForge. It only shows up on GitHub and Modrinth.
> Not sure if it's a glitch or something, but thought you should know.
> Thanks for all the work you put into these ports!

Just checked, it has been put for manual review, not seen this before as usually it get's automatically approved. Hopefully that won't take too long, for now use GitHub or Modrith. 

Thanks 👍

PR #3303, opened 2023-09-07
Include namespace in modded registry tag path

apple502j:
**Breaking change** for 1.20.2.

In #3163, modded **dynamic** registries had the registry namespace added to the tag paths. This change now also applies to static registries. This breaks data packs, including bundled ones, if they use tags for custom registries. **This does not affect vanilla tags** like block tags, item tags, or entity type tags.

Javadoc changes included, to document this behavior.


Technici4n (approved):

modmuss50 (approved):
Looks good, I dont think this will affect many mods at all, but its still a break change that should only go into 1.20.2+

PR #3305, opened 2023-09-08
Pass handler to clientside ConfigurationPacketHandler

apple502j:
Breaking change, needs to be merged ASAP.

Turns out you cannot query CCNH from MinecraftClient. We first need to get ClientConnection (which is also PITA - the impl already does that elsewhere), then get the handler through Netty internals and several casting.

Also adds a helper method to disconnect, as there is currently no way to do so. I wasn't sure if interface injection or static method was better; you can revert the commit if this needs to be discussed later.


Technici4n:
Question: why is the handler useful?

apple502j:
Closing, on second thought this doesn't look like it's needed in many use cases.

PR #3311, opened 2023-09-11
Support trade rebalance experiment

apple502j:
Continuation from #3306, with much cleaner design and code.


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/trade/TradeOfferHelper.java
```diff
@@ -66,4 +113,111 @@ public static void refreshOffers() {
 
 	private TradeOfferHelper() {
 	}
+
+	@FunctionalInterface
+	public interface VillagerOffersAdder {
```
Technici4n, line 119: The boolean parameter is a no-go for me.

liach: Agreed. Mojang can add multiple offer rebalance packs and mods can add multiple rebalance options; it's better for them to query the data pack status instead, so maybe pass an object that can provide such data pack/experiment enabled information.

apple502j: Mojang offering 2 trade rebalance packs is just a speculation - we should not over-complicate the API; and since this API does not allow mods to modify existing trades, I doubt there would be many mods that does that to warrant expansion of this API.

liach: Then Mojang can remove the trade rebalance pack, by integrating into mainline or dropping it, at any time. What would this boolean serve by then?

What I am looking for is another parameter which mods can query if ANY experimental pack, from mojang or mods, are added. Say a mod want to add a villager trade for bundle if bundle pack is enabled; then this parameter should be some sort of view/manager that allows the mod to query this information and thus decide.

apple502j: @liach The boolean-taking version would be deprecated and replaced with the existing, non-taking version. However, a deprecation is expected for any design we could take (apart from just outright not supporting the experiment).

The reason such parameter isn't added is simple: it is not possible. This method is called statically on game load before a world is even loaded, and calling the method `pow(2, experiments.length)` times (and storing the result, then applying on world load) is unreasonable. Requiring that this be called during world load is inconsistent with the existing API, and there is no convincing reason to do this.

I also question the "modded experiment" part of the comment. Fabric API does not provide an API to register an experiment yet, so there is no reason to support modded experiments.

Finally, a mod that wants to conditionally add bundles in the offer do not need a special parameter. All you need is a custom `TradeOffer.Factory`, which creates an offer only if the experiment is added. (This method can return `null` since this update.)

liach: Looked at the code, this is indeed some bad design from mojang. I think a better solution would be to patch `MerchantEntity.fillRecipes` implementations to fire 2 events where mods can modify the resulting TradeOffers.Factory list; you can pass the profession, level, and the server world in the regular villager event (and just the world in the wandering trader event), which should be sufficient for mods to make decisions mostly.

apple502j: Again, I did not want to make it too different from existing one; I also don't know why this is needed apart from some hypothetical scenarios.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
```diff
@@ -63,4 +92,53 @@ public static void printRefreshOffersWarning() {
...
+			ID_TO_INDEX.put(id, TradeOffers.REBALANCED_WANDERING_TRADER_TRADES.size());
+			TradeOffers.REBALANCED_WANDERING_TRADER_TRADES.add(pool);
+			return this;
+		}
+
+		@Override
+		public TradeOfferHelper.WanderingTraderOffersBuilder addOffersToPool(Identifier pool, TradeOffers.Factory... factories) {
+			if (!ID_TO_INDEX.containsKey(pool)) {
+				throw new IllegalArgumentException("pool %s is not registered".formatted(pool));
+			}
```
deirn: How does adding offers to modded pools work? From what I can tell this depends on mod order.
Should probably do the additions later (also a supplier maybe?).

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
```diff
@@ -63,4 +92,53 @@ public static void printRefreshOffersWarning() {
...
+			Pair<TradeOffers.Factory[], Integer> poolPair = TradeOffers.REBALANCED_WANDERING_TRADER_TRADES.get(poolIndex);
+			TradeOffers.Factory[] modified = ArrayUtils.addAll(poolPair.getLeft(), factories);
+			TradeOffers.REBALANCED_WANDERING_TRADER_TRADES.set(poolIndex, Pair.of(modified, poolPair.getRight()));
+			return this;
+		}
+
+		@Override
+		public Set<Identifier> getPoolIds() {
+			return Collections.unmodifiableSet(ID_TO_INDEX.keySet());
+		}
```
deirn: This depends on mod order, I don't think it's that useful.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
```diff
@@ -63,4 +92,53 @@ public static void printRefreshOffersWarning() {
...
+				TradeOffers.REBALANCED_WANDERING_TRADER_TRADES = new ArrayList<>(TradeOffers.REBALANCED_WANDERING_TRADER_TRADES);
+			}
+		}
+
+		@Override
+		public TradeOfferHelper.WanderingTraderOffersBuilder pool(Identifier id, int count, TradeOffers.Factory... factories) {
+			if (factories.length == 0) throw new IllegalArgumentException("cannot add empty pool");
+			if (count <= 0) throw new IllegalArgumentException("count must be positive");
+
+			Objects.requireNonNull(id, "id cannot be null");
```
deirn, line 117: Should also throw if the pool already exists.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
apple502j: Yeah, I suppose it's better to handle it ourselves.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/TradeOfferInternals.java
apple502j: Yeah, late evaluation seems like a better idea.

deirn (approved):

modmuss50 (approved):

PR #3324, opened 2023-09-18
Fix unhandled packets disconnecting the client.

modmuss50:


fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientConfigurationNetworkAddon.java
```diff
@@ -86,11 +86,7 @@ protected void receiveRegistration(boolean register, PacketByteBuf buf) {
 	 * @return true if the packet has been handled
 	 */
 	public boolean handle(PacketByteBufPayload payload) {
-		try {
-			return this.handle(payload.id(), payload.data());
-		} finally {
-			payload.data().release();
```
modmuss50, line 92: I think this is now correct, it now matches the server network addons.

apple502j (approved):

liach:
Yep, this indeed is an unintended side effect since mojang's networking update, where custom payloads have dedicated buf than the buf straight from netty :)

liach (approved):

SyedOmarNoor:
is this related to this? https://github.com/orgs/FabricMC/discussions/3391

PR #3328, opened 2023-09-20
Handle null being passed as the datafix type to PersistentState.Type

### LINKED ISSUE #3327: Persistent state types should support null data fixer types
As of Minecraft 1.20.2, persistent state types must be constructed with a data fixer type. If the data fixer type is `null`, then the following exception will be logged when loading the persistent state:

```
java.lang.NullPointerException: Cannot invoke "net.minecraft.datafixer.DataFixTypes.update(com.mojang.datafixers.DataFixer, net.minecraft.nbt.NbtCompound, int, int)" because "dataFixTypes" is null
```

modmuss50:
Closes #3327

The discoverbility is low, however I think most people will first try null and then realise it crashes. At the moment this is by far the easyist fix.

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/PersistentStateManagerMixin.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/PersistentStateManagerMixin.java
@@ -31,7 +31,7 @@ import net.minecraft.nbt.NbtCompound;
 import net.minecraft.world.PersistentStateManager;
 
 @Mixin(PersistentStateManager.class)
-public class PersistentStateManagerMixin {
+class PersistentStateManagerMixin {
 	/**
 	 * Handle mods passing a null DataFixTypes to a PersistentState.Type.
 	 */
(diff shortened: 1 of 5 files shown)


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/PersistentStateManagerMixin.java
```diff
@@ -0,0 +1,44 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import net.minecraft.datafixer.DataFixTypes;
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.world.PersistentStateManager;
+
+@Mixin(PersistentStateManager.class)
+public class PersistentStateManagerMixin {
```
Juuxel:
```suggestion
class PersistentStateManagerMixin {
```
(see conventions)

Juuxel (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/PersistentStateManagerMixin.java
modmuss50: Probably another thing we should validate 👍

haykam821 (approved):

PR #3341, opened 2023-09-26
Update to loom 1.4

modmuss50:
When merging all versions must be bumped.


modmuss50:
Now building with Java 21 from Microsoft. (Still targets J17 of course)

PR #3343, opened 2023-09-27
Document the getCodec method

apple502j:
Actual documentation on how to use codecs would be veeeery long for inclusion, therefore it is omitted.

### DIFF
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/CustomIngredientSerializer.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/CustomIngredientSerializer.java
@@ -20,6 +20,7 @@ import com.mojang.serialization.Codec;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.network.PacketByteBuf;
+import net.minecraft.recipe.Ingredient;
 import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.impl.recipe.ingredient.CustomIngredientImpl;
@@ -54,6 +55,15 @@ public interface CustomIngredientSerializer<T extends CustomIngredient> {
 	 */
 	Identifier getIdentifier();
 
+	/**
+	 * {@return the codec}.
+	 *
+	 * <p>Codecs are used to read the ingredient from the recipe JSON files.
+	 * If the custom ingredient
+	 *
+	 * @see Ingredient#ALLOW_EMPTY_CODEC
+	 * @see Ingredient#DISALLOW_EMPTY_CODEC
+	 */
 	Codec<T> getCodec(boolean allowEmpty);
 
 	/**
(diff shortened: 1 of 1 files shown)


fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/CustomIngredientSerializer.java
```diff
@@ -54,6 +55,15 @@ static CustomIngredientSerializer<?> get(Identifier identifier) {
 	 */
 	Identifier getIdentifier();
 
+	/**
+	 * {@return the codec}.
+	 *
+	 * <p>Codecs are used to read the ingredient from the recipe JSON files.
+	 * If the custom ingredient
```
haykam821: Is this sentence cut off?

apple502j: Intellij didn't save it properly hmm. Just gonna remove this
```suggestion
```

modmuss50 (approved):

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/CustomIngredientSerializer.java
```diff
@@ -54,6 +55,14 @@ static CustomIngredientSerializer<?> get(Identifier identifier) {
 	 */
 	Identifier getIdentifier();
 
+	/**
+	 * {@return the codec}.
```
haykam821, line 59: A small stylistic nitpick: should we enforce leaving out the period at the end of `{@return}`? Yarn primarily doesn't use punctuation with only 13 exceptions.

```suggestion
	 * {@return the codec}
```

apple502j: [Unfortunately, no.](https://github.com/checkstyle/checkstyle/issues/12600)

haykam821: In that case, should we change the Yarn convention to enforce the period? In any case, the style should be consistent disregarding any linter bugs, even if the exceptions aren't relevant to this pull request or repository specifically.

Juuxel: I'd say Yarn's convention is the real one, and the one used in FAPI is just a workaround for a bug. The output is the same anyway iirc.

apple502j: No, the output in FAPI afaik produces double periods. Better than failing build.

Juuxel:
```java
public class Test {
    /** {@return a number}. */
    public int x() { return 0; }
}
```



Just tested (with JDK 21), I can confirm that. (I thought I confirmed the opposite once?) So in this case, we should absolutely not change Yarn's conventions.

Issue #3316, opened 2023-09-14
[Suggestion] Data-Driven Biome Modification API


maityyy:
The Biome Modification API is most often used for simple operations like add/remove/replace a feature/mob/etc., which, IMO, can be implemented as simple JSON files in data packs. Datagen can also be used, since most worldgen is implemented through data packs and datagen. In fact, NeoForge already has [similar functionality](https://forge.gemwire.uk/wiki/Biome_Modifiers) and we can turn to its experience.

Idk if it is worth redoing the entire API or just adding the ability to create simple modifications via JSON files. I personally would not want to force the use of a datagen, since there are many simple mods that want to remain that way, although they can write a few files by hand + the current API may be difficult to expand. In any case, BMAPI is marked as an experimental API, so we have a lot of freedom and can indirectly opt out of some features.

I also think this is a great opportunity to unify the JSON file format with the NeoForge API

modmuss50:
A data driven system could possibly be built ontop of the existing API to provide the common operations. I dont think we should be making breaking changes or removing the existing API. Even its made stable that does not prevent v2 module, or new additions.

PR #3348, opened 2023-09-29
Biome API: javadoc fixes

apple502j:


fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModification.java
```diff
@@ -27,7 +27,10 @@
 import net.fabricmc.fabric.impl.biome.modification.BiomeModificationImpl;
 
 /**
- * <b>Experimental feature</b>, may be removed or changed without further notice.
+ * Provides methods for modifying biomes. To create an instance, call
+ * {@link BiomeModifications#create(Identifier)}.
+ *
+ * <p><b>Experimental feature</b>, may be removed or changed without further notice.
```
modmuss50, line 33: Not for this PR: We should really make this stable, I think ill take a look at the modules and figure out what can become stable.

modmuss50 (approved):

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModification.java
maityyy: I do not think so because Forge's API looks much better and we can do something similar (at least that is what I am planning to do when I have free time). See #3316

PR #3352, opened 2023-09-30
Add LootTableEvents.LOADED event 

LLytho:
Hey,

The current loot table mixin <https://github.com/FabricMC/fabric/blob/1.20.1/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/mixin/loot/LootManagerMixin.java#L55> updates the return value which will automatically blocks all modded mixins with the same injection but higher priority. 

So I added an event which can be used to track down when all loot tables are loaded + modified by fabric events.
My first idea was to move the mixin to `apply` again like in previous versions but the current mixin requires the `ResourceManager` which does not exist in `apply` and I wasn't sure if collecting the `ResourceManager` first, storing it in the class and passing it later on is a preferable way for fabric.

### DIFF
--- a/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
+++ b/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
@@ -92,9 +92,9 @@ public final class LootTableEvents {
 	/**
 	 * This event can be used for post-processing after all loot tables have been loaded and modified by fabric.
 	 */
-	public static final Event<Loaded> LOADED = EventFactory.createArrayBacked(Loaded.class, listeners -> (resourceManager, lootManager) -> {
+	public static final Event<Loaded> ALL_LOADED = EventFactory.createArrayBacked(Loaded.class, listeners -> (resourceManager, lootManager) -> {
 		for (Loaded listener : listeners) {
-			listener.onLootTableLoaded(resourceManager, lootManager);
+			listener.onLootTablesLoaded(resourceManager, lootManager);
 		}
 	});
 
@@ -133,6 +133,6 @@ public final class LootTableEvents {
 		 * @param resourceManager the server resource manager
 		 * @param lootManager     the loot manager
 		 */
-		void onLootTableLoaded(ResourceManager resourceManager, LootManager lootManager);
+		void onLootTablesLoaded(ResourceManager resourceManager, LootManager lootManager);
 	}
 }
--- a/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
+++ b/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
@@ -89,6 +89,15 @@ public final class LootTableEvents {
 		}
 	});
 
+	/**
+	 * This event can be used for post-processing after all loot tables have been loaded and modified by fabric.
+	 */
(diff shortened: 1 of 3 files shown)


apple502j (approved):

modmuss50:
General question, whats the use case for this event? I see the test mod is only checking that loot table is not empty, I imagine you plan to do more with it?

apple502j:
@modmuss50 [Very likely yes](https://github.com/AlmostReliable/lootjs).

LLytho:
> General question, whats the use case for this event? I see the test mod is only checking that loot table is not empty, I imagine you plan to do more with it?

Yes. I want to add loot table modification to LootJS and I want to trigger the LootJS event when all loot tables are loaded and modified by mods. So packdevs can further modify them by their needs.

fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
```diff
@@ -116,4 +125,14 @@ public interface Modify {
...
 	}
+
+	public interface Loaded {
+		/**
+		 * Called when all loot tables have been loaded and {@link LootTableEvents#REPLACE} and {@link LootTableEvents#MODIFY} have been invoked.
+		 *
+		 * @param resourceManager the server resource manager
+		 * @param lootManager     the loot manager
+		 */
+		void onLootTableLoaded(ResourceManager resourceManager, LootManager lootManager);
```
Technici4n:
```suggestion
		void onLootTablesLoaded(ResourceManager resourceManager, LootManager lootManager);
```
technically there's multiple of them? :P

maityyy: Maybe then this event should be called `ALL_LOADED`? Previous events are called per loot table, so the new event may be confusing.

Also I think this event should be after the `MODIFY`/`REPLACE` events in the code, but that does not matter, it is about checkstyle.

Btw, this event is very similar to [CommonLifecycleEvents.TAGS_LOADED](https://github.com/FabricMC/fabric/blob/670b2ea08de60d952f62d5333f93bfe18c27f8ed/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/CommonLifecycleEvents.java#L31)

LLytho: > Also I think this event should be after the MODIFY/REPLACE events in the code, but that does not matter, it is about checkstyle.

wdym by that?

maityyy: > wdym by that?

Ah, sorry, nevermind. I am inattentive

Technici4n (approved):

modmuss50 (approved):

fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
```diff
@@ -89,6 +89,15 @@ private LootTableEvents() {
 		}
 	});
 
+	/**
+	 * This event can be used for post-processing after all loot tables have been loaded and modified by fabric.
```
Juuxel:
```suggestion
	 * This event can be used for post-processing after all loot tables have been loaded and modified by Fabric.
```

Juuxel (approved):

PR #3363, opened 2023-10-05
[1.20.3] Move clientPickBlock APIs to client source set.

modmuss50:
Wont break the API for mods that are correctly using these events in the client sourceset (or merged mods).

Seems this was just missed when the inital split was made.


fabric-events-interaction-v0/src/main/resources/fabric.mod.json
```diff
@@ -25,7 +25,7 @@
       "net.fabricmc.fabric.impl.event.interaction.InteractionEventsRouter"
     ],
     "client": [
-      "net.fabricmc.fabric.impl.event.interaction.InteractionEventsRouterClient"
+      "net.fabricmc.fabric.test.client.event.interaction.InteractionEventsRouterClient"
```
Juuxel: Surely it's not supposed to be in the test mod?

modmuss50: Opps, lol.

PR #3369, opened 2023-10-07
Fix inconsistency of placed feature locations

### LINKED ISSUE #3366: Instaling fabric API slightly changes the world generation
Feature (grass, trees etc.) generation changes when i install fabric API

video [link](https://drive.google.com/file/d/1Y5XLmh8ugr6wU_buA6sCdnee2PpOMB73/view?usp=sharing)

minecraft version: 1.20.1
fabric API version: 0.89.3 (latest for 06.10.23)
seed: 8182091268832203360
teleportation command: _/tp @s -1472.30 64.00 27999.85 12.04 -12.60_
clipboard hotkey: win + v

i spent an entire evening disabling my mods in modrinth app to find out which of 100+ of them causes this bug

pretty sure FAPI for 1.20.2 has this bug too

this is not MC-55596 because it happens only when installing FAPI
there are several (probably ~20) variants of feature placement but none of them is equal to vanilla

apple502j:
`BiomeSource#getBiomes` mixin applies to all biome sources, including one for Overworld. The return value is a set; however one caller in the worldgen code iterates over it: `PlacedFeatureIndexer`. Using a hash set here randomizes the return value, affecting feature placement. Use a linked hash set instead.

Requesting backport for all supported versions.

Fixes #3366

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/BiomeSourceMixin.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/BiomeSourceMixin.java
@@ -17,7 +17,7 @@
 package net.fabricmc.fabric.mixin.biome;
 
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/mixin/biome/BiomeSourceMixin.java
```diff
@@ -33,7 +33,7 @@
 public class BiomeSourceMixin {
 	@Redirect(method = "getBiomes", at = @At(value = "INVOKE", target = "Ljava/util/function/Supplier;get()Ljava/lang/Object;"))
 	private Object getBiomes(Supplier<Set<RegistryEntry<Biome>>> instance) {
-		var biomes = new HashSet<>(instance.get());
+		var biomes = new LinkedHashSet<>(instance.get());
 		fabric_modifyBiomeSet(biomes);
 		return Collections.unmodifiableSet(biomes);
```
modmuss50: As I said on discord, I think we should change this to just retrun the original set when no modifications have taken place. In TheEndBiomeSourceMixin you would end up with something along the lines of:

Screenshot 2023-10-07 at 14 11 05

phuchau1989 (approved):

PR #3373, opened 2023-10-08
Fix FabricBlockSettings copyOf missing settings

Shnupbups:
Fixes FabricBlockSettings' copyOf method missing several new settings introduced as a result of the removal of Materials.

Specifically `burnable`, `liquid`, `forceNotSolid`, `forceSolid`, `pistonBehavior`, `instrument`, and `replaceable`. These are all copied by the vanilla `BlockSettings.copy` as well.
The `emissiveLightingPredicate` is also now copied in vanilla, so I moved the copy of that to reflect that change.

This should probably also be backported to 1.20.1, and frontported to 1.20.3 (at least for now, until it gets fully json'd away)


AlphaMode:
Shouldn't FabricBlockSettings also just be deprecated in favor of interface injection?

modmuss50:
> Shouldn't FabricBlockSettings also just be deprecated in favor of interface injection?

Probably, I would suggest waiting to see what (if anything) 1.20.3 does to this, its possible it might not exist in its current form for much longer.

modmuss50 (approved):

gniftygnome:
If you want a real-world example of something broken because of the missing copies, Every Compat (WoodGood) apparently uses the instrument as one of the checks for whether something is a wood block, which has broken their support for some Fabric mods using `FabricBlockSettings.copyOf`.  It would be nice to know sooner than 1.20.3 whether this will get fixed or we should switch to `Block.Settings.copy` or `AbstractBlock.Settings.copy`.

PR #3374, opened 2023-10-08
Fix 23w40a port issues

apple502j:
One code bug (possible memory leak when using reconfiguration), one javadoc issue, and one testmod issue.


modmuss50 (approved):
Will merge with the snapshot tomorrow. Thanks 👍

PR #3377, opened 2023-10-13
Properly detach network addon instance on config <-> play transition

deirn:
When transitioning between config and play stage `GlobalReceiverRegistry#endSession` is not called, creating a memory leak.
Also removed some code repetition.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
@@ -26,11 +26,15 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import net.minecraft.network.NetworkState;
 import net.minecraft.util.Identifier;
 
 public final class GlobalReceiverRegistry<H> {
+	private static final Logger LOGGER = LoggerFactory.getLogger(GlobalReceiverRegistry.class);
+
 	private final NetworkState state;
 
 	private final ReadWriteLock lock = new ReentrantReadWriteLock();
@@ -135,6 +139,7 @@ public final class GlobalReceiverRegistry<H> {
 
 		try {
 			this.trackedAddons.add(addon);
+			this.logTrackedAddonSize();
 		} finally {
 			lock.unlock();
 		}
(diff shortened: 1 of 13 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
```diff
@@ -145,17 +150,29 @@ public void endSession(AbstractNetworkAddon<H> addon) {
...
 			lock.unlock();
 		}
 	}
 
+	/**
+	 * In practice, trackedAddons should never contain more than one instance.
+	 */
+	private void logTrackedAddonSize() {
+		if (LOGGER.isDebugEnabled() && this.trackedAddons.size() > 1) {
+			LOGGER.error("{} receiver registry tracks {} addon instances where it should only tracks one!", state.getId(), trackedAddons.size());
```
apple502j:
```suggestion
			LOGGER.error("{} receiver registry tracks {} addon instances where it should only track one!", state.getId(), trackedAddons.size());
```

liach:
Can we inject `endSession` at `NetworkStateTransitionHandler#handle` instead of at individual packet handlers?

modmuss50 (approved):

modmuss50:
> Can we inject `endSession` at `NetworkStateTransitionHandler#handle` instead of at individual packet handlers?

Maybe, but I think the changes in this PR make it a bit more clear when its going to happen.

liach:
How does injecting at each packet more clear? These packets' handlers themselves don't transition; all vanilla transitions are done in NSTH.handle or disconnection.

If the addon registration and unregistration model is too bug-prone, we might move the global receiver propagation to find addons every time the registrations change. We have the client->server connection on MinecraftClient and the server->client connections in one of the server classes, and we can thus derive the addons we need.

deirn:
I actually have a POC for handling the session automatically in https://github.com/deirn/fabric/commit/0596ed73a3c495771de67a0e9cdc75dd97f243db, it ends the session in `ClientConnection#setPacketListener`. Seems to be working on my limited test, though I became too busy to continue it.

modmuss50:
> I actually have a POC for handling the session automatically in [deirn@0596ed7](https://github.com/deirn/fabric/commit/0596ed73a3c495771de67a0e9cdc75dd97f243db), it ends the session in `ClientConnection#setPacketListener`. Seems to be working on my limited test, though I became too busy to continue it.

Thanks, I will take a proper look into this another day to figure out whats best. 👍

modmuss50:
Merged as part of https://github.com/FabricMC/fabric/pull/3394 Many Thanks :)

PR #3383, opened 2023-10-19
Add a registration system for spreadable block compatibility.

gniftygnome:
This PR adds a registration system for spreadable blocks (blocks which extend SpreadableBlock).  Each "type" of spreadable block has its own registry instance, which maps a bare block (f.e. Dirt) to a spreadable block (f.e. Grass or Mycelium).  Two instances are provided for mods extending vanilla grass and mycelium.  Additional registries can be requested for new modded spreadable types.  I believe this is the simplest implementation appropriate for FAPI which implements enough features to be useful and is minimally intrusive.  For example, this provides enough functionality I expect to be able to use it in Terraform API.

Goals:
* Vanilla grass can spread to modded dirt and create the correct modded grass.
* Modded grass can spread to vanilla dirt or other modded dirt and create the correct vanilla or modded grass.
* The same goes for mycelium.
* Additional registries can be easily created for new blocks which extend SpreadableBlock.

This PR does not address decay of spreadable blocks, because that can be easily handled by mods via mixin.

This PR also does not address allowing mods to change the conditions for spread (such as minimum and maximum light levels), although I could imagine this registry optionally storing a reference to an alternative canSpread method to be used in lieu of vanilla's call to the static method in SpreadableBlock (if only they used an instance method instead ... *sigh*).

I have added tests to the test mod and tested using runTestmodClient, and also by modifying Terrestria to use the API in production.


fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+	Identifier MYCELIUM = new Identifier("mycelium");
+
+	/**
+	 * Get the registry for a given registry ID, or create a new registry for the ID
+	 * if none currently exists.
+	 *
+	 * @param type The registry type Identifier for the desired spreadable block registry
+	 * @return The SpreadableBlockRegistry for the given ID
+	 */
+	static SpreadableBlockRegistry getOrCreateInstance(Identifier type) {
```
modmuss50: Maybe just `get` here? Im not sure the `OrCreateInstance` is really that useful to the API user.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SpreadableBlockMixin.java
```diff
@@ -0,0 +1,73 @@
...
+
+@Mixin(SpreadableBlock.class)
+public class SpreadableBlockMixin extends SnowyBlock {
+	private SpreadableBlockMixin(Settings settings) {
+		super(settings);
+	}
+
+	@Shadow
+	private static boolean canSpread(BlockState state, WorldView world, BlockPos pos) {
+		return false;
```
modmuss50: Nit: throw in the shadow method

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SpreadableBlockMixin.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	@Inject(
+			method = "randomTick",
+			at = @At(
+					value = "INVOKE_ASSIGN",
+					target = "Lnet/minecraft/util/math/BlockPos;add(III)Lnet/minecraft/util/math/BlockPos;",
```
modmuss50, line 57: Nit: a comment sayind where its injecting into might be helpful when porting this.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+	public static @Nullable SpreadableBlockRegistry getInstanceBySpreadable(@Nullable BlockState blockState) {
+		if (blockState == null) {
+			return null;
+		}
+
+		return getInstanceBySpreadable(blockState.getBlock());
+	}
+
+	public static @Nullable SpreadableBlockRegistry getInstanceBySpreadable(@Nullable Block block) {
+		if (block == null || !SPREADABLE_TYPE.containsKey(block)) {
```
modmuss50: Why would block ever be null?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+		return SpreadableBlockRegistryImpl.getOrCreateInstance(type);
+	}
+
+	/**
+	 * Gets the spreadable block state (if any) for a given bare block state.
+	 *
+	 * @param bareBlockState The bare block state to search the registry for
```
modmuss50, line 83: Just reading this is not super clear to me what `The bare block state` means? Would this be dirt as opposed to grass?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
gniftygnome: It creates so the API user can add a custom SpreadableBlock type.  While I don't have a specific need for this myself at the moment, I can imagine a lot of potential uses.  The test cases I provided include creating a spreadable version of Podzol that spreads to Coarse Dirt as an example.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
gniftygnome: It's Dirt for the two vanilla registries.  It could be anything for any modded registries.  I'll think about how to clarify that a bit.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SpreadableBlockMixin.java
gniftygnome: Done in 11a0801.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SpreadableBlockMixin.java
gniftygnome: Done in 11a0801.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
gniftygnome: I'm a C programmer.  I have trust issues.  :P
I've changed the methods to require non-null, in 11a0801.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
gniftygnome: In 11a0801 I've reworded a bunch of the API javadocs.  Hopefully this is better now, but let me know.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
```diff
@@ -0,0 +1,144 @@
...
+import net.minecraft.block.Blocks;
+import net.minecraft.registry.tag.TagKey;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.registry.SpreadableBlockRegistry;
+
+public final class SpreadableBlockRegistryImpl implements SpreadableBlockRegistry {
+	private static final Logger LOGGER = LoggerFactory.getLogger(SpreadableBlockRegistryImpl.class);
+
+	private static final Map<Block, Identifier> SPREADABLE_TYPE;
```
kevinthegreat1: This seems to only allow a block to be spreadable. What if I only want one block state of a block to be spreadable?

gniftygnome: I was told to keep it really simple, so I did.  Then I was told it has to cover a whole bunch more use cases (and here's another)...  Effectively at least for now this is pretty much a dead PR.  I intend to support all this stuff over in Terraform API where there aren't so many constraints on the implementation and I can already use MixinExtras.

kevinthegreat1: I understand your frustration, however, once you add something to FAPI, you basically can't make a breaking change at all to ensure compatibility. Therefore, a new API often goes through a lot of deliberation and review. Also, I'm just asking a question about something I noticed.

gniftygnome: I'm not frustrated, though.  I have an alternative avenue for this.  The only way it impacts me is other mod authors will likely use incompatible mixins because FAPI doesn't provide something.  I don't see this as a case of "diligence" on FAPI's end though.  It's more like a lack of consensus or leadership on what is acceptable for inclusion.  If I had written this just to get it into FAPI then it would be frustrating, but I had to do the work anyway for Terraform API.

modmuss50 (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+ * }</pre>
+ */
+@ApiStatus.NonExtendable
+public interface SpreadableBlockRegistry extends Block2ObjectMap<BlockState> {
+	/**
+	 * Registry ID for Minecraft Grass Block type spreadable blocks.
+	 */
+	Identifier GRASS = new Identifier("grass");
+	/**
+	 * Registry ID for Minecraft Mycelium Block type spreadable blocks.
```
Juuxel:
```suggestion
	 * Registry ID for mycelium-type spreadable blocks.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * <pre>{@code
+ * SpreadableBlockRegistry.getOrCreateInstance(SpreadableBlockRegistry.MYCELIUM)
+ *        .add(MyModBlocks.MY_DIRT, MyModBlocks.MY_MYCELIUM.getDefaultState());
+ * }</pre>
+ */
+@ApiStatus.NonExtendable
+public interface SpreadableBlockRegistry extends Block2ObjectMap<BlockState> {
+	/**
+	 * Registry ID for Minecraft Grass Block type spreadable blocks.
```
Juuxel:
```suggestion
	 * Registry ID for grass-type spreadable blocks.
```
No need to say "Minecraft" here, and can be shorter in general

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
Juuxel: That name does seem too long considering there's no `get` that doesn't have the "or create" functionality.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.util.Block2ObjectMap;
+import net.fabricmc.fabric.impl.content.registry.SpreadableBlockRegistryImpl;
+
+/**
+ * Registries of Blocks to BlockStates, defining the spreadable BlockState to replace
```
Juuxel:
```suggestion
 * Registries of blocks to block states, defining the spreadable block state to replace
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+ * a bare Block with when a particular type of spreadable Block spreads to the bare black.
```
Juuxel:
```suggestion
 * a bare block with when a particular type of spreadable block spreads to the bare block.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * <p>For example, to get the registry for Mycelium block spread and then register
+ * a modded block pair where Mycelium spreading to MY_DIRT generates MY_MYCELIUM:
```
Juuxel:
```suggestion
 * <p>For example, to get the registry for mycelium block spread and then register
 * a modded block pair where mycelium spreading to {@code MY_DIRT} generates {@code MY_MYCELIUM}:
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 * Registry ID for Minecraft Mycelium Block type spreadable blocks.
+	 */
+	Identifier MYCELIUM = new Identifier("mycelium");
+
+	/**
+	 * Get the registry for a given registry ID, or create a new registry for the ID
+	 * if none currently exists.
+	 *
+	 * @param type The registry type Identifier for the desired spreadable block registry
+	 * @return The SpreadableBlockRegistry for the given ID
```
Juuxel:
```suggestion
	 * @return The {@code SpreadableBlockRegistry} for the given ID
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 * @return An immutable copy of the spreadable block types map
+	 */
+	static ImmutableMap<Block, Identifier> getSpreadableTypes() {
+		return SpreadableBlockRegistryImpl.getSpreadableTypes();
+	}
+
+	/**
+	 * Gets the spreadable block state (if any) for a given bare block state.
+	 *
+	 * <p>For example, the GRASS registry should return the modded grass block state
```
Juuxel:
```suggestion
	 * <p>For example, the {@link #GRASS} registry should return the modded grass block state
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * @param bareBlockState The bare block state to search the registry for
+	 * @return The replacement spreadable block state for this registry, if any
+	 */
+	@Nullable BlockState get(BlockState bareBlockState);
+
+	/**
+	 * Gets the spreadable block state (if any) for a given bare block.
+	 *
+	 * <p>For example, the GRASS registry should return the modded grass block
```
Juuxel:
```suggestion
	 * <p>For example, the {@link #GRASS} registry should return the modded grass block
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
```diff
@@ -0,0 +1,144 @@
...
+	private static final Map<Identifier, SpreadableBlockRegistry> REGISTRIES;
+
+	static {
+		SPREADABLE_TYPE = new IdentityHashMap<>();
+		REGISTRIES = new HashMap<>();
```
Juuxel: Nit: These can be inline which is a bit more concise and typical in Fabric API

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 * @param bareBlock The bare block to search the registry for
+	 * @return The replacement spreadable block state for this registry, if any
+	 */
+	@Nullable BlockState get(Block bareBlock);
+
+	/**
+	 * Adds a registry entry to this registry for the given bare block to spreadable block conversion.
+	 *
+	 * <p>For example, a mod adding a custom dirt type would register its modded dirt block and
+	 * the corresponding modded grass block state to the GRASS registry.
```
Juuxel:
```suggestion
	 * the corresponding modded grass block state to the {@link #GRASS} registry.
```

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 * @param type The registry type Identifier for the desired spreadable block registry
+	 * @return The SpreadableBlockRegistry for the given ID
+	 */
+	static SpreadableBlockRegistry getOrCreateInstance(Identifier type) {
+		return SpreadableBlockRegistryImpl.getOrCreateInstance(type);
+	}
+
+	/**
+	 * Gets an immutable copy of the spreadable block types map, which maps the
+	 * canonical or "primary" block of the type to the registry for the type.
```
Juuxel, line 66: Doesn't `SpreadableBlockRegistry.add` map *all* blocks of the type to registry, not just a "primary" block?

https://github.com/FabricMC/fabric/blob/725e336a53e8714456674d2250107f2479cbfc4e/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java#L97-L102

Juuxel (changes requested):
Mostly javadoc notes

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
gniftygnome: Yes, `SpreadableBlockRegistry.add` is called to add all the spreadable block transitions to the associated registries.  The types map is basically a map of registry key (which is the primary spreadable block) to registry.  The map is used by the vanilla mixin to implement vanilla-to-mod spread.  Spread from a modded block must be implemented in the block.

You could argue the API implementation is not complete without a `FabricSpreadableBlock` type thing with a default implementation of the mod->any logic.  I wouldn't argue against it; this PR probably should be rewritten anyway, now that MixinExtras is available in the API.  When I have some time in January, I'd like to take another pass at trying to implement some of the more complicated requests (such as allowing mods to set the new block state).

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
gniftygnome: *shrugs*  There's a static constructor either way; I always prefer to init in the constructor personally but I've changed these back to inline.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * <p>For example, this could be used when implementing a custom block which
+	 * becomes whichever spreadable block there is the most of nearby.
+	 *
+	 * @return An immutable copy of the spreadable block types map
+	 */
+	static ImmutableMap<Block, Identifier> getSpreadableTypes() {
```
Juuxel, line 73: Is there any reason to use indirect `Identifier`s as the values instead of the registry instance directly?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
Juuxel: Right. I was just wondering if it's really the "primary" spreadable block if it applies to *all* spreadable blocks using the registry.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
modmuss50: I wonder if a RegistryKey is the soltuion here? or maybe leave that until we are forced into it?

Juuxel: It's the id of a SpreadableBlockRegistry, not a vanilla registry entry like Block. In other words, just a key of this map:
https://github.com/FabricMC/fabric/blob/fee11564048c76a1273685b0496d559d5c1ce5a4/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java#L41

The instances for each id are constant once they're created (and the spreadable block registries that are in the map have been created), so it'd be simpler to just have the registry itself in the map IMO.

modmuss50: Ah yeah, sorry I didnt look at the full context. I think this would be a good idea to change.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
```diff
@@ -0,0 +1,141 @@
...
+import net.minecraft.registry.tag.TagKey;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.registry.SpreadableBlockRegistry;
+
+public final class SpreadableBlockRegistryImpl implements SpreadableBlockRegistry {
+	private static final Logger LOGGER = LoggerFactory.getLogger(SpreadableBlockRegistryImpl.class);
+
+	private static final Map<Block, Identifier> SPREADABLE_TYPE = new IdentityHashMap<>();
+	private static final Map<Identifier, SpreadableBlockRegistry> REGISTRIES = new HashMap<>();
```
modmuss50, line 41: is there ever a case where they may be multiple `SpreadableBlockRegistry` for a given `Block`. I was looking at fixing it up following Juzz's comments and it seems there might not be. Unless im missing something this could just be a single Map<Block, SpreadableBlockRegistry>, without having any sort of ID at all?

modmuss50:
Going to keep this one in last call for a little longer, I think there are still some design questions to be clarified.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/impl/content/registry/SpreadableBlockRegistryImpl.java
gniftygnome: It took me a while to remember why I did that.  :)  If mod A adds a spreadable block and mod B optionally extends A, I thought it would be nice for mod B to be able to get the registry for that spreadable block without having to import mod A (and guard the reference) to get the block itself.  Using an identifier makes it possible to just create the required identifier in mod B when fetching the registry.

gniftygnome:
> Going to keep this one in last call for a little longer, I think there are still some design questions to be clarified.

Yes, I am still not quite to a good point to pick this up again (and as I mentioned, it could probably be improved now that MixinExtras is available, anyway).

modmuss50:
> Yes, I am still not quite to a good point to pick this up again

No worries at all, Im quite happy to get this to the finish line myself, its 99% of the way there.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/SpreadableBlockMixin.java
```diff
@@ -0,0 +1,79 @@
...
+	 * Shift by 2 so the target block position is available as a local variable.
+	 */
+	@Inject(
+			method = "randomTick",
+			at = @At(
+					value = "INVOKE_ASSIGN",
+					target = "Lnet/minecraft/util/math/BlockPos;add(III)Lnet/minecraft/util/math/BlockPos;",
+					shift = At.Shift.BY, by = 2
+			),
+			locals = LocalCapture.CAPTURE_FAILHARD
```
Syst3ms, line 60: As you mentioned yourself, `@Local` time. That being said, shifts and conditionals are always a bit weird, have you checked that the output is what we want?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/SpreadableBlockRegistry.java
Syst3ms: I concur with the other comments on this, should probably map to registry instances directly.

Syst3ms (changes requested):

PR #3389, opened 2023-10-24
Fix custom ingredient serialization with allowEmpty

Technici4n:
@jaredlll08 told me that custom ingredients did not serialize correctly with the `allowEmpty` codec. This is because the serialization code in our modified codec actually tried to serialize using the vanilla codec first, which fails when the ingredient is empty if `allowEmpty` is false **but succeeds if `allowEmpty` is true, thus writing an empty array `[]`**. I rewrote this using an `either` codec, cleaning up the code a bit and fixing the bug.

### DIFF
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
@@ -23,10 +23,8 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
 
-import com.mojang.datafixers.util.Pair;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.DataResult;
-import com.mojang.serialization.DynamicOps;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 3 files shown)


fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/IngredientMixin.java
```diff
@@ -35,9 +37,18 @@
 public class IngredientMixin implements FabricIngredient {
 	@Inject(method = "createCodec", at = @At("RETURN"), cancellable = true)
 	private static void injectCodec(boolean allowEmpty, CallbackInfoReturnable<Codec<Ingredient>> cir) {
-		final Codec<CustomIngredient> customIngredientCodec = allowEmpty ? CustomIngredientImpl.ALLOW_EMPTY_INGREDIENT_CODECS : CustomIngredientImpl.DISALLOW_EMPTY_INGREDIENT_CODECS;
-		Codec<Ingredient> ingredientCodec = customIngredientCodec.xmap(CustomIngredient::toVanilla, FabricIngredient::getCustomIngredient);
-		cir.setReturnValue(CustomIngredientImpl.first(cir.getReturnValue(), ingredientCodec));
```
apple502j, line 40: 1. This is the only use of `CustomIngredientImpl#first`. Consider removing it if it's no longer used?
2. I think the fix is as simple as flipping the two here.

Technici4n: No, we don't want to risk skipping the `type` field when reading IMO (e.g. there could be a custom ingredient with an `item` field). I meant to remove the codec but forgot, now done.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/CustomIngredientImpl.java
```diff
@@ -137,33 +132,4 @@ public boolean isEmpty() {
 	private <T> T coerceIngredient() {
 		return (T) customIngredient;
 	}
-
-	public static <T> Codec<T> first(Codec<T> first, Codec<T> second) {
-		return new First<>(first, second);
-	}
-
-	// Decode/encode the first codec, if that fails return the result of the second.
-	record First<T>(Codec<T> first, Codec<T> second) implements Codec<T> {
```
modmuss50, line 146: Nice to get rid of this, thanks ;)

fabric-recipe-api-v1/src/testmod/java/net/fabricmc/fabric/test/recipe/ingredient/SerializationTests.java
```diff
@@ -64,19 +65,28 @@ public void testArrayDeserialization(TestContext context) {
...
+			String ingredientJson = """
+					{"ingredients":[{"item":"minecraft:stone"}],"fabric:type":"fabric:all"}
+					""".trim();
 
-		var ingredient = new AllIngredient(List.of(
-				Ingredient.ofItems(Items.STONE)
-		));
-		String json = ingredient.toVanilla().toJson(false).toString();
-		context.assertTrue(json.equals(ingredientJson), "Unexpected json: " + json);
+			var ingredient = DefaultCustomIngredients.all(
```
modmuss50: Checkstyle, use the full type here.

modmuss50 (approved):
Looks good, just fix the build.

apple502j (approved):

Issue #3384, opened 2023-10-21
ClientLoginNetworkAddon does not handle unsuccessful query responses 


masmc05:
https://github.com/FabricMC/fabric/blob/1fbc78fabb035f7902f21f9f1f34c13ad4d02aff/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientLoginNetworkAddon.java#L87-L89

As we can see here, even though the listener correctly states that the buffer can be null, it doesn't handle that, resulting in NPE on serialization

Maybe it should have been like that?
```java
ServerboundCustomQueryAnswerPacket packet = new ServerboundCustomQueryAnswerPacket(queryId, result == null ? null : new PacketByteBufLoginQueryResponse(result));
```

modmuss50:
Just to confirm are you using Minecraft 1.20.2?

masmc05:
yes

modmuss50:
Out of intrest whats your usecase for using the login networking APIs? In 1.20.2 in most cases the configuration phase should be a tottal replacement for the login phase.

masmc05:
For several reasons from the past we add an additional security level for top admins, we'll feel safer that way

modmuss50:
I would highly reccomend looking into doing this during the configuration phase, its the perfect place to send addional data such as this.

apple502j:
No, this logic should be kept in the login phase.

liach:
> I would highly reccomend looking into doing this during the configuration phase, its the perfect place to send addional data such as this.

The login query request is guaranteed to receive a response from a vanilla client. In comparison, detecting whether a packet is understood with ping and pong to and attaching pending authentication info onto server-side client objects are more error-prone and is more invasive. And the login network handler has a hold of the vanilla authentication information that modders wish to access.

modmuss50:
> detecting whether a packet is understood with ping and pong to and attaching pending authentication info onto server-side client objects are more error-prone and is more invasive.

This is handled by Fabric API, if its error-prone I need to know about it ASAP as reg sync and other critical parts of fabric API wont be working as expected.

apple502j:
The problem here is not that configuration networking is doomed (it isn't), it is that for this specific use (security check) you don't want to even proceed to configuration in the first place.

PR #3394, opened 2023-10-29
Network addon session refactor

modmuss50:
Big thanks to @deirn for 98% of the changes in this PR, this is a follow on from #3377 to correctly handle the transition between config and play.

This PR also contains a fix for #3384 and now correctly validates the packet size in a few places where it wasnt previously. Testing still needs to be done, will likely release this in a beta.


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
```diff
@@ -145,17 +153,29 @@ public void endSession(AbstractNetworkAddon<H> addon) {
...
 		} finally {
 			lock.unlock();
 		}
 	}
 
+	/**
+	 * In practice, trackedAddons should never contain more than one instance.
+	 */
+	private void logTrackedAddonSize() {
+		if (LOGGER.isDebugEnabled() && this.trackedAddons.size() > 1) {
```
liach: Does this `size <= 1` constraint hold true on servers, given each server player has one network handler which has one network addon?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractNetworkAddon.java
```diff
@@ -64,13 +75,32 @@ public H getHandler(Identifier channel) {
 		}
 	}
 
+	private void assertNotReserved(Identifier channel) {
+		if (this.isReservedChannel(channel)) {
+			throw new IllegalArgumentException(String.format("Cannot register handler for reserved channel with name \"%s\"", channel));
```
liach: This IAE may be also thrown if it's for unregistering; consider updating the message.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
deirn: Actually, yeah. I forgot to check about it 😅.

modmuss50: I think ill drop this log to trace, to only be used for testing. I dont see a nice way to make this work for servers.

PR #3397, opened 2023-10-29
Fix Doubles not being editable in the game-rule menu

Estecka:
Hi,

Currently, game-rules of type Double are impossible to edit using the "Edit Gamerules" screen on world creation; any change to those rules is discarded upon exiting the screen.

This is due to `DoubleRule.validate` not updating the rule's value when called by the `DoubleRuleWidget`. For other gamerule types, this is how their widget update their value.

I have tested the fix in this PR on 1.20.2 using the artifacts generated on my fork. However the issue dates from furher back; at least since 1.19.4 where I first identified the issue. If this is accepted, I'd be interested in a backport of this fix to that version.

### DIFF
--- a/fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/rule/DoubleRule.java
+++ b/fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/rule/DoubleRule.java
@@ -113,7 +113,12 @@ public final class DoubleRule extends GameRules.Rule<DoubleRule> implements Vali
 		try {
 			final double d = Double.parseDouble(value);
 
-			return this.inBounds(d);
+			if (!this.inBounds(d)) {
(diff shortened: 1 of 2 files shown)


Technici4n:
`validate` is a poor name for this method, unfortunately not changeable in the API. Could you at least update the javadoc of `ValidateableRule` to say that it should also set the value?

Estecka:
By the way, I noticed DoubleRule is the only one that implements `ValidateableRule`. Shouldn't `BoundedInt` implement it as well ?

Technici4n:
> By the way, I noticed DoubleRule is the only one that implements `ValidateableRule`. Shouldn't `BoundedInt` implement it as well ?

It can't due to remapping (the method would not be available in production).

Technici4n (approved):

modmuss50 (approved):

phuchau1989 (approved):

PR #3407, opened 2023-11-06
Avoid serializing object-based payload into `PacketByteBuf` on the main thread

### LINKED ISSUE #3335: [Enhancement] Serialize custom packet in networking thread instead of the client/server thread
Currently, for the new object-based networking API, when creating the packet, it serializes the packet in the clinet/server main thread. Vanilla packets are serialized on networking thread. If the packet is large, moving the serialization to networking thread could improve performance. What's more, in singleplayer, the packet objects are passed by object reference, not serialized or deserialized, to improve performance, so it could also improve performance for singleplayer case.

https://github.com/FabricMC/fabric/blob/1.20.2/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerNetworkingImpl.java#L66C21-L66C21

The payload object carries `PacketByteBuf`. It's possible to make it to carry the packet object instead of buffer. https://github.com/FabricMC/fabric/blob/1.20.2/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/payload/PacketByteBufPayload.java

This cannot be done to the old networking API because it accepts the `PacketByteBuf` as argument.

deirn:
Resolves #3335

```mermaid
graph TD
  ResolvablePayload["<i>ResolvablePayload</i>"] --> RetainedPayload["RetainedPayload(Identifier, PacketByteBuf)"]
  
  ResolvablePayload --> ResolvedPayload["<i>ResolvedPayload</i>"]
  RetainedPayload -.-> ResolvedPayload
  
  ResolvedPayload --> TypedPayload["TypedPayload(FabricPacket)"]
  ResolvedPayload --> UntypedPayload["UntypedPayload(Identifier, PacketByteBuf)"]
```

On local connection (singleplayer), `ResolvedPayload` instances will be used directly, bypassing the serialization altogether.

On remote connection, the sender sends a `ResolvedPayload` instance. The receiver then receives a `PacketByteBuf` that gets retained and wrapped into `RetainedPayload`. **This doesn't create a new buffer, unlike the current implementation.**

Then the receiver checks for instance of `ResolvablePayload` and resolves it into `ResolvedPayload` if it's not yet as such.
- For `TypedPayload`, `PacketType#read` will be called.
- For `UntypedPayload`, the buffer from `RetainedPayload` will be copied into a new one.

Then the handler for the payload will be called as usual.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
@@ -54,9 +54,7 @@ public final class NetworkingImpl {
 
 	static {
 		if (FORCE_PACKET_SERIALIZATION) {
-			LOGGER.warn("Force Packet Serialization is enabled");
-			if (FabricLoader.getInstance().isDevelopmentEnvironment()) LOGGER.warn("This is the default on dev env");
-			LOGGER.warn("Disable it with -Dfabric-api.networking.force-packet-serialization=false");
+			LOGGER.info("Force Packet Serialization is enabled to mimic remote connection on single player, this is the default behaviour on dev env. Add -Dfabric-api.networking.force-packet-serialization=false JVM arg to disable it.");
 		}
 	}
 }
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
@@ -21,9 +21,22 @@ import org.slf4j.LoggerFactory;
 
 import net.minecraft.util.Identifier;
 
+import net.fabricmc.fabric.impl.networking.payload.TypedPayload;
+import net.fabricmc.fabric.impl.networking.payload.UntypedPayload;
+import net.fabricmc.loader.api.FabricLoader;
+
 public final class NetworkingImpl {
 	public static final String MOD_ID = "fabric-networking-api-v1";
 	public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
+
+	/**
+	 * Force {@link TypedPayload} to be serialized into {@link UntypedPayload}, mimicking remote connection.
+	 *
+	 * <p>Defaults to {@code true} in dev env and {@code false} in production.
+	 */
+	public static final boolean FORCE_PACKET_SERIALIZATION = Boolean.parseBoolean(System.getProperty(
+			"fabric-api.networking.force-packet-serialization",
+			Boolean.toString(FabricLoader.getInstance().isDevelopmentEnvironment())));
 
 	/**
 	 * Id of packet used to register supported channels.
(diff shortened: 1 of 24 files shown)


Technici4n:
Does this add a way to enable serialization for singleplayer, for example in dev?

deirn:
> Does this add a way to enable serialization for singleplayer, for example in dev?

No, is it really needed? If you want to check for serialization you can always run the server.

deirn:
~~One thing that I forgot to mention: if you send a packet using raw buffer when it's registered using the `PacketType` it will fail on local connection as it expects it to be `TypedPayload` when it is a `UntypedPayload`. Should we also handle the conversion?~~

Not true anymore, it will now convert between the two accordingly. Still, should it really do so?

modmuss50:
> > Does this add a way to enable serialization for singleplayer, for example in dev?
> 
> No, is it really needed? If you want to check for serialization you can always run the server.

I think by default it should always serialise when in dev, with a system property to control the behaviour.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadC2SPacketMixin.java
```diff
@@ -28,21 +26,18 @@
...
-import net.fabricmc.fabric.impl.networking.payload.PacketByteBufPayload;
-import net.fabricmc.fabric.impl.networking.payload.PayloadHelper;
+import net.fabricmc.fabric.api.networking.v1.PacketByteBufs;
+import net.fabricmc.fabric.impl.networking.payload.RetainedPayload;
 
 @Mixin(CustomPayloadC2SPacket.class)
 public class CustomPayloadC2SPacketMixin {
-	@Shadow
-	@Final
-	private static int MAX_PAYLOAD_SIZE;
```
modmuss50: We still need to check the max payload size, in the same way that vanilla does. I previously moved this logic to `PayloadHelper.read`.  It may be bestto just copy the vanilla behvaiour into each of these mixins.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
```diff
@@ -38,4 +51,12 @@ public final class NetworkingImpl {
 	public static boolean isReservedCommonChannel(Identifier channelName) {
 		return channelName.equals(REGISTER_CHANNEL) || channelName.equals(UNREGISTER_CHANNEL);
 	}
+
+	static {
+		if (FORCE_PACKET_SERIALIZATION) {
+			LOGGER.warn("Force Packet Serialization is enabled");
+			if (FabricLoader.getInstance().isDevelopmentEnvironment()) LOGGER.warn("This is the default on dev env");
+			LOGGER.warn("Disable it with -Dfabric-api.networking.force-packet-serialization=false");
+		}
```
modmuss50: Im not sure these need to be warning levels, or logged at all in development tbh.

modmuss50 (approved):
PR is looking good now, I just want to test it out myself next. One small thing regarding the logging around the new system prop.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/NetworkingImpl.java
deirn: How does one know to disable the serialization then? I guess most devs won't need to be concerned about this but still.

modmuss50: I would imagine that 99% of people wont care about this. A single .info log line would be fine, people see warnings and think they have done something wrong. Whats here makes it sound like it should be disabled not that it can be disabled.

Technici4n: Why would one want to disable it in the first place? ... ;)

Patbox: To make sure it doesn't bug on singleplayer, because of improper/not using copy of an object?

modmuss50 (approved):

modmuss50:
I will initally release this as a beta, many thanks for sorting this. 👍

PR #3413, opened 2023-11-11
Backport #2901 to 1.19.2

bluebear94:
This PR backports the core shader API in `fabric-rendering-v1` to 1.19.2. This consists mostly of renames of the 1.19.3 API’s code, plus a minor change in the test mod to use the pre-JOML `Matrix4f`.

## Motivation

While I want to use the standard Forge and Fabric APIs for custom core shaders in Caxton, Fabric’s APIs are not available on 1.19.2 or 1.18.2, both of which Caxton still supports.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramMixin.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.client.rendering.shader;
+
+import net.minecraft.client.gl.Program;
+import net.minecraft.client.render.Shader;
+
+import org.spongepowered.asm.mixin.Final;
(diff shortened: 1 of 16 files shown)


modmuss50:
I think its unlikely that we will backport this to 1.18.2, but 1.19.2 should be possible, is there anything else (anyone) may want backporting while we are at it?

fabric-rendering-v1/src/testmod/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.test.rendering.client;
+
+import com.mojang.blaze3d.systems.RenderSystem;
+
+import net.minecraft.client.render.Shader;
```
modmuss50: Please fix the checkstyle issues.

bluebear94:
I’ll try to backport this to 1.18.2 after this PR is merged.

Right now, I’d also prefer to have the `ResourceManagerHelper#registerBuiltinResourcePack(Identifier, ModContainer, Text, ResourcePackActivationType)` overload available on 1.19.2 and earlier (the original API used `String` instead of `Text`). But this isn’t a huge priority.

modmuss50:
> I’ll try to backport this to 1.18.2 after this PR is merged.

Im a little hesitant to do a 1.18.2 feature release as its bene so long, and is before we split making it a bit hard to backport, I guess we can? If so It would need to go in at the same time as this PR.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/ShaderProgramMixin.java
```diff
@@ -0,0 +1,58 @@
...
+
+import net.minecraft.client.gl.Program;
+import net.minecraft.client.render.Shader;
+import net.minecraft.resource.ResourceFactory;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.client.rendering.FabricShaderProgram;
+
+@Mixin(Shader.class)
+abstract class ShaderProgramMixin {
```
modmuss50: This fails the mixin validation step, can you please test that it builds and functions correctly (in prod and in your mod dev env).

PR #3419, opened 2023-11-12
add `Codec`s for `FluidVariant` and `ResourceAmount<FluidVariant>`

### LINKED ISSUE #2524: "Standard" format for fluid stack serialization
cc @MaxNeedsSnacks

reoseah:
Adds `Codec` fields for these two types, which might be useful for the new recipe serializers, which now use `RecipeSerializer#codec` for reading from json, instead of receiving `JsonObject` and reading it imperatively like it was before.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
@@ -53,15 +53,15 @@ import net.fabricmc.fabric.impl.transfer.fluid.FluidVariantImpl;
 public interface FluidVariant extends TransferVariant<Fluid> {
 	Codec<FluidVariant> CODEC = RecordCodecBuilder.create(instance ->
 			instance.group(
-				Registries.FLUID.getCodec().fieldOf("fluid").forGetter(FluidVariant::getFluid),
-				NbtCompound.CODEC.optionalFieldOf("nbt").forGetter(variant -> Optional.ofNullable(variant.getNbt()))
-		).apply(instance, (fluid, optionalNbt) -> FluidVariant.of(fluid, optionalNbt.orElse(null))));
+					Registries.FLUID.getCodec().fieldOf("fluid").forGetter(FluidVariant::getFluid),
+					NbtCompound.CODEC.optionalFieldOf("nbt").forGetter(variant -> Optional.ofNullable(variant.getNbt()))
+			).apply(instance, (fluid, optionalNbt) -> FluidVariant.of(fluid, optionalNbt.orElse(null))));
 
 	Codec<ResourceAmount<FluidVariant>> AMOUNT_CODEC = RecordCodecBuilder.create(instance ->
-                instance.group(
-                    CODEC.fieldOf("variant").forGetter(ResourceAmount::resource),
-                    Codec.LONG.fieldOf("amount").forGetter(ResourceAmount::amount)
-            ).apply(instance, ResourceAmount::new));
+			instance.group(
+					CODEC.fieldOf("variant").forGetter(ResourceAmount::resource),
+					Codec.LONG.fieldOf("amount").forGetter(ResourceAmount::amount)
+			).apply(instance, ResourceAmount::new));
 
     /**
 	 * Retrieve a blank FluidVariant.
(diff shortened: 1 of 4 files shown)


reoseah:
I thought IDEA had automatically added imports needed when I copy-pasted these lines from a mod of mine, but apparently it didn't

modmuss50:
> I thought IDEA had automatically added imports needed when I copy-pasted these lines from a mod of mine, but apparently it didn't

Please add some unit tests for this, make sure that the change works as expected.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/storage/base/ResourceAmount.java
```diff
@@ -27,4 +30,9 @@
  */
 @ApiStatus.Experimental
 public record ResourceAmount<T> (T resource, long amount) {
+	public static final Codec<ResourceAmount<FluidVariant>> FLUID_VARIANT_CODEC = RecordCodecBuilder.create(instance ->
```
Technici4n: This has no business being here, fluid-specific code should go in the `fluid` subpackage of the transfer API. This could probably go in `FluidVariant`.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
```diff
@@ -43,6 +50,12 @@
 @ApiStatus.Experimental
 @ApiStatus.NonExtendable
 public interface FluidVariant extends TransferVariant<Fluid> {
+	public static final Codec<FluidVariant> CODEC = RecordCodecBuilder.create(instance ->
```
Technici4n: This should probably a `MapCodec` in case some people want to use it inline with other fields?

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
```diff
@@ -43,6 +50,12 @@
...
+			instance.group(
+				Registries.FLUID.getCodec().fieldOf("fluid").forGetter(FluidVariant::getFluid),
+				NbtCompound.CODEC.optionalFieldOf("nbt").forGetter(variant -> Optional.ofNullable(variant.getNbt()))
```
Technici4n: If I remember correctly, the NBT parsing code might read small numbers as a different datatype than `int`, that should really be checked.

Technici4n (changes requested):
The specific format is something that needs more discussion. 81000 is not the most user friendly, and it would be nice if users also had the option to use a different format.

reoseah:
Moved the resource amount codec. This was enough for my use case and I'm not interested in putting more work into it. I'm also not familiar with data fixers beyond the common uses, so you'll have to find someone else for any codec shenanigans  

If the 81000 denominator for fluids is not user-friendly, why was it used in the first place? Is there a non-leaking abstraction for it now?

Technici4n:
> If the 81000 denominator for fluids is not user-friendly, why was it used in the first place? Is there a non-leaking abstraction for it now?

It is used internally, but should not necessarily be exposed to the user. I think being able to input as floating point buckets or millibuckets is more user-friendly, and should be offered as an option. Not sure about the exact format, here are a few options:
```json5
{
    "amount": {"unit": "bucket", "amount": 0.1 }, // a bit verbose
    "amount": "0.1 B", // parsing is a bit tricky but probably reasonable
    "amount": "100 mB", // could also support mB with the same format
    "amount": "8100 d", // maybe also droplets
}
```

Personally I like the string with a mandatory unit suffix. I can add the code to your PR if you want.

reoseah:
If doing fancy codecs, why not have one of "amount", "millibuckets" or "buckets". Basically – with variant also "flattened" – like this:
```json
{
  "fluid": "minecraft:water",
  "amount": 40500
},
{
  "fluid": "minecraft:water",
  "millibuckets": 500
},
{
  "fluid": "minecraft:water",
  "buckets": 0.5
}
```
If hiding that fluids use denominator of 81000, why not hide the fact that there are some "fluid variants", which users might never have heard of?

Technici4n:
Yeah we should definitely hide the "variant" too, IMO. I think `fluid`, `amount` and `nbt` should be 3 keys at the same nesting level.

reoseah:
If there are codecs for `FluidVariant` and its' `ResourceAmount`, there's an argument to be made that `ItemVariant` and its' `ResourceAmount` should get a codec too - at least for consistency. But then you might want these codecs to be consistent codecs inside `ItemStack` and `Ingredient`. Also, mods like TechReborn use `data` key for recipe outputs with NBT, not `nbt`.

Technici4n:
we can start with the fluid codecs only IMO, that makes things simpler

cputnam-a11y:
Hmm

reoseah:
Oh, I'm pretty sure the transfer-api types do have codecs, this was obsolete for some time, and GitHub doesn't allow to create another fork before deleting older one

PR #3421, opened 2023-11-13
23w45a porting fixes

apple502j:
- Fixes `openableByWindCharge` not being copied. Testing this the easy way (copying vanilla) turned out to be a little difficult because you have to compare `BlockSetType`, and the ID field has a colon when constructing with our builder.
- Re-adds one TAW that was accidentally removed.
- Updates testmod to use the proper way of throwing.
- Comment update to remain accurate.


apple502j:
I also request Object Builder minor version to be bumped, as we did not when we added wind charge field in the initial port.

Shnupbups (approved):

Technici4n (approved):

modmuss50 (approved):
Many thanks, ill merge with the next snapshot.

PR #3431, opened 2023-11-21
Fix `FluidStorageUtil#moveWithSound` Play Sound

Phoupraw:
It plays sound to all players only once now.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
@@ -111,7 +111,7 @@ public final class FluidStorageUtil {
 						if (!fill && handItem == Items.POTION) sound = SoundEvents.ITEM_BOTTLE_EMPTY;
 					}
 
-					player.getWorld().playSound(null, player.getX(), player.getEyeY(), player.getZ(), sound, SoundCategory.PLAYERS, 1, 1);
+					player.getWorld().playSound(player, player.getX(), player.getEyeY(), player.getZ(), sound, SoundCategory.PLAYERS, 1, 1);
 
 					return true;
 				}
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
@@ -111,7 +111,7 @@ public final class FluidStorageUtil {
 						if (!fill && handItem == Items.POTION) sound = SoundEvents.ITEM_BOTTLE_EMPTY;
 					}
 
-					player.getWorld().playSound(player, player.getX(), player.getEyeY(), player.getZ(), sound, SoundCategory.PLAYERS, 1, 1);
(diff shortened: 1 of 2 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
```diff
@@ -111,7 +111,7 @@ private static boolean moveWithSound(Storage<FluidVariant> from, Storage<FluidVa
 						if (!fill && handItem == Items.POTION) sound = SoundEvents.ITEM_BOTTLE_EMPTY;
 					}
 
-					player.playSound(sound, SoundCategory.BLOCKS, 1, 1);
+					player.getWorld().playSound(player, player.getX(), player.getEyeY(), player.getZ(), sound, SoundCategory.PLAYERS, 1, 1);
```
modmuss50, line 114: Has this been tested, I believe you need to pass null as the first param otherwise this is going to play for everyone BUT the player.

modmuss50, line 114:
```suggestion
					player.getWorld().playSound(null, player.getX(), player.getEyeY(), player.getZ(), sound, SoundCategory.PLAYERS, 1, 1);
```

modmuss50 (approved):
Just fixed and tested this myself, Thanks.

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Phoupraw: This method should be called both on client and server. `ClientWorld.playSound` will play sound to the `except`, while `ServerWorld.playSound` will play sound to all players but the `except`. This usage is according to vanilla usage, such as `FenceGateBlock#onUse`.

Phoupraw: This works, too.
Well this is better because it doesn't force modder to call it on both sides. It can only be called on server, but it also works if it's called on client.

modmuss50: Humm, I tested with the chute block in the FAPI test mods, they only call this on the server. I wonder if its worth wrappning this is a `!world.isClient`? Maybe @Technici4n would have some input on this?

Technici4n: Well it depends on which side modders are expected to call it. It's probably worth documenting.

modmuss50: Yes, you were right, it makes sense to call this on both sides. I have gone ahead and updated the testmod, this now works as expected. 👍

Technici4n (approved):

fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidStorageUtil.java
Phoupraw: Should we indicate it needs to be called on both sides in the doc?

PR #3432, opened 2023-11-21
Fix registry resource conditions on loot tables.

### LINKED ISSUE #3322: [Non functional] fabric-resource-conditions-api-v1
Seems that the `fabric:registry_contains` is non functional for loot tables.

Reasoning behind the claim :

When the following is present in loot_table json file 

```       
"condition": "fabric:registry_contains",
"registry": "minecraft:block",
"values": [
"modid:block_name"
```

You get the following error in console :  
`(Fabric Resource Conditions) Can't retrieve current registries. Failing registry_contains resource condition check.` 
this occurs if the block is present or not.
Game loads fine, no further console errors regarding the loot table json, however the loot table does not function (no item drops)

Removing the `fabric:registry_contains` does allow the loot to drop, but for the niche case i'm using it for (to stop console spam when another mod's block is not present - i'm already checking the mod is present) 
if the block is not present you get the console error as expected 
``` Couldn't parse element loot_tables:mymodid:blocks/missing_block_name ```

fabric_version=0.88.1+1.20.1
minecraft_version=1.20.1
yarn_mappings=1.20.1+build.10
loader_version=0.14.22
'fabric-loom' version '1.3-SNAPSHOT'

modmuss50:
If anyone has a better idea for passing the DRM instance to the load method please let me know. Closes #3322


apple502j:
I believe this is as simple as storing the DRM in a normal field rather than in ThreadLocal.

PR #3451, opened 2023-11-29
Update to loader 0.15

modmuss50:
Ensure all modules depend on the version of loader they are built against, a number of modules had mixins that depended on the "legacy" local cpature behaviour, I have replaced those broken with @Local

Going forward all modules will depend on the loader version that FAPI is built against.

I will combine this PR with a Gradle + other deps update as well, due to the need to update all modules.

PR #3471, opened 2023-12-16
Fix sleeping direction becoming `null`

### LINKED ISSUE #3470: Villagers and players appear rotated in beds when Fabric API is installed
The sleeping pose incorrectly renders entities' rotation, causing them to show up incorrectly in beds. This happens with Fabric API and no other mods installed. Removing Fabric API fixes the issue.

MC 1.20.4, Fabric 0.15.2, FAPI 0.91.2+1.20.4:

MC 1.20.4, Fabric 0.15.2, no other mods:

Extra information:
MC 1.20.4
Fabric 0.15.2
Fabric API 0.91.2+1.20.4
OS: Arch Linux
Java: openjdk 17.0.9 (same happens on 21)

apple502j:
Fixes #3470 
Reverts part of #3451

`getReturnValue` must be used in `RETURN`, not `INVOKE`.


modmuss50 (changes requested):
This mixin no longer applies, I think a better solution is to use a @WrapOperation

PR #3473, opened 2023-12-18
Refactor resource loader internals

### LINKED ISSUE #3302: pack.mcmeta functionality is not available

apple502j:
Resolves #3302
Supersedes #3312

Likely the biggest refactor to Resource Loader since its inception??? Work ongoing.

Note, if you've been using FAPI resource pack internals, this might break it (as expected).

- [x] Split each mod's resource packs.
- [x] Make sure not to break programmer art/high contrast.
- [x] Add pack dependency system.
- [x] Add a placeholder pack, to be used only for UI purposes.
- [x] Hide internal/per-mod packs.
- [x] Improve documentation.
- [x] Improve internal names.
- [x] Test.


fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/PackScreenMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.client.gui.screen.pack.ResourcePackOrganizer;
+
+import net.fabricmc.fabric.impl.client.resource.loader.FabricAbstractPack;
+
+@Mixin(PackScreen.class)
+public class PackScreenMixin {
+	@Inject(method = "method_29672", at = @At("HEAD"), cancellable = true)
+	private void hideHiddenPacks(PackListWidget packListWidget, String string, ResourcePackOrganizer.Pack pack, CallbackInfo ci) {
+		// (Fabric)AbstractPack is the only vanilla implementation of Pack, but we cast anyway.
+		if (pack instanceof FabricAbstractPack fabricAbstractPack && fabricAbstractPack.isHidden()) {
```
maityyy: Maybe interface inject `ResourcePackOrganizer.Pack` instead of `ResourcePackOrganizer.AbstractPack` to avoid this ugly `instanceof` check?

Edit: just replace `class_5369$class_5372` with `class_5369$class_5371`

apple502j: Yeah, maybe I should.

NebelNidas (changes requested):
Not an in-depth review yet, but I've noticed that the misleading `ResourcePack#getName()` Yarn name still hasn't been fixed (partly my fault, I forgot to create an [issue](https://github.com/FabricMC/yarn/issues/3758)).
I'd rename the corresponding fields and translation keys accordingly to avoid confusion later down the line. Also pinging @PepperCode1 to check if Continuity has any special requirements regarding the format

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
```diff
@@ -124,7 +155,7 @@ public static DataConfiguration createDefaultDataConfiguration() {
 		// as the data pack screen automatically put any data pack as disabled except the Default data pack.
 		for (ResourcePackProfile profile : moddedResourcePacks) {
 			try (ResourcePack pack = profile.createResourcePack()) {
-				if (pack instanceof FabricModResourcePack || (pack instanceof ModNioResourcePack && ((ModNioResourcePack) pack).getActivationType().isEnabledByDefault())) {
+				if (pack instanceof ModNioResourcePack && ((ModNioResourcePack) pack).getActivationType().isEnabledByDefault()) {
```
apple502j, line 194: todo: this is definitely wrong

apple502j:
Marked ready for review. Will need some testing.

apple502j:
Test successful, but will need more checks on mod compatibility, ResourceManagerHelper pack interactions, etc.

apple502j:
TODO:

- [x] Fix reordering of required packs being broken

fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsWriteVisitorMixin.java
```diff
@@ -0,0 +1,61 @@
...
+			// Nonexistent pack profiles should be handled in the same way as vanilla
+			if (profile == null || !profile.isHidden()) copy.add(pack);
+		}
+
+		return copy;
+	}
+
+	@SuppressWarnings("unchecked")
+	@ModifyArg(method = "visitObject", at = @At(value = "INVOKE", target = "Ljava/util/function/Function;apply(Ljava/lang/Object;)Ljava/lang/Object;"))
+	private <T> T skipHiddenPacks(T value, @Local String key) {
```
apple502j, line 56: Injecting here is much easier than maintaining the ordinals among the `visitObject` calls in `accept`.

fabric-resource-loader-v0/src/main/resources/fabric.mod.json
```diff
@@ -28,6 +28,9 @@
   ],
   "accessWidener": "fabric-resource-loader-v0.accesswidener",
   "custom": {
-    "fabric-api:module-lifecycle": "stable"
+    "fabric-api:module-lifecycle": "stable",
+    "loom:injected_interfaces": {
+      "net/minecraft/class_3288": ["net/fabricmc/fabric/impl/resource/loader/FabricResourcePackProfile"]
```
Juuxel: This will make an impl package interface appear to modders on a vanilla type, which encourages the use of impl methods.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackManagerMixin.java
```diff
@@ -65,4 +82,50 @@ public void construct(ResourcePackProvider[] resourcePackProviders, CallbackInfo
...
+		Set<String> seen = new LinkedHashSet<>();
+
+		while (it.hasNext()) {
+			ResourcePackProfile profile = it.next();
+			seen.add(profile.getName());
+
+			for (ResourcePackProfile p : this.profiles.values()) {
+				if (p.isHidden() && p.parentsEnabled(seen) && seen.add(p.getName())) {
+					it.add(p);
+					LOGGER.info("cur @ {}, auto-enabled {}, currently enabled: {}", profile.getName(), p.getName(), seen);
```
modmuss50: Nit: drop to debug level, see others as well.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/PlaceholderResourcePack.java
```diff
@@ -0,0 +1,118 @@
...
+import net.minecraft.resource.ResourcePack;
+import net.minecraft.resource.ResourcePackProfile;
+import net.minecraft.resource.ResourceType;
+import net.minecraft.resource.metadata.ResourceMetadataReader;
+import net.minecraft.text.Text;
+import net.minecraft.text.TextCodecs;
+import net.minecraft.util.Identifier;
+
+public record PlaceholderResourcePack(ResourceType type) implements ResourcePack {
+	public static final Text DESCRIPTION_TEXT = Text.translatable("pack.description.modResources");
```
modmuss50: NIT: private

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
```diff
@@ -67,33 +73,58 @@ public static void appendModResourcePacks(List<ModResourcePack> packs, ResourceT
...
+	public static JsonObject getMetadataPackJson(int packVersion, JsonElement description) {
 		JsonObject pack = new JsonObject();
 		pack.addProperty("pack_format", packVersion);
-		pack.addProperty("description", description);
+		pack.add("description", description);
+		return pack;
+	}
+
+	public static String serializeMetadata(int packVersion, String description) {
+		JsonObject pack = getMetadataPackJson(packVersion, new JsonPrimitive(description));
```
modmuss50: Might be nice for this to use the existing vanilla PackMetadata classes + codecs to create the json. This is something I changed in my PR here: https://github.com/FabricMC/fabric/pull/3312/files#diff-abb5e49de03dd72c0540388d57e30fcb1bbb255800703e9e69ce9e1e97aff21cR87

Not a blocker for this PR, just maybe nice to have.

modmuss50 (commented):
Looks good, nothing major is standing out to me. The main thing with this is going to be testing, to A) make sure it works as expected, and B) make sure its not slower anywhere.

Ensuing that the test mod fully utilises the features, and documents how to test it works correctly would go a long way. Im very happy to help with this if needed.

fabric-resource-loader-v0/src/main/resources/fabric.mod.json
apple502j: What's the best way to solve this? Mark it as ApiStatus.Internal? Or just not use II at all?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackManagerMixin.java
apple502j: Yeah, they will be removed when it's done (or dropped to debug).

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/FabricResourcePackProfile.java
```diff
@@ -0,0 +1,47 @@
...
+	 * if each of the parent is enabled. Note that non-Fabric packs always return {@code true}.
+	 *
+	 * @return whether every parent is enabled.
+	 */
+	default boolean parentsEnabled(Set<String> enabled) {
+		return true;
+	}
+
+	default void setParentsPredicate(Predicate<Set<String>> predicate) {
+	}
```
modmuss50: Nit: these methods should be prefixed with `fabric_` as they are not part of the API.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
```diff
@@ -67,33 +78,85 @@ public static void appendModResourcePacks(List<ModResourcePack> packs, ResourceT
...
 	}
 
-	public static boolean containsDefault(ModMetadata info, String filename) {
-		return "pack.mcmeta".equals(filename);
+	public static void refreshAutoEnabledPacks(List<ResourcePackProfile> enabledProfiles, Map<String, ResourcePackProfile> allProfiles) {
+		LOGGER.debug("[Fabric] Starting internal pack sorting with: {}", enabledProfiles.stream().map(ResourcePackProfile::getName).toList());
+		enabledProfiles.removeIf(profile -> ((FabricResourcePackProfile) profile).isHidden());
+		LOGGER.debug("[Fabric] Removed all internal packs, result: {}", enabledProfiles.stream().map(ResourcePackProfile::getName).toList());
+		ListIterator<ResourcePackProfile> it = enabledProfiles.listIterator();
+		// LinkedHashSet for debug log sort, will switch to normal HashSet once finished
```
modmuss50: Does this need doing?

fabric-resource-loader-v0/src/test/java/net/fabricmc/fabric/test/resource/loader/unit/ModResourcePackUtilTests.java
```diff
@@ -0,0 +1,248 @@
...
+		ResourcePackProfile programmerArt = mockProfile(profiles, "programmer_art", null);
+		ResourcePackProfile highContrast = mockProfile(profiles, "high_contrast", null);
+		ResourcePackProfile userPackA = mockProfile(profiles, "user_pack_a", null);
+		ResourcePackProfile userPackB = mockProfile(profiles, "user_pack_b", null);
+		modAProfiles.putAll(profiles);
+		modBProfiles.putAll(profiles);
+		allProfiles.putAll(modAProfiles);
+		allProfiles.putAll(modBProfiles);
+
+		testRefreshAutoEnabledPacks(
```
modmuss50, line 77: Nit: You would usually have a test method for each of these, not the end of the world but would make it more clear if just one of these tests broke.

Happy to leave this be.

modmuss50 (approved):
Looks good, what kind of testing has been done on this? There is quite a bit of room for things to go wrong so want to be 100% sure that we are confident with it.

Im happy to move it to last call.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
apple502j: Will just remove the comment, since we keep the debug? Shouldn't hurt perf that much, as the injection target is IO-intensive anyway.

catter1:
Thank you so much for working on this! I am excited to test it out, and be able to easily create updates for Fabric again.

jellysquid3:
This has [caused breakages in Sodium](https://github.com/CaffeineMC/sodium-fabric/issues/2286), because the high contrast/programmer art packs for mods are given the same resource pack name, which it expects to not be possible.

jellysquid3:
Additionally, this causes massive regressions in UX since now every mod's resource pack is shoved into the space constrained list. Doing this makes the user interface borderline unusable, since it's completely overloaded with entries that 99.9% of users will not understand what to do with.

image

Maybe this specific aspect of the pull request should be reverted until a better user interface is engineered? I really don't think this is helpful for users.

apple502j:
@jellysquid3 Uhh, my test environment did have Sodium 0.5.5, and I did not encounter any errors while booting up or using High Contrast. It could be due to changes in 0.5.6+, I am honestly unsure.

> because the high contrast/programmer art packs for mods are given the same resource pack name, which it expects to not be possible.

[Each `ResourcePackProfile` should have a different name](https://github.com/FabricMC/fabric/pull/3473/files#diff-316ac23c11543330b8a7db8932b07521cc6fc3c95508d1ab618a43b296708776R125), but it seems like I forgot to override `ResourcePack#getName`. This is my oversight, yep.

> Additionally, this causes massive regressions

The change should have [removed any "internal" packs from `PackResourceOrganizer`](https://github.com/FabricMC/fabric/pull/3473/files#diff-cb2a01758c3e03a48be5aa8189058fc0fe378f7bc93e9218f6d36868677cf2a2R50), making them invisible to the client. The screenshot obviously says that was not the case, though. Could you provide more info on how to reproduce the issue?

jellysquid3:
> It could be due to changes in 0.5.6+, I am honestly unsure.

We [made a change](https://github.com/CaffeineMC/sodium-fabric/commit/5c04268709f8cc27f7b9bc1cc063cd1da63e3d8f) very shortly before Fabric API 0.95.3 was published (a few days ago) that put the resource scan results into a hash table, which doesn't tolerate duplicate keys. It was just really bad luck, and we have since fixed the code to be more tolerant of that case.

> The screenshot obviously says that was not the case, though. Could you provide more info on how to reproduce the issue?

This may have since been fixed. Another user reported it to us with Fabric API 0.95.3, but will see if I can reproduce it myself.

dicedpixels:
This happens when "[Resource Pack Overrides](https://modrinth.com/mod/resource-pack-overrides)" is present.

PR #3476, opened 2023-12-19
Data Attachment API

Syst3ms:
Trying to revive #2172, with some ideas from @Technici4n's NF implementation. 

The Data Attachment API provides a unified and convenient way for modders to attach arbitrary pieces of data to various game objects (as of writing: entities, block entities, worlds and chunks). Importantly, all of the logic specific to each game object type is abstracted away behind an interface and is not the modder's concern.

Additionally, attached data can optionally be made to persist across server restarts. A subsequent PR should work to implement a robust and flexible client-server syncing mechanism, staying true to the API's "target type should not matter in the API" philosophy.

### API Overview

The API contains three classes: `AttachmentType`,`AttachmentRegistry` and `AttachmentTarget`.

An `AttachmentType<A>` is a type of way to attach objects of type `A` to game objects. It specifies whether the attachments are persistent and/or synced, and if one of these is enabled, provides a `Codec<A>` for convenient encoding/decoding. The API assumes the objects of type `A` are immutable or, more generally, do not share any state with other objects used in separate attachments.

Attachment types are created and registered using `AttachmentRegistry`. This class provides simple `create`, `createDefaulted` and `createPersistent` methods for quick creation and registration of attachments, but when finer control is needed, provides an `AttachmentRegistry.Builder` to set each desired property separately.

Finally, `AttachmentTarget` defines the game objects to which data can be attached. Currently, `Entity`, `BlockEntity`, `ServerWorld` and `WorldChunk` are made to implement it using interface injection and mixin. It provides methods for getting, setting, modifying and removing attachments, as well as `hasAttached` for check if an attachment exists.

### Example use

Attachment types are created and registered using methods in `AttachmentRegistry`, and should be stored in `static final` fields, like so:
```java
public class ModAttachmentTypes {
    public static final AttachmentType<String> INTERNAL_NAME = AttachmentRegistry.create(
        new Identifier("example:internal_name")
    ); // String attachment type, no defaults, not persisted, not synced
    public static final AttachmentType<Integer> MANA = AttachmentRegistry.createDefaulted(
        new Identifier("example:mana"), 
        () -> 0
    ); // Integer attachment with default value 0
    public static final AttachmentType<String> PERSISTENT_NAME = AttachmentRegistry.createPersistent(
        new Identifier("example:persistent_name"),
        Codec.STRING
    );  // String attachment, persists across server restarts
    public static final AttachmentType<Integer> PERSISTENT_MANA = AttachmentRegistry.<Integer>builder() // Builder for finer control
       .codec(Codec.INT) // required codec for persistence
       .persistent() // persistent
       .copyOnDeath() // will persist over entity death and respawn
       .initializer(() -> 0) // default value
       .buildAndRegister(new Identifier("example:persistent_mana"));
}
```
Once attachments have been registered, they can be used on all `AttachmentTarget`s. In this example, we assume all of the relevant targets start with no attachments.
```java
Entity entity = /* omitted */;
ServerPlayerEntity player = /* omitted */;

String internalName = entity.getAttached(ModAttachmentTypes.INTERNAL_NAME); // null
entity.setAttached(ModAttachmentTypes.INTERNAL_NAME, "silly entity");
internalName = entity.getAttached(ModAttachmentTypes.INTERNAL_NAME); // "silly entity"

boolean hasAttachment = player.hasAttached(ModAttachmentTypes.MANA); // false
int mana = player.getAttachedOrCreate(ModAttachmentTypes.MANA); // 0, auto-initialized !
hasAttachment = player.hasAttached(ModAttachmentTypes.MANA); // now true
player.modifyAttached(ModAttachmentTypes.MANA, mana -> mana+1); // increment mana

player.setAttached(ModAttachmentTypes.PERSISTENT_MANA, 10); // will be saved
```

### Future enhancements

Client-server syncing would be a great feature to have, but has been left out of this PR on account of being a lot trickier than the rest, and not as essential. It should instead be the subject of a future PR. However, this feature is important because syncing rendering data with the client seems like a very obvious addition to this API. I'll outline a couple of the constraints any potential addition should try to fit into:

- The syncing API should be target-agnostic, i.e not expose or involve too many mechanics related to the underlying target game objects, in line with the overall design of `AttachmentTarget`.
- It should allow partial syncing, i.e give modders the ability to only sync a part of attachments with the client. This is useful when only a couple fields in a complex attached datatypes are useful for rendering, as to not send useless data.
- It should give control over which clients the data is synced to. For instance, attached data relating to HUD rendering should only be synced with its target.
- Optionally, it could allow incremental syncing, i.e try to sync only what has changed (imagine an attached Map, only the added/removed key-value pairs need to be synced). While I don't see a way to make this fully automatic, it could perhaps be helped with some helper interfaces for attached datatypes to implement.

*Maintainer note: the unit tests test persistence for entities and blockentities as those can be mocked without too much effort, but mocking worlds and chunks is such a pain that I opted to make those into a separate testmod. It works by launching a server (on the same world) twice from the client to check the saving/loading.*

TODO:

- [x] Javadoc
- [x] Tests
- [x] ~~Implement server-client syncing~~ Scrapped in favor of future PR


fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -0,0 +1,139 @@
...
+		 */
+		Builder<A> initializer(Supplier<A> supplier);
+
+		/**
+		 * Sets the {@link AttachmentSerializer} for this attachment.
+		 *
+		 * @param serializer the serializer
+		 * @return the builder
+		 */
+		Builder<A> serializer(AttachmentSerializer<A> serializer);
```
maityyy: Can we have here a simplification for codecs like this?
```java
default Builder<A> serializer(Codec<A> codec) {
    return serializer(AttachmentSerializer.fromCodec(codec));
}
```
In fact, I don't see the point in giving anything other than codecs, but I understand that some people don't like codecs. Although I think modders should just get used to codecs since they'll be everywhere in the future

modmuss50: I agree, with the way Mojang are changing seemingly everything over to Codecs, we should do the same with new APIs.

Syst3ms: I can certainly add shortcuts for Codecs, but should I go further and ditch AttachmentSerializer in favor of them?

modmuss50: I dont see why not?

maityyy: Yes, that's exactly what I meant

Syst3ms: I guess the old interface design was more appropriate for the old proposal's receiver-aware design, now there's nothing this can do than codecs can't.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,121 @@
...
+	default void readAttachmentsFromNbt(NbtCompound nbt) {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * @return whether the {@link AttachmentTarget} has any attachment data to serialize
+	 */
+	default boolean hasSerializableAttachments() {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
```
modmuss50: Do these need to be part of the public API?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/Attachment.java
```diff
@@ -0,0 +1,61 @@
...
+	 *
+	 * @return the serializer, may be null
+	 */
+	@Nullable
+	AttachmentSerializer<A> serializer();
+
+	/**
+	 * @return whether the attached data is synced between server and client
+	 */
+	boolean synced();
```
modmuss50: As discussed lets leave syncing to the mod, at least for this inital PR

modmuss50 (commented):
Looks like a reasonable start, it should be nice and easy to write some good unit tests for this, alongside a testmod.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/Attachment.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * An attachment allows "attaching" arbitrary data to various game objects (entities, block entities, worlds and chunks at the moment).
+ * Use the methods provided in {@link AttachmentRegistry} to create and register attachments.
+ *
+ * <p>Attachments can optionally be made to persist between restarts using {@link AttachmentSerializer}s, and
+ * can optionally be automatically synced between server and client.</p>
+ *
+ * @param <A> type of the attached data
+ */
+public interface Attachment<A> {
```
maityyy: I think `Attachment` should be renamed to `AttachmentType` (as in the original PR) since it is really a type or key and not an instance. This is also similar to what vanilla does with `IntProvider`, (Tree)`TrunkProvider`, (Brain)`MemoryType`, etc.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
Syst3ms: I wasn't entirely sure how to make this impl-only, but I've figured out a way.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.attachment.v1.Attachment;
+import net.fabricmc.fabric.api.attachment.v1.AttachmentRegistry;
+
+public final class AttachmentRegistryImpl {
+	private static final Logger LOGGER = LoggerFactory.getLogger("fabric-api-data-attachment-api-v1");
```
maityyy: fabric-**api**-data-atachment-**api**-v1? :D

Syst3ms: Makes more sense, but this seems a bit inconsistent across FAPI. I'll change it though.

maityyy: I think this can be automated via gradle (and then we can use smth like `"$module_name"` like in FMJ), but that is out of scope of this PR

Syst3ms:
We need to figure out a couple things regarding how the API of `AttachmentTarget` should look. There are two main paradigms with endless in-between delineations:

1. All attachments have a mandatory initializer (usually in the form of a `Supplier`), and if no data is attached (i.e `hasAttached` returns `false`), then `getAttached` automatically calls the initializer and returns the result.
    - Pros:
        - Aligns with the implementation in Neoforge by @Technici4n 
        - Avoids nullability
        - Many types (like primitives) have very obvious default values
    - Cons:
        - Requires a lot of `hasAttached`/`getAttached` boilerplate to deal with no attached data, e.g if another method expects a nullable parameter, one would probably have to use a ternary.
        - Many types have no good default/dummy value, and the auto-initializing behavior may be undesirable then.
2. Attachments are not automatically initialized, and `getAttached` returns `null` if no data is attached.
    - Pros:
        - Avoids implicit initialization (whether that is a pro is arguable)
        - Avoids initializing complex types with no good dummy values
        - Akin to `Map#get`, which is a very familiar pattern
    - Cons:
        - Having to deal with nullable values
        - Would require a `getOrCreate` helper method, which can be annoying boilerplate in many cases.

Naturally, many of these cons can be solved or at least alleviated with suitable helper methods. In 1, a `getAttachedOrNull` can be provided to emulate 2's behavior at little extra cost. In 2, a `getAttachedOrThrow` method removes nullability concerns where they are uncalled for, but can also introduce hard failure (exception). 

An in-between may be to provide partial/optional auto-initialization capability. For example, an optional initializer that, if given, is called when there is no attached data as in 1. Or, for a less intrusive alternative, an optional 'default' initializer that is used as the parameter of `getOrCreate`, for a more concise `target.getOrCreate(attachment)` callsite. But again, that can introduce exceptions. To remedy this and also potentially other issues, it is possible to introduce subinterfaces of `Attachment` to prevent misuse through the type system, but this may be gross overengineering.

In any case, the choice should be guided by which paradigm will cause the least boilerplate in the majority of cases. Whatever behavior `getAttached` ends up having, it should be what's likely to be most commonly used. Any additional insight on the matter will be helpful here. @modmuss50 @maityyy

Syst3ms:
After some discussion, I have implemented the following compromise:

By default, `AttachmentType` does *not* automatically initialize attachments on first query. However, various factory methods allow the user to create a `DefaultedAttachmentType<A>`, which comes with an initializer to provide a default (non-null) value. Accordingly, AttachmentTarget has a new, non-nullable `getAttached` overload taking in a `DefaultedAttachmentType` that will automatically initialize non-existing attachments. 

As a result, whether the attachment automatically initializes or not, the call site looks like `target.getAttached(TYPE)`. To prevent users from calling the wrong `getAttached` (for example due to a not-specific-enough type when storing the registered type, which the javadoc will warn against doing), a runtime check is performed along with a warning. This should prevent improper use of the system.

maityyy:
What if I registered my `AttachmentType` with an initializer but my field is of type `AttachmentType` instead of `DefaultedAttachmentType`? Then all my calls will escape initialization (because at compile time it will be `getAttached(AttachmentType)` instead of `getAttached(DefaultedAttachmentType)`) when there is no value and I will most likely get an NPE (you also pointed this out in the docs). I don't like the idea of sub-interfaces, it's very non-obvious and difficult to expand in the future.

I can suggest the following:

1. Rename `getAttached(DefaultedAttachmentType)` to be different from `getAttached(AttachmentType)` — then modders will likely realize they are calling the wrong method
2. Get rid of `DefaultedAttachmentType` and its `getAttached` overload, call the initializer if it is not `null` in the only remaining `getAttached` every time (like fastutil's maps)
3. The same thing, but in a separate method called `getAttachedOrDefault`
 - 2 & 3 will always have the `@Nullable` annotation

For now I prefer option 2, but it's not ideal, so I'd like to think more...

Syst3ms:
I mentioned this potential pitfall, and if you check, the (normal) getAttached implementation issues a warning and auto-inits if you pass a defaulted one.

But you're right about bad extensibility, the cursory amount of thought I've given syncing tells me that is quickly going to be unsustainable, so back to the drawing board.

That being said, the issue with 2 is that static checkers will complain about nullability even for defaulted attachments (could be solved with a getAttachedOrThrow).

The problem with 3 is overarching, and it's that I'm unsure how common defaulted will be vs non-defaulted, and I don't want to have a longer method name be used for the most common case.

fabric-data-attachment-api-v1/src/testmod/resources/fabric.mod.json
```diff
@@ -0,0 +1,15 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-data-attachment-api-v1-testmod",
+  "name": "Fabric Data Attachment API (v1) Test Mod",
+  "version": "1.0.0",
+  "environment": "*",
+  "license": "Apache-2.0",
+  "depends": {
+  },
```
modmuss50: nit nit: remove

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/PersistentAttachmentTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class PersistentAttachmentTest implements ModInitializer {
+	public static final String MOD_ID = "fabric-data-attachment-api-v1-testmod";
+	public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
+
+	private boolean firstLaunch = true;
+
+	@Override
+	public void onInitialize() {
+		AttachmentType<String> dummy = AttachmentRegistry.createPersistent(
```
modmuss50: Nit: the test mods will likely be used as an example, I would make the attachment a public static final field to better match how its likely going to be used.

fabric-data-attachment-api-v1/src/test/java/net/fabricmc/fabric/test/attachment/CommonAttachmentTests.java
```diff
@@ -0,0 +1,196 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.chunk.WorldChunk;
+
+import net.fabricmc.fabric.api.attachment.v1.AttachmentRegistry;
+import net.fabricmc.fabric.api.attachment.v1.AttachmentTarget;
+import net.fabricmc.fabric.api.attachment.v1.AttachmentType;
+import net.fabricmc.fabric.impl.attachment.AttachmentPersistentState;
+import net.fabricmc.fabric.impl.attachment.AttachmentSerializingImpl;
+
+public class CommonAttachmentTests {
```
modmuss50, line 60: Very nice, great job 👍

fabric-data-attachment-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,34 @@
...
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.15.1"
+  },
+  "description": "TODO",
```
modmuss50: TODO

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
```diff
@@ -0,0 +1,15 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin",
+  "compatibilityLevel": "JAVA_17",
+  "mixins": [
+    "attachment.AttachmentTargetsMixin",
+    "attachment.BlockEntityMixin",
+    "attachment.ChunkSerializerMixin",
+    "attachment.EntityMixin",
+    "attachment.ServerWorldMixin"
```
modmuss50:
```suggestion
    "AttachmentTargetsMixin",
    "BlockEntityMixin",
    "ChunkSerializerMixin",
    "EntityMixin",
    "ServerWorldMixin"
```

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
```diff
@@ -0,0 +1,15 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin",
```
modmuss50:
```suggestion
  "package": "net.fabricmc.fabric.mixin.attachment",
```

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
```diff
@@ -0,0 +1,55 @@
...
+	@Inject(at = @At("TAIL"), method = "<init>")
+	private void createAttachmentsPersistentState(MinecraftServer server, Executor workerExecutor, LevelStorage.Session session, ServerWorldProperties properties, RegistryKey<World> worldKey, DimensionOptions dimensionOptions, WorldGenerationProgressListener worldGenerationProgressListener, boolean debugWorld, long seed, List<Spawner> spawners, boolean shouldTickTime, RandomSequencesState randomSequencesState, CallbackInfo ci) {
+		// Force persistent state creation
+		ServerWorld world = (ServerWorld) (Object) this;
+		var type = new PersistentState.Type<>(
+				() -> new AttachmentPersistentState(world),
+				nbt -> AttachmentPersistentState.read(world, nbt),
+				DataFixTypes.LEVEL
+		);
+		world.getPersistentStateManager().getOrCreate(type, AttachmentPersistentState.ID);
```
modmuss50, line 40: Question: Does this save the nbt to the data folder even if no attachments are used?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
```diff
@@ -0,0 +1,92 @@
...
+
+import net.fabricmc.fabric.api.attachment.v1.AttachmentType;
+import net.fabricmc.fabric.impl.attachment.AttachmentSerializingImpl;
+import net.fabricmc.fabric.impl.attachment.AttachmentTargetImpl;
+
+@Mixin(value = {BlockEntity.class, Entity.class, World.class, WorldChunk.class})
+public class AttachmentTargetsMixin implements AttachmentTargetImpl {
+	@Unique
+	@Nullable
+	protected IdentityHashMap<AttachmentType<?>, Object> fabric_dataAttachments = null;
```
modmuss50: Nit: `fabric_` prefix is not needed with @Unique

modmuss50:
```suggestion
	private IdentityHashMap<AttachmentType<?>, Object> fabric_dataAttachments = null;
```

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
```diff
@@ -0,0 +1,92 @@
...
+	@Override
+	@Nullable
+	public <T> T getAttached(AttachmentType<T> type) {
+		return fabric_dataAttachments == null ? null : (T) fabric_dataAttachments.get(type);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	@Nullable
+	public <T> T setAttached(AttachmentType<T> att, @Nullable T value) {
```
modmuss50: Nit: Id possibly rename `att` to `attachment`?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
```diff
@@ -0,0 +1,48 @@
...
+	default void readAttachmentsFromNbt(NbtCompound nbt) {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * @return whether the {@link AttachmentTarget} has any attachment data to serialize
+	 */
+	default boolean hasPersistentAttachments() {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
```
modmuss50: As theser are impl methods I would prefix them with `fabric_` to prevent possible conflicts with other mods. (Its unlikely but trivial to do)

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentSerializingImpl.java
```diff
@@ -0,0 +1,96 @@
...
+		var compound = new NbtCompound();
+
+		for (Map.Entry<AttachmentType<?>, ?> entry : attachments.entrySet()) {
+			AttachmentType<?> type = entry.getKey();
+
+			if (type.persistent()) {
+				Codec<Object> codec = (Codec<Object>) type.codec();
+				// non-nullity enforced by builder API
+				codec.encodeStart(NbtOps.INSTANCE, entry.getValue())
+						.result()
```
modmuss50: Question: What happens with errors when (de)serialising? What does the error look like? Can we catch the error and possibly show something useful with the attachment type id?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.world.PersistentState;
+
+/**
+ * Backing storage for server-side world attachments.
+ * Thanks to custom [.isDirty] logic, the file is only written if something needs to be persisted.
+ */
+public class AttachmentPersistentState extends PersistentState {
+	public static final String ID = "fabric_data_attachments";
```
modmuss50: Question: Should this `fabric:data_attachments` ?

Should this include some sort of version to allow for possible changes later down the line? Maybe a version should go into the NBT? Or not at all?

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
Syst3ms: Wait, it isn't convention to have the mixins in a different folder for each module? Must've looked wrong at other modules

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
Syst3ms: I think it doesn't. I haven't tested it myself, but this mixin and the persistent state class (which is supposed to not do this on purpose) was copied from Technici4an (missing a couple `@author`s that reminds me), and I recall him saying it indeed doesn't, but I can't find where.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
Syst3ms: `@Unique` mangles the name to guarantee uniqueness, correct?

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
modmuss50: Keep the mixins in the same package, you can just clean up the json a little.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
Syst3ms: probably even better to `type`

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentSerializingImpl.java
Syst3ms: Well, that would only happen if:

1. Someone manually wrote some stuff in Fabric's attachment section and fucked it up
2. A codec was changed in a completely backwards-incompatible way (which is highly unlikely due to the modularity of Codecs and the stability of the base types, even more so in vanilla code).

Both seem pretty unlikely, but I'll try to test what kind of error we'd get.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
Syst3ms: unsure of the general conventions about this tbh, are there any other modules that add sections to NBT like this?

fabric-data-attachment-api-v1/src/main/resources/fabric-data-attachment-api-v1.mixins.json
Syst3ms: oh right i'm stupridr

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
```diff
@@ -0,0 +1,51 @@
...
+import net.minecraft.nbt.NbtCompound;
+
+import net.fabricmc.fabric.impl.attachment.AttachmentTargetImpl;
+
+/**
+ * @author Technici4an
+ */
+@Mixin(BlockEntity.class)
+public class BlockEntityMixin implements AttachmentTargetImpl {
+	@Dynamic
```
LlamaLad7: Shouldn't be necessary?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
```diff
@@ -0,0 +1,58 @@
...
+
+import net.fabricmc.fabric.impl.attachment.AttachmentPersistentState;
+
+/**
+ * @author Technici4an
+ */
+@Mixin(ServerWorld.class)
+public class ServerWorldMixin {
+	@Inject(at = @At("TAIL"), method = "<init>")
+	private void createAttachmentsPersistentState(MinecraftServer server, Executor workerExecutor, LevelStorage.Session session, ServerWorldProperties properties, RegistryKey<World> worldKey, DimensionOptions dimensionOptions, WorldGenerationProgressListener worldGenerationProgressListener, boolean debugWorld, long seed, List<Spawner> spawners, boolean shouldTickTime, RandomSequencesState randomSequencesState, CallbackInfo ci) {
```
LlamaLad7: Leaving out the args would make this more stable.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
```diff
@@ -0,0 +1,90 @@
...
+	@Override
+	@Nullable
+	public <T> T getAttached(AttachmentType<T> type) {
+		return fabric_dataAttachments == null ? null : (T) fabric_dataAttachments.get(type);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	@Nullable
+	public <T> T setAttached(AttachmentType<T> type, @Nullable T value) {
```
LlamaLad7, line 51: Does this need any kind of synchronization?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
```diff
@@ -0,0 +1,51 @@
...
+	@Inject(
+			at = @At(value = "INVOKE", target = "net/minecraft/block/entity/BlockEntity.readNbt(Lnet/minecraft/nbt/NbtCompound;)V"),
+			method = "method_17897" // lambda body in BlockEntity#createFromNbt
+	)
+	private static void injectReadNbt(NbtCompound nbt, String id, BlockEntity blockEntity, CallbackInfoReturnable<BlockEntity> cir) {
```
LlamaLad7: Needs a better name like `readBlockEntityAttachments`. Same applies to most other handler methods in this PR.

LlamaLad7 (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.world.PersistentState;
+
+/**
+ * Backing storage for server-side world attachments.
+ * Thanks to custom [.isDirty] logic, the file is only written if something needs to be persisted.
+ * @author Technici4n
```
Technici4n: No author annotations are needed in FAPI. Even if it's me ;)

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentSerializingImpl.java
maityyy: Maybe use yarn `Util::getResult`/mojmap `Util::getOrThrow` like vanilla and [`FabricRecipeProvider::run`](https://github.com/FabricMC/fabric/blob/1.20.4/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricRecipeProvider.java#L99)?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
Syst3ms: I copied it from Technici4an, I honestly don't get what this does anyway

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
Syst3ms: oh aight

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentSerializingImpl.java
Syst3ms: Whatever the error looks like, I have changed the implementation to display it along with a warning that the attachment has been skipped altogether.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
Syst3ms: That's the convention for JSON, so I guess I'll change it accordingly

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
Syst3ms: Hmmm, I guess a ThreadLocal wouldn't hurt, if only for singleplayer client-server syncing. I'm a bit unexperienced with synchronized code though.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java
```diff
@@ -0,0 +1,95 @@
...
+			Objects.requireNonNull(serializer, "codec cannot be null");
+
+			this.codec = serializer;
+			return this;
+		}
+
+		@Override
+		public AttachmentType<A> buildAndRegister(Identifier id) {
+			if (codec == null && persistent) {
+				throw new IllegalArgumentException("A persistent must have an associated codec");
```
Syst3ms: tyop lol

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
Syst3ms: After some discussion, it has been established that client/server threads won't pose an issue (they operate on separate targets), and that FAPI (much like vanilla) shouldn't bother with thread-safety in general.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
Syst3ms: Testing update: as the comment on isDirty indicates, this saves data if there are persistent attachments OR if persistent attachments have been read previously. I believe the latter condition is necessary so removed persistent attachments get deleted properly. If all of the attachments are non-persistent on a fresh save file, nothing is saved.

Perhaps there is a way to make it so the file is properly deleted if no persistent attachments remain, but this doesn't seem very necessary.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+	 * @param id    the identifier of this attachment
+	 * @param codec the codec used for serialization
+	 * @param <A>   the type of attached data
+	 * @return the registered {@link AttachmentType} instance
+	 */
+	public static <A> AttachmentType<A> createPersistent(Identifier id, Codec<A> codec) {
+		Objects.requireNonNull(id, "identifier cannot be null");
+		Objects.requireNonNull(codec, "codec cannot be null");
+
+		return AttachmentRegistry.<A>builder().persistent(true).codec(codec).buildAndRegister(id);
```
maityyy: Hm, maybe then `Builder::persistent(Boolean)` should be replaced with `Builder::persistent(Codec)`? Is there any point in asking for boolean? Moreover, [it is impossible to `Builder::build`](https://github.com/FabricMC/fabric/blob/19d21da1fe3cbc091a3122ce347978d2e81ff73c/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java#L86-L88) when the type is persistent but don't have a codec. Maybe even `Builder::codec` should be renamed to "serializable" or something

maityyy:
Just a random idea, maybe it's stupid: maybe the initializer should not be `Supplier<T>` but `Function<*AttachmentTarget*, T>`? (`<Entity, T>`, `<Chunk, T>`, etc.). I think this might be useful for entities: `entity -> entity.ticksFrozen + entity.fireTicks`

Syst3ms:
Unfortunately that can't be correctly typed because you'd need a generic self-type. Usually you'd implement it by declaring `AttachmentTarget<S extends AttachmentTarget<S>>` and then using `S`, but injected interfaces cannot have generics. Factoring in the target type in the type system more generally was in the original PR, but it's a generic mess.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
Syst3ms: Originally, an independent codec argument was added because we thought it would be useful for syncing. Syncing has been removed, but having a codec on hand might still be useful in the future. That being said, I agree the current way is a bit of boilerplate, I'll ask about what can be done.

maityyy:
I didn't actually think about `AttachmentTarget<S extends AttachmentTarget<S>` since `S` will always `implements AttachmentTarget` anyway (Java don't have Higher-Kinded Types sadly), but yeah, that's not the problem...

Syst3ms:
I have added a convenience `Builder#persistent(Codec)`, but *just in case* it is deemed useful to have a codec for some other purpose, I left persistence and codecs decoupled. Though, I don't think the future syncing API will need one outside of convenience methods. If anyone wants to be less conservative and commit to "codec == persistence", I can make that change however.

EDIT: I have committed to this change after discussing it

maityyy (approved):

modmuss50 (approved):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.world.PersistentState;
+
+/**
+ * Backing storage for server-side world attachments.
+ * Thanks to custom [.isDirty] logic, the file is only written if something needs to be persisted.
```
Technici4n: Nit: you probably want `{@code ...}` or `{@link #isDirty}`

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentPersistentState.java
Technici4n: `NBT_ATTACHMENT_KEY` is `fabric:attachments`. Should this be `fabric_attachments`? I think it's fine as is, just commenting.

IMB11 (approved):
Looking forward to using this instead of CCA! Very nice.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package net.fabricmc.fabric.impl.attachment;
+
+import net.minecraft.nbt.NbtCompound;
+
+import net.fabricmc.fabric.api.attachment.v1.AttachmentTarget;
+
+public interface AttachmentTargetImpl extends AttachmentTarget {
+	default void fabric_writeAttachmentsToNbt(NbtCompound nbt) {
```
Technici4n, line 56: We might want to move this to the public API. (For ItemStacks, it is important to have as part of the API at least - maybe less so for the other types). I think there was some use case for entities at least but I don't remember the details.

settings.gradle
```diff
@@ -64,3 +65,4 @@ include 'deprecated:fabric-models-v0'
 include 'deprecated:fabric-renderer-registries-v1'
 include 'deprecated:fabric-rendering-data-attachment-v1'
 include 'deprecated:fabric-rendering-v0'
+
```
Technici4n: Note: extra newline

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -0,0 +1,73 @@
...
+ * An attachment allows "attaching" arbitrary data to various game objects (entities, block entities, worlds and chunks at the moment).
+ * Use the methods provided in {@link AttachmentRegistry} to create and register attachments.
+ *
+ * <p>Attachments can optionally be made to persist between restarts using a provided {@link Codec}.</p>
+ *
+ * @param <A> type of the attached data. It is encouraged for this to be an immutable type.
+ */
+@ApiStatus.NonExtendable
+@ApiStatus.Experimental
+public interface AttachmentType<A> {
```
Technici4n, line 54: I think that more attention should be given to entity copies.
1. For NeoForge, I added a `copyOnDeath` boolean to optionally copy a serializable attachment on player or entity death.
2. We should automatically copy serializable attachments when a player is cloned due to non-death circumstances, e.g. when returning from the end.

Technici4n (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
maityyy: I agree but because I think someone will want to make their own targets for attachment, but then they will have to use the internal API

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
maityyy: ~~Why should this be automatic? I understand that this may seem convenient, but vanilla doesn't copy all the data and we have `AFTER_RESPAWN`/`COPY_FROM`/`AFTER_ENTITY/PLAYER_CHANGE_WORLD`~~

EDIT: Okay, I just checked the vanilla code and it really copies all the data, but only in one case: when teleporting to the end after beating the game

maityyy: And yes, attachments don't only apply to entities

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,206 @@
...
+	/**
+	 * Sets the data associated with the given {@link AttachmentType}. Passing {@code null} removes the data.
+	 *
+	 * @param type  the attachment type
+	 * @param value the new value
+	 * @param <A>   the type of the data
+	 * @return the previous data
+	 */
+	@Nullable
+	default <A> A setAttached(AttachmentType<A> type, @Nullable A value) {
```
Technici4n, line 188: The data will not be saved automatically for block entities and chunks. Any change to attached data for BEs or chunks needs to mark the target for saving (`setChanged()` and `setUnsaved(true)` respectively). Additionally, `setAttached` should be overriden by chunks and block entities to call `setChanged` and `setUnsaved` as appropriate.

From the NF docs:
image

Technici4n (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
Syst3ms: Attachments aren't supported on ItemStacks in this PR, by your recommendation.

Syst3ms:
Added all the entity copy stuff, with an extra customizable flag for player respawn. The API automatically copies attachments when returning from the End, as well as entity conversion (like zombie → drowned). I'll write a unit test to check this behaves as intended.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerPlayerEntityMixin.java
```diff
@@ -0,0 +1,18 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.server.network.ServerPlayerEntity;
+
+import net.fabricmc.fabric.impl.attachment.AttachmentTargetImpl;
+
+@Mixin(ServerPlayerEntity.class)
+abstract class ServerPlayerEntityMixin implements AttachmentTargetImpl {
+	@Inject(method = "copyFrom", at = @At("TAIL"))
```
maityyy: I will assume that this should be `ServerPlayerEvents.COPY_FROM` so as not to duplicate mixins and simplify FAPI updates for modmuss :D

Syst3ms: Sure thing.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentEntrypoint.java
```diff
@@ -0,0 +1,13 @@
+package net.fabricmc.fabric.impl.attachment;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.entity.event.v1.ServerPlayerEvents;
+
+public class AttachmentEntrypoint implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		ServerPlayerEvents.COPY_FROM.register((oldPlayer, newPlayer, alive) ->
+				AttachmentTargetImpl.copyOnRespawn((AttachmentTargetImpl) oldPlayer, (AttachmentTargetImpl) newPlayer, alive)
```
maityyy: Maybe then iface inject `AttachmentTargetImpl` instead of `AttachmentTarget`?
edit: Not important actually

Syst3ms: No, as its name suggests, `AttachmentTargetImpl` is implementation-only and shouldn't be exposed to users with interface injection. This cast is fine, as this call is made in impl as well.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
```diff
@@ -47,6 +49,15 @@ public <T> T getAttached(AttachmentType<T> type) {
...
 	@Nullable
 	public <T> T setAttached(AttachmentType<T> type, @Nullable T value) {
+		// Extremely inelegant, but the only alternative is separating out these two mixins and duplicating code
+		Object thisObject = this;
+
+		if (thisObject instanceof BlockEntity) {
+			((BlockEntity) thisObject).markDirty();
+		} else if (thisObject instanceof WorldChunk) {
+			((Chunk) thisObject).setNeedsSaving(true);
+		}
```
Technici4n, line 59: Users also need to do this manually for mutable data. This should be in the javadoc! Personally I would include target-specific documentation in `AttachmentTarget`. See for example what I did for NeoForge in `AttachmentType` where I list the specifics of every single attachment target.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -114,6 +114,13 @@ public interface Builder<A> {
 		 */
 		Builder<A> persistent(Codec<A> codec);
 
+		/**
+		 * Declares that when a player dies and respawns, the attachments corresponding of this type should remain.
+		 *
+		 * @return the builder
+		 */
+		Builder<A> copyOnPlayerRespawn();
```
Technici4n: This should only apply to serializable attachments.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
```diff
@@ -16,11 +16,36 @@
...
+		Map<AttachmentType<?>, ?> attachments = original.fabric_getAttachments();
+
+		for (Map.Entry<AttachmentType<?>, ?> entry : attachments.entrySet()) {
+			AttachmentType<Object> type = (AttachmentType<Object>) entry.getKey();
+
+			if (alive || type.copyOnPlayerRespawn()) {
+				target.setAttached(type, entry.getValue());
+			}
+		}
+	}
```
Technici4n: This should only be done for serializable attachments, and it should be done by serializing and deserializing to make sure that the instances remain separate. (For example the data might hold a reference to its target).

This round-trip is why we can only copy serializable attachments in the first place.

Technici4n (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
Syst3ms: Why should it? Imagine data that records something about a player's current play session. It should clearly persist over respawns, but by definition it should be reset on disconnect.

EDIT: hadn't seen your next comment. That is somewhat convincing.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentTargetImpl.java
Syst3ms: Wait, what about cases where we always copy (like return from end or mob conversion), wouldn't that mean everything is required to be serializable? Or that non-serializable things get lost on entity copy even in these very simple cases?

Syst3ms:
I overhauled the API and mechanics for copying entity attachments. There is a new `EntityCopyHandler` functional subinterface of `AttachmentType` that is used to handle the copying properly, as to avoid references to the old `Entity` instance when copying.

However, specifying such a handler manually is unnecessary most of the time, and one will automatically be deduced if a `Codec` is provided. Also, since there are now two independent uses for codecs, the old "codec method + flag methods" design has been reinstated, with convenience shorthands.

Additionally, the `copyOnDeath` (name changed) flag now also includes entity conversion. I was rather on the fence about whether that should count as a "respawn", but I decided to align with NF for the sake of convenience. Only teleportation across worlds is universally desirable.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -66,13 +65,40 @@ default boolean isPersistent() {
...
+
+	/**
+	 * A functional interface to handle copying attachment data from an old entity instance to a new one, for example
+	 * during player respawn, entity conversion, or when an entity is teleported between worlds.
+	 *
+	 * <p>It is <i>imperative</i> that the data attached to the new entity doesn't hold a reference to the old entity,
+	 * as that can and will break in unexpected ways.</p>
+	 */
+	@FunctionalInterface
+	interface EntityCopyHandler<T> {
```
modmuss50: I dont think this is needed, the data should not be holding onto refrences beyond what it plans to save. Lets discuss this before you revert it.

Syst3ms:
Discussion was had, and the complex previous change was reverted, and replaced with a javadoc warning about keeping entity instances around. I'd also missed copying on cross-world teleportation. 

By the way, I was surprised to find there is no event for mob conversion. For easier maintenance, maybe this could be added to the event API?

modmuss50 (approved):

modmuss50:
> By the way, I was surprised to find there is no event for mob conversion. For easier maintenance, maybe this could be added to the event API?

Yeah, thats a good idea. 👍

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+	public static <A> Builder<A> builder() {
+		return AttachmentRegistryImpl.builder();
+	}
+
+	/**
+	 * A builder for creating {@link AttachmentType}s with finer control over their properties.
+	 *
+	 * @param <A> the type of the attached data
+	 */
+	public interface Builder<A> {
```
apple502j, line 108: nit: add NonExtendable

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+
+	/**
+	 * Gets the data associated with the given {@link AttachmentType}, throwing an exception if it doesn't yet exist.
+	 *
+	 * @param type the attachment type
+	 * @param <A>  the type of the data
+	 * @return the attached data
+	 */
+	default <A> A getAttachedOrThrow(AttachmentType<A> type) {
+		return Objects.requireNonNull(getAttached(type));
```
apple502j: Should give a more concise message.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+	 * @param <A>  the type of the data
+	 * @return the attached data
+	 */
+	@Nullable
+	default <A> A getAttached(AttachmentType<A> type) {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * Gets the data associated with the given {@link AttachmentType}, throwing an exception if it doesn't yet exist.
```
apple502j: Should document which exception is thrown.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+	 * Gets the data associated with the given {@link AttachmentType}, or returns the provided default value if it doesn't exist.
+	 * Unlike {@link #getAttachedOrCreate(AttachmentType, Supplier)}, this doesn't initialize the attachment with the default value.
+	 *
+	 * @param type         the attachment type
+	 * @param defaultValue the default value to use as fallback
+	 * @param <A>          the type of the attached data
+	 * @return the attached data, or the default value
+	 */
+	default <A> A getAttachedOrElse(AttachmentType<A> type, A defaultValue) {
+		Objects.requireNonNull(defaultValue, "default value cannot be null");
```
apple502j: Why? `Optional#orElse` for example accepts null. If this is for typing purposes, we can use Contract.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+	 * argument if it doesn't exist. Unlike {@link #getAttachedOrCreate(AttachmentType, Supplier)}, this doesn't
+	 * initialize the attachment with the default value.
+	 *
+	 * @param type         the attachment type
+	 * @param defaultValue the default value supplier to use as fallback
+	 * @param <A>          the type of the attached data
+	 * @return the attached data, or the default value
+	 */
+	default <A> A getAttachedOrGet(AttachmentType<A> type, Supplier<A> defaultValue) {
+		Objects.requireNonNull(defaultValue, "default value cannot be null");
```
apple502j: This message is inaccurate; this allows the supplier to return null (should be documented as such), but the supplier itself cannot be.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+	 * @param <A>   the type of the data
+	 * @return the previous data
+	 */
+	@Nullable
+	default <A> A setAttached(AttachmentType<A> type, @Nullable A value) {
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * Tests whether the given {@link AttachmentType} has any associated data.
```
apple502j: Please document the behavior for defaulted ones. Do they return true at all times or not?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -0,0 +1,220 @@
...
+	default <A> A removeAttached(AttachmentType<A> type) {
+		return setAttached(type, null);
+	}
+
+	/**
+	 * Modifies the data associated with the given {@link AttachmentType}. Functionally the same as calling {@link #getAttached(AttachmentType)},
+	 * applying the modifier, then calling {@link #setAttached(AttachmentType, Object)} with the result.
+	 *
+	 * @param type     the attachment type
+	 * @param modifier the operation to apply to the current data
```
apple502j: The modifier, importantly, has to accept null for initialization.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java
```diff
@@ -0,0 +1,91 @@
...
+	public static <A> AttachmentRegistry.Builder<A> builder() {
+		return new BuilderImpl<>();
+	}
+
+	public static class BuilderImpl<A> implements AttachmentRegistry.Builder<A> {
+		@Nullable
+		private Supplier<A> defaultInitializer = null;
+		@Nullable
+		private Codec<A> persistenceCodec = null;
+		private boolean copyOnPlayerRespawn = false;
```
apple502j: Needs rename.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.nbt.NbtCompound;
+
+import net.fabricmc.fabric.impl.attachment.AttachmentTargetImpl;
+
+@Mixin(BlockEntity.class)
+abstract class BlockEntityMixin implements AttachmentTargetImpl {
```
apple502j, line 30: Some mods may attach data that should not be synced to clients. Since some BEs (but not all) sync the NBT to clients on chunk load, this needs further documentation.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
```diff
@@ -0,0 +1,43 @@
...
+@Mixin(ServerWorld.class)
+abstract class ServerWorldMixin {
+	@Inject(at = @At("TAIL"), method = "<init>")
+	private void createAttachmentsPersistentState(CallbackInfo ci) {
+		// Force persistent state creation
+		ServerWorld world = (ServerWorld) (Object) this;
+		var type = new PersistentState.Type<>(
+				() -> new AttachmentPersistentState(world),
+				nbt -> AttachmentPersistentState.read(world, nbt),
+				DataFixTypes.LEVEL
```
apple502j: This is definitely incorrect. Fortunately, if we declare dependency to Object Builder, its mixin will handle `null` argument.

apple502j (changes requested):
Glad this wasn't merged yet. Sorry for late review.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
Syst3ms: Do you reckon Contract could be added to some of the others?

EDIT: nevermind, I just fully got what you meant.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/ServerWorldMixin.java
Syst3ms: I didn't write this code and frankly I don't know how it works. Why is it broken?

EDIT: I got it, you'd want no datafixing, and the object builder mixin allows that by passing `null`.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
Syst3ms: If a mod wants to prevent some data from being synced, how should it go? Could something be provided by the API to make it easier?

EDIT: I reckon that since this version has no syncing behavior whatsoever, it would make the most sense to just up and remove any attachments from synced data.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
```diff
@@ -0,0 +1,64 @@
...
+	)
+	private void writeBlockEntityAttachments(CallbackInfoReturnable<NbtCompound> cir) {
+		this.fabric_writeAttachmentsToNbt(cir.getReturnValue());
+	}
+
+	@ModifyReturnValue(
+			method = "toInitialChunkDataNbt",
+			at = @At("RETURN")
+	)
+	private NbtCompound removeAttachmentsForChunkLoadSync(NbtCompound nbt) {
```
apple502j: Does this actually work?

apple502j (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java
Syst3ms: I'll write a test for it. Is it enough to check that the packet nbt contains no attachments?

Syst3ms: Writing the gametest was a good idea, because this mixin does, in fact, not work whatsoever.

apple502j (approved):

Syst3ms:
Two notes:
- It was brought to my attention that on respawn, client players are recreated and copied, but not through any of handled copy methods, so a mixin was added in [c37c705](https://github.com/FabricMC/fabric/pull/3476/commits/c37c705d9a5f7613286d42598780d325f36973f5) to take care of it. Unlike general mob conversion, this is completely covered by the mixin. Nevertheless, is it worth adding to the event API?
- Currently, all mob conversions are treated as deaths for attachment copy, which makes use of the copyOnDeath flag. This makes sense for most conversions, but what about tadpoles growing up? Answer: we'll just treat it the same.

modmuss50 (approved):

PR #3477, opened 2023-12-21
De-hardcode shears (again)

### LINKED ISSUE #1226: Shears not completely de-hardcoded

AnAwesomGuy:
Adds a boolean method to `FabricItem` and `FabricItemStack`, `isShears`. It functions as a stack-aware check that allows items to function as shears. By default, the method returns true if the item is either an instance of `ShearsItem` or in the `fabric:shears` tag.

If a stack `isShears`, it is able to shear pumpkins, de-arm tripwire, be enchanted with efficiency, have `ShearsDispenserBehavior` (if there isn't already a `DispenserBehavior` for the item registered), and shear snow golems, sheep, and mooshrooms.

This also mixes into `MatchToolLootCondition` and adds a [custom predicate](https://github.com/FabricMC/fabric/pull/3493) to it if its predicate contains shears. The custom predicate is a boolean value registered under `fabric:allow_shears` and defaults to `true`. If the value is `false`, the mixin will not add the custom predicate even if the predicate contains shears.

Supersedes/closes #1287 and closes #1226. Also depends on/includes #3493, so either merge that first or just merge this PR.


ZestyBlaze:
Aren't tags saved under `c:shears` not `fabric:shears`? You might want to ammend the comments to correct that if that's the case

deirn:
Apply the license header with `gradlew spotlessApply`. You also need to fix the style issues.

Juuxel:
@ZestyBlaze Generally, `fabric` tags are used for behaviour and `c` tags just for categorisation. (Compare with FAPI's mining level + mineable tags etc)

AnAwesomGuy:
this works from 23w32a onwards (`MatchToolLootCondition` became a record with an optional in that version)

for below that, you would remove `DirectRegistryEntryListAccessor<T>` and replace `MatchToolLootConditionMixin` with something like [this](https://github.com/AnAwesomGuy/wsmlmb/blob/b958f56e92ae1cd7c013b3c85df9b8987edb8e4f/src/main/java/net/anawesomguy/wsmlmb/mixin/shears/MatchToolLootConditionMixin.java#L25-L61). that will work down until somewhere between 1.18.1 and 1.18.2, when `TagKey<T>`s were added

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,57 @@
...
+			@SuppressWarnings("deprecation")
+			RegistryEntry<Item> entry = item.getRegistryEntry();
+
+			if (entry.isIn(FABRIC_SHEARS) || item instanceof ShearsItem) {
+				builder.add(entry);
+			}
+		}
+
+		SHEARS = builder.build();
+	}
```
deirn: This will fail if there are items that get registered after this class is loaded.

You also need to listen to [`RegistryEntryAddedCallback`](https://github.com/FabricMC/fabric/blob/1.20.4/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/RegistryEntryAddedCallback.java) and [`RegistryEntryRemovedCallback`](https://github.com/FabricMC/fabric/blob/1.20.4/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/RegistryEntryRemovedCallback.java)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
```diff
@@ -0,0 +1,72 @@
...
+	@Unique
+	private static final List<ItemPredicate> MATCH_TOOL_PREDICATES = new ArrayList<>();
+
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void shearsLoot(CallbackInfo ci) {
+		// allows anything in fabric:shears to mine grass (and other stuff) and it will drop
+		((MatchToolLootCondition) (Object) this).predicate().ifPresent(MATCH_TOOL_PREDICATES::add);
+	}
+
+	static {
```
deirn: Not really sure about using `static` block on Mixin, whether it overwrites the target class' or not. Just to be safe inject to `<clinit>` instead.

Syst3ms: Mixin `static` blocks are appended at the end of the class's, it is not overwritten. However, this has mostly niche uses and injecting in clinit often works just as well if not better.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/BlockInteractShearsMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.fabricmc.fabric.impl.item.ShearsHelper;
+
+@Mixin({PumpkinBlock.class, BeehiveBlock.class})
+public abstract class BlockInteractShearsMixin {
+	@ModifyExpressionValue(at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isOf(Lnet/minecraft/item/Item;)Z"), method = "onUse")
+	private boolean isShears(boolean original, @Local ItemStack stack) {
+		// allows anything in fabric:shears to shear pumpkins and beehives
+		return original || ShearsHelper.isShears(stack);
+	}
```
LlamaLad7: Use a `@WrapOperation` instead of capturing locals.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/DispenserBlockMixin.java
```diff
@@ -0,0 +1,53 @@
...
+import net.fabricmc.fabric.impl.item.ShearsHelper;
+
+@Mixin(DispenserBlock.class)
+public abstract class DispenserBlockMixin {
+	@Shadow
+	@Final
+	private static Map<Item, DispenserBehavior> BEHAVIORS;
+
+	@ModifyReturnValue(at = @At("TAIL"), method = "getBehaviorForItem")
+	private DispenserBehavior registerShearsBehavior(DispenserBehavior original, @Local ItemStack stack) {
```
LlamaLad7: Don't use `@Local`, `@ModifyReturnValue` can get the enclosing method's args natively as explained in its docs.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/DispenserBlockMixin.java
```diff
@@ -0,0 +1,53 @@
...
+		// allows anything in fabric:shears to have the dispenser behavior of shears,
+		// but only if there isn't a dispenser behavior already registered
+		if (!BEHAVIORS.containsKey(stack.getItem())
+				&& ShearsHelper.isShears(stack)) {
+			return new ShearsDispenserBehavior();
```
LlamaLad7: Should this not be cached?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/EfficiencyEnchantmentMixin.java
```diff
@@ -0,0 +1,36 @@
...
+
+import net.fabricmc.fabric.impl.item.ShearsHelper;
+
+@Mixin(EfficiencyEnchantment.class)
+public abstract class EfficiencyEnchantmentMixin {
+	@ModifyExpressionValue(at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isOf(Lnet/minecraft/item/Item;)Z"), method = "isAcceptableItem")
+	private boolean isShears(boolean original, @Local ItemStack stack) {
+		// allows anything in fabric:shears to be enchanted with efficiency
+		return original || ShearsHelper.isShears(stack);
+	}
```
LlamaLad7: Use a `@WrapOperation` instead of capturing locals.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/EntityInteractShearsMixin.java
```diff
@@ -0,0 +1,38 @@
...
+
+import net.fabricmc.fabric.impl.item.ShearsHelper;
+
+@Mixin({SheepEntity.class, SnowGolemEntity.class, MooshroomEntity.class})
+public abstract class EntityInteractShearsMixin {
+	@ModifyExpressionValue(at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isOf(Lnet/minecraft/item/Item;)Z"), method = "interactMob")
+	private boolean isShears(boolean original, @Local ItemStack stack) {
+		// allows anything in fabric:shears to shear sheep and snow golems
+		return original || ShearsHelper.isShears(stack);
+	}
```
LlamaLad7: Use a `@WrapOperation` instead of capturing locals.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
```diff
@@ -0,0 +1,72 @@
...
+
+@Mixin(MatchToolLootCondition.class)
+public abstract class MatchToolLootConditionMixin implements LootCondition {
+	@Unique
+	private static final List<ItemPredicate> MATCH_TOOL_PREDICATES = new ArrayList<>();
+
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void shearsLoot(CallbackInfo ci) {
+		// allows anything in fabric:shears to mine grass (and other stuff) and it will drop
+		((MatchToolLootCondition) (Object) this).predicate().ifPresent(MATCH_TOOL_PREDICATES::add);
```
LlamaLad7: `Shadow`, don't cast. See <https://github.com/SpongePowered/Mixin/wiki/Introduction-to-Mixins---Understanding-Mixin-Architecture>.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
LlamaLad7: A static block is in almost all cases preferable to an injection. For example when re-assigning fields you don't need to make them non-final. That doesn't apply here, and they would function the same, but a static block is still much cleaner.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/TripwireBlockMixin.java
```diff
@@ -0,0 +1,36 @@
...
+
+import net.fabricmc.fabric.api.tag.convention.v1.ConventionalItemTags;
+
+@Mixin(TripwireBlock.class)
+public abstract class TripwireBlockMixin {
+	@ModifyExpressionValue(at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;isOf(Lnet/minecraft/item/Item;)Z"), method = "onBreak")
+	private boolean isShears(boolean original, @Local PlayerEntity player) {
+		// allows anything in fabric:shears to silently break tripwire (string)
+		return original || player.getMainHandStack().isIn(ConventionalItemTags.SHEARS);
+	}
```
LlamaLad7: Use a `@WrapOperation` instead of capturing locals, and especially instead of recomputing the receiver.

LlamaLad7 (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
AnAwesomGuy: so, should i use a static initializer or an inject into <clinit>?
too many opinions lol

modmuss50: I wouldnt put this in the static initializer or clinit, as its registering the event handler based on classload.

I would move this to a mod inititliser entrypoint.

AnAwesomGuy: I think it would be better to not move it into another class, and I also don't think that registering an event early will do any harm.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
AnAwesomGuy: Fixed this by putting it into the tags loaded event. All the items definitely should be registered by then.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+						builder.addAll(accessor.getEntries());
+						builder.addAll(shears);
+						accessor.setEntries(builder.build());
+						accessor.setEntrySet(null);
+					}
+				}
+			}
+
+			MATCH_TOOL_PREDICATES.clear();
+		});
```
modmuss50: Some comments on what this does/why its needed would be a helpful. Im a little lost just looking at it.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public static boolean isShears(ItemStack stack) {
+		return stack.isIn(FABRIC_SHEARS) || stack.getItem() instanceof ShearsItem;
+	}
+
+	@Override
+	@SuppressWarnings("deprecation")
+	public void onInitialize() {
+		CommonLifecycleEvents.TAGS_LOADED.register((registries, client) -> {
+			if (!client) {
```
modmuss50: Nit: return early, instead of going more blocks deep:
```
if (client) {
   return;
}
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+	public static boolean isShears(ItemStack stack, Item item) {
+		return item == Items.SHEARS && isShears(stack);
+	}
+
+	public static boolean isShears(ItemStack stack) {
+		return stack.isIn(FABRIC_SHEARS) || stack.getItem() instanceof ShearsItem;
+	}
+
+	@Override
+	@SuppressWarnings("deprecation")
```
modmuss50: Question: Whats deprecated?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+public final class ShearsHelper implements ModInitializer {
+	public static final TagKey<Item> FABRIC_SHEARS = TagRegistration.ITEM_TAG_REGISTRATION.registerFabric("shears");
+	public static final List<ItemPredicate> MATCH_TOOL_PREDICATES = new ArrayList<>();
+	private static final Set<RegistryEntry<Item>> SHEARS_ITEMS = new HashSet<>();
+
+	public static boolean isShears(ItemStack stack, Item item) {
+		return item == Items.SHEARS && isShears(stack);
+	}
+
+	public static boolean isShears(ItemStack stack) {
```
modmuss50: Should this be on FabricItem to allow modders to be able to adjust this logic based on the item stack? This would enable powered shears or something?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+import net.fabricmc.fabric.api.event.lifecycle.v1.CommonLifecycleEvents;
+import net.fabricmc.fabric.impl.tag.convention.TagRegistration;
+import net.fabricmc.fabric.mixin.item.shears.accessors.DirectRegistryEntryListAccessor;
+
+public final class ShearsHelper implements ModInitializer {
+	public static final TagKey<Item> FABRIC_SHEARS = TagRegistration.ITEM_TAG_REGISTRATION.registerFabric("shears");
+	public static final List<ItemPredicate> MATCH_TOOL_PREDICATES = new ArrayList<>();
+	private static final Set<RegistryEntry<Item>> SHEARS_ITEMS = new HashSet<>();
+
+	public static boolean isShears(ItemStack stack, Item item) {
```
modmuss50: Whats the need for this, the ItemStack contains the item.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+import net.minecraft.registry.entry.RegistryEntry;
+import net.minecraft.registry.tag.TagKey;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.lifecycle.v1.CommonLifecycleEvents;
+import net.fabricmc.fabric.impl.tag.convention.TagRegistration;
+import net.fabricmc.fabric.mixin.item.shears.accessors.DirectRegistryEntryListAccessor;
+
+public final class ShearsHelper implements ModInitializer {
+	public static final TagKey<Item> FABRIC_SHEARS = TagRegistration.ITEM_TAG_REGISTRATION.registerFabric("shears");
```
modmuss50: Any new tags need to be defined in a public API.

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/ShearsTest.java
```diff
@@ -0,0 +1,37 @@
...
+import net.fabricmc.fabric.api.item.v1.FabricItemSettings;
+
+public class ShearsTest implements ModInitializer {
+	public static final Item REAL_SHEARS = new ShearsItem(new FabricItemSettings().maxDamage(38)); // to show that ShearsItem will work
+	public static final Item FAKE_SHEARS = new Item(new FabricItemSettings().maxDamage(38)); // to show that anything in fabric:shears will work
+
+	@Override
+	public void onInitialize() {
+		Registry.register(Registries.ITEM, new Identifier("fabric-item-api-v1-testmod", "real_shears"), REAL_SHEARS);
+		Registry.register(Registries.ITEM, new Identifier("fabric-item-api-v1-testmod", "fake_shears"), FAKE_SHEARS);
```
modmuss50: Some form of automated tests would be very nice. Either similar to some other tesst mods that run some basic checks on first tick, or maybe even a junit unit test?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
modmuss50: Would it be wroth passing in some sort of context, allowing shears that only work in certain situations? Im not quite sure on a nice API though. Might need further discussion on github.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,88 @@
...
+	public static final List<ItemPredicate> MATCH_TOOL_PREDICATES = new ArrayList<>();
+	private static final Set<RegistryEntry<Item>> SHEARS_ITEMS = new HashSet<>();
```
modmuss50: Why does this need to be a static field, its only used in the TAGS_LOADED.register callback?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Mixin(MatchToolLootCondition.class)
+public abstract class MatchToolLootConditionMixin implements LootCondition {
+	@Shadow
+	public abstract Optional<ItemPredicate> predicate();
+
+	@Inject(at = @At("RETURN"), method = "<init>")
+	private void shearsLoot(CallbackInfo ci) {
+		// allows anything in fabric:shears to mine grass (and other stuff) and it will drop
+		predicate().ifPresent(ShearsHelper.MATCH_TOOL_PREDICATES::add);
```
modmuss50: Just reading the code here without any context or understanding of the vanilla code, im a little confused as it seem that shears will be added to every `MatchToolLootCondition`? Why?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/DispenserBlockMixin.java
```diff
@@ -0,0 +1,55 @@
...
+
+	@ModifyReturnValue(at = @At("TAIL"), method = "getBehaviorForItem")
+	private DispenserBehavior registerShearsBehavior(DispenserBehavior original, ItemStack stack) {
+		// allows anything in fabric:shears to have the dispenser behavior of shears,
+		// but only if there isn't a dispenser behavior already registered
+		Item item = stack.getItem();
+
+		if (ShearsHelper.isShears(stack) && !BEHAVIORS.containsKey(item)) {
+			DispenserBehavior behavior = new ShearsDispenserBehavior();
+			BEHAVIORS.put(item, behavior);
```
modmuss50: Wont this still be regsistered if the tags are reloaded to no longer contain this item as shears?

modmuss50 (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
AnAwesomGuy: `Item#getRegistryEntry()`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
AnAwesomGuy: it also checks the `item` argument is `Items.SHEARS` just in case because there are more than just the shears checks using `ItemStack#isOf` in the mixins (it shouldnt even matter since the shears checks are always first, but who knows)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
```diff
@@ -0,0 +1,94 @@
...
+				if (entryList.contains(shearsRegistryEntry)) {
+					@SuppressWarnings("unchecked") // `ItemPredicate`s' RegistryEntryList should ALWAYS be Direct
+					DirectRegistryEntryListAccessor<Item> accessor = (DirectRegistryEntryListAccessor<Item>) entryList;
+					ImmutableList.Builder<RegistryEntry<Item>> builder = new ImmutableList.Builder<>();
+					builder.addAll(accessor.getEntries()); // add the original items
+					builder.addAll(shears); // then add the shears
+					accessor.setEntries(builder.build()); // set the RegistryEntryList's entries
+					accessor.setEntrySet(null); // the entryList contains an immutable Set copy of entries, so we have to clear that
+				}
+			}
```
deirn: All these comments seem unnecessary. The code already explains what it does.
Also seem to be the case with most comments.

deirn (commented):
Not sure about modifying all loot tables, we should probably only do it for vanilla.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsHelper.java
AnAwesomGuy: idk. modmuss earlier suggested adding some comments to clarify what the code does, so i just added this.

deirn: Comment what the entire snippet does, not individual lines. 
It's pretty clear what the individual lines do, it's more of what is the purpose of the snippet.

AnAwesomGuy:
> Not sure about modifying all loot tables, we should probably only do it for vanilla.

If someone could help me find a way to do that that would be amazing. Though it would be even harder now that we're moving towards making it stack-based, which means I'm mixing into `MatchToolLootCondition#test`, making it so there is basically no context about its creation.

I also don't really see why anyone would want to target only vanilla shears but ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/¯.

deirn:
The way Forge does it is to just replace the loot tables with their own. Maybe we can move the modification into a datagen instead.

AnAwesomGuy:
Then it wouldn't be stack-based, and that seems like what modmuss and some others want.

deirn:
You can make a custom predicate that checks if the item is shears with #3493 😉

AnAwesomGuy:
That looks pretty good, but how would I make my PR depend on yours?
Should I just wait for it to get merged?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsItemPredicate.java
```diff
@@ -0,0 +1,26 @@
...
+import com.mojang.serialization.Codec;
+
+import net.fabricmc.fabric.api.item.v1.CustomItemPredicate;
+
+import net.minecraft.item.ItemStack;
+
+public class ShearsItemPredicate implements CustomItemPredicate {
+	public static final ShearsItemPredicate INSTANCE = new ShearsItemPredicate();
+
+	private final Codec<ShearsItemPredicate> CODEC = Codec.unit(this);
```
deirn: It should probably be a boolean codec, that way you can check if the item is not a shears.

AnAwesomGuy: I've already done that, but I can't push the changes I've made, as GitHub on my laptop doesn't work in China and my VPN hasn't been working these few days. (GitHub mobile seems to work tho)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsItemPredicate.java
```diff
@@ -0,0 +1,57 @@
...
+	 *
+	 * @see net.fabricmc.fabric.api.item.v1.FabricItem#isShears
+	 */
+	ALLOW,
+	/**
+	 * Disallows Fabric API to replace any occurrences of {@link net.minecraft.item.Items#SHEARS minecraft:shears} with a custom predicate for stack-aware shears.
+	 *
+	 * @see net.fabricmc.fabric.api.item.v1.FabricItem#isShears
+	 */
+	DENY;
```
deirn, line 40: I'm a little confused by this, what's with the allow and deny? Why can't it be a boolean that checks if it is shears or not?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
```diff
@@ -0,0 +1,65 @@
...
+			CustomItemPredicate[] custom = predicate.custom().toArray(new CustomItemPredicate[0]);
+			ImmutableList.Builder<CustomItemPredicate> builder = ImmutableList.builderWithExpectedSize(custom.length + 1);
+
+			for (CustomItemPredicate customPredicate : custom) {
+				if (customPredicate == ShearsItemPredicate.DENY) return;
+				builder.add(custom);
+			}
+
+			((ItemPredicateExtensions) (Object) predicate).fabric_setCustom(builder.add(ShearsItemPredicate.ALLOW).build());
+		}
```
deirn, line 63: Does this make it replace all predicates with the `ShearsItemPredicate.ALLOW` by default? If anything, it should be defaults to deny.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsItemPredicate.java
AnAwesomGuy: > This also mixes into `MatchToolLootCondition` and adds a [custom predicate](https://github.com/FabricMC/fabric/pull/3493) to it if its predicate contains shears. The custom predicate is a boolean value registered under `fabric:allow_shears` and defaults to `true`. If the value is `false`, the mixin will not add the custom predicate even if the predicate contains shears.

true is `ALLOW` and false is `DENY`
this is made so mods can also target only shears using DENY if they want to for some reason

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
AnAwesomGuy: yes it does
i chose to make it default to ALLOW cuz im lazy and dont want to datagen
also i doubt anyone should want to make it only work for shears so its opt-out

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ShearsItemPredicate.java
deirn: What I had in mind is a boolean predicate, maybe called `fabric:is_shears` and `value && item.isShears()` check that entirely replaces the vanilla `items` predicate. People who want to match strictly vanilla shears can do so like before without the `deny` shenanigans.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/shears/MatchToolLootConditionMixin.java
deirn: Can't you change the conditions at runtime with [`LootTableEvents.MODIFY`](https://github.com/FabricMC/fabric/blob/6dfc9a786d41732780b9543a722a2319cabfeb97/fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java#L86)? CC: @Juuxel

cech12:
Thanks for all your work! I think it would be great if the changes were included as I own 2 mods that add shears to Minecraft. They would benefit greatly from this. Are there any plans to bring this PR to 1.20.5? What is needed to reach that? :)

AnAwesomGuy:
it should already work on 1.20.5, but this pr is kinda in a dead place rn

cech12:
I think it needs a rework after #3493 was closed, because of the vanilla's sub predicate system.

cech12:
For porting this to 1.20.6 I think it would be easier to add a new boolean component "IS_SHEARS" or something similar and add it to vanilla shears and also to custom shears. This component could be checked in all your mixins instead of adding a new tag, a new `isShears` method and the whole custom predicate logic. That should be much cleaner and easier to use.

Edit: and don't forget the newly added functionalities: shearable Bogged entity & Wolf armor shearing

Patbox:
Adding new component would make fabric api incompatible with vanilla

AnAwesomGuy:
yes, that is why i havent ported this to 1.20.5+ yet
i could also go the old route and just add stuff to `FabricItem` (including the custom predicate) but i dont really know if thats the way to go yet

PR #3487, opened 2023-12-28
feat: add fabric-api-bom and fabric-api-catalog

jpenilla:
This adds two new publications, a [BOM](https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import) (`fabric-api-bom`) and a [version catalog](https://docs.gradle.org/current/userguide/platforms.html#sec:importing-published-catalog) (`fabric-api-catalog`).

**Example use of the BOM**

This will also work with `include` configuration in loom 1.5+
```kotlin
dependencies {
  modImplementation(platform("net.fabricmc.fabric-api:fabric-api-bom:VERSION"))
  modImplementation("net.fabricmc.fabric-api:fabric-api-lookup-api-v1") // version is from BOM
}
```

**Example use of the catalog**

`settings.gradle(.kts)`:
```kotlin
dependencyResolutionManagement {
  repositories {
    maven("https://maven.fabricmc.net/")
  }
  versionCatalogs {
    create("fabricApiLibs") {
      from("net.fabricmc.fabric-api:fabric-api-catalog:VERSION")
    }
  }
}
```

`build.gradle(.kts)`:
```kotlin
dependencies {
  modImplementation(platform(fabricApiLibs.bom)) // if you want to use the BOM too
  modImplementation(fabricApiLibs.api.lookup.api.v1)

  // deprecated modules in own group
  modImplementation(fabricApiLibs.deprecated.command.api.v1)
}
```

**Future ideas**
For modules that depend on others, we could also include a bundle in the catalog that has flattened dependencies (for use with `include`).

### DIFF
--- a/fabric-api-bom/build.gradle
+++ b/fabric-api-bom/build.gradle
@@ -1,0 +1,31 @@
+plugins {
+	id 'java-platform'
+}
+
+version = rootProject.version
+
+publishing.publications {
+	register('mavenJava', MavenPublication) {
+		from(components['javaPlatform'])
+	}
+}
+
+tasks.withType(GenerateModuleMetadata) {
+	// todo: RemoteSignJar fails when this is false (as set by parent build script)
+	enabled = true
+}
+
+dependencies {
+	constraints {
+		for (proj in rootProject.allprojects) {
+			if (proj == project) { // the bom itself
+				continue
+			}
+			if (proj.name == 'fabric-api-catalog') {
+				continue
+			}
+
+			api(project(proj.path))
+		}
+	}
+}
--- a/fabric-api-catalog/build.gradle
+++ b/fabric-api-catalog/build.gradle
@@ -6,7 +6,7 @@ version = rootProject.version
 
 publishing.publications {
 	register('mavenJava', MavenPublication) {
-		from(components['versionCatalog'])
+		from components.versionCatalog
 	}
 }
 
@@ -21,7 +21,7 @@ tasks.register('configureCatalog') {
 		doConfigureCatalog()
 	}
 }
-tasks.generateCatalogAsToml {
+tasks.named('generateCatalogAsToml') {
 	dependsOn('configureCatalog')
 }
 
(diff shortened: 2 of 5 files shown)


fabric-api-bom/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+version = rootProject.version
+
+publishing.publications {
+	register('mavenJava', MavenPublication) {
+		from(components['javaPlatform'])
+	}
+}
+
+tasks.withType(GenerateModuleMetadata) {
+	// todo: RemoteSignJar fails when this is false (as set by parent build script)
```
modmuss50, line 14: Is this something you want me to look into? I forget the reason why I disabled it on the parent projects as well.

modmuss50 (approved):
This is super cool, especially the version catalog, I wasnt aware that was a thing :)

maityyy:
Maybe mention this in README.md?

fabric-api-bom/build.gradle
jpenilla: I think it's probably fine, I didn't notice any issues resulting from the metadata being present. Just left the notes as it does differ from the configuration done to the other subprojects.

fabric-api-catalog/build.gradle
```diff
@@ -0,0 +1,57 @@
+plugins {
+	id 'version-catalog'
+}
+
+version = rootProject.version
+
+publishing.publications {
+	register('mavenJava', MavenPublication) {
+		from(components['versionCatalog'])
```
Juuxel:
```suggestion
		from components.versionCatalog
```
More conventional in the Groovy DSL

fabric-api-catalog/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+	enabled = true
+}
+
+// Avoid configuration ordering issues by creating the catalog entries during task execution time
+tasks.register('configureCatalog') {
+	doFirst {
+		doConfigureCatalog()
+	}
+}
+tasks.generateCatalogAsToml {
```
Juuxel: This should be enough in Groovy DSL buildscripts:
```suggestion
generateCatalogAsToml {
```
Or lazily:
```suggestion
tasks.named('generateCatalogAsToml') {
```

Juuxel (approved):

PR #3493, opened 2023-12-31
Custom item predicates

deirn:
Allows mods to add custom item predicates to be used in advancements and loot tables.
Required to the stack-based approach in #3477.


fabric-item-api-v1/build.gradle
```diff
@@ -1,7 +1,26 @@
...
+	'fabric-api-base',
+	'fabric-registry-sync-v0'
+])
 
 testDependencies(project, [
 	':fabric-content-registries-v0',
+	':fabric-data-generation-api-v1'
 ])
+
+loom {
```
maityyy, line 13: `fabricApi { configureDataGeneration() }` does not working for sub modules or something?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicate.java
```diff
@@ -0,0 +1,65 @@
...
+ *     "fabric:type": "minecraft:default",
+ *     "durability": { "max": 20 },
+ *     "count": { "min": 3 }
+ *   }]
+ * }
+ * }</pre>
+ */
+public interface CustomItemPredicate extends Predicate<ItemStack> {
+	RegistryKey<Registry<Codec<? extends CustomItemPredicate>>> REGISTRY_KEY = RegistryKey.ofRegistry(new Identifier("fabric", "custom_item_predicates"));
+	Registry<Codec<? extends CustomItemPredicate>> REGISTRY = FabricRegistryBuilder.createSimple(REGISTRY_KEY).buildAndRegister();
```
maityyy: As far as I know, FAPI avoids using a vanilla registry because it tends to change + hash map is simpler.
In any case, no matter how you implement the registry, you should move it to a separate class as is done in other modules (this improves consistency)

fabric-item-api-v1/build.gradle
deirn: Is that a new loom feature? I just copied from the datagen gradle file.

maityyy: Ah, then it's okay

> Is that a new loom feature?

Yes

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicate.java
deirn: Using registry makes changing the codec easier, though I guess we can hide it and provide a static method.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemPredicateMixin.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	@Inject(method = "<clinit>", at = @At("TAIL"))
+	private static void modifyCodec(CallbackInfo ci) {
+		Codec<ItemPredicate> vanillaCodec = CODEC;
+		VanillaItemPredicate.register(vanillaCodec);
+
+		Codec<CustomItemPredicate> customCodec = CustomItemPredicate.REGISTRY.getCodec().dispatch("fabric:type", CustomItemPredicate::getCodec, c -> c);
+
+		CODEC = Codecs.either(Codecs.nonEmptyList(customCodec.listOf()), vanillaCodec).xmap(e -> e.map(l -> {
```
Syst3ms: Is there a particular reason to require custom predicates to be part of a list instead of a single object like vanilla? Seems like boilerplate.

fabric-item-api-v1/build.gradle
modmuss50: Id keep them all the same. I can look into moving to the new loom feature everwhere in a seperate PR.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicate.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.registry.FabricRegistryBuilder;
+
+/**
+ * Allows for adding custom {@linkplain ItemPredicate item predicate}s, used in advancements and loot tables.
+ *
+ * <p><b>JSON format</b>: along with the vanilla object ({@code {}}) syntax,
+ * Fabric also adds an alternate syntax using a list ({@code []}) that will allow for custom predicates.
```
Syst3ms: It should be specified that putting multiple predicates in a list will act as a conjunction.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemPredicateMixin.java
deirn: AFAIK there's no way to dispatch codec using the key specified in the parent codec, unless we want to make a custom codec impl.

Syst3ms: Couldn't a snippet like this do the trick? 
```java
Codecs.either(vanillaCodec, customCodec)
    .xmap(
        e -> e.map(l -> l, r -> /* wrapper */),
        predicate -> {
            CustomItemPredicate custom = predicate.custom(); // imagine an API like FabricIngredient
            if (custom == null) {
                return Either.left(predicate);
            } else {
                return Either.right(custom);
            }
        }
    )
```
That way symmetry is maintained between the presentation of each type on the JSON level.

deirn: That makes custom and vanilla predicate can't be used at the same time.

Syst3ms: I see your point. It could be solved with more manipulation, but it's not strictly necessary. I just dislike that it's impossible to use a custom predicate without brackets.

fabric-item-api-v1/src/test/java/net/fabricmc/fabric/test/item/unit/CustomItemPredicateCodecTests.java
```diff
@@ -0,0 +1,148 @@
...
+	@BeforeAll
+	static void setup() {
+		SharedConstants.createGameVersion();
+		Bootstrap.initialize();
+
+		new CustomItemPredicateTest().onInitialize();
+	}
+
+	@Test
+	void custom_codec_results_same_serialized_form_as_vanilla() {
```
Syst3ms: Nit: this should still use lowerCamelCase

Syst3ms (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicateRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.base.Preconditions;
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.item.FabricItemPredicateCodec;
+
+public final class CustomItemPredicateRegistry {
+	/**
+	 * Register a codec to be used to serialize/deserialize a custom item predicate
```
apple502j:
```suggestion
	 * Registers a codec to be used to serialize/deserialize a custom item predicate.
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicateRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * @param id    the predicate id
+	 * @param codec the predicate codec
+	 */
+	@SuppressWarnings("unchecked")
+	public static void register(Identifier id, Codec<? extends CustomItemPredicate> codec) {
+		Preconditions.checkArgument(FabricItemPredicateCodec.REGISTRY.containsKey(id), "There's already a codec that registered with this id");
```
apple502j: This is inverted (the first must be true), so no entry can actually be registered.

apple502j (changes requested):

AnAwesomGuy (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemPredicateMixin.java
```diff
@@ -0,0 +1,77 @@
...
+	@Override
+	public void fabric_setCustom(List<CustomItemPredicate> custom) {
+		this.custom = custom;
+	}
+
+	@Inject(method = "<clinit>", at = @At("TAIL"))
+	private static void modifyCodec(CallbackInfo ci) {
+		Codec<ItemPredicate> vanillaCodec = CODEC;
+		CODEC = FabricItemPredicateCodec.init(vanillaCodec);
+	}
```
LlamaLad7: Use a `static` block, also means you don't need `@Mutable`

LlamaLad7 (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemPredicateMixin.java
deirn: What about changing it to `ModifyExpressionValue`?

LlamaLad7: Messier IMO and more prone to breaking

deirn: MCDev says that `Mutable` still missing, but runtime works fine.

Syst3ms: That is a known bug with MCDev that has been [reported](https://github.com/minecraft-dev/MinecraftDev/issues/2148)

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/FabricItemPredicateCodec.java
```diff
@@ -0,0 +1,112 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.predicate.item.ItemPredicate;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.dynamic.RuntimeOps;
+
+import net.fabricmc.fabric.api.item.v1.CustomItemPredicate;
+
+public final class FabricItemPredicateCodec extends MapCodec<ItemPredicate> {
+	public static final BiMap<Identifier, Codec<CustomItemPredicate>> REGISTRY = HashBiMap.create();
```
maityyy, line 42: Shouldn't this be in a separate class? e.g. `CustomItemPredicateRegistryImpl`. Just to maintain consistency + not mix registration logic and codec logic

deirn: Not worth creating a separate class for one field IMO.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/CustomItemPredicateRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+
+public final class CustomItemPredicateRegistry {
+	/**
+	 * Registers a codec to be used to serialize/deserialize a custom item predicate.
+	 *
+	 * @param id    the predicate id
+	 * @param codec the predicate codec
+	 */
+	@SuppressWarnings("unchecked")
+	public static void register(Identifier id, Codec<? extends CustomItemPredicate> codec) {
```
AnAwesomGuy, line 34: i think that this method could be moved to the main `CustomItemPredicate` class, removing the need for another class here

Apollounknowndev:
Superseded by vanilla's item sub-predicate system, which provides the same exact functionality.

apple502j:
Closed, just add to the vanilla registry.

AnAwesomGuy:
should this be reopened, as im porting #3477 to 1.20.6 and this retains vanilla parity, while registering components doesnt

Apollounknowndev:
All this PR added was the ability to add custom item predicates, which you can do in vanilla now. Why would you need a new component?

PR #3495, opened 2023-12-31
Fix local login query response handler not working

### LINKED ISSUE #3494: ServerLoginNetworking.registerReceiver from fabric-networking-api-v1 does not work
Registering a channel using `ServerLoginNetworking.registerReceiver` does not work. The vanilla response of kicking the client when the server receives the client's query response packet is triggered instead of `LoginQueryResponseHandler.receive`.

The global counterpart `ServerLoginNetworking.registerGlobalReceiver` works fine.

I've made a demo mod here: https://github.com/DiffuseHyperion/TestingLoginQuery. Attempting to join the modded server should see you getting kicked with the message "Unexpected custom data from client", instead of being able to join the server normally.

deirn:
Closes https://github.com/FabricMC/fabric/issues/3494

Happens all the way from the start of the Networking API v1, though with the issue only encountered now means that no one actually uses the local handler registry, so not sure if we need to release a fix for all the versions.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerLoginNetworkAddon.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerLoginNetworkAddon.java
@@ -147,7 +147,7 @@ public final class ServerLoginNetworkAddon extends AbstractNetworkAddon<ServerLo
 		}
 
 		boolean understood = originalBuf != null;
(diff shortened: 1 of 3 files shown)


modmuss50 (approved):
Looks good, unless people need this for a sepecific older version I wont bother backporting it.

liach:
Good fix. In the original draft, I don't think the local handler existed, because back then login stage had so little information available that having client-sensitive channels doesn't make much sense.

PR #3508, opened 2024-01-04
Fix race condition when handling unknown packets.

### LINKED ISSUE #3505: Issue with Fabric Networking API
I've been getting random disconnects from Hypixel for the past week and I cannot seem to find out what's causing those disconnects, they happen every 5 Minutes to 3 Hours for seemingly no reason.

The only error message I'm getting is this:
`io.netty.util.IllegalReferenceCountException: refCnt: 0`

Looking into the logs gives this, but nothing else
[fabricapi-kicks.txt](https://github.com/FabricMC/fabric/files/13822366/fabricapi-kicks.txt)

I remember getting disconnected with a slighty different error
`io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1`
But I could't find anything in the logs other than that exact message

Minecraft 1.20.4
Fabric-Loader 0.15.3
Fabric-API 0.92.0
Java 21

modmuss50:
Fixes #3505

This was a fun one to figure out, this change ensures that custom payloads are always relaesed on the network thread, and no further processing of them happens on the main thread.

This caused a rare race codition where a user might be disconnected from a server sending unknown packets with the following error: `io.netty.util.IllegalReferenceCountException: refCnt: 0`

### DIFF
--- a/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/ClientCommonNetworkHandlerMixin.java
+++ b/fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/ClientCommonNetworkHandlerMixin.java
@@ -16,7 +16,10 @@
 
 package net.fabricmc.fabric.mixin.networking.client;
 
+import org.slf4j.Logger;
+import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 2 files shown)


christopherplaysminecraft:
Thankful that it got fixed so fast, this has been annoying me for more than two weeks now.

I have found this playing BedWars on the Hypixel Server.
The server sends custom `badlion:timers` packets which allow the client to display information about the game (Time until new resources spawn, Game Events, etc.) and it does this even if your client doesn't even support those features.

This doesn't cause any issues, typically.

deirn (approved):

PR #3509, opened 2024-01-05
Add a system property for a custom output directory for gametest structures

ErrorCraft:
This PR adds a system property to specify the output directory for gametest structures, so you don't have to move them manually after exporting them using the `test export` command. If the property is not specified, it still uses the default `gameteststructures` directory in your run directory.

Example `build.gradle`:
```gradle
loom {
	runs {
		gametestClient {
			inherit client
			name "GameTest Client"
			vmArg "-Dfabric-api.gametest.structures.output-dir=${file("src/gametest/resources/data/mod_id/gametest/structures")}"

			source sourceSets.gametest
		}
	}
}
```
This then exports the structures to `src/gametest/resources/data/mod_id/gametest/structures`, which can then read by Fabric API again when running your tests. I require the entire path to be specified, as [modmuss talked about additional search directories in Discord](https://discord.com/channels/507304429255393322/566276937035546624/1192157128744505536) too. On top of that, the `test` command does not allow you to specify identifiers, so the mod id has to be included in the path.

### DIFF
--- a/fabric-gametest-api-v1/build.gradle
+++ b/fabric-gametest-api-v1/build.gradle
@@ -2,6 +2,17 @@ version = getSubprojectVersion(project)
 
 loom {
 	accessWidenerPath = file("src/main/resources/fabric-gametest-api-v1.accesswidener")
+
+	runs {
+		testmodClient {
+			client()
+			name = "Testmod Client"
+			vmArg "-Dfabric-api.gametest.structures.output-dir=${file("src/testmod/resources/data/fabric-gametest-api-v1-testmod/gametest/structures")}"
+
+			ideConfigGenerated = true
+			source sourceSets.testmodClient
+		}
+	}
 }
 
 moduleDependencies(project, [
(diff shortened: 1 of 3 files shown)


deirn (approved):

fabric-gametest-api-v1/build.gradle
```diff
@@ -2,6 +2,17 @@ version = getSubprojectVersion(project)
...
 loom {
 	accessWidenerPath = file("src/main/resources/fabric-gametest-api-v1.accesswidener")
+
+	runs {
+		testmodClient {
+			client()
+			name = "Testmod Client"
+			vmArg "-Dfabric-api.gametest.structures.output-dir=${file("src/testmod/resources/data/fabric-gametest-api-v1-testmod/gametest/structures")}"
+
+			ideConfigGenerated = true
```
modmuss50: I think this is fine to have, just dont think it needs to be added to the IDE runs.

```suggestion
			ideConfigGenerated = false
```

modmuss50 (approved):

PR #3510, opened 2024-01-06
Check for the correct BE type in the lookup API

SquidDev:
Apologies if this has been raised before - I couldn't see anything obvious, but may have missed some discussion!

The block lookup API's `registerForBlockEntities` method currently just registers the passed provider for each valid block for that block entity type. While this allows for efficient lookup of providers, it doesn't guarantee that we'll only receive block entities of the correct type to our `BlockEntityApiProvider`!

In some situations, the wrong block entity will be present for a particular block (caused by update suppression or a misbehaving mod, e.g. https://github.com/cc-tweaked/CC-Tweaked/issues/1670). This (incorrect) block entity is then passed off to the `BlockEntityApiProvider`. As providers (especially those registered with the generic `registerForBlockEntity`) will often blind-cast to the expected type, we then get class-cast crashes.

The wrong BE being present is definitely a bug (and arguably it would be bettter to catch this early rather than silently failing!). However, I think in this case we should follow vanilla's lead and handle this more gracefully. In this case, we just check whether the correct BE type is present before forwarding it to the user-supplied `BlockEntityApiProvider`.

### DIFF
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiLookupImpl.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/impl/lookup/block/BlockApiLookupImpl.java
@@ -153,16 +153,16 @@ public final class BlockApiLookupImpl<A, C> implements BlockApiLookup<A, C> {
 			throw new IllegalArgumentException("Must register at least one BlockEntityType instance with a BlockEntityApiProvider.");
 		}
 
-		BlockApiProvider<A, C> nullCheckedProvider = (world, pos, state, blockEntity, context) -> {
-			if (blockEntity == null) {
-				return null;
-			} else {
-				return provider.find(blockEntity, context);
-			}
-		};
-
 		for (BlockEntityType<?> blockEntityType : blockEntityTypes) {
 			Objects.requireNonNull(blockEntityType, "Encountered null block entity type while registering a block entity API provider mapping.");
+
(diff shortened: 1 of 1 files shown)


apple502j (approved):

Juuxel (approved):

modmuss50 (approved):

Technici4n:
Does this warrant more than a silent ignore? In principle, block entities and block states should not get out of sync.

It would also be problematic for a block having different block entity types depending on the state to use this function because there can only be a single provider for the whole block.

SquidDev:
> In principle, block entities and block states should not get out of sync.

Oh agreed! The problem is that in practice (though rare), it does happen! And in that case, we can either crash (the current behaviour) or ignore (which is what vanilla tries to do). I guess we could log and remove the problematic BE, but I think that's too invasive.

Worth mentioning [NeoForge has the same check](https://github.com/neoforged/NeoForge/blob/4a41fbafc644fb8f6299fed381fc72bf031957eb/src/main/java/net/neoforged/neoforge/capabilities/RegisterCapabilitiesEvent.java#L61-L64)  that I'm proposing here, so I think it's probably reasonable. :p.

> It would also be problematic for a block having different block entity types depending on the state to use this function because there can only be a single provider for the whole block.

In that case, shouldn't the modder be registering via the block rather than BE type? Though yes, I guess this is technically a breaking change, Hyrum's law strikes again!

Technici4n:
I think the PR is fine, just food for thought. ;)

I wouldn't worry about the potential BC, it's really a bug that it would have been working previously.

SquidDev:
Thank you!

PR #3511, opened 2024-01-06
Add mob conversion event

Syst3ms:
While working on #3476, I was surprised to find there is no event for mob conversion. This is a generally useful mechanic to expose, so this PR adds a notification event to listen for it. There is no "allow" event as mob conversion happens in a variety of contexts, and what it means to "cancel" conversion can mean a lot of very different things.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
@@ -18,6 +18,7 @@ package net.fabricmc.fabric.api.entity.event.v1;
 
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.damage.DamageSource;
+import net.minecraft.entity.mob.MobEntity;
 
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
@@ -73,6 +74,18 @@ public final class ServerLivingEntityEvents {
 		}
 	});
 
+	/**
+	 * An event that is called when a mob has been converted to another type.
+	 *
+	 * <p>When this event is called, the old instance has not yet been discarded, and the new one has not yet been spawned.
+	 * Mods may use this event to copy some of the old entity's data to the converted one.</p>
+	 */
+	public static final Event<MobConversion> MOB_CONVERSION = EventFactory.createArrayBacked(MobConversion.class, callbacks -> (previous, converted, keepEquiment) -> {
+		for (MobConversion callback : callbacks) {
+			callback.onConversion(previous, converted, keepEquiment);
+		}
+	});
+
 	@FunctionalInterface
 	public interface AllowDamage {
 		/**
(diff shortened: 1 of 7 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/MobEntityMixin.java
```diff
@@ -0,0 +1,36 @@
...
+
+import net.minecraft.entity.EntityType;
+import net.minecraft.entity.mob.MobEntity;
+
+import net.fabricmc.fabric.api.entity.event.v1.ServerLivingEntityEvents;
+
+@Mixin(MobEntity.class)
+public class MobEntityMixin {
+	@Inject(method = "convertTo", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;spawnEntity(Lnet/minecraft/entity/Entity;)Z"))
+	private <T extends MobEntity> void afterEntityConverted(EntityType<T> entityType, boolean keepEquipment, CallbackInfoReturnable<T> cir, @Local T converted) {
```
LlamaLad7: Prefer `@ModifyArg` over local capture. You can get `keepEquipment` with `@Local(argsOnly = true)`

Syst3ms: Ohhh, that's quite sneaky, hadn't thought of that.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
```diff
@@ -73,6 +74,18 @@ public final class ServerLivingEntityEvents {
...
+	 * An event that is called when a mob has been converted to another type.
+	 *
+	 * <p>When this event is called, the old instance has not yet been discarded, and the new one has not yet been spawned.
+	 * Mods may use this event to copy some of the old entity's data to the converted one.</p>
+	 */
+	public static final Event<MobConversion> MOB_CONVERSION = EventFactory.createArrayBacked(MobConversion.class, callbacks -> (previous, converted, keepEquiment) -> {
+		for (MobConversion callback : callbacks) {
+			callback.onConversion(previous, converted, keepEquiment);
+		}
+	});
```
haykam821: The `keepEquiment` parameter name contains a typo:
 
```suggestion
	public static final Event<MobConversion> MOB_CONVERSION = EventFactory.createArrayBacked(MobConversion.class, callbacks -> (previous, converted, keepEquipment) -> {
		for (MobConversion callback : callbacks) {
			callback.onConversion(previous, converted, keepEquipment);
		}
	});
```

modmuss50 (approved):

modmuss50:
Is the plan to get this in before the data attachment API? I think it makes sense to?

Syst3ms:
Yes, preferrably, that avoids having to make a secondary PR to fix this in data attachment.

modmuss50:
Added to last call, should only need a few days at most 👍

Shnupbups:
A special case from vanilla that isn't caught by this is Tadpoles growing up into Frogs, in the `TadpoleEntity#growUp()` method. They are indeed separate entity types, and it utilizes `SpawnReason.CONVERSION` so I believe this is justified to fit this event.

Shnupbups:
Another issue raised on Discord is that perhaps there needs to be clear JavaDoc explaining that this event only pertains to when the resultant entity is different from the original mob, perhaps using the Red Mooshroom to Brown Mooshroom transformation as an example of something that is explicitly NOT covered.

maityyy:
Charged Creeper is also an example

Shnupbups (approved):

apple502j (approved):

Issue #3068, opened 2023-05-17
Add result NBT and count to crafting and smelting recipes


Technici4n:
I propose the following extensions for vanilla recipe outputs:
- crafting recipes: allow specifying NBT data of the result stack.
- smelting recipes: allow specifying NBT data **and count** of the result stack.

We need 3 kinds of patches: JSON format patches to read the extra data, data generation patches to write the extra data, and possibly patches to fix existing logic: furnaces will need a patch to correctly handle count and NBT.

The main reason I am opening this issue is to discuss what the extended JSON format should be: other format extensions such as process conditions and custom ingredient types have always been prefixed with `fabric:`.

## Current JSON formats
Crafting:
```json5
{
  "result": {
    "count": 2, // optional
    "item": "<item id>"
  }
}
```

Smelting:
```json5
{
  "result": "<item id>"
}
```

## Forge JSON format
Forge implements both of these extensions with the following syntax (valid for both crafting and smelting):
```json5
{
  "result": {
    "count": 2, // optional
    "nbt": {}, // optional, can either be an object or a string representation, defaults to null
    "item": "<item id>"
  }
}
```
The old syntax for smelting is still supported.

The main problem with this approach is that there is no way to distinguish vanilla fields and Forge-added fields, however the consistency is nice. In the following I assume that we want a `fabric:` prefix for any Fabric addition.

## Potential JSON format

For crafting recipes it is fairly obvious that we can add a `fabric:nbt` field to the result. For example:
```json5
{
  "result": {
    "count": 2, // optional
    "fabric:nbt": {}, // optional, can either be an object or a string representation, defaults to null
    "item": "<item id>"
  }
}
```
That would solve the problem nicely. Moving on...

Extended smelting recipes are less clear to me. Here are a few options, but each has drawbacks:
- Add `fabric:result_count` and `fabric:result_nbt` to the top level. Looks a bit weird. Introduces an asymmetry between crafting and smelting.
- Add `fabric:result` that supports an object with the same fields as crafting recipes. Replaces vanilla instead of adding to it.
- Extend `result` to accept objects, and support the same syntax as crafting recipes. Replaces vanilla instead of adding to it. Either we do `"count": 2` (no `fabric:` tag to mark the extension), or we do `"fabric:count": 2` (inconsistent with crafting recipes).

## JSON format proposal
Here I will propose a JSON format for these extensions, under the following assumptions:
- We want to mark any addition with a `fabric:` prefix.
[...]

PR #3512, opened 2024-01-06
Cooking recipes working with count.

SufficientlyPositive:
Alters the processing of cooking recipes so that they accept the following JSON formats for deserialising recipe results:

Vanilla
```
"result": "minecraft:example"
```
Additional (Added by this pull request)
```
"result": {
    "item": minecraft:example",      (required)
    "count": 1                       (optional)
 }
```

Serialisation of recipe results is altered from the vanilla format to the additional object format.

There has been fruitful discussion on NBT recipe additions in the following thread:
[Add result NBT and count to crafting and smelting recipes](https://github.com/FabricMC/fabric/issues/3068)
This modification adds the Mojang crafting result Codec to the cooking result deserialisation Codec. Should a Codec that also accepts NBT be produced for crafting, it should be a simple matter to edit `CookingRecipeSerializerMixin` to use that instead. This should give you NBT smithing recipes for (nearly, see below) free.

Suggested modifications should NBT recipes be enabled:
(Yarn names) `canAcceptRecipeOutput` in `AbstractFurnaceBlockEntity.class` does not check NBT data when checking whether a recipe is valid. In order to avoid unintended gain/loss of NBT data for recipes, this function would need to check NBT of the outputs too.

### DIFF
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/AbstractFurnaceBlockEntityMixin.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/AbstractFurnaceBlockEntityMixin.java
@@ -17,11 +17,10 @@
 package net.fabricmc.fabric.mixin.recipe.cooking;
 
 import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
-import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
-import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
 
 import net.minecraft.block.entity.AbstractFurnaceBlockEntity;
 import net.minecraft.item.ItemStack;
@@ -35,32 +34,30 @@ import net.minecraft.registry.DynamicRegistryManager;
 @Mixin(AbstractFurnaceBlockEntity.class)
 public abstract class AbstractFurnaceBlockEntityMixin {
 	/**
-	 * A wrapper altering the result of a smelting recipe when outputting into a partially filled output slot. 1.20.4
-	 * functionality increments the output by 1. This wrapper instead increases the count by the amount defined in the
-	 * recipe.
+	 * Argument modifier that alters the increment amount of a smelting recipe when outputting into a partially filled
+	 * output slot. The current (1.20.4) functionality is to always increment by 1. This is not compatible with multiple
+	 * item outputs. As such this modifies the argument and sets it to exactly the recipe output. Note: this overrides
+	 * any previous modifications to "amount".
 	 *
 	 * <p>Assumes that the increment will be valid and does not stop recipe if it would go over. E.g. an increment of
 	 * 65 would always be performed, if not stopped by other logic prior to this.
 	 *
-	 * @param outputStack the instance of the itemstack to increment, in the furnace's output slot.
-	 * @param increment the original incremented amount (1.20.4 this is always 1).
-	 * @param incrementItemStack the method responsible for increasing the item stack count.
-	 * @param registryManager the registry manager to access for the recipe
-	 * @param recipe the (!= null) recipe.
+	 * @param amount the original increment amount (1.20.4 this is always 1).
+	 * @param registryManager the registry manager to access for the recipe.
+	 * @param recipe the (!= null) cooking recipe.
+	 * @return the amount specified by the recipe, rather than the default 1.
 	 */
-	@WrapOperation(
+	@ModifyArg(
 			method = "craftRecipe",
 			at = @At(
 					value = "INVOKE",
 					target = "Lnet/minecraft/item/ItemStack;increment(I)V"
 			)
 	)
-	private static void incrementByRecipe(ItemStack outputStack, int increment, Operation<Void> incrementItemStack, DynamicRegistryManager registryManager, RecipeEntry<?> recipe) {
-		incrementItemStack.call(
-				outputStack,
-				recipe.value()
-						.getResult(registryManager)
-						.getCount());
+	private static int incrementByRecipe(int amount, @Local(argsOnly = true) DynamicRegistryManager registryManager, @Local(argsOnly = true) RecipeEntry<?> recipe) {
+		return recipe.value()
+				.getResult(registryManager)
+				.getCount();
 	}
 
 	/**
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/AbstractFurnaceBlockEntityMixin.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/AbstractFurnaceBlockEntityMixin.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


fabric-recipe-api-v1/src/testmod/resources/data/minecraft/recipes/iron_ingot_from_smelting_deepslate_iron_ore.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "minecraft:smelting",
+  "group": "iron_ingot",
+  "ingredient": {
+    "item": "minecraft:deepslate_iron_ore"
+  },
+  "result": {
+    "item": "minecraft:iron_ingot",
+    "count": 3
```
maityyy, line 9: I'm assuming it should be `fabric:count` instead of `count`

maityyy:
You change the codec, okay, but what about networking?

fabric-recipe-api-v1/src/testmod/resources/data/minecraft/recipes/iron_ingot_from_smelting_deepslate_iron_ore.json
SufficientlyPositive: This is using the Vanilla Codec for processing crafting recipe items with count. There is merit to renaming it in order to differentiate it from Vanilla datapack processing, but personally I prefer the consistency with vanilla.
I'm open to changing this however, but for now I am partial towards the proposal Shadows-of-Fire mentioned in the NBT thread.

SufficientlyPositive:
> You change the codec, okay, but what about networking?

Iirc networking uses the codecs for both serialisation and deserialisation also?

I assume there may be some slowdown from packets switching from the String format to the JSON object format, but I don't imaging this is particularly significant. Crafting uses the same Codec.

maityyy:
I decided to recheck the code and everything seems fine, so I'm sorry. :D My concern was that network packets might ignore the item NBT. And btw, no, recipe network packets doesn't use codecs

SufficientlyPositive:
> I decided to recheck the code and everything seems fine, so I'm sorry :D Btw, no, recipe network packets doesn't use codecs

Ah thanks, must me my faulty memory. I was under the impression that the pipeline was something like `(object) --codec--> (JSON) --magic--> (packets)` but it's been a while. I will need to recheck it 😄

fabric-recipe-api-v1/src/testmod/resources/data/minecraft/recipes/iron_ingot_from_smelting_deepslate_iron_ore.json
maityyy: Well, I personally agree with your point of view, but fabric adds a suffix to all of its changes to vanilla formats (look at resource conditions and custom ingredients), so consistency should be assessed from fabric's point of view, not vanilla's

SufficientlyPositive: Thanks for pointing me to these, I will take a look at them. 
I agree that consistency should be prioritised over the mod as a whole, so I will look into building a separate codec to use with `fabric:count` and build and test it over the next few days.

SufficientlyPositive: How about changing the new result format to `fabric:result`?, Leaving out the normal `result`. 
So something like this:
```
{
  "type": "minecraft:smelting",
  "group": "iron_ingot",
  "ingredient": {
    "item": "minecraft:deepslate_iron_ore"
  },
  "fabric:result": {
    "item": "minecraft:iron_ingot",
    "count": 3
  },
  "experience": 0,
  "cookingtime": 1
}
```

My reasoning is as follows: This change does not simply add to an additional option to an existing format, but instead adds an entirely new way of expressing results. For example, simply naming the (optional) count value `fabric:count` does not imply that a result of the format 
```
"result": {
  "item": "minecraft:iron_ingot"
}
```
would fail to compile in Vanilla (which it would). Renaming to `fabric:result` over `result` would express that this is a fabric-specific datapack file, even in the case where an item count is not provided.

maityyy: How will this work with datagen?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/AbstractFurnaceBlockEntityMixin.java
```diff
@@ -0,0 +1,90 @@
...
+	 * @param recipe the (!= null) recipe.
+	 */
+	@WrapOperation(
+			method = "craftRecipe",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/item/ItemStack;increment(I)V"
+			)
+	)
+	private static void incrementByRecipe(ItemStack outputStack, int increment, Operation<Void> incrementItemStack, DynamicRegistryManager registryManager, RecipeEntry<?> recipe) {
```
LlamaLad7: Should be a `@ModifyArg` with `@Local(argsOnly = true)`s to get the enclosing params.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/CookingRecipeSerializerMixin.java
```diff
@@ -0,0 +1,86 @@
...
+	@WrapOperation(
+			method = "method_53766(ILnet/minecraft/recipe/AbstractCookingRecipe$RecipeFactory;Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;",
+			at = @At(
+					value = "INVOKE",
+					target = "Lcom/mojang/serialization/Codec;fieldOf(Ljava/lang/String;)Lcom/mojang/serialization/MapCodec;", // Lcom/mojang/serialization/Codec;fieldOf(Ljava/lang/String;)Lcom/mojang/serialization/MapCodec;
+					ordinal = 1,
+					remap = false
+			)
+	)
+	private static MapCodec<? extends ItemStack> addAlternativeCodec(Codec<? extends ItemStack> originalCodec, String fieldName, Operation<MapCodec<? extends ItemStack>> toMapCodec) {
```
LlamaLad7: `@ModifyReceiver`?

LlamaLad7 (changes requested):

fabric-recipe-api-v1/src/testmod/resources/data/minecraft/recipes/iron_ingot_from_smelting_deepslate_iron_ore.json
SufficientlyPositive: Some form of additional functionality would be required for modders to be able to use this functionality with datagen. I do not intend to implement this within this PR.

maityyy: I mean, to what extent is `fabric:result` a realistic option? I would prefer the option that is easier from a datagen point of view, because a large number of recipes are created through datagen. I'm not saying you need to implement datagen support, I'm saying we can't make a choice without datagen

modmuss50 (commented):
- Datagen support is needed
- The tests should be improved, have a look at adding some game or unit tests to ensure that vanilla behaviour is presevered. (let me know if you need help with this)

Generally we would prefix it with fabric:

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/cooking/CookingRecipeSerializerMixin.java
```diff
@@ -0,0 +1,79 @@
...
+			method = "method_53766(ILnet/minecraft/recipe/AbstractCookingRecipe$RecipeFactory;Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;",
+			at = @At(
+					value = "INVOKE",
+					target = "Lcom/mojang/serialization/Codec;fieldOf(Ljava/lang/String;)Lcom/mojang/serialization/MapCodec;",
+					ordinal = 1,
+					remap = false
+			)
+	)
+	private static Codec<? extends ItemStack> addAlternativeCodec(Codec<? extends ItemStack> originalCodec, String fieldName) {
+		return Codecs.alternatively(
```
Technici4n, line 75: This changes the default format, you should use an xmapped either codec. To give you some ideas:
image
(this also supports NBT)

maityyy: Here is a link for easy copy :DDDD https://github.com/neoforged/NeoForge/pull/300

PR #3515, opened 2024-01-07
Update to loom 1.5

modmuss50:
Disabling the Mixin AP is quite a large change, with plently of room for things to go wrong, so this will need some good testing.


modmuss50:
Thinking of changing this to just be the 1.5 update, we can disable the mixin AP later.

PR #3295, opened 2023-09-01
Stack aware `getFoodComponent`

Phoupraw:
Similar to `getRecipeRemainder`.

PR #3520, opened 2024-01-12
[1.20.4] Stack aware `getFoodComponent`

modmuss50:
99% of the work was done in #3295 by @Phoupraw just a 1.20.4 version

~~This is the same, just targgeting 1.20.4 and using @WrapOperation in place of @Redirect~~

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
@@ -17,12 +17,14 @@
 package net.fabricmc.fabric.api.item.v1;
 
 import com.google.common.collect.Multimap;
+import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.block.BlockState;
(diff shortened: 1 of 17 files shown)


LlamaLad7:
I wouldn’t use WrapOperation for any of those. They are all meaningful unconditional replacements and as such they should always be the innermost thing in a wrapper chain, and if another mod wants to do a meaningful unconditional replacement as well then they should conflict. For those reasons Redirect is the right choice. The ThreadLocal should however be replaced with a `@Share`.

modmuss50:
> I wouldn’t use WrapOperation for any of those. They are all meaningful unconditional replacements and as such they should always be the innermost thing in a wrapper chain, and if another mod wants to do a meaningful unconditional replacement as well then they should conflict. For those reasons Redirect is the right choice. The ThreadLocal should however be replaced with a `@Share`.

Reverted that change and moved to share. Thats a good point, I didnt really give it much thought 👍

LlamaLad7 (approved):
LGTM. There are situations in which you would want to wrap even without calling the original, but as with everything in mixin it’s nuanced.

Issue #3230, opened 2023-07-31
CoreShaderRegistrationCallback is called multiple times on shader reload in 1.20.x


Draylar:
### Issue

`CoreShaderRegistrationCallback` class docs:

```
Called when core shaders ({@linkplain ShaderProgram shader programs} loaded from {@code assets/<namespace>/shaders/core})
are loaded to register custom modded shaders.
```

My understanding of this callback is that it should run _once_ when core shader assets are reloaded - but in 1.20.1, it seems to trigger *every time* a core vanilla shader is loaded in `GameRenderer#loadPrograms`. In the following test suite example, the shader is actually loaded about **60** times in a vanilla environment. You can verify this by writing a core shader file with an unreferenced uniform - the game will spit out a warning ~60 times (equal to the number of vanilla core shaders loaded in the inject method target).

https://github.com/FabricMC/fabric/blob/d9acd820fc8e56c8ac18c79d650efb9c413999a0/fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java#L45-L49

You can also stick a breakpoint in here & see that any custom program is added to `programs` ~60 times:

https://github.com/FabricMC/fabric/blob/d9acd820fc8e56c8ac18c79d650efb9c413999a0/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/GameRendererMixin.java#L49-L54

### Solution

I think the Slice is messed up in the inject, but I don't know enough about them to propose a fix. I personally would drop the Slice and inject at `List#add`, ordinal 0 (which would give the expected behavior of the callback only running once) - or inject at the `clearPrograms` call, which would let modded core shaders load after vanilla core shaders.

### Versions

```
minecraft_version=1.20.1
yarn_mappings=1.20.1+build.8
loader_version=0.14.21
```

PR #3521, opened 2024-01-12
Fix injection point for CoreShaderRegistrationCallback

Draylar:
Relevant issue: https://github.com/FabricMC/fabric/issues/3230

From 1.20.1(?) to 1.20.4, the injection point for the CoreShaderRegistrationCallback event has been broken: it triggers _every time_ vanilla registers a core shader, rather than once during the shader reload cycle. In practice, this means if you rely on the event for loading shaders and use the event as modeled in the test suite, your shader is parsed/validated about _60 times per reload_. An extra implication of this is that if your shader has any unused fields, it will spam console ~60 times telling you to remove it (compared to a single warning per reload, which is what happens in a vanilla environment).

I removed the slice in favor of injecting at the first add call. A potential adjustment might be injecting at the last add call (to load modded shaders after vanilla shaders), but I'm not sure how to hit that inject point without leaving the try/catch block.

### Testing / Replicating

https://github.com/FabricMC/fabric/blob/a462da68c68e109073526be5dd1ae8803e642435/fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudAndShaderTest.java#L45-L49

Place a print statement in this callback, or register a shader that will intentionally print a warning (such as one with an unused uniform). The print/warning will log approximately 60 times, once per vanilla core shader registration. The callback should only be triggered once, which is what happens after this patch is applied.

---

Thank you in advance for reviews and/or change requests! 💗

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/GameRendererMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/GameRendererMixin.java
@@ -24,7 +24,6 @@ import com.mojang.datafixers.util.Pair;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.Slice;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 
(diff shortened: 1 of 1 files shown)


Chocohead:
> I removed the slice in favor of injecting at the first add call. A potential adjustment might be injecting at the last add call (to load modded shaders after vanilla shaders), but I'm not sure how to hit that inject point without leaving the try/catch block.

You can do `@At(value = "INVOKE_ASSIGN:LAST"` to inject immediately after the final `List#add`

LlamaLad7:
You cannot, selectors like that are only for use in `@Slice`s.

LlamaLad7:
You can however use a slice to get that selector functionality. This should also absolutely not use locals. I would suggest the following:
```java
@WrapOperation(
		method = "loadPrograms",
		slice = @Slice(
				from = @At(value = "INVOKE:LAST", target = "Ljava/util/List;add(Ljava/lang/Object;)Z", remap = false)
		),
		at = @At(value = "INVOKE", target = "Ljava/util/List;add(Ljava/lang/Object;)Z", remap = false)
)
private boolean registerShaders(List<Pair<ShaderProgram, Consumer<ShaderProgram>>> programs, Object e, Operation<Boolean> original, ResourceFactory factory) throws IOException {
	boolean result = original.call(programs, e);
	CoreShaderRegistrationCallback.RegistrationContext context = (id, vertexFormat, loadCallback) -> {
		ShaderProgram program = new FabricShaderProgram(factory, id, vertexFormat);
		programs.add(Pair.of(program, loadCallback));
	};
	CoreShaderRegistrationCallback.EVENT.invoker().registerShaders(context);
	return result;
}
```

modmuss50:
Superseeded by: https://github.com/FabricMC/fabric/pull/3886

Many thanks for the PR

PR #3522, opened 2024-01-12
Allow `FabricCodecDataProvider` to access dynamic registries

ErrorCraft:
This PR allows `FabricCodecDataProvider` to access the dynamic registries if it is passed in the constructor. Now, I have a couple of issues with this implementation myself because I didn't want to cause a breaking change, the biggest one being that if you do not override the correct method in your subclass, it won't generate any files. Using the `FabricCodecDataProvider#FabricCodecDataProvider(FabricDataOutput, CompletableFuture, DataOutput.OutputType, String, Codec)}` constructor and then the `FabricCodecDataProvider#configure(BiConsumer)` method won't generate anything as it's never called.

This could all be avoided by always requiring a `CompletableFuture<RegistryWrapper.WrapperLookup>` in the constructor and by adding said `RegistryWrapper.WrapperLookup` to the `configure` method, but that would be a breaking change. We would end up with a much more graceful implementation though.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
@@ -52,7 +52,7 @@ public abstract class FabricCodecDataProvider<T> implements DataProvider {
 	/**
 	 * @deprecated Please use {@link FabricCodecDataProvider#FabricCodecDataProvider(FabricDataOutput, CompletableFuture, DataOutput.OutputType, String, Codec)}.
 	 */
-	@Deprecated(forRemoval = true)
+	@Deprecated()
 	protected FabricCodecDataProvider(FabricDataOutput dataOutput, DataOutput.OutputType outputType, String directoryName, Codec<T> codec) {
 		this.pathResolver = dataOutput.getResolver(outputType, directoryName);
 		this.registriesFuture = null;
@@ -107,7 +107,7 @@ public abstract class FabricCodecDataProvider<T> implements DataProvider {
 	 * @param provider A consumer that accepts an {@link Identifier} and a value to register.
 	 * @deprecated Please use {@link FabricCodecDataProvider#configure(BiConsumer, RegistryWrapper.WrapperLookup)}.
 	 */
-	@Deprecated(forRemoval = true)
+	@Deprecated()
 	protected void configure(BiConsumer<Identifier, T> provider) {
 		throw new UnsupportedOperationException("Override this method.");
 	}
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
@@ -109,6 +109,7 @@ public abstract class FabricCodecDataProvider<T> implements DataProvider {
 	 */
 	@Deprecated(forRemoval = true)
 	protected void configure(BiConsumer<Identifier, T> provider) {
+		throw new UnsupportedOperationException("Override this method.");
 	}
 
 	/**
@@ -118,6 +119,7 @@ public abstract class FabricCodecDataProvider<T> implements DataProvider {
 	 */
 	protected void configure(BiConsumer<Identifier, T> provider, RegistryWrapper.WrapperLookup lookup) {
 		// TODO: Make abstract once the deprecated method is removed.
+		throw new UnsupportedOperationException("Override this method.");
 	}
 
 	private JsonElement convert(Identifier id, T value) {
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
@@ -26,10 +26,12 @@ import com.google.gson.JsonElement;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.DataResult;
 import com.mojang.serialization.JsonOps;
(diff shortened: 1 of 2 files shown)


deirn:
I'd argue that since the datagen module only used in dev, it's fine to break the API.

modmuss50:
> I'd argue that since the datagen module only used in dev, it's fine to break the API.

No, I dont think this is a something we want to do. A few people actually use this in prod as well (for runtime resource gen)

apple502j:
I recommend breaking this in 1.20.5.

deirn:
> > I'd argue that since the datagen module only used in dev, it's fine to break the API.
> 
> No, I dont think this is a something we want to do. A few people actually use this in prod as well (for runtime resource gen)

Interesting, I thought people uses different API for that. Still, I don't think the datagen module needs to follow the strict backwards compatibility promise as other modules.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -69,11 +105,27 @@ public CompletableFuture<?> run(DataWriter writer) {
 	 * Implement this method to register entries to generate.
 	 *
 	 * @param provider A consumer that accepts an {@link Identifier} and a value to register.
+	 * @deprecated Please use {@link FabricCodecDataProvider#configure(BiConsumer, RegistryWrapper.WrapperLookup)}.
 	 */
-	protected abstract void configure(BiConsumer<Identifier, T> provider);
+	@Deprecated(forRemoval = true)
+	protected void configure(BiConsumer<Identifier, T> provider) {
+	}
```
modmuss50: This should throw, saying that the method needs to be overriden. Previously this was forced by being abstract.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -69,11 +105,27 @@ public CompletableFuture<?> run(DataWriter writer) {
...
+
+	/**
+	 * Implement this method to register entries to generate using a {@link RegistryWrapper.WrapperLookup}.
+	 * @param provider A consumer that accepts an {@link Identifier} and a value to register.
+	 * @param lookup A lookup for registries.
+	 */
+	protected void configure(BiConsumer<Identifier, T> provider, RegistryWrapper.WrapperLookup lookup) {
+		// TODO: Make abstract once the deprecated method is removed.
```
modmuss50, line 121: This should throw, see above.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -19,38 +19,74 @@
...
  */
 public abstract class FabricCodecDataProvider<T> implements DataProvider {
 	private final DataOutput.PathResolver pathResolver;
+	private final CompletableFuture<RegistryWrapper.WrapperLookup> registriesFuture;
 	private final Codec<T> codec;
 
+	/**
+	 * @deprecated Please use {@link FabricCodecDataProvider#FabricCodecDataProvider(FabricDataOutput, CompletableFuture, DataOutput.OutputType, String, Codec)}.
+	 */
+	@Deprecated(forRemoval = true)
```
modmuss50: Whats our stance on `forRemoval` I always forget. I think its reasonable to be used here as we will likely be more aggressive in removing it?

modmuss50 (changes requested):
LGTM, I always forget our stance on `forRemoval` I thought we generally didnt use it?

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
ErrorCraft: Added an `UnsupportedOperationException` with the message, not sure if that is the desired exception here though.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
ErrorCraft: [CONTRIBUTING.md mentions that `forRemoval` should be avoided](https://github.com/FabricMC/fabric/blob/1.20.4/CONTRIBUTING.md#annotations), but given some of the other comments on the PR and [your message in Discord](https://discord.com/channels/507304429255393322/566276937035546624/1195461399523106946) I don't think it is that much of an issue here.

modmuss50 (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
modmuss50:
```suggestion
	@Deprecated()
```

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricCodecDataProvider.java
```diff
@@ -69,11 +105,29 @@ public CompletableFuture<?> run(DataWriter writer) {
 	 * Implement this method to register entries to generate.
 	 *
 	 * @param provider A consumer that accepts an {@link Identifier} and a value to register.
+	 * @deprecated Please use {@link FabricCodecDataProvider#configure(BiConsumer, RegistryWrapper.WrapperLookup)}.
 	 */
-	protected abstract void configure(BiConsumer<Identifier, T> provider);
+	@Deprecated(forRemoval = true)
```
modmuss50:
```suggestion
	@Deprecated()
```

Goint to remove forRemoval, I dont think its needed.

PR #3523, opened 2024-01-12
fabric-model-events-api

Sollace:
This is a draft!

I'm still working on it, but in the meantime I want to get feedback and ideas on how to improve.

---------------

# What is this?

The goal of this PR is to introduce API to Fabric that will allow mods to more easily render additions to player and entity models. I wanted to make it possible for a mod to subscribe to receieve events whenever a specific named part of a model is rendered (such as the player's head). They should then recieve as part of the event enough information to render additional accessories (like a hat) onto the player's head using the top face as a point of fixture.

The head is a trivial example, though, so to try for something a bit more complicated, consider this scenario:

- Mod A is a backpack mod. It wants to render a backpack onto the player's back whenever their item is equipped.
- Mod B is "Mine Little Pony". It replaces the player's entire model with that of a four-legged creature.

If Mod A hardcodes the position and orientation of its backpack, when Mod B is installed, the backpack may appear in the wrong location (clipping into the pony's back). Mod A would have to add specific support for Mod B in order for them both to render in a compatible way, which puts the burden of implenting this onto Mod A's developers who may not be interested in doing this.

This API is intended to resolve this (and potentially several others), in addition to providing better tools for mods to add more complex additions to player models).

If instead, Mod A implemented the rendering for their backpack using this API, they can instead add an event for when the player's torso is rendered, grab the north face, and orient their backpack to that.

Then, so long as the part for the player's torso is oriented correctly by Mod B, the backpack from Mod A will always appear correctly on the differently-shaped model.

# Things this API adds

There are three types of events:
- The general event
   This is fired when a model part is rendered. Doesn't matter when, you get notified of it. The only context you get for this event listener is the part and whatever was given when rendering it.
   
- The entity event
   This one is specifically only fired when a model is being rendered for an entity (via EntityRendererDispatcher). 
   You get notified when the model is being rendered for an entity matching the type you specify, and you receive the entity as part of your context.

- The block entity event
   Same as the entity event but for block entities

- PartView, CubeData, FaceData
    The ModelPart being rendered is provided in the form of a PartView which gives you readonly access to pretty much everything you need to know about the ModelPart and its contents.

- ModelVisitor
    PartViews can be traversed using a visitor pattern. You can create a visitor using ModelVisitor.builder() and specify actions to perform for a Part, a Cube, and a Face. Or you can just query the data using PartView.cubes()

When registering for an event you specify the name/path of the part you want to get events for, and register with one of the three types of event listeners.

Example that would give the player a second head:
```
ModelPartCallbacks.get(PartTreePath.of(EntityModelPartNames.HEAD)).register(EntityType.PLAYER,
        (player, partView, matrices, vertexConsumer, tickDelta, light, overlay, r, g, b, a) -> {
     partView.part().pivotX = -5;
     partView.part().render(matrices, vertexConsumer, light, overlay, r, g, b, a);
     partView.part().pivotX = 4;
});
```

You can also specify a `MatchingStrategy` (default is ENDS_WITH) to change the behaviour in how it decides which parts match the path given.

This example will fire for any parts contained within the player's head:

```
ModelPartCallbacks.get(ModelPartCallbacks.MatchingStrategy.STARTS_WITH, PartTreePath.of(EntityModelPartNames.HEAD)).register(EntityType.PLAYER, 
        (player, partView, matrices, vertexConsumer, tickDelta, light, overlay, r, g, b, a) -> {
    // ...
});
```

image

image


fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/EmptyDataCollection.java
```diff
@@ -0,0 +1,56 @@
...
+package net.fabricmc.fabric.impl.client.modelevents;
+
+import java.util.Iterator;
+import java.util.Optional;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.fabricmc.fabric.api.client.modelevents.v1.data.DataCollection;
+
+@ApiStatus.Internal
```
Syst3ms, line 26: This is unnecessary, loom will add this annotation to all impl packages automatically.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Optional;
+import net.fabricmc.fabric.impl.client.modelevents.EmptyDataCollection;
+
+/**
+ * Represents a managed collection of data.
+ *
+ * @param <T> The data elements this collection contains.
+ */
+public interface DataCollection<T> extends Iterable<T> {
```
Syst3ms, line 50: Is there a particular reason why this can't just be an ImmutableList or something?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
```diff
@@ -0,0 +1,84 @@
...
+import net.fabricmc.fabric.api.client.modelevents.v1.data.FaceData;
+import net.fabricmc.fabric.impl.client.modelevents.FaceDataImpl;
+import net.minecraft.client.model.Dilation;
+import net.minecraft.client.model.ModelPart;
+import net.minecraft.client.model.ModelPart.Cuboid;
+import net.minecraft.client.model.ModelPart.Quad;
+import net.minecraft.util.math.Direction;
+
+@ApiStatus.Internal
+@Mixin(value = ModelPart.Cuboid.class, priority = Integer.MAX_VALUE)
```
Syst3ms: Mixin priority is generally useless for FAPI. Besides, a lower priority value means the mixin gets triggered earlier, so you gave this the smallest priority possible.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
```diff
@@ -0,0 +1,84 @@
...
+abstract class ModelPart_CuboidMixin implements CubeData {
```
Syst3ms: No underscores in class names, even for nested classes. If Eclipse has a plugin for checkstyle install it ASAP. Otherwise run one of the checkstyle gradle tasks.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
Syst3ms, line 1: I'm frankly amazed this is even legal java code, but it makes sense since all of these are package-private. Regardless, don't do this, and split these mixins over multiple files instead. Plus, I don't know how these declarations behave in bytecode but there's a nonzero chance it completely breaks mixin. EDIT: if the tests run, then maybe not, but please don't do this anyway.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
```diff
@@ -0,0 +1,90 @@
...
+ * - player arm rendering
+ * - mods that render an entity by doing {@code dispatcher.getRenderer(entity).render(...)}
+ *
+ * Without this injection (or rendering via other methods) normal events still work, but events that specifically
+ * depend on the entity being rendered being known, will be skipped.
+ */
+@ApiStatus.Internal
+@Mixin(value = EntityRenderDispatcher.class, priority = Integer.MAX_VALUE)
+abstract class EntityRenderDispatcherMixin {
+    private static final String RENDER = "render(Lnet/minecraft/entity/Entity;DDDFFLnet/minecraft/client/util/math/MatrixStack;Lnet/minecraft/client/render/VertexConsumerProvider;I)V";
```
Syst3ms, line 48: No need for this. Line length isn't really a concern if that's your worry.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    @Accessor("cuboids")
+    public abstract List<Cuboid> getCuboids();
+
+    @Override
+    @Accessor("children")
+    public abstract Map<String, ModelPart> getChildren();
+
+    @Dynamic("Compiler-generated class constructor method")
```
Syst3ms: Probably unnecessary. `<init>` properly exists in bytecode and Mixin should have no trouble targeting it.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
```diff
@@ -0,0 +1,86 @@
...
+abstract class ModelPartMixin implements FabricPartHooks.Container {
+    @Shadow
+    public boolean visible;
+    @Shadow
+    public boolean hidden;
+
+    private FabricPartHooks fabric_hooks;
+
+    @Override
+    @Accessor("cuboids")
```
Syst3ms, line 55: Does this work?

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+	    test(errors, () -> {
+	        assert PartTreePath.of().isRoot();
+	    });
+	    test(errors, ModelEventsTest::checkEmptyPartPathCreation);
+	    test(errors, ModelEventsTest::checkPartPathParsing);
+	    test(errors, ModelEventsTest::checkPathIndexOfForShortPath);
+	    test(errors, ModelEventsTest::checkPathIndexOfForLongPath);
+	    test(errors, ModelEventsTest::checkPathComparisons);
+        test(errors, ModelEventsTest::registerEntityModelPartListener);
```
Syst3ms, line 88: What happened with the indentation there? Again, run checkstyle, it will definitely catch that and then some.

gradle.properties
```diff
@@ -41,6 +41,7 @@ fabric-message-api-v1-version=6.0.4
 fabric-mining-level-api-v1-version=2.1.59
 fabric-model-loading-api-v1-version=1.0.7
 fabric-models-v0-version=0.4.6
+fabric-model-events-api-v1-version=0.1.0
```
Syst3ms: This should be `1.0.0`, 0.x versions are only for legacy modules.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Gets a single element at a given position within this collection.
+     * If the collection is empty of the requested index does not correspond to
+     * an element contained in this collection, will instead return an empty optional.
+     *
+     * @param index Element index
+     * @return Optional element at the requested index
+     */
+    Optional<T> getAt(int index);
```
Syst3ms, line 72: Optionals are generally discouraged in FAPI, prefer an `@Nullable`.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/ModelRenderContext.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import it.unimi.dsi.fastutil.Stack;
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.entity.Entity;
+
+@ApiStatus.Internal
+public final class ModelRenderContext {
+    static final Stack<Entity> CURRENT_ENTITY = new ObjectArrayList<>();
```
Syst3ms, line 28: Should this be thread-safe?

Syst3ms (changes requested):
I'm not familiar with rendering code, so I can't judge the core API design, so this is just about general code structure and stuff I caught. This direly needs checkstyle.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
Sollace: Yeah. It works.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
Sollace: This is perfectly legal and I've done it semi-frequently to group related mixins.

Fair. I'll separate them.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
Sollace: What format must I use for inner classes then?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
Sollace: Yes, that's the idea. It's meant to run last.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
Sollace: Yes. I'm implementing custom logic, and if you read the rest of the class you'll see the interfaces are different.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
Sollace: It's being used in two places.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
Juuxel: `ModelPartCuboidMixin` should be the expected format.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
Sollace: Regardless, I add it so builds don't produce warnings when it can't find it.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
Sollace: No. :)

Juuxel:
The module name should be `fabric-model-events-v1`.

https://github.com/FabricMC/fabric/blob/a462da68c68e109073526be5dd1ae8803e642435/CONTRIBUTING.md?plain=1#L289-L291

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/ModelRenderContext.java
Sollace: I'm not concerned about thread safety since all rendering is done on the same thread.

Sollace:
> The module name should be `fabric-model-events-v1`.
> 
> https://github.com/FabricMC/fabric/blob/a462da68c68e109073526be5dd1ae8803e642435/CONTRIBUTING.md?plain=1#L289-L291

The thing you quoted appears to disagree...?

Juuxel:
> Event modules should have the `-events` suffix instead.

This seems to be an event module?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
Juuxel: This is also in the contributing guidelines:

https://github.com/FabricMC/fabric/blob/a462da68c68e109073526be5dd1ae8803e642435/CONTRIBUTING.md?plain=1#L97-L98

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
Sollace: This file is a bit weird with spaces and tabs. It's kind of hard to fix inside the IDE since it renders both the same and (confusingly) retains the mix of spaces and tabs when adding new lines.

I'll fix it later.

Sollace:
> > Event modules should have the `-events` suffix instead.
> 
> This seems to be an event module?

Hm okay

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
Sollace: To offer a better explanation:
This works because I specified the field name on the @Accessor annotation instead of letting try to inspect the method name to find one. That's leaves the method name open for me to name it anything I want, in this case to implement the interface.

fabric-model-events-api-v1/src/client/resources/assets/fabric-mining-level-api-v1/icon.png
Juuxel, line 1: The file path is not quite right ("mining level api"), will also need adjusting anyway if the module name is changed

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
Syst3ms: This kind of duplication isn't a big deal at all for mixin classes, IMO. They're too rigid and static for this kind of constant reuse to remotely be a problem.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
Syst3ms: This constructor is completely explicit in the source class, I have no clue why Mixin would be raising warnings.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
Syst3ms: `MAX_VALUE` seems a bit overkill for that. The vast majority of mixins don't use priority at all, so a priority of 1001 could be enough (FAPI uses a priority of 999 in many places to trigger before mods). I'd like a second opinion on this though.

apple502j:
Oh no, another module. Could we just add this to Model Loader or Rendering APIs?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+ *      matrixStack.push();
+ *      matrixStack.translate(headCenter.x, headCenter.y, headCenter.z);
+ *      // Draw your model
+ *      // Remember to pop your stack
+ *      matrixStack.pop();
+ * });
+ * </pre>
+ * }
+ */
+public interface ModelPartCallbacks {
```
apple502j, line 46: NonExtendable.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/PartTreePath.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+package net.fabricmc.fabric.api.client.modelevents.v1;
+
+import net.fabricmc.fabric.impl.client.modelevents.PartTreePathImpl;
+
+/**
+ * Represents the absolute or relative path of a part within a model's tree.
+ */
+public interface PartTreePath extends Iterable<String> {
```
apple502j, line 24: NonExtendable.

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
```diff
@@ -0,0 +1,190 @@
...
+import net.minecraft.client.render.WorldRenderer;
+import net.minecraft.client.render.entity.model.EntityModelPartNames;
+import net.minecraft.entity.EntityType;
+
+public final class ModelEventsTest implements ClientModInitializer {
+	private static final Logger LOGGER = LoggerFactory.getLogger(ModelEventsTest.class);
+
+	private static final ModelVisitor DEBUG_VISITOR = ModelVisitor.builder()
+            .visitCubes((matrices, cube) -> {
+                var debugConsumer = MinecraftClient.getInstance().getBufferBuilders().getEntityVertexConsumers().getBuffer(RenderLayer.getLines());
```
apple502j: Use `var` only in constructors. Also, why do users need to code like this? Can't we at least pass client instance?

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+	private static void test(List<AssertionError> errors, Runnable runnable) {
+		try {
+			runnable.run();
+		} catch (AssertionError e) {
+			errors.add(e);
+		}
+	}
+
+	static void checkEmptyPartPathCreation() {
```
apple502j, line 110: Ideally these should be written in JUnit.

fabric-model-events-api-v1/src/client/resources/fabric.mod.json
```diff
@@ -0,0 +1,30 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-model-events-api-v1",
+  "name": "Fabric Mining Level API (v1)",
```
apple502j: 👀

fabric-model-events-api-v1/src/client/resources/fabric.mod.json
```diff
@@ -0,0 +1,30 @@
...
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.15.1",
+    "fabric-api-base": "*",
+    "fabric-lifecycle-events-v1": "*"
```
apple502j: This dependency is not declared in `build.gradle`.

fabric-model-events-api-v1/src/client/resources/fabric.mod.json
```diff
@@ -0,0 +1,30 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-model-events-api-v1",
+  "name": "Fabric Mining Level API (v1)",
+  "version": "${version}",
+  "environment": "*",
```
apple502j:
```suggestion
  "environment": "client",
```

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_QuadMixin.java
```diff
@@ -0,0 +1,54 @@
...
+import net.fabricmc.fabric.api.client.modelevents.v1.data.FaceData;
+import net.fabricmc.fabric.impl.client.modelevents.FaceDataImpl;
+import net.minecraft.client.model.ModelPart;
+import net.minecraft.util.math.Direction;
+
+@ApiStatus.Internal
+@Mixin(ModelPart.Quad.class)
+abstract class ModelPart_QuadMixin implements FaceDataImpl.Container {
+    private Direction fabric_direction;
+    private FaceDataImpl fabric_face_data;
```
apple502j: Mark as Unique (prefix unnecessary), and don't use snake case

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/CubeData.java
```diff
@@ -0,0 +1,116 @@
...
+import net.fabricmc.fabric.api.client.modelevents.v1.traversal.Traversable;
+import net.minecraft.client.model.Dilation;
+import net.minecraft.client.model.ModelPart;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.util.math.Direction;
+
+/**
+ * Data pertaining to an individual cube.
+ */
+public interface CubeData extends Traversable {
```
apple502j, line 34: NonExtendable.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/DataCollection.java
apple502j, line 50: NonExtendable.

apple502j: No, please use List here.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/FaceData.java
```diff
@@ -0,0 +1,86 @@
...
+
+import net.fabricmc.fabric.api.client.modelevents.v1.traversal.ModelVisitor;
+import net.fabricmc.fabric.api.client.modelevents.v1.traversal.Traversable;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.util.math.Direction;
+
+/**
+ * Provides data pertaining to a cubes individual face.
+ */
+public interface FaceData extends Traversable {
```
apple502j, line 32: NonExtendable

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/PartView.java
```diff
@@ -0,0 +1,84 @@
...
+
+import net.fabricmc.fabric.api.client.modelevents.v1.PartTreePath;
+import net.fabricmc.fabric.api.client.modelevents.v1.traversal.Traversable;
+import net.fabricmc.fabric.impl.client.modelevents.FabricPartHooks;
+import net.minecraft.client.model.ModelPart;
+
+/**
+ * Provides access to information about a model part as it is being rendered.
+ */
+public interface PartView extends Traversable {
```
apple502j, line 33: NonExtendable.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/traversal/Traversable.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minecraft.client.util.math.MatrixStack;
+
+/**
+ * Represents an object that can be traversed.
+ *
+ * @see PartView
+ * @see CubeData
+ * @see FaceData
+ */
+public interface Traversable {
```
apple502j, line 28: Is this intended to be extendable? If not add NonExtendable

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/FabricPartHooks.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+
+        view.dispatchEvents(matrices, vertices, light, overlay, red, green, blue, alpha);
+    }
+
+    // called recursively
+    void setPath(PartTreePathImpl path) {
+        if (settingPath) {
+            LOGGER.warn("Loop detected in model tree at $", path);
```
apple502j: Logger arguments are curly-braces

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/ModelPartCallbacksImpl.java
```diff
@@ -0,0 +1,146 @@
...
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.block.entity.BlockEntityType;
+import net.minecraft.client.render.VertexConsumer;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityType;
+
+@ApiStatus.Internal
+public final class ModelPartCallbacksImpl implements ModelPartCallbacks {
+    private static volatile long UPDATE_TIME = 0;
```
apple502j: camelCase.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/ModelPartCallbacksImpl.java
```diff
@@ -0,0 +1,146 @@
...
+            for (var event : events) {
+                matrices.push();
+                event.onModelPartRendered(part, matrices, vertices, delta, light, overlay, r, g, b, a);
+                matrices.pop();
+            }
+        };
+    }
+
+    private static ModelPartListener guardRecursion(final ModelPartListener listener) {
+        var guard = new Object() {
```
apple502j: MutableBoolean.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/FaceData.java
```diff
@@ -0,0 +1,86 @@
...
+     * @return Quaternion rotation
+     */
+    default Quaternionf rotation() {
+        return direction().getRotationQuaternion();
+    }
+
+    /**
+     * The minimum axis vector of this face.
+     */
+    Vector3f min();
```
apple502j: Unless we want users to mutate this, we should be returning the one with C.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Gets a callbacks registry to register events upon a certain part is rendered.
```
apple502j:
```suggestion
     * Gets a callbacks registry to register events triggered upon rendering a certain part.
```

Registration occurs before rendering. Also, upon is a preposition.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+     * <p>
+     * This is a short-hand method and calling this is the equivalent of calling {@code ModelPartCallbacks.get(ModelPartCallbacks.MatchingStrategy.ENDS_WITH, path)}
+     *
+     * @param path The path to check for
```
apple502j: Should be all lowercase

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+     * @return ModelPartCallbacks object for registering event listeners.
```
apple502j: Type should not be included here, nor should it end with a period.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    static ModelPartCallbacks get(PartTreePath path) {
+        return get(MatchingStrategy.ENDS_WITH, path);
+    }
+
+    /**
+     * Gets a callbacks registry to register events upon a certain part is rendered.
+     * <p>
+     * The parts that cause events registered through this callback object is determined by the tree-path
```
apple502j:
```suggestion
     * The parts that trigger events registered through this callback object is determined by the tree-path
```

Or "invoke".

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+     * and the matching strategy.
+     *
+     * @param matchingStrategy Strategy for deciding whether a model part's path matches the path specified here.
```
apple502j: param should not end with a period either

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Registers an event to be invoked when a model part is rendered.
+     */
+    void register(ModelPartListener listener);
+
+    /**
+     * Registers an event to be invoked for a model part during entity rendering.
+     *
+     * @param <T> Entity class
+     * @param entityType entity type of the entity who's model rendering must trigger this event
```
apple502j:
```suggestion
     * @param entityType entity type of the entity whose model rendering must trigger this event
```

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
```diff
@@ -0,0 +1,125 @@
...
+     * Registers an event to be invoked for a model part during block entity rendering.
+     *
+     * @param <T> BlockEntity class
+     * @param entityType block entity type of the block entity who's model rendering must trigger this event
+     * @param listener Part rendering listener
+     */
+    <T extends BlockEntity> void register(BlockEntityType<T> entityType, BlockEntityModelPartListener<T> listener);
+
+    /**
+     * Represents the strategy used when determining whether a model's path matches with the requested path for an event.
```
apple502j:
```suggestion
     * Represents the strategy used when determining whether a model's path matches the requested path for an event.
```

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/PartTreePath.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Gets the last element in this path.
+     */
+    String last();
+
+    /**
+     * Finds the position at which a specific part appears on this path.
+     * If the part cannot be found will return -1.
```
apple502j, line 46: Use `{@code }`.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/PartTreePath.java
```diff
@@ -0,0 +1,83 @@
...
+     * Checks whether any of the part names on this path matches the requested part.
+     */
+    default boolean includes(PartTreePath path) {
+        return indexOf(path) != -1;
+    }
+
+    /**
+     * Checks whether the specified part name corresponds to the first element in this path.
+     */
+    default boolean beginsWith(PartTreePath path) {
```
apple502j, line 66: The usual Java name is "startsWith".

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/FaceData.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.util.math.Direction;
+
+/**
+ * Provides data pertaining to a cubes individual face.
+ */
+public interface FaceData extends Traversable {
+    /**
+     * The physical orientation of this face.
+     *
+     * If the face is mirrored, will be the opposite of {@see lightingDirection}.
```
apple502j, line 36: This is a link, not see.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/PartView.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Bridge method for converting from a ModelPart back into a PartView.
```
apple502j: ModelPart should be linked.

apple502j (changes requested):

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/EntityRenderDispatcherMixin.java
```diff
@@ -0,0 +1,90 @@
...
+}
+/**
+ * Common injection point for rendering a block entity
+ */
+@ApiStatus.Internal
+@Mixin(value = BlockEntityRenderDispatcher.class, priority = Integer.MAX_VALUE)
+abstract class BlockEntityRenderDispatcherMixin {
+    private static final String RUN_REPORTED = "runReported(Lnet/minecraft/block/entity/BlockEntity;Ljava/lang/Runnable;)V";
+
+    @ModifyVariable(method = RUN_REPORTED, at = @At("HEAD"))
```
LlamaLad7: `argsOnly = true`

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPartMixin.java
LlamaLad7: Mixin does not have any trouble resolving `<init>`. `@Dynamic` is unnecessary.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
```diff
@@ -0,0 +1,84 @@
...
+            float textureWidth, float textureHeight,
+            Set<Direction> enabledFaces,
+            CallbackInfo info) {
+        fabric_dilation = new Dilation(extraX, extraY, extraZ);
+    }
+
+    @Override
+    public Cuboid cuboid() {
+        return (ModelPart.Cuboid)(Object)this;
+    }
```
LlamaLad7: Added members need prefixing if their descriptors do not make them inherently unique.

LlamaLad7 (changes requested):
General points: 
- Never use `MAX_VALUE` as a priority. That just makes it incredibly annoying for mods who want to be after it. I see no reason to use anything over say 2000.
- Handler methods are meant to be descriptive of their purposes. Your names just restate what is shown in the annotations already and are not useful in a stacktrace. E.g. `renderFabricPartHooks` instead of `on_render`.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
Sollace: Sir, this is an interface.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/PartTreePath.java
Sollace: Sir, this is an interface.

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
Sollace: I dunno, man. I copied this from an existing module.

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
Sollace: Does it matter? This is the testClient. It's not part of the module itself.

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
Sollace: I dunno, man. I copied this from an existing module.

fabric-model-events-api-v1/src/client/resources/fabric-model-events-api-v1.accesswidener
```diff
@@ -0,0 +1,4 @@
+accessWidener v2 named
```
apple502j, line 1: This AW does not seem to be declared in FMJ. Also, the build.gradle path is wrong.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
Syst3ms: The NonExtendable annotation marks classes that shouldn't be extended by API users, but can't be made `final`, like an interface.

fabric-model-events-api-v1/src/testmod/java/net/fabricmc/fabric/test/client/modelevents/ModelEventsTest.java
Syst3ms: Testmods often serve as coding reference for modules that aren't very well-documented.

Sollace: Client instance is static and visible through `MinecraftClient.getInstance()`. I am open to (and have put thought into) passing other values, but for client instance specifically I don't see the point if they can just get it normally.

(note adding parameters entails me going back and adding it to everywhere up the call chain and these events, and would require the user to pass it down along with the MatrixStack, if I don't just drop a `MinecraftClient.getInstance()` somewhere along the way.

Sollace: Removed the vars

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/modelevents/ModelPart_CuboidMixin.java
Sollace: I lost track of who asked for this - I have changed it to 90000 with a comment explanation. I'm always up to change it further if more people will chime in.

fabric-model-events-api-v1/src/client/resources/fabric.mod.json
Sollace: Fixed?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/traversal/Traversable.java
Sollace: At present I don't know of any use cases where someone might want to extend it. There's no reason I need it to be NonExtendable either, so I'm opting to rather not rule out the posibility for some mods to extend it for their use cases.

(Might even just get rid of it entirely in a later revision)

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/impl/client/modelevents/ModelPartCallbacksImpl.java
Sollace: Even for statics?

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/data/FaceData.java
Sollace: Ah thanks. I was looking for an immutable variant but could not find it.

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
Sollace: Reworded and added more detail as it's "parts matching a path" rather than just "a certain part".

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/ModelPartCallbacks.java
Sollace: I feel like this is... not the best?
image

Sollace: As opposed to
image

fabric-model-events-api-v1/src/client/java/net/fabricmc/fabric/api/client/modelevents/v1/PartTreePath.java
Sollace: I kind of want to distenguish this from Strings though. `pathA.beginsWith(pathB)` isn't necessarily the same as `pathA.toString().startsWith(pathB.toString())`

(former is an array-based contents comparison and the latter is per-character which may create false positives)

Sollace:
Something I'd like some feedback on:

1. Thread safety in `ModelRenderContext.java`.
  Since this is entirely connected to rendering, I would presume this is always being accessed from the same thread. It's pretty easy for me to fix this, I just want to know if this is this a real concern I should address? 
  
2. Preconditions
   In a few places I am using `Objects` and `Preconditions` for doing sanity checks. Is there a preferred alternative I should be using here instead?

3. Event parameters
   I've included what I think is a reasonable set of parameters, basically the same as what you'd need to specify when calling ModelPart#render. I recognise there might be some use in providing the `VertexConsumers` in use as well. Is there something else worth considering for this?
   
   Options are either I can add it as a parameter to the event, change the existing parameter (not really a fan as they already have a lot) or add additional getters to PartView (as that is intended to be a convenient expansion point to the api as it can serve the role as a `Event` object.

4. Cancelling
   It's technically possible to allow mods to cancel part rendering by adding a boolean return. I thought about doing that, but then again the handler has all the power to change the model part instance they're given.
   (Although hiding it may be a special case for providing specific support for cancelling as simply setting `ModelPart#hidden = true` would also prevent future invokations of the event)
   
5. Traversal
   I'm on the fence about keeping that. It's technically possible to manually iterate the data views, the traverser can just do it a little more efficiently since it has access to api internals. I'm not a fan of constructing them the way they are being done, however.
   
   The traverser does have the small advantage that it has baked-in handling for all the matrix stack transforming and rotating needed to render stuff where the part is during traversal (which is how I produced the screenshots in the OP) 

6. No lists
   The DataCollection<T> class is not a list, nor did I ever have any intention for them to be used like one (even though they have random access and iteration). They're there as a way to get the first and last element as an optional, with values computed and cached internally upon demand so we're not heavily impacting render performance.
   
   I also anticipate that someone could optimize them further (and I likely will revise it myself during the course of this PR)

Juuxel:
> In a few places I am using Objects and Preconditions for doing sanity checks. Is there a preferred alternative I should be using here instead?

Both of those are good options, the contributing guidelines mention `Objects` directly too and `Preconditions` is also widely used.

PR #3528, opened 2024-01-14
Don't tick gametests if the server isn't ticking

jaredlll08:
The vanilla code now includes that `shouldTick` call.


fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/mixin/gametest/MinecraftServerMixin.java
```diff
@@ -18,7 +18,11 @@
 
 import java.util.function.BooleanSupplier;
 
+import net.minecraft.server.ServerTickManager;
```
deirn: Needs to be placed after the org.* imports.

deirn (approved):

apple502j (approved):

modmuss50 (approved):
Great spot, thanks.

PR #3533, opened 2024-01-16
1.20.4 - Server Lifecycle Events - Save Events

MrNavaStar:
This Pr adds two events (`StartSave` & `EndSave`). These events are called when the server starts and stops its periodic saving, as well as when the sever shuts down and when the `/save-all` command is run.

Common use cases mainly focus on mods that store data in a custom way (custom files, databases, etc) and need a simple way to periodically save data to reduce data loss.

I found myself using these mixins in a lot of my projects, and figured it wouldn't hurt to share them here. Please let me know if you would like any changes or If this pr isn't something your interested in.

Cheers and happy modding!


fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerLifecycleEvents.java
```diff
@@ -107,6 +107,24 @@ private ServerLifecycleEvents() {
 		}
 	});
 
+	/**
+	 * Called before a Minecraft server begins saving data.
+	 */
+	public static final Event<StartSave> START_SAVE = EventFactory.createArrayBacked(StartSave.class, callbacks -> server -> {
```
modmuss50: Would `BEFORE_SAVE` and `AFTER_SAVE` be better? it would fit better with other events we have.

modmuss50: Please add some tests for this, see [ServerLifecycleTests.java](https://github.com/FabricMC/fabric/blob/1.20.4/fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerLifecycleTests.java)

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/MinecraftServerMixin.java
```diff
@@ -101,4 +101,14 @@ private void endResourceReload(Collection<String> collection, CallbackInfoReturn
 			return value;
 		}, (MinecraftServer) (Object) this);
 	}
+
+	@Inject(method = "save", at = @At("HEAD"))
+	private void startSave(boolean suppressLogs, boolean flush, boolean force, CallbackInfoReturnable<Boolean> cir) {
```
modmuss50, line 106: Question: Are any of these params useful? Such as `flush` or `force`, im not 100% sure what they do without looking at the code.

apple502j: `force` is whether `save-off`ed servers should save. `flush`, see https://minecraft.wiki/w/Commands/save#save-all

Syst3ms: This kind of info is definitely worth passing, `force` at the very least.

MrNavaStar: I was curious about the params, I thought they would be nice to have. wanted to discuss it. I can certainly pass this info

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerLifecycleEvents.java
MrNavaStar: oh sure thing

MrNavaStar: Sure! I have no preference on the names

MrNavaStar:
Let me know what other changes you would like, however I wont be able to get to them until tomorrow.

modmuss50 (approved):

Syst3ms (approved):

zOnlyKroks (approved):

PR #3534, opened 2024-01-16
Deprecate FabricBlockSettings & FabricItemSettings.

modmuss50:
- Deprecate FabricBlockSettings & FabricItemSettings.
- Add injected FabricItem.Settings interface replacing FabricItemSettings functionality. 
- The existing vanilla Block Settings should provide most if not all of the functionality that FabricBlockSettings provided

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
@@ -35,15 +35,9 @@ import net.fabricmc.fabric.mixin.object.builder.AbstractBlockAccessor;
 import net.fabricmc.fabric.mixin.object.builder.AbstractBlockSettingsAccessor;
 
 /**
- * Fabric's version of Block.Settings. Adds additional methods and hooks
- * not found in the original class.
- *
- * <p>Make note that this behaves slightly different from the
- * vanilla counterpart, copying some settings that vanilla does not.
- *
- * <p>To use it, simply replace Block.Settings.of() with
- * FabricBlockSettings.of().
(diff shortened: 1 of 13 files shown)


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
```diff
@@ -97,56 +91,72 @@ protected FabricBlockSettings(AbstractBlock.Settings settings) {
...
 	 */
 	@Deprecated
 	public static FabricBlockSettings of() {
 		return create();
 	}
 
+	/**
+	 * @deprecated replace with {@link AbstractBlock.Settings#copy(AbstractBlock)}
+	 */
 	public static FabricBlockSettings copyOf(AbstractBlock block) {
```
liach, line 113: Should add `@Deprecated` annotation in addition to the JD tag. Same for the other `copyOf`.

Technici4n, line 113: vanilla's copy does not copy everything I thought? worth double-checking. (they have two copy methods now?)

Shnupbups: Pretty sure they do now copy everything

apple502j: They have one that worked like the previous one and one that copies everything.

modmuss50: Yep ^

apple502j:
Incorporated into #3537.

PR #3537, opened 2024-01-17
Port to 24w03b

modmuss50:
## Changes
- Previously deprecated `fabric-containers-v0` has been removed
- FabricBlockSettings and FabricItemSettings have been deprecated, 
  - You should now use `AbstractBlock.Settings` and `Item.Settings`
  - Previous functionality has been interface injected onto the vanilla classes.
- `CustomDamageHandler` now passes the `EquipmentSlot` and a `Runnable` as the break callback.
- Major refactor to the networking APIs (see bellow)
- FabricParticleTypes now supports the `Codec` and `PacketCodec`
- Custom ingridents now provide a `PacketCodec` instead of the old read/write methods using `PacketByteBuf`'s

## Networking

Mojang heavily refactored packets in this version to use the new `PacketCodec` class, and also changed their CustomPayload class to be an almost perfect replacement for our `FabricPacket` and `PacketType` classes. All APIs that previously worked around using a standalone `PacketByteBuf` have been removed, you must now use a `CustomPayload` with `PacketCodec`'s.

Each `CustomPayload` must have a `CustomPayload.Id` and `PacketCodec` registered with the new `PayloadTypeRegistry` before you are able to send or recieve them.

```java
PayloadTypeRegistry.playC2S().register(MyExamplePayload.ID, MyExamplePayload.CODEC);

private record MyExamplePayload(String value) implements CustomPayload {
  public static final CustomPayload.Id<MyExamplePayload> ID = CustomPayload.id("fabric:c2s_play");
  public static final PacketCodec<RegistryByteBuf, MyExamplePayload> CODEC = PacketCodecs.STRING.xmap(MyExamplePayload::new, MyExamplePayload::value).cast();

  @Override
  public Id<MyExamplePayload> getId() {
    return ID;
  }
}
```

The above example shows a basic example of a pyload, that has been registered for the play phase and will be sent from the client to the server. Payload types need to be registered on both the client and server regardless of the drection they are being sent. Bi-directional packets will need to be registered twice for both directions.

Listening to the payload is very similar to the old packet based handlers, using `ServerPlayNetworking.registerGlobalReceiver`

### Removed classes:
- FabricPacket
- FutureListeners
- PacketType

A full blog post will be written closer to 1.20.5 release, detailing all of the changes.


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadPacketCodecMixin.java
```diff
@@ -0,0 +1,50 @@
...
+		this.customPayloadTypeProvider = customPayloadTypeProvider;
+	}
+
+	@WrapOperation(method = {
+			"encode(Lnet/minecraft/network/PacketByteBuf;Lnet/minecraft/network/packet/CustomPayload$Id;Lnet/minecraft/network/packet/CustomPayload;)V",
+			"decode(Lnet/minecraft/network/PacketByteBuf;)Lnet/minecraft/network/packet/CustomPayload;"
+	}, at = @At(value = "INVOKE", target = "Lnet/minecraft/network/packet/CustomPayload$1;getCodec(Lnet/minecraft/util/Identifier;)Lnet/minecraft/network/codec/PacketCodec;"))
+	private PacketCodec<B, ? extends CustomPayload> wrapGetCodec(@Coerce PacketCodec<B, CustomPayload> instance, Identifier identifier, Operation<PacketCodec<B, CustomPayload>> original, B packetByteBuf) {
+		if (customPayloadTypeProvider == null) {
+			throw new IllegalStateException("Payload codec provider is not set!");
```
Patbox: I don't think it should throw when customPayloadTypeProvider is null. I think it would be better to just skip fabric logic in this case, as there isn't much benefit and it would break if someone else created codec for any threoretical reason

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
```diff
@@ -48,33 +46,17 @@
...
-	private int hookDamage(int amount) {
+	@WrapOperation(method = "damage(ILnet/minecraft/entity/LivingEntity;Lnet/minecraft/entity/EquipmentSlot;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;damage(ILnet/minecraft/util/math/random/Random;Lnet/minecraft/server/network/ServerPlayerEntity;Ljava/lang/Runnable;)V"))
+	private void hookDamage(ItemStack instance, int amount, Random random, ServerPlayerEntity serverPlayerEntity, Runnable runnable, Operation<Void> original) {
 		CustomDamageHandler handler = ((ItemExtensions) getItem()).fabric_getCustomDamageHandler();
 
 		if (handler != null) {
-			return handler.damage((ItemStack) (Object) this, amount, fabric_damagingEntity, fabric_breakCallback);
+			// TODO 1.20.5, what to do about break callback
+			assert false;
+			amount = handler.damage((ItemStack) (Object) this, amount, serverPlayerEntity, null);
```
ErrorCraft: Can't you just pass the `Runnable` like this for now, as the previous behaviour just seemed to use the vanilla break callback?
```suggestion
			amount = handler.damage((ItemStack) (Object) this, amount, serverPlayerEntity, e -> runnable.run());
```

As a side note: deprecating the `CustomDamageHandler::damage` method and using a new one would be a bit odd, as it would still require people to upgrade to the new one, so I don't think that would be that viable? For example, this
```java
CustomDamageHandler handler = (stack, amount, entity, breakCallback) -> {
	breakCallback.accept(entity);
	return amount;
};
```
would still have to be changed to this
```java
CustomDamageHandler handler = (stack, amount, entity, breakCallback) -> {
	breakCallback.run();
	return amount;
};
```
Then again, `FabricItemSettings` is getting deprecated too in #3534, so maybe it could be a different case here?

apple502j: We don't mind breaking changes during port if it makes sense.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadPacketCodecMixin.java
modmuss50: Fixed. Thanks

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
modmuss50: This was resolved.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
```diff
@@ -48,33 +48,23 @@
...
+			// Track whether an item has been broken by custom handler
+			MutableBoolean mut = new MutableBoolean(false);
+			amount = handler.damage((ItemStack) (Object) this, amount, serverPlayerEntity, slot, () -> {
+				mut.setTrue();
+				runnable.run();
+			});
+
+			// If item is broken, there's no reason to call the original.
+			if (mut.booleanValue()) return;
 		}
```
ErrorCraft, line 65: Since we're not returning anything and using a boolean anyways, wouldn't `WrapWithCondition` be easier to use here?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
```diff
@@ -97,56 +91,72 @@ protected FabricBlockSettings(AbstractBlock.Settings settings) {
...
 	 */
 	@Deprecated
 	public static FabricBlockSettings of() {
 		return create();
 	}
 
+	/**
+	 * @deprecated replace with {@link AbstractBlock.Settings#copy(AbstractBlock)}
+	 */
 	public static FabricBlockSettings copyOf(AbstractBlock block) {
```
ErrorCraft, line 114: This should include the `@Deprecated` annotation.
```suggestion
	@Deprecated
	public static FabricBlockSettings copyOf(AbstractBlock block) {
```

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java
```diff
@@ -97,56 +91,72 @@ protected FabricBlockSettings(AbstractBlock.Settings settings) {
...
 		return new FabricBlockSettings(((AbstractBlockAccessor) block).getSettings());
 	}
 
+	/**
+	 * @deprecated replace with {@link AbstractBlock.Settings#copy(AbstractBlock)}
+	 */
 	public static FabricBlockSettings copyOf(AbstractBlock.Settings settings) {
```
ErrorCraft, line 122: Same thing here.
```suggestion
	@Deprecated
	public static FabricBlockSettings copyOf(AbstractBlock.Settings settings) {
```

ErrorCraft (changes requested):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -151,82 +102,38 @@ public static Set<Identifier> getGlobalReceivers() {
...
-	 *
-	 * @param channelName the id of the channel
-	 * @return the previous handler, or {@code null} if no handler was bound to the channel name
-	 */
-	@Nullable
-	public static PlayChannelHandler unregisterReceiver(ServerPlayNetworkHandler networkHandler, Identifier channelName) {
-		Objects.requireNonNull(networkHandler, "Network handler cannot be null");
-
-		return unwrapUntyped(ServerNetworkingImpl.getAddon(networkHandler).unregisterChannel(channelName));
+	public static <T extends CustomPayload> boolean registerReceiver(ServerPlayNetworkHandler networkHandler, CustomPayload.Id<T> type, PlayPayloadHandler<T> handler) {
```
deirn, line 125: The local registry uses `Id` while the global registry uses `Type`. Also present on the other local registries.

modmuss50: Thanks for spotting that. It now only uses Id.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/ClientConfigurationNetworkHandlerMixin.java
```diff
@@ -51,7 +51,7 @@ private void initAddon(CallbackInfo ci) {
 		this.addon.lateInit();
 	}
 
-	@Inject(method = "onReady", at = @At(value = "INVOKE", target = "Lnet/minecraft/network/ClientConnection;setPacketListener(Lnet/minecraft/network/listener/PacketListener;)V", shift = At.Shift.BEFORE))
+	@Inject(method = "onReady", at = @At(value = "INVOKE", target = "Lnet/minecraft/network/ClientConnection;transitionInbound(Lnet/minecraft/network/NetworkState;Lnet/minecraft/network/listener/PacketListener;)V", shift = At.Shift.BEFORE))
```
LlamaLad7: Minor point but the `shift` should go

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
```diff
@@ -109,17 +106,8 @@ private void disconnectAddon(CallbackInfo ci) {
...
-		if (callbacks instanceof GenericFutureListenerHolder holder) {
-			channelFuture.addListener(holder.getDelegate());
-			channelFuture.addListener(ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE);
-			ci.cancel();
-		}
-	}
-
 	@Override
-	public Collection<Identifier> getPendingChannelsNames(NetworkState state) {
+	public Collection<Identifier> getPendingChannelsNames(NetworkPhase state) {
```
LlamaLad7: Needs prefixing? Or is it too late

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadC2SPacketMixin.java
```diff
@@ -16,33 +16,42 @@
...
+	@WrapOperation(
+			method = "<clinit>",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/network/packet/CustomPayload;createCodec(Lnet/minecraft/network/packet/CustomPayload$CodecFactory;Ljava/util/List;)Lnet/minecraft/network/codec/PacketCodec;"
+			)
 	)
-	private static void readPayload(Identifier id, PacketByteBuf buf, CallbackInfoReturnable<CustomPayload> cir) {
-		cir.setReturnValue(PayloadHelper.readCustom(id, buf, MAX_PAYLOAD_SIZE, NetworkingImpl.FACTORY_RETAIN.get()));
+	private static PacketCodec<PacketByteBuf, CustomPayload> wrapCodec(CustomPayload.CodecFactory<PacketByteBuf> unknownCodecFactory, List<CustomPayload.Type<PacketByteBuf, ?>> types, Operation<PacketCodec<PacketByteBuf, CustomPayload>> original) {
```
LlamaLad7, line 44: Should probably just be an assignment in a `static` block? I feel like that'd be more stable.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadS2CPacketMixin.java
```diff
@@ -16,33 +16,51 @@
...
-	private static int MAX_PAYLOAD_SIZE;
+	@WrapOperation(
+			method = "<clinit>",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/network/packet/CustomPayload;createCodec(Lnet/minecraft/network/packet/CustomPayload$CodecFactory;Ljava/util/List;)Lnet/minecraft/network/codec/PacketCodec;",
+					ordinal = 0
+			)
+	)
+	private static PacketCodec<RegistryByteBuf, CustomPayload> wrapPlayCodec(CustomPayload.CodecFactory<RegistryByteBuf> unknownCodecFactory, List<CustomPayload.Type<RegistryByteBuf, ?>> types, Operation<PacketCodec<RegistryByteBuf, CustomPayload>> original) {
```
LlamaLad7, line 45: Definitely think these two should be assignments in a `static` block. There's currently no relation to the actual field other than the `ordinal`.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/IngredientMixin.java
```diff
@@ -51,26 +52,14 @@ private static void injectCodec(boolean allowEmpty, CallbackInfoReturnable<Codec
...
-			if (serializer == null) {
-				throw new IllegalArgumentException("Cannot deserialize custom ingredient of unknown type " + type);
-			}
-
-			cir.setReturnValue(serializer.read(buf).toVanilla());
-		} else {
-			// Reset index for vanilla's normal deserialization logic.
-			buf.readerIndex(index);
-		}
+	private static PacketCodec<RegistryByteBuf, Ingredient> useCustomIngredientPacketCodec(PacketCodec<RegistryByteBuf, Ingredient> original) {
```
LlamaLad7, line 62: Should probably just be a field assignment as well.

LlamaLad7 (changes requested):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadC2SPacketMixin.java
modmuss50: I tried this first, but the codecs have a sneaky xmap: 
Screenshot 2024-01-22 at 17 05 16

I need the codec before its xmaped

modmuss50: I could do the same as IngredientMixin using ModifyExpressionValue on the xmap? Its prob a little better?

LlamaLad7: Oh right, noted. I think the others still apply.

modmuss50: The others including the ingredients also have the xmap.

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/mixin/networking/client/ClientConfigurationNetworkHandlerMixin.java
modmuss50: Done

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ClientConnectionMixin.java
modmuss50: Done, this is impl so can be renamed freely.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadC2SPacketMixin.java
LlamaLad7: Yeah nvm you're right

LlamaLad7 (approved):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadC2SPacketMixin.java
modmuss50: No worries, thanks a lot for looking. If you want to place money on whats going to break next this might be a good start :D

LlamaLad7: This is something my new `@Expression`s would help a lot with, you'd be able to target it with something like a `@ModifyReceiver` on 
```java
"CODEC = @(?.xmap(?, ?))"
```
to make sure it's actually relevant to the field access.
Long term point of course.

modmuss50:
Going to merge this, I fully expect there to be issues so please make sure to open issues so we can fix them 👍 Thanks everyone for the help, its highly appreciated.

PR #3555, opened 2024-01-28
Fix porting issues in Dynamic Registries API

### LINKED ISSUE #3554: [1.20.4] `CommonLifecycleEvents#TAGS_LOADED` isn't called on client thread when joining a world
The callback is supposed to run with the `client` flag set to `true` when the client receives the tag sync packet from the server. 

Currently this does not happen for both joining a singleplayer world and clients connecting to a dedicated server.

The callback however does run perfectly fine with the `client` flag set to `true` after executing `/reload`.

The callback on the server thread seems to work fine (when `client` is set to `false`).

apple502j:
Resolves #3554 

The first commit (only) should be ported to 1.20.2 and 1.20.4.

Fixed bugs:

- `TAGS_LOADED` not being invoked on join (#3554, also in 1.20.2+)
- `networkCodec` being ignored in custom dynamic registry
- `SKIP_EMPTY_SYNC_REGISTRIES` being ignored in actually syncing registry & client making the registry (previously it only applied to tags)
- Custom dynamic registry tag syncing relying on class loading order


modmuss50 (approved):
Looks great many thanks.

PR #3558, opened 2024-01-31
Expose a function for querying the non-default fluid renderer

jellysquid3:
This pull request exposes the existing functionality located at `FluidRenderHandlerRegistryImpl#getOverride` so that other mods can use it without depending on internals.

The reason for this is because Sodium (see [issue](https://github.com/CaffeineMC/sodium-fabric/issues/2298)) implements its own "default" for fluids, which allows it to use per-vertex colors to create a smooth gradient across the water surface. Since it's currently not possible to detect whether the default implementation is being returned by Fabric API, we can't know when to override it with our own in rendering.

I'm not sure if this is the best solution, but it seems _relatively_ harmless as far as changes go.

### DIFF
--- a/fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
+++ b/fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
@@ -42,6 +42,15 @@ public interface FluidRenderHandlerRegistry {
 	FluidRenderHandler get(Fluid fluid);
 
 	/**
+	 * Get a {@link FluidRenderHandler} for a given Fluid, if it is not the
+	 * default implementation. Supports vanilla and Fabric fluids.
+	 *
+	 * @param fluid The Fluid.
+	 * @return The FluidRenderHandler.
+	 */
+	FluidRenderHandler getOverride(Fluid fluid);
+
+	/**
 	 * Register a {@link FluidRenderHandler} for a given Fluid.
 	 *
 	 * <p>Note that most fluids have a still and a flowing type, and a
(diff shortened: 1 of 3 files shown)


jellysquid3:
Note: This does introduce one difference, being that the renderer will now _always_ invoke the fluid handler for water/lava fluids in `FluidRendererMixin.tesselate`. I don't know what the consequences of this are exactly, but there seems to be an escape hatch to deal with the recursion and it appears to work fine.

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
```diff
@@ -41,6 +41,15 @@ public interface FluidRenderHandlerRegistry {
...
 	FluidRenderHandler get(Fluid fluid);
 
+	/**
+	 * Get a {@link FluidRenderHandler} for a given Fluid, if it is not the
+	 * default implementation. Supports vanilla and Fabric fluids.
+	 *
+	 * @param fluid The Fluid.
+	 * @return The FluidRenderHandler.
+	 */
+	FluidRenderHandler getOverride(Fluid fluid);
```
Technici4n, line 55: should this be `@Nullable`?

modmuss50 (approved):

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/api/client/render/fluid/v1/FluidRenderHandlerRegistry.java
```diff
@@ -41,6 +43,16 @@ public interface FluidRenderHandlerRegistry {
 	 */
 	FluidRenderHandler get(Fluid fluid);
```
PepperCode1, line 45: Should this be marked with `@Nullable` as well?

modmuss50: Done, thanks.

PR #3560, opened 2024-01-31
Port to 24w05a

modmuss50:


fabric-biome-api-v1/src/testmod/generated/data/fabric-biome-api-v1-testmod/tags/worldgen/biome/biome_tag_test.json
```diff
@@ -1,5 +1,4 @@
 {
-  "replace": false,
```
modmuss50, line 2: I think this is fine, as it should default to false.

PR #3563, opened 2024-02-01
Fix WorldRenderEvents.AFTER_TRANSLUCENT

### LINKED ISSUE #3561: The `AFTER_TRANSLUCENT` world render event no longer fires as of 24w05a
There was an oversight with the FAPI 24w05a port for the rendering api. The `MatrixStack#push` call that the `AFTER_TRANSLUCENT` world render event relied on was removed meaning that the event never gets fired, thus breaking all mods that rely on the event.

FAPI Version: `0.95.4+1.20.5`

kevinthegreat1:
Fix #3561

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
@@ -58,13 +58,11 @@ public abstract class WorldRendererMixin {
 	@Shadow
 	private MinecraftClient client;
 	@Unique private final WorldRenderContextImpl context = new WorldRenderContextImpl();
-	@Unique private boolean didRenderParticles;
 
 	@Inject(method = "render", at = @At("HEAD"))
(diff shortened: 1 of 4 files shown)


kevinthegreat1:
Made it work but I feel like there is still some jank going on. The test is from an existing mod on 1.20.4.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
```diff
@@ -86,6 +85,12 @@ private void afterTerrainSolid(CallbackInfo ci) {
 		WorldRenderEvents.BEFORE_ENTITIES.invoker().beforeEntities(context);
 	}
 
+	@ModifyExpressionValue(method = "render", at = @At(value = "NEW", target = "net/minecraft/client/util/math/MatrixStack"))
+	private MatrixStack setMatrixStack(MatrixStack matrixStack) {
+		context.setMatrixStack(matrixStack);
```
modmuss50, line 90: The context object is re-used, so you need to set it back to null in context.prepare

modmuss50 (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
kevinthegreat1: Done, but other objects such as `frustum` and block outline stuff such as `entity` and `blockState` are also not reset.

modmuss50 (approved):

PR #3576, opened 2024-02-08
[1.20.1] Backport mob conversion event

Syst3ms:
Required for data attachment. Testmod works fine.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/MobEntityMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/MobEntityMixin.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


modmuss50 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/MobEntityMixin.java
```diff
@@ -0,0 +1,36 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.entity.event;
+
+import com.llamalad7.mixinextras.sugar.Local;
```
modmuss50, line 19: As you are using ME, can you make sure that the module's fabric.mod.json depends on loader 0.15.6. I decided not to do the same as I did on 1.20.4 where I updated all of them.

modmuss50 (changes requested):

modmuss50 (approved):

PR #3578, opened 2024-02-09
24w06a porting fixes

### LINKED ISSUE #3554: [1.20.4] `CommonLifecycleEvents#TAGS_LOADED` isn't called on client thread when joining a world

apple502j:
Surprisingly, Resource Loader and RegSync needed zero changes with regard to vanilla RegSync feature. We can utilize this, but it is kept for later.

- Bump yarn
- Refactor some code to 1) adopt Yarn rename of "resource pack names" to IDs, 2) to match vanilla coding style (with respect to isInCreativeMode)
- Resolves #3554 
- Fixes a bug where using a custom elytra did not clear the new field, `ignoreFallDamageAboveY`.


modmuss50 (approved):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/elytra/PlayerEntityMixin.java
```diff
@@ -56,6 +61,7 @@ void injectElytraCheck(CallbackInfoReturnable<Boolean> cir) {
 
 		if (EntityElytraEvents.CUSTOM.invoker().useCustomElytra(self, false)) {
 			startFallFlying();
+			this.ignoreFallDamageAboveY = null;
```
modmuss50, line 64: Great spot with this one 👍

Issue #1049, opened 2020-08-27
Hooks to allow registration of custom tracked data fields on entities.


i509VCB:
Note half the stuff here may be gibberish, this is just a concept

# Some context

In vanilla Minecraft, tracked data is typically used to sync data values of an entity to the client.

For example this is used for tracking the amount of stingers or arrows a living entity has stuck in itself.

## The problem

However there are two fundamental issues with this system that make it harder for mods to use this system for adding custom data to track on existing entities:

Tracked data values are keyed via network id, between 0 and 254. So each entity is only allowed to have 254 registered tracked data values at once. The id of each tracked data value is assigned is based on registration order at <clinit>.

Mods CANNOT register their tracked data values before vanilla or they risk having the client crash. Typically the solution to this has been to create the tracked data fields in a mixin so load order isn't affected. And then injecting at TAIL of `initDataTracker` to start tracking those values.

So a slightly different load order can cause a tracked data value to be incorrectly tracked and cause the client to crash due to de-serializing tracked data of the wrong type. This makes the built-in tracked data system fragile if mods directly touch entity code for these matters.

Also custom entities with their own tracked data is another important aspect to consider if we wish to use a system to work around vanilla's system.

## What about working with vanilla's system

In my opinion, vanilla's system could work but we would need to do the following:
- Abstract the entire process of registering mod added tracked data onto an entity.
- In the case of mismatches, remap any mismatched custom tracked data keys (whether we do this on client or server is a different question).

### Issues
- Mods which currently mixin to add custom tracked data are effectively in the wild west till all those mods adopt this system.
- Vanilla system has a limit of 255 entries per entity. In some absurdly large modpacks this could cause issues
- Clients joining other servers with more mods than server could has tracked data id mismatches which are very hard to track.

### Summary of vanilla
The above explanation leads me to believe salvaging vanilla's system is inadequate due to size limit and reliance on int ids.

## A new subsystem

So I shall propose a concept for a separate subsystem that replicates the functionality of tracked data while being much more flexible and compatible.

### Setting course

In order for this new system to be resilient to the flaws that the vanilla data tracker system has, we should avoid use of int value keys for syncing. Unless we have a specific sync packet that will ensure we have the proper client-server mapping

### Using registry(ies)?

Possibly handle these via registries of custom tracked data values? This would allow us to use more efficient int ids for synchronizing tracked data since registry sync will handle this. However registry sync mismatches like the block id mismatch could be fatal in this case due to serialization of data.

[...]

PR #3584, opened 2024-02-11
Add a tracked data handler registry

### LINKED ISSUE #3482: Syncing for TrackedDataHandlerRegistry
Currently, `TrackedDataHandlerRegistry` does not use a registry, which means that when `TrackedDataHandler`s are registered, their numeric ID is simply incremented. As such, if the mod loading order differs between client and server, or if their mod lists don't match *exactly*, there is a very real possibility that these IDs will mismatch, corrupting any `EntityTrackerUpdateS2CPacket` that involve custom `TrackedDataHandler`s.

This issue proposes that some system be added to sync these IDs. This could be resolved by either migrating the class to a registry, or by setting up some kind of sync to synchronize these IDs on join. I personally believe using a registry would be the ideal approach here since registry syncing already exists.

haykam821:
This pull request introduces a new API for registering tracked data handlers, which reduces conflicts between mods registering tracked data handlers in an order that is inconsistent between clients and servers.

To use the API, simply replace usages of the vanilla method:

```diff
- TrackedDataHandlerRegistry.register(TRACKED_DATA_HANDLER);
+ FabricTrackedDataRegistry.registerHandler(TRACKED_DATA_HANDLER_ID, TRACKED_DATA_HANDLER);
```

An example of a mod introducing tracked data handlers is provided within the test mod. This entity uses tracked data handlers for global positions, items, and optional item groups, all of which are not available as one of the vanilla tracked data handlers.

A few remarks regarding the implementation of this API:

- Documentation has not been written yet.
- Note that the class name is `FabricTrackedDataRegistry` rather than `FabricTrackedDataHandlerRegistry` so that tracked data support can be added in the future within the same class; see #1049 for more information.
- Each registration recomputes the map used by vanilla. I'm not sure whether optimizing this behavior is worthwhile; the work required to delay computation until the map is accessed would be more invasive than the current setup.
- Mods that do not switch to this API will still work as before. Said mods will not gain the benefits of using this API and will still lack a guarantee of tracked data handlers matching between clients and servers.
- There isn't any guard against registering a tracked data handler after mod initialization. Should one be added? Currently, the unspoken idea is that mods will only use this API during mod initialization.
- Mods assuming that tracked data handler IDs are stable during mod initialization (or possibly later, per the last remark) will be broken. I assume this breakage is fine, as I doubt many mods interact with tracked data handlers in this way.

Fixes #3482


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	/**
+	 * Reorders handlers in {@link net.minecraft.entity.data.TrackedDataHandlerRegistry#DATA_HANDLERS} to have a consistent order between client and server.
+	 *
+	 * <p>The order used is the following:
+	 *
+	 * <ul>
+	 *   <li>Vanilla handlers</li>
+	 *   <li>Handlers in the Fabric API registry (sorted by ID)</li>
```
modmuss50, line 68: This is going to fall appart if you have diffrent handlers on the client and server. We have a similar issue with the blockstate ids, it works 99% of the time, but when it doesnt odd things happen.

I wonder if it would be better to actually sync the id -> int map during configuration?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,113 @@
...
+import net.minecraft.entity.data.TrackedDataHandler;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.collection.Int2ObjectBiMap;
+
+import net.fabricmc.fabric.mixin.object.builder.TrackedDataHandlerRegistryAccessor;
+
+public class FabricTrackedDataRegistryImpl {
+	private static final Logger LOGGER = LoggerFactory.getLogger(FabricTrackedDataRegistryImpl.class);
+
+	private static List<TrackedDataHandler<?>> VANILLA_HANDLERS = new ArrayList<>();
```
modmuss50, line 44: Question: How many vanilla handlers are there? I assume too many to give them their own IDs?

haykam821: There are 29 tracked data handlers in vanilla as of Minecraft snapshot 24w06a. I don't think it's necessary for these tracked data handlers to get their own identifiers, since this implementation doesn't manage them, and mismatches are separated by protocol version.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
modmuss50: We discussed on Discord about possiblying registering a full Registry and then allowing reg sync to handle the syncing of the IDs. Im not sure how we would handle mods that arent using the API, but beyond that it shouldnt be too hard. Let me know if you want some help

haykam821: I think mods that aren't using the API would just be prone to desynchronization, which matches the behavior both before and in the current state of this pull request. The use of a synced registry would just change the order of the middle section of tracked data handlers.

haykam821: Fixed in 03d9015e09128ec82fb711017b022d3270aae62b

Linguardium:
would it be beneficial to add a warning if a non-vanilla handler is registered? (outside of the api)

haykam821:
@Linguardium A warning already will be logged in that scenario.

Linguardium:
Yep, i missed that

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
```diff
@@ -0,0 +1,31 @@
...
+import net.minecraft.entity.data.TrackedDataHandler;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.object.builder.FabricTrackedDataRegistryImpl;
+
+public final class FabricTrackedDataRegistry {
+	private FabricTrackedDataRegistry() {
+	}
+
+	public static void registerHandler(Identifier id, TrackedDataHandler<?> handler) {
```
modmuss50, line 28: Needs some docs.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.collection.Int2ObjectBiMap;
+
+import net.fabricmc.fabric.api.event.registry.FabricRegistryBuilder;
+import net.fabricmc.fabric.api.event.registry.RegistryAttribute;
+import net.fabricmc.fabric.api.event.registry.RegistryIdRemapCallback;
+import net.fabricmc.fabric.mixin.object.builder.TrackedDataHandlerRegistryAccessor;
+
+public class FabricTrackedDataRegistryImpl {
```
modmuss50, line 38: Nit: final

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+			HANDLER_REGISTRY = FabricRegistryBuilder
+					.createSimple(HANDLER_REGISTRY_KEY)
+					.attribute(RegistryAttribute.SYNCED)
+					.buildAndRegister();
+
+			RegistryIdRemapCallback.event(HANDLER_REGISTRY).register(state -> reorderHandlers());
+		}
+
+		Registry.register(HANDLER_REGISTRY, id, handler);
+		reorderHandlers();
```
modmuss50, line 125: I dont think this needs to happen here, if the above change is made.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+			throw new IllegalArgumentException("Cannot register tracked data handler added via TrackedDataHandlerRegistry.register");
+		}
+
+		if (HANDLER_REGISTRY == null) {
+			HANDLER_REGISTRY = FabricRegistryBuilder
+					.createSimple(HANDLER_REGISTRY_KEY)
+					.attribute(RegistryAttribute.SYNCED)
+					.buildAndRegister();
+
+			RegistryIdRemapCallback.event(HANDLER_REGISTRY).register(state -> reorderHandlers());
```
modmuss50, line 121: Would Client/ServerPlayConnectionEvents.JOIN be a better place to do this? I dont think it would be a bad idea to sort them event when a modded one isnt registered.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/TrackedDataHandlerRegistryMixin.java
```diff
@@ -0,0 +1,34 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.entity.data.TrackedDataHandlerRegistry;
+
+import net.fabricmc.fabric.impl.object.builder.FabricTrackedDataRegistryImpl;
+
+@Mixin(TrackedDataHandlerRegistry.class)
+public class TrackedDataHandlerRegistryMixin {
```
modmuss50, line 29: Would it be worth while injecting into `register` and logging a warning (or even crashing in dev?) if a handler is registered the vanilla way? 

Logging the warning with a stacktrace would be more helpful when trying to track down mods that arent using the API.

fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/TrackStackEntity.java
```diff
@@ -0,0 +1,104 @@
...
+import net.minecraft.sound.SoundEvents;
+import net.minecraft.text.Text;
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.Formatting;
+import net.minecraft.util.Hand;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.GlobalPos;
+import net.minecraft.world.World;
+
+public class TrackStackEntity extends MobEntity {
```
modmuss50, line 43: Might be nice to have a gametest that spawns the entity? Just make sure things dont crash?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/TrackedDataHandlerRegistryMixin.java
```diff
@@ -0,0 +1,34 @@
...
+	@Inject(method = "<clinit>", at = @At("TAIL"))
+	private static void storeVanillaHandlers(CallbackInfo ci) {
```
LlamaLad7, line 31: Use a static block over an inject.

LlamaLad7 (changes requested):

fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/TrackStackEntity.java
haykam821: I'm not sure how to implement game tests, but using one here seems reasonable. The entity interaction should be simulated as well.

modmuss50: https://github.com/FabricMC/fabric/blob/892bf04af77eebeb135b611785ce6efa97ebd475/fabric-command-api-v2/src/testmod/java/net/fabricmc/fabric/test/command/EntitySelectorGameTest.java might be a good starting point. Happy to guide you through them.

The game tests only run as the server, so they wouldnt be able to check that the data is sycning correctly, but it can at least assume it.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
haykam821: If there are no tracked data handlers registered via Fabric API, the sort would have no effect.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
haykam821: The reviews seem to be ordered differently for me; are you talking about [changing the event at which reordering occurs](https://github.com/FabricMC/fabric/pull/3584#discussion_r1493778731)? The idea of reordering after registration is so that every tracked data handler registered via Fabric API is immediately available in the map. If that behavior is unnecessary, then this reordering step can be removed.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/TrackedDataHandlerRegistryMixin.java
haykam821: Note that `TheEndBiomeSourceMixin`, `RenderLayersMixin`, `DefaultAttributeRegistryMixin`, `BlockEntityRendererFactoriesMixin`, and `EntityRenderersMixin` also seem to use a `<clinit>` inject in a similar way to this mixin; should those be changed as well (at a later time)?

modmuss50:
@haykam821 are you intrested in finishing this off? I dont think there was anything too crazy left to do on it.

PepperCode1:
Superseded by #4599.

PR #3588, opened 2024-02-12
Fix empty data attachments being saved.

modmuss50:
Ensure that empty or null data attachments are never saved or read to nbt.


Syst3ms (approved):
Sneaky oversight. Just make sure to fix checkstyle before merging.

PR #3589, opened 2024-02-13
Overlay Conditions API

Apollounknowndev:
## Summary

1.20.2 added pack overlays (explained in the [23w31a changelog](https://www.minecraft.net/en-us/article/minecraft-snapshot-23w31a)), which conditionally overlay resource/data pack files from sub-directories onto the base pack. This PR adds an API for custom resource pack overlays, which allow overlays to be applied based on conditions other than just the current pack format.  

### Don't resource conditions do the same thing?

There are a few reasons why overlay conditions and resource conditions are distinct from each other and why both should exist.

- Resource conditions only apply to data pack files that are not part of dynamic registries such as advancement and recipe files. Overlay conditions can be used on *all* data pack *and* resource pack files.
- Overlay conditions can replace files in base packs. For example, a pack overlay could change a combat-based advancement based on if a mod with a special weapon is loaded. This would keep the advancement's identifier the same, meaning parent/child advancements could still connect with the advancement the same regardless of whether or not the mod with the special weapon is loaded.
- The main trade-off of overlay conditions compared to resource conditions is that overlays are loaded before the game has loaded the contents of the packs, meaning there's no way to do registry checks or tag populated checks. By default the Overlay Conditions API only has `fabric:not`, `fabric:all_of`, `fabric:any_of`, and `fabric:mod_loaded` conditions.

### Overlay condition format example
```json
{
  "pack": {
    "pack_format": 26,
    "description": "[Insert description here]"
  },
  "fabric:overlays": {
    "entries": [
      {
        "directory": "example_mod_compat",
        "type": "fabric:mod_loaded",
        "mod_id": "example_mod"
      },
      {
        "directory": "worldgen_incompatible_with_terralith",
        "type": "fabric:not",
        "condition": {
          "type": "fabric:mod_loaded",
          "mod_id": "terralith"
        }
      }
    ]
  }
}
```

### Overlay registration example
```java
private static final Identifier MOD_LOADED = new Identifier("fabric:mod_loaded");

static void init() {
	// init static
}

static {
	OverlayConditions.register(MOD_LOADED, object -> {
		JsonElement modId = JsonHelper.getElement(object, "mod_id");
		if (modId instanceof JsonPrimitive) {
			return FabricLoader.getInstance().isModLoaded(modId.getAsString());
		} else {
			throw new JsonParseException("Invalid mod_id element: " + modId);
		}
	});
}
```


apple502j:
:iapprove: Was just thinking about this idea! Thanks for PRing. Now onto the code review...

fabric-overlay-conditions-api-v1/build.gradle
```diff
@@ -0,0 +1,6 @@
+version = getSubprojectVersion(project)
```
apple502j: This should either be part of Resource Conditions or Resource Loader.

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.JsonHelper;
+
+import net.fabricmc.loader.api.FabricLoader;
+
+/**
+ * Registration and access to overlay conditions.
+ * An overlay condition is an identified {@code Predicate<JsonObject>} that can decide whether a resource pack overlay is applied.
+ * <ul>
+ *     <li>A pack.mcmeta file that contains an overlay condition will be parsed in {@link net.fabricmc.fabric.impl.overlay.conditions.OverlayConditionsMetadata#getAppliedOverlays()}.</li>
```
apple502j: Implementation should not be linked in API javadoc.

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+ *     <li>The lower-level {@link #conditionsMatch} and {@link #conditionMatches} may be useful when implementing conditions.</li>
+ *     <li>Conditions are registered with {@link #register} and queried with {@link #get}.</li>
+ * </ul>
+ *
+ * <h3>Details of the format</h3>
```
apple502j: I wonder if we can reuse vanilla code here.

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+		return REGISTERED_CONDITIONS.get(identifier);
+	}
+
+	/**
+	 * If {@code matchAll} is true, check if all the passed conditions match.
+	 * If it is false, check if at least one of the passed conditions matches.
+	 *
+	 * @throws RuntimeException If some condition failed to parse.
+	 */
+	public static boolean conditionsMatch(JsonArray conditions, boolean matchAll) throws RuntimeException {
```
apple502j: `RuntimeException` should not be declared.

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+	@Nullable
+	public static Predicate<JsonObject> get(Identifier identifier) {
+		return REGISTERED_CONDITIONS.get(identifier);
+	}
+
+	/**
+	 * If {@code matchAll} is true, check if all the passed conditions match.
+	 * If it is false, check if at least one of the passed conditions matches.
+	 *
+	 * @throws RuntimeException If some condition failed to parse.
```
apple502j: This should use `JsonParseException`.

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+
+		return matchAll;
+	}
+
+	/**
+	 * Check if the passed condition object matches.
+	 *
+	 * @throws RuntimeException If some condition failed to parse.
```
apple502j: same

fabric-overlay-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/overlay/conditions/v1/OverlayConditions.java
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	public static boolean conditionMatches(JsonObject condition) throws RuntimeException {
+		Identifier conditionId = new Identifier(JsonHelper.getString(condition, CONDITION_TYPE_KEY));
```
apple502j: Use `tryParse`.

apple502j (changes requested):
Several places in Resource Loader currently discard overlays, which also needs checking.

Apollounknowndev:
~~Marking the PR as a draft for the moment, as I plan to merge the module into the Resource Conditions API and there's two key issues to sort out as a result of that.~~

~~First, there's the confusing naming. The Resource Conditions API will contain resource conditions for loading individual files (but only some data pack files and no resource pack files) and overlay conditions for loading pack overlays. Perhaps there's a naming scheme that would work better?~~

~~Second, there's the conditions themselves. Some conditions can apply to both resource conditions and overlay conditions (`not`, `all_of`, `any_of`, `mod_loaded`) while some can only apply to resource conditions (`features_enabled`,`tags_populated`). Should these remain as separated condition maps, or should they be stored as one map with some way to indicate if a condition is only for resource conditions?~~

Upon further consideration, it'd make more sense to have overlays as a part of the Resource Loader API. I'll unmark the PR as a draft once that is done and some cases where pack overlays are scrapped get fixed.

Apollounknowndev:
Fully migrated overlay conditions over to the Resource Loader API. In addition, I've made pack overlays function in builtin resource packs (which ended up being a very small change code-wise).

Technici4n:
Resource conditions are the correct tool for the job here IMO, not a new system that is essentially a copy.

> meaning there's no way to do registry checks or tag populated checks.

That would be a user error for using these conditions, IMO. We could also throw in such cases.

Apollounknowndev:
> Resource conditions are the correct tool for the job here IMO, not a new system that is essentially a copy.

Resource conditions still only apply to some datapack files (e.g. loot tables and recipes) and not to files tied to dynamic registries (e.g. worldgen files) and any resource pack files. In addition, conditional overlays make large collections of pack files with the same resource conditions far easier to organize and handle.

Technici4n:
I mean that you should use resource conditions for the implementation of the overlay conditions, ideally. Resource conditions are fundamentally JSON predicates, which is what you have reinvented here. :P

Apollounknowndev:
Made the initial jump to Resource Conditions, will work on handling unsupported overlay conditions (such as the aforementioned `features_enabled` and `tags_populated` conditions) next.

Apollounknowndev:
Most of the unsupported overlay conditions ended up already giving a warning in the logs and returning false for when datapack contents haven't been initialized yet. I've applied it to the `features_enabled` condition as well, which should be the only condition that didn't previously have such check.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/ResourcePackProfileMixin.java
```diff
@@ -70,4 +70,5 @@ public boolean fabric_parentsEnabled(Set<String> enabled) {
 	public void fabric_setParentsPredicate(Predicate<Set<String>> predicate) {
 		this.parentsPredicate = predicate;
 	}
+
```
modmuss50: Nit: Revert this change

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
```diff
@@ -120,8 +123,17 @@ public ResourcePack open(String name) {
 
 					@Override
 					public ResourcePack openWithOverlays(String string, ResourcePackProfile.Metadata metadata) {
-						// Don't support overlays in builtin res packs.
-						return entry.getRight();
+						ModNioResourcePack pack = entry.getRight();
+						if (metadata.overlays().isEmpty()) {
+							return pack;
+						} else {
```
modmuss50: Code style nit, no need for the else block, the code is slightly easier to read without it. Also make sure that checkstyle passes. I reccomend building using Gradle and/or using an IDE plugin for it.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
```diff
@@ -216,6 +216,24 @@ public static <T> ConditionJsonProvider registryContains(RegistryKey<Registry<T>
 		return ResourceConditionsImpl.registryContains(REGISTRY_CONTAINS, registry.getValue(), entries);
 	}
 
+	/**
+	 * Creates a condition with a directory field, for use in overlay conditions.
+	 */
+	public static ConditionJsonProvider withDirectory(ConditionJsonProvider condition, String directory) {
```
modmuss50: Where is this used/tested? It might be nice to show how to datagen these if possibe? It might not be a bad idea to rename this as its only used for overlays?

fabric-resource-conditions-api-v1/src/testmod/resources/resourcepacks/overlay_test/pack.mcmeta
```diff
@@ -0,0 +1,27 @@
+{
+  "pack": {
+    "pack_format": 26,
+    "description": "Fabric Overlay Conditions Test Datapack."
+  },
+  "fabric:overlays": {
```
modmuss50, line 6: Could we add the conditions to the vanilla `overlays`? If not whats the reasoning?

Apollounknowndev: Vanilla overlays parse entries with a codec and Fabric overlays don't, so either resource conditions would need to be refactored to use codecs or there'd have to be mixin hackery to the vanilla overlays to parse both vanilla and fabric overlays at once. Separating the overlays was the cleanest implementation that I saw.

I also attempted to parse them separately but have the `overlays` key the same, but then any instance of the `overlays` key in pack.mcmeta files that didn't contain both an `entries` field and a `fabric:entries` field failed to parse.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/DefaultResourceConditions.java
Apollounknowndev: Opting to remove this as there doesn't appear to be a way to datagen `pack.mcmeta` files.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/mixin/resource/conditions/ResourcePackProfileMixin.java
```diff
@@ -0,0 +1,46 @@
...
+
+import net.minecraft.resource.ResourcePack;
+import net.minecraft.resource.ResourcePackProfile;
+
+import net.fabricmc.fabric.impl.resource.conditions.OverlayConditionsMetadata;
+
+@Mixin(ResourcePackProfile.class)
+public class ResourcePackProfileMixin {
+	@ModifyVariable(method = "loadMetadata", at = @At("STORE"))
+	private static List<String> fabric_applyOverlayConditions(List<String> overlays, @Local ResourcePack resourcePack) throws IOException {
```
LlamaLad7, line 36: Should probably be a `@ModifyArg`, don't rely on locals where unnecessary.

LlamaLad7 (changes requested):

fabric-resource-conditions-api-v1/src/testmod/resources/resourcepacks/overlay_test/pack.mcmeta
Technici4n: It should be easy enough to parse JSON with a codec

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/OverlayConditionsMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.gson.JsonParseException;
+
+import net.minecraft.resource.metadata.ResourceMetadataSerializer;
+import net.minecraft.util.JsonHelper;
+
+import net.fabricmc.fabric.api.resource.conditions.v1.ResourceConditions;
+
+public record OverlayConditionsMetadata(JsonArray overlays) {
+	private static final Pattern DIRECTORY_NAME_PATTERN = Pattern.compile("[-_a-zA-Z0-9.]+");
+	public static final ResourceMetadataSerializer<OverlayConditionsMetadata> SERIALIZER = new ResourceMetadataSerializer<>() {
```
maityyy, line 35: I don't agree with the fact that it's in the impl package because it makes working with the datagen difficult. Look at `MetadataProvider`. I believe this entire class should be in the API package

maityyy: This also means that you should replace `JsonArray overlays` with a separate class, similar to what vanilla does in `PackOverlaysMetadata`

fabric-resource-conditions-api-v1/src/testmod/resources/resourcepacks/overlay_test/pack.mcmeta
Apollounknowndev: Wouldn't that require rewriting how resource conditions are created to parse with a Codec rather than with grabbing fields from a JsonObject manually?

Technici4n: It will need a custom `Codec` implementation, but it is doable. You can obtain a JSON instance in a `Decoder` implementation via `ops.convertTo(JsonOps.INSTANCE, input)`.

Apollounknowndev:
Superceded by #3702 to target 1.20.5.

PR #3593, opened 2024-02-14
Fluid rendering fixes

### LINKED ISSUE #3592: Ice/Honey block/Slime block/Frosted Ice cause water to show flowing animation at the sides
Fabric API:

Vanilla:

it would be nice if this bug was fixed before Sodium 0.6.0

PepperCode1:
- Fix default overlay block check using `TransparentBlock` instead of `TranslucentBlock` (fix #3592)
- Fix `setBlockTransparency`/`isBlockTransparent` not being thread-safe
- Reuse the same render handler objects for water and lava
- Fix custom geometry being buffered twice if `FluidRenderHandler#renderFluid` is invoked directly
- Fix calling `FluidRenderHandler.super.renderFluid` not using passed arguments to calculate color
- Fix calling `FluidRenderHandler.super.renderFluid` more than once producing incorrect geometry
- Fix fluids with no handler never receiving water overlay instead of using default behavior
- Add static `FluidRendering#render`, which allows invoking `FluidRenderHandler#renderFluid` with a specified, possibly non-vanilla default renderer


modmuss50 (approved):

Technici4n (approved):
Looks good from a very quick look.

PR #3594, opened 2024-02-15
[1.20.5] Misc networking cleanup

modmuss50:
- Deduplicate some logic in the client play/common network addon. 
- (Slight breaking change) limit the lenght of a network channel name to 128.
- Set a limit of 8192 channel, can be configured by a system prop.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
@@ -216,6 +216,10 @@ public final class GlobalReceiverRegistry<H> {
 		if (payloadTypeRegistry.get(channelName) == null) {
 			throw new IllegalArgumentException(String.format("Cannot register handler as no payload type has been registered with name \"%s\" for %s %s", channelName, side, phase));
 		}
+
+		if (channelName.toString().length() >= 128) {
+			throw new IllegalArgumentException(String.format("Cannot register handler for channel with name \"%s\" as it exceeds the maximum length of 128 characters", channelName));
+		}
 	}
 
 	public NetworkPhase getPhase() {
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
@@ -34,6 +34,7 @@ import net.minecraft.network.NetworkSide;
 import net.minecraft.util.Identifier;
 
 public final class GlobalReceiverRegistry<H> {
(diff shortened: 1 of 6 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/GlobalReceiverRegistry.java
```diff
@@ -216,6 +216,10 @@ public void assertPayloadType(Identifier channelName) {
 		if (payloadTypeRegistry.get(channelName) == null) {
 			throw new IllegalArgumentException(String.format("Cannot register handler as no payload type has been registered with name \"%s\" for %s %s", channelName, side, phase));
 		}
+
+		if (channelName.toString().length() >= 128) {
```
haykam821: Is this constant meant to be used in a different way from `AbstractChanneledNetworkAddon.MAX_CHANNEL_NAME_LENGTH`?

modmuss50: Yes, Its possible (unlikely tbh) there may be a need for other modded clients to connect. All fabric mods will be forced to use ids that are 128 long. I believe Paper has a similar option.

haykam821: Does that mean that the Fabric API channel ID length is restricted to 128 characters, while the protocol channel ID length is restricted to the value of the system property, which is by default 128 characters? In that case, I believe the protocol restriction will allow IDs one character longer than what registration allows due to the use of `>` versus `>=`.

modmuss50: Ah, good spot. Yes you are correct. By default they should be the same, I will fix this shortly 👍

haykam821: Would it make sense for `AbstractChanneledNetworkAddon.MAX_CHANNEL_NAME_LENGTH` to have a minimum of 128 then? You would be able to express the protocol maximum length as a value that is optionally larger than the Fabric API registration maximum length but equal by default.

modmuss50: > Would it make sense for `AbstractChanneledNetworkAddon.MAX_CHANNEL_NAME_LENGTH` to have a minimum of 128 then?

Sure, I honestly doubt anyone will change things, but its easy enough to do. I have updated the PR.

haykam821: The changes look good, and you're right, I don't think any channel ID will get anywhere near 128 characters.

PR #3597, opened 2024-02-17
Remove unnecessary recreation of identifiers in VillagerTypeHelper.register

### LINKED ISSUE #3408: Unnecessary recreation of identifiers in the VillagerTypeHelper.register method
The deprecated `VillagerTypeHelper.register` method uses `new Identifier(id.toString())` rather than `id`:

https://github.com/FabricMC/fabric/blob/65c120eded0fb3d65c10cd2a616e38e22746963d/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/villager/VillagerTypeHelper.java#L57-L60

There shouldn't be any observable difference between the two arguments, but the behavior is a bit silly and seems to have been introduced in 1e232e13aed7429f4ce5dda15c3754e8b1ddadcb.

Syst3ms:
Closes #3408.


modmuss50 (commented):
Sure, barely worth the effort required 👍

PR #3606, opened 2024-02-18
Update versions

Spinoscythe:
What the title says


gradle.properties
```diff
@@ -4,62 +4,62 @@ fabric.loom.multiProjectOptimisation=true
...
+fabric-resource-loader-v0-version=1.0.0
+fabric-screen-api-v1-version=3.0.0
+fabric-screen-handler-api-v1-version=2.0.0
+fabric-sound-api-v1-version=2.0.0
+fabric-transfer-api-v1-version=5.0.0
 fabric-transitive-access-wideners-v1-version=5.0.14
-fabric-convention-tags-v1-version=1.5.10
-fabric-client-tags-api-v1-version=1.1.7
+fabric-convention-tags-v1-version=2.0.0
+fabric-client-tags-api-v1-version=2.0.0
```
modmuss50: Can you revert the changes here please, this is not right at all, and something I do when preparing a release.

gradle.properties
```diff
@@ -4,62 +4,62 @@ fabric.loom.multiProjectOptimisation=true
 
 version=0.96.3
 minecraft_version=1.20.4
-yarn_version=+build.1
-loader_version=0.15.1
+yarn_version=+build.3
+loader_version=0.15.7
```
modmuss50, line 8: This should be causing the build to fail, can you check that it builds.

modmuss50 (commented):
Im not too sure about this, I generally only update the depedencies when needed.

There are other things that should also be checked for newer versions. (e.g checkstyle, github actions).

Spinoscythe:
ok, noted

PR #3611, opened 2024-02-22
Add render sky event

AlphaMode:
Currently I would like to use such a event in galacticraft for overriding vanillas skybox above a certain height

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
@@ -224,6 +224,28 @@ public final class WorldRenderEvents {
 	});
 
 	/**
+	 * Called when rendering the world's skybox.
+	 * This happens right before terrain setup.
+	 *
+	 * <p>Use this for rendering custom sky boxes in existing dimensions,
+	 * such as rendering more stars above a current y value.
+	 *
+	 * <p>Returning false will cancel rendering the skybox.  This has no
+	 * effect on other subscribers to this event - all subscribers will always be called.
+	 */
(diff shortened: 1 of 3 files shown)


modmuss50:
Would the existing `DimensionRenderingRegistry.registerSkyRenderer` handle your use case? Its not totally clear to me why we also need this event.

AlphaMode:
> Would the existing `DimensionRenderingRegistry.registerSkyRenderer` handle your use case? Its not totally clear to me why we also need this event.

No it doesn't since I want to override the existing skybox in a dimension. Also DimensionRenderingRegistry.registerSkyRenderer doesn't pass the fog callback.
I'd also think it might be useful for the SkyRenderer to return a boolean whether to cancel vanilla rendering (for forge parity mostly but I haven't really seen a clear usecase for that yet.)
Here's a old image of my use case
image

modmuss50:
> No it doesn't since I want to override the existing skybox in a dimension.

`registerSkyRenderer` can be used to override, assuming there isnt an existing one. Happy to maybe change the API if needed.

> DimensionRenderingRegistry.registerSkyRenderer doesn't pass the fog callback.

This could be added to `WorldRenderContext`

AlphaMode:
> > No it doesn't since I want to override the existing skybox in a dimension.
> 
> `registerSkyRenderer` can be used to override, assuming there isnt an existing one. Happy to maybe change the API if needed.
> 
> > DimensionRenderingRegistry.registerSkyRenderer doesn't pass the fog callback.
> 
> This could be added to `WorldRenderContext`

Is it possible to unregister it once I'm done?

modmuss50:
> Is it possible to unregister it once I'm done?

Do you mean you want to temporarily change the sky rendering? 🤔 I havent looked at the code but you might be able to get the exisitng/default sky renderer before registering yours, you can then choose to fall back to that if you dont want to use yours?

PR #3623, opened 2024-02-29
24w09a

modmuss50:
Removed:
- FabricItemSettings (previously deprecated)
- NbtIngredient (replaced with ComponentIngredient)
- ModifyItemAttributeModifiersCallback

Disabled (for now):
- Transfer API

Breaking changes:
- FabricAdvancementProvider passes a wrapper lookup
- FabricBlockLootTableProvider passes a wrapper lookup
- FabricLanguageProvider passes a wrapper lookup


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricBlockLootTableProvider.java
```diff
@@ -72,7 +72,7 @@ public void excludeFromStrictValidation(Block block) {
 	}
 
 	@Override
-	public void accept(BiConsumer<Identifier, LootTable.Builder> biConsumer) {
+	public void accept(RegistryWrapper.WrapperLookup registryLookup, BiConsumer<Identifier, LootTable.Builder> biConsumer) {
```
maityyy, line 75: The same should be done with `FabricAdvancementProvider::generateAdvancement` as some advancement triggers now require `RegistryEntry`

modmuss50: Im going to take this opportunity to just add it to them all, there is no reason not to

Issue #3080, opened 2023-05-26
Enchantment API


Technici4n:
We need some features related to enchantments. This issue exists to collect them all in the same place, to ensure a cohesive design. If you have more suggestions please add a comment.

## Creating new enchantments
We need control over which items can be enchanted in an anvil (override `isAcceptableItem`) and in an enchantment table (problematic, involves `EnchantmentTarget`).

## Adding items to existing enchantments
Sometimes the enchantment does not allow an item. The item should be able to override this check.

## Intrinsically enchanted items
Some items should behave as if they were enchanted for gameplay reasons, for example an electric tool with a mode to enable silk touch. Using the vanilla enchantment system is problematic as it can cause xp dupe glitched with the grindstone, among other things.

## Misc hooks
- Only damageable items can be enchanted using books in an anvil. This is a problem for some mods that want some non-damageable items to be enchantable (e.g. AE2 annihilation planes).

maityyy:
`EnchantmentTarget` is а really big pain. Maybe replace it with tags?

maityyy:
#1102 was actually not bad. I like the idea of using events because they allow you to edit the list of items for an existing enchantment

Phoupraw:
Any progress with Intrinsically enchanted items?

Linguardium:
this api is mostly OBE. in 1.21+ you can supply a prefilled DataComponentTypes.ENCHANTMENT component in the item settings

Phoupraw:
> this api is mostly OBE. in 1.21+ you can supply a prefilled DataComponentTypes.ENCHANTMENT component in the item settings

You mean `stack.getDefaultComponents().get(DataComponentTypes.ENCHANTMENTS)`? That is enough for many cases, but it would be better if there is an API to allow `ComponentChanges` to modify intrinsic enchantments (stack-aware).

maityyy:
can be closed?

PR #3627, opened 2024-03-02
Enchantment API [24w03a+]

Syst3ms:
Snapshot 24w03a added a system of tags which broke a lot of the rigidity of the previous `EnchantmentTarget` system. Nonetheless, some ideas from #3080 are still worth implementing.

This PR adds several ways to modify what enchantments are and are not allowed on certain items:

- A `canBeEnchantedWith` method in `FabricItem` that allows modded items to override their behavior with regards to existing enchantments (vanilla or from other mods). There is also a `FabricItemStack` counterpart, which should ideally be called in future FAPI design/mods instead of `Enchantment#isAcceptableItem`.
- A new `ALLOW_ENCHANTING` event that allows doing the same thing using a global event-based system, useful for cases where both the items and the enchantments may be vanilla or from another mod.

When it comes to vanilla enchantments, these are not meant to replace the robust tag system that was introduced in 24w03a, which should be used instead when possible, but rather allow finer control than what the tag system allows; for example, having the receivable enchantments depend on stack data, or discriminating between enchantments that apply to the same tag (e.g. Mending and Unbreaking). This API might prove more useful in older versions that don't have the new tag system, which may be the subject of a future backport PR.

The PR also adds a new system of "intrinsic enchantments", which are a way for developers to give their items enchant-like effects while avoiding the unintended side-effects that may result from doing so using purely the vanilla system. For example, it could be useful to code electric multi-tools that can have Silk Touch-like or Fortune-like effects depending on their configuration.

TODO:
- [x] Intrinsically enchanted items (as in #3080)
- [x] Testmod (check for grindstone interaction)
- [x] Item-based overrides for modded items & vanilla/other mod enchantments?
- [x] Optional: Backport to 1.20.4 or below


maityyy:
Is "Intrinsically enchanted items" really needed in FAPI?

Syst3ms:
I think it's reasonable that mods may want to give their items gameplay properties as if they had a vanilla enchantment. For example, some electric multi-tool that has a Silk Touch effect. However, using the vanilla system is sort of messy and requires a bit of boilerplate cleanup to make sure it doesn't look out of place. Off the top of my mind, removing the glint, adding HideFlags, making sure the grindstone doesn't bug out, and perhaps tweaking anvil behavior. Nothing mods can't add themselves, but then again nothing in FAPI is.

I think it's a nice convenience feature to have.

fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/CustomDamageTest.java
```diff
@@ -79,5 +103,15 @@ public ItemStack getRecipeRemainder(ItemStack stack) {
 
 			return ItemStack.EMPTY;
 		}
+
+		@Override
+		public ItemEnchantmentsComponent getIntrinsicEnchantments(ItemStack stack) {
+			if (stack.isDamaged()) {
+				System.out.println("Damaged, adding silk touch");
```
maityyy: I think it should be `player.sendMessage` or `LOGGER.log` :D

Syst3ms: Oop, that's a debug message I forgot to remove

dhyces:
> I think it's reasonable that mods may want to give their items gameplay properties as if they had a vanilla enchantment. For example, some electric multi-tool that has a Silk Touch effect. However, using the vanilla system is sort of messy and requires a bit of boilerplate cleanup to make sure it doesn't look out of place. Off the top of my mind, removing the glint, adding HideFlags, making sure the grindstone doesn't bug out, and perhaps tweaking anvil behavior. Nothing mods can't add themselves, but then again nothing in FAPI is.
> 
> I think it's a nice convenience feature to have.

unfortunately, will become unnecessary in 1.20.5, as components are very easily leveraged to create this behavior
```java
public static final Item MY_ENCHANTED_SWORD = register("sword", new SwordItem(ToolMaterials.NETHERITE, 1, -1.3, new Item.Settings().component(DataComponentTypes.ENCHANTMENTS, ...).component(DataComponentTypes.ENCHANTMENT_GLINT, false)));
```

Syst3ms:
This is a good first approximation, however I believe this doesn't prevent incorrect interactions with grindstones (xp dupe), and it introduces some changes in anvils. To circumvent this you'd need to implement some more methods and you might need a mixin.

It is still possible to iron out all of these issues and do it using only vanilla means, but it's not completely trivial (nor do components make a big difference with NBT), hence why this convenience API exists.

modmuss50:
Is having an event and the item mehod the best way to do this? As it does mean that you have to check 2 things to know if an item can be enchanted? Im not sure there is an obvious solution, but I think its worth thinking about.

Syst3ms:
I hadn't thought about how, in its current state, devs would have to separately check both the event and the FabricItemStack method to find out whether an item can be enchanted. Technici4n suggests moving the event logic to the FabricItemStack method instead, which makes a lot of sense. This way the dev usage is very simple: just call `FabricItemStack#canBeEnchantedWith`.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EnchantmentHelperMixin.java
```diff
@@ -0,0 +1,58 @@
...
+		int intrinsicLevel = stack.getItem().getIntrinsicEnchantments(stack).getLevel(ench);
+		return Math.max(original, intrinsicLevel);
+	}
+
+	// Other injectors require a code reference to EnchantmentHelper$Consumer which would need an AW.
+	@ModifyArgs(
+			method = "forEachEnchantment(Lnet/minecraft/enchantment/EnchantmentHelper$Consumer;Lnet/minecraft/item/ItemStack;)V",
+			at = @At(value = "INVOKE", target = "Lnet/minecraft/enchantment/EnchantmentHelper$Consumer;accept(Lnet/minecraft/enchantment/Enchantment;I)V")
+	)
+	private static void iterateOverIntrinsicEnchantments(Args args, @Local(argsOnly = true) ItemStack stack) {
```
LlamaLad7: Prefer WrapOperation over ModifyArgs, much cleaner.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,80 @@
...
+	public interface AllowEnchanting {
+		/**
+		 * Checks whether an {@link Enchantment} should be applied to a given {@link ItemStack}.
+		 *
+		 * @param enchantment the enchantment that may be applied
+		 * @param target the target item
+		 * @param enchantingContext the enchanting context in which this check is made
+		 * @return {@link ActionResult#SUCCESS} if the enchantment may be applied, {@link ActionResult#FAIL} if it
+		 * may not, {@link ActionResult#PASS} to fall back to other callbacks/vanilla behavior
+		 * @see EnchantingContext
```
Technici4n: A `TriState` seems more appropriate.

Technici4n (changes requested):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EnchantmentHelperMixin.java
Syst3ms: Ok sure, I'll add the AW

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
Syst3ms: I originally started with a TriState, but FAPI is a bit inconsistent about this. If you look at EntitySleepEvents you'll see ActionResult is used in exactly the same way.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EnchantmentHelperMixin.java
LlamaLad7: Why do you need an AW 🤔

Syst3ms: `EnchantmentHelper.Consumer` is package-private

LlamaLad7: Ah it’s the receiver. You can just use `@Coerce Object` instead. But in fact I didn’t look closely enough and a single `@ModifyArg` will do. You can get all the args with that as explained in the javadocs.

Syst3ms: Oh, I forgot ModifyArg could just do that. But well, you learn a new obscure Mixin annotation every day.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
Technici4n: The problem with using `ActionResult` like this is that a number of values have unclear behavior. In general, `ActionResult` should only be used if the vanilla code expects one.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,80 @@
...
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.api.util.TriState;
+
+/**
+ * Events relating to enchantments, allowing for finer control of what enchantments can apply to different items.
+ */
+public class EnchantmentEvents {
```
apple502j: Mark as final

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,80 @@
...
+		 * may not, {@link TriState#DEFAULT} to fall back to other callbacks/vanilla behavior
+		 * @see EnchantingContext
+		 */
+		TriState allowEnchanting(
+				Enchantment enchantment,
+				ItemStack target,
+				EnchantingContext enchantingContext
+		);
+	}
+}
```
apple502j, line 84: Needs private ctor

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/EnchantmentEvents.java
```diff
@@ -0,0 +1,80 @@
...
+import net.fabricmc.fabric.api.util.TriState;
+
+/**
+ * Events relating to enchantments, allowing for finer control of what enchantments can apply to different items.
+ */
+public class EnchantmentEvents {
+	/**
+	 * An event that allows overriding whether an {@link Enchantment} can be applied to an {@link ItemStack}.
+	 *
+	 * <p>This should only be used to modify the behavior of <i>external</i> items with regards to <i>external</i> enchantments,
```
apple502j: We prefer using `em`

apple502j (changes requested):

modmuss50 (approved):

modmuss50 (approved):
Trying again in last call.

PR #3630, opened 2024-03-03
[1.20.5] Fix loot table provider not working after 24w09a port

maityyy:
Use `thenCompose` instead of `thenApply` for the loot table provider because `thenCompose` is needed when we return `CompletionStage` (`CompletableFuture`)


maityyy:
Exactly the same in [AdvancementProvider](https://github.com/FabricMC/fabric/blob/d20e1f08cb6dcc9bb0d0d0bd9891f1c20cdc35fe/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricAdvancementProvider.java#L83)

modmuss50 (approved):

Juuxel (approved):

PR #3644, opened 2024-03-12
Continuation of Fabric Data Fixer API (v1)

EthanRStokes:
This is a continuation of #2544 aimed to resolve some issues originally found.

From the previous PR, I have made the following changes
- Added an entrypoint (`fabric-data-fixer`) to register block entities and entities to the vanilla schema (applied in the 1.14 schemas, since that is around the time when Fabric started)
- Added the ability for mods to register multiple data fixers by saving and loading different keys appended to the mod id (used when saving and loading the data version)
- Updated the implementation to be compatible with the latest version of Minecraft
- Converted the mixins to more compatible alternatives, patching everything with the exception of `WORLD_GEN_SETTINGS`, and also using the power of Mixin Extras

I have also made changes to Fabric’s entity type and blockentity type builders to initialize the vanilla datafixer. These changes may not be wanted by all and can be reverted if needed.

Todo:
- [ ] Port the biome rename fix to post-1.19.3
- [ ] Add tests for the schema registration entrypoint
- [x] Remove `DataFixerEvents` in favor of the entrypoint
- [x] Add choice fixers for newly registered (block) entities to custom "vanilla" schema 1903

Any feedback on this would be greatly appreciated!


AViewFromTheTop:
Tree don't forget I was the one that found out how to get modded block entities and entities datafixed
I need clout for what I did lol

Treetrain1:
> Tree don't forget I was the one that found out how to get modded block entities and entities datafixed I need clout for what I did lol

Yes
I applaud you for originally learning about it and teaching me on the matter, however this is very off-topic.

apple502j:
One of the fundamental issues in the DataFixerUpper (which resulted in me abandoning that PR) is that there are N parallel versions to maintain. I still haven't figured out how to fix this; Quilt's API, from what I've heard, is basically abandoned.

Let's say a mod has a Metal Chest block. That block gets renamed to Iron Chest. That ID change is a mod-specific change tied to mod data version. The datafixer has to correct the block and item IDs in both vanilla and modded places. These are registered in a schema.

But wait! Each DataFixer is a combination of schemas and datafixes. If we add the Metal Chest datafix and schema to Metal Chest datafixer alone, then the datafix would not affect vanilla places. If we add it to both schemas, then the Metal Chest schema is no longer independent of the vanilla data version. That is practically same as doing a Mixin to various schema classes.

@Treetrain1 Could you explain the answer to this paradox?

apple502j:
In short: I cannot find a way to support all four at the same time without scrapping per-mod data version.

- Datafixing rename of vanilla items in vanilla chest
- Datafixing rename of vanilla items in modded chest
- Datafixing rename of modded items in vanilla chest
- Datafixing rename of modded items in modded chest

modmuss50:
An alternative solution that might cover the most common use cases could be a way to register aliases for registry entries.

Basically if you have `mymod:old` and want to migrate people to `mymod:new`. You would register the new entry as normal, but then add an alias so anything that queries the registry for `mymod:old` will just be returned the new entry.

EthanRStokes:
> An alternative solution that might cover the most common use cases could be a way to register aliases for registry entries.
> 
> Basically if you have `mymod:old` and want to migrate people to `mymod:new`. You would register the new entry as normal, but then add an alias so anything that queries the registry for `mymod:old` will just be returned the new entry.

Yes, that could definitely be used for the most common use cases. However, this doesn’t support transformations to the actual data, which may be more important than an id change, depending on how you look at it across Minecraft or mod versions.

EthanRStokes:
> In short: I cannot find a way to support all four at the same time without scrapping per-mod data version.
> 
> * Datafixing rename of vanilla items in vanilla chest
> * Datafixing rename of vanilla items in modded chest
> * Datafixing rename of modded items in vanilla chest
> * Datafixing rename of modded items in modded chest

I believe the solution to that is to register the block entities and entities into an early version of the vanilla schema, which is covered by the `fabric-data-fixer` entrypoint.

Tonight, I’m going to test the difference in behavior between registered and non-registered block entities.

EthanRStokes:
The only downside to this is that modders will have to register all of their block entities and entities for them to work with datafixers, however this is exactly what mojang does every time they add a new entity or block entity.

I don’t think modders should be *required* to register their content, but I think they should be greatly advised to.

AViewFromTheTop:
About the datafixing of Block Entities themselves brought up by apple.

I entirely agree with modmuss.
I'm quite certain this is exactly what the vanilla fixer does as the "registry" is just a map.
Looking through old schemas before telling Tree about datafixing modded things and how to make it work, 99 registers (of course) everything without a namespace and improper capitalization.
The goal of the fixer is to load almost any old version's worlds and to keep it up with the new one, so it must maintain both the old and new id.
I don't see why anyone would consider this an issue. It's how vanilla works and the proposed API will do it the same way, and I personally don't see a feasible alternative.

Treetrain1:
This is unrelated to the conversation, but would anyone know why my recent commits are not showing up here?
The branch on my repository shows them all, but I guess GitHub is acting up here

EthanRStokes:
After looking at how the schema registration works, I believe there needs to be a way to update the block entities and entities after the vanilla datafixer but before modded ones, without touching the vanilla data version

I wonder if an intermediary datafixer can resolve this in some way. Perhaps adding a sub version to the latest vanilla schema could work?

AViewFromTheTop:
> After looking at how the schema registration works, I believe there needs to be a way to update the block entities and entities after the vanilla datafixer but before modded ones, without touching the vanilla data version
> 
> I wonder if an intermediary datafixer can resolve this in some way. Perhaps adding a sub version to the latest vanilla schema could work?

If I remember correctly, sub-versions can only go up to 9. Be careful.

AViewFromTheTop:
Does anyone else here think that adding the rename fixes should be doing prior to the schema that the Block Entities/Entities are registered in, by the way?
I think that would be a solution to the whole issue apple brought up, actually. Tree did some testing and had some trouble with just registering the old and new ids.

AViewFromTheTop:
Though now that I think of it, that wouldn't run, would it.
Is there any way to denote a specific schema as having a separate data version from the rest? One pulling from Vanilla and the other from the mod? Or should there perhaps be an entire separate injected fixer inbetween two schemas solely for renaming to fulfill this purpose?
@Treetrain1 perhaps that would work.

AViewFromTheTop:
Nevermind.
@Treetrain1 Scratch the "inbetween two schemas" idea because that can't work.
But a datafixer for renaming that runs before the Vanilla fixer instead?

EthanRStokes:
I believe I ran into the same paradox I swore to destroy
I believe apple is right. There isn’t a way to sync changes across all modded fixers with the current implementation.

To fix this, I wonder if it’s possible to bundle everything into a single datafixer and sort each mod’s schemas

EthanRStokes:
This definitely isn’t going to work. I’m going to try again with a different approach, combining fixers instead of isolating them to specific mods. Might require a whole new system, especially for supporting mod-specific data versions.

apple502j:
In my view, any attempt to make datafixers work in modded environment properly would require forking DFU first. This is way outside our scope, though.

PR #3662, opened 2024-03-22
Codec-based resource conditions

Apollounknowndev:
This PR moves resource conditions to use codecs.

The `ConditionJsonProvider` interface has been removed and been replaced by `ResourceCondition`. `ResourceCondition` has a `test` method for determining whether the condition should pass and `getType`, which returns a `ResourceConditionType`. The `ResourceCondition` class also holds two codecs: `CODEC` (the base codec for a `ResourceCondition` instance) and `CONDITIONS_CODEC` (the codec used for deserializing resource condition objects in json files).

A `ResourceConditionType` has the Identifier for the resource condition (used in codec parsing) and has the Codec for the resource condition. They are used to register resource conditions with the `ResourceConditions.register` method.

**Changes to conditions themselves**
- `fabric:true` has been added, which will always pass.
- `fabric:[block/item/fluid]_tags_populated` have been removed.


modmuss50:
As this is very much a breaking change it must target 1.20.5

Apollounknowndev:
Moved to #3670 to target 1.20.5

PR #3664, opened 2024-03-24
Fully port to 24w12a

apple502j:
- **BREAKING**: Mining Level API was removed due to changes in this snapshot making it obsolete. An event to modify item components will be added later, which should replace it.
- **BREAKING**: Changed parameters passed to Loot API events. `ResourceMaanger` was omitted, although it is still possible to pass one (let me know if this is worth a ThreadLocal). `LootManager` is gone, and in the `LOAD_ALL` event it was replaced with `Registry<LootTable>`. ID is replaced with registry key. In addition, because mutating registry is pain, the loot tables are modified before addition; this means that per-table events don't get the table registry anymore.
- `VillagerInteractionRegistries#registerGiftLootTable` now take `RegistryKey` of the loot table. The previous method is now deprecated.
- Map decoration type registry is now synced.


fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
```diff
@@ -66,9 +66,9 @@ private LootTableEvents() {
 	 *
 	 * <pre>
 	 * {@code
-	 * LootTableEvents.MODIFY.register((resourceManager, lootManager, id, tableBuilder, source) -> {
+	 * LootTableEvents.MODIFY.register((key, tableBuilder, source) -> {
 	 *     // If the loot table is for the cobblestone block and it is not overridden by a user:
-	 *     if (Blocks.COBBLESTONE.getLootTableId().equals(id) && source.isBuiltin()) {
+	 *     if (Blocks.COBBLESTONE.getLootTableId().equals(key) && source.isBuiltin()) {
```
Juuxel: Is the equals check still correct?

fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/mixin/loot/LootManagerMixin.java
```diff
@@ -43,58 +42,43 @@
 /**
  * Implements the events from {@link LootTableEvents}.
  */
-@Mixin(LootManager.class)
+@Mixin(ReloadableRegistries.class)
 abstract class LootManagerMixin {
```
modmuss50: Class name needs changing

modmuss50 (commented):
Looks good, the build is failing a mixin class name needs changing.

fabric-loot-api-v2/src/main/java/net/fabricmc/fabric/api/loot/v2/LootTableEvents.java
apple502j: Technically works, but identity is the convention.

Juuxel: Nice, I was just wondering if the LHS type was still the same since `getLootTableId` still refers to an id and not a key.

apple502j: Yarn needs update :)

modmuss50 (approved):

pentyum:
I don't know how to get the mining level of a block since 1.20.5, could you please give me an example?

apple502j:
There is no mining level in 1.20.5.

PR #3666, opened 2024-03-25
Update to Loom 1.6

modmuss50:
Already used in 1.20.5


.github/workflows/release.yml
```diff
@@ -11,8 +11,6 @@ jobs:
       image: mcr.microsoft.com/openjdk/jdk:21-ubuntu
       options: --user root
     steps:
-      - run: apt update && apt install git -y && git --version
-      - run: git config --global --add safe.directory /__w/fabric/fabric
```
modmuss50: Note to self: This is still needed.

PR #3671, opened 2024-03-26
Enhance Lookup API

Phoupraw:
- Added preliminary providers, which are called before all other providers.
- You can register multiple providers for one item/block/entity type now.
- Fallback providers list is replaced by `Event`, where you can add phase ordering.
- All the changes are backward compatible. Some methods are marked with `@Deprecated`.


Technici4n:
I don't like this PR, it seems like a bunch of random opinionated changes, and many go against my original design. I would like each change to be submitted separately, and with a motivating use case for each.

PR #3676, opened 2024-03-27
24w13a

### LINKED ISSUE #3669: Random mod loading order sometimes causes the screen to not open when using an ExtendedScreenHandlerType
This is the part of the code that's causing this issue:
https://github.com/FabricMC/fabric/blob/e9d2a72b4f1ba820c88830c84e199b07b05f4ae9/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/impl/screenhandler/Networking.java#L84C1-L88C6

When the Fabric API initializes after a mod registered its (extended) screen handler, the packet codec will not be put into the `CODEC_BY_ID` map because the callback hasn't been registered yet.

One possible solution might be to create a new callback that's executed before the registry freezes to foreach all entries that are an instance of ExtendedScreenHandlerType.

Using Fabric `0.96.12+1.20.5`

modmuss50:
Big thanks to @DrexHD for most of this

Also fixes https://github.com/FabricMC/fabric/issues/3669


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
```diff
@@ -275,7 +275,8 @@ public FabricEntityTypeBuilder<T> requires(FeatureFlag... requiredFeatures) {
 	public EntityType<T> build() {
 		// Modded DFU is a dream, currently not possible without screwing it up.
 
-		return new FabricEntityType<>(this.factory, this.spawnGroup, this.saveable, this.summonable, this.fireImmune, this.spawnableFarFromPlayer, this.specificSpawnBlocks, dimensions, trackRange, trackedUpdateRate, forceTrackedVelocityUpdates, this.requiredFeatures);
+		//TODO 1.20.5, new field
+		return new FabricEntityType<>(this.factory, this.spawnGroup, this.saveable, this.summonable, this.fireImmune, this.spawnableFarFromPlayer, this.specificSpawnBlocks, dimensions, 1, trackRange, trackedUpdateRate, forceTrackedVelocityUpdates, this.requiredFeatures);
```
modmuss50, line 279: Need to figure out what this new field is.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/ingredient/CustomIngredientSerializer.java
```diff
@@ -64,7 +64,7 @@ static CustomIngredientSerializer<?> get(Identifier identifier) {
 	 * @see Ingredient#ALLOW_EMPTY_CODEC
 	 * @see Ingredient#DISALLOW_EMPTY_CODEC
 	 */
-	Codec<T> getCodec(boolean allowEmpty);
+	MapCodec<T> getCodec(boolean allowEmpty);
```
modmuss50, line 67: Note: Breaking change

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityTypeBuilder.java
modmuss50: It seems to be the inital scale, so 1 is fine for now (matches the vanilla builder). We should prob replace this whole class with injected interfaces.

PR #3677, opened 2024-03-28
Deprecate FabricEntityTypeBuilder & FabricBlockEntityTypeBuilder in favour of the vanilla classes

modmuss50:
Depends on https://github.com/FabricMC/fabric/pull/3666 due to generic interface injection


zenfyrdev:
Can this be done to FabricBlockEntityTypeBuilder too? Since there's already BlockEntityType.Builder.

modmuss50:
> Can this be done to FabricBlockEntityTypeBuilder too? Since there's already BlockEntityType.Builder.

Good idea, I'll take a look. 👍

modmuss50:
> Can this be done to FabricBlockEntityTypeBuilder too? Since there's already BlockEntityType.Builder.

Just had a look as as far as I can tell FabricBlockEntityTypeBuilder is useless? The only thing we add in FabricBlockEntityTypeBuilder is methods to add the supported blocks, but this can be set when creating the vanilla builder using `BlockEntityType.Builder.create` I think ill just deprecate it.

modmuss50:
Deprecated FabricEntityTypeBuilder and interface injected a `build()` (without the datafix type) overload.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
```diff
@@ -0,0 +1,151 @@
...
+		default EntityType.Builder<T> alwaysUpdateVelocity(boolean alwaysUpdateVelocity) {
+			throw new AssertionError("Implemented in Mixin");
+		}
+
+		/**
+		 * Build the entity type from the builder. Same as {@link EntityType.Builder#build(String)} but without an id.
+		 *
+		 * @return the entity type instance
+		 */
+		default EntityType<T> build() {
```
LlamaLad7, line 58: Both need `fabric_` prefixing.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/mixin/object/builder/EntityTypeBuilderMixin.java
```diff
@@ -0,0 +1,114 @@
...
+		return (EntityType.Builder<T>) (Object) this;
+	}
+
+	@Override
+	public EntityType<T> build() {
+		return build(null);
+	}
+
+	@Inject(method = "build", at = @At("RETURN"))
+	public void build(String id, CallbackInfoReturnable<EntityType<T>> cir) {
```
LlamaLad7: Handler methods should always be `private` and have names that describe their goal, not just a copy of the target method's name.

LlamaLad7 (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
modmuss50: No, this is part of the API.

LlamaLad7: Oh is this an existing API?

modmuss50: No, this is new (based off the old one). The vanilla build() method takes a string that will almost always be null for mods, this is an overload of the vanilla build().

LlamaLad7: Is the consensus to not prefix user-facing things and just hope other mods do, then?

modmuss50: Yes exactly.

LlamaLad7: Fair enough, will bear in mind 👍

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
```diff
@@ -0,0 +1,151 @@
...
+		 * @param <T> the type of entity
+		 *
+		 * @return a new living entity type builder
+		 */
+		static <T extends LivingEntity> EntityType.Builder<T> createLiving() {
+			return FabricEntityTypeImpl.Builder.createLiving(UnaryOperator.identity());
+		}
+
+		/**
+		 * Creates an entity type builder for a mob entity.
```
apple502j, line 77: Where do we set the constructor? `PlayerEntity` is the only one in vanilla where constructor is skipped. Otherwise client sync/biome spawn/etc wouldn't work.

apple502j (changes requested):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
modmuss50: Ah, I see good point. The previous API you would set the factory via the builder methods, of course this cannot happen now.

PR #3695, opened 2024-04-09
Allow passing multiple BiomeTags at the same time for BiomeSelectors

JT122406:
I have run into situations where with ores for example I want to add them to more than one biomeTag biomes and have to make a separate entry, this solves that by allowing developers to pass in more than one.


haykam821:
Does creating a tag that references other tags not work for your use case? In most cases, `element.isIn(ElementTags.COMBINED)` is preferred over `element.isIn(ElementTags.FIRST) || element.isIn(ElementTags.SECOND)`.

JT122406:
You could do that but I feel like it's unnecessary and you'd need to make tags for every combo you do

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
```diff
@@ -115,4 +115,16 @@ default boolean hasPlacedFeature(RegistryKey<PlacedFeature> key) {
 	 * {@return true if this biome is in the given {@link TagKey}}.
 	 */
 	boolean hasTag(TagKey<Biome> tag);
+
+	/**
+	 * {@return true if this biome is in any of the given {@link TagKey}s}.
+	 */
+	default boolean hasTag(TagKey<Biome>... tags) {
```
maityyy, line 122: It doesn't sound obvious. I think we need to add `anyTagMatch`, `allTagMatch` and `noneTagMatch`

maityyy (changes requested):

apple502j:
Hmm, can't we just use `&&` or `Predicate.and`?

maityyy:
> Hmm, can't we just use `&&` or `Predicate.and`?

We can, but we have convenient helper methods and the author suggests expanding it

modmuss50:
I'm unsure how useful this boilerplate code is, as pointed about above this is really easy to do in your own mod with `Predicate.and`.

fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext.java
JT122406: I agree I think something like this would be ideal here

maityyy: *To be clear: I'm suggesting the same thing as `Stream::anyMatch`, `Stream::allMatch`, `Stream::noneMatch`

PR #3696, opened 2024-04-09
[1.20.4] Alter StartTracking with Before and After Variants

Dragon-Seeker:
The Point of this PR is to have an After Variant for StartTracking as the current implementation calls the `StartTracking` event before any packets for the Entity's creation making such an event pointless if you need to sync things like data attachment to the Client as such will require some delayed method of handling any packet sent until the entity exists. Evidence of such an event being useful can be found within [Cardinal Components](https://github.com/Ladysnake/Cardinal-Components-API/blob/1a9bf0a0f67d0fd5f7a9195a0c927c2e5eae3443/cardinal-components-entity/src/main/java/dev/onyxstudios/cca/mixin/entity/common/MixinEntityTrackerEntry.java#L43) and [NeoForge](https://github.com/neoforged/NeoForge/blob/9f0d3d489e3746bfe5664ca989f309c261b90c89/patches/net/minecraft/server/level/ServerEntity.java.patch#L29). 

Additionally, this PR attempts to keep binary compatibility with previous API versions with such old variants extending the renaming of the old event.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
@@ -26,14 +26,25 @@ import net.fabricmc.fabric.api.event.EventFactory;
  * Events related to a tracking entities within a player's view distance.
  */
 public final class EntityTrackingEvents {
+
 	/**
 	 * An event that is called before player starts tracking an entity.
 	 * Typically, this occurs when an entity enters a client's view distance.
-	 * This event is called before the player's client is sent the entity's {@link Entity#createSpawnPacket() spawn packet}.
+	 * This event is called <strong>before</strong> the player's client is sent the entity's {@link Entity#createSpawnPacket() spawn packet}.
 	 */
-	public static final Event<StartTracking> START_TRACKING = EventFactory.createArrayBacked(StartTracking.class, callbacks -> (trackedEntity, player) -> {
-		for (StartTracking callback : callbacks) {
-			callback.onStartTracking(trackedEntity, player);
+	public static final Event<BeforeStartTracking> BEFORE_START_TRACKING = EventFactory.createArrayBacked(BeforeStartTracking.class, callbacks -> (trackedEntity, player) -> {
+		for (BeforeStartTracking callback : callbacks) {
+			callback.beforeStartTracking(trackedEntity, player);
+		}
+	});
+
+	/**
+	 * An event that is called after player starts tracking an entity.
+	 * This event is called <strong>after</strong> the player's client is sent the entity's {@link Entity#createSpawnPacket() spawn packet}.
+	 */
+	public static final Event<AfterStartTracking> AFTER_START_TRACKING = EventFactory.createArrayBacked(AfterStartTracking.class, callbacks -> (trackedEntity, player) -> {
+		for (AfterStartTracking callback : callbacks) {
+			callback.afterStartTracking(trackedEntity, player);
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
```diff
@@ -72,4 +93,23 @@ public interface StopTracking {
 
 	private EntityTrackingEvents() {
 	}
+
+	//--
+
+	/**
+	 * @deprecated Use {#BEFORE_START_TRACKING} or {@link #AFTER_START_TRACKING} instead
+	 */
+	@Deprecated(forRemoval = true)
```
modmuss50: This shouldnt be `forRemoval = true`

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
```diff
@@ -72,4 +93,23 @@ public interface StopTracking {
...
+	public static final Event<BeforeStartTracking> START_TRACKING = BEFORE_START_TRACKING;
```
modmuss50: I would prob just leave this a tottaly seperate event, and keep it the same as it was before.

modmuss50 (commented):
Seems like a reasonable change, could you also add some tests for this please.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
Dragon-Seeker: Little clarification, do you mean:
1. Keep the original Event as it was but with deprecation while also adding the specific before and after variants 
OR
2. Keep the original Event but add the after-variant as the only addition

apple502j: Think it's no. 2; start tracking and after start tracking events.

modmuss50: Hi, sorry. I was more thinking about 1:

Keep the current event untouched, no need to try and be clever about ensuring backwards compatibility just duplicate it. You would then end up with:

- BEFORE_START_TRACKING
- START_TRACKING (existing, deprecated)
- AFTER_START_TRACKING

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
Dragon-Seeker: Adjusted such

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/EntityTrackingEvents.java
Dragon-Seeker: I did ensure the code was backward compatible but made the desired adjustment to the PR.

Fuzss:
Not sure if this really needs a before variant. After all the point of the event is to sync custom data to the client which is not really possible if the client doesn't know about the entity yet.

Forge/Neo also only have the after variant (called at the end of the method instead of beginning).

So this PR would probably work just as a bugfix for moving the existing event to the end of the method instead of keeping it at the beginning.

Dragon-Seeker:
I do agree that this should frankly be just a change to push the injection of the mixin invocation at the tail of the method. Unless someone has a reason for the original behavior which doesn't seem likely.

PR #3702, opened 2024-04-10
[1.20.5] Resource Conditions in Pack Overlays

Apollounknowndev:
Supercedes #3589, rebased to target 1.20.5.

> 1.20.2 added pack overlays (explained in the [23w31a changelog](https://www.minecraft.net/en-us/article/minecraft-snapshot-23w31a)), which conditionally overlay resource/data pack files from sub-directories onto the base pack. This PR extends this system to allow for overlays based on resource conditions, which allow overlays to be applied based on conditions other than just the current pack format.

The main difference between the 1.20.4 and 1.20.5 versions of the PR is that `OverlayConditionsMetadata` has been completely refactored for the new codec-based resource conditions, which means they are parsed with a codec now. The class has also been moved to api for datagen purposes.


fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/OverlayConditionsMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+
+		return appliedOverlays;
+	}
+
+	public record Entry(ResourceCondition condition, String overlay) {
+		static final Codec<Entry> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+				ResourceCondition.CODEC.fieldOf("condition").forGetter(Entry::condition),
+				Codec.STRING.validate(OverlayConditionsMetadata::validateDirectory).fieldOf("directory").forGetter(Entry::overlay)
+		).apply(instance, Entry::new));
+		public boolean shouldApply() {
```
maityyy: Space missing above

PR #3706, opened 2024-04-12
Finish port to 1.20.5-pre1

apple502j:
**Breaking Changes**
- Particle API: `FabricParticleTypes` no longer takes a `Factory` (which was removed). Also, one method erroneously took an unused param, `codecGetter`, which has been removed. (This was introduced in #3537.)
- Content Registries: Added an injected interface `FabricBrewingRecipeRegistryBuilder`, adding back support for `Ingredient`s and additionally allowing users to specify feature flags. (The event callback is now an inner-interface `BuildCallback`.

**Other Changes**
- Yarn and installer bumped.
- Some general refactors.
- `VariantCodecs` now validates the components like the vanilla `ItemStack`.
- Javadoc fixes.

**Opinions Wanted**
- Should the builder have methods taking feature flags, or should we pass the feature set along with the builder in the event? In vanilla, it just checks whether the items are valid; we can't do that easily with `Ingredient`s, so I added the feature flag vararg - but maybe passing the features might be better.
- Should `VariantCodecs` validate components?
- How should we handle `SUCCESS_NO_ITEM_USED` action result? This is accepted and swings hands in vanilla, just like `SUCCESS`. Its only use in vanilla is sitting a pet wolf.


modmuss50:
> Should the builder have methods taking feature flags

I think you can just interface inject a getting for the feature set

> Should VariantCodecs validate components?

Yes, especially if vanilla is doing it for item stacks now

> How should we handle SUCCESS_NO_ITEM_USED action result?

Im not tottaly sure where we would need to handle it, Ideally id expect one of the methods in the enum to be used to better handle new entries?

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FabricBrewingRecipeRegistryBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.registry.entry.RegistryEntry;
+import net.minecraft.resource.featuretoggle.FeatureFlag;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * An extension of {@link BrewingRecipeRegistry.Builder} to support ingredients.
+ */
+public interface FabricBrewingRecipeRegistryBuilder {
```
modmuss50, line 32: This is great, as I said in the comment I think we should interface inject a getter for the feature set.

modmuss50 (approved):

apple502j:
As for the SUCCESS thing, because vanilla does not seem to differentiate between the two (yet) I don't know the proper response. While they are equivalent method-wise, they are also compared directly (though not for SUCCESS from what I can find).

Issue #3719, opened 2024-04-19
1.20.5-rc1 - adding modded ore crashes the game


kwpugh:
Since updating from 1.20.5-pre1 to pre3, I have been getting crashes on my three mods that add ore gen to the game.

https://gist.github.com/kwpugh/87739bee296c3d30188706eab6c0fe9f

Same occurs in pre4 and now rc1

in chatting with TelepathicGrunt, it was pointed out the modded features might not be getting registered by the api.

![Screenshot 2024-04-18 at 9 59 06 PM](https://github.com/FabricMC/fabric/assets/8809210/1b978624-6566-4ace-8ca1-d200007abe2b)

My ore gen setup is pretty standard, following the approach from KaupenJoe's ore gen tutorial

modmuss50:
Can you share your code please?

apple502j:
I think I know why. Not sure why FAPI testmod didn't catch this.

kwpugh:
https://gist.github.com/kwpugh/e95b1df33e46df050c254c88d9d6b62e

https://gist.github.com/kwpugh/35a5cbafc020649f6101cad9d69a3fb4

https://gist.github.com/kwpugh/2b318bb3315020b748153c16afdec7bc

KedisPL:
I have the same issue

KedisPL:
```
[18:27:35] [Server thread/ERROR] (Minecraft) Error executing task on Chunk source main thread executor for minecraft:overworld
 net.minecraft.ReportedException: Biome decoration
	at net.minecraft.server.level.ChunkMap.method_39934(ChunkMap.java:691) ~[minecraft-merged-2052ebfe04-1.20.5-rc2-loom.mappings.1_20_5_rc2.layered+hash.2198-v2.jar:?]
	at net.minecraft.util.thread.BlockableEventLoop.doRunTask(BlockableEventLoop.java:162) ~[minecraft-merged-2052ebfe04-1.20.5-rc2-loom.mappings.1_20_5_rc2.layered+hash.2198-v2.jar:?]
	at net.minecraft.server.level.ServerChunkCache$MainThreadExecutor.doRunTask(ServerChunkCache.java:561) ~[minecraft-merged-2052ebfe04-1.20.5-rc2-loom.mappings.1_20_5_rc2.layered+hash.2198-v2.jar:?]
[...]

PR #3721, opened 2024-04-19
Add an ore to the biome test mod

modmuss50:
Wrote this to try and reproduce #3719 I wasnt able to reproduce the issue, but having a test for this is a good idea.


apple502j:
FYI, this test wasn't necessary (desert wells did crash as well)

modmuss50:
Ah, its fine to have anyway, even as an example for others. No harm in having too many tests :D

PR #3726, opened 2024-04-20
Use knownPackInfo on mod data packs to avoid registry syncing

jacobsjo:
This PR adds `knownPackInfo` (introduced by vanilla in 24w06a) to mod data packs, and adds them to the `ClientDataPackManager`. This allows them to be skipped when sending datapack content from server to client.

- Currently using `fabric:<packid>:<mod-version>` as the packInfo.
- Changed the `onSelectKnownPacks` method in `SynchronizeRegistriesTask` to also allow the client to only know a subset of packs on the server, and only synchronize the rest, i.e. when the client doesn't have all server mods installed.
- Removed the `RegistryLoaderMixin` as vanilla currently treats all packs with `knownPackInfo` as stable in `RegistryLoader.RESOURCE_ENTRY_INFO_GETTER` (that might be unintended my Mojang, but even when fixed that would be a much better place to change; i.e. treat the `fabric` namespace as stable)

Testing this is a bit tricky. Most of the tests only show that the registries are properly in sync at the end, but don't show that some data isn't send via the network. In the testmod I have added a mixin that prints an error should the client not know all knownPacks during synchronization (to be ignored if testing with a different client). There is also logging of all common knownPacks that don't get send.

The testmod packs now override the `woods` wolf_variant with a different texture location (green head). Wolf variants are a good test as they don't affect other parts of the world and still need to be available on the client for the texture references. The `test2` builtin pack also overrides the `woods` wolf_variant with yet another texture location (pink head).

In particular, I have tested all 3 cases:

- mod present on server and client (i.e. in singleplayer):
  - The woods wolf gets a new texture with a green head.
  - With the `test2` built-in datapack (reload the world after activation), the woods wolf gets a pink head.
  - In both cases, the common knownPacks include all data packs. (no error message)
- mod present only on server (i.e. using a vanilla client):
  - the woods wolf gets the pink and black missing texture, so the registries get synced properly.
  - the server will print out the error about the client not knowing all datapacks, that is expected in this case
- mod present only on client (i.e. connecting to a vanilla server): The woods wolf keeps its normal texture, so the mod-bundled datapack gets properly disabled.


jacobsjo:
I've added knownPackInfo to built-in datapacks too (in addition to the bundled datapacks) and fixed some issues with the datapacks not getting properly disabled on the client when not all mods are present on the server and improved. I've also updated the testmod to test those cases.

The PR description has been updated to include more detail about the tests.

Patbox:
Is there need for it to be logged by default? It would be just extra mostly useless/garbage information in a lot of cases (don't all mods get added as datapacks/resourcepacks, even if they don't have any data? That means just having fabric api on would cause log spam, or any mod intended to work with vanilla clients)

modmuss50:
> Is there need for it to be logged by default? It would be just extra mostly useless/garbage information

Is that not the point of having it as debug logging? Debug logging is already very noisy so adding a few things from FAPI doesnt seem too bad to me.

jacobsjo:
I could see it becoming spamy when there are a lot of mods and many players on a server. 

Maybe it makes sense to move that logging to the client (would then only log once each)

We could also only log it (as info?) when some datapacks aren't known.

apple502j:
This is one of those PRs that I think should be tested in a big modpack, but I suppose there are hardly any.

apple502j:
Also, I think this might fail if there are 100 or so mods (not uncommon in modpack setup), as there's a cap on the amount of KPIs you can send in a packet.

Patbox:
From quickly checking the code, the limitation only exists on C2S packet (S2C one doesn't have any limits), so it should be patchable while keeping vanilla compatibility

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/SelectKnownPacksC2SPacketMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import org.spongepowered.asm.mixin.injection.Constant;
+import org.spongepowered.asm.mixin.injection.ModifyConstant;
+
+import net.minecraft.network.packet.c2s.config.SelectKnownPacksC2SPacket;
+
+import net.fabricmc.fabric.impl.resource.loader.ModResourcePackCreator;
+
+@Mixin(SelectKnownPacksC2SPacket.class)
+public class SelectKnownPacksC2SPacketMixin {
+	@ModifyConstant(method = "<clinit>", constant = @Constant(intValue = 64))
```
apple502j: Changing a constant int field has no effect.

apple502j (changes requested):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/SelectKnownPacksC2SPacketMixin.java
jacobsjo: I've just verified: Changing this method to always return 1 does cause `io.netty.handler.codec.EncoderException: 3 elements exceeded max size of: 1` so it does have an effect.

apple502j: Ah, so you're changing the method argument. Makes sense.

jacobsjo: I've changed it to `@ModifyArg` now, so its a bit easier to understand.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackCreator.java
```diff
@@ -68,12 +68,23 @@ public boolean canBeEnabledLater() {
 		}
 	};
 	public static final ModResourcePackCreator CLIENT_RESOURCE_PACK_PROVIDER = new ModResourcePackCreator(ResourceType.CLIENT_RESOURCES);
-	private static final ResourcePackPosition ACTIVATION_INFO = new ResourcePackPosition(true, ResourcePackProfile.InsertionPosition.TOP, false);
+	/**
+	 * The maximum ammount of known data packs requested from the client, including vanilla data packs.
+	 */
+	public static final int MAX_KNOWN_PACKS = Optional.ofNullable(System.getProperty("fabric-resource-loader-v0:maxKnownPacks")).map(Integer::parseInt).orElse(1024);
```
modmuss50:
```suggestion
	public static final int MAX_KNOWN_PACKS = Integer.getInteger("fabric-resource-loader-v0:maxKnownPacks", 1024);
```

fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/ClientDataPackManagerMixin.java
```diff
@@ -0,0 +1,55 @@
...
+
+	@Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/resource/VanillaDataPackProvider;createClientManager()Lnet/minecraft/resource/ResourcePackManager;"))
+	public ResourcePackManager createClientManager() {
+		return ModResourcePackUtil.createClientManager();
+	}
+
+	@ModifyReturnValue(method = "getCommonKnownPacks", at = @At("RETURN"))
+	List<VersionedIdentifier> getCommonKnownPacksReturn(List<VersionedIdentifier> original) {
+		if (original.size() > ModResourcePackCreator.MAX_KNOWN_PACKS) {
+			LOGGER.warn("Too many knownPacks: Found {}; max {}. Increase -Dfabric-resource-loader-v0:maxKnownPacks to prevent unnecesary registry syncing.", original.size(), ModResourcePackCreator.MAX_KNOWN_PACKS);
```
modmuss50:
```suggestion
			LOGGER.warn("Too many knownPacks: Found {}; max {}", original.size(), ModResourcePackCreator.MAX_KNOWN_PACKS);
```

I dont think suggesting users to set a system prop is the solution here, as it would need changing on both sides, and cannot be done in mod packs. If this does become an issue we might need to raise it, i imagine its unlikely?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/SynchronizeRegistriesTaskMixin.java
```diff
@@ -0,0 +1,73 @@
...
+
+	@Inject(method = "syncRegistryAndTags", at = @At("HEAD"))
+	public void syncRegistryAndTags(Consumer<Packet<?>> sender, Set<VersionedIdentifier> commonKnownPacks, CallbackInfo ci) {
+		LOGGER.debug("Syncronizing registries with common known packs: {}", commonKnownPacks);
+	}
+
+	@Inject(method = "sendPacket", at = @At("HEAD"), cancellable = true)
+	private void sendPacket(Consumer<Packet<?>> sender, CallbackInfo ci) {
+		if (this.knownPacks.size() > ModResourcePackCreator.MAX_KNOWN_PACKS) {
+			LOGGER.warn("Too many knownPacks: Found {}; max {}. Increase -Dfabric-resource-loader-v0:maxKnownPacks to prevent unnecesary registry syncing.", this.knownPacks.size(), ModResourcePackCreator.MAX_KNOWN_PACKS);
```
modmuss50:
```suggestion
			LOGGER.warn("Too many knownPacks: Found {}; max {}", this.knownPacks.size(), ModResourcePackCreator.MAX_KNOWN_PACKS);
```

modmuss50 (approved):
Changes look good expect a few minor things, I do feel like this is quite a risky change so we should make sure its tested to the best of our abbility and then release as a beta.

fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/ClientDataPackManagerMixin.java
jacobsjo: I think the warning is more for the case of "someone decreased it, but decreased it too much", so then changing it back on that side is all you need. I don't think a scenario with more than 1024 packs is at all likely.

But I'm happy to remove the suggestion.

modmuss50: I think the sysytem prop should be for developers not users. As I said we can keep an eye on this to see if people do hit the limit or have issues and possibly think again.

PR #3728, opened 2024-04-22
Add API to modify default item components

modmuss50:
```java
DefaultItemComponentsCallback.MODIFY.register(context -> {
	context.modify(Items.GOLD_INGOT, builder -> {
		builder.add(DataComponentTypes.ITEM_NAME, Text.literal("Fool's Gold").formatted(Formatting.GOLD));
	});
	context.modify(Items.GOLD_NUGGET, builder -> {
		builder.add(DataComponentTypes.FIREWORKS, new FireworksComponent(1, List.of(
			new FireworkExplosionComponent(FireworkExplosionComponent.Type.STAR, IntList.of(0x32a852), IntList.of(0x32a852), true, true)
		)));
	});
	context.modify(Items.BEEF, builder -> {
		// Remove the food component from beef
		builder.add(DataComponentTypes.FOOD, null);
	});
});

// Make all fireworks glint
DefaultItemComponentsCallback.AFTER_MODIFY.register(context -> {
	context.modify(DataComponentTypes.FIREWORKS, (fireworksComponent, builder) -> {
		builder.add(DataComponentTypes.ENCHANTMENT_GLINT_OVERRIDE, true);
	});
});
```

Does what it says on the tin really, allows mods to add or remove default item components.


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentsCallback.java
```diff
@@ -0,0 +1,91 @@
...
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events to modify the default {@link ComponentMap} of items.
+ */
+public interface DefaultItemComponentsCallback {
+	/**
+	 * Event used to add new data components to known items.
```
florensie: It also allows modifying and removing existing components, right?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentsCallback.java
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * Event used to modify the default data components of items after they have been modified by other mods during {@link #MODIFY}.
+	 */
+	Event<AfterModifyCallback> AFTER_MODIFY = EventFactory.createArrayBacked(AfterModifyCallback.class, listeners -> context -> {
+		for (AfterModifyCallback listener : listeners) {
+			listener.afterModify(context);
+		}
+	});
+
+	interface ModifyContext {
```
florensie: The Context interfaces should probably both also have `@FunctionalInterface`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/DefaultItemComponentImpl.java
```diff
@@ -0,0 +1,69 @@
...
+	static class AfterModifyContextImpl implements DefaultItemComponentsCallback.AfterModifyContext {
+		private static final AfterModifyContextImpl INSTANCE = new AfterModifyContextImpl();
+
+		private AfterModifyContextImpl() {
+		}
+
+		@Override
+		public <T> void modify(DataComponentType<T> type, BiConsumer<T, ComponentMap.Builder> builderConsumer) {
+			for (Item item : Registries.ITEM) {
+				if (item.getComponents().contains(type)) {
```
florensie: Should this really be filtered by FAPI instead of just firing for every item and letting the listener set their own condition? You might want to check the existence of multiple components, or modify any item without certain components.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentsCallback.java
florensie: Scratch that, I got confused

apple502j (changes requested):
It appears this recreates the builder for each invocation, wonder if we can reuse one?

Shnupbups:
In addition to being able to modify an item's default components, we should ideally also be able to modify the global `DEFAULT_ITEM_COMPONENTS` component map in `DataComponentTypes` that gets added to all items by default.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentsCallback.java
```diff
@@ -0,0 +1,91 @@
...
+import net.minecraft.component.DataComponentType;
+import net.minecraft.item.Item;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events to modify the default {@link ComponentMap} of items.
+ */
+public interface DefaultItemComponentsCallback {
```
Juuxel: This is not a callback interface so it should be a `final class DefaultItemComponentEvents`

modmuss50:
> In addition to being able to modify an item's default components, we should ideally also be able to modify the global `DEFAULT_ITEM_COMPONENTS` component map in `DataComponentTypes` that gets added to all items by default.

I think thats a little harder, as it gets copied on item construction. Thus it would need to be modified before the vanilla items. You can use the event added here to add or remove a component from every item.

modmuss50:
> It appears this recreates the builder for each invocation, wonder if we can reuse one?

Yeah, im not super happy about this, and how we loop over all the items each time for the modify event. I might re-think how this works somewhat 🤔

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/DefaultItemComponentImpl.java
modmuss50: I have changed this so you can now have a custom Predicate<Item> to match on items as you wish.

modmuss50:
Im going to put this into last call as I think its ok, would be nice if someone who plans to use this could take a look and check that it covers your usecase.

florensie:
> Im going to put this into last call as I think its ok, would be nice if someone who plans to use this could take a look and check that it covers your usecase.

This works for me.

Technici4n:
What is the point of `DefaultItemComponentsCallback.AFTER_MODIFY`? We have event priorities.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentEvents.java
```diff
@@ -0,0 +1,115 @@
...
+
+	public interface AfterModifyContext {
+		/**
+		 * Modify the default data components of any item with the specified {@link DataComponentType}.
+		 *
+		 * @param itemPredicate A predicate to match items to modify
+		 * @param builderConsumer A consumer that provides a {@link ComponentMap.Builder} to modify the item's components.
+		 * @param <T> The value type of the data component
+		 */
+		<T> void modify(Predicate<Item> itemPredicate, Consumer<ComponentMap.Builder> builderConsumer);
```
Juuxel: This `<T>` probably shouldn't be here?

Shnupbups:
Been testing this, and I think it would be nice if the `builderConsumer` was a `BiConsumer<ComponentMap.Builder, Item>` that was passed the item being modified.

So, for example, I could do
```java
Map<Item, Integer> fuelTimeMap = AbstractFurnaceBlockEntity.createFuelTimeMap();
context.modify(fuelTimeMap::containsKey, (builder, item) -> builder.add(FUEL_TIME, fuelTimeMap.get(item)));
```

Obviously this would be quite redundant when using the non-predicate version of `modify`, so perhaps that one could stick to just taking a standard `Consumer<ComponentMap.Builder>`.

Syst3ms (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DefaultItemComponentEvents.java
```diff
@@ -0,0 +1,115 @@
...
+	public static final Event<ModifyCallback> MODIFY = EventFactory.createArrayBacked(ModifyCallback.class, listeners -> context -> {
+		for (ModifyCallback listener : listeners) {
+			listener.modify(context);
+		}
+	});
+
+	/**
+	 * Event used to modify the default data components of items after they have been modified by other mods during {@link #MODIFY}.
+	 */
+	public static final Event<AfterModifyCallback> AFTER_MODIFY = EventFactory.createArrayBacked(AfterModifyCallback.class, listeners -> context -> {
```
Technici4n: Why two events?

modmuss50: The idea was to allow mods to add based on other mods additions, I think I can move it into one but with some pref overhead. As it only happens once it should be fine. It's on my todo list to look at, but 1.21 comes first

Syst3ms: From my understanding, the first is for modifying default components of specific (known) items or groups of items, whereas the second one is for modifying every default component of a certain type. Doing so before all specific items have been modified is undesirable in the great majority of cases, so splitting it into two dedicated phases makes sense. I was on the fence about whether to suggest having the javadoc clarify this in greater detail.

modmuss50: I think it can be done as one if each call to modify iterates all the items, then you can just use event priorities to ensure you go later.

modmuss50: Now 1 event, makes the code much simpler even if it is theoretically slower.

modmuss50:
> Been testing this, and I think it would be nice if the `builderConsumer` was a `BiConsumer<ComponentMap.Builder, Item>` that was passed the item being modified.
> 
> So, for example, I could do
> 
> ```java
> Map<Item, Integer> fuelTimeMap = AbstractFurnaceBlockEntity.createFuelTimeMap();
> context.modify(fuelTimeMap::containsKey, (builder, item) -> builder.add(FUEL_TIME, fuelTimeMap.get(item)));
> ```
> 
> Obviously this would be quite redundant when using the non-predicate version of `modify`, so perhaps that one could stick to just taking a standard `Consumer<ComponentMap.Builder>`.

Done 👍

Shnupbups (approved):
Would it be worth adding a `modify` that takes a `Collection<Item>` and/or one that takes an `Item...` varargs? I suppose it's not really needed since modders can easily just pass `items::contains` as the predicate, but might be nice to have just for simplicity's sake.

modmuss50:
> Collection<Item>

Sure, ill add a `Collection<Item>` I dont think the varargs is really needed, as you can just do `List.of(item1, item2)`

PR #3734, opened 2024-04-26
Add specific error when encoding packet with no encoder

Moulberry:
Adds a specific error when encoding a packet that doesn't have a registered encoder

To explain a bit more about the issue:
CustomPayload#getCodec returns a dummy decoder using unknownCodecFactory. This is fine when decoding, but not when encoding. When encoding, you'll end up with a confusing exception about casting to DiscardedPayload (or UnknownCustomPayload).

An ideal solution would not use #getCodec for encode, and would instead throw the more helpful exception directly when a codec is missing. Unfortunately I couldn't figure out a good way to do this since map is part of the anonymous class (perhaps someone with more knowledge of mixins could point me in the right direction).

Overwriting the entire #getCodec method is also an option, but I think is generally unfavoured.

For this PR, I just throw the error if the namespace isn't "minecraft", which might be problematic if mods add directly to the map instead of using Fabric's system.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadPacketCodecMixin.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadPacketCodecMixin.java
@@ -46,15 +46,31 @@ public abstract class CustomPayloadPacketCodecMixin<B extends PacketByteBuf> imp
 	}
 
 	@WrapOperation(method = {
-			"encode(Lnet/minecraft/network/PacketByteBuf;Lnet/minecraft/network/packet/CustomPayload$Id;Lnet/minecraft/network/packet/CustomPayload;)V",
 			"decode(Lnet/minecraft/network/PacketByteBuf;)Lnet/minecraft/network/packet/CustomPayload;"
 	}, at = @At(value = "INVOKE", target = "Lnet/minecraft/network/packet/CustomPayload$1;getCodec(Lnet/minecraft/util/Identifier;)Lnet/minecraft/network/codec/PacketCodec;"))
-	private PacketCodec<B, ? extends CustomPayload> wrapGetCodec(@Coerce PacketCodec<B, CustomPayload> instance, Identifier identifier, Operation<PacketCodec<B, CustomPayload>> original, B packetByteBuf) {
+	private PacketCodec<B, ? extends CustomPayload> wrapGetCodecDecode(@Coerce PacketCodec<B, CustomPayload> instance, Identifier identifier, Operation<PacketCodec<B, CustomPayload>> original, B packetByteBuf) {
 		if (customPayloadTypeProvider != null) {
 			CustomPayload.Type<B, ? extends CustomPayload> payloadType = customPayloadTypeProvider.get(packetByteBuf, identifier);
 
 			if (payloadType != null) {
 				return payloadType.codec();
+			}
+		}
+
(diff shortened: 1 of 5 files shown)


Patbox:
This does break existing mods such as polymer library (it injects into getCodec method) and carpet (adds to vanilla list). It also requires more specific mixins to work around it.

modmuss50:
Hi, im just testing out this PR and im not sure its working as expected? Running `/networktestcommand sendMissingCodec` shows me the following error screen? Putting a breakpoint in the `encode` doesnt seem to trigger.

Screenshot 2024-05-01 at 09 27 43

modmuss50:
You might be able to add a unit test for this, see `PayloadTypeRegistryTests` for an example.

Moulberry:
> Hi, im just testing out this PR and im not sure its working as expected? Running `/networktestcommand sendMissingCodec` shows me the following error screen? Putting a breakpoint in the `encode` doesnt seem to trigger.

The current behaviour (which may be possible to change) is that attempting to encode a packet without a codec is still an error, so the client will be disconnected.
The difference is that the logs will show the new RuntimeException ("has no registered codec" etc.), instead of a ClassCastException to DiscardedPayload.

The thrown exception is caught somewhere higher up and the client is disconnected with the "Failed to encode packet" error message, with our RuntimeException being a child.

Below is what appears in the logs (at least on my machine) when running `/networktestcommand sendMissingCodec`:
```
[20:16:50] [Netty Server IO #1/ERROR] (Minecraft) Error sending packet clientbound/minecraft:custom_payload
io.netty.handler.codec.EncoderException: Failed to encode packet 'clientbound/minecraft:custom_payload' (fabric-networking-api-v1-testmod:no_codec)
        at net.minecraft.network.handler.PacketCodecDispatcher.handler$zcm000$fabric-networking-api-v1$encode(PacketCodecDispatcher.java:547) ~[minecraft-common-38e832ecca-1.20.5-net.fabricmc.yarn.1_20_5.1.20.5+build.1-v2.jar:?]
        at net.minecraft.network.handler.PacketCodecDispatcher.encode(PacketCodecDispatcher.java:55) ~[minecraft-common-38e832ecca-1.20.5-net.fabricmc.yarn.1_20_5.1.20.5+build.1-v2.jar:?]
        ~~SNIP~~
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[netty-common-4.1.97.Final.jar:?]
        at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]
Caused by: java.lang.RuntimeException: Custom payload 'fabric-networking-api-v1-testmod:no_codec' has no registered codec
        at net.minecraft.network.packet.UnknownCustomPayload$Anonymous$a4ca0709779d43999db59af12d303b0c.encode(UnknownCustomPayloadMixin.java:41) ~[?:?]
        at net.minecraft.network.packet.UnknownCustomPayload$Anonymous$a4ca0709779d43999db59af12d303b0c.encode(UnknownCustomPayloadMixin.java:33) ~[?:?]
        ~~SNIP~~
        at net.minecraft.network.handler.PacketCodecDispatcher.encode(PacketCodecDispatcher.java:53) ~[minecraft-common-38e832ecca-1.20.5-net.fabricmc.yarn.1_20_5.1.20.5+build.1-v2.jar:?]
        ... 31 more
```

I can't say for sure why the breakpoint isn't working, but I'd think it has something to do with the mixin+anonymous class?

> You might be able to add a unit test for this, see PayloadTypeRegistryTests for an example.

I'll give it a shot!

PR #3735, opened 2024-04-26
Always get the latest ServerPlayerEntity from ServerPlayNetworkHandler

modmuss50:
The player instance in `ServerPlayNetworkHandler` is replaced when the player dies and respawns.


mysticdrew (approved):
👍

PR #3744, opened 2024-04-30
Support extending dynamic registries in datagen

SquidDev:
Data generators are passed a `CompletableFuture<RegistryWrapper.WrapperLookup>` which contains the default contents of vanilla's dynamic registries (and mods', if they implement `buildRegistry`). However, there are times when it is useful to extend this registry, for instance when generating multiple datapacks.

In vanilla, this is done by constructing a new `RegistryBuilder`, and passing that and the existing registries to `ExperimentalRegistriesValidator.validate`[^1]. However, there's a couple of issues in Fabric which prevent this from working — this PR attempts to resolve those.

 - `RegistryBuilder.createWrapperLookup` expects all registries to be present in the base registries (as it calls `getWrapperOrThrow` rather than `getWrapper`[^2]). This is true for vanilla's dynamic registries, but *not* custom Fabric ones. To resolve that, we ensure all dynamic registries are present (though possibly empty) in the initial datagen registries.

 - Dynamic registries are not added to `CloneableRegistries` in `ExperimentalRegistriesValidator.validate`. We simply replace the usage of vanilla's `DYNAMIC_REGISTRIES` with Fabric's own collection.

 - In order to extend the registries, we need access to the original `CompletableFuture<RegistryWrapper.WrapperLookup>`. However, this is only exposed when adding a pack, and is not generally available. This PR adds a getter for the registry future.

   I do have some mixed feelings about this change — it does feel a bit error prone to have both `RegistryDependentFactory` and this, but the alternatives I can think of feel more clunky.

[^1]: I think the Yarn name is a little misleading here, but I've spent too much time in the Mojmap version of this code, that I'm not sure I'd feel comfortable suggesting alternatives.
[^2]: The "correct" fix would be to use `getWrapper` here, [which is what NeoForge does](https://github.com/neoforged/NeoForge/blob/92b21d1e38d7bf23d19b1b96aaf69b8b65398437/patches/net/minecraft/core/RegistrySetBuilder.java.patch#L18-L35). However, doing that in Mixins is quite tricky!

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/FabricDataGenerator.java
@@ -25,6 +25,9 @@ import org.jetbrains.annotations.ApiStatus;
 import net.minecraft.SharedConstants;
 import net.minecraft.data.DataGenerator;
 import net.minecraft.data.DataProvider;
+import net.minecraft.registry.BuiltinRegistries;
+import net.minecraft.registry.ExperimentalRegistriesValidator;
+import net.minecraft.registry.RegistryBuilder;
 import net.minecraft.registry.RegistryWrapper;
 import net.minecraft.util.Identifier;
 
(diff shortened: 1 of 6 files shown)


modmuss50 (approved):

SquidDev:
Thank you!

PR #3749, opened 2024-05-01
Support loading a single resource condition instead of array

Apollounknowndev:
This PR adds the ability for resource conditions to be parsed from a single condition rather than an array of conditions. This can be convenient if you only have a single condition to check against.
```json
{
  ...
  "fabric:load_conditions": {
    "values": [
      "modid"
    ]
  }
}
```

Technical details:
There's a new codec added to `ResourceCondition`: `CONDITIONS_CODEC`. It accepts either a resource condition or a list of resource conditions, and if it is a list it will wrap up the list in a `fabric:and` resource condition.

### DIFF
--- a/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceCondition.java
+++ b/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceCondition.java
@@ -38,6 +38,11 @@ public interface ResourceCondition {
 	 * A codec for a list of conditions.
 	 */
 	Codec<List<ResourceCondition>> LIST_CODEC = CODEC.listOf();
+	/**
+	 * A codec for parsing load conditions. Can either be an array of conditions or a single condition.
+	 */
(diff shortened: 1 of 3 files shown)


modmuss50 (approved):
LGTM, might be nice to expand the test to cover the array case for completeness sake.

fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/api/resource/conditions/v1/ResourceCondition.java
```diff
@@ -38,6 +38,10 @@ public interface ResourceCondition {
 	 * A codec for a list of conditions.
 	 */
 	Codec<List<ResourceCondition>> LIST_CODEC = CODEC.listOf();
+	/**
+	 * A codec for parsing load conditions. Can either be an array of conditions or a single condition.
+	 */
+	Codec<ResourceCondition> CONDITION_CODEC = Codec.withAlternative(CODEC, CODEC.listOf(), conditions -> ResourceConditions.and(conditions.toArray(new ResourceCondition[0])));
```
Juuxel:
```suggestion
	Codec<ResourceCondition> CONDITION_CODEC = Codec.withAlternative(CODEC, LIST_CODEC, conditions -> ResourceConditions.and(conditions.toArray(new ResourceCondition[0])));
```
Might as well use the constant for `listOf` since we have it in the other field

Juuxel (approved):

PR #3752, opened 2024-05-02
Registry resource conditions

### LINKED ISSUE #3632: [Suggestion] Resource Conditions support for dynamic registries
Currently, I don't think there's anything for this functionality. Imo, it'd be very neat to have this, for stuff like biomes.
This'll probably apply to blocks in the future too, so I'd say it's a priority when that happens.

SquidDev:
This adds support for resource conditions within dynamic registries.

The registry-contains condition (and all other conditions) have access to the parent registries  but *not* registries in the current `ServerDynamicRegistryType`. In practice, this means conditions can reference the `STATIC` registries, but not dynamic registries (`WORLDGEN`) or loot tables (`RELOADABLE`).

For datagen, this adds additional overrides to `FabricDynamicRegistryProvider.Entries` to add items with resource conditions. This is a bit of a divergence from the other resource condition generators (which use `FabricDataGenHelper.consumeConditions`), but I think this makes more sense here.

```java
entries.add(
  RegistryKey.of(MY_REGISTRY, new Identifier("my_mod", "my_object")), new MyObject(),
  ResourceConditions.allModsLoaded("another_mod")
);
```

Closes #3632.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
@@ -31,6 +31,7 @@ import com.mojang.serialization.DynamicOps;
 import com.mojang.serialization.Encoder;
 import com.mojang.serialization.JsonOps;
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,6 +53,8 @@ import net.minecraft.world.gen.feature.PlacedFeature;
 
 import net.fabricmc.fabric.api.datagen.v1.FabricDataOutput;
 import net.fabricmc.fabric.api.event.registry.DynamicRegistries;
+import net.fabricmc.fabric.api.resource.conditions.v1.ResourceCondition;
+import net.fabricmc.fabric.impl.datagen.FabricDataGenHelper;
 import net.fabricmc.fabric.impl.registry.sync.DynamicRegistriesImpl;
 
 /**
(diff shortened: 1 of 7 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
```diff
@@ -166,11 +214,14 @@ <T> RegistryEntries<T> getQueuedEntries(RegistryKey<T> key) {
 		}
 	}
 
+	private record EntryWithConditions<T>(T value, @Nullable ResourceCondition... conditions) {
```
maityyy: `ConditionalEntry` sounds better IMO but doesn't matter

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
```diff
@@ -227,21 +274,31 @@ private <T> CompletableFuture<?> writeRegistryEntries(DataWriter writer, Registr
...
 			LOGGER.error("Couldn't serialize element {}: {}", path, error);
 		});
 
 		if (optional.isPresent()) {
-			return DataProvider.writeToPath(cache, optional.get(), path);
+			JsonElement jsonElement = optional.get();
+
+			if (conditions != null && conditions.length > 0) {
+				if (!jsonElement.isJsonObject()) {
+					LOGGER.error("Cannot add conditions to {}: JSOn is a non-object value", path);
```
modmuss50:
```suggestion
					LOGGER.error("Cannot add conditions to {}: JSON is a non-object value", path);
```

modmuss50: I wonder if this would be better off to throw? I dont see a reason not to.

modmuss50 (approved):

modmuss50:
Looks good, ill add to last call, 1 or 2 trivial things to be looked at before merging. Thanks :)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
SquidDev: Agreed! I based this off the fact that `writeToPath` logs an error if it cannot serialise, but maybe that should throw too?

modmuss50: Humm, I see. I dont think we should change DataProvider to throw (even though I think it should). But we should throw here, its much better to force the mod dev to fix their issue than lead to possible issues at runtime.

apple502j:
Are there any registries where this might fail? I don't remember if there was one that didn't tolerate additional keys, probably zero but maybe we should check,

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java
SquidDev: Oh, I meant this one within `FabricDynamicRegistryProvider.writeToPath`:

https://github.com/FabricMC/fabric/blob/1997dc55222c6bc9183dbf592672a56189a24a14/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricDynamicRegistryProvider.java#L239-L241

modmuss50: Oh right, this does match vanilla, I dont think there is a strong need to change existing stuff as part of this PR.

Im happy either way with the PR, ill let you choose what you think is best.

SquidDev:
Thanks!

PR #3758, opened 2024-05-07
Add c:indestructible tag

Dev0Louis:
This PR adds the c:indestructible tag to items. 

This tag is there to represent items which shall not be destroyed. 

I don't know if c:indestructible is the correct naming. But I don't have a better idea. 

I have not added a block tag. As I would be unsure if bedrock would go into it? For some mods bedrock may be considered as destructible. 

A different idea would be to use fabric:indestructible and change all the vanilla cases where items are destroyed or used in crafting. But this would likely be to out of scope for fabric-api.


TelepathicGrunt:
Don't forget to add tag translation to EnglishTagLangGenerator as well as making this tag in the normal datagen file. That way the tag has a translation and has a blank json file made for people that look at the json rather than the tag classes. Better visibility that way

Dev0Louis:
The tag has been discussed on Neo Forge and it seems like that it is not wanted from their side. 
I also got new naming suggestions such as `fabric:cannot_be_destroyed`. The way forward with such a tag would be to actually patch game behavior based on it. 

Question that need to be answered first is if that is in scope of fabric api.

Issue #3701, opened 2024-04-10
Wolf Variant Biome Tags


catter1:
## Summary
I would like to see the creation of a set of biome tags for spawning the wolf variants. Something along the lines of `#c:has_wolf_variant/<variant>` for each wolf variant, and any future variants added by Vanilla or other mods.

## Reasoning
With the current implementation, the biome(s) wolves can generate are currently hardcoded in the `wolf_variant` files. Three of the variants use biome tags (`#minecraft:is_jungle`, `#minecraft:is_savanna`, and `#minecraft:is_badlands` for the `rusty`, `spotted`, and `striped` variants, respectively), but the rest of the varaints only reference one biome. If another biome-adding mod/datapack wants to make the variants generate in their own biomes, they will have to override that file. This is an issue for compatibility reasons.

If they are implemented, a biome-adding mod/datapack would be able to add their biomes to the tags like normal. Then they would have the additional responsibility of overriding those `wolf_variant` files to reference the `fabric` tags instead of the hardcoded vanilla biome(s). Do keep in mind that I am speaking from the perspective of a datapack developer, who heavily relies on the tags.

## Clarification
**Important:** This convention would __not__ edit the actual `wolf_variant` files. That is what makes it a convention and not a behavior modification. It is up to each individual biome mod and datapack to implement the tag into the files. Obviously this is not ideal, but it is the best option we have for compatibility. There are downsides associated (i.e. what if a mod wants to change a Vanilla wolf texture), but this would not be at fault of Fabric/Neoforge/the convention, but rather an understanding. 

Of course, if Mojang suddenly adds their own tags, that would solve this entire issue.

## Application
Terralith and Regions Unexplored are popular mods that have agreed on using this tag convention. Both Terralith and Regions Unexplored will modify the `wolf_variant` files to reference the same `#c:has_wolf_variant/<variant>` tag, as well as add custom biomes to the tag. This allows both mods to add new biomes which can spawn wolf variants without completely overriding each other.

## Semantics
The easy option is to say "just have separate `fabric` and `neoforge` tags for this". However, it has its own issues. For datapack developers that add their own biomes, what tag should they reference? The Fabric one, or the Neoforge one? One would have to choose which modloader to be compatible with.

## Final Notes
[...]

PR #3807, opened 2024-05-27
Tag cleanup and additions for 1.21

### LINKED ISSUE #3731: Leather item tag for convention-tags-v2
Would be nice to have for mods that add "leathers" comparable to regular leather.

### LINKED ISSUE #3742: Should `c:coal` item tag be removed in favor of minecraft's coal tag?
It is a bit late now but wanted to document it here so it is not forgotten about. Nouns should be plural so this tag not being called coals was a typo. It isn't in neo side so no mismatch there. But it can't be fixed on Fabric's side without it being a breaking change in code unless a second coals tag is made, deprecate the singular one, and pull in the singular one into the plural one as an optional entry.

### LINKED ISSUE #3755: Redundant tags for raw storage blocks
Copied from: https://github.com/neoforged/NeoForge/issues/919

Please reply on the neo issue on what Fabric community thinks about these tags and if fabric will keep `raw blocks` Item Tag with better javadocs for them or remove the `raw blocks` item tags while keeping the `raw storage block` item tags.

image

image

From the neo issue report:
---

Minecraft Version: 1.20.6

NeoForge Version: 20.6.24-beta

Description of issue:

Neo currently has two sets of tags for raw ore blocks: [storage_blocks/raw_](https://github.com/neoforged/NeoForge/blob/1.20.x/src/main/java/net/neoforged/neoforge/common/Tags.java#L559-L561) ([with a block form](https://github.com/neoforged/NeoForge/blob/1.20.x/src/main/java/net/neoforged/neoforge/common/Tags.java#L213-L215)) and [raw_blocks/](https://github.com/neoforged/NeoForge/blob/1.20.x/src/main/java/net/neoforged/neoforge/common/Tags.java#L496-L499) (only item form, no block). Both just contain the vanilla raw blocks, and neither has any javadocs explaining the distinction, leading to modders being unsure of which one to use.

I think the storage block tag pretty clearly means it can be converted to and from the raw ore item, just like other storage block tags. However, raw blocks is unclear as we lack examples that are not vanilla.

The best answer I could get was that raw blocks are blocks that can drop 1 or more raw ore when mined, however that definition was unsatisfying to me as raw ore blocks do not drop raw ore (they drop themselves) and regular ores drop 1 or more raw ore items (but are not tagged as raw blocks). Plus, raw_blocks/ is exclusively an item tag, talking about block drops with an item tag does not make sense.
I was also told the tags exist because they came from Fabric, however Fabric lacked a mirror to the raw storage block tags suggesting the tags potentially should have just been merged with the forge tag for these vanilla block.
Thus, I propose we take one of the following actions:

Remove the raw_blocks/ tags as they lack any conventions tied to their usage. Since are not in a breaking change window, that would consist of deprecating for removal until the next breaking changes window.
Document exactly what it means to be a raw block, and how that is different from raw storage blocks and ore blocks and potentially make it a block tag as well. Make it clear what guarantees the tag provides and what other tags are required to make the tag meaningful (e.g. do we need to do tag intersection like with ores for sizes?)

### LINKED ISSUE #3790: Add `c:music_discs` in 1.21
Mojang removed the tag. I think it was worth having for things like dropping a random music disc from a musical mob or spawning a random music disc in special chests or music disc input recipes
image

https://github.com/neoforged/NeoForge/issues/993

TelepathicGrunt:
With 1.21 shaping up to be another large breaking change for many modders, we can use this for our advantage to do some tag cleanup work. This is paired up with this PR: https://github.com/neoforged/NeoForge/pull/1030

The changes includes:

- Adding a `c:music_discs` tag since vanilla is removing `minecraft:music_discs` in 1.21. A log warning was added for this so people know they can migrate over to the c tag for music disc stuff that should not be reliant on the JUKEBOX_PLAYABLE item component.

- deprecated of `c:coal` tag that I missed before. I could not find anyone to say why this singular-named tag exists when vanilla has a `minecraft:coals` tag that would work for modder's use cases. So it seems better to deprecated this tag and add a log warning if found to tell people to migrate to the vanilla tag instead.

- `raw_blocks` item tags are deprecated because the `storage_blocks` tag for the raw blocks already exist and is the same. Log warning added for this migration cleanup. The deprecated of these tags should reduce confusion in this area for modders.

- Added `c:leathers` tag on request to match NeoForge side.

- Removed Eroded Badlands from Sparse Vegetation Overworld since the biome has no trees at all and that normal badllands and desert isn't in the tag in first place.

- Tool tags were misspelled. So instead of `c:tools/bows`, it should had been `c:tools/bow` because tools is the noun to be plural and bow is the adjective. So it would read as Bow Tools instead of the awkward Bows Tools. Attempted to be added in non-breaking way as possible.

- Deprecated Shulker Boxes block tag because Minecraft has a Shulker Box block tag 

- Glazed Terracotta block/item tag, Concrete block/item tag, Concrete Powder item tag.

Please comment here if there's any change here that you dislike or love! Or if you have suggestions of new tags to add, to sync from other loader, or should also be removed/adjusted.

See the below issue reports for more details and reasoning for each of this PR's changes: 

Closes https://github.com/FabricMC/fabric/issues/3742
Closes https://github.com/FabricMC/fabric/issues/3755
Closes https://github.com/FabricMC/fabric/issues/3790
Closes https://github.com/FabricMC/fabric/issues/3731


TelepathicGrunt:
Note there is talk about the wolf variant biome tags on the other PR here:
https://github.com/neoforged/NeoForge/pull/1030#issuecomment-2133915916

Is wolf variant biome tags something Fabric is willing to allow? It's an odd case that requires overriding some vanilla files to work properly in a sensible way. (Original request is https://github.com/FabricMC/fabric/issues/3701)

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -74,15 +74,11 @@ private ConventionalItemTags() {
 	public static final TagKey<Item> NUGGETS = register("nuggets");
 	public static final TagKey<Item> ORES = register("ores");
 	public static final TagKey<Item> RAW_MATERIALS = register("raw_materials");
-	public static final TagKey<Item> RAW_BLOCKS = register("raw_blocks");
```
modmuss50, line 77: Breaking change, im not super keen on making more breakig changes like this, it should be deprecated.

TelepathicGrunt: just adding the deprecated tag and not datagen the files?

modmuss50: No, that still sounds breaking to me. Ill have to look at this properly, I would expect to levae it as it, but just deprecate it.

TelepathicGrunt: Sounds good, I added the tags back with their datagen. But I moved them to the bottom of ConventialItemTags, put a deprecated tag on them, and javadoc ont hem saying they are redundant with what replacement to use. Plus the console will also warn about using them if detected as well. That might be good enough to get people off the tags to replacements without breaking

TelepathicGrunt:
Tool tags were misspelled. 

so instead of `c:tools/bows`, it should had been `c:tools/bow` because tools is the noun to be plural and bow is the adjective. So it would read as Bow Tools instead of the awkward Bows Tools.

Tried to implement the change as non-breaking as I can. Caused an awkward naming of one field `MINING_TOOL_TOOLS` because the original field was `MINING_TOOLS` which is now deprecated

modmuss50 (approved):
Looks great, many thanks. I will put this into last call, I will aim for it to be slightly shorter than normal (maybe 3 days?) so we can get it in well before the stable release.

deprecated/fabric-convention-tags-v1/src/main/java/net/fabricmc/fabric/impl/tag/convention/ConventionLogWarnings.java
```diff
@@ -209,8 +210,26 @@ private enum LogWarningMode {
...
+			createMapEntry(RegistryKeys.ITEM, "tools/brushes", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.BRUSH_TOOLS),
+			createMapEntry(RegistryKeys.ITEM, "tools/melee_weapons", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.MELEE_WEAPON_TOOLS),
+			createMapEntry(RegistryKeys.ITEM, "tools/ranged_weapons", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.RANGED_WEAPON_TOOLS),
+			createMapEntry(RegistryKeys.ITEM, "tools/mining_tools", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.MINING_TOOL_TOOLS),
+			createMapEntry(RegistryKeys.ITEM, "raw_blocks", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.STORAGE_BLOCKS),
+			createMapEntry(RegistryKeys.ITEM, "raw_blocks/copper", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.STORAGE_BLOCKS_RAW_COPPER),
+			createMapEntry(RegistryKeys.ITEM, "raw_blocks/gold", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.STORAGE_BLOCKS_RAW_GOLD),
+			createMapEntry(RegistryKeys.ITEM, "raw_blocks/iron", net.fabricmc.fabric.api.tag.convention.v2.ConventionalItemTags.STORAGE_BLOCKS_RAW_IRON),
```
modmuss50, line 231: These are still provided by FAPI so always trigger the warning. Im not too sure what the solution is.

TelepathicGrunt: Ah that’s true. I missed that. Not sure either. It stems from the fact that fabric doesn’t want to remove the actual tag file and field. I guess for fabric, you can remove it from the warnings and then just keep the deprecated annotation. Then in some far future when the deprecated tags are finally removed, you can add the entries to the warning

purejosh:
~~Which tag form is favored, the plural form or the singular form?~~

After checking the `ConventionalItemTags`, it appears that the singular form is now preferred.

TelepathicGrunt:
@purejosh which tag are you talking about.nouns are plural. Adjectives are not. So c:tools is plural. c:tools/spear is technically plural as you read it as spear tools. Spear is the adjective for the tool

purejosh:
> @purejosh which tag are you talking about.nouns are plural. Adjectives are not. So c:tools is plural. c:tools/spear is technically plural as you read it as spear tools. Spear is the adjective for the tool

Apologies, I posted hastily. I checked the `ConventionalItemTags` and found my answer.

PR #3797, opened 2024-05-24
Add mixins for FabricItem.getAttributeModifiers

opekope2:
Added an ItemStack mixin to call `FabricItem.getAttributeModifiers(ItemStack)`.
Also added a mixin when Minecraft creates the item tooltip, because unlike 1.20.4, which get attribute modifiers using `getAttributeModifiers`, 1.20.6 loads them from data components.

I tested it on a production client with a mod of mine (source not yet upstream), and it worked. I'll test it on a server later.

Resolves #3787

PR #3809, opened 2024-05-28
Add mixins for FabricItem.getAttributeModifiers (1.21)

### LINKED ISSUE #3787: FabricItem.getAttributeModifiers(ItemStack) is defunct
`Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> getAttributeModifiers(ItemStack, EquipmentSlot)` was refactored to `AttributeModifiersComponent getAttributeModifiers(ItemStack)` in f1e646e to reflect Minecraft changes, however, `ItemStackMixin.hookGetAttributeModifiers` was removed earlier in 1172e89, which called this API.
GitHub search
IDEA search

opekope2:
Added an ItemStack mixin to call `FabricItem.getAttributeModifiers(ItemStack)`.
Similar to #3797, but for 1.21.
Resolves #3787

I tested it on a dev client&server and a prod client&server, and I didn't encounter any issue.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
@@ -25,7 +25,10 @@ import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
 
+import net.minecraft.component.ComponentType;
+import net.minecraft.component.type.AttributeModifiersComponent;
 import net.minecraft.entity.EquipmentSlot;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.item.Item;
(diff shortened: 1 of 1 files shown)


apple502j (approved):

modmuss50:
I will merge this in 1.21 now, and get the 1.20.6 PR out another time 👍

modmuss50:
This fix is broken (https://github.com/FabricMC/fabric/issues/3812) FabricItem.getAttributeModifiers should be tottaly removed as you can now use the component. I will revert and deprecate for 1.20.6, and remove fully in 1.21

opekope2:
This API allowed `Item`s to override attribute modifiers based on certain conditions of the `ItemStack`. [Here](https://github.com/opekope2/StaffMod/blob/e06e25f55067ca3909f429dfc4bf2a37429e7ccc/FabricMod/src/main/kotlin/opekope2/avm_staff/internal/fabric/item/FabricStaffItem.kt#L55-L57)'s an example where I override it, which delegates to [AnvilHandler](https://github.com/opekope2/StaffMod/blob/e06e25f55067ca3909f429dfc4bf2a37429e7ccc/StaffMod/src/main/kotlin/opekope2/avm_staff/internal/staff_handler/AnvilHandler.kt#L120) and other classes. The alternative I see is changing the attribute modifier [where I modify the `ItemStack`](https://github.com/opekope2/StaffMod/blob/e06e25f55067ca3909f429dfc4bf2a37429e7ccc/StaffMod/src/main/kotlin/opekope2/avm_staff/util/StaffUtil.kt#L56-L63), but that only gets called when a player presses a key, and not when `/give` is used, so initially it would not be functional as intended, only after the player presses the key. NeoForge supports this using `IItemExtension.getAttributeModifiers(ItemStack)` as of 1.20.6, but I don't know about their future.

modmuss50:
Unfortunately I'm struggling to follow the Kotlin code sorry, you can set the default item componet in the Item settings, this would then apply to /give.

The vanilla none stack aware method is deprecated, and ingored if the stack has a attribute modifier component.

opekope2:
My use case is to change the attribute modifiers based on my custom attribute modifier. For `/give`, I'll try `Item.postProcessComponents(ItemStack)`

opekope2:
The removal makes sense if I look at mojmap

PR #3814, opened 2024-05-30
Add API to control creative inventory screen

### LINKED ISSUE #3813: [Suggestion] Public API for dealing with creative mode pages/tabs
Currently it's impossible for mods to control FAPI's page system in creative inventory without using impl.

Here are some features this API should include:
- Allow getting and setting current page index (CreativeGuiExtensions impl)
- Getting the page count
- Getting the page associated with a specific tab (CreativeInventoryScreenMixin#fabric_getPage)
- Get a sorted list of tabs on a specific page
- Get/set the currently selected tab (vanilla accessor)

modmuss50:
Closes #3813


Syst3ms (approved):

BoogieMonster1O1 (approved):

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	/**
+	 * Returns whether there are additional pages to show on top of the default vanilla pages.
+	 *
+	 * @return true if there are additional pages
+	 */
+	default boolean hasAdditionalPages() {
+		throw new AssertionError("Implemented by mixin");
+	}
```
NebelNidas, line 97: > whether there are additional pages to show on top of the default vanilla pages.

If "default vanilla pages" implies "pages present in the base game", how would cases like `24w14potato` be handled?

Vanilla on the right, Fabric API on the left:
image

I'd split this into two methods, `hasMultiplePages` and perhaps `hasNonVanillaItemGroups`.

NebelNidas (changes requested):

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
modmuss50: We didnt handle that, its an edge case in one april fools version that we didnt care about. Thus I dont think we should design the API around this. If an official stream of the game was to add another tab we would need to make changes to the API. I dont expect this to happen though.

haykam821: I don't see an issue with that scenario, since this method mostly seems useful for determining whether the page selector will be rendered.

NebelNidas: Fair. I would mark this comment as resolved, though I don't seem to have a button for that 🤔

NebelNidas (approved):

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
```diff
@@ -0,0 +1,117 @@
...
+	 * Returns an ordered list containing the item groups on the requested page.
+	 */
+	default List<ItemGroup> getItemGroupsOnPage(int page) {
+		throw new AssertionError("Implemented by mixin");
+	}
+
+	/**
+	 * Returns the page index of the given item group.
+	 *
+	 * <p>The common item groups always return the current page index.
```
haykam821: 'Common' might be vague in this case:

```suggestion
	 * <p>Item groups appearing on every page always return the current page index.
```

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
```diff
@@ -0,0 +1,117 @@
...
+import net.minecraft.client.gui.screen.ingame.CreativeInventoryScreen;
+import net.minecraft.item.ItemGroup;
+
+/**
+ * Fabric provided extensions to {@link CreativeInventoryScreen}.
+ * This interface is automatically implemented on all creative inventory screens via Mixin and interface injection.
+ */
+public interface FabricCreativeInventoryScreen {
+	/**
+	 * Switches to the page with the given index.
```
haykam821:
```suggestion
	 * Switches to the page with the given index if it exists.
```

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
```diff
@@ -0,0 +1,117 @@
...
+	 *
+	 * @param page the index of the page to switch to
+	 * @return Returns true when the page was changed
+	 */
+	default boolean switchToPage(int page) {
+		throw new AssertionError("Implemented by mixin");
+	}
+
+	/**
+	 * Switches to the next page.
```
haykam821:
```suggestion
	 * Switches to the next page if it exists.
```

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/api/client/itemgroup/v1/FabricCreativeInventoryScreen.java
```diff
@@ -0,0 +1,117 @@
...
+	 *
+	 * @return Returns true when the page was changed
+	 */
+	default boolean switchToNextPage() {
+		return switchToPage(getCurrentPage() + 1);
+	}
+
+	/**
+	 * Switches to the previous page.
```
haykam821:
```suggestion
	 * Switches to the previous page if it exists.
```

haykam821 (commented):
Parts of the Javadoc could be clarified.

haykam821 (approved):

PR #3824, opened 2024-06-02
Port Model Loading API v1 to 1.21

PepperCode1:


modmuss50 (approved):
Echoing what I said on discord:

Using Either does feel little bit convoluted, especially if you don't care about the variant and just want the ID. The Context interfaces could maybe have a method to get the Identifier from the either? I did think about having a sealed interface that wraps both a ModelIdentifier and Identifier but I think that would be way too over engineered and I couldn't come up with a good name for it 😄  This might be a none issue though, I'm not too sure how people use this?

I think its great that you managed to keep the API basically the same, I'm not really able to give an in-depth review on the impl. Other than the one minor checkstyle issue id be happy to merge it, we can always make changes later once people start using it.

Thanks again for looking at this.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BlockStatesLoaderMixin.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Mixin(BlockStatesLoader.class)
+abstract class BlockStatesLoaderMixin implements BlockStatesLoaderHooks {
+	@Unique
+	@Nullable
+	private LoadingOverride loadingOverride;
+
+	@WrapWithCondition(method = "load()V", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/model/BlockStatesLoader;loadBlockStates(Lnet/minecraft/util/Identifier;Lnet/minecraft/state/StateManager;)V"))
+	private boolean shouldDoVanillaLoading(BlockStatesLoader self, Identifier id, StateManager<Block, BlockState> stateManager, @Local Block block) {
+		return loadingOverride == null || !loadingOverride.loadBlockStates(block, id);
```
embeddedt: To be able to load one blockstate file at a time, I currently call the inner `loadBlockStates` method directly. Simply replacing the iterator in `load` with a singleton isn't ideal for me because it loads all the static definitions as well. For my approach to work with this loading override system, the Fabric API mixin here would need to be changed to an injector in `loadBlockStates` itself. Injecting at HEAD and conditionally canceling would probably work?

PR #3861, opened 2024-06-16
Add eggs common tag

ThatGravyBoat:
This adds the eggs tag which is present on neoforge but is not on fabric, this is a very useful tags as it allows use of common eggs between platforms especially for recipes, several mods add their own egg to this tag but vanillas chicken egg is not in there by default unless a mod adds it which causes issues or is just slightly annoying that if you want to use it you either have to have an or for the tag and the vanilla item or you have to add the vanilla item yourself to the tag.

### DIFF
--- a/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/EnglishTagLangGenerator.java
+++ b/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/EnglishTagLangGenerator.java
@@ -274,6 +274,7 @@ public class EnglishTagLangGenerator extends FabricLanguageProvider {
 		translationBuilder.add(ConventionalItemTags.ROPES, "Ropes");
 		translationBuilder.add(ConventionalItemTags.CHAINS, "Chains");
 		translationBuilder.add(ConventionalItemTags.HIDDEN_FROM_RECIPE_VIEWERS, "Hidden From Recipe Viewers");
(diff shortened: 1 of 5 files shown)


haykam821:
Should a tag of eggs that doesn't contain sniffer, turtle, or dragon eggs (and arguably frogspawn, but maybe not spawn eggs) be named `c:eggs`?

ThatGravyBoat:
> Should a tag of eggs that doesn't contain sniffer, turtle, or dragon eggs (and arguably frogspawn, but maybe not spawn eggs) be named `c:eggs`?

I would say yes, as I said this is already a tag used by the community in several mods, making the tag either include the other eggs would cause a discrepancy between existing mods that have already standardized using only the chicken egg in this tag, from all public mods available on GitHub not a single one adds the dragon, sniffer, or turtle egg, and if it were to be changed to another name it would not put it inline with the existing tag that didn't seem to be added to fabric during the tag pr.

haykam821:
I think this tag should be named something along the lines of `c:baking_eggs` instead of adding a new tag that is wrong.

modmuss50:
Yes, I agree with @haykam821 here, if its known to be a bad name then we shouldn't add it.

ThatGravyBoat:
> Yes, I agree with @haykam821 here, if its known to be a bad name then we shouldn't add it.

I understand, I do think the `c:baked_eggs` does not convey effectively what the tag is commonly and already used by many for, I think a better name then could be `c:food_ingredients/eggs`

ThatGravyBoat:
Was added in https://github.com/FabricMC/fabric/pull/4186

PR #3881, opened 2024-06-24
Fix legacy tag warning text to be accurate

TelepathicGrunt:
DEV_SHORT and DEV_VERBOSE are not valid entries. Only these are:
image


modmuss50 (approved):
Thanks, I forgot to change these 👍

PR #3886, opened 2024-06-25
Fix CoreShaderRegistrationCallback mixin

### LINKED ISSUE #3230: CoreShaderRegistrationCallback is called multiple times on shader reload in 1.20.x

modmuss50:
Superseeds #3521
Fixes #3230

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/GameRendererMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/shader/GameRendererMixin.java
@@ -20,13 +20,12 @@ import java.io.IOException;
 import java.util.List;
 import java.util.function.Consumer;
 
+import com.llamalad7.mixinextras.sugar.Local;
 import com.mojang.datafixers.util.Pair;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.Slice;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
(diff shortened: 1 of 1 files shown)


mysticdrew (approved):
👍 

Looks good, thanks for the quick turnaround!

PR #3887, opened 2024-06-26
Replace redirect in lifecycle events with wrap operation

AlphaMode:
Noticed a old redirect while updating porting lib to 1.21. So I thought I might create a quick PR to fix this.


Dragon-Seeker:
Seems fine

AlphaMode:
Sugar could also be used here but I figured that would be just as hacky as using locals

modmuss50:
I assume there is a reason for making this change? Generally my rule in moving stuff to mixin extras is to only do it when there is actually a need/benefit or when the mixin need changing anyway (Usually during a port).

AlphaMode:
> I assume there is a reason for making this change? Generally my rule in moving stuff to mixin extras is to only do it when there is actually a need/benefit or when the mixin need changing anyway (Usually during a port).

I found it because it was causing a conflict in porting lib. I fixed this by just changing the injection point.
Anyway there isn't really not a reason to change this as it would prevent future possible conflicts

modmuss50 (approved):

PR #3889, opened 2024-06-26
Add `Event#hasListener`, remove unused junk

apple502j:
Should be self-descriptive.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
@@ -88,4 +88,15 @@ public abstract class Event<T> {
 	public void addPhaseOrdering(Identifier firstPhase, Identifier secondPhase) {
 		// This is not abstract to avoid breaking existing Event subclasses, but they should really not be subclassing Event.
 	}
+
+	/**
+	 * Returns whether the event has any listeners. Useful for skipping steps related to
+	 * event invocation, such as making context objects, when the event has no listeners.
+	 * This method is unnecessary when the only code guarded by this check is the event invocation.
+	 *
+	 * @return whether the event has any listeners
+	 */
+	public boolean hasListener() {
+		return true;
+	}
 }
--- a/fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
+++ b/fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
@@ -25,11 +25,8 @@ import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-import net.minecraft.Bootstrap;
-import net.minecraft.SharedConstants;
 import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.api.event.Event;
@@ -37,12 +34,6 @@ import net.fabricmc.fabric.api.event.EventFactory;
 import net.fabricmc.fabric.impl.base.toposort.NodeSorting;
 
 public class EventTests {
-	@BeforeAll
-	static void beforeAll() {
-		SharedConstants.createGameVersion();
-		Bootstrap.initialize();
-	}
-
 	private static final Function<TestCallback[], TestCallback> INVOKER_FACTORY = listeners -> () -> {
 		for (TestCallback test : listeners) {
 			test.onTest();
--- a/fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
+++ b/fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
@@ -16,15 +16,20 @@
 
 package net.fabricmc.fabric.test.base;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
 
+import net.minecraft.Bootstrap;
+import net.minecraft.SharedConstants;
 import net.minecraft.util.Identifier;
 
 import net.fabricmc.fabric.api.event.Event;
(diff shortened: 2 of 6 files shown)


modmuss50 (commented):
Please add some tests.

modmuss50 (approved):

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
```diff
@@ -16,63 +16,59 @@
...
-		NodeSorting.ENABLE_CYCLE_WARNING = false;
-		testDeterministicOrdering();
-		testTwoCycles();
-		NodeSorting.ENABLE_CYCLE_WARNING = true;
-
-		long time2 = System.currentTimeMillis();
-		LOGGER.info("Event unit tests succeeded in {} milliseconds.", time2 - time1);
+	@BeforeAll
+	static void beforeAll() {
+		SharedConstants.createGameVersion();
```
Technici4n: Are these needed for the tests?

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
```diff
@@ -88,4 +88,15 @@ public void register(Identifier phase, T listener) {
...
 	}
+
+	/**
+	 * Returns whether the event has any listeners. Useful for skipping steps related to
+	 * event invocation, such as making context objects, when the event has no listeners.
+	 * This method is unnecessary when the only code guarded by this check is the event invocation.
+	 *
+	 * @return whether the event has any listeners
+	 */
+	public boolean hasListener() {
```
Technici4n, line 99: IMO this is a bad pattern. If you are concerned about allocation, reuse the context object. In typical modpacks, most events will be used and these checks will be pointless anyway.

modmuss50: Thats a fair point, @apple502j did you have a specific use case in mind?

BasiqueEvangelist: A usecase I had for this was for a deprecated event, where actually properly servicing it required setting up some expensive stuff

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
apple502j: Probably not, but doesn't really matter. Also note that our testmods are used by devs as templates for other testmods.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
apple502j: Also, not all context objects can be reused.

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
Technici4n: Well, I don't think that devs should initialize MC if they don't need it.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
Technici4n: Can you link some specific examples?

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/EventTests.java
modmuss50: If its not needed, please remove. At the very least it will make the tests a second or two faster when running on their own.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
Technici4n, line 99: This is not resolved, please don't mark it as resolved.

modmuss50:
Still wating on a good usecase for this, in places where it matter we already re-use a context object, and firing an event with no listeners should have little overhead.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/event/Event.java
BasiqueEvangelist: https://github.com/BasiqueEvangelist/ne-v-seti/blob/sanity-1.17/src/main/java/me/basiqueevangelist/nevseti/OfflineDataCache.java#L30
I guess I misremembered specifically how much it would take to service this event, but I think it will still be useful for events that are deprecated and require running extra logic to run their invokers

apple502j:
I think I had some events in my mind when I filed this PR, I need to figure that out again.

apple502j:
**Events with non-zero invocation cost**

- RegSync: Remap event constructs `RemapStateImpl`. Very negligible as this is called only once per connection.
- `ItemGroupEvents`: Involves creation of several objects. Once per startup, so doesn't really matter, unless Mojang decuples the number of items or something.
- `LootTableEvents`: This **might** benefit from a check. An event (invoked per loot table) requires `LootTable` to be converted back to builder.

Will do a more thorough research later.

zenfyrdev:
> `LootTableEvents`: This might benefit from a check. An event (invoked per loot table) requires `LootTable` to be converted back to builder.

IMO this would only benefit if there was a separation between singular and wildcard events (like `MODIFY_ALL`, `modify(RegistryKey)` or `modify(Predicate<LootTable>)`).

With the current implementation, if a single mod wants to modify a single loot table, all the benefits of such a check disappear.

Juuxel:
`modify(Identifier)` or `modify(RegistryKey<LootTable>)` would actually make sense for the API since most modifications are probably targeted to just a single loot table.

ishland:
When a synchronous event that's only meant to be invoked from the server thread needs to be invoked from a worker thread, it needs to make a round-trip to the server thread. Such round-trip can take 50ms or more depending on the MSPT of the server. 
Wasting that amount of time invoking an event that nobody listens to is just a waste of time. Having such API can avoid doing such costly round-trips. 

I encountered this when implementing #4541 where I have to make round-trips to the server thread to invoke the event: 
https://github.com/RelativityMC/C2ME-fabric/commit/0a04fdc6b1c9c96bd985adb2fc2fc279420656e2
There's one such case in the `ServerBlockTicking` class and another two in `ServerEntityTicking`.

Technici4n:
Wouldn't this indicate a flaw with the event design? The moment any mod registers a listener, all the benefits of C2ME will be gone, no? @ishland

ishland:
> Wouldn't this indicate a flaw with the event design? 

I can't think of any better designs other than splitting that event into 6 events.  Relaxing where the event can be invoked just adds more trouble for modders that have no idea about threading. 

> The moment any mod registers a listener, all the benefits of C2ME will be gone, no? 

No. Multiple pending synchronizations with the server thread can be in flight at the same time. But adding extra server thread synchronization points (for just invoking the event) adds more server thread overhead and increases latency.

modmuss50:
I dont have any major concers with this, the use case seems good enough to me. The PR will need updating before moving forward.

PR #3900, opened 2024-06-30
Updating food tags to have separation of golden & non-golden foods

### LINKED ISSUE #3899: Fabric Standard Tags for Food are limited.
As i develop both datapacks & mods more, i am realizing it is important for modded Compatibility to have different tags for golden foods & normal foods. However, the current tags are used by many mods, and removing the normal tags would cause incompatibility when updating mods.

And for people who don't need this separation, the normal tags should still exist. 

Golden foods such as golden apple, enchanted golden apple, and golden carrot need too have a tag separate from the rest. and they also need to be included in the normal tag, used for targeting all fruits & vegies including golden.

I am opening a pull request myself, however as i am still new to the modding scene i'm not sure i did everything right. If this is the case, i would love to try to fix it myself, unless fabric peeps just wanna make the change thsemselves, but hopefully if i explain bad the PR should at least demonstrate what i want to happen.

Bowswa:
Modify Standard Tags Data Generator
-Structured to not break mods
-Added tags for golden VS nongolden fruits & veggies
-Those tags are included in the original tag as "optional".

Meant to fix #3899

### DIFF
--- a/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/ItemTagGenerator.java
+++ b/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/ItemTagGenerator.java
@@ -210,17 +210,29 @@ public final class ItemTagGenerator extends FabricTagProvider.ItemTagProvider {
 	}
 
 	private void generateFoodTags() {
-		getOrCreateTagBuilder(ConventionalItemTags.FRUITS_FOODS)
-				.add(Items.APPLE)
+    getOrCreateTagBuilder(ConventionalItemTags.GOLDEN_FRUITS_FOODS)
 				.add(Items.GOLDEN_APPLE)
 				.add(Items.ENCHANTED_GOLDEN_APPLE);
+    
+    getOrCreateTagBuilder(ConventionalItemTags.NORMAL_FRUITS_FOODS)
+				.add(Items.APPLE);
 
-		getOrCreateTagBuilder(ConventionalItemTags.VEGETABLES_FOODS)
+		getOrCreateTagBuilder(ConventionalItemTags.FRUITS_FOODS)
+				.addOptionalTag(ConventionalItemTags.GOLDEN_FRUITS_FOODS)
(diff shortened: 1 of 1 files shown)


TelepathicGrunt:
You are missing the lang file datagen as well. And you need to run the datagenerators. Also, these new fields are not even added to ConventionalItemTags so this would not even compile. I would encourage only making edits while having the code opened in an IDE so you can run all the needed tasks, spotless, and datagenerators

TelepathicGrunt:
Also, would the actual tag be 

`c:foods/golden_fruit` 
`c:foods/normal_fruit` 
and then collect those two tags into `c:foods/fruit`

Or do 
`c:foods/fruit/golden` 
`c:foods/fruit/golden` 
and then collect those two tags into `c:foods/fruit`

Leaning towards
`c:foods/fruit/golden`
`c:foods/fruit/regular`
and then collect those two tags into `c:foods/fruit`

TelepathicGrunt:
@Bowswa I added `c:foods/golden` in this PR instead.
https://github.com/FabricMC/fabric/pull/3912

The reason i did not split the Fruits and Vegetable tags is because people can add other golden foods like meat and such. So rather than open that can of worms of golden tags everywhere, I added a single tag for golden foods.

To make best use of that new golden tag, please checkout `net.fabricmc.fabric.api.recipe.v1.ingredient.DefaultCustomIngredients` as it give you the ability to do tag intersections and differences. This allows recipes to do an intersection on Fruits and Golden to get the two golden apple foods plus other modded golden fruits. You can also do the difference of fruits and golden to get all non-golden fruits.

Now if you have any datapacks that is supposed to work in vanilla and Fabric at same time, take a look into Overlays. 
https://gist.github.com/lukebemish/91431e6bdabd495e692e7069d822f1ae
It is a vanilla feature where you can load a vanilla-based recipe if not on fabric (you own tag of fruits you want) and when on fabric, then load the special fabric ingredient based recipe. This will become viable one https://github.com/FabricMC/fabric/pull/3872 is merged

modmuss50:
Closing this in favor of @TelepathicGrunt 's. Many thanks for the help. 👍

Issue #3895, opened 2024-06-29
[Convention Tags] Entries missing from certain tags/categories


Patbox:
- [ ] Maces are missing from c:tools/melee_weapons
- [ ] Flint and steel is missing from c:tools

TelepathicGrunt:
Flint and steel is a weird one. It's like on border of tool and not. Where would the line be drawn of what is a tool? What about Name Tags or Lead? Would people have any negative experience if flint and steel was in Tools tag when compared to the other items in Tools tag?

apple502j:
I believe a tool is something with a durability (excluding armor and perhaps weapons).

Patbox:
Weapons are included in tools tag via inclusion of their tags. But yes, flint and steel is a tool in both looks and usecase, it's just not mining tool. But yeah, you can go with mc wiki definition for that https://minecraft.wiki/w/Tool

TelepathicGrunt:
Should flint and steel only be in tools or should be there be a subcategory for it?
`c:tools/firestarter`?

PR #3912, opened 2024-07-05
Several tag fixes and additions

### LINKED ISSUE #3869: Missing a few c tags,
These tags are missing: c:glass_blocks/colorless, c:ender_pearls

### LINKED ISSUE #3899: Fabric Standard Tags for Food are limited.

TelepathicGrunt:
Changes includes:

- Renaming all food tags under `c:foods/` to be singular. Example is `c:foods/fruits` to be `c:foods/fruit`. This is because you would say Fruit Foods and not Fruits Foods. Basically, foods is the noun. Fruit is the descriptor of the food noun. Backwards compat added here to reduce breakage for people.

- Added `c:foods/golden` to collect all gold based foods. Which then can be paired ingredient intersections and differences to target or exclude from the other food tags.

- Added `c:tools/igniter` to now have Flint and Steel be considered a tool properly.

- Adding `c:ender_pearls`, `c:glass_blocks/colorless`, `c:glass_blocks/cheap`, `c:glass_blocks/tinted`, `c:glass_panes/colorless` tags to match what Neoforge has closer.

Neoforge PR: https://github.com/neoforged/NeoForge/pull/1238

Closes: https://github.com/FabricMC/fabric/issues/3899

Closes: https://github.com/FabricMC/fabric/issues/3869

Closes half of : https://github.com/FabricMC/fabric/issues/3895


modmuss50 (approved):
Looks good, thanks for the contiuned support with this.

TelepathicGrunt:
@modmuss50 One second, I'll deal with the merge conflicts

TelepathicGrunt:
@modmuss50 Should be good now

modmuss50:
Lol, I was doing the same :D Wondered why I couldnt push it lol. Thanks 👍

ffuentesm:
does this mean the duplicated plural tags will be removed? 
Like how we still have c:foods/vegetables & c:foods/vegetable in the API.

TelepathicGrunt:
@ffuentesm The deprecated (typoed) tags are to be removed in 1.22 iirc unless the fabric team prefers to remove it earlier. They were a mistake to begin with but removing them is a breaking change that fabric api is trying to avoid

PR #3913, opened 2024-07-06
Implement built-in mod resource pack sorting

Apollounknowndev:
Overriding data/resource pack files from other mods is currently quite difficult as mod loading order isn't consistent. This can be especially annoying for compatibility mods that exist to override data/resource files from other mods.

This PR addresses the issue by implementing a system for mods to define in their `fabric.mod.json` what their priority is in the internal Fabric Mods data/resource pack. There are five phases, with each occuring in a deterministic order (intra-phase sorting isn't guaranteed). These phases are currently `bottom`, `lower`, `default`, `upper`, and `top` but can be changed if there is a better naming convention.

Mods that wish to change their priority in the load order can add this to their `fabric.mod.json` file:
```json
"custom": {
    "fabric:resource_load_order": {
        "data": "fabric:lower",
        "assets": "fabric:default"
    }
}
```


OroArmor:
This should be on the `pack.mcmeta` file. There is no reason to restrict this to the whole mod container.

Apollounknowndev:
> This should be on the `pack.mcmeta` file. There is no reason to restrict this to the whole mod container.

Why, exactly? This is just for ordering inside of the `Fabric Mods` data/resource pack. Other packs cannot be put somewhere inside the ordering of the `Fabric Mods` data/resource pack.

modmuss50:
This is a great start, but I think we need to come up with a concreate plan for what we want. I see a few problems with the current approach of only allowing entire mods to be lower or higher. I think this is going to cause issues where mods just put them selves higher for the sake of it, just moving the problem elsewhere. Being able to specify their relation with specific mod ids might be a better solution? e.g higher than `modx`.

I agree it does seem strange to have this in the fmj, although im not going to outright dismiss it. If it was in the packmc.meta could it allow sorting builtin packs diffrently? The meta file already contains the conditions, it seems like a logicial place to put this. The split between the assets and data feels a little strange to me as well.

Apollounknowndev:
Screwed up some Github thing, creating fresh PR to fix it.

PR #3925, opened 2024-07-07
Add tests for convention tag keys matching field names

haykam821:
Issues such as #3921 can be prevented through automated tests, so this pull request introduces said tests.

### DIFF
--- a/fabric-convention-tags-v2/src/testmod/java/net/fabricmc/fabric/test/tag/convention/v2/TagUtilTest.java
+++ b/fabric-convention-tags-v2/src/testmod/java/net/fabricmc/fabric/test/tag/convention/v2/TagUtilTest.java
@@ -16,12 +16,20 @@
 
 package net.fabricmc.fabric.test.tag.convention.v2;
 
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+
+import org.objectweb.asm.Opcodes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Great idea.

fabric-convention-tags-v2/src/testmod/java/net/fabricmc/fabric/test/tag/convention/v2/TagUtilTest.java
```diff
@@ -16,26 +16,145 @@
...
+			ConventionalItemTags.RED_SANDSTONE_BLOCKS,
+			ConventionalItemTags.RED_SANDSTONE_SLABS,
+			ConventionalItemTags.RED_SANDSTONE_STAIRS,
+			ConventionalItemTags.UNCOLORED_SANDSTONE_BLOCKS,
+			ConventionalItemTags.UNCOLORED_SANDSTONE_SLABS,
+			ConventionalItemTags.UNCOLORED_SANDSTONE_STAIRS
+	);
+
+	private static boolean isValidTagField(Field field) {
+		if ((field.getModifiers() & Opcodes.ACC_STATIC) == 0) {
```
Juuxel: Nit: [`Modifier.STATIC`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/reflect/Modifier.html#STATIC) is better here as it's part of the same reflection API instead of ASM

modmuss50: You can even use `Modifier.isStatic()` 👍

PR #3927, opened 2024-07-08
Add API to create reload listeners with a registry lookup.

### LINKED ISSUE #3924: [1.21] How to obtain registry access instance for data pack reload listeners registered to `ResourceManagerHelper`?
A lot of data pack reload listeners such as for recipes, tags, and loot tables in vanilla require a registry access instance for decorating `JsonOps` when deserializing `JsonElement`s.

Personally my `JsonElement`s contain item references. How can I obtain a registry access instance for deserializing those elements?

The equivalent NeoForge event provides the necessary `RegistryAccess` instance: https://github.com/neoforged/NeoForge/blob/1.21.x/src/main/java/net/neoforged/neoforge/event/AddReloadListenerEvent.java

modmuss50:
Closes #3924

Its a bit of a hack to get the registry lookup from the recipe manager, but its much easier than trying to pass it though.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceManagerHelper.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceManagerHelper.java
@@ -55,6 +55,8 @@ public interface ResourceManagerHelper {
 	/**
 	 * Register a resource reload listener for a given resource manager type.
 	 *
+	 * <p>Note: This is only supported for server data reload listeners.
+	 *
 	 * @param identifier The identifier of the listener.
 	 * @param listenerFactory   A function that creates a new instance of the listener with a given registry lookup.
 	 */
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceManagerHelper.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceManagerHelper.java
@@ -16,8 +16,11 @@
 
 package net.fabricmc.fabric.api.resource;
 
+import java.util.function.Function;
+
 import org.jetbrains.annotations.ApiStatus;
 
+import net.minecraft.registry.RegistryWrapper;
 import net.minecraft.resource.ResourceManager;
 import net.minecraft.resource.ResourceType;
 import net.minecraft.text.Text;
@@ -48,6 +51,14 @@ public interface ResourceManagerHelper {
 	 * @param listener The resource reload listener.
 	 */
 	void registerReloadListener(IdentifiableResourceReloadListener listener);
+
+	/**
+	 * Register a resource reload listener for a given resource manager type.
+	 *
+	 * @param identifier The identifier of the listener.
+	 * @param listenerFactory   A function that creates a new instance of the listener with a given registry lookup.
+	 */
+	void registerReloadListener(Identifier identifier, Function<RegistryWrapper.WrapperLookup, IdentifiableResourceReloadListener> listenerFactory);
 
 	/**
 	 * Get the ResourceManagerHelper instance for a given resource type.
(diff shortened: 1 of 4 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
```diff
@@ -160,15 +170,21 @@ public static List<ResourceReloader> sort(ResourceType type, List<ResourceReload
 	}
 
 	protected void sort(List<ResourceReloader> listeners) {
-		listeners.removeAll(addedListeners);
+		listeners.removeAll(listenerFactories);
```
Juuxel: This doesn't seem right since the types don't match (`Collection.removeAll` has an unhelpful `Collection<?>` wildcard param)

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
```diff
@@ -200,15 +216,78 @@ protected void sort(List<ResourceReloader> listeners) {
...
+		@Override
+		public IdentifiableResourceReloadListener get(RegistryWrapper.WrapperLookup registry) {
+			return listener;
+		}
+	}
+
+	private record RegistryResourceReloaderFactory(Identifier id, Function<RegistryWrapper.WrapperLookup, IdentifiableResourceReloadListener> listenerFactory) implements ListenerFactory {
+		@Override
+		public IdentifiableResourceReloadListener get(RegistryWrapper.WrapperLookup registry) {
+			Objects.requireNonNull(listenerFactory);
```
Juuxel, line 285: It'd be better to check this at registration/`RegistryResourceReloaderFactory` construction imo.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
modmuss50: Oh yeah, great spot! thanks. I will fix tomorrow.

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourceManagerHelper.java
```diff
@@ -49,6 +52,16 @@ default void addReloadListener(IdentifiableResourceReloadListener listener) {
...
 
+	/**
+	 * Register a resource reload listener for a given resource manager type.
+	 *
+	 * <p>Note: This is only supported for server data reload listeners.
+	 *
+	 * @param identifier The identifier of the listener.
+	 * @param listenerFactory   A function that creates a new instance of the listener with a given registry lookup.
+	 */
+	void registerReloadListener(Identifier identifier, Function<RegistryWrapper.WrapperLookup, IdentifiableResourceReloadListener> listenerFactory);
```
zenfyrdev, line 63: I know that this is already merged, but what is the point of having an Identifier here? We're already registering an **Identifiable**ResourceReloadListener and later this just gets wrapped.

Wouldn't this be better something like this?
```java
void registerReloadListener(Function<RegistryWrapper.WrapperLookup, IdentifiableResourceReloadListener> listenerFactory);

void registerReloadListener(Function<RegistryWrapper.WrapperLookup, Collection<IdentifiableResourceReloadListener>> listenersFactory);
```

Since the helper does not expose all reloaders to the API, we could change the internals to support something like this?

modmuss50: We need the ID before creating the listner, this was very much a stop gap solution that didnt require re-writing the whole module. `IdentifiableResourceReloadListener` is a very old API, and is not how we would create this kind of API these days.

At somepoint I hope there will be a replacement resource loader module that improves this.

PR #3937, opened 2024-07-13
Add `ShadeMode` to FRAPI

PepperCode1:
`ShadeMode` is a material property and enum with two fields, `ENHANCED` and `VANILLA`. This material property is a hint to the renderer about how the quad is intended to be shaded, for example through ambient occlusion and diffuse shading. The renderer is free to ignore this hint. Similar functionality already existed in Indigo when using `AoConfig.HYBRID`, but a vanilla shade mode was hardcoded to always only be applied to quads output by the default implementation of `emitBlockQuads`. With a dedicated property exposed in the API, Indigo's code becomes cleaner and users have more control over this behavior.


fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/material/MaterialFinder.java
```diff
@@ -34,45 +35,56 @@
...
+	 *
+	 * @see BlendMode
 	 */
 	MaterialFinder blendMode(BlendMode blendMode);
 
 	/**
-	 * Vertex color(s) will be modified for quad color index unless disabled.
+	 * Controls whether vertex colors should be modified for quad coloring.
+	 *
+	 * <p>The default value is {@code false}.
```
Technici4n: That's not very clear. The default value is `{@code false}`, which means that the vertex colors _will be modified_. What about this?

```suggestion
	 * Controls whether vertex colors should be modified for quad coloring.
	 *
	 * <p>The default value is {@code false}, meaning that vertices will be colored according to the color index.
```

fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/renderer/v1/material/MaterialFinder.java
```diff
@@ -34,45 +35,56 @@
...
 	 */
 	MaterialFinder emissive(boolean isEmissive);
 
 	/**
-	 * Vertex color(s) will be modified for diffuse shading unless disabled.
+	 * Controls whether vertex colors should be modified for diffuse shading.
+	 *
+	 * <p>The default value is {@code false}.
```
Technici4n, line 73:
```suggestion
	 * <p>The default value is {@code false}, which enables diffuse shading.
```

Technici4n (changes requested):
I did not run the code, but I had a thorough look at every change. Looks very good.

Technici4n (approved):

modmuss50 (approved):

PR #3957, opened 2024-07-19
Add `c:slime_balls` tag, add more melee weapons to melee tag, add `c:fertilizers` tag, fix entries for some biome and food tags

### LINKED ISSUE #3946: Convention tags are missing tag for slimeballs
It seems like the fabric convention tags are missing slimeballs:
https://github.com/FabricMC/fabric/tree/1.21/fabric-convention-tags-v2/src/generated/resources/data/c/tags/item
https://github.com/neoforged/NeoForge/blob/1.21.x/src/generated/resources/data/c/tags/item/slimeballs.json
Usage in RS: https://github.com/refinedmods/refinedstorage2/blob/develop/refinedstorage-platform-common/src/main/resources/data/refinedstorage/recipe/processor_binding.json#L13

### LINKED ISSUE #3949: Should `c:tools/melee_weapon` contain Hoes, Pickaxes, and Shovels? (Maybe Trident too)
Someone brought up that the melee weapon tag inherited from Fabric API's past only contains Swords, Axes and Maces by default.
https://github.com/FabricMC/fabric/blob/1.21/fabric-convention-tags-v2/src/generated/resources/data/c/tags/item/tools/melee_weapon.json

Should this be changed to also have Hoes, Shovels, and Pickaxes? Specifically, any tool that has an ATTACK_DAMAGE attribute? (Maybe paired with ATTACK_SPEED attribute too?) Might also mean we should add Trident to it as it can be used both ranged and melee. 

If not, what would be a better way to describe this tag to someone to make it clear that only Swords, Axes, and Maces belong in this tag? Like weapons that are a viable melee weapon where being up close with physical attacks is the main attacking method as oppose to throwing or attacking being a weak secondary use like shovels.

Neo: https://github.com/neoforged/NeoForge/issues/1327

### LINKED ISSUE #3956: Add `c:fertilizers` tag
Someone started a convo here asking for `c:fertilizers` tag
https://github.com/neoforged/NeoForge/issues/1135

The tag would be for collection Bonemeal and other consumable items that can grow plants. I checked and it looks like Bonemeal simply calls a method on bonemealable blocks to grow plants. So there's no place or need for a functional tag under modloader namespace. This would be more for simply collecting the items for categorization usages.

I'm asking the original user rn for their exact use case but wanted to post this here in the meanwhile for feedback while I try and figure out what exactly is the usefulness of this tag

TelepathicGrunt:
- Adds Sparse Jungle to `c:is_sparse_vegetation/overworld` biome tag as it was missing from before. The biome belongs there. It got sparse in the name!

- Clarified the melee weapon tag's javadoc to make it clear pickaxes are excluded because being a weapon is not a pickaxe's primary purpose.

- Added `c:slime_balls` tag to match NeoForge and because someone requested it on Fabric. Showing there is use case for this tag likely due to mods that do Slimes that drop ore-based slimeballs or something.

- Added `c:fertilizers` tag so farming machines can know what items can be consumed/damaged in order to grow nearby crops/plants. This is needed as some mods do add their own bonemeal variant items such as Minecolonies Compost item. The growing logic is on the crop/plant so no hooks needed in item. Just need to know what item is a bonemeal-like item.

- Added Old Growth Spruce Taiga, Windswept Hills, Windswept Gravelly Hills, Windswept Forest, and Stony Shores to the `c:is_cold/overworld` tag.

- Moved Melon Slice from `c:foods/vegetable` to `c:foods/fruit` as melons are usually considered fruits.

- Removed Glistering Melon Slice from `c:foods/golden` because it is not actually edible and thus, not food.

Neo PR: https://github.com/neoforged/NeoForge/pull/1348

closes https://github.com/FabricMC/fabric/issues/3946
closes https://github.com/FabricMC/fabric/issues/3949
closes https://github.com/FabricMC/fabric/issues/3956


fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -57,8 +57,19 @@ private ConventionalItemTags() {
 	public static final TagKey<Item> MACE_TOOLS = register("tools/mace");
 
 	// Action-based tool tags
+	/**
+	 * A tag containing melee-based weapons for recipes and loot tables.
+	 * Tools are considered melee if they contain a GENERIC_ATTACK_DAMAGE attribute.
```
Chocohead:
```suggestion
	 * Tools are considered melee if they contain a {@link net.minecraft.entity.attribute.EntityAttributes#GENERIC_ATTACK_DAMAGE} attribute.
```

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -313,6 +324,12 @@ private ConventionalItemTags() {
 	public static final TagKey<Item> ROPES = register("ropes");
 	public static final TagKey<Item> CHAINS = register("chains");
 	public static final TagKey<Item> ENDER_PEARLS = register("ender_pearls");
+	public static final TagKey<Item> SLIMEBALLS = register("slimeballs");
+	/**
+	 * For bonemeal-like items that can grow plants.
+	 * (Note: Could include durability-based modded bonemeal-like items. Check for durability DataComponentType to handle them properly)
```
Chocohead:
```suggestion
	 * (Note: Could include durability-based modded bonemeal-like items. Check for durability {@link net.minecraft.component.ComponentType} to handle them properly)
```

Chocohead (changes requested):
Would be good to link Minecraft references so any changes in the future won't get missed in the Javadocs

Crendgrim:
Hm, as a mod dev I'm not a big fan of the change to `c:tools/melee_weapons`. It was nice to have a common tag that basically meant "this is a melee weapon". I don't see a good way to filter things out again because there's no common tags for shovels/hoes/paxels/whatever mods add that do damage but aren't actually melee weapons. Also, it is confusing that `c:tools` now includes `c:spear_tools`, but `c:tools/melee_weapons` hardcodes the trident. If a mod adds a trident, it automatically appears in one collective tag but not the other.

TelepathicGrunt:
@Crendgrim You can use the diff ingredients to filter out shovels, hoes, and pickaxes. Maybe even axes too if that's not something you want. So adding more entries is fine since we can filter them later in datapacks using modloader's diff ingredients if desired. It’s the same logic we encourage with the dyed tags. See the javadoc here for more info: https://github.com/FabricMC/fabric/blob/f9000df732297d583eb96c98ec0341da06c1d1c6/fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java#L245

the reason I did not nest tags into the melee tag is because someone can make a sword that does not have the attack damage attribute and instead, shoots fireball projectiles. That sword would be in sword tag because it is a sword. But it is not a melee weapon. Same goes for axes, spears, etc. it is better to individually define each item in the melee tag so modders can add their weapons only if it is truly a melee weapon and not some case where it doing unique behavior than the rest of the similar tools do

TelepathicGrunt:
@Crendgrim to clarify a bit more on first point, there are shovel and hoe tags. They are under the Mc namespace https://github.com/misode/mcmeta/blob/data-json/data/minecraft/tags/item/shovels.json

But again, if they define an attack damage attribute, they are defined to deal more damage than hand or other normal items. That makes them well, a weapon. More efficient at killing than empty hand. Now you may have specific use case for *specific* kinds of weapons and that’s where ingredient difference/intersection comes into play as that is its entire purpose to allow you to pick exactly what you need from a tag

modmuss50 (approved):

PR #3964, opened 2024-07-22
Update Dependencies

CanerKaraca23:


apple502j:
We do not accept dependency update-only PRs from first-time contributors. (These are generally updated without PRs.)

CanerKaraca23:
> We do not accept dependency update-only PRs from first-time contributors. (These are generally updated without PRs.)

It will be good then if you update all of these dependency updates.

PR #3966, opened 2024-07-22
Add a registry for registering additional blocks that a BlockEntityType supports

FoxSamu:
A `BlockEntity` can only be supported by a specific set of blocks. These sets are hardcoded by Minecraft per `BlockEntityType`. Modded block entities can define their own (also hardcoded) lists of supported blocks. However, some vanilla block entities are perfectly reusable on a mod's custom blocks. This PR adds an API to add to the sets of supported blocks that Minecraft hardcoded into the game.

Without this API, you can't use vanilla block entities on custom blocks easily. You must create a subclass of the desired `BlockEntity` and register a custom `BlockEntityType` for it. Potentially you should also add an appropriate renderer. This process is very cumbersome, since you're basically adding stuff that already exists.

For example, suppose you were to create a custom sign. You create and register a `SignBlock`. When you place this `SignBlock` in the world, it just looks like an invisible block. That's because the `SignBlockEntity` does not support your custom sign, so Minecraft removed it and the sign block entity renderer will not be invoked. This is where this PR comes to the rescue.

Fabric did already have mixins to tell Minecraft where to load modded sign textures from. It seems thus quite lacking to me that there was no API to allow modded blocks to use vanilla block entities. Though I'd make it, since I had this pretty much already laying around.

# Usage
Under the `fabric-content-registries` module, there is now a class `BlockEntitySupportedBlocksRegistry`. This class has a single static method `register` that takes in a `BlockEntityType` and a bunch of `Block`s that the `BlockEntityType` must support.

So let's say you registered a sign at `ModBlocks.MY_SIGN`, you can tell Minecraft to accept your block as a valid block for sign block entities as follows:
```java
BlockEntitySupportedBlocksRegistry.register(BlockEntityType.SIGN, ModBlocks.MY_SIGN);
```

# Internal working
A `BlockEntityType` has a private field `blocks` of type `Set<Block>`. This is the hardcoded set of blocks that Minecraft assigned to be valid for this block entity type. A mixin makes this set mutable and allows adding to it via an injected interface (`BlockEntityTypeHooks`). This is what the registry class calls after some small sanity checks.

# Controversies
I'm not sure whether the content registries module was the correct place to add this to. Maybe it should have been somewhere else but this did seem like the most fitting module to me.

An alternative way of implementing could have been by interface injection. A `FabricBlockEntityType` interface could have been implemented on `BlockEntityType` and it would have a method to add more supported blocks. I felt like that was kind of inconsistent with the rest so I didn't do that.


Fuzss:
This is very useful. 👍

But to achieve this it's not at all necessary to use a mixin + duck interface. Just access-widen `BlockEntityType#blocks`, making it `accessible` and `mutable` from `BlockEntitySupportedBlocksRegistry::register`.

FoxSamu:
Oh you're right, didn't think of it that way.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/BlockEntitySupportedBlocksRegistry.java
```diff
@@ -0,0 +1,51 @@
...
+	private BlockEntitySupportedBlocksRegistry() {
+	}
+
+	/**
+	 * Adds additional blocks that the given block entity type supports.
+	 *
+	 * @param type            The block entity type to add support for.
+	 * @param supportedBlocks The blocks to add as additional supported blocks.
+	 */
+	public static void register(BlockEntityType<?> type, Block... supportedBlocks) {
```
modmuss50, line 41: Could this not just be an injected interface?

modmuss50 (commented):
Would it be a good idea to limit this API to only be used before registry freezing? Thus any mods that want to query this can do so after the registry has been frozen, and can be sure that nothing will be added later?

modmuss50:
As @Fuzss said maybe a transitive accesswidener is enough? Makes things a lot simpler? Im not sure, its worth discussing.

FoxSamu:
It has no semantical difference to use an access widener, it's either that or the Mixin which makes the field mutable and accessible. The Mixin may be less compatible though so I changed it to the access widener form now.

FoxSamu:
> Would it be a good idea to limit this API to only be used before registry freezing? Thus any mods that want to query this can do so after the registry has been frozen, and can be sure that nothing will be added later?

I would have no clue how to do that. But I doubt if it's really that necessary, because of 3 reasons:
- The API only allows adding additional supported blocks.
- The `BlockEntityType.blocks` field is only used in `.supports(...)`. Thus any added block will be supported for the rest of the game's runtime.
- Any mod could just access-widen the same field and mutate it anyway after freezing.

However I can imagine it may be undesired when a block suddenly supports a block entity from some mid-game instant and onward, as some mod decided to just register it mid-game (though, a decent modder would know not to do that).

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/BlockEntitySupportedBlocksRegistry.java
FoxSamu: I thought about it but I felt it was a bit off with the other content registries. It's also because most injected interfaces are there to assist with modded objects/classes (because Mojang loves to make things unnecessarily tricky for mods sometimes), whereas this API alters the behaviour of a vanilla object.

Though, that's how I see it. There were a variety of ways to achieve the same result so I eventually just picked the one that I would be the most comfortable with using.

fabric-content-registries-v0/src/testmod/java/net/fabricmc/fabric/test/content/registry/ContentRegistryTest.java
```diff
@@ -169,6 +175,19 @@ public void onInitialize() {
 				builder.registerPotionRecipe(Potions.AWKWARD, Ingredient.ofItems(Items.BUNDLE), Potions.LUCK);
 			}
 		});
+
+		BlockEntitySupportedBlocksRegistry.register(BlockEntityType.SIGN, TEST_SIGN);
+
+		try {
+			BlockEntitySupportedBlocksRegistry.register(BlockEntityType.SIGN, null);
+			BlockEntitySupportedBlocksRegistry.register(BlockEntityType.SIGN, null, null);
+			BlockEntitySupportedBlocksRegistry.register(null, TEST_SIGN);
```
haykam821: Since the first method call is meant to throw an exception which exits this block, the behavior of the following method calls won't be tested.

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/BlockEntitySupportedBlocksRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+		Objects.requireNonNull(type, "Block entity type cannot be null!");
+		Objects.requireNonNull(supportedBlocks, "Supported blocks cannot be null!");
+
+		for (Block block : supportedBlocks) {
+			Objects.requireNonNull(block, "A supported block cannot be null!");
+		}
+
+		if (!(type.blocks instanceof HashSet<Block>)) {
+			type.blocks = new HashSet<>();
```
haykam821: Wouldn't this statement clear the default entries from the blocks set?

FoxSamu: Good point

fabric-content-registries-v0/src/testmod/java/net/fabricmc/fabric/test/content/registry/ContentRegistryTest.java
FoxSamu: This construct appears several times in this file actually, and they all have the same problem.

If I fix it, I would fix them all.

FoxSamu: Fixed using a JUnit like assertThrows method

FoxSamu:
I thought of an injected interface, it's also an option. However, I would put that into the block API module, which has injected interfaces for `Block` and `BlockState` too. I may make a different PR for that to compare the difference, if it's significant enough.

@modmuss50 any opinions?

modmuss50:
> I thought of an injected interface, it's also an option. However, I would put that into the block API module, which has injected interfaces for `Block` and `BlockState` too. I may make a different PR for that to compare the difference, if it's significant enough.
> 
> @modmuss50 any opinions?

I do think that an injected interface would be best here, I dont see the need for a full registry.

modmuss50:
Hey, @FoxSamu surprisingly we need this as a priority after some changes made in 1.21.1-rc1! I hope you dont mind me taking over this PR and moving it to interface injection (I will create a sperate PR).

FoxSamu:
~~[A distortion in the fabric of the universe made this knowledge come to me ahead of time]~~

@modmuss50 That's ok tho. I'm quite busy preparing a study exchange lately so I don't think I'll find time to work on this anyway

modmuss50:
> @modmuss50 That's ok tho. I'm quite busy preparing a study exchange lately so I don't think I'll find time to work on this anyway

No worries at all, I hope you have a great time! Thanks a lot for this PR. Ill close it as superseeded by https://github.com/FabricMC/fabric/pull/4009 Thanks again

PR #3977, opened 2024-07-29
Don't track fake player network addons

### LINKED ISSUE #3974: Memory leaks occur when players log in or log out.
Memory leaks occur every time a player logs in or logs out.

1. To facilitate testing, use Carpet's /player command to control player logins.
2. Place two command blocks, one to control player logins and the other to control logouts.
3. Enable the command blocks to run in a loop.
4. Open the Task Manager, and you can observe that the process memory keeps growing.
5. Although Carpet is used to assist in testing, after testing, it is found that memory leaks do not occur when fabric api is not installed.

每次玩家上线或下线时都会发生内存泄漏

1. 为了方便测试，使用Carpet的/player命令来控制玩家上线
2. 放置两个命令方块，一个控制玩家上线，一个控制下线
3. 将命令方块循环开启
4. 打开任务管理器，可以看到进程内存一直在增长
5. 虽然使用了Carpet来辅助测试，但经过测试，未安装fabric api时并没有发生内存泄漏
image
image
image
image

modmuss50:
Fixes #3974

Its possible we dont actually need the addon at all, but this is a much safer change if mods were expecting an addon.

### DIFF
--- a/fabric-events-interaction-v0/build.gradle
+++ b/fabric-events-interaction-v0/build.gradle
@@ -1,3 +1,3 @@
 version = getSubprojectVersion(project)
 
-moduleDependencies(project, ['fabric-api-base'])
+moduleDependencies(project, ['fabric-api-base', 'fabric-networking-api-v1'])
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/UntrackedNetworkHandler.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/UntrackedNetworkHandler.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


fabric-events-interaction-v0/build.gradle
```diff
@@ -1,3 +1,3 @@
 version = getSubprojectVersion(project)
 
-moduleDependencies(project, ['fabric-api-base'])
+moduleDependencies(project, ['fabric-api-base', 'fabric-networking-api-v1'])
```
apple502j, line 3: I believe this has to be declared on FMJ as well. see https://github.com/FabricMC/fabric/blob/1.21/fabric-screen-handler-api-v1/src/main/resources/fabric.mod.json

modmuss50: Yes, good catch. Fixed.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/UntrackedNetworkHandler.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.networking;
+
+// An internal marker interface used by the fake player API to prevent the network addon from being tracked.
+public interface UntrackedNetworkHandler {
```
Octol1ttle, line 20: should this be `@ApiStatus.Internal`?

modmuss50: No, this annotation is automatically applied to all impl packages, meaning as its in the impl package it will effectivity already have this annotation. See: https://github.com/FabricMC/fabric/blob/1.21/gradle/package-info.gradle

PR #3999, opened 2024-08-05
Revert registry sync changes. Allowing duplicate entries and vanilla clients on servers with modded registries.

### LINKED ISSUE #3993: Network Protocol Error when joining Hypixel
After updating to Fabric API 0.101.0 attempting to join Hypixel results in my client being kicked because of a Network Protocol Error related to registries, downgrading to 0.100.8 fixes the issue.

Here's the Network Protocol Error Report: https://mclo.gs/ADMTl1z
Regular Log: https://mclo.gs/rf4O2dJ

### LINKED ISSUE #3998: After update Bedrock/Geyser clients cannot join fabric servers
Hello, early this morning, I updated all my FabricMC server mods, including Fabric API. 

A bunch of users that connect using Bedrock client through GeyserMC reported that the server was returning a message requiring them to instal Fabric loader or Fabric API.

I think the change that triggered the issue is this one:

image

Does this represent a real issue or is something that should be fixed on Geyser's side?

Thx!!!!

modmuss50:
Revert the 2 registry sync changes as they are causing more trouble than they are solving, we can re-introduce these in 1.21.1.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
@@ -41,7 +41,6 @@ import org.jetbrains.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import net.minecraft.nbt.NbtCompound;
 import net.minecraft.network.packet.Packet;
 import net.minecraft.registry.Registries;
 import net.minecraft.registry.Registry;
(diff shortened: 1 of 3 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/RegistrySyncManager.java
```diff
@@ -159,9 +148,9 @@ public static <T extends RegistryPacketHandler.RegistrySyncPayload> CompletableF
 	}
 
 	/**
-	 * Creates a {@link Map} used to sync the registry ids.
+	 * Creates a {@link NbtCompound} used to sync the registry ids.
 	 *
-	 * @return a {@link Map} to sync, null when empty
+	 * @return a {@link NbtCompound} to sync, null when empty
```
Octol1ttle: Keep the correction?

modmuss50: Sure.

apple502j: (`an NBT`?)

PR #4009, opened 2024-08-07
Add an API to add additional supported blocks to block entity types.

modmuss50:
For both 1.21.1 and 1.21


Juuxel (approved):
LGTM (+ I already have a use case for this).

Issue #463, opened 2019-12-16
Network handshake design


Player3324:
The network handshake ensures that all mods have compatible counterparts between client and server. It also determines when mods can do their own handshakes, send initial data and when the game may continue the login process.

## Compatibility levels
The API supports the following approaches for various compatibility schemes:
1. don't check or communicate
2. require absence or...
2.1 some version range on the other side(from both sides, may be `*` for any)
2.2 same version on the other side
3. require presence...
3.1 with some version range on the other side (from both sides, may be `*` for any)
3.2 with the same version on both sides

Option 1 is for mods that don't communicate with their own copy on other side and have nothing that has to be registered/present on both sides. It is also suitable for anything with a custom handshake or its own presence detection+validation, e.g using the `register` plugin channel.

Option 2 is for mods that can run stand-alone like in 1, but have an optional communication component depending on presence and compatibility. This is frequently seen in client mods that improve or enable extra features if their server part is available and vice versa. In order to let the mods using this option enable those features, the Fabric handshake carries the mod presence and version information back across the network. Specifying `*` in the version range disables any checking, making it announce only.

Option 3 is what mods content mods should use. It guarantees that the desired version is present after the handshake, otherwise the connection will have been dropped.

2.2 and 3.2 can be represented by a version range as in 2.1/3.1, it just eliminates the need to specify the own mod's version in multiple places.

Most mods will likely use 3.2, the default may have to be 1 for a while to retain compatibility.

### Interaction with differently specified requirements

Mismatched requirement specifications happen when using different versions of a mod and the mod changed its requirements. The allowed version range acts as the intersection of the two ranges, the presence requirement as the strictest. Announcements from 2.x happen as usual.

## Mod-side declaration

Mods specify their c<->s compatibility requirements declaratively using their fabric-mod.json. The specific format/keys are TBD, but likely standardized properties outside the custom block with the rationale that Fabric API acts as an implementation of a certain Loader feature provision.

One option is:
```
requireNetworkPresence=true
acceptNetworkVersion="same"
```
where requireNetworkPresence is true for 3.x, false for 1 and 2.x and acceptNetworkVersion is same for 2/3.2 or a version range as used by dependencies for 2/3.1. The mod id to check against is the same as the declaring mod.

## Implementation

Instead of relying on sending the entire mod list to the other side or asking for versions of certain mods and verifying the response, a demand based scheme can save both bandwidth and latency.

[...]

PR #4011, opened 2024-08-07
The Mod Protocol API, aka better validation of mod presence of both-side mods with extra protocol version syncing

Patbox:
This api implements Mod Protocols, a simple way for a mods to quickly check if they are present on client/server, while also allowing for prevention of joining if it's missing on either side (through not being forced to). It can be also used for syncing compatible protocol versions for general networking api usage or feature toggling, without making the server/client incompatible with vanilla clients/servers.

Example screenshots of mismatch screen:
## Vanilla client joining modded server:
image
## Modded client with some mismatched mods joining modded server:
image
## Modded client (with server required entries) joining vanilla server:
image

Vanillish clients get disconnected by the server, while rest is validated client side.

(Also see https://github.com/Patbox/fabric/blob/mod-protocol/fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java for info)

Worth noting for people who might get confused, this is not an "anticheat"! This api allows any mod to register any protocol names and even then cheat clients could easily work around it. This api is purely to improve experience of regular users / limit amount of reports for mod devs or pack makers.

One thing to do with this pr would be bit more testing + maybe writing some unit tests, through from what I've checked everything should work. Otherwise I think it should be all done.


fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/ModProtocolLocator.java
```diff
@@ -0,0 +1,131 @@
...
+import net.fabricmc.loader.api.metadata.CustomValue;
+import net.fabricmc.loader.api.metadata.ModMetadata;
+
+public class ModProtocolLocator {
+	public static void provide(BiConsumer<ModContainer, ModProtocolImpl> consumer) {
+		for (var mod : FabricLoader.getInstance().getAllMods()) {
+			create(mod, consumer);
+		}
+	}
+	private static void create(ModContainer container, BiConsumer<ModContainer, ModProtocolImpl> consumer) {
```
maityyy, line 38: Is it possible to implement this more nicely via `Codec::validate`? (`ModProtocolImpl.CODEC::validate`)

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+/**
+ * The Mod Protocol API (client side), version 1.
+ *
+ * See {@link net.fabricmc.fabric.api.modprotocol.v1}
+ */
+
+@ApiStatus.Experimental
+package net.fabricmc.fabric.api.client.modprotocol.v1;
```
modmuss50, line 24: I wonder if this will clash with the generated package-info files 🤔 Need to see a green build to be sure after the checkstyle issues are solved.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/ModProtocolLocator.java
Patbox: Wrong line? This code parses it by hand, as you can't use codecs with loaders custom values (or at least I didn't know about it being an option). Also this is more flexible than that codec, which is only used for attaching mods to server's metadata

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.modprotocol.v1.ModProtocol;
+import net.fabricmc.fabric.api.modprotocol.v1.ModProtocolIds;
+import net.fabricmc.fabric.impl.modprotocol.RemoteProtocolStorage;
+
+/**
+ * Utility methods allowing to get protocol versions supported by the server.
+ *
+ * <p>Protocol identifier's can be any valid identifier, through by default mods defining it will use "mod" namespace and path equal to its id.
```
apple502j:
```suggestion
 * <p>Protocol identifiers can be any valid {@link Identifier}. The default is {@code mod:(mod ID)}.
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+ * See {@link ModProtocolIds} for more information.</p>
```
apple502j:
```suggestion
 * @see ModProtocolIds
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public final class ClientModProtocolLookup {
+	public static final int UNSUPPORTED = -1;
```
apple502j, line 40: Should have a javadoc.

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	private ClientModProtocolLookup() { }
+
+	/**
+	 * Gets protocol version supported by the server.
```
apple502j:
```suggestion
	 * Gets the protocol version supported by the server.
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @param handler the network handler connected to the server
+	 * @param protocolId protocol's id
+	 * @return Protocol version supported by the server
```
apple502j:
```suggestion
	 * @return the protocol version supported by the server
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public static int getSupportedProtocol(ClientCommonNetworkHandler handler, Identifier protocolId) {
+		return RemoteProtocolStorage.getProtocol(handler, protocolId);
+	}
+
+	/**
+	 * Gets protocol version supported by the server.
```
apple502j:
```suggestion
	 * Gets the protocol version supported by the server.
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @param connection the ClientConnection connected to the server
+	 * @param protocolId protocol's id
+	 * @return Protocol version supported by the server
```
apple502j:
```suggestion
	 * @return the protocol version supported by the server
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public static int getSupportedProtocol(ClientConnection connection, Identifier protocolId) {
+		return RemoteProtocolStorage.getProtocol(connection, protocolId);
+	}
+
+	/**
+	 * Gets protocol version supported by the server.
```
apple502j:
```suggestion
	 * Gets the protocol version supported by the server.
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @param handler the network handler connected to the server
+	 * @param protocol protocol to check against
+	 * @return Protocol version supported by the server
```
apple502j:
```suggestion
	 * @return the protocol version supported by the server
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public static int getSupportedProtocol(ClientCommonNetworkHandler handler, ModProtocol protocol) {
+		return RemoteProtocolStorage.getProtocol(handler, protocol.id());
+	}
+
+	/**
+	 * Gets protocol version supported by the server.
```
apple502j:
```suggestion
	 * Gets the protocol version supported by the server.
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @param connection the ClientConnection connected to the server
+	 * @param protocol protocol to check against
+	 * @return Protocol version supported by the server
```
apple502j:
```suggestion
	 * @return the protocol version supported by the server
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public static int getSupportedProtocol(ClientConnection connection, ModProtocol protocol) {
+		return RemoteProtocolStorage.getProtocol(connection, protocol.id());
+	}
+
+	/**
+	 * Gets all protocols supported by the server.
+	 *
+	 * @param handler the network handler connected to the server
+	 * @return Map of protocols supported by the server
```
apple502j:
```suggestion
	 * @return the map of protocols to the versions supported by the server
```

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/ClientModProtocolLookup.java
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public static Object2IntMap<Identifier> getAllSupportedProtocols(ServerCommonNetworkHandler handler) {
+		return RemoteProtocolStorage.getMap(handler);
+	}
+
+	/**
+	 * Gets all protocols supported by the server.
+	 *
+	 * @param connection the ClientConnection connected to the server
+	 * @return Map of protocols supported by the server
```
apple502j:
```suggestion
	 * @return the map of protocols to the versions supported by the server
```

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/ModProtocolLocator.java
maityyy: Ah, yes, Fabric Loader doesn't give a JsonElement, I forgot about that

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * @param identifier the identifier of protocol
+	 * @param name display name in protocol, shown if it's missing
+	 * @param version display version of the protocol, shown if it's missing
+	 * @param protocol list of protocol versions
+	 * @param requireClient marks protocol as required on client
+	 * @param requireServer marks protocol as required on server
+	 * @return registered Mod Protocol
+	 */
+	public static ModProtocol register(Identifier identifier, String name, String version, IntList protocol, boolean requireClient, boolean requireServer) {
```
apple502j, line 66: I feel like this should be a builder.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * <p>While lookup can be done anytime, registration is only possible before registries are frozen.</p>
+ */
+public final class ModProtocolRegistry {
+	private ModProtocolRegistry() { }
+
+	/**
+	 * Allows to get a locally registered protocol.
+	 * @param identifier protocol's id
+	 * @return requested ModProtocol or null if not found
```
apple502j:
```suggestion
	 * @return the requested ModProtocol or {@code null} if not found
```

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
```diff
@@ -0,0 +1,79 @@
...
+	 * @param requireClient marks protocol as required on client
+	 * @param requireServer marks protocol as required on server
+	 * @return registered Mod Protocol
+	 */
+	public static ModProtocol register(Identifier identifier, String name, String version, IntList protocol, boolean requireClient, boolean requireServer) {
+		return ModProtocolManager.add(null, new ModProtocolImpl(identifier, name, version, IntList.of(protocol.toIntArray()), requireClient, requireServer));
+	}
+
+	/**
+	 * Allows to customise priority of namespaces when displaying missing protocols.
```
apple502j:
```suggestion
	 * Allows to customize priority of namespaces when displaying missing protocols.
```

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ServerModProtocolLookup.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	/**
+	 * Gets protocol version supported by the player.
+	 *
+	 * @param player the player
+	 * @param protocolId protocol's id
+	 * @return Protocol version supported by the player
+	 */
+	public static int getSupportedProtocol(ServerPlayerEntity player, Identifier protocolId) {
```
apple502j, line 50: Not sure how useful this would be, as SPE is created after the Config phase concludes.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *     		<dt>"fabric:mod_protocol": {
+ *     		 "protocol": [1, 2],
+ *     		 "id: "custom:id",
+ *     		 "name": "Mod Name",
+ *     		 "version": "v1.2.3",
+ *     		 "require_client": false,
+ *     		 "require_server": true
+ *     		}</dt>
+ *      	<dd><p>Full object. Only required value is "protocol", which can be set directly for single version or as an array for multiple.</p>
+ *      	<p>"id" is protocols identifiers, which can have any namespace and path, as long as it's valid.
```
apple502j:
```suggestion
 *      	<p>"id" is the protocols identifier, which can have any namespace and path, as long as it's valid.
```

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *      	It's optional and defaults to id with "mod" namespace and path equal to mod's id.
```
apple502j:
```suggestion
 *      	It is optional and defaults to an ID with "mod" namespace and path equal to mod's id.
```

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *       		 "version": "v1.2.3",
+ *       		 "require_client": false,
+ *       		 "require_server": true
+ *       		}]</dt>
+ *        	<dd><p>Array of full objects. Allows to define multiple versions of the protocol. The inner objects use the same format as single-full object format,
+ *        	with main exception being that fields "id", "name" and "version" aren't defaulted and need to be always set</p></dd>
+ *     </dl>
+ *     </dd>
+ *     <dt>{@link net.fabricmc.fabric.api.modprotocol.v1.ModProtocolRegistry}</dt>
+ *     <dd>This is the simplest way to define it. Can be useful when you don't need to change it depending on mods configuration
```
apple502j, line 66: Needs changing

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/ModProtocolImpl.java
```diff
@@ -0,0 +1,82 @@
...
+	public static final Codec<List<ModProtocolImpl>> LIST_CODEC = CODEC.listOf();
+	public static final PacketCodec<PacketByteBuf, ModProtocolImpl> PACKET_CODEC = PacketCodec.ofStatic(ModProtocolImpl::encode, ModProtocolImpl::decode);
+
+	private static ModProtocolImpl decode(PacketByteBuf buf) {
+		Identifier id = buf.readIdentifier();
+		String name = buf.readString();
+		String version = buf.readString();
+		IntList protocols = IntList.of(buf.readIntArray());
+		byte b = buf.readByte();
+		boolean requireClient = (b & 0b10) != 0;
```
apple502j, line 51: Maybe just use readBoolean?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/TextUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.text.Text;
+
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+
+public class TextUtil {
+	private static final Map<String, String> FALLBACK_TRANSLATIONS = new HashMap<>();
+
+	static {
+		try {
```
apple502j, line 38: Is this necessary?

gradle.properties
```diff
@@ -38,6 +38,7 @@ fabric-lifecycle-events-v1-version=2.3.12
 fabric-loot-api-v2-version=3.0.14
 fabric-loot-api-v3-version=1.0.2
 fabric-message-api-v1-version=6.0.13
+fabric-mod-protocol-api-v1-version=0.0.0
```
apple502j: This starts at `1.0.0`.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -293,6 +293,32 @@ public static void send(ServerPlayerEntity player, CustomPayload payload) {
 		player.networkHandler.sendPacket(createS2CPacket(payload));
 	}
 
+	/**
+	 * Gets the brand of the client player connected with.
+	 *
+	 * @param player the player
+	 * @return client's brand, which might be null if not sent.
```
apple502j:
```suggestion
	 * @return client's brand, or {@code null} if not sent
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -293,6 +293,32 @@ public static void send(ServerPlayerEntity player, CustomPayload payload) {
...
+		Objects.requireNonNull(player, "Server player entity cannot be null");
+
+		return getBrand(player.networkHandler);
+	}
+
+	/**
+	 * Gets the brand of the client player connected with.
+	 *
+	 * @param handler the network handler, representing the connection to the player/client
+	 * @return client's brand, which might be null if not sent.
```
apple502j:
```suggestion
	 * @return client's brand, or {@code null} if not sent
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerConfigurationNetworking.java
```diff
@@ -224,6 +224,19 @@ public static void send(ServerConfigurationNetworkHandler handler, CustomPayload
 
 	// Helper methods
 
+	/**
+	 * Gets the brand of the client player connected with.
+	 *
+	 * @param handler the network handler, representing the connection to the player/client
+	 * @return client's brand, which might be null if not sent.
```
apple502j:
```suggestion
	 * @return client's brand, or {@code null} if not sent
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -293,6 +293,32 @@ public static void send(ServerPlayerEntity player, CustomPayload payload) {
...
+	 */
+	@Nullable
+	public static String getBrand(ServerPlayerEntity player) {
+		Objects.requireNonNull(player, "Server player entity cannot be null");
+
+		return getBrand(player.networkHandler);
+	}
+
+	/**
+	 * Gets the brand of the client player connected with.
```
apple502j:
```suggestion
	 * Gets the brand of the client the player connected with.
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerPlayNetworking.java
```diff
@@ -293,6 +293,32 @@ public static void send(ServerPlayerEntity player, CustomPayload payload) {
 		player.networkHandler.sendPacket(createS2CPacket(payload));
 	}
 
+	/**
+	 * Gets the brand of the client player connected with.
```
apple502j:
```suggestion
	 * Gets the brand of the client the player connected with.
```

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerConfigurationNetworking.java
```diff
@@ -224,6 +224,19 @@ public static void send(ServerConfigurationNetworkHandler handler, CustomPayload
 
 	// Helper methods
 
+	/**
+	 * Gets the brand of the client player connected with.
```
apple502j:
```suggestion
	 * Gets the brand of the client the player connected with.
```

fabric-mod-protocol-api-v1/src/testmod/resources/fabric.mod.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "Fabric Mod Protocol API (v1) Test Mod",
+  "version": "1.0.0",
+  "environment": "*",
+  "license": "Apache-2.0",
+  "depends": {
+    "fabric-mod-protocol-api-v1": "*"
+  },
+  "entrypoints": {
+    "main": [
+      "net.fabricmc.fabric.test.modprotocol.ModProtocolTestmods::init"
```
apple502j: We prefer to implement `ModInitializer`.

fabric-mod-protocol-api-v1/src/testmod/resources/fabric.mod.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "client": [
```
apple502j: Remove this

fabric-mod-protocol-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,42 @@
...
+    "irc": "irc://irc.esper.net:6667/fabric",
+    "issues": "https://github.com/FabricMC/fabric/issues",
+    "sources": "https://github.com/FabricMC/fabric"
+  },
+  "authors": [
+    "FabricMC"
+  ],
+  "entrypoints": {
+    "main": [
+      "net.fabricmc.fabric.impl.modprotocol.ModProtocolInit::init"
```
apple502j: same, please use `ModInitializer`

fabric-mod-protocol-api-v1/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,42 @@
+{
```
apple502j, line 1: Indent appears to be messed up.

fabric-mod-protocol-api-v1/src/main/resources/assets/fabric-mod-protocol-api-v1/lang/en_us.json
```diff
@@ -0,0 +1,8 @@
+{
+  "text.fabric.mod_protocol.mismatched.title": "Required protocols are mismatched!",
```
apple502j, line 2: I think this title is unclear, but I haven't come up with a better one 🤷🏼

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/payload/ModProtocolResponseC2SPayload.java
```diff
@@ -0,0 +1,42 @@
...
+import net.minecraft.network.packet.CustomPayload;
+import net.minecraft.util.Identifier;
+
+public record ModProtocolResponseC2SPayload(Object2IntMap<Identifier> supported) implements CustomPayload {
+	public static final Id<ModProtocolResponseC2SPayload> ID = new Id<>(Identifier.of("fabric", "mod_protocol/response"));
+	public static final PacketCodec<PacketByteBuf, ModProtocolResponseC2SPayload> PACKET_CODEC =
+			PacketCodecs.map(ModProtocolResponseC2SPayload::createMap, Identifier.PACKET_CODEC, PacketCodecs.INTEGER)
+			.xmap(ModProtocolResponseC2SPayload::new, ModProtocolResponseC2SPayload::supported).cast();
+
+	private static Object2IntMap<Identifier> createMap(int i) {
```
apple502j, line 34: Maybe there should be a cap on the number of protocols clients can send.

apple502j (changes requested):

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ServerModProtocolLookup.java
Patbox: The protocols are fully usable/accessible all the time and can be used together with play networking api. Thats why I included it.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/TextUtil.java
Patbox: Alternative option would be to duplicate the translations in code and then keeping both synced. This felt like a cleaner approach (as a bonus it makes it work when resource loader isn't present).

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/payload/ModProtocolResponseC2SPayload.java
Patbox: Would require writing a custom lazy codec since the class is initialized on init, but should be doable. Client should never sent more than amount server did.

fabric-mod-protocol-api-v1/src/main/resources/assets/fabric-mod-protocol-api-v1/lang/en_us.json
Patbox: Tbf it's best one I could think of too

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
Patbox: The only fields it could safely default would be `requireX` ones, so not sure if builder makes sense (since you need to specify rest of them anyway).

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/TextUtil.java
apple502j: Hmm, I always thought vanilla loaded the English one as a fallback.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/payload/ModProtocolResponseC2SPayload.java
apple502j: Can be a hardcoded cap, just so that bad clients won't request a map of size 1M.

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/impl/modprotocol/client/ClientModProtocolInit.java
```diff
@@ -28,8 +29,8 @@
 import net.fabricmc.fabric.impl.modprotocol.payload.ModProtocolRequestS2CPayload;
 import net.fabricmc.fabric.impl.modprotocol.payload.ModProtocolResponseC2SPayload;
 
-public final class ClientModProtocolInit {
-	public static void clientInit() {
+public final class ClientModProtocolInit implements ClientModInitializer {
+	public void onInitializeClient() {
 		ClientConfigurationNetworking.registerGlobalReceiver(ModProtocolRequestS2CPayload.ID, (payload, context) -> {
```
apple502j, line 34: Please use `@Override`.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/TextUtil.java
Patbox: It does, but only for vanilla (server.jar) one. Fabric has a patch for loading modded ones, but it might not be present when this is used standalone.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/payload/ModProtocolResponseC2SPayload.java
Patbox: There is no hardcap on api side, so ideally it would be dynamic

christopherplaysminecraft:
Just wondering, wouldn't this allow any server to continuously annoy users by requesting them to install a mod?

Also wondering about the wording in the original message
> through not being forced to

Does that mean that one can simply tell it (click a button) to ignore any missing mods?

modmuss50:
It has always been possible (and quite easy) for a server mod to write code that disconnects the player if the corresponding client side mod isn't present. This PR also won't allow server owners to request any client mod to be installed.

christopherplaysminecraft:
Okay
Tbh I understood this whole PR to basically just do that... whoops.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *     		with display name and version being copied form mods metadata. It also marks the protocol as required on both client and server.</p>
+ *     		</dd>
+ *     		<dt>"fabric:mod_protocol": {
+ *     		 "protocol": [1, 2],
+ *     		 "id: "custom:id",
+ *     		 "name": "Mod Name",
+ *     		 "version": "v1.2.3",
+ *     		 "require_client": false,
+ *     		 "require_server": true
+ *     		}</dt>
```
modmuss50, line 43: This should be versioned, e.g `fabric:mod_protocol_v1`

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *      	<dd><p>Full object. Only required value is "protocol", which can be set directly for single version or as an array for multiple.</p>
+ *      	<p>"id" is the protocols identifier, which can have any namespace and path, as long as it's valid.
+ *      	It is optional and defaults to an ID with "mod" namespace and path equal to mod's id.
+ *      	</p>
+ *     		<p>"name" is a name displayed if protocol doesn't match. It's optional and by default it uses one from mod's metadata.</p>
```
modmuss50, line 48: Im not sure this is that useful tbh, why would a mod have 2 names?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *          <p>"version" is a version displayed if protocol doesn't match. It's optional and by default it uses one from mod's metadata.</p>
```
modmuss50, line 49: Again, why duplicate data?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *          <p>"require_client" controls if clients without this protocol can join the server, defaults to true, preventing joining</p>
+ *          <p>"require_server" controls if clients can join servers without this mod, defaults to true, preventing joining</p>
```
modmuss50, line 51: Is true the correct default here? I wonder if most mods would be happy to join a server without their mod 🤔 Maybe there should be no default for these values?

modmuss50, line 51: Could these be merged into one field, with 3 values, e.g `both`, `client_required` and `server_required`?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
```diff
@@ -0,0 +1,75 @@
...
+ *     		</dd>
+ *     		<dt>"fabric:mod_protocol": {
+ *     		 "protocol": [1, 2],
+ *     		 "id: "custom:id",
+ *     		 "name": "Mod Name",
+ *     		 "version": "v1.2.3",
+ *     		 "require_client": false,
+ *     		 "require_server": true
+ *     		}</dt>
+ *      	<dd><p>Full object. Only required value is "protocol", which can be set directly for single version or as an array for multiple.</p>
```
modmuss50, line 44: Further expansion on how the protocol works be useful here, its the most important bit!

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Interface representing registered ModProtocol. Can be used for further lookups.
+ */
+@ApiStatus.NonExtendable
+public interface ModProtocol {
+	/**
+	 * @return Identifier associated with this Mod Protocol
+	 */
+	Identifier id();
```
modmuss50, line 32: Question (ive not looked at the impl), where does this ID come from when a protocol is specified in the fabric.mod.json?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
```diff
@@ -0,0 +1,53 @@
...
+	 */
+	String name();
+	/**
+	 * @return Display version of this protocol
+	 */
+	String version();
+	/**
+	 * @return Protocol versions supported by this protocol
+	 */
+	IntList protocol();
```
modmuss50, line 44: Nit: any reason not to use `int[]`?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
```diff
@@ -0,0 +1,53 @@
...
+	 */
+	Identifier id();
+	/**
+	 * @return Display name of this protocol
+	 */
+	String name();
+	/**
+	 * @return Display version of this protocol
+	 */
+	String version();
```
modmuss50, line 40: Somewhat realted to my above question, why are these needed, would it be better tieing this to the `ModContainer`/modid?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolIds.java
```diff
@@ -0,0 +1,41 @@
...
+
+import net.minecraft.util.Identifier;
+
+/**
+ * Utility methods allowing to create Identifiers targeting default protocols.
+ */
+public final class ModProtocolIds {
+	public static final String MOD = "mod";
+	public static final String SPECIAL = "special";
+	public static final String FEATURE = "feature";
```
modmuss50, line 27: Explain what these do and are used for in the javadoc. I have no clue just looking at the API.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
modmuss50: I agree a builder would be best here, especially if/when more fields may be added in the future. As I alluded to above, I wonder if this would be better off being more strictly tied to the `ModContainer`, removing the need to specify the name/version.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ServerModProtocolLookup.java
modmuss50: Whats the usecase for getting this later on? I do agree this doesnt seem that helpful.

Im also not too keen of APIs that return -1, I think it should throw (name the method to suggest it throws) and also provide one that returns an `OptionalInt`

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientLoginNetworkAddon.java
```diff
@@ -103,6 +103,11 @@ protected void handleRegistration(Identifier channelName) {
 	protected void handleUnregistration(Identifier channelName) {
 	}
 
+	@Override
+	public @Nullable String getBrand() {
```
modmuss50, line 107: Ideally this wouldnt exist on the login addon, could the root method go in `ClientCommonNetworkAddon`? Im not too worried about this as its impl, and can easily be refactored later. Or even sooner before this PR goes in as its own thing.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerConfigurationNetworking.java
```diff
@@ -224,6 +224,19 @@ public static void send(ServerConfigurationNetworkHandler handler, CustomPayload
...
 	// Helper methods
 
+	/**
+	 * Gets the brand of the client the player connected with.
+	 *
+	 * @param handler the network handler, representing the connection to the player/client
+	 * @return client's brand, or {@code null} if not sent
+	 */
+	@Nullable
+	public static String getBrand(ServerConfigurationNetworkHandler handler) {
```
modmuss50, line 234: What are your toughts on moving this out into its own PR? Will make things a bit easier to review, this change can likely go in quite quickly.

modmuss50: It would also be nice to have a way to get the servers brand.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractChanneledNetworkAddon.java
```diff
@@ -87,6 +90,10 @@ public boolean handle(CustomPayload payload) {
 			}
 		}
 
+		if (payload instanceof BrandCustomPayload brand) {
+			this.brand = brand.brand();
```
modmuss50, line 94: I'd return false after this point. False becuase we only need to "peek" at this packet and not consume it, allowing vanilla or any exisitng mod to also read it.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ServerPlayNetworkHandlerMixin.java
```diff
@@ -44,9 +45,13 @@ abstract class ServerPlayNetworkHandlerMixin extends ServerCommonNetworkHandler
 	}
 
 	@Inject(method = "<init>", at = @At("RETURN"))
-	private void initAddon(CallbackInfo ci) {
+	private void initAddon(CallbackInfo ci, @Local(argsOnly = true) ClientConnection connection) {
 		this.addon = new ServerPlayNetworkAddon((ServerPlayNetworkHandler) (Object) this, connection, server);
 
+		if (connection.getPacketListener() instanceof NetworkHandlerExtensions extension) {
+			this.addon.setBrand(extension.getAddon().getBrand());
+		}
```
modmuss50, line 53: A comment explaing this may help.

modmuss50 (commented):
This looks like a great start, just got a few points/questions regarding the API and networking changes before looking at the impl in detail.

I think the protocols should be attached to the `ModContainer`/modid in a stronger fashion, I dont think a general ID has much benefit.

I am still a little bit skeptical of using int ids as apposed to a semver version, I worry people wont remeber to change their protocol id. My inital thoughts about this problem were that we would re-use the same versioning system as used by loading when determining mod compatibility. I would expect most mod developers will just do a strict match of the specific versions, so its a setup and forget deal. I think we should discuss this and agree on the spec before spending ages perfecting the impl. See https://github.com/FabricMC/fabric/issues/463 for some prior thoughts on this topic.

I like having an API to query the client/server branding, I think this would be better off as a seperate PR so its not lost within this one, and could likely go in quite quickly. I would strongly suggest splitting that into another PR.

fabric-mod-protocol-api-v1/src/client/java/net/fabricmc/fabric/api/client/modprotocol/v1/package-info.java
modmuss50: Its a none issue, as its in the api package of course.

modmuss50:
In its current form I dont believe this supports the use case where you have an optional mod but want to have a requiment when it does exist.

E.g you have a mod that is optional on the server, but the protocol must match if it does exist on the client. Care should be taken to allow a vanilla client to join in this example.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
Patbox: The protocol array support (and naming/etc) is mostly done to mirror "provides", allowing people to check against old ids when they change/include other mods protocol easily.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
Patbox: Same as above

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
Patbox: Wasn't sure myself, but though it makes sense to default to something since the just number inline does. But can be changed

fabric-networking-api-v1/src/client/java/net/fabricmc/fabric/impl/networking/client/ClientLoginNetworkAddon.java
Patbox: Originally I wanted to use it purely for changing the message (through ended up using the same one for everyone), so yeah I can drop this for now

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
Patbox: Provided IntList is inmutable, preventing accidental modification

Patbox:
> This looks like a great start, just got a few points/questions regarding the API and networking changes before looking at the impl in detail.
> 
> I think the protocols should be attached to the `ModContainer`/modid in a stronger fashion, I dont think a general ID has much benefit.
> 
> I am still a little bit skeptical of using int ids as apposed to a semver version, I worry people wont remeber to change their protocol id. My inital thoughts about this problem were that we would re-use the same versioning system as used by loading when determining mod compatibility. I would expect most mod developers will just do a strict match of the specific versions, so its a setup and forget deal. I think we should discuss this and agree on the spec before spending ages perfecting the impl. See #463 for some prior thoughts on this topic.
> 
> I like having an API to query the client/server branding, I think this would be better off as a seperate PR so its not lost within this one, and could likely go in quite quickly. I would strongly suggest splitting that into another PR.

Protocols are dettached from containers due to in api usage, which can be useful for other tools. This is why it's namespaced.

Using strings would be bit more annoying when trying to use the getSupportedProtocol api's for any checks, since you might need to do a bit more matching. Aside of not all mods using valid semver in the first place. Ints are simplest way to check against version compat (which most software, including Minecraft does for protocol). My idea for improving visibility, is to put it alongside other values in `gradle.properties`, which makes it as simple of a change as version one. Having dependency style matching won't work for more complex cases or will make people more likely to make it way stricter than it needs to be, which could end up being annoying for modpack server makers which might want to get a hotfix for their server quickly without needing to wait for full modpack update.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
Patbox: By default it's "mod:(modid)", but can be overriten with any other identifier

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
Patbox: Since (currently) mod can provide multiple protocols (or make them in code) it still makes sense for this to exist. As a bonus people might want to have it different/more specific with something like `feature` namespace.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolIds.java
Patbox: By default mostly to indicate the source and display priority, special being always displayed first, then mod and ending up with feature. But these can be added upon in event-ordering styled fashion

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerConfigurationNetworking.java
Patbox: Sure thing

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
modmuss50: Being able to change other mods protocols seems like a massive recipe for confusion. Is this not what the dependency system already handles?

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocol.java
modmuss50: I do think tieing it to the mod solves a lot of problems mostly by removing a lot of duplicated data from the fmj. You also get support for "provides" for free.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolIds.java
modmuss50: I'd just keep it simple and sort alphabetically for now. Maybe add it later as a follow up, or even have a custom UI drawn by the client making it easier to consume the data.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
Patbox: This is not really for changing other mods protocol. It's mostly to mirror the "provides" functionality (your own mod changing it's id, but still being network compatible). Through to be fair it could be moved to allow this in code only

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolIds.java
Patbox: Sorting is already done and yeah it still sorts alphabetically for paths. It's mostly useful as another mod could register a modpack version/name in the special namespace, making it always display as the first thing

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/TextUtil.java
apple502j: Resource Loader is expected to be always there, I think.

modmuss50: I would suggest always getting the client to initiate the disconnect, this way the error can always be localised, and possibly replaced with a custom UI at a later date.

Patbox: It does work like this currently (at least for clients with fabric api), but the text is created (and does setup the fallback property) in common so the same message can be displayed on vanilla clients

modmuss50:
> Protocols are dettached from containers due to in api usage, which can be useful for other tools. This is why it's namespaced.

Can you expand on this? I dont see what issues attaching them to containers/modids would cause.

> Using strings would be bit more annoying when trying to use the getSupportedProtocol api's for any checks, since you might need to do a bit more matching.

I am not a massive fan of `getSupportedProtocol` I dont think this whole PR should be used for versioning packets, the main goal is to check that the client is compatible. Versioning packets is better done by using `canSend` imo.

> Aside of not all mods using valid semver in the first place.

We could easily have a fallback to support basic comparsions, such as the version being identical. Mods are highly encouraged to use semver, you already miss out on a lot of stuff and get a warning from loader if you dont.

> Ints are simplest way to check against version compat

The code for comparing versions already exists in loader, we need to make this easy for most people to use in their mods. I really dont see people updating an int every time they break protocol, it gets even more complex when you have 2 or more branches for differing minecraft versions. These problems have all been solved with semver.

> it alongside other values in gradle.properties

I dont think we can do this, processResources is not suited to this. One solution may be to move everything to the fmj but this is another discussion.

> Having dependency style matching won't work for more complex cases

Such as? There are complex cases that this PR in its current state doesnt handle.

> will make people more likely to make it way stricter than it needs to be

I agree with this, I would expect the default to allow patch version changes. We need to find a good middle ground as like I said I dont see people every changing this int value, I know I will certainly forget.

apple502j:
> Can you expand on this? I dont see what issues attaching them to containers/modids would cause.

String ID would allow mods to define multiple protocols, though I am unsure why you need them.

re: `canSend` - I agree this module should only be used for the automatic disconnection, fail-soft protocol checks should continue to use `canSend`.

Patbox:
> Can you expand on this? I dont see what issues attaching them to containers/modids would cause.
A single mod can for example change vanilla protocol depending on it's configuration, which then could just use mod protocol to inform about it / enforce it without requiring it's own syncing (or ending up with bad errors). Other thing this could be used for is a mod to register configurable protocol (for example for the modpack), to inform player that they run on outdated version. In both cases you want custom name and version.

> I am not a massive fan of `getSupportedProtocol` I dont think this whole PR should be used for versioning packets, the main goal is to check that the client is compatible. Versioning packets is better done by using `canSend` imo.

Can send is a just a boolean, which doesn't indicate packet version, just whatever that identifier is handled (which by itself is bit different concept, through it would require some extensions to be more useful with fabric's net api). It's pretty much similar to how network api has c:version packet, just more generic. I have similar system in polymer and I can say it's pretty useful.

> The code for comparing versions already exists in loader, we need to make this easy for most people to use in their mods. I really dont see people updating an int every time they break protocol, it gets even more complex when you have 2 or more branches for differing minecraft versions. These problems have all been solved with semver.

"Semver" doesn't really tell much about network compatibility, as Mojang themselves break it in minor versions, but also keeps it the same if they can. I can imagine modders doing things similarly to that. To be fair I myself don't use semver correctly, just have mods that use semver-like in format.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ModProtocolRegistry.java
Patbox: In case you wanted to use it as a `feature:modid/feature` you need unique id (and version, which might be just config state).

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/ServerModProtocolLookup.java
Patbox: Returning -1 is bit cleaner if someone wanted to switch(...) over it. Through OptionalInt makes sense too. I think making it throw would make it bit more annoying to use.

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/impl/modprotocol/ModProtocolImpl.java
Patbox: Tried to compact it, since readBoolean takes full byte (and extra boolean could be added easily with it without requiring protocol changes). Through it would make it cleaner

fabric-mod-protocol-api-v1/src/main/java/net/fabricmc/fabric/api/modprotocol/v1/package-info.java
Patbox: Yeah fair, through I imagine ideally it would never need incompatible format change, but can do it.

Fuzss:
Any update on this?

PR #4013, opened 2024-08-07
Crafter support for the transfer api.

### LINKED ISSUE #4004: Transfer api does not support the crafter
The crafter block uses a special method to move its result into an inventory and fabric api doesn't patch this method like it does with the dropper and hopper, causing transfer-api-only storages to not accept items from crafters. It might also be a good idea to check that the transfer api is able to correctly insert into crafters as it has special item spreading logic.

Orginally reported as MattiDragon/ExtendedDrawers#124

modmuss50:
Closes #4004

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/CrafterBlockMixin.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/CrafterBlockMixin.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/mixin/transfer/CrafterBlockMixin.java
```diff
@@ -0,0 +1,75 @@
...
+		if (itemStack.isEmpty()) {
+			// Nothing left to do, in theory should never get here.
+			return;
+		}
+
+		final Storage<ItemVariant> target = ItemStorage.SIDED.find(world, pos.offset(direction), direction.getOpposite());
+
+		if (target != null) {
+			// Attempt to move the entire stack, and decrement the size of success moves.
+			long moved = StorageUtil.move(
```
Technici4n: Why not `insert` on the target directly?

modmuss50: Good question, thats what I get for copy/pasting from the hopper without thinking. Updated it now, its much nicer :)

Technici4n (approved):

PR #4033, opened 2024-08-15
Port to 24w33a

modmuss50:
### TODO:
- [x] Get ingame, waiting on files added by @haykam821
- [x] fabric-data-generation-api-v1
- [x] fabric-biome-api-v1
- [ ] fabric-renderer-api-v1 / fabric-renderer-indigo
- [x] fabric-resource-conditions-api-v1
- [x] fabric-data-attachment-api-v1
- [x] fabric-client-tags-api-v1/fabric-convention-tags-v2
- [x] //TODO's (search for 1.21.2 and 24w33a)
- [ ] New fuel api, likely an event
- [ ] Validate rendering APIs
- [x] Check creative inventory buttons
- [x] Figure out what to do with FabricBlockEntityTypeBuilder, is it needed or is a TAW ok?
- [x] Bump loader version in all FMJs to 0.16.2, maybe do the mixin java version at the same time. (Do last as it will make the diff a bit of a mess)

Note: Not everything needs to be done for an inital release, we can release without all modules and missing APIs.

### Major breaking changes:
- Removed FuelRegistry
- Many entity rendering APIs now pass the new render state as opposed to the entity
- Removed ResourceReloadListenerKeys.TAGS
- UseItemCallback now returns a `ActionResult` instead of a `TypedActionResult<ItemStack>`
- Removed injected interface FabricBlockEntityType.Builder as the target class was removed.
- Biome carver related APIs no longer take a `GenerationStep.Carver`

### Post release tasks:
- Revive https://github.com/FabricMC/fabric/pull/3992


IMS212:
putting this here as note; BakedQuad now contains integer light emission, might want to replace `emissive` in material with that.

PepperCode1:
My current plan to support `BakedQuad` emissivity is as follows:
- Add an overload for `MutableQuadView#fromVanilla` that also accepts a `boolean`; this boolean says whether to respect the `BakedQuad`'s light. The default value is `true`.
- `MutableQuadView#fromVanilla` will apply the `BakedQuad`'s light by setting all per-vertex lightmap values to the quad light.
- `VanillaModelEncoder` will use the overload. `emitBlockQuads` will pass `true` while `emitItemQuads` will pass `false`.
- `QuadView#toBakedQuad` will set the resulting `BakedQuad`'s light to the minimum of all per-vertex lightmap values.

IMS212:
Just to clarify; `toBakedQuad` should use the minimum, or maximum? I think maximum would make more sense...

KedisPL:
By the way, modded blocks crash when they try to render in the world.

modmuss50:
> By the way, modded blocks crash when they try to render in the world.

Ah, yes I can reproduce, ill look into this. Thanks.

modmuss50:
Should now be good for release, plenty still to do after this first release but at least most of it is there.

PR #4037, opened 2024-08-17
Some more porting work on 24w33a

hypherionmc:
The following modules are still HARD broken

- fabric-recipe-api-v1

Other modules compile, but may have missing code. I have marked those with TODOs. I also marked code where I am not 100% sure that it is the correct change

~~It ran on my machine, once....~~


modmuss50:
Thanks, I will merge this right away and will work through the todos

PR #4045, opened 2024-08-20
Fix operator item group event

### LINKED ISSUE #4044: OP tab still showing when stacks have been added, even though it's disabled in options
I'm adding item stacks to the OP tab like this:
``` java
ItemGroupEvents.modifyEntriesEvent(ItemGroups.OPERATOR).register(entries -> {
  entries.add(stack);
});
```

However when I do this and the player has the Operator Items Tab set to OFF, the tab will still show, though only with the item stacks I've added and not the ones from vanilla mc. On top of that these buttons for switching pages shows up that do nothing (see image). My expectation is that the tab won't show at all when it's disabled in the options.

image

AshyBoxy:
This fixes #4044


modmuss50 (approved):
I think this is reasonable, previously it left it in the modders hand but there was little reason to do that and as you found broke the page count.

PR #4051, opened 2024-08-23
After Damage Event

TheDeathlyCow:
Adds a server-side only entity event after an entity is damaged. The event is fired after damage is applied, and includes both damage dealt and damage taken. However, damage taken is a bit weird. Its not the final amount of health reduction, but the damage the entity receives *just* armour / enchantments are applied and *after* shield, `freeze_hurts_extra_types`, and helmet damage reduction. The event is not fired if the damage killed the entity. 

I'm not sure if the difference between damage dealt and damage taken is clear enough that they should be separate here. Maybe it would be best to only supply the damage taken? These two numbers are both supplied to the `player_hurt_entity` and `entity_hurt_player` advancement criteria, so I thought their inclusion here made sense. However, as this is a fairly easy change to make I thought I would just open this now to see what people think.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
@@ -125,7 +125,7 @@ public final class ServerLivingEntityEvents {
 
 		/**
 		 * Called after a living entity took damage, unless they were killed. The damage taken is given as damage taken
-		 * before armor or enchantments are applied.
+		 * before armor or enchantments are applied, but after other effects like shields are applied.
 		 *
 		 * @param entity the entity that was damaged
 		 * @param source the source of the damage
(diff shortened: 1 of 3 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerLivingEntityEvents.java
```diff
@@ -42,6 +42,22 @@ public final class ServerLivingEntityEvents {
 		return true;
 	});
 
+	/**
+	 * An event that is called after an entity is damaged. This is fired from {@link LivingEntity#damage} after damage
+	 * is applied, or after that damage was blocked by a shield.
+	 *
+	 * <p>Damage dealt is the damage initially applied to the entity. Damage taken is the amount of damage the entity
+	 * actually took, after effects such as shields and extra freezing damage are applied. Damage taken does NOT include
```
modmuss50: I wonder if it should be renamed to `baseDamageTaken` or something else to make it clear its not the final amount of damage?

PR looks good otherwise 👍

TheDeathlyCow: sounds good, pushed change 👍

modmuss50 (approved):

PR #4052, opened 2024-08-23
Print the full stack trace from the dedicated server watchdog 

TelepathicGrunt:
Turns out 1.21.2 broke my FullStack Watchdog mod but instead of merging someone's PR to port my mod to that version, I figured maybe it'll be better to have it in Fabric API itself.

This code simply swaps the ThreadInfo being passed to StringBuilder with a full stacktrace string instead. The reason for this is ThreadInfo's toString method will truncate the stacktrace for itself. We need the full dump in order to see problems so doing our own copy of that toString method but without the truncating should do the trick

### DIFF
--- a/fabric-crash-report-info-v1/src/main/java/net/fabricmc/fabric/impl/crash/report/info/ThreadPrinting.java
+++ b/fabric-crash-report-info-v1/src/main/java/net/fabricmc/fabric/impl/crash/report/info/ThreadPrinting.java
@@ -21,6 +21,9 @@ import java.lang.management.MonitorInfo;
 import java.lang.management.ThreadInfo;
 
 public class ThreadPrinting {
+	/**
+	 * A modified copy of {@link ThreadInfo#toString} without the MAX_FRAMES check.
+	 */
 	public static String fullThreadInfoToString(ThreadInfo threadInfo) {
 		StringBuilder sb = new StringBuilder("\"" + threadInfo.getThreadName() + "\""
 				+ (threadInfo.isDaemon() ? " daemon" : "")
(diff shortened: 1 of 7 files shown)


modmuss50:
Great idea 👍 Would it be possible to create a testmod with a way to trigger this for testing?

fabric-crash-report-info-v1/src/main/java/net/fabricmc/fabric/impl/crash/report/info/ThreadPrinting.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+package net.fabricmc.fabric.impl.crash.report.info;
+
+import java.lang.management.LockInfo;
+import java.lang.management.MonitorInfo;
+import java.lang.management.ThreadInfo;
+
+public class ThreadPrinting {
+	public static String fullThreadInfoToString(ThreadInfo threadInfo) {
```
modmuss50, line 27: Please add a comment saying this was copied from ThreadInfo.toString(), I do wonder if there is a better way that doesnt require copy/pasting almost 100 lines from java 🤔

TelepathicGrunt:
testmod added. Command is `/print_thread_dump_test_command`. Will work in single player or on server as it calls the DedicatedServerWatchdog.createCrashReport method directly to get the string to print.
image

fabric-crash-report-info-v1/src/main/java/net/fabricmc/fabric/impl/crash/report/info/ThreadPrinting.java
TelepathicGrunt: comment is added now with link. Don't see a way to reuse code due to the java code using a static final field for their constant

modmuss50 (approved):

PR #4070, opened 2024-09-05
[1.21.5] Implement builtin mod resource/data pack sorting

### LINKED ISSUE #3543: [Suggestion]: Allow no-code mods to specify that their resources must apply after another mod
My use case is I have some mod-compat datapack to add compatibility between my mods and other mods. Some of these datapacks require overriding datapack files within my original mod in order to work.

I was planning on turning these datapacks into datapacks-as-mods by adding just a fabric.mod.json file to them. No code what-so-ever. I can specify a dependency in that file but I do not see any way to specify that my datapack-as-mod should load its resources on top and override another mod.

Without this ordering being allowed, I cannot make these datapack-as-mods projects because it'll become random of whether the resources properly overrides or not and thus, the datapack mod could work some times and not work other times. Not ideal.

Apollounknowndev:
> Supersedes #3913, targets 1.21.2 and changed how sorting works.

Overriding data/resource pack files from other mods is currently quite difficult as mod loading order isn't consistent. This can be especially annoying for compatibility mods that exist to override data/resource files from other mods.

This PR addresses the issue by implementing a system for mods to define in their fabric.mod.json what their priority is in the internal Fabric Mods data/resource pack relative to other mods. For example, a pack that overrides trees from a biome mod can set itself to load above the biome mod's datapack internally to make sure files are consistently overriden.

Mods that wish to change their priority relative to other in the load order can add this to their fabric.mod.json file:

```json
"custom": {
    "fabric:resource_load_order": {
        "before": [
          "mod_a",
          "mod_c",
        ],
        "after": [
          "mod_d"
        ]
    }
}
```

Mods in the `before` list will load their resources before/under the mod, mods in the `after` list will load their resources after/above the mod.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
@@ -21,6 +21,7 @@ import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
@@ -57,6 +58,7 @@ import net.fabricmc.fabric.api.resource.ModResourcePack;
 import net.fabricmc.fabric.api.resource.ResourcePackActivationType;
 import net.fabricmc.loader.api.FabricLoader;
 import net.fabricmc.loader.api.ModContainer;
+import net.fabricmc.loader.api.metadata.CustomValue;
 import net.fabricmc.loader.api.metadata.ModMetadata;
 
 /**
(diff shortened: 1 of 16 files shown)


modmuss50:
I think this is fine, I havent looked at it in depth, but the idea seems solid. I think we discussed this before, but are we happy to have this in the fabric.mod.json as opposed to the pack.mcmeta file? I think im happy either way.

This clearly needs some tests, if you want some help setting up multiple testmods for one module let me know, it should be tottaly possible to do.

Apollounknowndev:
It would be possible to move this to the `pack.mcmeta` file, but it'd require refactoring how mod packs are registered in an awkward way (Mod resource packs would need to have their profiles built, *then* sorted, *then* registered) and requires an additional accessor to get the resource pack metadata so I didn't opt for it. If there's an advantage to moving it to `pack.mcmeta` that makes it worth the hassle lmk.
I've tried to setup test mods but I'm not sure how to do the gradle wizardry, some help on that would be much appreciated.

modmuss50:
I have added 3 blank test mods, hopefully this will help you make some tests out of them. I also made some small code cleanup/changes with unit testing in mind, but ran out of time. Feel free to take a look at this, if not I might come back to it.

Apollounknowndev:
I've tried getting the testmod to load but it crashes with `NullPointerException: Cannot invoke "java.util.jar.Manifest.getMainAttributes()" because "manifest" is null` :P

modmuss50:
> I've tried getting the testmod to load but it crashes with `NullPointerException: Cannot invoke "java.util.jar.Manifest.getMainAttributes()" because "manifest" is null` :P

Sorry I missed this. Where are you seeing this problem?

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
```diff
@@ -64,29 +66,65 @@
...
+		return sorter.getPacks();
+	}
+
+	public static void addLoadOrdering(CustomValue.CvObject object, List<String> allIds, ModResourcePackSorter sorter, Order order, String currentId) {
+		List<String> modIds = new ArrayList<>();
+		CustomValue array = object.get(order.jsonKey);
 
-			if (pack != null) {
-				packs.add(pack);
+		if (array != null && array.getType() == CustomValue.CvType.ARRAY) {
```
Player3324: I think it should reject or at least warn about unsupported types. It might be nice to support a direct string value since that's what would happen for most mods using this feature (`"after": "mod_xy"`)

Apollounknowndev:
> Sorry I missed this. Where are you seeing this problem?

This occurs when trying to run `runTestmodClient` or `runTestmodServer`. The mod list shows `testmod-a` and `testmod-b` but doesn't get to `testmod-c` without crashing with a `Failed to remap mods!` error.

Apollounknowndev:
Finally have a few tests in using the three test mods.

modmuss50 (approved):

PR #4081, opened 2024-09-10
Implement client watchdog

apple502j:
Note: will refactor once #4052 is merged.

Client watchdog works similarly to the dedicated server watchdog - but for render thread. When the client ceases to tick for 30 seconds (or `fabric.clientWatchdog.maxTimeMs`), the game forcibly crashes. Useful for debugging infinite loops and the like.

This feature is currently enabled only in dev (or with `fabric.clientWatchdog.enabled`).

TODO:
- Maybe try to save the integrated server?
- This currently logs non-fatal error due to off-thread native access. Not sure if this is worth fixing. The error is not present in the watchdog crash report.

### DIFF
--- a/fabric-crash-report-info-v1/src/client/java/net/fabricmc/fabric/impl/client/crash/report/info/ClientWatchdog.java
+++ b/fabric-crash-report-info-v1/src/client/java/net/fabricmc/fabric/impl/client/crash/report/info/ClientWatchdog.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.client.crash.report.info;
+
+import java.nio.file.Path;
+import java.util.Locale;
+
+import com.mojang.logging.LogUtils;
+import org.slf4j.Logger;
+
+import net.minecraft.Bootstrap;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.server.dedicated.DedicatedServerWatchdog;
+import net.minecraft.util.Util;
+import net.minecraft.util.crash.CrashReport;
+import net.minecraft.util.crash.ReportType;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientLifecycleEvents;
+import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientTickEvents;
+import net.fabricmc.fabric.mixin.client.crash.report.info.MinecraftClientAccessor;
+import net.fabricmc.loader.api.FabricLoader;
+
+public class ClientWatchdog implements ClientModInitializer {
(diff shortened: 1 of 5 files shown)


fabric-crash-report-info-v1/src/client/java/net/fabricmc/fabric/impl/client/crash/report/info/ClientWatchdog.java
```diff
@@ -0,0 +1,86 @@
...
+		ClientLifecycleEvents.CLIENT_STARTED.register((client) -> {
+			Thread thread = new Thread(() -> run(client));
+			thread.setName("Fabric Client Watchdog");
+			thread.setDaemon(true);
+			thread.start();
+		});
+	}
+
+	public void run(MinecraftClient client) {
+		while (client.isRunning()) {
```
modmuss50, line 58: Should there not be a sleep in this while loop, there is no need to constantly check the value. See: DedicatedServerWatchdog

fabric-crash-report-info-v1/src/client/java/net/fabricmc/fabric/impl/client/crash/report/info/ClientWatchdog.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.util.crash.CrashReport;
+import net.minecraft.util.crash.ReportType;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientLifecycleEvents;
+import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientTickEvents;
+import net.fabricmc.fabric.mixin.client.crash.report.info.MinecraftClientAccessor;
+import net.fabricmc.loader.api.FabricLoader;
+
+public class ClientWatchdog implements ClientModInitializer {
```
modmuss50, line 38: really needs a testmod, id add a client command that sets a bool, and in the client ticket event just have a while (bool)

modmuss50 (commented):
im not sure how useful this is, but its easy enough. a few minor points.

fabric-crash-report-info-v1/src/client/java/net/fabricmc/fabric/impl/client/crash/report/info/ClientWatchdog.java
apple502j: Ah yeah, whoops.

apple502j:
> im not sure how useful this is

Well, a watchdog is useful in any part of code that might do infinite loop.

embeddedt:
IMO, a watchdog for the integrated server thread (in production) would be vastly more useful than one on the client thread. It's pretty rare that released mods deadlock the render thread. However, it's quite common for the integrated server to freeze with no explanation.

modmuss50:
I agree with @embeddedt however I think we also need to be careful to not trigger it too egarly, and maybe not crash to desktop? It could possibly open a UI allowing the user to ingore it?

With this PR as is, if you stall the main render thread will windows show the not responding UI?

apple502j:
Ah yeah, integrated server needs a watchdog as well. Will do that. I doubt it will have that much of a side effect if it only applies on dev/is opt-in.

If the server is not ticking for 30 seconds, I think it's fair to say it crashed. I don't want to overcomplicate things (remember this is FAPI).

> With this PR as is, if you stall the main render thread will windows show the not responding UI?

Need to check later. Without this PR it does show that UI.

modmuss50:
By UI I mean just a simple miecraft yes/no screen asking the user if they want to quit. I just worry someone is going to want to run a massive command that takes a long time to complete, and 30 seconds wont be enough.

embeddedt:
IMO the integrated server watchdog should *not* be enabled in dev, only in prod. At the very least, it should be disabled if a debugger is attached, otherwise pausing at a breakpoint for a while is enough to trip it, which makes it an inconvenience.

Player3324:
It might have some extra uses in-dev with a shorter delay to ungrab the mouse, especially on Linux.

I'd like to see less use of MC's utility methods where the JDK has suitable alternatives to reduce porting efforts.

PR #4082, opened 2024-09-10
Add ItemVariant#withComponentChanges and FluidVariant#withComponentChanges

BasiqueEvangelist:
Right now making a new transfer variant with some changed components is a bit cumbersome.
This PR adds a method to both ItemVariant and FluidVariant for easily making a copy of a variant with your component changes applied.

### DIFF
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
@@ -84,4 +84,11 @@ public interface FluidVariant extends TransferVariant<Fluid> {
 	default RegistryEntry<Fluid> getRegistryEntry() {
 		return getFluid().getRegistryEntry();
 	}
+
+	/**
+	 * Creates a copy of this FluidVariant with the provided component changes applied.
+	 * @param changes the changes to apply
+	 * @return the new variant with the changes applied
+	 */
+	FluidVariant withChanges(ComponentChanges changes);
 }
--- a/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
+++ b/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
@@ -90,5 +90,6 @@ public interface FluidVariant extends TransferVariant<Fluid> {
 	 * @param changes the changes to apply
 	 * @return the new variant with the changes applied
 	 */
+	@Override
 	FluidVariant withComponentChanges(ComponentChanges changes);
 }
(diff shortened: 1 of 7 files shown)


fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/api/transfer/v1/fluid/FluidVariant.java
```diff
@@ -84,4 +84,11 @@ default Fluid getFluid() {
 	default RegistryEntry<Fluid> getRegistryEntry() {
 		return getFluid().getRegistryEntry();
 	}
+
+	/**
+	 * Creates a copy of this FluidVariant with the provided component changes applied.
+	 * @param changes the changes to apply
+	 * @return the new variant with the changes applied
+	 */
+	FluidVariant withChanges(ComponentChanges changes);
```
Juuxel: `withComponentChanges` is a better name to clarify what we're changing IMO

BasiqueEvangelist: :+1:

Juuxel (approved):

Juuxel:
I wonder if this method should also be in TransferVariant as `TransferVariant<O> withComponentChanges(ComponentChanges)` with overridden specialisations in ItemVariant and FluidVariant since it makes sense for any TransferVariant 🤔

BasiqueEvangelist:
> I wonder if this method should also be in TransferVariant as `TransferVariant<O> withComponentChanges(ComponentChanges)` with overridden specialisations in ItemVariant and FluidVariant since it makes sense for any TransferVariant 🤔

i was thinking about doing something like that, but i dunno if that would be useful
like i guess it would make sense if you were doing something that was generic over item and fluid variants?
but is that that common though

i guess it doesn't really hurt that much to add it to TransferVariant though

Juuxel:
On second thought, adding it would be a breaking change since there's really no default impl you could provide for all variants, unless you made it throw. We are in the snapshot cycle though so I suppose it's possible to add the method to the superitf too if it's not ported to 1.21.1.

modmuss50:
> On second thought, adding it would be a breaking change since there's really no default impl you could provide for all variants, unless you made it throw. We are in the snapshot cycle though so I suppose it's possible to add the method to the superitf too if it's not ported to 1.21.1.

i think it makes sense to add it, and throw by default.

BasiqueEvangelist:
> On second thought, adding it would be a breaking change since there's really no default impl you could provide for all variants, unless you made it throw. We are in the snapshot cycle though so I suppose it's possible to add the method to the superitf too if it's not ported to 1.21.1.

Added. I haven't made it actually throw, since this PR is targetting 1.21.2 where compatibility breaks are expected anyways, but that could be added when this is backported to 1.21.1.

modmuss50:
> 1.21.2 where compatibility breaks are expected anyways

No, this is not how we work, this should very much try to maintain compatibility. The only time we make breaking changes is when we are forced to by Minecraft or there is a big benefit in doing so. Here I dont think there is.

BasiqueEvangelist:
> > 1.21.2 where compatibility breaks are expected anyways
> 
> No, this is not how we work, this should very much try to maintain compatibility. The only time we make breaking changes is when we are forced to by Minecraft or there is a big benefit in doing so. Here I dont think there is.

Right, I guess i'll make it throw then.

modmuss50 (approved):

PR #4085, opened 2024-09-11
Modify Enchantment and Fabric Component Map Builder Extensions

TheDeathlyCow:
This PR adds two things, that are fairly closely related so I felt their inclusion in a single PR was appropriate. 

First, this adds a new event that allows for external enchantments to be modified without the need to fully overwrite the enchantment in a data pack. This was motivated by TheDeathlyCow/scorchful#62 and TheDeathlyCow/scorchful#32 where I want to make changes to the Fire Protection and Impaling enchantment effects, but I did not want to fully overwrite them. 

~~I chose specifically to focus on the enchantment effects because attempting to modify the description or definition would just involve overwriting values anyway, and the exclusive set can be more easily modified with tags. However, the effects themselves are much easier to extend, as *most* of them are given as lists, which can of course easily be extended.~~
I switched to using `Enchantment.Builder` for a nicer to use API for adding effects that doesn't involve massive code duplication. This does come with the ability to modify the exclusive set, but I have indicated in the javadoc that this should be done through tags instead. 

Second, this adds a new extension interface to `ComponentMap.Builder` that enables retrieval of component values from the builder without requiring accessors. ~~This is very closely related to the modify enchantment effects event, particularly because the `getOrEmpty` method can be used to extend the effect lists as mentioned previously without overwriting.~~ Originally I made this change to make inter-compatible enchantment effects building easier, but after switching to the enchantment builder class this is no longer really needed for that purpose. However, this class may still be useful for the default component event, as many users may wish to modify or decorate the existing component values, and not fully overwrite them.

So, for example, you can add the Fire Aspect effect to Impaling like so:
```java
EnchantmentEvents.MODIFY.register(
	(key, builder, source) -> {
		if (source.isBuiltin() && key == WEIRD_IMPALING) {
			builder.addEffect(
				EnchantmentEffectComponentTypes.POST_ATTACK,
				EnchantmentEffectTarget.ATTACKER,
				EnchantmentEffectTarget.VICTIM,
				new IgniteEnchantmentEffect(EnchantmentLevelBasedValue.linear(4.0f)),
				DamageSourcePropertiesLootCondition.builder(
					DamageSourcePredicate.Builder.create().isDirect(true)
				)
			);
		}
	}
);
```

~~Currently, this is a bit cumbersome, so I may try to create a helper class that lets you define the effects in a way similar to `Enchantment.Builder`.~~ Done, I just switched to using `Enchantment.Builder` itself. 

~~I may also add something like the LootTableSource from the loot API, so that listeners can choose to only overwrite the enchantment if its a builtin enchantment and not touch datapacks that are meant to overwrite it.~~ Done. However, the similarity to the loot table source indicates to me that a potentially more general API (like a `ResourceSource` enum) could be needed. 

I'm not really sure how necessary `FabricComponentMapBuilder` is anymore now, since the only two item component types in vanilla that use lists are bees and recipes, so I'd like to know what are people's thoughts on whether it should be kept. Perhaps the ability to decorate non-list components is still enough of a use case to keep it? For example:

```java
DefaultItemComponentEvents.MODIFY.register(context -> {
	context.modify(Items.DIAMOND_PICKAXE, builder -> {
		Text baseName = builder.getOrCreate(
				DataComponentTypes.ITEM_NAME,
				Items.DIAMOND_PICKAXE::getName
		);
		MutableText modifiedName = Text.literal("Modified ")
				.append(baseName);

		builder.add(DataComponentTypes.ITEM_NAME, modifiedName);
	});
});
```


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RegistryLoaderMixin.java
```diff
@@ -0,0 +1,91 @@
...
+			MutableRegistry<E> registry,
+			Decoder<E> decoder,
+			RegistryOps<JsonElement> ops,
+			RegistryKey<E> key,
+			Resource resource,
+			RegistryEntryInfo entryInfo,
+			CallbackInfo ci
+	) {
+		if (key.isOf(RegistryKeys.ENCHANTMENT)) {
+			FABRIC_API$SOURCE.set(EnchantmentUtil.determineSource(resource));
```
modmuss50: Can you use @Share from mixin extras instead of a thread local?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EnchantmentBuilderAccessor.java
```diff
@@ -0,0 +1,43 @@
...
+
+import net.minecraft.component.ComponentMap;
+import net.minecraft.component.ComponentType;
+import net.minecraft.enchantment.Enchantment;
+import net.minecraft.registry.entry.RegistryEntryList;
+
+@Mixin(Enchantment.Builder.class)
+public interface EnchantmentBuilderAccessor {
+	@Accessor("definition")
+	Enchantment.Definition fabric_api$getDefinition();
```
modmuss50: Accessors dont need the modid prefix, this can just be `getDefinition()`

modmuss50 (approved):
This looks really good, not a lot to say. Just a few tweaks could be made to the mixins.

I really like the automated tests.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RegistryLoaderMixin.java
LlamaLad7: Why is this separate to the WrapOperation in the first place?

LlamaLad7: If it's just to get the enclosing method's params, you can do that from within a WrapOperation by simply adding them to the end of your handler params, as covered in the docs.

apple502j:
I do wonder if, at some point, generic dynamic registry modification API makes sense. We have 2 (biome & loot) APIs that already do similar things, and reducing code duplication is a good idea. However, it's also true many of the parts are likely non-reusable...

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RegistryLoaderMixin.java
```diff
@@ -0,0 +1,91 @@
...
+import net.fabricmc.fabric.api.item.v1.EnchantmentSource;
+import net.fabricmc.fabric.impl.item.EnchantmentUtil;
+
+@Mixin(RegistryLoader.class)
+abstract class RegistryLoaderMixin {
+	@Unique
+	private static final ThreadLocal<EnchantmentSource> FABRIC_API$SOURCE = ThreadLocal.withInitial(() -> EnchantmentSource.DATA_PACK);
+
+	@Inject(
+			method = "parseAndAdd",
```
maityyy: Is it possible to use `ServerWorldEvents.LOAD` here? I use this event to modify enchantments too

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RegistryLoaderMixin.java
TheDeathlyCow: Done, added it to the end of the method args so no `@Share` needed (poor person named share getting tagged here lol). Thanks for letting me know about this, mixin extras is still a little new to me :p

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/EnchantmentBuilderAccessor.java
TheDeathlyCow: Done

TheDeathlyCow:
@apple502j yeah I did think it was a bit weird to be targeting the generic registry loader class for just the specific case of enchantments. However I think the biggest hurdle with a generic registry modification API would be creating generic mutable builder objects that can be passed to the event. Such as in this case, creating a `Enchantment.Builder` object over `Enchantment`. 

Perhaps it could be done through direction modification of the JsonElement before it gets parsed into the final object? Instead of a Java-based event mods could define their modification patches in JSON with their mod datapack too? Though not sure how that would work with synchronized registries like Enchantment here.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/RegistryLoaderMixin.java
TheDeathlyCow: I don't think so. Otherwise, the enchantment source may not be retrievable without the `Resource` it came from (which `parseAndAdd` provides).

apple502j:
Supporting merge at this moment, however I expect this to be refactored (with possible breaking change) when we do https://github.com/FabricMC/fabric/pull/4085#issuecomment-2349832892

Not sure if I have time for that though. thus, 👍🏼

PR #4087, opened 2024-09-11
Add Dev Module

IThundxr:
Also adds properties for enabling certain things controlled by SharedConstants.isDevelopment so developers can pick and choose what they want to enable instead of everything being enabled


fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
+`Weight#validate`:
```
modmuss50: This should be documented in the code, that way it ends up in the javadoc and viewable in your mod dev env.

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric;
+
+public class FabricDev {
```
modmuss50: This class name is not descriptive

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
```diff
@@ -0,0 +1,43 @@
...
+	public static final boolean THROW_ON_INVALID_BLOCK_BOXES = getProperty("throwOnInvalidBlockBoxes", true);
+	public static final boolean ENABLE_UNPRIMED_HEIGHTMAP_LOGGING = getProperty("enableUnprimedHeightmapLogging", true);
+	public static final boolean ENABLE_LOADING_STRUCTURES_FROM_GAMETESTS = getProperty("enableLoadingStructuresFromGameTests", true);
+	public static final boolean THROW_ON_MISSING_DATA_FIXERS = getProperty("throwOnMissingDataFixers", false);
+	public static final boolean ENABLE_SUPPLIER_AND_RUNNABLE_DEBUGGING = getProperty("enableSupplierAndRunnableDebugging", false);
+	public static final boolean ENABLE_EXCEPTION_IDE_PAUSING = getProperty("enableExceptionIdePausing", true);
+
+	private static boolean getProperty(String name, boolean defaultValue) {
+		try {
+			return "true".equalsIgnoreCase(System.getProperty(name));
```
modmuss50: These should all be prefixed with `fabric.` at the very least, maybe even include the module name. Take a look at what other modules do.

fabric-api-dev/src/main/java/net/fabricmc/fabric/mixin/dev/ArgumentTypesMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+import net.minecraft.command.argument.ArgumentTypes;
+
+import net.fabricmc.fabric.FabricDev;
+
+@Mixin(ArgumentTypes.class)
+public class ArgumentTypesMixin {
+	@Dynamic("@ModifyExpressionValue's the FIELD GET of SharedConstants.isDevelopment to add a OR condition for FabricDev.REGISTER_TEST_ARGUMENTS")
```
modmuss50: @Dynamic doesnt add any useful infomation, I think its quite clear from the mixin what they do.

fabric-api-dev/src/main/java/net/fabricmc/fabric/mixin/dev/ArgumentTypesMixin.java
```diff
@@ -0,0 +1,35 @@
...
+	@ModifyExpressionValue(method = "register(Lnet/minecraft/registry/Registry;)Lnet/minecraft/command/argument/serialize/ArgumentSerializer;", at = @At(value = "FIELD", target = "Lnet/minecraft/SharedConstants;isDevelopment:Z"))
+	private static boolean fabric$mevIsDevelopmentForDevModule(boolean original) {
```
modmuss50: The mixins dont need to be prefixed with `fabric$` this happens automatically.

fabric-api-dev/src/main/resources/fabric-api-dev.mixins.json
```diff
@@ -0,0 +1,21 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin",
+  "compatibilityLevel": "JAVA_17",
+  "mixins": [
+    "dev.ArgumentTypesMixin",
```
modmuss50: Move `dev.` to the package.

fabric-api-dev/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,28 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-api-dev",
+  "name": "Fabric API Dev Utilities",
```
modmuss50: Im not too keen on this module name, although I dont really have a better suggestion. I dont think there is a reason to prevent people from enabling these options in prod.

I dont think it fits into any existing module.

gradle.properties
```diff
@@ -13,6 +13,7 @@ curseforge_minecraft_version=1.21.1
 
 # Do not manually update, use the bumpversions task:
 fabric-api-base-version=0.4.42
+fabric-api-dev-version=0.0.1
```
modmuss50: 1.0.0

modmuss50 (changes requested):
Please make sure it also builds.

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+`EulaReader#<init>` & `EulaReader#createEulaFile`:
+- Argument: `fabric.dev.alwaysAgreeToEula`
+- Default: false
+- Note: By enabling this, you declare that you have agreed to the EULA.
+- Skips creating the eula.txt file and always agrees to the EULA
+
+`MinecraftServer#tickWorlds`:
+- Argument: `fabric.dev.tickTestManager`
+- Default: false
+- Ticks TestManager.INSTANCE
```
apple502j: Already handled by GameTest API.

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+`CommandManager#<init>`:
+- Argument: `fabric.dev.registerDebugCommands`
+- Default: true
+- Register's Minecraft's debug commands (TestCommand, RaidCommand, DebugPathCommand, DebugMobSpawningCommand, WardenSpawnTrackerCommand, SpawnArmorTrimsCommand, ServerPackCommand) and if on the server DebugConfigCommand
```
apple502j: typo

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+`CommandManager#execute`:
+- Argument: `fabric.dev.enableCommandExceptionLogging`
+- Default: true
+- Logs an error if a command threw an exception
+
+`ArgumentTypes#register`:
+- Argument: `fabric.dev.registerTestArguments`
+- Default: false
+- Register's test_argument and test_class command arguments
```
apple502j: Already handled by GameTest API

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+`BlockBox#<init>`
+- Argument: `fabric.dev.throwOnInvalidBlockBoxes`
+- Default: true
+- Throw's an exception if a bounding box is invalid
+
+`Chunk#sampleHeightmap`:
+- Argument: `fabric.dev.enable_unprimed_heightmap_logging`
```
apple502j: camel or snake?

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+- Default: true
+- Logs an error if the heightmap is null
+
+`StructureTemplateManager#<init>`:
+- Argument: `fabric.dev.enableLoadingStructuresFromGameTests`
+- Default: true
+- Adds a provider to load structure templates from GameTest files
```
apple502j: Already handled by GameTest API.

fabric-api-dev/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+`Util#getChoiceTypeInternal`:
+- Argument: `fabric.dev.throwOnMissingDataFixers`
+- Default: false
+- Throw's an exception if a DataFixer isn't registered
```
apple502j: Remove this, as it never makes sense in a modded environment

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
```diff
@@ -0,0 +1,43 @@
...
+		} catch (Throwable e) {
+			return defaultValue;
```
apple502j: This is never called.

apple502j (changes requested):

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
IThundxr: > This is never called.

It'll be called if getProperty throws a NPE if the property doesn't exist

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
IThundxr: That was the intention originally, but when i redid the method it was forgotten.

fabric-api-dev/src/main/resources/fabric.mod.json
IThundxr: I'm kinda trying to describe it like, "This is a module that generally adds stuff only useful for developers"

fabric-api-dev/src/main/java/net/fabricmc/fabric/FabricDev.java
apple502j: Neither `getProperty` nor `equalsIgnoreCase` throw NPE.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+	 * Default value: true</br>
+	 * {@link net.minecraft.util.Util#error(String)}, {@link net.minecraft.util.Util#error(String, Throwable)}
+	 * and {@link net.minecraft.util.Util#throwOrPause(Throwable)}
+	 */
+	public static final boolean ENABLE_EXCEPTION_IDE_PAUSING = getProperty("enableExceptionIdePausing", true);
+
+	private static final boolean IS_DEVELOPMENT_ENV = FabricLoader.getInstance().isDevelopmentEnvironment();
+
+	private static boolean getProperty(String name, boolean defaultValue) {
+		String propertyValue = System.getProperty("fabric.dev" + name);
```
modmuss50: Boolean.getBoolean

fabric-api-dev/src/main/java/net/fabricmc/fabric/mixin/dev/BlockAndItemMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.fabricmc.fabric.impl.FabricDevProperties;
+
+@Mixin({Block.class, Item.class})
+public class BlockAndItemMixin {
+	@ModifyExpressionValue(method = {
+			"<init>(Lnet/minecraft/block/AbstractBlock$Settings;)V",
+			"<init>(Lnet/minecraft/item/Item$Settings;)V"
+	}, at = @At(value = "FIELD", target = "Lnet/minecraft/SharedConstants;isDevelopment:Z"))
+	private boolean mevIsDevelopmentForDevModule(boolean original) {
```
modmuss50: remove the `mev` prefix, the method name should explain what is being done here.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+	public static final boolean LOG_CONVENTION_ISSUES = getProperty("logConventionIssues", true);
+
+	/**
+	 * Skips creating the eula.txt file and always agrees to the EULA</br>
+	 * <strong>Note: By enabling this, you declare that you have agreed to the EULA.</strong></br>
+	 * Property: <code>fabric.dev.alwaysAgreeToEula</code></br>
+	 * Default value: false</br>
+	 * {@link net.minecraft.server.dedicated.EulaReader#EulaReader} and {@link EulaReader#createEulaFile()}
+	 */
+	public static final boolean ALWAYS_AGREE_TO_EULA = getProperty("alwaysAgreeToEula", false);
```
modmuss50: Please remove this, we should not be touching the EULA at all.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Invokes a method in which you should have a breakpoint to debug errors
+	 * thrown with Util#error and exceptions thrown with Util#throwOrPause</br>
+	 * Property: <code>fabric.dev.enableExceptionIdePausing</code></br>
+	 * Default value: true</br>
+	 * {@link net.minecraft.util.Util#error(String)}, {@link net.minecraft.util.Util#error(String, Throwable)}
+	 * and {@link net.minecraft.util.Util#throwOrPause(Throwable)}
+	 */
+	public static final boolean ENABLE_EXCEPTION_IDE_PAUSING = getProperty("enableExceptionIdePausing", true);
```
modmuss50: Why would you ever want this enabled? It should not be the default.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+
+	private static final boolean IS_DEVELOPMENT_ENV = FabricLoader.getInstance().isDevelopmentEnvironment();
+
+	private static boolean getProperty(String name, boolean defaultValue) {
```
modmuss50: Id just disable all of these by default, we want to make a dev env as close to production as possible, at least by default.

fabric-api-dev/src/main/resources/fabric.mod.json
modmuss50: Im not happy with this name, it doesnt explain what it does. I dont have a good suggestion otherwise though.

settings.gradle
```diff
@@ -14,6 +14,7 @@ include 'fabric-api-bom'
 include 'fabric-api-catalog'
 
 include 'fabric-api-base'
+include 'fabric-api-dev'
```
modmuss50, line 17: This should be set as a dev only module, and not included in the prod jar.

fabric-api-dev/src/main/resources/fabric-api-dev.mixins.json
```diff
@@ -0,0 +1,18 @@
+{
+  "required": true,
+  "package": "net.fabricmc.fabric.mixin.dev",
+  "compatibilityLevel": "JAVA_17",
```
modmuss50: JAVA_21

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+import net.minecraft.server.dedicated.EulaReader;
+import net.minecraft.util.collection.Weight;
+import net.minecraft.world.Heightmap;
+
+import net.fabricmc.loader.api.FabricLoader;
+
+@SuppressWarnings("JavadocReference")
+public class FabricDevProperties {
+	/**
+	 * Logs an error when a weight is set to zero</br>
```
modmuss50: FYI: These java docs are all in impl classes, so wont end up in the published javadocs. They should be documented elsewhere.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	public static final boolean LOG_MISSING_TRANSLATIONS = getProperty("logMissingTranslations", true);
+
+	/**
+	 * Logs an error if Block classes don't end with Block and if Item classes don't end with Item</br>
+	 * Property: <code>fabric.dev.logConventionIssues</code></br>
+	 * Default value: true</br>
+	 * {@link net.minecraft.block.Block#Block} and {@link net.minecraft.item.Item#Item}
+	 */
+	public static final boolean LOG_CONVENTION_ISSUES = getProperty("logConventionIssues", true);
```
modmuss50: Not a desriptive name of what this does.

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	public static final boolean REGISTER_DEBUG_COMMANDS = getProperty("registerDebugCommands", true);
+
+	/**
+	 * Logs an error if a command threw an exception</br>
+	 * Property: <code>fabric.dev.enableCommandExceptionLogging</code></br>
+	 * Default value: true</br>
+	 * {@link CommandManager#execute(ParseResults, String)}
+	 */
+	public static final boolean ENABLE_COMMAND_EXCEPTION_LOGGING = getProperty("enableCommandExceptionLogging", true);
```
modmuss50: I wonder if these should be named such as `fabric.dev.command.logging.exception` (with all of the others following the same pattern?

modmuss50 (changes requested):
None of these should be enabled by default imo.

settings.gradle
IThundxr: The reason it's included in prod is since some developers requested that they would be able to use it outside of dev envs to be debugging in prod easier

fabric-api-dev/src/main/java/net/fabricmc/fabric/impl/FabricDevProperties.java
IThundxr: It's usually helpful to have a breakpoint there when debugging, i don't see any reason to have it off by default

fabric-api-dev/src/main/java/net/fabricmc/fabric/api/dev/v1/FabricDevProperties.java
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * Invokes a method in which you should have a breakpoint to debug errors
+	 * thrown with Util#error and exceptions thrown with Util#throwOrPause</br>
+	 * Property: <code>fabric.dev.enableExceptionIdePausing</code></br>
+	 * Default value: true</br>
+	 * {@link net.minecraft.util.Util#error(String)}, {@link net.minecraft.util.Util#error(String, Throwable)}
+	 * and {@link net.minecraft.util.Util#throwOrPause(Throwable)}
+	 */
+	public static final boolean ENABLE_EXCEPTION_IDE_PAUSING = getProperty("enableExceptionIdePausing");
```
skycatminepokie, line 106: I didn't know this existed, that's cool.

modmuss50:
Hi, sorry for not taking a look at this PR in a while, please don't heasiate to ask me to look at PRs its impossible for me to check them all. If you need to ask me multiple times please do.

I worry that very few people will opt into these features, most of whitch are some additional logging that may not always be useful to someone other than Mojang. Warnings that we log we need to be 100% sure that they are important otherwise developers will just start ingoring them all. If its bad enough that it should never happen we should crash the game instead. 

We already enable the most useful features that are locked behind `SharedConstants.isDevelopment` by default in a number of related modules such as `/test` command, `/debugconfig` command, and some argument type debug commands.

The module name also doesnt feel right as well, I think it maybe should include `api` as it does technically have an API (the system props). I also strongly think this should be a dev only module, it can still be installed in prod if someone wishes to by downloading it from maven.

Before you spend anymore time on this lets discuss this and get input from others to see what they think, I think we need to strongly consider if this is in [scope](https://github.com/FabricMC/fabric/blob/1.21.4/CONTRIBUTING.md#scope), adding a new sub-module has a none zero peformance cost let alone the maintaince this module will require on updates (its unlikely people are going to search for new usages of `isDevelopment` after an update).

IThundxr:
> Hi, sorry for not taking a look at this PR in a while, please don't heasiate to ask me to look at PRs its impossible for me to check them all. If you need to ask me multiple times please do.
> 
> I worry that very few people will opt into these features, most of whitch are some additional logging that may not always be useful to someone other than Mojang. Warnings that we log we need to be 100% sure that they are important otherwise developers will just start ingoring them all. If its bad enough that it should never happen we should crash the game instead.
> 
> We already enable the most useful features that are locked behind `SharedConstants.isDevelopment` by default in a number of related modules such as `/test` command, `/debugconfig` command, and some argument type debug commands.
> 
> The module name also doesnt feel right as well, I think it maybe should include `api` as it does technically have an API (the system props). I also strongly think this should be a dev only module, it can still be installed in prod if someone wishes to by downloading it from maven.
> 
> Before you spend anymore time on this lets discuss this and get input from others to see what they think, I think we need to strongly consider if this is in [scope](https://github.com/FabricMC/fabric/blob/1.21.4/CONTRIBUTING.md?rgh-link-date=2025-01-14T21%3A12%3A16Z#scope), adding a new sub-module has a none zero peformance cost let alone the maintaince this module will require on updates (its unlikely people are going to search for new usages of `isDevelopment` after an update).

An alternative could be that we enable more things by default, this is gonna be a bit subjective, so it would require some conversations to see what exactly people feel should be enabled by default, but for a start i think the following are good:
- zeroWeightWarning
- logMissingTranslations
- enableCommandExceptionLogging
- enableCommandArgumentLogging
- enableExceptionIdePausing

modmuss50:
> An alternative could be that we enable more things by default, this is gonna be a bit subjective, so it would require some conversations to see what exactly people feel should be enabled by default, but for a start i think the following are good:

I dont think there is a perfect solution either way, some people want these additional checks enabled, others dont care. I think the better of the two evils is having them disabled by default. I do worry having too many warnings will make people ingore them all. Again this is likely out of scope but having a way for these warnings to crash the game might nice, this way a project such as Fabric API can ensure that the warnings never happen again.

> enableExceptionIdePausing
This one I dont think thats useful for anyone (even Mojang tbh), just let the game crash, and if you want to breakpoint on an exception set a breakpoint on the exception. Id argue we shouldnt even have an option for this.

Earthcomputer:
Some of these should absolutely be enabled by default, such as `enableCommandExceptionLogging`.

Additionally, there are some other things that could be useful in the dev module that Mojang doesn't have, like fixing the server watchdog to not crash after being stopped at a breakpoint.

modmuss50:
> Some of these should absolutely be enabled by default, such as enableCommandExceptionLogging.

This can be done in the command module.

> like fixing the server watchdog to not crash after being stopped at a breakpoint.

I think this was discussed, and we agreed it was a good idea but couldnt come to a solid way to do it. Can be done in the crash reports module.

IThundxr:
> > An alternative could be that we enable more things by default, this is gonna be a bit subjective, so it would require some conversations to see what exactly people feel should be enabled by default, but for a start i think the following are good:
> 
> 
> 
> I dont think there is a perfect solution either way, some people want these additional checks enabled, others dont care. I think the better of the two evils is having them disabled by default. I do worry having too many warnings will make people ingore them all. Again this is likely out of scope but having a way for these warnings to crash the game might nice, this way a project such as Fabric API can ensure that the warnings never happen again.
> 
> 
> 
> > enableExceptionIdePausing
> 
> This one I dont think thats useful for anyone (even Mojang tbh), just let the game crash, and if you want to breakpoint on an exception set a breakpoint on the exception. Id argue we shouldnt even have an option for this.

Making the warnings crash wouldn't be a bad idea

As for exception pausing it's useful as you can go down the stacktrace whenever a exception is thrown with it and see exactly what every variable or field is set to while the error was thrown

modmuss50:
> As for exception pausing it's useful as you can go down the stacktrace whenever a exception is thrown with it and see exactly what every variable or field is set to while the error was thrown

But you can do this by setting a breakpoint on the exception its self, without needing to place a breakpoint in a weird location.

IThundxr:
> > As for exception pausing it's useful as you can go down the stacktrace whenever a exception is thrown with it and see exactly what every variable or field is set to while the error was thrown
> 
> 
> 
> But you can do this by setting a breakpoint on the exception its self, without needing to place a breakpoint in a weird location.

Sure, but now your needing to place a breakpoint at multiple different exceptions rather than just a single breakpoint in the Util class

modmuss50:
Added by vanilla in a much better way than we ever could have done. Thanks for the PR.

PR #4088, opened 2024-09-12
Create `c:obsidians` block and item tag

TelepathicGrunt:
Back in the original unify c tag work, most tags were unified. One tag, `forge:obsidian`, was converted to `c:obsidians` in NeoForge but was not ported to Fabric. Now might be a good time to see if Fabric community is ok with this tag to collect all obsidian variants. 

Vanilla has normal Obsidian and Crying Obsidian so it makes sense mods could add even more variants. Such a tag might be useful in recipes where you want to accept any Obsidian as an ingredient.

Thoughts? Any objections to this tag?

Edit: 
Neo got a request to add obsidian subtags of `c:obsidians/normal` and `c:obsidians/crying`. Adding subtags is ideal for narrowing down and clarifying usages. In this case, recipes can target `c:obsidians/normal` and won't have to worry about other mod's special rare obsidians making it inside too.


Juuxel (approved):

modmuss50 (approved):
LGTM

modmuss50 (approved):

PR #4089, opened 2024-09-13
add AllowWorldChange, AllowTeleport and AfterTeleport

fishshi:
~~I implemented an API that allows for custom actions to be executed before and after a player teleports to inject custom code at the beginning and end of the teleport method. This approach provides more control and flexibility over teleport behavior, such as restricting teleportation for some case or triggering specific effects after teleportation.~~
Added three APIs:
AllowWorldChange: Called when the player's dimension changes to check if the player is allowed to cross dimensions.
AllowTeleport: Called before the player teleports to check if the player is allowed to teleport.
AfterTeleport: Called after the player teleports to execute custom logic.
The first API can be used for portals, while the other two can be applied to various teleport commands and mods that use teleport commands.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -18,7 +18,7 @@ package net.fabricmc.fabric.api.entity.event.v1;
 
 import net.minecraft.entity.damage.DamageSource;
 import net.minecraft.server.network.ServerPlayerEntity;
-
+import net.minecraft.server.world.ServerWorld;
 import net.minecraft.util.math.Vec3d;
 
 import net.fabricmc.fabric.api.event.Event;
@@ -67,9 +67,9 @@ public final class ServerPlayerEvents {
     /**
      * Called when a player is about to be teleported.
      */
-    public static final Event<AllowTeleport> ALLOW_TELEPORT = EventFactory.createArrayBacked(AllowTeleport.class, callbacks -> (player, pos) -> {
+    public static final Event<AllowTeleport> ALLOW_TELEPORT = EventFactory.createArrayBacked(AllowTeleport.class, callbacks -> (player, world, pos) -> {
 		for (AllowTeleport callback : callbacks) {
-			if (!callback.allowTeleport(player, pos)) {
+			if (!callback.allowTeleport(player, world, pos)) {
 				return false;
 			}
 		}
@@ -116,10 +116,11 @@ public final class ServerPlayerEvents {
 		 * Called when a player is about to be teleported.
          * 
          * @param player the teleporting player
+         * @param world the world to teleport to
          * @param pos the new position to teleport to
          * @return true if the teleport should go ahead, false otherwise.
 		 */
-		boolean allowTeleport(ServerPlayerEntity player, Vec3d pos);
+		boolean allowTeleport(ServerPlayerEntity player, ServerWorld world, Vec3d pos);
 	}
 
     @FunctionalInterface
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
@@ -19,6 +19,8 @@ package net.fabricmc.fabric.api.entity.event.v1;
 import net.minecraft.entity.damage.DamageSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 
+import net.minecraft.util.math.Vec3d;
+
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
@@ -61,6 +63,28 @@ public final class ServerPlayerEvents {
 
 		return true;
 	});
+    
+    /**
+     * Called when a player is about to be teleported.
+     */
+    public static final Event<AllowTeleport> ALLOW_TELEPORT = EventFactory.createArrayBacked(AllowTeleport.class, callbacks -> (player, pos) -> {
+		for (AllowTeleport callback : callbacks) {
+			if (!callback.allowTeleport(player, pos)) {
+				return false;
+			}
+		}
+
+		return true;
+	});
+
+    /**
+     * Called when a player has been teleported.
+     */
+    public static final Event<AfterTeleport> AFTER_TELEPORT = EventFactory.createArrayBacked(AfterTeleport.class, callbacks -> (player) -> {
+		for (AfterTeleport callback : callbacks) {
+			callback.afterTeleport(player);
+		}
+	});
 
 	@FunctionalInterface
 	public interface CopyFrom {
@@ -84,6 +108,28 @@ public final class ServerPlayerEvents {
 		 * @param alive whether the old player is still alive
 		 */
 		void afterRespawn(ServerPlayerEntity oldPlayer, ServerPlayerEntity newPlayer, boolean alive);
+	}
+
+    @FunctionalInterface
+	public interface AllowTeleport {
+		/**
+		 * Called when a player is about to be teleported.
+         * 
+         * @param player the teleporting player
+         * @param pos the new position to teleport to
+         * @return true if the teleport should go ahead, false otherwise.
+		 */
+		boolean allowTeleport(ServerPlayerEntity player, Vec3d pos);
+	}
+
+    @FunctionalInterface
+	public interface AfterTeleport {
+		/**
+		 * Called when a player has been teleported.
+         * 
+         * @param player the teleported player
+		 */
+		void afterTeleport(ServerPlayerEntity player);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
```diff
@@ -86,6 +110,28 @@ public interface AfterRespawn {
 		void afterRespawn(ServerPlayerEntity oldPlayer, ServerPlayerEntity newPlayer, boolean alive);
 	}
 
+    @FunctionalInterface
+	public interface AllowTeleport {
+		/**
+		 * Called when a player is about to be teleported.
+         * 
+         * @param player the teleporting player
+         * @param pos the new position to teleport to
```
modmuss50: I expect the target world may also be useful.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
```diff
@@ -90,6 +95,21 @@ private void onCopyFrom(ServerPlayerEntity oldPlayer, boolean alive, CallbackInf
 		ServerPlayerEvents.COPY_FROM.invoker().copyFromPlayer(oldPlayer, (ServerPlayerEntity) (Object) this, alive);
 	}
 
+    @Inject(method = "teleport(Lnet/minecraft/server/world/ServerWorld;DDDLjava/util/Set;FF)Z", at = @At("HEAD"), cancellable = true)
+    private void beforeTeleport(ServerWorld world, double destX, double destY, double destZ, Set<PositionFlag> flags, float yaw, float pitch, CallbackInfoReturnable<Boolean> cir) throws CommandSyntaxException {
+        ServerPlayerEntity player = (ServerPlayerEntity) (Object) this;
+        boolean allowed = ServerPlayerEvents.ALLOW_TELEPORT.invoker().allowTeleport(player, new Vec3d(destX, destY, destZ));
+        if (!allowed) {
+            throw new CommandSyntaxException(null, new LiteralMessage(""));
```
modmuss50: Question (I havent got the code in front of me), why is a CommandSyntaxException thrown? Does it need a message, and does this prevent the player from being teleported when they use a portal?

fabric-entity-events-v1/src/testmod/java/net/fabricmc/fabric/test/entity/event/EntityEventTests.java
```diff
@@ -80,6 +80,20 @@ public void onInitialize() {
 			LOGGER.info("Respawned {}, [{}, {}]", oldPlayer.getGameProfile().getName(), oldPlayer.getWorld().getRegistryKey().getValue(), newPlayer.getWorld().getRegistryKey().getValue());
 		});
 
+        ServerPlayerEvents.ALLOW_TELEPORT.register((player, pos) -> {
+            if(pos.getY() < -256)
```
modmuss50: Might be better to prevent the player from teleporting if they are holding a specific item, this way its easier to test portals.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
fishshi: I threw an exception because, whether the teleportation is successful or not, a teleportation message will always pop up when using coordinates or player-to-player teleportation. An empty exception can somewhat resolve this issue.

fabric-entity-events-v1/src/testmod/java/net/fabricmc/fabric/test/entity/event/EntityEventTests.java
fishshi: Using portals to travel across dimensions is different from teleporting with coordinates or between players, making it difficult to unify the behaviors. AllowTeleport and AfterTeleport is only suitable for the latter, whether using the tp command or a map mod.

fishshi:
I added AllowChangeWorld to handle portal events, which will be called before the player goes through the portal.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.java
fishshi: Yes, I added it. Thank you!

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
```diff
@@ -77,6 +84,16 @@ private void notifyDeath(DamageSource source, CallbackInfo ci) {
 		ServerLivingEntityEvents.AFTER_DEATH.invoker().afterDeath((ServerPlayerEntity) (Object) this, source);
 	}
 
+	@Inject(method = "teleportTo", at = @At(value = "FIELD", target = "Lnet/minecraft/server/network/ServerPlayerEntity;inTeleportationState:Z", opcode = Opcodes.PUTFIELD, shift = At.Shift.BEFORE), cancellable = true)
```
LlamaLad7: `Shift.BEFORE` should never be used. `Inject` already places its code before the target instruction, this places is one instruction back unnecessarily.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
```diff
@@ -77,6 +84,16 @@ private void notifyDeath(DamageSource source, CallbackInfo ci) {
...
+	private void beforeWorldChanged(TeleportTarget teleportTarget, CallbackInfoReturnable<Entity> cir) {
+		ServerPlayerEntity player = (ServerPlayerEntity) (Object) this;
+		boolean allowed = ServerEntityWorldChangeEvents.ALLOW_PLAYER_CHANGE_WORLD.invoker().allowChangeWorld(player, player.getServerWorld(), teleportTarget.world());
+		if (!allowed) {
+			cir.setReturnValue(null);
+			cir.cancel();
```
LlamaLad7: You don't need to call `cancel`, that's the first thing `setReturnValue` does.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
```diff
@@ -90,6 +107,21 @@ private void onCopyFrom(ServerPlayerEntity oldPlayer, boolean alive, CallbackInf
...
+	@Inject(method = "teleport(Lnet/minecraft/server/world/ServerWorld;DDDLjava/util/Set;FF)Z", at = @At("HEAD"), cancellable = true)
+	private void beforeTeleport(ServerWorld world, double destX, double destY, double destZ, Set<PositionFlag> flags, float yaw, float pitch, CallbackInfoReturnable<Boolean> cir) throws CommandSyntaxException {
+		ServerPlayerEntity player = (ServerPlayerEntity) (Object) this;
+		boolean allowed = ServerPlayerEvents.ALLOW_TELEPORT.invoker().allowTeleport(player, world, new Vec3d(destX, destY, destZ));
+		if (!allowed) {
+			throw new CommandSyntaxException(null, new LiteralMessage(""));
+		}
+	}
+
+	@Inject(method = "teleport(Lnet/minecraft/server/world/ServerWorld;DDDLjava/util/Set;FF)Z", at = @At("TAIL"), cancellable = true)
```
LlamaLad7: There's no need for these to be `cancellable`. There's also a potential issue with the before event being called and the after event not being called if someone cancels the method. Not sure how worried we are about that, and if it'd be worth requiring loader 0.16 to get <https://github.com/LlamaLad7/MixinExtras/wiki/WrapMethod> instead.

LlamaLad7 (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
fishshi: I see, thank you!

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
fishshi: I see, thank you!

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java
fishshi: I understand, thank you for pointing that out!

fishshi:
Thank you all for your guidance and feedback! I've made the necessary changes based on your suggestions. Additionally, I've also addressed the issues that occurred during the build.

fishshi:
I have discovered a bug and am working on fixing it. I will submit part of the PR later once I have made some progress.

Issue #3294, opened 2023-09-01
[Suggestion] Add an entity/player teleport event 


TelepathicGrunt:
Much like the teleporting event on Forge that fires for changeDimension (entity) or teleportTo (serverplayer), this allows mods to be able to block teleporting to a disallowed dimension. 

Example use case: https://www.curseforge.com/minecraft/mc-mods/restricted-portals

My mod has a custom teleporting system and I am able to call the Forge event to be compatible with the above mod much better. On Fabric, the lack of an equivalent event means I need to do direct mod compat by calling Restricted Portal's code in my mod.

Ideally for 1.19.2+

AndiLeni:
I'd like to see this feature as well.

maityyy:
Already implemented https://github.com/FabricMC/fabric/blob/efd7c830ae75ef1af97b3e94c33a02312663d0cb/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java#L31

PR #4104, opened 2024-09-20
add ALLOW_PLAYER_CHANGE_WORLD

fishshi:
ALLOW_PLAYER_CHANGE_WORLD: Called when the player's dimension changes to check if the player is allowed to cross dimensions.

solve issue:
https://github.com/FabricMC/fabric/issues/3294#issue-1877370481

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java
@@ -46,6 +46,19 @@ public final class ServerEntityWorldChangeEvents {
 	});
 
 	/**
+	 * Called before a player is being moved to a different world.
+	 */
+	public static final Event<AllowPlayerChange> ALLOW_PLAYER_CHANGE_WORLD = EventFactory.createArrayBacked(AllowPlayerChange.class, callbacks -> (player, origin, destination) -> {
+		for (AllowPlayerChange callback : callbacks) {
+			if (!callback.allowChangeWorld(player, origin, destination)){
+				return false;
+			}
+		}
+
+		return true;
+	});
+
+	/**
 	 * An event which is called after a player has been moved to a different world.
 	 *
 	 * <p>This is similar to {@link ServerEntityWorldChangeEvents#AFTER_ENTITY_CHANGE_WORLD} but is only called for players.
@@ -73,6 +86,19 @@ public final class ServerEntityWorldChangeEvents {
 		 * @param destination the destination world the new entity is in
 		 */
 		void afterChangeWorld(Entity originalEntity, Entity newEntity, ServerWorld origin, ServerWorld destination);
+	}
+
+	@FunctionalInterface
+	public interface AllowPlayerChange {
+		/**
+		 * Called before a player is being moved to a different world.
+		 *
+		 * @param player the player
+		 * @param origin the original world the player is in
+		 * @param destination the new world the player is moving to
+		 * @return true if the player is allowed to change worlds, false otherwise
+		 */
+		boolean allowChangeWorld(ServerPlayerEntity player, ServerWorld origin, ServerWorld destination);
 	}
 
 	@FunctionalInterface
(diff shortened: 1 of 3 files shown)


fabric-entity-events-v1/src/testmod/java/net/fabricmc/fabric/test/entity/event/EntityEventTests.java
```diff
@@ -64,6 +64,16 @@ public void onInitialize() {
 			LOGGER.info("Entity {} Killed: {}", entity, killed);
 		});
 
+		ServerEntityWorldChangeEvents.ALLOW_PLAYER_CHANGE_WORLD.register((player, origin, destination) -> {
+			if (player.getStackInHand(Hand.MAIN_HAND).getItem() == Items.END_ROD) {
```
haykam821: Nitpick:

```suggestion
			if (player.getMainHandStack().isOf(Items.END_ROD)) {
```

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java
```diff
@@ -45,6 +45,19 @@ public final class ServerEntityWorldChangeEvents {
 		}
 	});
 
+	/**
+	 * Called before a player is being moved to a different world.
+	 */
+	public static final Event<AllowPlayerChange> ALLOW_PLAYER_CHANGE_WORLD = EventFactory.createArrayBacked(AllowPlayerChange.class, callbacks -> (player, origin, destination) -> {
+		for (AllowPlayerChange callback : callbacks) {
+			if (!callback.allowChangeWorld(player, origin, destination)){
```
haykam821: Checkstyle will fail:

```suggestion
			if (!callback.allowChangeWorld(player, origin, destination)) {
```

haykam821 (changes requested):

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java
fishshi: thank you!

fishshi:
Sorry, but I accidentally merged the base branch into this branch. Will this have any impact, and how can I recover from it?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/ServerEntityWorldChangeEvents.java
```diff
@@ -75,6 +88,19 @@ public interface AfterEntityChange {
 		void afterChangeWorld(Entity originalEntity, Entity newEntity, ServerWorld origin, ServerWorld destination);
 	}
 
+	@FunctionalInterface
+	public interface AllowPlayerChange {
```
modmuss50, line 92: Should this be limited to just players, other entities are allowed to use portals.

I think its also worth looking at 1.21.2 as I know they changed a lot of this teleportation code.

fishshi: Yes, this event is currently only used to restrict players when crossing dimensions; other entities are not affected. I believe there are very few scenarios where restricting entities from crossing dimensions would be necessary.

fishshi: Additionally, I will read through 1.21.2 as soon as possible, and I may be able to provide feedback by tomorrow.

fishshi: I have reviewed the relevant source code for 24w39a, and there are not many changes regarding the dimension-crossing code. Through testing, I can assure you that this event works correctly in the latest version as well.

TelepathicGrunt:
Big concern with this PR. Say you block a teleport from a mod but that mod still sends the packets relating to teleporting. Won’t that cause issues and weird states in the game? For example, if you block the teleportTo in Bumblezone, my other packets will still be sent. https://github.com/TelepathicGrunt/Bumblezone/blob/2dc4c2034f1d9caf52cbe4c798754721b4895039/common/src/main/java/com/telepathicgrunt/the_bumblezone/entities/teleportation/BzWorldSavedData.java#L314

This is probably gonna cause major problems for the client and possibly other people as well

modmuss50:
Hi, thanks for the PR unfortunately I think doing this well requires quite a lot of work, teleportation is really complex. Some things that you will likely need to take into account:

- None player entities
- The context of how the entity is being teleported. E.g was it a command, portal, death, end credits or something another mod has initiated.
- There should be 3 events, allow, on and canceled.
- What state will the client and the entity being teleported be left in when the teleportation is blocked.

The first thing to do is to likely define a clear set of requirements/use cases, this likely requires looking at what existing mods do.

fishshi:
I'm really sorry, but I've decided to give up on this. Anyone else is welcome to continue trying.

PR #4109, opened 2024-09-21
AttachmentType Registration Enhancements

forgetmenot13579:
## Register `AttachmentType`s using the builder with consumer pattern.
Today, using the builder pattern creates a `Builder<Object>` by default, making both the `persistent` method and the `initializer` method largely useless, which is especially problematic as this is currently the only way to create an attachment with both persistence and a default value.
DataType
While this can be avoided by explicitly providing type parameters to the static `AttachmentRegistry#builder` method, type parameterizing methods is unintuitive and should be avoided as an API pattern whenever possible.

The solution implemented in this PR is the addition of the `AttachmentRegistry#create(Identifier id, Consumer<Builder<A>>)` method. This allows users to configure the (better type inferred) builder directly.
Registration


modmuss50:
> The motivation for this change is to avoid holding onto the mod ID in magic static strings as much as possible.

This does not seem like the solution to this problem, and barely changes anything as it doesnt solve it in every other place you create an identifier. Just create a static method that returns an identifier so you can do `MyMod.id("path")`. This is a none issue that this change doesnt solve.

forgetmenot13579:
I disagree that it is a nonissue, but it is true that solving it here specifically does not make much of a difference when the problem persists across the rest of the ecosystem. I have removed that feature and updated the PR summary accordingly.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -66,12 +79,9 @@ public static <A> AttachmentType<A> create(Identifier id) {
 	 * @return the registered {@link AttachmentType} instance
 	 */
 	public static <A> AttachmentType<A> createDefaulted(Identifier id, Supplier<A> initializer) {
-		Objects.requireNonNull(id, "identifier cannot be null");
-		Objects.requireNonNull(initializer, "initializer cannot be null");
-
-		return AttachmentRegistry.<A>builder()
-				.initializer(initializer)
-				.buildAndRegister(id);
+		return create(id, builder -> {
```
Syst3ms: nit: use expression lambda

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -83,10 +93,9 @@ public static <A> AttachmentType<A> createDefaulted(Identifier id, Supplier<A> i
 	 * @return the registered {@link AttachmentType} instance
 	 */
 	public static <A> AttachmentType<A> createPersistent(Identifier id, Codec<A> codec) {
-		Objects.requireNonNull(id, "identifier cannot be null");
-		Objects.requireNonNull(codec, "codec cannot be null");
-
-		return AttachmentRegistry.<A>builder().persistent(codec).buildAndRegister(id);
+		return create(id, builder -> {
```
Syst3ms: likewise, use expression lambda

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/WheelInfo.java
```diff
@@ -0,0 +1,39 @@
...
+	public static final Codec<WheelInfo> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+			Codecs.POSITIVE_FLOAT.fieldOf("wheelDiameter").forGetter(WheelInfo::wheelDiameter),
+			Codecs.POSITIVE_FLOAT.fieldOf("tireDiameter").forGetter(WheelInfo::tireDiameter),
+			Codecs.POSITIVE_FLOAT.fieldOf("tireThickness").forGetter(WheelInfo::tireThickness)
+	).apply(instance, WheelInfo::new));
+
+	public static final AttachmentType<WheelInfo> ATTACHMENT = AttachmentRegistry.create(Identifier.of(AttachmentTestMod.MOD_ID, "wheel_info"),
+			attachment -> attachment.initializer(() -> new WheelInfo(100, 5432, 37))
+					.persistent(WheelInfo.CODEC)
+	);
```
Syst3ms: nit: formatting looks weird here

fabric-data-attachment-api-v1/src/test/java/net/fabricmc/fabric/test/attachment/CommonAttachmentTests.java
```diff
@@ -33,6 +33,7 @@
 import java.util.function.UnaryOperator;
 
 import com.mojang.serialization.Codec;
+
```
Syst3ms: will this build?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
```diff
@@ -43,6 +43,21 @@ public final class AttachmentRegistry {
...
 	}
 
+	/**
+	 * Creates <i>and registers</i> an attachment, configuring the builder used underneath.
+	 *
+	 * @param id  the identifier of this attachment
+	 * @param <A> the type of attached data
+	 * @return the registered {@link AttachmentType} instance
+	 */
+	public static <A> AttachmentType<A> create(Identifier id, Consumer<Builder<A>> consumer) {
```
Syst3ms, line 54: Since it's intended as a replacement for `builder()`, this pattern should be explicitly documented in the javadoc above.

Syst3ms (changes requested):
Mostly nitpicks, this is a reasonable change. I think the entrypoint thing was well out of scope, though. This won't be too big of a hassle to merge into attachment sync.

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/WheelInfo.java
forgetmenot13579: The important thing is illustrating functionality. I'd be happy to switch the formatting if you prefer any one of these in particular.
```
public static final AttachmentType<WheelInfo> ATTACHMENT = AttachmentRegistry.create(Identifier.of(AttachmentTestMod.MOD_ID, "wheel_info"),
	attachment -> attachment
		.initializer(() -> new WheelInfo(100, 5432, 37))
		.persistent(WheelInfo.CODEC)
);
```
```
public static final AttachmentType<WheelInfo> ATTACHMENT = AttachmentRegistry.create(Identifier.of(AttachmentTestMod.MOD_ID, "wheel_info"),
	attachment -> attachment.initializer(() -> new WheelInfo(100, 5432, 37)).persistent(WheelInfo.CODEC)
);
```
```
public static final AttachmentType<WheelInfo> ATTACHMENT = AttachmentRegistry.create(Identifier.of(AttachmentTestMod.MOD_ID, "wheel_info"),
	attachment -> {
		attachment.initializer(() -> new WheelInfo(100, 5432, 37));
		attachment.persistent(WheelInfo.CODEC);
	}
);
```

Syst3ms: I was thinking either the first and/or putting the arguments on separate lines, as in
```suggestion
	public static final AttachmentType<WheelInfo> ATTACHMENT = AttachmentRegistry.create(
	    Identifier.of(AttachmentTestMod.MOD_ID, "wheel_info"),
	    attachment -> attachment.initializer(() -> new WheelInfo(100, 5432, 37)).persistent(WheelInfo.CODEC)
	);
```
or something like that. Take your pick, this is just a nitpick after all.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
forgetmenot13579: Since `builder()` doesn't work as intended anyway, and this *is* an experimental module, would removing or deprecating the existing `builder()` method make sense?

Syst3ms: Removing is definitely too much. Honestly, I knew of the issue already, I just thought passing the type parameters was alright. There's absolutely no need to make a breaking change, considering the existing method works, even if it's a bit clunky.

Deprecation sounds a bit more reasonable, but I think @modmuss50 knows the policy better on that front.

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/WheelInfo.java
forgetmenot13579: It's in the testmod. I'm just going to leave it as-is.

Syst3ms (approved):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentRegistry.java
modmuss50: Deprecating the old methods (without forRemoval) seems fine, there is no benefit to removing them.

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/WheelInfo.java
```diff
@@ -0,0 +1,40 @@
...
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.dynamic.Codecs;
+
+import net.fabricmc.fabric.api.attachment.v1.AttachmentRegistry;
+import net.fabricmc.fabric.api.attachment.v1.AttachmentType;
+
+public record WheelInfo(float wheelDiameter, float tireDiameter, float tireThickness) {
```
modmuss50: This class is never loaded, so we never actually run this test code. I would suggest replacing this with a unit test in CommonAttachmentTests if possible.

forgetmenot13579: This was mostly included to show the API usage, rather than as an actual test, but I have moved it to the test package so it does get registered. I've also updated the existing tests to use the new API.

Syst3ms (approved):

modmuss50 (approved):

PR #4112, opened 2024-09-22
Add a event for modify player's movement during using item

### LINKED ISSUE #4103: Hook for modifying player movement during item use
Having the ability to modify player movement for items currently being used would be useful, as the current `ClientPlayerEntity` injection point is quite annoying and fickle to target:

```java
public void tickMovement() {
	// etc

	if (this.isUsingItem() && !this.hasVehicle()) {
		this.input.movementSideways *= 0.2F;
		this.input.movementForward *= 0.2F;
		this.ticksLeftToDoubleTapSprint = 0;
	}

	// etc
}
```

For example, a mod may wish to adjust or disable the slowdown when a player is using the mod's custom bow.

fishshi:
Fixes #4103 
### Add a event for modify player's movement during using item.

The event ADJUST_USING_ITEM_SPEED is used to adjust a player's movement speed while using an item. It allows mods to modify the player's movement speed based on specific conditions when they are using a item.

When a player is moving while using an item, the movement speed can be modified based on various conditions. The return value is a float representing the speed adjustment percentage (e.g., 0.8 for 80% of the original speed). If it returns null, it means there is no speed adjustment. Among multiple speeds, the maximum value is taken.

The most common usecase is to modify using item speed based on main hand item, most for custom items. Another possible usecase is to modify using item speed when certain conditions (e.g. equipped items) are met.

### DIFF
--- a/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/client/entity/event/ClientPlayerEvents.java
+++ b/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/client/entity/event/ClientPlayerEvents.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.entity.event;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+public final class ClientPlayerEvents {
+	/*
+	 * An event that is called when a player is moving during using an item.
+	 */
+	public static final Event<ModifyPlayerMovementDuringUsingitem> MODIFY_PLAYER_MOVEMENT_DURING_USINGITEM = EventFactory.createArrayBacked(ModifyPlayerMovementDuringUsingitem.class, callbacks -> player -> {
+		for (ModifyPlayerMovementDuringUsingitem callback : callbacks) {
+			callback.modifyPlayerMovementDuringUsingitem(player);
+		}
+	});
+
+	@FunctionalInterface
+	public interface ModifyPlayerMovementDuringUsingitem {
+		/**
+		 * @param player the player is moving during using an item.
+		 */
+		void modifyPlayerMovementDuringUsingitem(ClientPlayerEntity player);
+	}
+}
--- a/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/client/entity/event/ClientPlayerEvents.java
+++ b/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/client/entity/event/ClientPlayerEvents.java
@@ -22,7 +22,7 @@ import net.fabricmc.fabric.api.event.EventFactory;
 import net.minecraft.client.network.ClientPlayerEntity;
 
 public final class ClientPlayerEvents {
-	/*
+	/**
 	 * An event that is called when a player is moving during using an item.
 	 */
 	public static final Event<ModifyPlayerMovementDuringUsingitem> MODIFY_PLAYER_MOVEMENT_DURING_USINGITEM = EventFactory.createArrayBacked(ModifyPlayerMovementDuringUsingitem.class, callbacks -> player -> {
@@ -34,6 +34,8 @@ public final class ClientPlayerEvents {
 	@FunctionalInterface
 	public interface ModifyPlayerMovementDuringUsingitem {
 		/**
+		 * Called when a player is moving during using an item.
+		 * 
 		 * @param player the player is moving during using an item.
 		 */
 		void modifyPlayerMovementDuringUsingitem(ClientPlayerEntity player);
--- a/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
+++ b/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
@@ -16,37 +16,29 @@
 
 package net.fabricmc.fabric.api.entity.event.client;
 
+import org.jetbrains.annotations.Nullable;
+
 import net.minecraft.client.network.ClientPlayerEntity;
 
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 public final class ClientPlayerEvents {
-	/*
-	 * Flag for using default slowdown during using an item.
-	 */
-	public static final float USE_DEFAULT_SLOWDOWN_SPEED = -1.0F;
-
-	/*
-	 * Default percentage of speed slowdown when using an item in Minecraft.
-	 */
-	private static final float DEFAULT_SLOWDOWN_SPEED = 0.2F;
-
 	/**
 	 * An event that is called when a player is moving during using an item.
 	 */
 	public static final Event<AdjustUsingItemSpeed> ADJUST_USING_ITEM_SPEED = EventFactory.createArrayBacked(AdjustUsingItemSpeed.class, callbacks -> player -> {
-		float maxSpeed = USE_DEFAULT_SLOWDOWN_SPEED;
+		Float maxSpeedPercentage = null;
 
 		for (AdjustUsingItemSpeed callback : callbacks) {
-			float currentSpeed = callback.adjustUsingItemSpeed(player);
+			Float speedPercentage = callback.adjustUsingItemSpeed(player);
 
-			if (currentSpeed >= 0.0F && currentSpeed <= 1.0F) {
-				maxSpeed = currentSpeed > maxSpeed ? currentSpeed : maxSpeed;
+			if (speedPercentage != null && speedPercentage >= 0.0F && speedPercentage <= 1.0F) {
+				maxSpeedPercentage = maxSpeedPercentage == null ? speedPercentage : Math.max(speedPercentage, maxSpeedPercentage);
 			}
 		}
 
-		return maxSpeed == USE_DEFAULT_SLOWDOWN_SPEED ? DEFAULT_SLOWDOWN_SPEED : maxSpeed;
+		return maxSpeedPercentage;
 	});
 
 	@FunctionalInterface
@@ -55,9 +47,10 @@ public final class ClientPlayerEvents {
 		 * Called when a player is moving during using an item.
 		 *
 		 * @param player the player is moving during using an item.
-		 * @return the percentage of the player's speed from 0.0F to 1.0F.
-		 * Return {@link #USE_DEFAULT_SLOWDOWN_SPEED} indicates that no adjustment should be applied.
+		 * @return the percentage of player's speed from 0.0F to 1.0F,
+		 * or {@code null} indicates that no adjustment should be applied.
 		 */
-		float adjustUsingItemSpeed(ClientPlayerEntity player);
+		@Nullable
+		Float adjustUsingItemSpeed(ClientPlayerEntity player);
 	}
 }
--- a/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
+++ b/fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
@@ -33,7 +33,7 @@ public final class ClientPlayerEvents {
 		for (AdjustUsingItemSpeed callback : callbacks) {
 			Float speedPercentage = callback.adjustUsingItemSpeed(player);
 
-			if (speedPercentage != null && speedPercentage >= 0.0F && speedPercentage <= 1.0F) {
+			if (speedPercentage != null) {
 				maxSpeedPercentage = maxSpeedPercentage == null ? speedPercentage : Math.max(speedPercentage, maxSpeedPercentage);
 			}
 		}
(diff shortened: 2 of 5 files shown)


fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/client/entity/event/ClientPlayerEvents.java
```diff
@@ -0,0 +1,43 @@
...
+package net.fabricmc.fabric.api.client.entity.event;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+public final class ClientPlayerEvents {
+	/**
+	 * An event that is called when a player is moving during using an item.
```
modmuss50: Question: Whats an example usecase?

fishshi: For example, a mod may wish to adjust or disable the slowdown when a player is using the mod's custom bow.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package net.fabricmc.fabric.api.entity.event.client;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ClientPlayerEvents {
```
modmuss50, line 26: Question: Does this also need to be handled on the server somewhere? E.g if I wanted to increase the players speed?

fishshi: I don't think it's necessary to make any changes on the server, as all information related to movement speed is processed on the client and sent to the server. We only need to modify the movement speed on the client. In the test mod, if you want to increase the player's speed, you only need to modify the multiplier coefficient. If you have any thoughts or additional comments, please let me know with more detailed information.

fabric-entity-events-v1/src/testmodClient/java/net/fabricmc/fabric/test/entity/event/client/EntityEventTestsClient.java
```diff
@@ -16,17 +16,34 @@
...
 			return !player.getEquippedStack(EquipmentSlot.CHEST).isOf(EntityEventTests.DIAMOND_ELYTRA);
 		});
+
+		ClientPlayerEvents.MODIFY_PLAYER_MOVEMENT_DURING_USINGITEM.register(player -> {
+			LOGGER.info("Player {} is moving during using item.", player);
+
+			if (player.getMainHandStack().isOf(Items.BOW) && player.getEquippedStack(EquipmentSlot.FEET).isOf(Items.DIAMOND_BOOTS)) {
+				LOGGER.info("Player {} can move without slowdown becase of diamond boots on feet.", player);
+				player.input.movementForward *= 5F;
+				player.input.movementSideways *= 5F;
```
TelepathicGrunt: So this PR, if you want to prevent slowdown, you multiply the move speed by 5? Uhhhhhhhhh

Have you considered what happens if two mods’s condition passes for trying to turn off the move speed? The two multiplying will end up causing the player to move 5 times as fast.

This PR is not suitable for multiple mods trying to turn off item use slowdown. It will need a rethinking

fishshi: Got it, I will try as soon as possible. Thank you!

fishshi: I changed the event to DISABLE_USINGITEM_SLOWDOWN. When the return value is true, the player can move without being slowed down, which will also resolve the issue of multiple mods affecting this simultaneously.

fishshi:
At the same time, I can implement adjusting the slowdown multiplier. However, now I'm unsure what should happen if multiple mods adjust the slowdown multiplier simultaneously, or if some mods adjust the multiplier while others disable the slowdown altogether. This seems to create a conflict in the settings between mods.

fishshi:
Description:
In event ADJUST_USING_ITEM_SPEED, I am using a float in the range [0, 1] as the percentage of movement speed, where 0 means movement is prohibited and 1 means normal speed with no slowdown. Any return values outside this range are considered invalid. When multiple mods are in effect, the maximum speed will be taken.
Additionally, I have defined a USE_DEFAULT_SLOWDOWN_SPEED flag for mods, indicating that no modifications are needed. This return value will not be compared with the return values of other mods, which helps mitigate conflicts. Mods only need to return their desired slowdown percentage in special cases; otherwise, they can return USE_DEFAULT_SLOWDOWN_SPEED.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * An event that is called when a player is moving during using an item.
+	 */
+	public static final Event<AdjustUsingItemSpeed> ADJUST_USING_ITEM_SPEED = EventFactory.createArrayBacked(AdjustUsingItemSpeed.class, callbacks -> player -> {
+		float maxSpeed = USE_DEFAULT_SLOWDOWN_SPEED;
+
+		for (AdjustUsingItemSpeed callback : callbacks) {
+			float currentSpeed = callback.adjustUsingItemSpeed(player);
+
+			if (currentSpeed >= 0.0F && currentSpeed <= 1.0F) {
```
Octol1ttle: why is the speed capped at 1.0?

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ClientPlayerEvents {
+	/*
+	 * Flag for using default slowdown during using an item.
+	 */
+	public static final float USE_DEFAULT_SLOWDOWN_SPEED = -1.0F;
```
Octol1ttle: change the return type of the interface to `@Nullable Float` instead of using this

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,63 @@
...
+	 * An event that is called when a player is moving during using an item.
+	 */
+	public static final Event<AdjustUsingItemSpeed> ADJUST_USING_ITEM_SPEED = EventFactory.createArrayBacked(AdjustUsingItemSpeed.class, callbacks -> player -> {
+		float maxSpeed = USE_DEFAULT_SLOWDOWN_SPEED;
+
+		for (AdjustUsingItemSpeed callback : callbacks) {
+			float currentSpeed = callback.adjustUsingItemSpeed(player);
+
+			if (currentSpeed >= 0.0F && currentSpeed <= 1.0F) {
+				maxSpeed = currentSpeed > maxSpeed ? currentSpeed : maxSpeed;
```
Octol1ttle: use `Math.max`

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/mixin/client/entity/event/ClientPlayerMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.entity.event.client.ClientPlayerEvents;
+
+@Mixin(ClientPlayerEntity.class)
+abstract class ClientPlayerMixin {
+	@Inject(method = "tickMovement", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;isUsingItem()Z"))
+	private void beforeTickMovement(CallbackInfo ci) {
```
Octol1ttle:
```suggestion
	private void invokeAdjustUsingItemSpeedEvent(CallbackInfo ci) {
```

Octol1ttle (changes requested):

Alexander01998:
What do you think about naming it `ITEM_USE` instead of `USING_ITEM`? It's a few characters shorter and (to me) it seems a bit easier to read.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
Alexander01998: On a similar note, why doesn't it accept speeds below zero? Is it just because of the `-1.0F` fallback or do negative values crash the game? If it's the former, consider allowing negative values after changing the fallback to `@Nullable Float`. If it's the latter, consider adding a comment to document why negative values would break stuff.

fishshi: Since this event is used to adjust the player's movement speed while using items, a value of 0 means the player's speed is already 0. A value of 1 means the player will maintain their original movement speed without any slowdown.
When the value is negative, it implies that the player will actually move backward when trying to move forward, which is counterintuitive. Therefore, negative values are not needed. 
When the value is greater than 1, it can be understood as the player moving faster while using an item. I'm currently questioning whether this is even necessary.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
fishshi: This is indeed a good idea. Thank you!

fishshi:
> What do you think about naming it `ITEM_USE` instead of `USING_ITEM`? It's a few characters shorter and (to me) it seems a bit easier to read.

Sorry, my English level is not very high, so I can't be sure which name is better. Can anyone else provide a bit more advice?

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
fishshi: Thank you! I’ve made the changes.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/mixin/client/entity/event/ClientPlayerMixin.java
fishshi: Thank you! I’ve made the changes.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
Alexander01998: Personally, I find it perfectly intuitive that negative speed would move you backwards. Do we really ever know that this functionality will not be necessary for any current or future mod? It just seems like an artificial limit to me.

If it requires extra effort to make these values work, that would be a different story. But given that it seems to just work without the limit, I think we might as well remove the limit and give modders the flexibility to use crazy values if they want to.

fishshi: Thank you! I understand, and I will remove this limit.

Earthcomputer:
I'm reading `MODIFY_USING_ITEM_SPEED` as "modify using-item speed" where "using-item" is a modifier to "speed". When using verbs to modify nouns like this, normally what you're actually using is the [gerund](https://www.scribbr.co.uk/nouns/gerunds/), i.e. "using" here is a noun, not a verb. The "item" here is the object of the action of "using", but (experts correct me if I'm wrong) gerunds cannot directly have objects in the same way as verbs. Instead we often use the possessive ("a using of an item" or "an item's using"), or use the object as a noun modifier like an adjective ("an item using"). Hence what I think is more grammatical here is "modify item-using speed". However as a native speaker this still feels weird to me, and I think it's because for the specific case of the verb "use", we have a better-suited noun "use" than the gerund "using". Which takes us to "modify item-use speed" ("use" here is a noun).

fishshi:
Thank you all! I have already renamed the event to MODIFY_ITEM_USE_SPEED.

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,56 @@
...
+	});
+
+	@FunctionalInterface
+	public interface ModifyItemUseSpeed {
+		/**
+		 * Called when a player is moving during using an item.
+		 *
+		 * @param player the player is moving during using an item.
+		 * @return a Float representing the speed modifyment as a percentage (e.g., 0.8 for 80% speed),
+		 * or {@code null} indicates that no modifyment should be applied.
```
Alexander01998:
```suggestion
		 * Called when a player is moving while using an item.
		 *
		 * @param player the player that is moving while using an item.
		 * @return a Float representing the speed modification as a percentage (e.g., 0.8 for 80% speed),
		 * or {@code null} indicates that no modification should be applied.
```

fabric-entity-events-v1/src/client/java/net/fabricmc/fabric/api/entity/event/client/ClientPlayerEvents.java
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.network.ClientPlayerEntity;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+public final class ClientPlayerEvents {
+	/**
+	 * An event that is called when a player is moving during using an item.
```
Alexander01998:
```suggestion
	 * An event that is called when a player is moving while using an item.
```

Earthcomputer:
From the discussion on Discord. My last message in favour of `MODIFY_ITEM_USE_SPEED` was based on the mistaken assumption that this event was about how fast you can use an item, rather than modifying the entity's movement who is using the item. In that case, I'd recommend `MODIFY_ITEM_USE_MOVEMENT_SPEED`, which is consistent with the vanilla `movement_speed` attribute. It's not perfect because it doesn't directly refer to the entity in question but I can't think of a way to do that without making it something like `MODIFY_ITEM_USE_ENTITY_MOVEMENT_SPEED` which IMO is just too long. At least "movement" distinguishes it from the speed of an item use.

Alexander01998 (approved):

Octol1ttle (approved):

PR #4113, opened 2024-09-23
Add server tick events for BlockEntity, Chunk, and Entity

forgetmenot13579:
Also adds tests for the new (and existing) chunk events.


modmuss50:
Is this something we really want? In the past we have said no as they are a very common cause of peformance issues.

forgetmenot13579:
The people who need this functionality are going to mixin anyway. By adding these events, we will have load, tick, and unload events for all four `AttachmentTarget` types and can look at implementing some kind of component system on top as previously discussed in the Discord server.

modmuss50:
A mixin is likely a better solution here, as it allows you to inject into the specific entity at the specific time that suits a modders need. One idea I had for these kinds of events would be registering them to the target when its loaded however this has its own issues.

On a large servers there are going to vast numbers of BE's, chunks and entities loaded at once, its going to be way too easy for someone to bring the server to a crawl.

PR #4115, opened 2024-09-24
Add component system

forgetmenot13579:
This is a draft of an entity component system built on top of the data attachments API and existing events. More events will need to be added for the currently registered `AttachmentTarget`s, but I wanted to get opinions on API structure first.


modmuss50:
Hi, thanks for the PR. Unfortunately I dont think this is something that fits within the scope of what we want in Fabric API. We already have the events, so you could always build something like this ontop of those in your own mods if needed. This seems too complicated for very little added benefit for most people. Im more than happy to discuss this further with you.

In the future I would suggest creating an issue or message in the discord to get some inital feedback on an idea before spending time developing it.

PR #4118, opened 2024-09-24
Add a level screen provider registry

### LINKED ISSUE #3926: Registration helpers for level screen providers should be added
The `LevelScreenProvider.WORLD_PRESET_TO_SCREEN_PROVIDER` field is initialized using `Map.of`; a helper similar to other `Map.of` helpers would be useful for mods adding their own world presets that should have a screen like superflat and single biome worlds do for customization.

An event could also be used so that mods can control multiple world presets at once.

There are two parts required for level screen providers: declaring that a button should be shown and creating the screen that is opened by activating the button.

haykam821:
This pull request adds a registry for level screen providers in the `fabric-screen-api-v1` module, which control the screens available from clicking 'Customize' on the create world screen. It can be used in the following way:

```java
Identifier id = Identifier.of("example", "fabricland");
RegistryKey<WorldPreset> key = RegistryKey.of(RegistryKeys.WORLD_PRESET, key);

LevelScreenProviderRegistry.register(FABRICLAND_WORLD_PRESET, (parent, generatorOptionsHolder) -> {
	ChunkGenerator chunkGenerator = generatorOptionsHolder.selectedDimensions().getChunkGenerator();
	// Return a screen here that calls `parent.getWorldCreator().applyModifier(...)` to apply changes
});
```

To test this registry, a Fabricland chunk generator and world preset has been added to the `fabric-screen-api-v1` test mod. The create world screen allows opening a customization screen which allows the background and outline blocks that are generated to be changed:

'World Type: Fabricland' button next to 'Customize' button
Fabricland made of granite and red glazed terracotta

The corresponding names in Yarn mappings may be improved, so the `LevelScreenProviderRegistry` class name in this API should not be considered final.

Note that the `LevelScreenProviderMixin` class modifies `Map.of` rather than replacing the field value as other registry API mixins do; the `Map.of` method is more brittle, but seemingly is required as `LevelScreenProvider` is an interface. See [this Discord message](https://discord.com/channels/507304429255393322/566276937035546624/1288241478187094127) for more information.

Fixes #3926

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/LevelScreenProviderRegistry.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/LevelScreenProviderRegistry.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.screen.v1;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import net.minecraft.client.gui.screen.world.LevelScreenProvider;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.world.gen.WorldPreset;
+
+import net.fabricmc.fabric.mixin.screen.LevelScreenProviderAccessor;
+
+/**
+ * Adds registration hooks for {@link LevelScreenProvider}s.
+ */
+public final class LevelScreenProviderRegistry {
+	private static final Logger LOGGER = LoggerFactory.getLogger(LevelScreenProviderRegistry.class);
+
+	private LevelScreenProviderRegistry() {
+	}
+
+	/**
+	 * Registers a provider for a screen that allows users to adjust the generation options for a given world preset.
+	 *
+	 * @param worldPreset the world preset to register the provider for
+	 * @param provider    the provider for the screen
+	 */
+	public static void register(RegistryKey<WorldPreset> worldPreset, LevelScreenProvider provider) {
+		Objects.requireNonNull(worldPreset, "world preset cannot be null");
+		Objects.requireNonNull(provider, "level screen provider cannot be null");
+
+		Optional<RegistryKey<WorldPreset>> key = Optional.of(worldPreset);
+		LevelScreenProvider old = LevelScreenProviderAccessor.fabric_getWorldPresetToScreenProvider().put(key, provider);
+
+		if (old != null) {
+			LOGGER.debug("Replaced old level screen provider mapping from {} to {} with {}", worldPreset, old, provider);
+		}
+	}
+}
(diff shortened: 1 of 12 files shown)


fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/LevelScreenProviderRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	public static void register(RegistryKey<WorldPreset> worldPreset, LevelScreenProvider provider) {
+		Objects.requireNonNull(worldPreset, "world preset cannot be null");
+		Objects.requireNonNull(provider, "level screen provider cannot be null");
+
+		Optional<RegistryKey<WorldPreset>> key = Optional.of(worldPreset);
+		LevelScreenProvider old = LevelScreenProviderAccessor.fabric_getWorldPresetToScreenProvider().put(key, provider);
+
+		if (old != null) {
+			LOGGER.debug("Replaced old level screen provider mapping from {} to {} with {}", worldPreset, old, provider);
```
modmuss50, line 54: Generally we crash if there is a duplicate, is there a specific usecase for overriding them?

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/LevelScreenProviderRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+import net.minecraft.client.gui.screen.world.LevelScreenProvider;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.world.gen.WorldPreset;
+
+import net.fabricmc.fabric.mixin.screen.LevelScreenProviderAccessor;
+
+/**
+ * Adds registration hooks for {@link LevelScreenProvider}s.
+ */
+public final class LevelScreenProviderRegistry {
```
modmuss50, line 34: We generally use interfaces for most public API like this, I would move the logic (with the logger) to an impl class.

fabric-screen-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/screen/ScreenTests.java
```diff
@@ -48,6 +60,12 @@ public void onInitializeClient() {
 		});
 
 		ScreenEvents.AFTER_INIT.register(this::afterInitScreen);
+
+		Registry.register(Registries.CHUNK_GENERATOR, FABRICLAND_ID, FabriclandChunkGenerator.CODEC);
```
modmuss50, line 64: I assume its fine to register this only on the client? In a real mod this wouldnt be like this, maybe just put a comment on it for people using it as a reference..

modmuss50 (commented):
Looks good, just a few minor changes.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/LevelScreenProviderRegistry.java
haykam821: The `FlattenableBlockRegistry`, `LandPathNodeTypesRegistry`, `SculkSensorFrequencyRegistry`, `StrippableBlockRegistry`, `VillagerInteractionRegistries`, `FabricDefaultAttributeRegistry`, `MinecartComparatorLogicRegistry`, and `VillagerTypeHelper` classes provide registration methods which log replaced values the same way.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/LevelScreenProviderMixin.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+
+import net.minecraft.client.gui.screen.world.LevelScreenProvider;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.world.gen.WorldPreset;
+
+@Mixin(LevelScreenProvider.class)
+public interface LevelScreenProviderMixin {
+	@WrapOperation(method = "<clinit>", at = @At(value = "INVOKE", target = "Ljava/util/Map;of(Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Ljava/util/Map;"))
+	private static Map<Optional<RegistryKey<WorldPreset>>, LevelScreenProvider> makeMutable(Object k1, Object v1, Object k2, Object v2, Operation<Map<Optional<RegistryKey<WorldPreset>>, LevelScreenProvider>> operation) {
```
LlamaLad7, line 35: ModifyExpressionValue

PR #4119, opened 2024-09-25
Add Hud Render Events

### LINKED ISSUE #3908: [1.21] Issue rendering via HudRenderCallback
Hi, I'm using HudRenderCallback event to render custom guis and things but I have noticed that when chat lines, messages overlay and action bar messages fade out it overlaps my custom gui. I have read the event notes and says:

> Called after rendering the whole hud, which is displayed in game, in a world.

If the method is called after the hud renders, it should not be rendered on top of everything else? Am I doing anything wrong?

On the other hand, I have noticed that inside InGameHud.class there is a new thing called LayeredDrawer and all hud methods like hotbar, stats bars are added creating, I think, a ordered way to render things without overlap anything, but I'm not sure how it works...

The possible solution I have found is to inject my code into one of these methods, in particular inside of 'renderMiscOverlays' and now it renders fine..

Rendering via HudRenderCallback:

Rendering via mixin InGameHud => renderMiscOverlays at head:

Thank you very much in advance :)

kevinthegreat1:
Originally discussed [here](https://discord.com/channels/507304429255393322/566276937035546624/1287326807061958708).  

Deprecate `HudRenderCallback`.  

Fixes #3908.


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,74 @@
...
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Events for rendering elements on the HUD.
+ *
+ * <p>These events will not be called if the HUD is hidden with F1.
+ */
+public class HudRenderEvents {
```
apple502j: final

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,74 @@
...
+	public static final Event<HudRenderStage> LAST = createEventForStage();
+
+	private static Event<HudRenderStage> createEventForStage() {
+		return EventFactory.createArrayBacked(HudRenderStage.class, listeners -> (context, tickDelta) -> {
+			for (HudRenderStage listener : listeners) {
+				listener.onRender(context, tickDelta);
+			}
+		});
+	}
+
```
apple502j: needs private ctor

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,74 @@
...
+	@FunctionalInterface
+	public interface HudRenderStage {
+		/**
+		 * Called sometime during a specific HUD render stage.
+		 *
+		 * @param context     The {@link DrawContext} instance
```
apple502j: Might make sense to pass client.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudRenderEventsTests.java
```diff
@@ -0,0 +1,54 @@
...
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.rendering.v1.HudRenderEvents;
+
+public class HudRenderEventsTests implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		// Render a blue rectangle at the top right of the screen, and it should be blocked by misc overlays such as vignette, spyglass, and powder snow
+		HudRenderEvents.START.register((context, tickCounter) -> {
+			context.fill(context.getScaledWindowWidth() - 200, 0, context.getScaledWindowWidth(), 30, 0xFF0000FF);
```
apple502j: Use `Colors` consts

apple502j (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Called after the entire HUD is rendered.
+	 */
+	public static final Event<HudRenderStage> LAST = createEventForStage();
+
+	private static Event<HudRenderStage> createEventForStage() {
+		return EventFactory.createArrayBacked(HudRenderStage.class, listeners -> (context, tickDelta) -> {
+			for (HudRenderStage listener : listeners) {
+				listener.onRender(context, tickDelta);
```
Octol1ttle:
```suggestion
		return EventFactory.createArrayBacked(HudRenderStage.class, listeners -> (context, tickCounter) -> {
			for (HudRenderStage listener : listeners) {
				listener.onRender(context, tickCounter);
```

Octol1ttle (changes requested):

Fuzss:
Not sure this really goes far enough. The system used by NeoForge has always been pretty ideal (although quite invasive): splitting gui rendering into all the different parts (health, hotbar, status effects, etc.) and allowing mods to render their custom gui layers anywhere in-between.

Also it would be awesome to finally get shared height parameters for bars (health bar + armor bar on the left side, food + air on the right side) rendered above the hotbar in Fabric Api. So that multiple mods adding their own bars know at what height to render without interfering with others.

kevinthegreat1:
> Not sure this really goes far enough. The system used by NeoForge has always been pretty ideal (although quite invasive): splitting gui rendering into all the different parts (health, hotbar, status effects, etc.) and allowing mods to render their custom gui layers anywhere in-between.

I agree that Forge's implementation (last when I saw it) is much more versatile and comprehensive. It would be great to be able to cancel certain components too. However as you mentioned it is quite invasive, and in the past, this injection point hasn't exactly been the most stable. A more comprehensive api would need a bit more discussion with the maintainers, but I'd be happy to see it.

> Also it would be awesome to finally get shared height parameters for bars (health bar + armor bar on the left side, food + air on the right side) rendered above the hotbar in Fabric Api. So that multiple mods adding their own bars know at what height to render without interfering with others.

The height problem is just reading the code to see the height where bars and components render right? Are you suggesting we provide z constants for different HUD elements?

Fuzss:
> The height problem is just reading the code to see the height where bars and components render right? Are you suggesting we provide z constants for different HUD elements?

No. NeoForge adds `Gui#leftHeight` and `Gui#rightHeight`. Those update automatically as vanilla bars are drawn above the hotbar. Mods drawing their own bars should update them as well. So e.g. if I have a mod that draws an armor toughness bar above the vanilla armor bar on the left hotbar side I know at what height to draw my bar at, and then another mod that also draws a bar at the left hotbar side can use the left height variable that has been updated by me to draw above me once more.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,78 @@
...
+			for (HudRenderStage listener : listeners) {
+				listener.onHudRender(client, context, tickCounter);
+			}
+		});
+	}
+
+	@FunctionalInterface
+	public interface HudRenderStage {
+		/**
+		 * Called sometime during a specific HUD render stage.
```
Octol1ttle:
```suggestion
		 * Called during a HUD render stage.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,138 @@
...
+		 * @param context     the {@link DrawContext} instance
+		 * @param tickCounter the {@link RenderTickCounter} instance with access to tick delta
+		 */
+		void afterMiscOverlays(MinecraftClient client, DrawContext context, RenderTickCounter tickCounter);
+	}
+
+	@FunctionalInterface
+	public interface AfterMainHud {
+		/**
+		 * Called after the hotbar, status bars, and experience bar have been rendered, and before the status effects overlays are rendered.
```
Octol1ttle: Is there a reason to calling this before status effect overlays & bossbar?
```suggestion
		 * Called after the hotbar, status bars, and experience bar have been rendered, and before the status effects overlays and boss bar are rendered.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,138 @@
...
+	public interface BeforeChat {
+		/**
+		 * Called after the debug HUD, scoreboard, overlay message (action bar), and title and subtitle have been rendered, and before the {@link net.minecraft.client.gui.hud.ChatHud} is rendered.
+		 *
+		 * @param client      the {@link MinecraftClient} instance
+		 * @param context     the {@link DrawContext} instance
+		 * @param tickCounter the {@link RenderTickCounter} instance with access to tick delta
+		 */
+		void beforeChat(MinecraftClient client, DrawContext context, RenderTickCounter tickCounter);
+	}
```
Octol1ttle: Maybe add `AFTER_SLEEP_OVERLAY` before this event? This would allow rendering before (so, behind) debug HUD, scoreboard, etc.

kevinthegreat1: You can just use `AFTER_MAIN_HUD`.

Octol1ttle: that's called before the sleep overlay though

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
kevinthegreat1: No not really. Just thought this was better.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
kevinthegreat1: Done in ce182ed.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
kevinthegreat1: Done in ce182ed.

Octol1ttle (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,159 @@
...
+		 *
+		 * @param client      the {@link MinecraftClient} instance
+		 * @param context     the {@link DrawContext} instance
+		 * @param tickCounter the {@link RenderTickCounter} instance with access to tick delta
+		 */
+		void onHudStart(MinecraftClient client, DrawContext context, RenderTickCounter tickCounter);
+	}
+
+	@FunctionalInterface
+	public interface AfterMiscOverlays {
```
modmuss50: Question: Are all of these interfaces the same? I do wonder if its really worth having 5 identical interfaces, this class could likely be quite a bit more consises.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,19 +16,72 @@
...
 import net.fabricmc.fabric.api.client.rendering.v1.HudRenderCallback;
+import net.fabricmc.fabric.api.client.rendering.v1.HudRenderEvents;
 
 @Mixin(InGameHud.class)
 public class InGameHudMixin {
+	@Shadow
+	@Final
+	private MinecraftClient client;
+
+	@ModifyArg(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/LayeredDrawer;addLayer(Lnet/minecraft/client/gui/LayeredDrawer$Layer;)Lnet/minecraft/client/gui/LayeredDrawer;", ordinal = 0))
```
modmuss50: Please add comments explaining where there mixins are targgeting, it makes it a lot easier to update.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Events for rendering elements on the HUD.
+ *
+ * <p>These events will not be called if the HUD is hidden with F1.
+ */
+public final class HudRenderEvents {
+	/**
+	 * Called at the start of HUD rendering, right before anything is rendered.
+	 */
+	public static final Event<Start> START = EventFactory.createArrayBacked(Start.class, listeners -> (client, context, tickCounter) -> {
```
modmuss50: Are events like this the best solution here? Would it be better to allow mods to register their own LayeredDrawer.Layer in a similar way to how the vanilla hud elements work?

modmuss50 (commented):
Im not sure events like this are the best way to do this. An API that allows you to register LayeredDrawer.Layer's in the desired order may be better?

modmuss50:
Please let me know if you want to discuss this more

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
kevinthegreat1: I originally had one interface but using multiple interfaces allows people to use one class to implement multiple events.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
kevinthegreat1: Done in 6362810.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
kevinthegreat1: I have updated the code to use one interface, but I don't see a difference with registration methods. Wouldn't it still be implemented with an event?

modmuss50: I was thinking they could just be static? I dont think there is a need for an even unless im missing something.

kevinthegreat1: I was under the impression that not using events is more complicated than just using events? So there's no need for registration methods? Are you saying just implement it with a synchronized list? Events also allow phase ordering if some mod wants to render before another.

Fuzss:
> Im not sure events like this are the best way to do this. An API that allows you to register LayeredDrawer.Layer's in the desired order may be better?

This please! It would be really great if mods were to properly register their GUI layers with an identifier that allows for ordering with callbacks firing before and after those layers render, as is done on NeoForge. That way other mods can interact with those new layers.
Like the Raised mod (for moving the hotbar up) could selectively move modded layers via translating the pose stack, or a minimap could move a custom effects hud out of the way and so on.

Ideally vanilla layers should also be accessible via an identifier, although not sure how to implement that in a way that’s not too invasive.

kevinthegreat1:
> > Im not sure events like this are the best way to do this. An API that allows you to register LayeredDrawer.Layer's in the desired order may be better?
> 
> This please! It would be really great if mods were to properly register their GUI layers with an identifier that allows for ordering with callbacks firing before and after those layers render, as is done on NeoForge. That way other mods can interact with those new layers. Like the Raised mod (for moving the hotbar up) could selectively move modded layers via translating the pose stack, or a minimap could move a custom effects hud out of the way and so on.
> 
> Ideally vanilla layers should also be accessible via an identifier, although not sure how to implement that in a way that’s not too invasive.

I had a discussion with him on discord [here](https://discord.com/channels/507304429255393322/566276937035546624/1294326846888087624), and I think you're misunderstanding his proposal. Under his idea, the API would still have the same capability as the events here, since what you're describing uses many more mixins and is significantly more invasive. I have looked at neo's API, and the number of patches required will probably make it not acceptible for fabric.

modmuss50:
Yes, as discussed on discord I think it would be better to statically register (or maybe register in events similar to the ones in the current state of the PR a vanilla `LayeredDrawer.Layer`.

kevinthegreat1:
> Yes, as discussed on discord I think it would be better to statically register (or maybe register in events similar to the ones in the current state of the PR a vanilla `LayeredDrawer.Layer`.

Done in 0d2bcf2.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,19 +16,78 @@
...
+	private LayeredDrawer.Layer fabric$beforeChat(LayeredDrawer.Layer beforeChatLayer) {
+		return (context, tickCounter) -> {
+			HudRenderEvents.BEFORE_CHAT.invoker().render(context, tickCounter);
+			beforeChatLayer.render(context, tickCounter);
+		};
+	}
+
+	// Targeting the last addLayer call of the second layered drawer, currently the subtitles hud layer (subtitlesHud.render) as of 1.21.
+	@ModifyArg(method = "<init>", slice = @Slice(from = @At(value = "NEW", target = "Lnet/minecraft/client/gui/LayeredDrawer;", ordinal = 2)), at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/LayeredDrawer;addLayer(Lnet/minecraft/client/gui/LayeredDrawer$Layer;)Lnet/minecraft/client/gui/LayeredDrawer;", ordinal = 7))
+	private LayeredDrawer.Layer fabric$AfterSubtitlesHud(LayeredDrawer.Layer subtitlesHudLayer) {
```
Octol1ttle:
```suggestion
	private LayeredDrawer.Layer fabric$afterSubtitlesHud(LayeredDrawer.Layer subtitlesHudLayer) {
```

Octol1ttle:
> > Yes, as discussed on discord I think it would be better to statically register (or maybe register in events similar to the ones in the current state of the PR a vanilla `LayeredDrawer.Layer`.
> 
> Done in [0d2bcf2](https://github.com/FabricMC/fabric/commit/0d2bcf291eda8c1b5353412e421323e09eebd2c2).

I *think* what modmuss meant here is that we would have a `HudRenderRegistry` (or `HudRenderRegistrationEvents`) where mods would put their `LayeredDrawer.Layer`s. Then, all these layers would be added to the drawer directly with `addLayer`

kevinthegreat1:
> I _think_ what modmuss meant here is that we would have a `HudRenderRegistry` (or `HudRenderRegistrationEvents`) where mods would put their `LayeredDrawer.Layer`s. Then, all these layers would be added to the drawer directly with `addLayer`

He said it’s fine to register in events. I used events because it allows ordering the layers within one phase/event.

Also, directly adding to the vanilla list would require similarly many injection points or rely on the indices which is even more undesirable.

Octol1ttle (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderEvents.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Called after the debug HUD, scoreboard, overlay message (action bar), and title and subtitle have been rendered, and before the {@link net.minecraft.client.gui.hud.ChatHud} is rendered.
+	 */
+	public static final Event<LayeredDrawer.Layer> BEFORE_CHAT = createEventForStage();
+
+	/**
+	 * Called after the entire HUD is rendered.
+	 */
+	public static final Event<LayeredDrawer.Layer> LAST = createEventForStage();
```
maityyy: Maybe END? There is START, not FIRST or smth

kevinthegreat1: The names are taken from `WorldRenderEvents`, which has both `LAST` and `END`, and `LAST` is the last place where you should render stuff in most cases. However, I'm happy to change this if modmuss agrees.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricLayeredDrawer.java
```diff
@@ -0,0 +1,48 @@
...
+
+	default FabricLayeredDrawer addLayerAfter(Identifier after, Identifier identifier, LayeredDrawer.Layer layer) {
+		return addLayerAfter(after, IdentifiedLayer.wrapping(identifier, layer));
+	}
+
+	FabricLayeredDrawer removeLayer(Identifier identifier);
+
+	FabricLayeredDrawer replaceLayer(Identifier identifier, Function<IdentifiedLayer, IdentifiedLayer> replacer);
+
+	FabricLayeredDrawer addSubDrawer(Identifier identifier, LayeredDrawer drawer, BooleanSupplier shouldRender);
```
modmuss50: I know I added this, but im not too sure its really needed, I dont think there is any benefit to having it.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricLayeredDrawer.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.function.BooleanSupplier;
+import java.util.function.Function;
+
+import net.minecraft.client.gui.LayeredDrawer;
+import net.minecraft.util.Identifier;
+
+/**
+ * A layered drawer that has an identifier attached to each layer and methods to add layers in specific positions.
+ */
+public interface FabricLayeredDrawer {
```
modmuss50: I think this should be renamed, its not a `LayeredDrawer` at all, its a facade over the vanilla one to hide the complexity. `LayeredDrawerContext`, `LayeredDrawerManager`, `LayeredDrawerModifications` Im not really too sure of a good name, I wonder if anyone else has a good suggestion.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudLayerRegistrationCallback.java
```diff
@@ -0,0 +1,42 @@
...
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Callback for when hud layers are registered.
+ *
+ * <p>To register a layer, register a listener to this event and register your layers in the listener.
+ *
+ * @see FabricLayeredDrawer
+ */
+public interface HudLayerRegistrationCallback {
```
modmuss50, line 48: Yes, this is exactally what I had in mind. I would maybe add a code example to the javadocs.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
```diff
@@ -0,0 +1,48 @@
...
+	Identifier DEMO_TIMER = Identifier.ofVanilla("demo_timer");
+	Identifier DEBUG_HUD = Identifier.ofVanilla("debug_hud");
+	Identifier SCOREBOARD = Identifier.ofVanilla("scoreboard");
+	Identifier OVERLAY_MESSAGE = Identifier.ofVanilla("overlay_message");
+	Identifier TITLE_AND_SUBTITLE = Identifier.ofVanilla("title_and_subtitle");
+	Identifier CHAT = Identifier.ofVanilla("chat");
+	Identifier PLAYER_LIST = Identifier.ofVanilla("player_list");
+	Identifier SUBTITLES = Identifier.ofVanilla("subtitles");
+	Identifier SLEEP = Identifier.ofVanilla("sleep");
+	Identifier SUB_DRAWER_1 = Identifier.ofVanilla("sub_drawer_1");
```
modmuss50: No, I dont think we should expose the names of these as API, the sub drawers are impl details im not sure mods need to care about?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
```diff
@@ -0,0 +1,48 @@
...
+
+public interface IdentifiedLayer extends LayeredDrawer.Layer {
+	Identifier MISC_OVERLAYS = Identifier.ofVanilla("misc_overlays");
+	Identifier CROSSHAIR = Identifier.ofVanilla("crosshair");
+	Identifier MAIN_HUD = Identifier.ofVanilla("main_hud");
+	Identifier EXPERIENCE_LEVEL = Identifier.ofVanilla("experience_level");
+	Identifier STATUS_EFFECTS = Identifier.ofVanilla("status_effects");
+	Identifier BOSSBAR = Identifier.ofVanilla("bossbar");
+	Identifier DEMO_TIMER = Identifier.ofVanilla("demo_timer");
+	Identifier DEBUG_HUD = Identifier.ofVanilla("debug_hud");
```
modmuss50: Maybe remove `hud` from the names? Might be worth looking over these names in general to make sure they are consitent.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
```diff
@@ -0,0 +1,48 @@
...
+	Identifier CHAT = Identifier.ofVanilla("chat");
+	Identifier PLAYER_LIST = Identifier.ofVanilla("player_list");
+	Identifier SUBTITLES = Identifier.ofVanilla("subtitles");
+	Identifier SLEEP = Identifier.ofVanilla("sleep");
+	Identifier SUB_DRAWER_1 = Identifier.ofVanilla("sub_drawer_1");
+	Identifier SUB_DRAWER_2 = Identifier.ofVanilla("sub_drawer_2");
+
+	Identifier id();
+
+	static IdentifiedLayer wrapping(Identifier id, LayeredDrawer.Layer layer) {
```
modmuss50: `of` might be a better name.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,21 +16,198 @@
...
+
+	@Redirect(method = "<init>", at = @At(
+			value = "net.fabricmc.fabric.impl.client.rendering.LayerInjectionPoint",
+			target = "Lnet/minecraft/client/gui/hud/InGameHud;renderMainHud(Lnet/minecraft/client/gui/DrawContext;Lnet/minecraft/client/render/RenderTickCounter;)V")
+	)
+	private LayeredDrawer wrapMainHud(LayeredDrawer instance, LayeredDrawer.Layer layer) {
+		return wrap(MAIN_HUD, instance, layer);
+	}
+
+	// renderExperienceLevel
```
modmuss50: Nit: remove this comment was a left over from when I was adding all of these.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,21 +16,198 @@
...
+				value = "net.fabricmc.fabric.impl.client.rendering.LayerInjectionPoint",
+				target = "Lnet/minecraft/client/gui/hud/InGameHud;renderSleepOverlay(Lnet/minecraft/client/gui/DrawContext;Lnet/minecraft/client/render/RenderTickCounter;)V"
+			)
+	)
+	private LayeredDrawer wrapSleepOverlay(LayeredDrawer instance, LayeredDrawer.Layer layer) {
+		return wrap(SLEEP, instance, layer);
+	}
+
+	@Unique
+	private static LayeredDrawer wrap(Identifier identifier, LayeredDrawer instance, LayeredDrawer.Layer layer) {
```
modmuss50: Nit: Move the static method to the end.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,21 +16,198 @@
...
+		return instance.addLayer(IdentifiedLayer.wrapping(identifier, layer));
+	}
+
+	@Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/LayeredDrawer;addSubDrawer(Lnet/minecraft/client/gui/LayeredDrawer;Ljava/util/function/BooleanSupplier;)Lnet/minecraft/client/gui/LayeredDrawer;", ordinal = 0))
+	private LayeredDrawer wrapSubDrawer1(LayeredDrawer instance, LayeredDrawer drawer, BooleanSupplier shouldRender) {
```
modmuss50: Would it be better to handle the sub drawers by mixing into addSubDrawer? They wouldn't even need IDs that way?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -16,21 +16,198 @@
...
+		return instance.addLayer(new SubLayer(SUB_DRAWER_1, drawer, shouldRender));
+	}
+
+	@Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/LayeredDrawer;addSubDrawer(Lnet/minecraft/client/gui/LayeredDrawer;Ljava/util/function/BooleanSupplier;)Lnet/minecraft/client/gui/LayeredDrawer;", ordinal = 1))
+	private LayeredDrawer wrapSubDrawer2(LayeredDrawer instance, LayeredDrawer drawer, BooleanSupplier shouldRender) {
+		return instance.addLayer(new SubLayer(SUB_DRAWER_2, drawer, shouldRender));
+	}
+
+	@Inject(method = "<init>", at = @At("RETURN"))
+	private void fabric$registerLayers(CallbackInfo ci) {
```
modmuss50: Remove `fabric$` mixin will add the mod id on its own.

fabric-rendering-v1/src/test/java/net/fabricmc/fabric/test/client/rendering/FabricLayeredDrawerTest.java
```diff
@@ -0,0 +1,102 @@
...
+import net.minecraft.client.gui.LayeredDrawer;
+import net.minecraft.client.render.RenderTickCounter;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.client.rendering.v1.FabricLayeredDrawer;
+import net.fabricmc.fabric.api.client.rendering.v1.IdentifiedLayer;
+import net.fabricmc.fabric.impl.client.rendering.FabricLayeredDrawerImpl;
+
+public class FabricLayeredDrawerTest {
```
modmuss50: These unit tests need expanding to cover all of the functions.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudRenderEventsTests.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.rendering.v1.HudLayerRegistrationCallback;
+import net.fabricmc.fabric.api.client.rendering.v1.IdentifiedLayer;
+
+public class HudRenderEventsTests implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		HudLayerRegistrationCallback.EVENT.register(layeredDrawer -> layeredDrawer.addLayerBefore(IdentifiedLayer.MISC_OVERLAYS, Identifier.of("fabric", "test_before_misc_overlay"), (context, tickCounter) -> {
```
modmuss50: This is quite hard to read, I would split the render code out into their own functions in the same way that vanilla does.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudRenderEventsTests.java
```diff
@@ -0,0 +1,58 @@
...
+
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.util.Colors;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.rendering.v1.HudLayerRegistrationCallback;
+import net.fabricmc.fabric.api.client.rendering.v1.IdentifiedLayer;
+
+public class HudRenderEventsTests implements ClientModInitializer {
```
modmuss50: I wonder what automated testing we can do with the new client tests? Might need to wait until https://github.com/FabricMC/fabric/pull/4381 has been merged. Don't let this block this PR, but its a nice thing to keep in mind.

modmuss50 (commented):
This is looking really good now, I have just a few comments (mostly about my own changes :D ) Do let me know if you have any questions or disagree.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricLayeredDrawer.java
kevinthegreat1: `LayeredDrawerWrapper` or `FabricLayeredDrawerWrapper`?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
kevinthegreat1: There could be some problems with flattening all sub drawers, but I'll try that.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudRenderEventsTests.java
kevinthegreat1: Yep, I was planning for that. All the tests in here can be easily made automated after the screenshots stuff become available.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
kevinthegreat1: I've tried to update the names to be consistent.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudLayerRegistrationCallback.java
kevinthegreat1: Added javadocs.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderCallback.java
```diff
@@ -22,10 +22,14 @@
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
+/**
+ * @deprecated Use {@link HudLayerRegistrationCallback} instead. For the closest equivalent, register a layer after {@link IdentifiedLayer#SUBTITLES}.
```
Octol1ttle: > For the closest equivalent, register a layer after {@link IdentifiedLayer#SUBTITLES}.

I'm against this note. Devs can make uninformed decisions about fixing this deprecation based on this note

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
```diff
@@ -0,0 +1,167 @@
...
+	private static boolean matchesIdentifier(LayeredDrawer.Layer layer, Identifier identifier) {
+		return layer instanceof IdentifiedLayer il && il.id().equals(identifier);
+	}
+
+	private interface LayerVisitor {
+		/**
+		 * @return false to remove the layer, true to keep it
+		 */
+		boolean visit(LayeredDrawer.Layer layer, ListIterator<LayeredDrawer.Layer> iterator);
+	}
```
Octol1ttle: I cannot grasp the meaning of `LayerVisitor#visit` having a return value.

Look at removeLayer for example. When a layer is found, `iterator.remove()` is called. Makes sense, it should be removed from the list.

But then `false` is returned, which will result in another call to `iterator.remove()` by `visitLayers`, why? Either the `LayerVisitor` removes the item or `visitLayers` does. Doing this twice seems very confusing to me

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SubLayer.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.client.gui.LayeredDrawer;
+import net.minecraft.client.render.RenderTickCounter;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.client.rendering.v1.IdentifiedLayer;
+
+/**
+ * A layer that wraps another layered drawer that can be added to {@link net.fabricmc.fabric.api.client.rendering.v1.LayeredDrawerWrapper LayeredDrawerWrapper}.
+ *
+ * <p>This implementation is currently not used, as vanilla sub drawers are flattened, but exists for completeness and use if needed.
```
Octol1ttle:
```suggestion
 * <p>This implementation is currently not used, as vanilla sub drawers are flattened, but exists for completeness. Use if needed.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricLayeredDrawer.java
modmuss50: `LayeredDrawerWrapper` is fine, we tend to only use `Fabric` when its an extended version of a vanilla class. In this case its not the vanilla class at all.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
modmuss50: Yes, something isnt right here. I think the return can be removed I think it was from an eariler example of this code. The unit tests clearly need expanding to test this.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/LayeredDrawerMixin.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Mixin(LayeredDrawer.class)
+public abstract class LayeredDrawerMixin {
+	@Shadow
+	@Final
+	private List<LayeredDrawer.Layer> layers;
+
+	@Inject(method = "addSubDrawer", at = @At("HEAD"), cancellable = true)
+	private void flattenSubDrawer(LayeredDrawer drawer, BooleanSupplier shouldRender, CallbackInfoReturnable<LayeredDrawer> cir) {
+		layers.addAll(((LayeredDrawerAccessor) drawer).getLayers());
```
modmuss50: No,  we should flattern them I designed the wrapper around them being layerd, this also breaks vanilla as the `shouldRender` is never checked.

We only need to wrap the `drawer` here so we can get the sub layers in the visitor. I would go back to having the sublayer record I had before (without the id) but creating it here.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
kevinthegreat1: Huh, how did the unit test succeed just fine?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/LayeredDrawerMixin.java
kevinthegreat1: Ah, I misunderstood, done.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
kevinthegreat1: Oh I get it. It currently works, just a bit lacking in documentation. `removeLayer` calls `findLayer` with a lambda that returns false, but `findLayer` always calls `visitLayers` with a lambda that returns true and ignore the return value of the passed in lambda.

Octol1ttle (approved):

PepperCode1 (changes requested):
Currently, adding a layer relative to a layer in a `SubLayer` will also apply the `SubLayer`'s predicate to the new layer, which should not happen. This should be solved by flattening the `SubLayer` when such an addition is requested, so any new layers are always added directly to the layer list of `LayeredDrawerWrapperImpl.base` while preserving layer order.

kevinthegreat1:
I think it makes sense to have the default to respect hide hud, but I also see the need to register layers that always render. I tried to make a solution that does both, but I was only able to come up with a working solution when sub-layers are only one level deep (sub-layers don't contain sub-layers).

Basically I added a `ignoreShouldRender` method to our layer interface. In `SubLayer`, if the sub-layer shouldn't be rendered, we manually render all layers that returns true for `ignoreShouldRender`. This, however, does not work with nested sub-layers.

PepperCode1:
I don't agree that it makes sense for the default to respect the existing predicate, and I think adding additional methods only adds unnecessary complexity for the user when they can very easily add their own check for whether the HUD is hidden. The existing predicate also doesn't even exist all the time, nor do I think the API should need to expose exactly how layers are bundled internally.

PepperCode1:
To further illustrate my point, if I call `wrapper.addLayerBefore(SLEEP, myLayer)`, it will have no additional predicate besides what my layer checks because the `SLEEP` layer is not in a `SubLayer`. However, if I call `wrapper.addLayerAfter(BOSS_BAR, myLayer)`, it will now have the additional "HUD hidden" predicate because the `BOSS_BAR` layer is in a `SubLayer` with that predicate. This is not what a user would expect because `BOSS_BAR` and `SLEEP` are right next to each other in the vanilla layer order. This behavior is not even documented currently, but I am arguing that even if it was, that would be adding a lot of documentation (effectively "adding after, adding before, or replacing this layer will apply an additional predicate to the new layer" or "adding after, adding before, or replacing this layer will **not** apply an additional predicate to the new layer" for each layer in `IdentifiedLayer`) for this inconsistent behavior. Additionally, a method would have to be added to `IdentifiedLayer` (the proposed `ignoreShouldRender`) for whether to ignore an additional predicate, with up to 4 more overloads in `LayeredDrawerWrapper` to make it convenient to use. I do not understand the benefit of adding all of this complexity to the API when the alternative does not require any of it. The only benefit I see is that the user will sometimes not have to write `if (MinecraftClient.getInstance().options.hudHidden) { return; }` inside their layer.

kevinthegreat1:
I agree the behavior can be confusing. I am always happy to add documentation, and the current documentation is in no way final. However, one of the original goals of this pr and the original discussion is that hide hud is automatically applied, as `HudRenderCallback` originally did before it got broken, so I still believe that hide hud and predicate should be respected, at least an option to be respected.

Octol1ttle:
Maybe we can change the terminology: instead of adding a layer before/after a vanilla layer, one "attaches" their layer to the vanilla layer.

I also stand by having the original predicate respected in modded layers.

modmuss50:
This is a valid point that @PepperCode1 raises, I think there are maybe 3 soltuions:
1. Flattern the layers, this was something we were trying so hard not to do.
2. As @Octol1ttle suggested changing the naming/documentation to make it clear how it currently works.
3. Add a layer twice, once after the target layer and once after the sub layer, drawing only one at a time.

Im leaning towards saying option 2 is the way to go, we could maybe rename the layers hidden by this flag to `BOSS_BAR_HUD` for example. I dont think there is an ideal solution, we just need to pick the compromise we want to take.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
```diff
@@ -0,0 +1,176 @@
...
+		return modified.get();
+	}
+
+	private static boolean matchesIdentifier(LayeredDrawer.Layer layer, Identifier identifier) {
+		return layer instanceof IdentifiedLayer il && il.id().equals(identifier);
+	}
+
+	@VisibleForTesting
+	public interface LayerVisitor {
+		/**
```
Octol1ttle: If these methods/interfaces are public, they shouldn't need `@VisibleForTesting`?

apple502j: Quite the opposite...

kevinthegreat1: Moved the test and made them package private.

kevinthegreat1:
Fixed the random commits, I didn't notice that branches already diverged when I made this pr.

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotComparisonOptionsImpl.java
```diff
@@ -49,6 +49,11 @@ public TestScreenshotComparisonOptionsImpl(NativeImage templateImage) {
 		this.templateImage = Either.right(templateImage);
 	}
 
+	@Override
+	public TestScreenshotComparisonOptions save() {
+		return saveWithFileName(getTemplateImagePath());
```
Earthcomputer: Unfortunately this is wrong in a couple of ways. Firstly, it writes the template image before reading it, causing the test to always succeed (at least in dev). Secondly this path points to the runtime resources of the test mod, and not the location where you probably want to write it which is the resource folder in the gradle source set. At dev time this will actually be a folder, but in production this will be a path pointing inside your test mod jar and therefore not writable.

There is already a mechanism to automatically save these templates, I'd suggest you rely on that instead. If you set the system property `fabric.client.gametest.testModResourcesPath` to the path to your test mod resources folder, then screenshot comparisons will write the grabbed screenshot to the templates directory for next time instead of failing. This system property might not be super easy to discover right now but Loom 1.10 will be setting it by default.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudLayerTests.java
```diff
@@ -0,0 +1,147 @@
...
+		context.fill(context.getScaledWindowWidth() / 2 - 150, 0, context.getScaledWindowWidth() / 2 + 150, 15, Colors.YELLOW);
+		context.drawCenteredTextWithShadow(MinecraftClient.getInstance().textRenderer, "6. This yellow rectangle should block the player list.", context.getScaledWindowWidth() / 2, 0, Colors.WHITE);
+	}
+
+	@Override
+	public void runTest(ClientGameTestContext context) {
+		// Set up required test environment
+		context.getInput().resizeWindow(1708, 960); // Twice the default dimensions
+		context.runOnClient(client -> {
+			client.options.hudHidden = false;
```
Earthcomputer, line 108: This shouldn't be necessary as the client gametest runner resets all game options to the defaults before each test

Earthcomputer (changes requested):

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudLayerTests.java
kevinthegreat1: The test failed a previous time without these two lines. (`4303293` fails on github actions.)

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotComparisonOptionsImpl.java
kevinthegreat1: Sorry, but I don't understand your comment. I just wanted a shorthand for saving the screenshot with the same name as the template name, `.saveWithFileName(templateName)`, and this addition seems to work as expected for my case. It saves the screenshot with `templateName` in the screenshot directory as expected. The test fails normally in the same way as using `.saveWithFileName(templateName)`. `getTemplateImagePath()` just seems to return the `templateName`.

Earthcomputer: Hmm, okay I thought this was something else. The name `getTempalteImagePath()` misled me into thinking it's the full path, that function should probably be renamed.

kevinthegreat1: Do you want to add this shorthand in your clean-up pr and rename `getTemplateImagePath`? I think that would be best.

Earthcomputer: Hmm, I don't know, @modmuss50 which would be better?

modmuss50: Yes, please rename it in the test PR, thats ready to merge whenever, so I can do that fairly swiftly to not block this for too long.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudLayerTests.java
Earthcomputer: As discussed on Discord, the GUI scale was the problem, which makes sense because GUI scale 2 isn't the default.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LayeredDrawerWrapper.java
```diff
@@ -0,0 +1,136 @@
...
+	/**
+	 * Attaches a layer before the layer with the specified identifier.
+	 *
+	 * <p>The render condition of the layer being attached to, if any, also applies to the new layer.
+	 *
+	 * @param beforeThis the identifier of the layer to add the new layer before
+	 * @param layer      the layer to add
+	 * @return this layered drawer
+	 */
+	LayeredDrawerWrapper attachLayerBefore(Identifier beforeThis, IdentifiedLayer layer);
```
modmuss50, line 89: Good name 👍

modmuss50 (approved):
Looks great now, thanks for persisting with this, Im glad we spent the time on it and got something thats really good in the end. I also love the tests. 👍

kevinthegreat1:
Small bug fix where the return value of `visitLayers` of sub-layers, `modified`, was not propagated to parent drawers. Comes with a side of documentation and tests.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudLayerRegistrationCallback.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * <p>To register a layer, register a listener to this event and register your layers in the listener.
+ * For common use cases, see {@link LayeredDrawerWrapper}.
+ *
+ * <p>For example, the following code registers a layer after {@link IdentifiedLayer#MISC_OVERLAYS}:
+ * <pre>{@code
+ * HudLayerRegistrationCallback.EVENT.register(layeredDrawer -> layeredDrawer.attachLayerAfter(IdentifiedLayer.MISC_OVERLAYS, Identifier.of("example", "example_layer_after_misc_overlays"), (context, tickDelta) -> {
+ *     // Your rendering code here
+ * }));
+ * }</pre>
```
Juuxel: Nit: You can use `{@snippet}` here

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
```diff
@@ -0,0 +1,179 @@
...
+import org.jetbrains.annotations.VisibleForTesting;
+
+import net.minecraft.client.gui.LayeredDrawer;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.client.rendering.v1.IdentifiedLayer;
+import net.fabricmc.fabric.api.client.rendering.v1.LayeredDrawerWrapper;
+import net.fabricmc.fabric.mixin.client.rendering.LayeredDrawerAccessor;
+
+public class LayeredDrawerWrapperImpl implements LayeredDrawerWrapper {
```
Juuxel:
```suggestion
public final class LayeredDrawerWrapperImpl implements LayeredDrawerWrapper {
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayerInjectionPoint.java
```diff
@@ -0,0 +1,83 @@
...
+import org.objectweb.asm.Handle;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.InsnList;
+import org.objectweb.asm.tree.InvokeDynamicInsnNode;
+import org.spongepowered.asm.mixin.injection.InjectionPoint;
+import org.spongepowered.asm.mixin.injection.struct.InjectionPointData;
+import org.spongepowered.asm.mixin.injection.struct.MemberInfo;
+
+public class LayerInjectionPoint extends InjectionPoint {
```
Juuxel:
```suggestion
public final class LayerInjectionPoint extends InjectionPoint {
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LayeredDrawerWrapper.java
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @param layer the layer to add
+	 * @return this layered drawer
+	 */
+	LayeredDrawerWrapper addLayer(IdentifiedLayer layer);
+
+	/**
+	 * Attaches a layer before the layer with the specified identifier.
+	 *
+	 * <p>The render condition of the layer being attached to, if any, also applies to the new layer.
```
Juuxel, line 82: This is a bit unintuitive IMO, but I think there was a technical reason for this behaviour? I'd assume that people don't typically need to attach layers at specific spots anyway.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LayeredDrawerWrapper.java
```diff
@@ -0,0 +1,136 @@
...
+ * </table>
+ */
+public interface LayeredDrawerWrapper {
+	/**
+	 * Adds a layer to the end of the layered drawer.
+	 *
+	 * @param layer the layer to add
+	 * @return this layered drawer
+	 */
+	LayeredDrawerWrapper addLayer(IdentifiedLayer layer);
```
Juuxel, line 77: Are this and the replace/remove methods tested? I think this is still the key method in the API for most people.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LayeredDrawerWrapper.java
Octol1ttle: Read the full timeline of this pull request for an explanation

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LayeredDrawerWrapper.java
Octol1ttle: Please have a look at `net.fabricmc.fabric.impl.client.rendering.LayeredDrawerWrapperTest` and `net.fabricmc.fabric.test.rendering.client.HudLayerTests.java`

Juuxel: Ah, I only saw the game test which only uses the `attach*` methods.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
```diff
@@ -0,0 +1,179 @@
...
+			return false;
+		});
+	}
+
+	/**
+	 * @return true if a layer with the given identifier was found
+	 */
+	@VisibleForTesting
+	boolean findLayer(Identifier identifier, LayerVisitor visitor) {
+		AtomicBoolean found = new AtomicBoolean(false);
```
Earthcomputer: Imo it's better to use apache's `MutableBoolean` than `AtomicBoolean` for mutating from within a lambda, unless the lambda is called from multiple threads.

PepperCode1: I agree; this change should be made.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
```diff
@@ -0,0 +1,108 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.client.rendering.WrappedLayer;
+
+/**
+ * A hud layer that has an identifier attached for use in {@link LayeredDrawerWrapper}.
+ *
+ * <p>For common use cases, see {@link LayeredDrawerWrapper}.
+ */
+public interface IdentifiedLayer extends LayeredDrawer.Layer {
```
PepperCode1, line 35: The documentation of this class should state that its static final fields are in the order in which their corresponding layer is rendered.

Additionally, I think the documentation should make clear what the default render predicate for each layer is, that this API does not allow changing the existing predicate for a layer, and that operations relative to a layer will generally inherit that layer's predicate. The last two points might belong in the documentation of `LayeredDrawerWrapper` instead.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
PepperCode1: Also on line 149.

PepperCode1 (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/LayeredDrawerWrapperImpl.java
kevinthegreat1: Done in e3f65ea.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/IdentifiedLayer.java
kevinthegreat1: Done in 3384eae.

PepperCode1 (approved):

Juuxel (approved):

PR #4122, opened 2024-09-26
Update to loom 1.8

modmuss50:
WIll undraft once loom 1.8 is stable.


modmuss50:
> java.util.concurrent.CompletionException: java.lang.AbstractMethodError: Receiver class net.fabricmc.fabric.test.resource.loader.ResourceReloadListenerTestMod$2 does not define or inherit an implementation of the resolved method 'abstract void method_14491(net.minecraft.class_3300)' of interface net.minecraft.class_4013.

Looks like there is a remapping issue with the test mods, I will investigate.

PR #4126, opened 2024-09-28
Add script to only enable a subset of projects for development

modmuss50:
This should help those who struggle to build FAPI on their computers, it allows you to "focus" on specific modules.

Run `java gradle/Focus.java fabric-networking-api-v1` to focus on fabric-networking-api-v1 and its dependencies.

Is this something people would find useful to have?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ classes/
 run
 *.log
 *.log.gz
+focus.txt
 
 # Windows
 *.db
--- a/gradle/Focus.java
+++ b/gradle/Focus.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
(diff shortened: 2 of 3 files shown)


apple502j:
Not sure if this is useful for me (I just `runTestmodClient` on module directly), but seems good.

modmuss50:
> Not sure if this is useful for me (I just `runTestmodClient` on module directly), but seems good.

It might help you use an IDE, as this will only load the modules you care about in the IDE, I would strongly reccomend using the IDE run configs, while using intelij to build. Ill only merge this if someone is actually going to use it, otherwise its kinda pointless.

modmuss50:
Going to merge this, easy enough to remove if no one ends up using it.

embeddedt:
Very late, but this will definitely come in handy for me at some point, thank you!

PR #4134, opened 2024-10-01
Add more `c` fluid tags

TelepathicGrunt:
This PR could be a bit more controversial so let it sit for a bit. On Neoforge, I am [proposing an Experience fluid tag](https://github.com/neoforged/NeoForge/pull/1573) and figured it might be a good time to try and propose the missing fluid tags to Fabric if Fabric is ok with them.

The fluid tags were originally in long past, tags that mods in that ecosystem used to have. So there were lots of people with stuff like potion fluids or experience fluids and they tried to agree upon a standard for bottle/bowl rates to buckets for said tags. I am not sure how common these stuff are in fabric's ecosystem or what the agreed upon rates are.

According to the FluidConstants class, fabric ecosystem is 3 bottles to 1 bucket so people are expecting 27000 droplet for bottles and bowls. I added the FluidConstant class in the javadocs for the fluid tags class so new modders know what other modders are expecting when they find a fluid that matches the tag. I also added bowls to the FluidConstants class (assuming it is equal to bottles) so there is less confusions around bowls. But I can remove this new constant if preferred.

I also state in the Javadoc how to get what the potion type is for potion fluids and suspicious stew fluids by using data components (to help make mods know how to interact with each other)

For the experience fluid rate, that one is tricky. After some research, we found these fabric mods with Experience Fluids at different rates:

Kibe: 810 droplets to 1 xp
Create Enchantment Industry: 81 droplet to 1 xp and 81 droplet to 10 xp
XP Obelisk: 810 droplets to 1 xp
Sophisticated Core: 1620 droplets to 1 xp

So I went 810 droplet rate unless those devs decide in Fabric discord that they want a different rate. 810 seems to be a compromise in the middle of these mods. Standardizing the rate will help multiple mods with experience fluid to know how to read each other's volume of fluids properly.


TelepathicGrunt:
Was talking with some devs in discord. The general idea is that smashing a Bottle O' Enchanting yields less droplets of fluids when the experience is converted to the fluid. While extraction from Bottle O' Enchanting yields a perfect 27000 droplets.

The rate then is perfect extraction is 1/3th a bucket while smashing is roughly 1/14th a bucket. Forge ecosystem of mods had a similar thing going for years with 250mb for perfect extraction and roughly 140mb for smashing. They have not seen any issue with this so no xp duplicating rose from that and it is argued that smashing is a lossy conversion of turning the Bottle O' Enchanting to experience.

Fabric convo: https://discord.com/channels/507304429255393322/507982478276034570/1290832448166432850

Shnupbups (approved):
A bit different of a take on Experience than I used in my old mod Cauldrons, but I think it works.
Maybe add one more tag for Dragon's Breath, the last of the 'quasi-fluids' in Vanilla? `c:dragon_breath` would work fine.

TelepathicGrunt:
Dragons breath and ominous potion might be better as a separate pr as not even neo or forge had tags for them. Might be worth their own discussion separate?

Noaaan (approved):
The concept in general would open up for better/more compat in the future. I could imagine being able to pump liquid XP directly into various enchanting workplaces, or potentially mixing potions together

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalFluidTags.java
```diff
@@ -27,15 +27,63 @@
...
  * This is the only tag category exempted from many-different-types plural rule.
+ *
+ * <p>(See {@link net.fabricmc.fabric.api.transfer.v1.fluid.FluidConstants} for the correct droplet rates for containers)
  */
 public final class ConventionalFluidTags {
 	private ConventionalFluidTags() {
 	}
 
+	/**
+	 * Holds all fluids related to water without the behaviors attached to the vanilla lava fluid tag.
```
VoidLeech: This should say "fluids related to lava"

modmuss50 (approved):

Juuxel (approved):
Makes sense to standardise.

PR #4136, opened 2024-10-02
Update mod to Minecraft snapshot 24w40a

### LINKED ISSUE #4135: Dyed bundles are not in the respective dyed tags

haykam821:
This pull request ports Fabric API to Minecraft snapshot 24w40a without any breaking changes.

In addition, the `c:is_dark_forest` biome tag has been added. This biome tag includes the dark forest and pale garden biomes, as the latter is a variant of the former, as stated by [the Minecraft changelog](https://www.minecraft.net/en-us/article/minecraft-snapshot-24w40a):

> #### Pale Garden
>
> Pale Garden is a biome variation of Dark Forest.

The game runs, but when creating a new world after exiting another world, the game may crash with the message `Caused by: java.lang.IllegalStateException: Tags already captured, this should not happen`.


modmuss50 (approved):
Looks great, I cannot reproduce the crash, im happy to release it as-is. The recipe stuff can be improved over the coming weeks.

haykam821:
I can try to reproduce the crash again later. The only differences I can think of that my environment may have is worlds with feature flags enabled and the test worlds from #4118.

I do realize that the pale oak chest boat is missing from the `c:boats` tag; that (and probably #4135) should be fixed as well.

modmuss50:
> I do realize that the pale oak chest boat is missing from the c:boats tag; that (and probably https://github.com/FabricMC/fabric/issues/4135) should be fixed as well.

Fixed both, thanks.

I have tried worlds with the experiments enabled and all seems fine. Ill release this now and we can always fix the crash later.

fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/EntityTypeTagGenerator.java
```diff
@@ -52,6 +52,7 @@ protected void configure(RegistryWrapper.WrapperLookup registries) {
 				.add(EntityType.JUNGLE_CHEST_BOAT)
 				.add(EntityType.ACACIA_CHEST_BOAT)
 				.add(EntityType.CHERRY_CHEST_BOAT)
+				.add(EntityType.PALE_OAK_BOAT)
```
haykam821: Should this entry not be optional?

modmuss50: No, I dont think so, entity types are always loaded into a static registry, even when disabled.

fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/EntityTypeTagGenerator.java
```diff
@@ -46,15 +46,25 @@ protected void configure(RegistryWrapper.WrapperLookup registries) {
...
+				.add(EntityType.SPRUCE_BOAT)
 				.add(EntityType.SPRUCE_CHEST_BOAT)
+				.add(EntityType.BIRCH_BOAT)
 				.add(EntityType.BIRCH_CHEST_BOAT)
+				.add(EntityType.JUNGLE_BOAT)
 				.add(EntityType.JUNGLE_CHEST_BOAT)
+				.add(EntityType.ACACIA_BOAT)
 				.add(EntityType.ACACIA_CHEST_BOAT)
+				.add(EntityType.CHERRY_BOAT)
 				.add(EntityType.CHERRY_CHEST_BOAT)
```
haykam821: The regular boats are already added by the `minecraft:boat` entity type tag here. Is the duplication necessary?

modmuss50: Oh right, my bad.

PR #4143, opened 2024-10-05
Update add method for StatType in TranslationBuilder

solonovamax:
Use either the translation key from the name or fallback to Util.createTranslationKey formatting for StatType in TranslationBuilder

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLanguageProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLanguageProvider.java
@@ -207,7 +207,11 @@ public abstract class FabricLanguageProvider implements DataProvider {
 		 * @param value    The value of the entry.
 		 */
 		default void add(StatType<?> statType, String value) {
-			add("stat_type." + Registries.STAT_TYPE.getId(statType).toString().replace(':', '.'), value);
+		    if (statType.getName() instanceof TranslatableTextContent translatableTextContent) {
+		        add(translatableTextContent.getKey(), value);
+	        } else { // fallback to Util.createTranslationKey formatting (should it error instead?)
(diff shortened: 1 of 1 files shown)


solonovamax:
I was originally just going to have the `Util.createTranslationKey("stat_type", Registries.STAT_TYPE.getId(statType))` line, but I decided it might be a good idea to have semi-parity with `add(RegistryKey<ItemGroup> registryKey, String value)`, so I also added the test for if the name is a `TranslatableTextContent`

solonovamax:
oopsies, checkstyle issues. I'll fix that in a bit.

solonovamax:
~~Odd, it seems that when checking it out locally, the `check` task passes, but not when run in CI.~~ edit: nvm I was on the default branch not my own branch :skull:

but, the issue seems to be my use of spaces over tabs.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLanguageProvider.java
```diff
@@ -207,7 +207,11 @@ default void add(RegistryEntry<EntityAttribute> entityAttribute, String value) {
 		 * @param value    The value of the entry.
 		 */
 		default void add(StatType<?> statType, String value) {
-			add("stat_type." + Registries.STAT_TYPE.getId(statType).toString().replace(':', '.'), value);
+			if (statType.getName() instanceof TranslatableTextContent translatableTextContent) {
+				add(translatableTextContent.getKey(), value);
+			} else { // fallback to Util.createTranslationKey formatting (should it error instead?)
```
apple502j, line 212: Error? not sure how the translation would be used if the name is not TTC.

solonovamax: I just threw in a fallback because you never know, some person might try and have like some string literal concatenated with the translation. But I can always make it an error.

PR #4155, opened 2024-10-10
Fix After Translucent World Render Event

AzureAaron:
Fixes the `AFTER_TRANSLUCENT` world render event not working due to it not being attached to any render pass, which also caused the test mod to crash.

Also fixes the test mod using the wrong box drawing method and added some further specifications as to the exact type of box that should be rendered.


modmuss50:
Does this also apply to 1.21.1 or did it break during the 1.21.2 update?

apple502j:
@modmuss50 From the diff I believe this broke in 1.21.2 cycle. Not tested though.

AzureAaron:
> Does this also apply to 1.21.1 or did it break during the 1.21.2 update?

The event works fine in 1.21.1, it only broke in 1.21.2 with the introduction of the frame graph and render pass stuff.

modmuss50:
Ok, perfect thanks.

modmuss50 (approved):

PR #4171, opened 2024-10-17
Remove LevelProperties check

SemmieDev:
Removes the check that makes sure `saveProperties` on `MinecraftServer` is an instance of `LevelProperties`, as since the `LevelProperties` class is unused in the entire fabric API. This allows mods to use their own implementation of `SaveProperties`.


modmuss50 (approved):
Looks fine to me, seems to be left over from old times.

PR #4180, opened 2024-10-22
Make dynamic registry element path include namespace

### LINKED ISSUE #4179: Dynamic registry loading location ignores namespace
The dynamic registry loading code calls `RegistryKeys.getPath` which, unlike `getTagsPath`, is unpatched by Fabric API. This means that my custom `bingo:difficulty` dynamic registry gets loaded from `data/<namespace>/difficulty` instead of the namespaced `data/<namespace>/bingo/difficulty`.

Gaming32:
Fixes FabricMC/fabric#4179

Note that since this changes the loading location of modded dynamic registries, this is technically a breaking change to datapacks. However, with 1.21.2 being brand new, and that version already breaking datapack folder structures (with plurals becoming singular), I think now is the perfect time to fix this.


Gaming32:
An additional thing I will note is that fixing this realigns the behavior with the docs of `DynamicRegistries`. I do believe this used to be the old behavior before `RegistryKeys.getPath` was added.

modmuss50:
Change seems fine to me, would it be possible to add some tests to help ensure that this doesnt break again in the future? Could maybe expand one of the existing dynmatic registry tests instead of creating a whole new one.

I get that its a breaking change, but I think the best solution is to just go for it in 1.21.3 otherwise we will be waiting a long time to fix. If anyone disagrees please let your thoughts be known.

Gaming32:
I can add tests, sure.

Gaming32:
I'm not entirely sure how to add tests for this. However, `JsonDataLoader` is now broken without this.

PepperCode1 (approved):

modmuss50:
I have added a simple test for this, should stop it from happening again.

PR #4183, opened 2024-10-23
Add ServerChunkEvents.Generate

jpenilla:
Add a new chunk generate event, fired alongside the chunk load event when a chunk is first upgraded to full status.

My use case for this is a world map mod, I need to know when new chunks are generated to map them. I don't need to map a chunk that is just loading, only generated, I have other hooks to re-map chunks when blocks change.

Until now I've used a mixin to accomplish this, however, this makes compatibility more difficult with mods like Moonrise that rewrite the chunk system.

Some additional context (Mojang mappings):
- https://github.com/jpenilla/squaremap/commit/cbee307394cb417b4fcc9ac67fb1b0dd5b842070
- https://github.com/Tuinity/Moonrise/blob/a70073ae3e64ddf77829ada3c05a4c7e3c81554d/fabric/src/main/java/ca/spottedleaf/moonrise/fabric/FabricHooks.java#L88
- https://github.com/Tuinity/Moonrise/blob/a70073ae3e64ddf77829ada3c05a4c7e3c81554d/neoforge/src/main/java/ca/spottedleaf/moonrise/neoforge/NeoForgeHooks.java#L82

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -38,6 +38,17 @@ public final class ServerChunkEvents {
 	});
 
 	/**
+	 * Called when a newly generated chunk is loaded into a ServerWorld.
+	 *
+	 * <p>When this event is called, the chunk is already in the world.
+	 */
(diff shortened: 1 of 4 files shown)


modmuss50:
Please add a tests for this, fine if its something simple such as a long line or maybe place a block somewhere.

jpenilla:
I've added a simple logging-based test with instructions to verify behavior in the Javadoc.

modmuss50 (approved):

PR #4184, opened 2024-10-25
Revert "Prevent vanilla clients from joining servers that require modded registry entries. (#4169)"

modmuss50:
This reverts commit 56ec7ac6d8661d6698ccd35fc3c6c524242886af.

Reverts this again, I still want to make this change but it needs to be better documented and ensure that it matches the (unexpected) behaviour when there is a FAPI client.


zOnlyKroks (approved):
lgtm

PR #4187, opened 2024-10-26
Fix overriding vanilla translations

apple502j:
Fixes a longstanding bug where overriding vanilla translation can cause non-overridden ones to fail to load. Also adds a test.

Request porting to all maintained versions, if possible to 1.20.1.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/server/LanguageMixin.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/server/LanguageMixin.java
@@ -35,6 +35,8 @@ import org.spongepowered.asm.mixin.injection.Redirect;
 import net.minecraft.util.Language;
 
 import net.fabricmc.fabric.impl.resource.loader.ServerLanguageUtil;
+import net.fabricmc.loader.api.FabricLoader;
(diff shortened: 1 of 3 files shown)


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/mixin/resource/loader/server/LanguageMixin.java
```diff
@@ -51,6 +53,18 @@ private static Map<String, String> create(Map<String, String> map) {
...
 	}
 
+	@Redirect(method = "load(Ljava/util/function/BiConsumer;Ljava/lang/String;)V", at = @At(value = "INVOKE", target = "Ljava/lang/Class;getResourceAsStream(Ljava/lang/String;)Ljava/io/InputStream;"))
+	private static InputStream readCorrectVanillaResource(Class instance, String path) throws IOException {
+		ModContainer mod = FabricLoader.getInstance().getModContainer("minecraft").orElseThrow();
+		Path langPath = mod.findPath(path).orElse(null);
+
+		if (langPath == null) {
+			throw new IOException("Could not read %s from minecraft ModContainer".formatted(path));
+		} else {
```
modmuss50, line 63: Nit: no need for the else

modmuss50 (approved):

PR #4188, opened 2024-10-27
Update Loom and Loader

modmuss50:
Fixes some dev woes


modmuss50:
Going to merge this now even though loader is still in bets, ill just have to make sure that loader is no longer a beta before releasing.

PR #4198, opened 2024-11-01
Add tag aliases

Juuxel:
The tag alias API lets modders merge different but equivalent tags together so that they have the same content. For example, tag aliases can be used to unify matching `minecraft` and `c` tags. (This is the main purpose of the PR.)

Tag aliases are defined in data packs like tags. For example, a block tag alias group for fences would be located at `data/my_mod/fabric/tag_aliases/block/fences.json` with the contents:
```java
{
  "tags": [
    "minecraft:fences",
    "c:fences"
  ]
}
```
The files can also be generated with the data generation API. Alias groups can be replaced (or disabled) like most data pack files by having a different file with the same name at the same file path.

TODO:
- [x] Better docs
- Tests
  - [x] Functionality
  - [x] Data generation


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -116,6 +125,34 @@ protected FabricTagBuilder getOrCreateTagBuilder(TagKey<T> tag) {
 		return new FabricTagBuilder(super.getOrCreateTagBuilder(tag));
 	}
 
+	/**
+	 * Gets an {@link AliasGroupBuilder} with the given ID.
+	 *
+	 * @param groupId the group ID
+	 * @return the alias group builder
+	 */
+	protected AliasGroupBuilder getOrCreateAliasGroupBuilder(Identifier groupId) {
```
apple502j: I always wanted to drop the `getOrCreate`, but haven't gotten an opportunity to do so. Maybe do that for future ones?

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/impl/tag/TagAliasLoader.java
```diff
@@ -0,0 +1,176 @@
...
+
+			for (Map.Entry<Identifier, Resource> entry : resourceFinder.findResources(manager).entrySet()) {
+				Identifier resourcePath = entry.getKey();
+				Identifier groupId = resourceFinder.toResourceId(resourcePath);
+
+				try (Reader reader = entry.getValue().getReader()) {
+					JsonElement json = JsonParser.parseReader(reader);
+					Codec<TagAliasGroup<Object>> codec = TagAliasGroup.codec((RegistryKey<? extends Registry<Object>>) registryKey);
+
+					switch (codec.parse(JsonOps.INSTANCE, json)) {
```
apple502j, line 87: Shouldn't this be parsed with `RegistryOps`?

apple502j (commented):
Will test when I have time

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/impl/tag/TagAliasLoader.java
Juuxel: The codec doesn't depend on `RegistryOps` since it doesn't parse registry entries, only tag *keys* (which don't depend on registry content).

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
Juuxel: Sure, I just have it to match the existing methods. I think `aliasGroup` is probably enough (the "builder" part is also not very useful)

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -396,4 +433,49 @@ public final FabricTagBuilder add(RegistryKey<T>... registryKeys) {
 			return this;
 		}
 	}
+
+	/**
+	 * A builder for tag alias groups.
+	 */
+	public final class AliasGroupBuilder {
```
modmuss50, line 440: Nit: private constructor

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/api/tag/v1/TagAliasGroup.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * <p>If you want to have the contents of a {@code c} tag in your own tag, prefer including the {@code c} tag
+ * in your tag file directly. That way, data packs can modify your tag separately. Tag aliases make their contained
+ * tags almost fully indistinguishable since they get the exact same content, and you have to override the alias group
+ * in a higher-priority data pack to unlink them.
+ *
+ * @param tags the tags in the group, must be from the same registry
+ * @param <T> the type of registry entries in the tags
+ */
+public record TagAliasGroup<T>(List<TagKey<T>> tags) {
```
modmuss50: Does this have a use being part of API?

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/api/tag/v1/TagAliasGroup.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+
+import com.mojang.serialization.Codec;
+
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.tag.TagKey;
+
+/**
+ * A group of tags that refer to the same set of registry entries.
```
modmuss50: Maybe move this doc to the package-info, especially if this class is moved out of api.

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/impl/tag/TagAliasLoader.java
```diff
@@ -0,0 +1,176 @@
...
+
+			// Remove any groups of one tag, we don't need to apply them.
+			groupsByTag.values().removeIf(tags -> tags.size() == 1);
+
+			RegistryWrapper.Impl<?> wrapper = registries.getOrThrow(entry.getKey());
+
+			if (wrapper instanceof TagAliasEnabledRegistryWrapper aliasWrapper) {
+				aliasWrapper.fabric_loadTagAliases(groupsByTag);
+			} else {
+				LOGGER.error("[Fabric] Couldn't apply tag aliases to registry wrapper {} ({}) since it had an unknown type", wrapper, entry.getKey().getValue());
```
modmuss50: Nit: This sounds like a fatal error to me. If it happens its likely an issue with fabric or another mod and not the datapack.

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/mixin/tag/SimpleRegistryMixin.java
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * <p>Additionally, the {@link TagAliasEnabledRegistryWrapper} implementation is for dynamic registry tag loading.
+ */
+@Mixin(SimpleRegistry.class)
+abstract class SimpleRegistryMixin<T> implements SimpleRegistryExtension, TagAliasEnabledRegistryWrapper {
+	@Unique
+	private static final Logger LOGGER = LoggerFactory.getLogger("fabric-tag-api-v1");
+
+	@Unique
+	private Map<TagKey<?>, Set<TagKey<?>>> fabric_pendingTagAliasGroups;
```
modmuss50: Nit: no need for `fabric_` prefix when using @Unique

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/mixin/tag/SimpleRegistryMixin.java
```diff
@@ -0,0 +1,124 @@
...
+			List<RegistryEntry<T>> entriesAsList = List.copyOf(entries);
+
+			// Replace the old entry list contents with the merged list.
+			for (TagKey<?> tag : aliasGroup) {
+				RegistryEntryList.Named<T> entryList = tagLookup.getOptional((TagKey<T>) tag).orElseThrow();
+				entryList.entries = entriesAsList;
+			}
+		}
+
+		LOGGER.info("[Fabric] Loaded {} tag alias groups for {}", uniqueAliasGroups.size(), key.getValue());
```
modmuss50: Nit: debug

modmuss50 (commented):
Nothing major standing out to me

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/impl/tag/TagAliasLoader.java
Juuxel: Agreed, but I don't think throwing makes much sense here since it can put resource reloading in a loop.

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/api/tag/v1/TagAliasGroup.java
Juuxel: Technically it could be useful for people writing their own data generation code, but the javadoc should be enough to clarify the format. Then the codec method doesn't strictly need to be exposed in the API.

A class might be more visible than package docs though, so I'll also duplicate the most important parts to a README (which I forgot to create).

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/impl/tag/TagAliasLoader.java
modmuss50: I was thinking about this the other day, we should really do something to break out of that loop and display an error screen or just crash fatally. Not something for this PR of course.

apple502j: (The infinite loop is for resource packs only, and I am not sure if that happens in 1.21.4. There's some refactor around this code every other version or so...)

fabric-tag-api-v1/src/main/java/net/fabricmc/fabric/api/tag/v1/TagAliasGroup.java
Juuxel: Another thing to consider is that the data gen API needs to access this (the codec), and I don't really like modules using each other's internals.

modmuss50: Thats fair, are we happy to expose the record as API? Generally we dont.

modmuss50 (approved):
This looks good to me. Just one or two minor nit picks. Ill move to last call now (expect a full week).

@TelepathicGrunt would you be able to take a quick look at the readme/package doc to make sure that it covers your use cases/looks reasonable.

TelepathicGrunt (approved):

PR #4199, opened 2024-11-01
[1.21.4] Add stripped pale oak logs & wood to their respective conventional tags

### LINKED ISSUE #4181: Pale oak logs are not included in the stripped logs and stripped woods tags
#4146 introduced tags for stripped logs and stripped woods, but the relevant pale oak logs are not included in either.

### LINKED ISSUE #4182: The raw material item tag is built twice
https://github.com/FabricMC/fabric/blob/8cb06cc20b4da041da115518957249472f7e6091/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/ItemTagGenerator.java#L466-L473

As a result, each entry is listed in the tag twice.

Prospector:
Fixed stripped log tag issues

Fixes #4181


modmuss50:
I think this change would be best made as 2 PRs, it will make merging a lot easier as I can just cherry-pick the raw material changes into the other branches.

Prospector:
Split off raw materials into #4203

modmuss50 (approved):

PR #4201, opened 2024-11-01
Backport Fix for FabricMC/fabric#4182 to 1.21.2

Prospector:


modmuss50:
This PR is the same as https://github.com/FabricMC/fabric/pull/4200 just differs by title? They both target the same 1.21.3 branch? We also no longer support 1.21.2.

Please also run the generate resources task.

Prospector:
Whoops, I also didn't mean for this to be 1.21.2 anyway it was supposed to be 1.21.1 😅 

Closed these backports so you can just cherry-pick #4203 into any supported versions (looks like this error goes back quite a ways, looks like as far back as 1.20.5)

PR #4219, opened 2024-11-09
Fix AFTER_SETUP event not being able to render

### LINKED ISSUE #4195: WorldRenderEvents.AFTER_SETUP fires immediately before a glClear() preventing rendering in MC 1.21.3
Minecraft added a `glClear` call right after the `WorldRenderEvents.AFTER_SETUP` call is fired, preventing any rendering from happening during that event.

This is a blocking issue for Distant Horizons since the next event in the stack `BEFORE_ENTITIES` is fired too late in Minecraft's rendering pipeline (I need to draw into the skybox).
If this is expected behavior moving forward please let me know so I can look into other solutions, thanks!

### Previous behavior: MC 1.21.1 
Fabric API `0.107.0+1.21.1`

**Final Frame:**
image

**Render Doc OpenGL event list:**
image

------

### New behavior: MC 1.21.3
Fabric API `0.107.0+1.21.3`

**Final Frame:**
image

**Render Doc OpenGL event list:**
image

-----

Here are render doc snapshots of both Minecraft versions in case you want to look closer:
[render dock snapshots.zip](https://github.com/user-attachments/files/17564672/render.dock.snapshots.zip)

JustRed23:
Due to 1.21.2 changing the way how level rendering works, adding a `RenderSystem.clear` after the `AFTER_SETUP` event, rendering is not possible with this event, so I moved the event to a later point after the clear call but before the terrain rendering

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
@@ -69,6 +69,8 @@ public final class WorldRenderEvents {
 	 * identified and rebuilt but before chunks are uploaded to GPU.
 	 *
 	 * <p>Use for setup of state that depends on view frustum.
+	 *
+	 * <b>Can NOT be used to render anything after MC 1.21.2 due to a glClear call, use {@link WorldRenderEvents#BEFORE_TERRAIN} instead.
 	 */
 	public static final Event<AfterSetup> AFTER_SETUP = EventFactory.createArrayBacked(AfterSetup.class, context -> { }, callbacks -> context -> {
 		for (final AfterSetup callback : callbacks) {
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
@@ -70,7 +70,7 @@ public final class WorldRenderEvents {
 	 *
 	 * <p>Use for setup of state that depends on view frustum.
 	 *
-	 * <b>Can NOT be used to render anything after MC 1.21.2 due to a glClear call, use {@link WorldRenderEvents#BEFORE_TERRAIN} instead.
+	 * <b>Can NOT be used to render anything after MC 1.21.2 due to a glClear call, use {@link WorldRenderEvents#BEFORE_TERRAIN} instead.</b>
 	 */
 	public static final Event<AfterSetup> AFTER_SETUP = EventFactory.createArrayBacked(AfterSetup.class, context -> { }, callbacks -> context -> {
 		for (final AfterSetup callback : callbacks) {
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
@@ -264,11 +264,6 @@ public final class WorldRenderEvents {
 	}
 
 	@FunctionalInterface
-	public interface BeforeTerrain {
-		void beforeTerrain(WorldRenderContext context);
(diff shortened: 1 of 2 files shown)


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
```diff
@@ -83,6 +83,19 @@ private void afterTerrainSetup(Camera camera, Frustum frustum, boolean hasForced
...
 
+	@Inject(
+			method = "method_62214",
+			at = @At(
+				value = "INVOKE",
+				target = "Lnet/minecraft/util/profiler/Profiler;push(Ljava/lang/String;)V",
+				ordinal = 0,
+				shift = Shift.AFTER
+			)
+	)
```
Alexander01998: This points to `profiler.push("terrain");` if I'm not mistaken? Might be a good idea to note that with a comment since it's not obvious from the code.

JustRed23: It does, I'll add a comment

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
```diff
@@ -68,13 +69,26 @@ private WorldRenderEvents() { }
 	 * identified and rebuilt but before chunks are uploaded to GPU.
 	 *
 	 * <p>Use for setup of state that depends on view frustum.
+	 *
+	 * <b>Can NOT be used to render anything after MC 1.21.2 due to a glClear call, use {@link WorldRenderEvents#BEFORE_TERRAIN} instead.
```
watertrainer: building fails for me because of a missing closing tag for the bold text here, works after adding it

JustRed23: Weird.. the rest of the docs also don't use closing tags

Juuxel: The docs don't use closing tags for `<p>`. That format is pretty conventional for javadocs, and I think it's a special case in the Checkstyle rules. (In valid HTML, the end tag `</b>` can't be omitted but `</p>` can in this case.)

AsynchronousAI: I got this issue also, adding `</b>` fixed this.

AsynchronousAI:
LGTM!

Alexander01998 (approved):

modmuss50 (changes requested):
It was pointed out to me on discord that the old API is broken, this should be fixed to work as it did it before (or as close to it). There is no point adding a new api and leaving a broken API in its place.

JustRed23:
> It was pointed out to me on discord that the old API is broken, this should be fixed to work as it did it before (or as close to it). There is no point adding a new api and leaving a broken API in its place.

Removed new event, fixed existing event

modmuss50 (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
```diff
@@ -80,6 +80,18 @@ private void beforeRender(ObjectAllocator objectAllocator, RenderTickCounter tic
 	@Inject(method = "setupTerrain", at = @At("RETURN"))
 	private void afterTerrainSetup(Camera camera, Frustum frustum, boolean hasForcedFrustum, boolean spectator, CallbackInfo ci) {
 		context.setFrustum(frustum);
+	}
+
+	@Inject(
+			method = "method_62214",
+			at = @At(
+				value = "INVOKE",
```
Chocohead: Would this not be a good place to use `INVOKE_STRING`? That way this can specifically target the `terrain` profiler push call, even if it gets moved around

Verteiron (approved):

PR #4221, opened 2024-11-10
Change Creative Buttons Texture

matthewperiut:
# Screenshots of updated texture and position

image
image
image
image

Programmer art resource pack:
image  
image  

# **BELOW IS OUTDATED!**

image

tl;dr

https://github.com/user-attachments/assets/4685653d-0541-40f6-afea-477b2b22460d

# Proposal to Change Creative Buttons Texture
The creative buttons texture has haunted me for years in fabric-api. Why?  
They're blue. The entire interface for buttons hasn't been blue in over 6 years.  
Additionally they have too much contrast, sticking out like a sore thumb in the user interface.

# How do we fix it?
We have a great example of something interactable in the creative menu already, the scroll bar.
It has an inactive state and an active state:  
active:  
scroll_enabled
inactive:  
scroll_disabled

# Extrapolating the Features
Notice that the creative scroll bar is in an indented section where it can move freely.  
This creates a contrast strong enough to differentiate it from the rest of the menu.  
The "grip" lines horizontal on the bar also changes color depending on if you can or cannot scroll.    

# Using the Features
Including the indent to seperate it from the rest of the UI helps the contrast to make it visible, without making it stand out visually from the slots and the scroll bar.
Using the pallette of the scroll bar in its active and inactive state for the pallette in the buttons being active and inactive makes it more flush with the design and not stick out.

# Result:
Default appearance:
default_appearance
Hovering on right:
hover_right
Hovering on left:
hover_left
Disabled left:
disabled_left
Disabled right:
disabled_right

# Full screenshot view of change

# Previous appearance for reference:

# Small Feats
Reduced color pallete allows for greater compression  
5,512 bytes -> 382 bytes  
14x decrease in size  

Jar size improvement:  
1,992,580 -> 1,989,946  
0.13% total jar size improvement  
(small things matter LOL)  
detail: compiled on arch linux with Temurin-21.0.5+11-LTS   

# Footnote additional details  
The arrow is made white instead of the button outline because it becomes too noisy in a small area.  
Minecraft 1.15 removed blue backgrounds from buttons in Minecraft, and it came out in 2019, 6 years ago. 
Previous attempts, which the textures were worse rip- #2740, #2746


Laxystem:
I believe it makes a lot more sense to have secondary buttons there instead of primary ones.

In-game UI mostly doesn't have primary buttons (except for the `p` menu - but that one uses them as tabs, which means it's outdated anyway).

matthewperiut:
Another reference, back in 1.14, where hovered buttons were still blue this texture was awesome!
Screenshot 2024-11-11 at 1 45 55 PM
but nowadays it really doesn't make much sense with the colors chosen to be inside the existing gui as-is

embeddedt:
I like the new look, but the right edge of the buttons doesn't align with the right edge of the scrollbar. To be fair this was also an issue with the previous implementation, but it seems more obvious now.

matthewperiut:
Does this look better?
image
Alternatively, this?
image

embeddedt:
I like the first revision, yes.

matthewperiut:
image
I further modified it to have the disabled button be a bit nicer

IMB11:
I think this looks great :+1:

modmuss50 (approved):
Im happy with this, its a definite improvement. If anyone thinks they can be done better again please let it be known as I dont plan on touching this yet agian.

ghost:
Only issue I can think of is that by trying to fit into the vanilla scheme here the contrast on the active buttons end up being low.

That may not be an issue due to the physical location but I believe it is something that should still be thought about.

matthewperiut:
I just found a last minute thing, surprised it wasn't seen earlier by past me, but don't merge yet!
image

matthewperiut:
Alright, I'm thinking about the solutions, and here are the options,
1. Move the buttons back to their usual location
2. Shorten the search bar
3. Move the search bar

I'm not really sure what to do, but I think probably move the buttons back

embeddedt:
I actually think shortening the search bar would look better.

Laxystem:
@matthewperiut there's a fourth option, replace the outline with a normal cell border on the search tab.

IMB11:
Shortening the search bar would be perfectly acceptable in my opinion, users won't notice and it's aesthetically pleasing.

Syst3ms (changes requested):
Shorten the search bar, as per consensus. Making a review of this so it doesn't get forgotten when last call ends.

matthewperiut:
Yeah sorry guys work got busy, let me try and get this finished

matthewperiut:
image
getting there guys...

matthewperiut:
Ok, now it is ready (I think)

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
```diff
@@ -68,11 +73,11 @@ private void updateSelection() {
 		}
 	}
 
-	@Inject(method = "init", at = @At("RETURN"))
+	@Inject(method = "init", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/widget/TextFieldWidget;setEditableColor(I)V", shift = At.Shift.AFTER))
```
matthewperiut, line 71: This was changed to stop creating creative buttons when in survival mode, incase anyone asks the purpose

matthewperiut:
OK I'm taking my grubby hands off this code and various like OO this could be betters and handing it off to just be reviewed, do keep in mind that I tried to keep my code as minimally invasive as possible for the changes that I made. Feel free to ask questions for my decisions. This is what the search tab looks like now
image

IMB11:
Looks great IMO :+1:

matthewperiut:
Me thinking I'm done, having random thoughts of what ifs and how to fix them later... guys maybe its truly ready now.

fabric-item-group-api-v1/src/main/resources/assets/fabric/textures/gui/creative_buttons.png
IMB11, line 1: Does this PNG file need to be this big? (256x256 is a bit overkill)

matthewperiut:
I decided not to mess with the png size as that is how it was originally, I am truly solely trying to get the design of the button to be different and making sure the code I add isn't inefficient. But yes, in the current version widgets are usually just small images with a different image loading system.

IMB11 (approved):

cassiancc (changes requested):
Before this is merged, I'd like the old textures to be added to Programmer Art, so that Fabric can continue to match the older blue button style when that resource pack is enabled. I've gone ahead and set this up as a PR to matthewperiut's fork as https://github.com/matthewperiut/fabric/pull/1.

matthewperiut:
I feel that the user base for the addition of keeping the old textures in programmer art is a little too niche where it should just be an additional external resourcepack outside of fabric api. I will go ahead and merge it if @Syst3ms or @modmuss50 want me to though.

cassiancc:
There seems to be approval in the Discord. Mod Menu also ships blue buttons for [Programmer Art ](https://github.com/TerraformersMC/ModMenu/blob/1.21.2/src/main/resources/programmer_art/assets/modmenu/textures/gui/configure_button.png)and High Contrast (which I'll come back to at a later point), as they are both built-in resource packs.
image

modmuss50:
Yes, I think adding back the old textures to programmer art is a great idea. Lets do that.

I know this PR has been in last call for a very long time, but I think its been worth it to make sure that everyone can provide their input. I want to make sure we are all 100% (or as close as) happy with this before we merge it. Id rarther not come back to this anytime soon.

matthewperiut:
The people have spoken, its merged

cassiancc (approved):

matthewperiut:
Adding the programmer art textures has an unintended consequence of two pixels remaining that weren't masked over because in the new texture they are covered. See picture
Screenshot 2025-01-06 at 12 21 01 PM
The top left and bottom left pixels should not be there as they are a remnant of the previous search bar location pixels and were hidden before, so this next commit is to accommodate for the programmer art without this artifact, with this result:
Screenshot 2025-01-06 at 12 20 11 PM

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
```diff
@@ -197,4 +204,38 @@ public boolean setSelectedItemGroup(ItemGroup itemGroup) {
...
+	@WrapOperation(
+			method = "init",
+			at = @At(
+					value = "NEW",
+					target = "net/minecraft/client/gui/widget/TextFieldWidget"
+			)
+	)
+	private TextFieldWidget shortenSearchBarTextField(TextRenderer textRenderer, int x, int y, int width, int height, Text text, Operation<TextFieldWidget> original) {
+		// Modify only the available text space width to be 5 pixels shorter
+		if (hasAdditionalPages()) {
```
modmuss50: No, I am strongly against changing any vanilla UI element like this, we shouldnt make something smaller, why does this need to be done?

modmuss50 (changes requested):

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
matthewperiut: A while ago someone said that the location of the buttons should be further to the left, and I moved it and people said its better moved in line with the search bar. Then I noticed the search bar goes in where the buttons go, so I asked if I should move back the buttons or shorten the search bar or move the search bar, and @Syst3ms requested to shorten the search bar basically based off consensus, which seemingly most people agreed.

This specific part of the code doesn't let the text in the search bar go too far to the right.

Another part renders the end of the search bar a little to the left.

What do you suggest I do?

matthewperiut: Without this search bar change with the button locations as-is it appears like this:
image

matthewperiut: Additionally the search bar will keep going just like normal, because it allows for text longer than the length of the given area, with mojang's code making it look nice in the process

cassiancc: Fabric is already changing the vanillla UI by adding buttons, is making sure our buttons fit a concern? These changes will only be made if additional Creative groups have been added by a mod.

matthewperiut:
Would this be preferable?

Laxystem:
@matthewperiut that's my suggestion! and I, for one, like it a lot better.

matthewperiut:
image 

I removed the wrap element, and I adjusted the positioning of everything to still make it look nice, but not exactly perfectly aligned, but this removes an invasive modification.

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
matthewperiut: I removed the wrap operation from the code to avoid modifying the element.

matthewperiut:
I changed the size of the buttons, the search bar is not drawn any different in this iteration, as the code has been removed, only the buttons are modified.
image
image

matthewperiut:
What it looks like now, as of re-requesting review
Programmer art needed a slight modification
image  
image  
Here's various appearances of the regular button
image
image
image
image

williamistGitHub:
Hey, I really like the look of these new buttons (imo they're definitely better than the old ones) but I wanted to make a suggestion about the look of them when you hover them. The old buttons change the color of the background when you hover them, but the new buttons only change the color of the arrow which makes it less obvious that something's changed (especially if a mouse pointer is covering up some of it, it's a small button). My idea to improve this is to add a white overlay to the hovered state of the buttons, so that pretty much every pixel changes color noticeably on hover.

Here's a mockup I did of the idea, I think it works well:

image

The white overlay in that example is at 40% opacity, which I think isn't too low to be not noticeable, but not too high as to make the button unrecognizable.

Thanks again for making these new buttons by the way, they look so good.

Laxystem:
@williamistGitHub while I agree with the concept, I think the whitening should be done in the traditional minecraft way, outlines only. Having a white overlay on the entire thing looks like my screen's damaged.

matthewperiut:
The indentation is being lightened, which causes it to not match the other indents on the page. Additionally whitening around the edge is something I tried but it doesn't suit this format. I've decided that the way I have it is the best way to do it unless someone has a better idea and can send a screenshot, I think it's good as-is.

IMB11:
Is this going to get merged anytime soon? I think this is good as-is as well.

IMB11 (approved):

modmuss50 (approved):

PR #4223, opened 2024-11-14
Provide a RegistryByteBuf for attachment syncing

modmuss50:


fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -56,7 +56,7 @@
  * @param <A> type of the attached data. It is encouraged for this to be an immutable type.
  * @see AttachmentRegistry
  * @see AttachmentRegistry.Builder#persistent(Codec)
- * @see AttachmentRegistry.Builder#syncWith(PacketCodec, AttachmentSyncPredicate)
+ * @see AttachmentRegistry.Builder#syncedWith(PacketCodec, AttachmentSyncPredicate)
```
Juuxel: This is wrong/outdated

modmuss50: Thanks :)

Juuxel (approved):

Syst3ms (approved):

PR #4229, opened 2024-11-15
Add transitive access wideners for tracked data related creator entity methods

dvitski:
It seems desirable to have access to the tracked data fields of display entity classes.

Developers will have the option to call `Entity.dataTracker` methods with these `TrackedData` fields.

(I have not been able to find any other way to get/set the data values of these entities, however if there is a simpler way then I would love to know.)

As an alternative, it would be nice to have access to the getter/setter methods within the display entity classes, but I couldn't think of a clean generative way to achieve this.


fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -128,6 +130,24 @@ def generateRenderPhaseInnerClasses(List<String> lines, FileSystem fs) {
 	}
 }
 
+def generateTrackedDataFields(String className, List<String> lines, FileSystem fs) {
+	for (def field : loadClass(fs.getPath("${className}.class")).fields) {
+		// All protected static fields of RenderPhase
```
modmuss50: Comment is incorrect.

modmuss50 (commented):
Is exposing the fields the best way to go? It looks like there are getters and setters for most if all of these pieces of data. I think it would be better to TAW those.

dvitski:
I've hacked together a solution which checks for `dataTracker` accesses within get/set methods

1. Is getter/setter
2. If getter, has no parameters
3. If setter, has 1 parameter
4. Is private, is not static
5. Contains `dataTracker` field reference as get

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -128,16 +131,78 @@ def generateRenderPhaseInnerClasses(List<String> lines, FileSystem fs) {
...
+		new ClassVisitor(Opcodes.ASM9) {
+			@Override
+			MethodVisitor visitMethod(int access, String name, String descriptor, String signature, String[] exceptions) {
+				// check for desired methods
+				if (!name.startsWith("get") && !name.startsWith("set")) {
+					return null
+				}
+
+				// check methods as genuine basic getter/setters
+				long parameterCount = getMethodParameterCount(descriptor)
```
modmuss50: ASM has a helper for this [Type.getArgumentCount()](https://asm.ow2.io/javadoc/org/objectweb/asm/Type.html#getArgumentCount(java.lang.String))

modmuss50 (commented):
Great idea!

fabric-transitive-access-wideners-v1/build.gradle
dvitski: Ah I'll check that out

modmuss50 (approved):

PR #4231, opened 2024-11-15
Registry aliasing

Syst3ms:
Adds a small API to add aliases to registries, useful for when mods want to change IDs without losing compatibility with existing saves, and without having to resort to datafixer hacks.

The API just consists of a new `FabricRegistry` interface, implemented on all Registry instances via interface injection, with an `addAlias` method.


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -88,6 +91,19 @@ public abstract class SimpleRegistryMixin<T> implements MutableRegistry<T>, Rema
 	@Shadow
 	@Final
 	private Map<RegistryKey<T>, RegistryEntry.Reference<T>> keyToEntry;
+	@Shadow
+	@Final
+	private RegistryKey<? extends Registry<T>> key;
+	@Unique
+	private Event<RegistryEntryAddedCallback<T>> fabric_addObjectEvent;
```
modmuss50: Nit: no need for `fabric_` prefix when using @Unique.

modmuss50: Ah, scrolled down and seen this is existing code, so leave as is. Can be cleaned up in a seperate PR later if anyone cares enough.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -124,19 +132,21 @@ public Event<RegistryIdRemapCallback<T>> fabric_getRemapEvent() {
...
+						callback.onEntryAdded(rawId, id, object);
+					}
 				}
-			}
 		);
-		fabric_postRemapEvent = EventFactory.createArrayBacked(RegistryIdRemapCallback.class,
-			(callbacks) -> (a) -> {
-				for (RegistryIdRemapCallback<T> callback : callbacks) {
-					callback.onRemap(a);
+		fabric_postRemapEvent = EventFactory.createArrayBacked(
```
modmuss50: Has anything changed here? If not try not to reformat existing unrelated code.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+			throw new IllegalArgumentException(
+					"Tried adding %s as an alias for %s, but it is already an alias (for %s) in registry %s".formatted(
+							old,
+							newId,
+							fabric_aliases.get(old),
+							this.key
+					)
+			);
+		} else if (this.idToEntry.containsKey(old)) {
+			throw new IllegalArgumentException(
```
modmuss50, line 435: Question/thing to check: Make sure that it also throws if you try to register something after an alias.

fabric-registry-sync-v0/src/test/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
```diff
@@ -0,0 +1,92 @@
...
+	private static final Identifier NEW_ID = id("new");
+	private static final Identifier OTHER = id("other");
+
+	@BeforeAll
+	static void beforeAll() {
+		long time = System.currentTimeMillis();
+		SharedConstants.createGameVersion();
+		Bootstrap.initialize();
+
+		System.out.println("Bootstrap took " + (System.currentTimeMillis() - time) + "ms");
```
modmuss50: My fault for this, remove this timer, I was just intrested in how long this took and forgot to take it out.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+		if (fabric_aliases.containsKey(old)) {
+			throw new IllegalArgumentException(
+					"Tried adding %s as an alias for %s, but it is already an alias (for %s) in registry %s".formatted(
+							old,
+							newId,
+							fabric_aliases.get(old),
+							this.key
+					)
+			);
+		} else if (this.idToEntry.containsKey(old)) {
```
modmuss50: Codestyle nit: `else` is useless here. Just have seperate if statements.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+			// that doesn't already fall under the first condition
+			throw new IllegalArgumentException(
+					"Making %1$s an alias of %2$s would create a cycle, as %2$s is already an alias of %1$s (registry %3$s)".formatted(
+							old,
+							newId,
+							this.key
+					)
+			);
+		} else if (!this.idToEntry.containsKey(newId)) {
+			FABRIC_LOGGER.warn(
```
modmuss50, line 456: Question: Is there ever a use case where we want this?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+			alias = fabric_aliases.get(alias);
+		}
+
+		return alias;
+	}
+
+	@Unique
+	private RegistryKey<T> getAliased(RegistryKey<T> key) {
+		Identifier aliased = getAliased(key.getValue());
+		return aliased == key.getValue() ? key : RegistryKey.of(key.getRegistryRef(), aliased);
```
modmuss50: Question for perf experts: Is the extra allocation a peformance concern here? Would it be better if we also had a Map<RegistryKey, RegistryKey> ?

fabric-registry-sync-v0/src/test/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
```diff
@@ -0,0 +1,92 @@
...
+import net.minecraft.Bootstrap;
+import net.minecraft.SharedConstants;
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.registry.FabricRegistryBuilder;
+
+public class RegistryAliasTest {
+	private static final RegistryKey<Registry<String>> testRegistryKey = RegistryKey.ofRegistry(id("test"));
```
modmuss50: I know you copied this from me, but I've already had new ideas on how to test registry related things. Its up to you, feel free to ingore this:

In https://github.com/FabricMC/fabric/pull/4235/files I created a new registry with a random ID in each test. Thus the test has no side effects or depedencies. In your case here it might be better to make the registry none static and register it with a random in a beforeEach?

There is nothing wrong with what you have here but we are still finding new ideas and better ways to use unit testing for stuff such as this.

fabric-registry-sync-v0/src/test/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+		testRegistry = FabricRegistryBuilder.createSimple(testRegistryKey).buildAndRegister();
+
+		Registry.register(testRegistry, NEW_ID, "entry");
+		Registry.register(testRegistry, OTHER, "other");
+		testRegistry.addAlias(OBSOLETE_ID, NEW_ID);
+	}
+
+	private static Identifier id(String s) {
+		return Identifier.of("registry_sync_test", s);
```
modmuss50: Make this specific to this test class, see my newer registry related PR.

fabric-registry-sync-v0/src/test/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import net.minecraft.Bootstrap;
+import net.minecraft.SharedConstants;
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.registry.FabricRegistryBuilder;
+
+public class RegistryAliasTest {
```
modmuss50, line 39: I would also add a simple alais to a block or item in a testmod, maybe with a system prop or constant that can be changed to populate a world with the old item/block?

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+	}
+
+	@WrapMethod(method = "get(Lnet/minecraft/registry/RegistryKey;)Ljava/lang/Object;")
```
modmuss50: Defering to @LlamaLad7 to review these mixins, with a note that peformance is likely a concern here.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+		} else if (!this.idToEntry.containsKey(newId)) {
+			FABRIC_LOGGER.warn(
+					"Adding {} as an alias for {}, but the latter doesn't exist in registry {}",
+					old,
+					newId,
+					this.key
+			);
+		}
+
+		fabric_aliases.put(old, newId);
```
modmuss50: Nit: A debug log for each alias might be handy

modmuss50 (commented):
Looks like a great start, nothing drastic stands out.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
Syst3ms: No, nothing changed, I'll try to undo reformatting later

fabric-registry-sync-v0/src/test/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
Syst3ms: That sounds like a good idea, I'll take a look.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
Syst3ms: What do you mean? EDIT: nvm i understand, I'm not checking for the other way around rn

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
Syst3ms: I was thinking something like "the new id may not be in the registry, but if it is, we want aliasing", for example with dynamic registries? can go either way, it's not hard to check on the modder side if necessary

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +401,104 @@ public void unmap(String name) throws RemapException {
...
+			);
+		} else if (!this.idToEntry.containsKey(newId)) {
+			FABRIC_LOGGER.warn(
+					"Adding {} as an alias for {}, but the latter doesn't exist in registry {}",
+					old,
+					newId,
+					this.key
+			);
+		}
+
```
modmuss50, line 463: Was thinking last night, also add a call to `assertNotFrozen`

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/FabricRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+public interface FabricRegistry {
+	/**
+	 * Adds an alias for an entry in this registry. Once added, all queries to this registry that refer to the {@code old}
+	 * {@link Identifier} will be redirected towards {@code newId}. This is useful if a mod wants to change an ID without
+	 * breaking compatibility with existing worlds.
+	 * @param old the {@link Identifier} that will become an alias for {@code newId}
+	 * @param newId the {@link Identifier} for which {@code old} will become an alias
+	 */
+	default void addAlias(Identifier old, Identifier newId) {
+		// implemented via mixin
```
modmuss50: Nit: Throw, if you call this you expect it to work.

modmuss50 (approved):

fabric-registry-sync-v0/src/testmod/java/net/fabricmc/fabric/test/registry/sync/RegistryAliasTest.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.registry.Registry;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.RegistryKeys;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.api.ModInitializer;
+
+public class RegistryAliasTest implements ModInitializer {
+	private static final Logger LOGGER = LogUtils.getLogger();
+	public static final boolean USE_OLD_IDS = Boolean.parseBoolean(System.getProperty("fabric.registry.sync.test.alias.use_old_ids", "true"));
```
modmuss50, line 36: Nit: Boolean.getBoolean

modmuss50: Or not, this defaults to true unlike getBoolean.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
apple502j: `RegistryKey` is interned, so not a concern.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
Earthcomputer: It would probably be better to use `@ModifyVariable(at = @At("HEAD"), argsOnly = true)` to modify the parameter rather than wrapping the method.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
```diff
@@ -382,4 +414,112 @@ public void unmap(String name) throws RemapException {
...
+		aliases.put(old, newId);
+		FABRIC_LOGGER.debug("Adding alias {} for {} in registry {}", old, newId, this.key);
+	}
+
+	@Unique
+	private Identifier getAliased(Identifier id) {
+		Identifier alias = id;
+
+		while (aliases.containsKey(alias)) {
+			alias = aliases.get(alias);
```
Earthcomputer: You can use the fact that `get` returns `null` if the key is absent to avoid doing two map lookups. However I am thinking for performance reasons, the map should point directly to the final ID so you don't have to walk it. The hard work to "flatten" the map could be done in `addAlias` which hopefully doesn't get called as often.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/mixin/registry/sync/SimpleRegistryMixin.java
LlamaLad7: Sorry didn't get this ping for whatever reason, but I agree with Earth. It would also allow all the ones which have the same parameter type to be merged into 1 handler.

PR #4232, opened 2024-11-15
Fix includeData in pick events

apple502j:
Fixes an issue where the game mode check for `includeData` was omitted. Vanilla only allows `includeData` if the player is in Creative mode. Only affects the snapshots.

Also fixes several javadoc issues.


haykam821:
Is the include data flag set only for creative players by the client? If any player can request data to be included, I think the current behavior works well, just like how block interaction events are invoked for spectators. The parameter name and documentation could be changed to make the onus of authorization checking a bit more clear, though.

apple502j:
Spectators can interact with chests; vanilla `getPickedItem` is only called for Creative players.

modmuss50 (approved):

modmuss50:
Going to merge this, as it sems correct to me and I want to get a relaese out. Happy to discuss this more and change later if needed.

PR #4233, opened 2024-11-15
[1.21.4] Add optional registries & refactor registry sync

modmuss50:
- General code cleanup, a lot of stuff can now be made client only as we no longer persist ids.
  - There is scope to move even more (such as the registry remapping code) to the client, but ill leave this for later.
- Add some more unit tests :)
- Add optional registries

# Optional registries / stricter defaults

The goal of these changes is to document the expected behaviour of missing registry's when a client connects to a modded server. This behaviour is now consistent between a vanilla clients and modded clients. This has nothing to do with individual registry entries (yet).

Some mods have their own registries that only need to be synced when the connecting client has their mod, I have added a new `OPTIONAL` registry attribute that allows this. This attribute is sent in the registry sync packet.

The behaviour of a connecting client is as follows:

Screenshot 2024-11-15 at 20 25 17

These changes will only apply to 1.21.4 as they are breaking.

# TODO
- [x] Test with real clients and servers
- [x] Better error message when a registry is missing, I just reused the entries one for now?


fabric-registry-sync-v0/src/client/java/net/fabricmc/fabric/impl/client/registry/sync/ClientRegistrySyncHandler.java
```diff
@@ -0,0 +1,198 @@
...
+					// Registry was not found on the client, and is not optional.
+					missingEntries.put(registryId, new ArrayList<>(remoteRegistry.keySet()));
+				}
+
+				continue;
+			}
+
+			for (Identifier remoteId : remoteRegistry.keySet()) {
+				if (!registry.containsId(remoteId)) {
+					// Found a registry entry from the server that is
```
apple502j: that is... (missing?)

fabric-registry-sync-v0/src/client/java/net/fabricmc/fabric/impl/client/registry/sync/ClientRegistrySyncHandler.java
```diff
@@ -0,0 +1,198 @@
...
+			// If so we can just ignore it.
+			// Otherwise we throw an exception and disconnect.
+			if (registry == null) {
+				if (isRegistryOptional(registryId, data)) {
+					LOGGER.info("Received registry data for unknown optional registry: {}", registryId);
+					continue;
+				} else {
+					throw new RemapException("Received registry data for unknown registry: " + registryId);
+				}
+			}
```
apple502j: I believe this is duplicative, as `checkRemoteRemap` should've already thrown in this case?

apple502j (changes requested):

PR #4235, opened 2024-11-18
Add RegistryEntryAddedCallback.allEntries

modmuss50:
RegistryEntryAddedCallback.allEntries invokes a callback for all previous and future entries in a registry. This is a common bit of boiler plate that many mods use.

### DIFF
--- a/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/RegistryEntryAddedCallback.java
+++ b/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/RegistryEntryAddedCallback.java
@@ -54,6 +54,8 @@ public interface RegistryEntryAddedCallback<T> {
 	/**
 	 * Register a callback for all present and future entries in the registry.
 	 *
+	 * <p>Note: The callback is recursive and will be invoked for anything registered within the callback itself.
+	 *
 	 * @param registry the registry to listen to
 	 * @param consumer the callback that accepts a {@link RegistryEntry.Reference}
 	 */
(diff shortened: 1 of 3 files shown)


fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/RegistryEntryAddedCallback.java
```diff
@@ -16,17 +16,52 @@
...
+	 *
+	 * @param registry the registry to get the event for
+	 * @return the event
+	 */
 	static <T> Event<RegistryEntryAddedCallback<T>> event(Registry<T> registry) {
 		return ListenableRegistry.get(registry).fabric_getAddObjectEvent();
 	}
+
+	/**
+	 * Register a callback for all present and future entries in the registry.
```
Player3324, line 55: I think this should have a reentry/recursion warning - if a mod adds from within the consumer it'll see an invocation for itself

modmuss50: Yes, this was by design and I think correct/expected behaviour. Maybe just a note in the javadocs to point this out would be ok?

Player3324: Yes that's what I meant, the note references the Javadoc section but I should have specified.

modmuss50: Added a note about it. my bad I though you were suggesting to catch/prevent it.

Player3324 (approved):

apple502j (approved):

PR #4246, opened 2024-11-24
[1.21.1] Add Passive, Neutral, Hostile, and Boss entity tags.

crispytwig:
Thought it was silly that these did not exist in Minecraft nor Fabric's tags, so I think it would be a good idea to create generators for them based on the tags I've added here. 

As @haykam821 pointed out, these will need to have the generators edited, but I'm not familiar with data generation, so I'll close this and leave it up to someone else to decide if they'd like to contribute this feature.


crispytwig:
#c:bosses already exists, whoopsies

TelepathicGrunt:
A mob can change its hostile status based on conditions. Like I have a monster that will attack players unless they wear bee armor, then the mob is neutral to the player. And so forth. 

I do not think tags is usable for this as behavior is dynamic. Not static like tags wants

haykam821:
A mob that attacks based on armor would be neutral. For example, piglins are considered neutral.

haykam821 (changes requested):
From a technical standpoint, this pull request is incorrect; it modifies the generated directory directly.

Instead, the data generators should be updated and ran to produce the files currently included in this pull request. Fields should also be added to `ConventionalEntityTypeTags` so that mods can reference these tags without constructing their own tag keys.

PR #4247, opened 2024-11-24
Port FRAPI to 1.21.4 and refactor

PepperCode1:
- Port to 1.21.4
  - Change `glint` material property type from `TriState` to new `GlintMode` enum, due to addition of vanilla `ItemRenderState.Glint` enum
  - Remove `ItemStack` and `ModelTransformationMode` parameters from `emitItemQuads` to preserve contract of `ItemRenderState`
  - Implement `FabricBakedModel` methods on vanilla's `WrapperBakedModel`
- Apply refactors and breaking changes
  - Move `hasTransform`, `pushTransform`, and `popTransform` from `RenderContext` to `QuadEmitter` to make the system more flexible and allow mesh builder emitters to also have transforms
  - Remove `RenderContext` and inline remaining parameters to `emit*` methods; allows outputting model to arbitrary emitter, invoking model from arbitrary location, and having better control over early culling
  - Convert `MeshBuilder` to `MutableMesh`, which has more functionality necessary for advanced use cases
  - Add `Mesh#size`
  - Remove `RendererAccess` and add static `Renderer#get` and `#register` instead
  - Disallow `Renderer` retrieval from returning a null value
  - Set default vertex color to 0xFFFFFFFF instead of 0
  - Remove `QuadEmitter#hasTransform` as it is no longer necessary for its original purpose
  - Remove `disableColorIndex` material property
  - Rename `colorIndex` methods to `tintIndex` in `QuadView` and `MutableQuadView` to match https://github.com/FabricMC/yarn/commit/6d2cb315d002682c1a7124be1d0591f23c3c398b
  - Change `QuadView#faceNormal` to return a `Vector3fc` instead of a `Vector3f`
  - Remove `ForwardingBakedModel` as it is similar enough to vanilla's new `WrapperBakedModel`
  - Remove `WrapperBakedModel` as it was moved to Model Loading API v1 in #4243
  - Remove deprecated methods
- Change `MeshImpl` to use a static cursor pool to save on memory and to allow nesting access to cursor on a thread
- Disallow `MutableQuadViewImpl#material` from accepting a null material (in the API, the material parameter was already not nullable)
- Precompute default quad data buffer instead of using zeroed buffer and setting defaults every time in `MutableQuadViewImpl#clear`
- Optimize `MutableQuadViewImpl#copyFrom` by not precomputing geometry
- Remove "compatibility mode" from Indigo

## TODO

- [x] Allow `QuadTransform`s to say that they do not need to receive quads that will be culled in one way or another - "post-cull transforms". Utilizing this functionality in a custom model can significantly improve performance in some cases. **[May require a breaking change]**
  Resolution: Update documentation to make it clear that `emitter.hasTransforms()` does not need to be checked to perform early culling.
- [x] Add a `MutableMesh`/`QuadBuffer` which automatically grows and serves like a `List<MutableQuadView>`. This can make models that need to analyze submodels or perform 1:N transforms on submodels significantly more efficient, and make it easier to create such models. **[May require a breaking change]**
  Resolution: `MeshBuilder` was converted to `MutableMesh` with the necessary new methods.
- [x] Decide the fate of `WrapperBakedModel` and `ForwaringBakedModel` now that vanilla added its own `WrapperBakedModel`, which is very similar to `ForwardingBakedModel`. **[May require a breaking change]**
  Resolution: Remove `ForwardingBakedModel` and move `WrapperBakedModel` to Model Loading API with some changes.
- [x] Consider re-adding `ModelTransformationMode` parameter to `emitItemQuads` since it is an enum value and is passed to `ItemRenderer#renderItem` (where Indigo hooks into item rendering) anyway. **[May require a breaking change]**
  Resolution: Not re-added.
- [x] Consider removing `QuadEmitter#hasTransform` now that its initial purpose has been replaced by the `cullTest` parameter. **[May require a breaking change]**
  Resolution: Removed. Can be re-added later without a user-facing breaking change if necessary.
- [x] Consider exposing `MutableQuadView#clear` as public API.
  Resolution: Not exposed. Can be exposed later without a user-facing breaking change if necessary.
- [x] Finish porting FRAPI test mod and add new tests; Merging #4243 is a prerequisite.
- [x] Finish documentation.


modmuss50 (approved):
<3

PR #4256, opened 2024-11-29
Implement changes to client auto-test threading

Earthcomputer:
Implement changes to client auto-test threading as discussed on Discord. More detailed implementation notes can be found on the docs to `ThreadingImpl`.

I didn't put much thought into class naming, or what should or shouldn't be in the API, etc, as this is all still internal, and deciding on a nice clean API is left for future work.


fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/FabricApiAutoTestClient.java
```diff
@@ -136,18 +129,19 @@ private void runTest() {
...
 			waitForScreen(TitleScreen.class);
+			waitForServerStop();
 		}
 
 		try (var server = new TestDedicatedServer()) {
 			connectToServer(server);
 			waitForWorldTicks(5);
 
-			final GameProfile profile = submitAndWait(MinecraftClient::getGameProfile);
+			final GameProfile profile = computeOnClient(() -> MinecraftClient.getInstance().getGameProfile());
```
modmuss50: Is there a specific reason you decided not to pass the client though as a param to the lambda? What you have here isnt wrong, but my worry is this makes it possibly easier to call `MinecraftClient.getInstance()` on the wrong thread.

An option might be to just check the thread in `getInstance()`, but not something for now, at least until this becomes a public API. Ill take a proper look at the changes soon.

Earthcomputer: Honestly the only reason was that I didn't want to make more types of throwing functional interfaces. But it would definitely be convenient to have a `MinecraftClient` passed as the argument

Chocohead: > Honestly the only reason was that I didn't want to make more types of throwing functional interfaces.

There's already a wide range of options in [Apache Commons](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/function/package-summary.html). All the `Failable` interfaces match up to their normal Java equivalents.

Earthcomputer: I was looking for something like this, thanks!

Earthcomputer: > An option might be to just check the thread in `getInstance()`, but not something for now, at least until this becomes a public API. Ill take a proper look at the changes soon.

I wanted to do this but I thought it would be too controversial for this PR to be accepted, so I left the discussion for later. I can do it now if you want.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/mixin/MinecraftClientMixin.java
```diff
@@ -0,0 +1,142 @@
...
+				deferredTask.run();
+			}
+		}
+	}
+
+	@WrapMethod(method = "startIntegratedServer")
+	private void deferStartIntegratedServer(LevelStorage.Session session, ResourcePackManager dataPackManager, SaveLoader saveLoader, boolean newWorld, Operation<Void> original) {
+		if (FabricApiAutoTestClient.IS_AUTO_TEST && ThreadingImpl.taskToRun != null) {
+			// don't start the integrated server (which busywaits) inside a task
+			deferredTask = () -> original.call(session, dataPackManager, saveLoader, newWorld);
```
embeddedt: Using a MixinExtras `Operation` outside the scope of the injector function is undefined behavior according to LlamaLad7.

Earthcomputer: I thought about this but I couldn't find it documented anywhere and it worked so I left it as is. But now I found it searching back in Llama's Discord. Will change it.

LlamaLad7: The operation could have writes to bound LocalRefs which will throw if they occur after the handler returns

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/FabricClientTestHelper.java
```diff
@@ -182,41 +184,43 @@ public static void waitForTitleScreenFade() {
...
+		ThreadingImpl.runOnClient(() -> action.accept(MinecraftClient.getInstance()));
 	}
 
-	public static <T> T submitAndWait(Function<MinecraftClient, T> function) {
-		return submit(function).join();
+	public static <T, E extends Throwable> T computeOnClient(FailableFunction<MinecraftClient, T, E> action) throws E {
+		MutableObject<T> result = new MutableObject<>();
+		runOnClient(minecraft -> result.setValue(action.apply(minecraft)));
+		return result.getValue();
 	}
```
Alexander01998: Suggestion: Calling it `getFromClient()` instead of `computeOnClient()` would make it clearer that this method returns a value while `runOnClient()` doesn't.

Your use of `compute` here isn't wrong, I believe the Java API does something similar, but I think many people use `run` and `compute` synonymously and wouldn't immediately grasp that this is the difference between them. `get` vs `run` makes it really obvious.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/mixin/MinecraftClientMixin.java
```diff
@@ -0,0 +1,149 @@
...
+			postRunTasks(ci);
+		}
+	}
+
+	@Inject(method = "getInstance", at = @At("HEAD"))
+	private static void checkThreadOnGetInstance(CallbackInfoReturnable<MinecraftClient> cir) {
+		if (FabricApiAutoTestClient.IS_AUTO_TEST) {
+			Preconditions.checkState(Thread.currentThread() != ThreadingImpl.testThread, "MinecraftClient.getInstance() cannot be called from the test thread");
+		}
+	}
```
Alexander01998: This is a really nice addition that I'm sure will catch a lot of newbie mistakes. Maybe we should go one step further and also point them in the right direction. Something like:

> MinecraftClient.getInstance() cannot be called from the test thread. You probably want to use runOnClient() or computeOnClient() instead.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/TestDedicatedServer.java
```diff
@@ -51,25 +49,14 @@ public String getConnectionAddress() {
...
 	}
 
-	private <T> CompletableFuture<T> submit(Function<MinecraftDedicatedServer, T> function) {
-		return server.submit(() -> function.apply(server));
-	}
-
-	private <T> T submitAndWait(Function<MinecraftDedicatedServer, T> function) {
-		return submit(function).join();
-	}
-
```
Alexander01998, line 72: Why are there no `runOnServer()` / `computeOnServer()` helpers here? I mean, I can't say that I've used the old `TestDedicatedServer.submitAndWait()` in actual test code, but this seems like something that more server-heavy mods would find useful in their tests. Obviously one can run `ThreadingImpl.runOnServer()` directly instead, as you've done above, but it feels inconsistent to have these helpers on one side and not the other.

Alexander01998 (commented):
This looks really solid overall! 👍 I won't pretend to understand all of the multithreading logic, but we can see from the tests that it works and I can also see that you've put a lot of effort into making sure every edge case has safety checks included.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/TestDedicatedServer.java
Earthcomputer: They were private and unused so I deleted them, the equivalent is in `ThreadingImpl`. We can think about what helpers are useful to expose in a future PR

Alexander01998: OK.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/mixin/MinecraftClientMixin.java
Earthcomputer: I didn't want to add this yet because we don't know what the user-facing methods are going to be called

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/FabricClientTestHelper.java
Earthcomputer: I'm pretty sure this is the more common terminology, although I'm not super attached to it. But it should be immediately obvious what the difference between the two functions are from the parameter type

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/mixin/MinecraftClientMixin.java
Alexander01998: Yes, the method names might change later and would then need to be updated in this message, but I don't think the general concept of "lambda that runs stuff on the client thread" and "other lambda that does that and also returns a value" is going away any time soon.

The risk of forgetting to add this later seems worse. Remember, first-time modders / test writers are not used to the way this system works and may not even know that these methods exist. It's not obvious from the current code that the message is unfinished. Consider adding a `// TODO:` comment if you don't want to add the full message yet.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/FabricClientTestHelper.java
Alexander01998: I do like simple language, but yeah `compute` is pretty common too. I don't think parameter/return types are usually the first thing I would look at when working with a new API, but then again this will eventually have Javadoc that can make it obvious more quickly.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/ThreadingImpl.java
```diff
@@ -0,0 +1,203 @@
...
+
+	@Nullable
+	public static Thread testThread = null;
+	public static final Semaphore TEST_SEMAPHORE = new Semaphore(0);
+
+	@Nullable
+	public static Runnable taskToRun = null;
+
+	public static void enterPhase(int phase) {
+		while ((PHASER.getPhase() & PHASE_MASK) != phase) {
```
modmuss50, line 91: This is the only bit that is midly confusing to me (I think), is this method blocking until the current phase has been reached from somewhere else? My tottal lack of knowledge of Phaser doesnt help

modmuss50 (approved):
Ill be honest I dont full understand the impl, but the principle  is sound.

fabric-api-base/src/testmodClient/java/net/fabricmc/fabric/test/base/client/ThreadingImpl.java
Earthcomputer: This method blocks until the given tick phase has been reached on all threads registered to the `Phaser` ("parties"), while allowing other parties to advance through the other tick phases (which is what the loop is for). The `arriveAndAwaitAdvance` is a barrier that all parties must call before they are all unblocked at once.

A `Phaser` maintains a counter called the "phase" which increments each time a barrier is crossed. It starts at 0 and when it surpasses `Integer.MAX_VALUE` it resets back to 0 again (negative values happen when the `Phaser` is terminated, which is not functionality we're using). Using the lower 2 bits of this counter that already exists tells us which of our 4 tick phases we're currently in. It's possible to track that separately by subclassing the `Phaser` and overriding its `onAdvance` method, but I figured why not take advantage of a counter that already exists.

PR #4259, opened 2024-11-30
Make data generation .cache reproducible

modmuss50:
As far as I can tell this timestamp is never used, by setting it to a constant value Gradle will be able to consider skipping datageneration tasks if the input never changes. As datagen is an input to its self it does take 2 runs to be skipped.

Also see: https://github.com/FabricMC/fabric-loom/pull/1226

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/DataCacheCachedDataMixin.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/DataCacheCachedDataMixin.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/DataCacheCachedDataMixin.java
```diff
@@ -0,0 +1,50 @@
...
+@Mixin(targets = "net.minecraft.data.DataCache$CachedData")
+public abstract class DataCacheCachedDataMixin {
+	@WrapOperation(method = "write", at = @At(value = "INVOKE", target = "Lcom/google/common/collect/ImmutableMap;entrySet()Lcom/google/common/collect/ImmutableSet;"))
+	private ImmutableSet<Map.Entry<Path, HashCode>> sortPaths(ImmutableMap<Path, HashCode> instance, Operation<ImmutableSet<Map.Entry<Path, HashCode>>> original) {
+		return original.call(instance).stream()
+				.sorted(Map.Entry.comparingByKey(Comparator.comparing(k -> normalizePath(k.toString()))))
+				.collect(ImmutableSet.toImmutableSet());
+	}
+
+	@WrapOperation(method = "write", at = @At(value = "INVOKE", target = "Ljava/nio/file/Path;toString()Ljava/lang/String;"))
```
Earthcomputer: I think both of these could be `@ModifyExpressionValue`

modmuss50: Yes, thanks.

Earthcomputer (approved):

PR #4261, opened 2024-11-30
Add more data generation TAWs

modmuss50:
CrossType is needed for many BSMG methods.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.accesswidener
+++ b/fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.accesswidener
@@ -47,6 +47,7 @@ transitive-accessible method net/minecraft/client/data/TexturedModel makeFactory
 
 transitive-accessible class net/minecraft/client/data/BlockStateModelGenerator$BlockTexturePool
 transitive-accessible class net/minecraft/client/data/BlockStateModelGenerator$LogTexturePool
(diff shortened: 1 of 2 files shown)


fabric-data-generation-api-v1/src/main/resources/fabric-data-generation-api-v1.accesswidener
```diff
@@ -39,6 +39,10 @@ transitive-accessible method net/minecraft/data/family/BlockFamilies register (L
 
 transitive-accessible field net/minecraft/client/data/BlockStateModelGenerator blockStateCollector Ljava/util/function/Consumer;
 transitive-accessible field net/minecraft/client/data/BlockStateModelGenerator modelCollector Ljava/util/function/BiConsumer;
+transitive-accessible field net/minecraft/client/data/BlockStateModelGenerator itemModelOutput Lnet/minecraft/client/data/ItemModelOutput;
+
+accessible field net/minecraft/client/data/ItemModelGenerator output Lnet/minecraft/client/data/ItemModelOutput;
+accessible field net/minecraft/client/data/ItemModelGenerator modelCollector Ljava/util/function/BiConsumer;
```
SquidDev, line 45: I think these two lines are wrong. They should be `transitive-accessible`!

modmuss50: Uhh, yes thats correct.

PR #4269, opened 2024-12-03
Allow retrieving model loading plugins

PepperCode1:
This is accomplished though static `ModelLoadingPlugin#getAll` and `PreparableModelLoadingPlugin#getAll`. This functionality is necessary for alternative model loading implementations to support Model Loading API v1, such as the one Embeddedt's ModernFix uses to load models lazily.

### DIFF
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/PreparableModelLoadingPlugin.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/PreparableModelLoadingPlugin.java
@@ -16,9 +16,12 @@
 
 package net.fabricmc.fabric.api.client.model.loading.v1;
 
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Supplier;
+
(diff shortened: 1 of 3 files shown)


embeddedt (approved):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/PreparableModelLoadingPlugin.java
```diff
@@ -63,4 +74,11 @@ interface DataLoader<T> {
 		 */
 		CompletableFuture<T> load(ResourceManager resourceManager, Executor executor);
 	}
+
+	/**
+	 * Bundles a {@link PreparableModelLoadingPlugin} with its corresponding {@link DataLoader}
+	 * for retrieval through {@link #getAll()}.
+	 */
+	record Holder<T>(DataLoader<T> loader, PreparableModelLoadingPlugin<T> plugin) {
```
modmuss50: Generally we dont expose a record as API, I would suggest using an interface that has a `DataLoader<T> loader()` and `PreparableModelLoadingPlugin<T> plugin()`. You can then impliment that interface with a record like this one but in impl.

modmuss50 (approved):

PR #4275, opened 2024-12-05
[1.21.1] Add HudPreRenderCallback Event and HudPostRenderCallback.EVENT

JT122406:
Adds in a Pre Render Callback to allow developers to render things with the hud appearing over them similar to how carved pumpkins show the hud over the texture.  With this addition I decided to rename HudRenderCallback.EVENT to HudPostRenderCallback.EVENT and then Deprecate HudRenderCallback and point devs to using the new Post one.


Linguardium:
would probably make more sense here to interact with the LayeredDrawer to allow adding layers or add new layers to do so

kevinthegreat1:
There is already an open pr for hud render events. See #4119. I’m happy to discuss this further in the fabric discord.

JT122406:
> There is already an open pr for hud render events. See #4119. I’m happy to discuss this further in the fabric discord.

My bad didn't even see that when I was looking through

PR #4276, opened 2024-12-05
Fix custom ingredient implementation

### LINKED ISSUE #4270: CustomIngredient broken equality
CustomIngredient instances (the vanilla Ingredient wrapped ones) dont have a custom equality implementation. This can cause all sorts of issues as a custom igredient will be considered equal to an empty one

PepperCode1:
- Override `isEmpty` in `CustomIngredientImpl`
- Override `acceptsItem` in `CustomIngredientImpl`
- Override `equals` and `hashCode` in `CustomIngredientImpl` so `CustomIngredient`s can also implement them
  - Adjust the class doc of `CustomIngredient` to encourage implementors to do so
- Implement `Ingredient#hashCode`
- Remove incorrect `Nullable` annotation from `ItemStack` parameter in `CustomIngredientImpl#test`
- Simplify default implementation of `CustomIngredient#toDisplay` as created `RegistryEntryList` would always contain right storage
- Implement `equals` and `hashCode` on all builtin custom ingredient classes

Fixes #4270.


Technici4n (approved):
Looks good at a very quick glance.

modmuss50 (approved):

apple502j (approved):

PR #4280, opened 2024-12-05
Update checkstyle

modmuss50:
Fixes some fishy html tags that now fail.


fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/package-info.java
```diff
@@ -17,37 +17,38 @@
 /**
  * The Resource Loader, version 0.
  *
- * <p><h3>Quick note about vocabulary in Minecraft:</h3></p>
+ * <h3>Quick note about vocabulary in Minecraft:</h3>
+ *
  * <p>Resource Pack refers to both client-sided resource pack and data pack.</p>
```
Juuxel: Should the end tag be here either?

modmuss50: No, the whole tag isnt needed at all. Removed.

PR #4286, opened 2024-12-08
Fix block outline rendering

### LINKED ISSUE #4285: The High Contrast Block Outline is broken with Fabric API
Fabric Loader: 0.16.9
Fabric API: 0.111.0+1.21.4
Minecraft: 1.21.4

## Actual
When the "high contrast block outlines" accessibility option is enabled and fabric api installed, only the blue line is visible

## Expected
There should be a blue line as well as a thicker black line, like there is in vanilla and fabric loader (without fabric api)

## Additional
As seen in the following screenshot, the black line is still visible if the thickness of the blue line is decreased
image

PepperCode1:
- Move outline event invoker to before either normal or high contrast outlines are rendered
- Remove outline layer reset, fixing high contrast outline being rendered with wrong layer
- Pass translucent parameter as `WorldRenderContext#translucentBlockOutline`, usable in the `BEFORE_BLOCK_OUTLINE` and `BLOCK_OUTLINE` events

Fixes #4285.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -44,12 +44,6 @@ public interface WorldRenderContext {
 	 */
 	WorldRenderer worldRenderer();
 
-	/**
-	 * The matrix stack is only not null in {@link WorldRenderEvents#AFTER_ENTITIES} or later events.
-	 */
-	@Nullable
-	MatrixStack matrixStack();
-
 	RenderTickCounter tickCounter();
 
 	boolean blockOutlines();
@@ -58,9 +52,9 @@ public interface WorldRenderContext {
 
 	GameRenderer gameRenderer();
 
-	Matrix4f projectionMatrix();
-
 	Matrix4f positionMatrix();
+
+	Matrix4f projectionMatrix();
 
 	/**
 	 * Convenient access to {WorldRenderer.world}.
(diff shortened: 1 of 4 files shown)


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
```diff
@@ -110,12 +109,6 @@ public interface WorldRenderContext {
 	 * {@code WorldRenderer.drawBlockOutline}.
 	 */
 	interface BlockOutlineContext {
-		/**
-		 * @deprecated Use {@link #consumers()} directly.
-		 */
-		@Deprecated
-		VertexConsumer vertexConsumer();
```
modmuss50, line 117: No, we cannot remove API mid version, can this be kept around (even if somewhat broken?).

modmuss50 (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderEvents.java
```diff
@@ -279,13 +279,15 @@ public interface BeforeBlockOutline {
...
+		 * @param translucent If {@code true}, current block outline is being rendered after translucent terrain.
+		 *                    Otherwise, it is being rendered after solid terrain.
 		 * @param hitResult The game object currently under the crosshair target.
 		 * Normally equivalent to {@link MinecraftClient#crosshairTarget}. Provided for convenience.
 		 * @return true if vanilla block outline rendering should happen.
 		 * Returning false prevents {@link WorldRenderEvents#BLOCK_OUTLINE} from invoking
 		 * and also skips the vanilla block outline render, but has no effect on other subscribers to this event.
 		 */
-		boolean beforeBlockOutline(WorldRenderContext context, @Nullable HitResult hitResult);
+		boolean beforeBlockOutline(WorldRenderContext context, boolean translucent, @Nullable HitResult hitResult);
```
modmuss50: Sorry, I missed this earlier. This is also a breaking change. You could you maybe pass this in via the context (not ideal but at least is not breaking), otherwise it needs a whole new event.

Juuxel: Or another method in the interface delegating to the old one by default, but implementing it would be clunky.

PepperCode1: `BEFORE_BLOCK_OUTLINE_V2` and `BeforeBlockOutlineV2`? Are there better names?

modmuss50: I think the easy option is to add it to the context as `translucentBlockOutline`, and maybe document its designed for this event.

apple502j:
Was about to say I think it's fine to break this API now - however turns out the API is much more frequently used than I thought.

modmuss50 (approved):

PR #4301, opened 2024-12-14
Ensure that there are no duplicate files across projects.

modmuss50:
Prevents https://github.com/FabricMC/fabric/pull/4300 from happening again, depends on that being merged first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -744,18 +744,41 @@ dependencies {
 	}
 }
 
-remapJar {
-	afterEvaluate {
-		subprojects.each {
-			if (it.name in devOnlyModules || metaProjects.contains(it.name)) {
-				return
-			}
-
-			// Include the signed or none signed jar from the sub project.
-			nestedJars.from project("${it.path}").tasks.getByName("signRemapJar")
-		}
+configurations {
+	nestedJars {
+		transitive = false
 	}
 }
+
+dependencies {
+	subprojects.each {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -744,19 +744,42 @@ dependencies {
...
+	from {
+		configurations.nestedJars.files.collect { zipTree(it) }
 	}
+
+	// We expect these files to be duplicated, so exclude them.
+	exclude 'META-INF/**'
+	exclude 'fabric.mod.json'
 }
 
+test.dependsOn "checkNoDuplicateFiles"
```
Juuxel: `check` might be a better parent task since this is separate from running the code tests

build.gradle
```diff
@@ -744,19 +744,42 @@ dependencies {
...
+		nestedJars project("${it.path}")
+	}
+}
+
+remapJar {
+	nestedJars.from configurations.nestedJars
+}
+
+// Attempt to create a single jar with all files from all nested jars, this will fail if there are duplicate files.
+tasks.register("checkNoDuplicateFiles", Zip) {
```
Juuxel, line 768: This could be a `Copy` too, but I guess this saves some disk space since the files aren't even used

modmuss50: Yes it could be a `Sync`, However I think Zip is likely best, if I had to guess zip is likely faster than dealing with 1000s of small files. In the end it doesnt really matter.

PR #4302, opened 2024-12-15
Allow setting `canPotentiallyExecuteCommands` in BE/E type builders

PepperCode1:
- Add `FabricBlockEntityTypeBuilder#canPotentiallyExecuteCommands`
- Add `FabricEntityType.Builder#canPotentiallyExecuteCommands`
- Add `FabricBlockEntityTypeBuilder#addBlocks(Collection)`
- Deprecate `FabricBlockEntityTypeBuilder#build(Type)`

### DIFF
--- a/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
+++ b/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
@@ -52,6 +52,16 @@ public interface FabricEntityType {
 		}
 
 		/**
+		 * Sets whether the entity is able to execute commands.
+		 *
+		 * @param canPotentiallyExecuteCommands whether the entity is able to execute commands
+		 * @return this builder
+		 */
+		default EntityType.Builder<T> canPotentiallyExecuteCommands(boolean canPotentiallyExecuteCommands) {
+			throw new AssertionError("Implemented in Mixin");
+		}
+
+		/**
 		 * Creates an entity type builder for a living entity.
 		 *
 		 * <p>This entity's spawn group will automatically be set to {@link SpawnGroup#MISC}.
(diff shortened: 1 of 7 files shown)


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
```diff
@@ -51,6 +51,16 @@ default EntityType.Builder<T> alwaysUpdateVelocity(boolean alwaysUpdateVelocity)
 			throw new AssertionError("Implemented in Mixin");
 		}
 
+		/**
+		 * Sets whether the entity is able to execute commands.
+		 *
+		 * @param canPotentiallyExecuteCommands whether the entity is able to execute commands
+		 * @return this builder
+		 */
+		default EntityType.Builder<T> canPotentiallyExecuteCommands(boolean canPotentiallyExecuteCommands) {
```
modmuss50, line 60: Nit: Should the 2 methods in either builder be the same? This one takes the boolean, while the block entity type is just `canPotentiallyExecuteCommands()` without the bool arg?

PepperCode1: The method in the entity type builder takes a `Boolean` because the other method in this interface also takes a `Boolean`. I suppose the block entity type builder can also take this parameter.

modmuss50 (approved):

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricEntityType.java
haykam821: Should both ways be available on both classes then? The methods setting the value to `true` could be default methods for both classes.

modmuss50: It does feel a little odd to add two diffrent methods that should do the same thing. I dont have a peference either way but I think it would be good for them to match.

modmuss50: Ive added the bool to the block entity type builder.

PR #4318, opened 2024-12-21
Add various lifetime-bound try-with-resources APIs to client gametests

Earthcomputer:
See changes to `ClientGameTestTest` for an overview of the changes to the API.


fabric-client-gametest-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/client/gametest/ClientGameTestTest.java
```diff
@@ -16,141 +16,85 @@
...
+				context.takeScreenshot("server_in_game", 0);
+
+				{ // Test that we can enter and exit configuration
+					final GameProfile profile = context.computeOnClient(MinecraftClient::getGameProfile);
+					server.runCommand("debugconfig config " + profile.getName());
+					context.waitForScreen(ReconfiguringScreen.class);
+					context.takeScreenshot("server_config");
+					server.runCommand("debugconfig unconfig " + profile.getId());
+					// TODO: better way to wait for reconfiguration to end
+					context.waitTicks(100);
```
Earthcomputer, line 95: Not having this `waitTicks(100)` here breaks the assertion in `TestServerConnection.close()` because reconfiguration sets `MinecraftClient.world` to `null`. Ideally I want the `close()` method to wait for any configuration to finish before trying to disconnect. I [tried](https://discord.com/channels/507304429255393322/566276937035546624/1319691385598640168) to fix this for this PR, but I didn't manage to figure out a good way to fix it, so I've left the problem for later. I imagine it's a pretty rare case anyway.

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/api/client/gametest/v1/TestServerContext.java
```diff
@@ -0,0 +1,59 @@
...
+ * <p>Functions in this class can only be called on the client gametest thread.
+ */
+@ApiStatus.NonExtendable
+public interface TestServerContext {
+	/**
+	 * Runs a command on the server.
+	 *
+	 * @param command The command to run
+	 */
+	void runCommand(String command);
```
modmuss50, line 38: A thing for the future, but getting the result of the command might be handy, not quite how it would work but I could see it being useful.

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/DedicatedServerImplUtil.java
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+
+	private static void setupServer(Properties customServerProperties) {
+		Properties serverProperties = new Properties();
+		serverProperties.putAll(DEFAULT_SERVER_PROPERTIES);
+		serverProperties.putAll(customServerProperties);
+
+		try {
+			Files.writeString(Path.of("eula.txt"), "eula=true");
```
modmuss50: Ive always been hesitent about accepting the eula for people, maybe it would be best to skip the check via a mixin thats only appied when testing. This way we dont accept it for someone if they start a none test server in the same dir. I'm likely being overly pedantic about this, but I think its best to be safe.

fabric-client-gametest-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/client/gametest/ClientGameTestTest.java
modmuss50: Yeah, could you maybe wait until the world has rendered? I dont think it matters a vast amount, as this test is something that I dont expect many if anyone to run. I think I added it after we broke reconfiguration, im not even sure the debugconfig command is registered by default.

modmuss50 (approved):
This is really great, my only slight worry is around agreeing to eula for people.

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/api/client/gametest/v1/TestServerContext.java
Earthcomputer: This function was just copied over from `TestDedicatedServer`. But yes, could be handy. Annoyingly the current function it's implemented with doesn't return the result of the command but it should hopefully be a case of finding a different function to use.

PR #4321, opened 2024-12-23
Add `UnbakedModelDeserializer`

PepperCode1:
This pull request introduces the `UnbakedModelDeserializer` interface, which allows overriding the deserialization process of a JSON model file based on its `fabric:type` JSON key.

### Points to Discuss

- [x] Should optional types be allowed? That is, if a type is declared as optional in the JSON file and the type does not exist, the file will be loaded as if no type were declared instead of failing to load. I am not familiar with how Fabric API load conditions work or whether they are used when deserializing models, but if they are, then they could be an alternative approach to this idea.
  Resolution: Yes. The behavior and syntax of this feature matches those of NeoForge and is explained in the documentation of this pull request. Resource conditions may in theory work as a functionally equivalent alternative but would be significantly less convenient to use, requiring the creation of multiple files per model.
- [x] Should utilities to deserialize vanilla model elements (`elements`, `parent`, `textures`, `ambientocclusion`, `display`, `gui_light`) be provided? If so, how?
  Resolution: They will not be included in this pull request but may be added in a later one.
- [x] Should data generation classes be extended to make it easier to generate models of custom types? If so, how?
  Resolution: They will not be included in this pull request but may be added in a later one.
- [x] Which default custom types should be registered by Fabric API, if any? NeoForge provides a composite model type and an OBJ model type.
  Resolution: None will be included in this pull request but some may be added in a later one.
- [x] Should `UnbakedModelDeserializer` have a method to retrieve its `Identifier` like `CustomIngredientSerializer` instead of the ID being passed to the `register` method directly?
  Resolution: No; not necessary.
- [x] Should `UnbakedModelDeserializer` be generic? It represents a "type" which can create instances, and such classes are almost always generic both in Fabric API and vanilla.
  Resolution: No; not necessary.
- [x] Are there better names for certain things?
  Resolution: Only naming of the single added API class, `UnbakedModelDeserializer`, and its members matters. These names are already good. Class and member names in the implementation can be changed at any later time.

### TODO

- [x] Documentation
- [x] Optional types

### DIFF
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
@@ -63,8 +63,8 @@ public interface UnbakedModelDeserializer {
 	}
 
 	/**
-	 * @return the custom model deserializer registered with the given identifier, or {@code null} if there is no such
-	 * deserializer
+	 * {@return the custom model deserializer registered with the given identifier, or {@code null} if there is no such
+	 * deserializer}
 	 */
 	@Nullable
 	static UnbakedModelDeserializer get(Identifier id) {
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.model.loading.v1;
+
+import java.io.Reader;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.render.model.UnbakedModel;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.client.model.loading.UnbakedModelDeserializerRegistry;
+
+public interface UnbakedModelDeserializer {
+	static void register(Identifier id, UnbakedModelDeserializer deserializer) {
+		UnbakedModelDeserializerRegistry.register(id, deserializer);
+	}
+
+	@Nullable
+	static UnbakedModelDeserializer get(Identifier id) {
+		return UnbakedModelDeserializerRegistry.get(id);
+	}
+
+	static UnbakedModel deserialize(Reader reader) throws JsonParseException {
+		return UnbakedModelDeserializerRegistry.deserialize(reader);
+	}
+
+	UnbakedModel deserialize(JsonObject jsonObject, JsonDeserializationContext context) throws RuntimeException;
+}
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
@@ -24,23 +24,77 @@ import com.google.gson.JsonParseException;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.client.render.model.UnbakedModel;
+import net.minecraft.client.render.model.json.JsonUnbakedModel;
+import net.minecraft.client.render.model.json.ModelElement;
+import net.minecraft.client.render.model.json.ModelElementFace;
+import net.minecraft.client.render.model.json.ModelElementTexture;
+import net.minecraft.client.render.model.json.ModelTransformation;
+import net.minecraft.client.render.model.json.Transformation;
 import net.minecraft.util.Identifier;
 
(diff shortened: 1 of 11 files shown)


cputnam-a11y:
Is load conditions the same thing as resource conditions? if so, https://github.com/FabricMC/fabric/blob/1.21.4/fabric-resource-conditions-api-v1/src/main/java/net/fabricmc/fabric/impl/resource/conditions/ResourceConditionsImpl.java#L66-L87 is the implentation. If a type is not specified, the resource is loaded as normal, as if all conditions passed

Technici4n:
- optional types: yes, they are useful
- utilities: only if we need them in FAPI for now, otherwise defer for later
- datagen: don't think about it for now
- custom types: can come later 
- how to pass the identifier: it does not really matter. Passing it when registering is closer to how vanilla normally registers things.
- should the deserializer be generic? It doesn't really matter I think.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.render.model.UnbakedModel;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.client.model.loading.UnbakedModelDeserializerRegistry;
+
+public interface UnbakedModelDeserializer {
```
Technici4n, line 55: Missing javadoc

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+	@Nullable
+	static UnbakedModelDeserializer get(Identifier id) {
+		return UnbakedModelDeserializerRegistry.get(id);
+	}
+
+	static UnbakedModel deserialize(Reader reader) throws JsonParseException {
+		return UnbakedModelDeserializerRegistry.deserialize(reader);
+	}
+
+	UnbakedModel deserialize(JsonObject jsonObject, JsonDeserializationContext context) throws RuntimeException;
```
Technici4n: Missing javadoc, especially with regards to the `context`.

Technici4n: Why RuntimeException rather than a Gson one?

PepperCode1: It is acceptable for the deserializer to throw any `RuntimeException`, such as `InvalidIdentifierException`. The vanilla `JsonUnbakedModel` deserializer does that.

XFactHD:
- Optional types: yes, they are very useful if the basic model can be loaded by vanilla and the custom loader just enhances the model
- Deserialization utilities: something like Neo's [StandardModelParameters](https://github.com/neoforged/NeoForge/blob/e082c32c532199cd2d803c5d05fee765906df854/src/main/java/net/neoforged/neoforge/client/model/StandardModelParameters.java) could certainly be useful but I don't think it should be a blocker for this PR
- Datagen: I don't know to which degree Fabric extends model datagen for custom models beyond parent and textures, so I can't really comment on this
- Custom types: better suited for subsequent PRs IMO
- Identifier lookup: I can only see this being useful for datagen and for that it should be perfectly fine to make the registry a `BiMap` and look up the key from that (assuming of course that the deserializer is even involved in datagen at all)
- Generics: I don't think there is much value in a generic type in this case

Apart from optional types and the docs TODO, the implementation looks good to me.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
PepperCode1: The `throws` clause has been removed and replaced with a note in the method javadoc stating that the method is allowed and encouraged to throw exceptions as they will be caught by the caller.

modmuss50 (approved):
Looks great, nice and simple API surface.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
```diff
@@ -0,0 +1,105 @@
...
+	 *
+	 * @throws IllegalArgumentException if the deserializer is already registered
+	 */
+	static void register(Identifier id, UnbakedModelDeserializer deserializer) {
+		UnbakedModelDeserializerRegistry.register(id, deserializer);
+	}
+
+	/**
+	 * @return the custom model deserializer registered with the given identifier, or {@code null} if there is no such
+	 * deserializer
```
Juuxel:
```suggestion
	 * {@return the custom model deserializer registered with the given identifier, or {@code null} if there is no such
	 * deserializer}
```

Juuxel (approved):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/UnbakedModelDeserializer.java
PepperCode1: What is the purpose of this change?

Juuxel: It renders in a nicer way. From the [spec](https://docs.oracle.com/en/java/javase/23/docs/specs/javadoc/doc-comment-spec.html#return):
> As an inline tag, provides content for the first sentence of a method's main description, and a "Returns" section, as if "`@return description`" were also present. In the default English locale, the first sentence is "Returns *description*.".

I don't think there's any description for the method in the method list if you use the block tag form (`@return` without braces).

Technici4n: Only `@return` with no text before means that there won't be any text in the summary of the method. I personally prefer not using any tag in that case.

Juuxel: The whole point of the inline `{@return}` is that it also adds a short method summary ("Returns ....") as well as the return value section. IMO only `@return` < no tag ≤ only `{@return}`, but this is very bikesheddy and anything works in practice.

Issue #4170, opened 2024-10-16
Issues with fabric:components custom ingredient type


vico93:
I think i'm the only one that uses the very niche `"fabric:type": "fabric:components"` [custom ingredient](https://github.com/FabricMC/fabric/blob/1.21.1/fabric-recipe-api-v1/src/testmod/resources/data/fabric-recipe-api-v1-testmod/recipe/test_shapeless_match.json) that Fabric introduces for .json recipes to fine-grained ingredient picking (basically, select only item with specific components - or NBT for older versions).

I did a recipe to make mud with a dirt block and a water bottle with the following .json:

```json
{
    "type": "minecraft:crafting_shapeless",
    "ingredients": [
        {
            "item": "minecraft:dirt"
        },
		{
			"fabric:type": "fabric:components",
			"base": {
				"item": "minecraft:potion"
			},
			"components": {
				"minecraft:potion_contents": {
					"potion": "minecraft:water"
				}
			},
		  "strict": true
		}
    ],
    "result": {
        "id": "minecraft:mud",
        "count": 1
    }
}
```

Although it worked - shows in REI and when i manually put the dirt and the water bottle on the crafting it outputs the mud block - i noticed some issues with this recipe:

First thing is that the water potion isn't pulled to the crafting table (while the other "normal" ingredient - dirt - goes normally) when the recipe is select with either the vanilla recipe book or REI:
![2024-10-16_15 04 07](https://github.com/user-attachments/assets/680f01f3-b750-4f7f-a79a-0c51b78b5614)
While the recipe is shown normally on REI's recipe preview (but clicking on + gives the same issue of the recipe book):
![2024-10-16_15 04 13](https://github.com/user-attachments/assets/69ebf998-e415-4713-a8e4-ef32ae3ad67d)

The other thing is that, at least with water bottles, an empty bottle isn't returned to the inventory when the crafting is done - similar to how crafting with water/lava/milk buckets works. Maybe adding another field to specify the "return" item could be a solution to achieve consistency between bottle and bucket recipes, or just defining internally that filled bottles should return empty ones would be enough.

MC 1.21.1, Fabric 0.16.7 and Fabric API 0.106.0+1.21.1

modmuss50:
Hi, this sounds like a valid problem, I do wonder how this behaves in 1.21.2 as the recipe book works quite diffrently now.

I expect REI has its own logic for this so might be worth seeing if anyone has had this issue before over there?

[...]

PR #4322, opened 2024-12-23
[1.21.1-1.21.4] Custom Ingredients sync fix

### LINKED ISSUE #4225: [1.21.1-1.21.4] CustomIngredientPacketCodec always falls back.
**Fabric Loader:** 0.16.9
**Fabric API:** 0.108.0+1.21.1

Hello, I've found a bit of an edge case with the class I mentioned in the title.

Whenever a custom ingredient is attempted to be synced, it will always fall back.

I think that this might be because of the current supported ingredients set being cleared unconditionally. https://github.com/FabricMC/fabric/blob/5197d62800bbd09169ced6c73bc45767a028cd41/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/EncoderHandlerMixin.java#L57-L76

Salandora:
Fixes #4225.
The synced ```fabric_supportedCustomIngredients``` list was lost due to minecraft removing the ```EncoderHandler``` before joining the world.
Therefore all recipes were synced with the fallback method instead.
This PR fixes that issue by storing the list in the ```ClientConnection``` class instead and making use of the ```packet_handler``` pipeline which points to the ```ClientConnection``` instance.

@MerchantPug could you maybe test this fix against your implementation?

### DIFF
--- a/fabric-recipe-api-v1/build.gradle
+++ b/fabric-recipe-api-v1/build.gradle
@@ -5,9 +5,9 @@ loom {
 }
 
 moduleDependencies(project, [
-	"fabric-networking-api-v1",
+	'fabric-networking-api-v1',
 ])
 
 testDependencies(project, [
-		':fabric-lifecycle-events-v1',
+	':fabric-lifecycle-events-v1',
 ])
--- a/fabric-recipe-api-v1/build.gradle
+++ b/fabric-recipe-api-v1/build.gradle
@@ -7,3 +7,7 @@ loom {
 moduleDependencies(project, [
 	"fabric-networking-api-v1",
 ])
+
+testDependencies(project, [
+		':fabric-lifecycle-events-v1',
+])
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/SupportedIngredientsPacketEncoder.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/SupportedIngredientsPacketEncoder.java
@@ -26,4 +26,6 @@ import net.minecraft.util.Identifier;
  */
 public interface SupportedIngredientsPacketEncoder {
 	void fabric_setSupportedCustomIngredients(Set<Identifier> supportedCustomIngredients);
+
+	Set<Identifier> fabric_getSupportedCustomIngredients();
 }
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/EncoderHandlerMixin.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/EncoderHandlerMixin.java
@@ -19,14 +19,16 @@ package net.fabricmc.fabric.mixin.recipe.ingredient;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
-import net.fabricmc.fabric.impl.recipe.ingredient.CustomIngredientSync;
-import net.fabricmc.fabric.impl.recipe.ingredient.SupportedIngredientsPacketEncoder;
-import net.minecraft.network.handler.EncoderHandler;
-import net.minecraft.network.packet.Packet;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.network.handler.EncoderHandler;
+import net.minecraft.network.packet.Packet;
+
+import net.fabricmc.fabric.impl.recipe.ingredient.CustomIngredientSync;
+import net.fabricmc.fabric.impl.recipe.ingredient.SupportedIngredientsPacketEncoder;
 
 @Mixin(EncoderHandler.class)
 public class EncoderHandlerMixin {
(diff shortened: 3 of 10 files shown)


modmuss50:
Hi, got a few questions: is there a reason why this targets 1.21.1 and not 1.21.4? Could this be a backport of https://github.com/FabricMC/fabric/pull/4287 instead? Ill have to have a proper look with the code in front of me as well.

Salandora:
I targeted 1.21.1 because my Sophisticated Storage port is still on 1.21.1 so I would need the fix there. I want to make 2 more PRs against .3 and .4 just wanted to get the okay first that the changes are okay and not do the work twice.

#4287 is unfortunately not a fix to the issue, I tried to explain my findings in #4225.

I can look into backporting #4276 and what applies from #4287 and either add them here or make a new PR for it. 
The thing with #4287 is that in 1.21.1 for example there is no ```Ingredient.OPTIONAL_PACKET_CODEC``` so all of that part does not apply.

Yeah please take your time, I would like to have MerchantPug confirm the fix works for them too so we can be sure that there isn't a second bug.

fabric-recipe-api-v1/build.gradle
```diff
@@ -7,3 +7,7 @@ loom {
 moduleDependencies(project, [
 	"fabric-networking-api-v1",
 ])
+
+testDependencies(project, [
+		':fabric-lifecycle-events-v1',
+])
```
PepperCode1: Nit: Files should end with a newline.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/SupportedIngredientsPacketEncoder.java
```diff
@@ -26,4 +26,6 @@
  */
 public interface SupportedIngredientsPacketEncoder {
```
PepperCode1: This interface should be renamed to `SupportedIngredientsClientConnection` to match the name of the class it's actually implemented on.

fabric-recipe-api-v1/src/main/resources/fabric-recipe-api-v1.mixins.json
PepperCode1, line 1: It's not necessary to change the indentation of the whole file. All Fabric API mixin json files use 2 spaces for indentation.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/EncoderHandlerMixin.java
```diff
@@ -51,7 +40,11 @@ public void fabric_setSupportedCustomIngredients(Set<Identifier> supportedCustom
 			method = "encode(Lio/netty/channel/ChannelHandlerContext;Lnet/minecraft/network/packet/Packet;Lio/netty/buffer/ByteBuf;)V"
 	)
 	private void capturePacketEncoder(ChannelHandlerContext channelHandlerContext, Packet<?> packet, ByteBuf byteBuf, CallbackInfo ci) {
-		CustomIngredientSync.CURRENT_SUPPORTED_INGREDIENTS.set(fabric_supportedCustomIngredients);
+		ChannelHandler client = channelHandlerContext.pipeline().get("packet_handler");
```
PepperCode1: Nit: `client` as a variable name here doesn't make sense to me. I think a simple `channelHandler` would be fine.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/ingredient/EncoderHandlerMixin.java
```diff
@@ -51,7 +40,11 @@ public void fabric_setSupportedCustomIngredients(Set<Identifier> supportedCustom
...
+
+		if (client != null) {
```
PepperCode1: It might be better to replace this check with `client instanceof SupportedIngredientsPacketEncoder`.

PepperCode1 (changes requested):

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/ingredient/SupportedIngredientsPacketEncoder.java
Salandora: Will change that and also fix the comment as it was still on EncoderHandler

fabric-recipe-api-v1/build.gradle
Salandora: Will add the newline and also fix the indentation and quotes

fabric-recipe-api-v1/src/main/resources/fabric-recipe-api-v1.mixins.json
Salandora: Ups, will undo that

ChrysanthCow:
Will test this soonish.

ChrysanthCow:
## Before PR
Before

## After PR
After

Looks like it works for me! :+1:

PepperCode1 (approved):

modmuss50 (approved):

vico93:
dumb question but does this changes the behavior i reported in #4170 in any way?

Salandora:
@vico93 yes and no, the recipe book will still not work with this.

But your client side is now aware of custom ingredients again meaning that if REI has support for custom ingredients it should work but from my own testings with REI I at least need to write a custom TransferHandler, unfortunately I'm not sure if this is required for you too.

EMI is a bit nicer in that regard as I only needed to override the recipe and let EMI know that it needs to use a ListIngredient but for that I need to know about the current recipe and its custom ingredient on the client side,

PR #4323, opened 2024-12-23
Biomemodifications: Check to make sure Feature isn't present before adding it

JT122406:
Fixes an issue where a feature order cycle error will happen because a feature may be added to a biome twice.  This can occur if a biome appears in more than one tag and you want a feature in biomes from both tags, currently it gets added twice to the biome causing a feature order cycle error


apple502j:
Maybe a log?

JT122406:
Like print it to the console log that this is happening?

JT122406:
Thinking maybe a debug or a warn or? In my head putting it in like info doesn't make sense as this could happen very easily with tags and it's not like a "bad" thing it's just something that happens

modmuss50:
I think this seems fine, as it can happen in a valid case. We would only tend to log if something was broken.

modmuss50 (approved):

JT122406:
I wanna get this back ported to 1.21.1 how do yall handle that do I need to make a pr into that version?

PR #4327, opened 2024-12-25
Fix world closing in client gametests

Earthcomputer:
There was an issue where the assertion that the game is on the title screen would fail when ending a test by closing a singleplayer world (via a try-with-resources close of `TestSingleplayerContext`).

This was caused by the threading system deferring the `client.disconnect` call until after the client thread task wait loop (i.e. until the gametest thread waits a tick). Therefore, the order of operations inside `GameMenuScreen.disconnect` was reversed and the title screen was shown first before disconnecting. Moreover, the title screen was visible to the gametest thread until ticks were waited, which allowed `context.waitForScreen(TitleScreen.class)` to return immediately after waiting 0 ticks.

Rather than relying on navigating the game menu which may have been changed by mods, I thought it was better to refactor the code to disconnect directly, and I've written it in a way which waits for the `world` to be `null` directly after calling `client.disconnect`, which allows for the deferred call to run.

### DIFF
--- a/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestServerConnectionImpl.java
+++ b/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestServerConnectionImpl.java
@@ -16,7 +16,6 @@
 
 package net.fabricmc.fabric.impl.client.gametest;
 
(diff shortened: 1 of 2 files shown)


kevinthegreat1 (approved):
Works for me. (Tested by manually using this version.)

modmuss50 (approved):

PR #4329, opened 2024-12-26
Client gametest screenshot changes

Earthcomputer:
- Screenshots are now instant, you no longer have to wait a tick for them to render the current state.
   - The frame is re-rendered explicitly for the purpose of taking the screenshot. This allows us to set the `tickDelta` value to `1`, improving consistency.
   - The option to delay before taking a screenshot has been removed. You can still manually create a delay before a screenshot by using `context.waitTick()` and `context.waitTicks(delay)`.
   - This has exposed other footguns with client gametests, though. In some cases, the game needs to tick to be able to process an action, such as with some key binds which are polled, and when a packet needs to be sent to the server (which we have fixed to take exactly 1 tick in gametest singleplayer). To be able to see the effects of your actions, you may need to wait ticks before taking a screenshot. I have decided to not wait a tick before taking screenshots though, as this is observable in other ways than just this, and unlike before, the screenshots are a reflection of the real state of the game.
- Screenshots now add a counter prefix to their filename by default, so that they appear in the same order in the file explorer as they were taken.
- Added more configuration options to taking a screenshot:
   - Option to disable the counter prefix.
   - Option to change the `tickDelta` to something other than `1`.
   - Option to change the screenshot resolution.
   - Option to change the screenshot save directory.
- Consistent window sizes
   - The window size according to the game always starts at (854, 480) (or whatever is set in the command line options), regardless of if the window manager overrides the physical size of the window.
   - If the physical window is resized either by the user or window manager, this will not affect the game (e.g. how widgets are layed out in GUIs, the screenshot resolutions, etc). The content of the window may appear stretched to the user, but this does not affect screenshots.
   - Added a programmatic way to resize the window in a way that is visible to the game.
- Automated screenshot comparisons are left to a future PR.

### DIFF
--- a/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotOptionsImpl.java
+++ b/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotOptionsImpl.java
@@ -45,7 +45,7 @@ public class TestScreenshotOptionsImpl implements TestScreenshotOptions {
 
 	@Override
 	public TestScreenshotOptions withTickDelta(float tickDelta) {
-		Preconditions.checkArgument(tickDelta >= 0 && tickDelta <= 1, "tickDelta should be between 0 and 1");
+		Preconditions.checkArgument(tickDelta >= 0 && tickDelta <= 1, "tickDelta must be between 0 and 1");
 
 		this.tickDelta = tickDelta;
 		return this;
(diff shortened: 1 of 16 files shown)


fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotOptionsImpl.java
Alexander01998, line 1: nit: The use of public, non-final fields in this class and the mix of `should` and `must` in the preconditions text seems weird and inconsistent.

The preconditions seem to imply that you can set the tick delta to a value outside the 0-1 range, while you can't set negative width and height values. But actually both of these restrictions can be bypassed by modifying the `tickDelta` and `size` fields directly.

Earthcomputer: To bypass it you have to downcast to an impl class, so I'm not too worried about it. I'll fix the wording of should

modmuss50 (approved):

PR #4334, opened 2024-12-28
Solve inconsistency issues in client gametests

Earthcomputer:
This PR fixes the two biggest remaining consistency issues with client gametests that I have encountered:

- I had assumed that the vanilla task queue on the client runs every tick, but it actually runs every frame, leading to ticks sometimes not running when `waitTick` is run.
   - I now allow multiple iterations of the client task loop to run (i.e. multiple frames) within the client-server tick synchronizer, and only hand control back to the gametest thread when there is going to be a tick this frame.
   - I also limit the number of ticks in a single frame to 1. This simplifies the implementation compared to what it otherwise would be, and also prevents there from being ticks where tasks are not run (and packets are not handled).
- I had assumed that packets in singleplayer reach the other side immediately, and that synchronizing the vanilla task loops was enough to make packets consistent. In fact, even though packets are not serialized in singleplayer, they can still take an arbitrarily long amount of time to reach the receiving side (but typically don't take long) just as they can in multiplayer. This was a silly assumption in hindsight.
   - I have introduced `NetworkSynchronizer` to solve this issue.
   - A packet can be either "in-flight", which is between the time it is sent and the time it is handled on the Netty thread on the receiving side, or it can be queued for handling on the receiving main thread, which it between when it is handled on the Netty thread and it is removed from the main thread task queue.
      - Some packets are handled directly on the Netty thread and never enter the second stage. The `NetworkSynchronizer` is careful not to assume that all packets must be added to the task queue.
   - Once the packets are tracked in this way, the key change is that the client and server now continue running their task loops until there are no in-flight packets and no packets waiting to be handled in the vanilla task queues. 
   - Network synchronization can be disabled via a system property, which is useful for mods which directly interface with the Netty pipeline, which would desynchronize the in-flight packet counter.


modmuss50 (approved):
Im not going to pretend to understand all the logic, but as with the previous PRs the reasoning is sound. Thanks for all the effort with this.

PR #4337, opened 2024-12-29
Clean run directory on startup for client gametests

Earthcomputer:
This is only one possibility for solving the problem of cleaning the run directory for client gametests. We can decide later whether we want to actually do it this way. With this PR we can see what such an implementation would look like.

Note: if this PR is to be merged, the run directory of FAPI's own client gametest needs to be changed.


apple502j:
I feel like this is something Loom should handle.

PR #4338, opened 2024-12-30
[1.21.1] Biomemodifications: Check to make sure Feature isn't present before adding it. BackPort of #4323

JT122406:
Fixes an issue where a feature order cycle error will happen because a feature may be added to a biome twice. This can occur if a biome appears in more than one tag and you want a feature in biomes from both tags, currently it gets added twice to the biome causing a feature order cycle error


modmuss50:
Hi, just seen this PR is identical to https://github.com/FabricMC/fabric/pull/4323, it will be easier for me to manage if I cherry pick it down. We only need a seperate PR when the change cannot be easily chery picked. Thanks.

JT122406:
Sounds good I wasn't sure if you just cherry-picked or not

PR #4352, opened 2025-01-05
Add `c:flowers`, `c:flowers/tall`, and `c:flowers/small` block and item tags

TelepathicGrunt:
Mojang has recently rid `minecraft:tall_flowers` block tag and `minecraft:tall_flowers`/`minecraft:flowers` item tags.

What this PR aims to do is bring back and duplicate these flower tags under `c` namespace to make it safer for Modders to use without worry about Mojang deleting the rest of the flower tags in the future. We had multiple people ask where the flower item tag went so there is quite a number of people relying on a tag of flowers.

The non-tag entries of the flower tags are in same order as the tags in past (with exception to eye blossom because vanilla has the open and closed entry far apart in the vanilla for no logical reason which drove me insane)

For Fabric side, the `c` flower tags that still have a vanilla equivalent are aliased together.

I also added entries to log warning to find people using vanilla tags that no longer exist and guide them to the new `c` tags

Blocks:
`c:flowers/tall`
`c:flowers/small`
`c:flowers`

Items:
`c:flowers/tall`
`c:flowers/small`
`c:flowers`

Any issues with this? Is there any flower subcategory that is strongly desired beyond tall/small?

NeoForge PR: https://github.com/neoforged/NeoForge/pull/1825


fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -694,6 +718,9 @@ private void generateTagAlias() {
 		aliasGroup("fences").add(BlockTags.FENCES, ConventionalBlockTags.FENCES);
 		aliasGroup("fences/wooden").add(BlockTags.WOODEN_FENCES, ConventionalBlockTags.WOODEN_FENCES);
 		aliasGroup("fence_gates").add(BlockTags.FENCE_GATES, ConventionalBlockTags.FENCE_GATES);
+
+		aliasGroup("flowers/small").add(BlockTags.SMALL_FLOWERS, ConventionalBlockTags.SMALL_FLOWERS);
+		aliasGroup("flowers").add(BlockTags.FLOWERS, ConventionalBlockTags.FLOWERS);
```
Juuxel, line 722: The vanilla tags have behaviour changes here, but IMO it's fine for the `flowers` tag (the only gameplay features relate to bees).

According to the wiki, `small_flowers` also changes the effective tool to swords, which might not be something modders want for all their small flowers?

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -308,6 +308,24 @@ private ConventionalItemTags() {
 	public static final TagKey<Item> UNCOLORED_SANDSTONE_SLABS = register("sandstone/uncolored_slabs");
 	public static final TagKey<Item> UNCOLORED_SANDSTONE_STAIRS = register("sandstone/uncolored_stairs");
 
+	// Flower
+	/**
+	 * Contains living ground-based flowers that are 1 block tall such as Dandelions or Poppy.
+	 * Equivalent to the "minecraft:small_flowers" item tag.
```
Juuxel:
```suggestion
	 * Equivalent to the {@code minecraft:small_flowers} item tag.
```
Code formatting looks nicer for ids

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -308,6 +308,24 @@ private ConventionalItemTags() {
...
+	 * Aliased with {@link ItemTags#SMALL_FLOWERS}.
+	 */
+	public static final TagKey<Item> SMALL_FLOWERS = register("flowers/small");
+	/**
+	 * Contains living ground-based flowers that are 2 block tall such as Rose Bush or Peony.
+	 * Equivalent to the "minecraft:tall_flowers" item tag in past Minecraft version.
```
Juuxel:
```suggestion
	 * Equivalent to the {@code minecraft:tall_flowers} item tag in past Minecraft versions.
```

fabric-convention-tags-v2/src/main/java/net/fabricmc/fabric/api/tag/convention/v2/ConventionalItemTags.java
```diff
@@ -308,6 +308,24 @@ private ConventionalItemTags() {
...
+	 */
+	public static final TagKey<Item> TALL_FLOWERS = register("flowers/tall");
+	/**
+	 * Contains any living plant block that contains flowers or is a flower itself.
+	 * Equivalent to the "minecraft:flowers" item tag in past minecraft versions.
```
Juuxel:
```suggestion
	 * Equivalent to the {@code minecraft:flowers} item tag in past Minecraft versions.
```

fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
TelepathicGrunt: hmm would we then rid the small flower item tag alias as well? it would be weird to alias only on the item tag and not block tag

TelepathicGrunt: I removed the alias and added why the alias is not done in the javadoc to help reduce confusion

modmuss50 (approved):

PR #4363, opened 2025-01-08
Fix focus script to always include "fabric-client-gametest-api-v1"

modmuss50:


gradle/Focus.java
```diff
@@ -56,6 +56,7 @@ public static void main(String[] args) throws IOException {
 
 		// All modules depend on fabric-gametest-api-v1 and fabric-registry-sync-v0
```
Earthcomputer: This comment needs updating too

PR #4365, opened 2025-01-08
feat: Equipment model provider

IMB11:
In 1.21.4, Mojang added equipment models to replace the previous texture format for armour. This PR adds datagen support for these equipment model definitions.

**Example Usage (testmod)**

https://github.com/IMB11/fabric/blob/1.21.4/fabric-data-generation-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/datagen/client/DataGeneratorClientTestEntrypoint.java#L80-L102


modmuss50 (changes requested):
As discussed on discord, dont re-eningger something that has already been done for us. This should be a thing wrapper around the existing vanilla `EquipmentAssetProvider`.

Extend the class, replace the static method with an abstract one, and it will be done in about 5 lines of code.

IMB11:
Will make separate PR.

PR #4381, opened 2025-01-14
Client gametest screenshot comparison API

Earthcomputer:
Adds APIs to assert that a screenshot matches or contains a template image.


fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/api/client/gametest/v1/TestScreenshotComparisonOptions.java
```diff
@@ -0,0 +1,112 @@
...
+ * matching unless you specifically require an exact match, because the exact screenshot can vary slightly on different
+ * frames and in rare cases other factors such as the GPU/driver version.
+ *
+ * <p>Template images must be fully opaque, this API will throw if the template image has any transparent pixels.
+ *
+ * <p>Template images loaded from a path are expected to be in the {@code templates} directory inside the resources
+ * directory of the mod which registered the gametest. If the template image is not found, and the
+ * {@code fabric.client.gametest.testModResourcesPath} system property is set to the resources folder of the mod which
+ * registered the gametest, then the screenshot will be saved to the {@code templates} folder so that it can be used
+ * next time the gametest is run.
```
Earthcomputer, line 43: I'm wondering if loom could set this system property by default?

modmuss50: Yes, we can assume the default path and allow the user to override it in the DSL, can you open an loom issue so its not forgotten about?

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotComparisonAlgorithms.java
```diff
@@ -0,0 +1,231 @@
...
+
+					for (int needleX = 0; needleX <= maxNeedleX; needleX++) {
+						if (predicate.isAt(needleX, needleY)) {
+							return new Vector2i(needleX, needleY);
+						}
+					}
+
+					return null;
+				})
+				.filter(Objects::nonNull)
```
modmuss50, line 200: There is nothing wrong with this, but if im not mistaken there will be wasted time on all of the none matching y values. The cheap way of doing it would be a an atomic boolean that checks within the for loop if set return null. This saves searching for something that has already been found by another thread.

Edit, just read it again and I think the `findAny()` handles this, close this comment if so.

modmuss50 (approved):
This is really good, the default algorithms look good, im glad we arent pulling in natives to do it :D

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/TestScreenshotComparisonAlgorithms.java
Earthcomputer: Yeah I believe nonterminal operations in streams are lazy, so it won't generate any more once `findAny()` has found something.

modmuss50: Very clever, I dont usually think to use streams to manage parallel tasks like this. 👍

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/api/client/gametest/v1/TestScreenshotComparisonAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * A thin wrapper around a raw image data array, so that algorithms can directly access the array which may be more
+	 * efficient than going through a {@link NativeImage} each time.
+	 *
+	 * @param <DATA> The type of the image data array
+	 */
+	@ApiStatus.NonExtendable
+	interface RawImage<DATA> {
```
Juuxel, line 104: This is a non-extendable interface, but are there any public APIs that create instances of it (ie can it even be used with API code)?

Earthcomputer: There are no public APIs to create instances of `RawImage`, but it can be used by custom implementations of `TestScreenshotComparisonAlgorithm`

PR #4385, opened 2025-01-17
25w03a game tests

modmuss50:
Tottaly overhalled Game Test module to accomdate the changes made in 25w03a.

I have added back a `@GameTest` that functions almost the same as the previous one vanilla had. The new vanilla system makes it much harder to write a test than it was before. As there was a lot of value in the magic annotation and tests that are easy to write I think it makes sense to have something similar to what we had before.

The options in the new Fabric provided `@GameTest` annotation directly map the vanilla data driven options, removing the need to have a json file for each test function. Data driven tests will still work if you wish to use the vanilla system.

The previous `fabric-api.gametest` and `fabric-api.gametest.report-file` system properties work as they did before. `fabric-api.gametest.filter` and `fabric-api.gametest.verify` have been added that map directly to the vanilla gametest entrypoint options.

Tests can still be ran/debuged from within a client using the `/test` command. Structures can still be loaded from snbt files as before.


fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestModInitializer.java
```diff
@@ -16,49 +16,47 @@
...
-		if (!GAME_TEST_IDS.containsKey(testClass)) {
-			throw new UnsupportedOperationException("The test class (%s) was not registered using the '%s' entrypoint".formatted(testClass.getCanonicalName(), ENTRYPOINT_KEY));
+		for (TestAnnotationLocator.TestMethod testMethod : locator.getTestMethods()) {
+			LOGGER.debug("Registering test method: {}", testMethod.identifier());
+			Registry.register(Registries.TEST_FUNCTION, testMethod.identifier(), testMethod.testFunction());
 		}
 
-		return GAME_TEST_IDS.get(testClass);
+		DynamicRegistrySetupCallback.EVENT.register(registryView -> {
+			// We ideally need an after loaded event, but this should hopefully work for now
```
Earthcomputer: Yeah we do, this code is kinda horrible. If Mojang adds other builtin test instances, then the modded ones will be before them in the registry.

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/TestAnnotationLocator.java
```diff
@@ -0,0 +1,159 @@
...
+		findMagicMethods(entrypoint, testClass, methods);
+
+		if (methods.isEmpty()) {
+			LOGGER.warn("No methods with the FabricGameTest annotation were found in {}", testClass.getName());
+		}
+
+		return methods;
+	}
+
+	// Recursively find all methods with the GameTestData annotation
```
Earthcomputer:
```suggestion
	// Recursively find all methods with the GameTest annotation
```

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestModInitializer.java
modmuss50: I might change this to a dedicated mixin, will need to duplicate some stuff from reg sync tho.

haykam821 (changes requested):
Many tests get broken by the change from a 8x8x8 empty structure template to the vanilla 1x1x1 one.

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/FabricGameTestModInitializer.java
modmuss50: Fixed, now uses a dedicated mixin. Id be somewhat hesitant to add an api for this into reg sync, I dont think we should encorage people to register dynmatic reg entries in code. Anyway thats a discussion for another time.

PR #4401, opened 2025-01-26
Client gametest cleanup

Earthcomputer:
A few bits of cleanup to the client gametest API. I recommend reviewing the commits individually so that the non-repackage changes are easier to spot.


modmuss50 (approved):

fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/api/client/gametest/v1/package-info.java
```diff
@@ -3,27 +3,38 @@
...
+ * <p>Loom provides an API to configure client gametests in your {@code build.gradle}. It is recommended to run
+ * gametests from a separate source set and test mod:
+ * <pre>
+ *     {@code
+ *     fabricApi.configureTests {
+ *         createSourceSet = true
+ *         modId = 'your-gametest-mod-id'
+ *     }
+ *     }
+ * </pre>
```
kevinthegreat1, line 15: Might want to add a quick and simple `since loom 1.10` notice.

Earthcomputer: I don't think there's any need. I don't think we will be making this API non-experimental before there is a stable version of Loom which supports its configuration.

modmuss50: I agree with @Earthcomputer I'm wrapping up the final touches to loom now. If we added this note it wont ever get removed.

PR #4404, opened 2025-01-29
Fix Client Registry Remapping Only Remapping One Registry

AViewFromTheTop:
- return was used instead of continue, causing only the last modified registry to sync to the client.
Fix is simply using an if-else statement instead of calling return within the if statement.


modmuss50 (approved):
Added a unit test, many thanks for getting to the bottom of this.

PR #4416, opened 2025-02-04
Use page up/down to change creative inventory pages.

modmuss50:


fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
```diff
@@ -108,6 +109,19 @@ private void renderTabIcon(DrawContext drawContext, ItemGroup itemGroup, Callbac
...
+	private void keyPressed(int keyCode, int scanCode, int modifiers, CallbackInfoReturnable<Boolean> cir) {
+		if (keyCode == GLFW.GLFW_KEY_PAGE_UP) {
+			if (switchToNextPage()) {
+				cir.setReturnValue(true);
+			}
+		} else if (keyCode == GLFW.GLFW_KEY_PAGE_DOWN) {
+			if (switchToPreviousPage()) {
+				cir.setReturnValue(true);
+			}
+		}
```
Octol1ttle: I don't like PG UP being next page. To me it makes sense that vanilla is the "top" page and mods are the "bottom" pages.

Juuxel:
Should the keys be configurable?

Octol1ttle:
> Should the keys be configurable?

I don't think so. It's an extremely minor feature, would require either adding some sort of configuration file to FAPI (the only one I know of is for Indigo) or polluting the keybinds menu (which IMO isn't worth it when the active context is so rare)

fabric-item-group-api-v1/src/client/java/net/fabricmc/fabric/mixin/itemgroup/client/CreativeInventoryScreenMixin.java
LambdAurora: To add to this, on some keyboards those keys will be labeled "previous page" for page up, and "next page" for page down, so I support this suggestion.

modmuss50: I agree, I have swapped these around.

Octol1ttle (approved):

PR #4432, opened 2025-02-09
Add getAllMatches and getAllOfType method to ServerRecipeManager

Patbox:
Adds 2 methods to ServerRecipeManager via interface injection:
- Stream<RecipeEntry> getAllMatches(RecipeType, RecipeInput, World) - gets all matches of a recipe, instead of only first one (like getFirstMatch)
- Collection<RecipeEntry> getAllOfType(RecipeType type) - returns all recipes of selected type.

Both of these are just simple accessors to methods provided by PreparedRecipes class.

### DIFF
--- a/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricServerRecipeManager.java
+++ b/fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricServerRecipeManager.java
@@ -1,7 +1,22 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricServerRecipeManager.java
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * Creates a stream of all recipe entries of the given {@code type} that match the
+	 * given {@code input} and {@code world}.
+	 *
+	 * <p>If {@code input.isEmpty()} returns true, the returned stream will be always empty.
+	 *
+	 * @return the stream of matching recipes
+	 */
+	default <I extends RecipeInput, T extends Recipe<I>> Stream<RecipeEntry<T>> getAllMatches(RecipeType<T> type, I input, World world) {
+		return Stream.of();
```
modmuss50: These should throw, we tend to just do `throw new AssertionError("Implemented in Mixin");`

modmuss50 (approved):

PR #4436, opened 2025-02-11
Fix modded oxidizable blocks not oxidizing

### LINKED ISSUE #4226: Random ticks are cached as disabled when blocks are made oxidizable after being registered
A block registered and then added to the oxidizable blocks map using `OxidizableBlockRegistry` will never randomly tick, because the random tick boolean, which is computed based on the oxidizable blocks map, will be cached upon block registration.

Refer to [this Discord message](https://discord.com/channels/507304429255393322/566276937035546624/1306670090753802240) for the original context.

### LINKED ISSUE #4371: `OxidizableBlocksRegistry.registerOxidizableBlockPair` does not work for Vanilla oxidizable Block implementations
When registering a pair of custom oxidizable blocks created from a Vanilla `Block` implementation (such as `WeatheringCopperFullBlock` (mojmap)). These blocks override the `isRandomlyTicking` method, and check the field `WeatheringCopper.NEXT_BY_BLOCK` to see if they can oxidize further. This field is updated by `OxidizableBlocksRegistry.registerOxidizableBlockPair`, so it would work _if it wasn't for `BlockState` caching the return value of `isRandomlyTicking` at the moment the `Block` is created._ Hence, upon the time of caching this value, the block pair isn't registered yet and the block will not be randomly ticked.

```java
// WeatheringCopperFullBlock.java (mojmap!)
@Override
protected boolean isRandomlyTicking(BlockState state) {
    // Here it accesses the WeatheringCopper.NEXT_BY_BLOCK registry, updated by OxidizableBlocksRegistry
    // This is called and cached before we get the chance to call OxidizableBlocksRegistry!
    return WeatheringCopper.getNext(state.getBlock()).isPresent();
}
```

It's not actually possible to override this by setting `Block.Properties.randomTicks()`, since the `Block` implementations completely override that. The only workaround is to extend the vanilla block logic and override `isRandomlyTicking` to check for the weathering state instead of looking in the registry.

The cause lays somewhere with the Fabric Registry Sync API initializing the block states early. The problem could potentially be fixed if `OxidizableBlocksRegistry.registerOxidizableBlockPair` re-caches the `isRandomlyTicking` for all the block states of the registered blocks. The only reason it does work for Vanilla blocks is because the Vanilla blocks are hardcoded into the registry in the `WeatheringCopper` interface.

Shnupbups:
Fixes #4371 and #4226

`OxidizableBlocksRegistry.registerOxidizableBlockPair` will now re-cache the value of the `ticksRandomly` field of the passed blocks' states to ensure that the default implementation of oxidization will actually work on the blocks.

This may be considered a slightly dirty workaround, but it works well and isn't particularly intrusive, despite using 2 mixins.

I've also included a test (though not a GameTest) to ensure that this works, and split the `OxidizableBlocksRegistry` into separate API and impl classes to more easily allow for potential extensibility in future while not breaking compatibility.


Shnupbups:
Um, not sure why the data attachment API gametest failed, I didn't touch that at all...

Earthcomputer:
It fails intermittently, don't worry about it

PepperCode1 (approved):

modmuss50 (approved):

PR #4437, opened 2025-02-11
Fix incorrect HUD z index

### LINKED ISSUE #4435: Fabric Rendering API breaking vanilla core shaders
### Description
Since `0.116.0+1.21.4` several UI elements no longer render on the vanilla positions. I have not tested every UI element, only noticed it being off at the player list no longer rendering at z-Index 2800, instead it renders at z-index 1600.

### Steps to Reproduce
1. Install Fabric API `0.116.0+1.21.4`
2. Install this resource pack [test.zip](https://github.com/user-attachments/files/18751354/test.zip)
3. Join a multiplayer server
4. Observe the player list being colored in magenta, instead of cyan.

### Expected Behavior
The player list should be colored in cyan (screenshot vanilla): 
Image

### Observed Behavior
The player list is colored in magenta (screenshot fabric api `0.116.0+1.21.4`):
Image

### Additional Info
This change in z-index did not occur on fabric api `0.115.1+1.21.4`.

modmuss50:
Fixes #4435


kevinthegreat1 (approved):
Ah, z offset in sub drawers weren't kept. This makes sense.

PR #4447, opened 2025-02-16
Change elytra event mixin injection point

PrimordialMoros:
The `EntityElytraEvents.CUSTOM` allows specifying your own condition for enabling elytra gliding. However, vanilla code checks all applicable slots for items that allow gliding. If you are using a custom condition instead of an item then an empty list is supplied which causes a crash due to an index check in `Util.getRandom`.

Moved the injection point for the `EntityElytraEvents` to avoid this issue.


modmuss50 (approved):
Looks good, makes sense. Thanks.

PR #4450, opened 2025-02-18
Update Model Loading API to 1.21.5

PepperCode1:
- Add `ModelModifier.*BakeItem` events. Their utility is currently limited as the render state cannot be inspected, but they will be much more useful after the 1.21.5 FRAPI port.
- Remove `ModelModifier.*Bake` events as `UnbakedModel`s are not really baked anymore
- `ModelModifier.*BakeBlock` events now accept a `BlockState` instead of a `ModelIdentifier`
- Temporarily remove extra model API. Porting it directly is not possible and more time needs to be spent designing a new API that will cover the same use cases.
- Remove `UnwrappableBakedModel` as it is potentially less useful with the changes to the model system. It will be re-added later if necessary.
- Add `WrapperBlockStateModel` which simply delegates all method calls to a different `BlockStateModel`. This class existed in vanilla in 1.21.4 as `WrapperBakedModel` but was removed.
- Add `WrapperUnbakedItemModel` which simply delegates all method calls to a different `ItemModel.Unbaked`
- Add `WrapperItemModel` which simply delegates all method calls to a different `ItemModel`
- Move all model wrapper-related classes to a separate package


modmuss50:
FYI: Build failuire should be fixed when you merge.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/WrapperUnbakedItemModel.java
```diff
@@ -0,0 +1,51 @@
...
+
+package net.fabricmc.fabric.api.client.model.loading.v1;
+
+import com.mojang.serialization.MapCodec;
+
+import net.minecraft.client.render.item.model.ItemModel;
+
+/**
+ * A simple implementation of {@link ItemModel.Unbaked} that delegates all method calls to the {@link #wrapped} field.
+ * Implementations must set the {@link #wrapped} field somehow.
```
modmuss50: Question: I assume there is a good reason why wrapped isnt final, and the mod is expected to set the field.

Generally we tend not expose fields as part of the API, would an `abstract ItemModel.Unbaked getWrapped()` be more suitable for these classes.

modmuss50 (commented):
Looks really good to my untrained eye, just one small question about the api design.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/WrapperUnbakedItemModel.java
PepperCode1: Fabric API had the `ForwardingBakedModel` class for the longest time, which used the design of having a protected, non-final field. Any new model wrapper classes use the same design. The field could be final, or the class could be removed entirely, but it's so simple that the field may as well be non-final to cover a couple extra use cases so the user doesn't need to copy the class.

modmuss50: Ok, thats fair enough. It seems unlikely we would ever want to evolve these apis in a none breaking way so its unlikely going to be an issue.

modmuss50 (approved):

PR #4452, opened 2025-02-19
Add vararg helper methods for multi-tag support in the FabricTagBuilder

Starexify:
This makes it possible to use multiple arguments in the new addTags methods so that modders using datagen <sub><sub>(like me)</sub></sub> no loger have to call the `addTag()` method for each `TagKey<T>`.

This PR will add the following vararg methods:
```java
/**
 * Add multiple tags to this tag.
 *
 * @return the {@link FabricTagBuilder} instance
 */
@SafeVarargs
public final FabricTagBuilder addTags(TagKey<T>... tags) {
	for (TagKey<T> tag : tags) {
		addTag(tag);
	}

	return this;
}

/**
 * Add multiple optional tags to this tag.
 *
 * @return the {@link FabricTagBuilder} instance
 */
@SafeVarargs
public final FabricTagBuilder addOptionalTags(TagKey<T>... tags) {
	for (TagKey<T> tag : tags) {
		addOptionalTag(tag);
	}

	return this;
}

/**
 * Add multiple tags to this tag, ignoring any warning.
 *
 * <p><b>Note:</b> only use this method if you are sure that the tags will be always available at runtime.
 * If not, use {@link #addOptionalTags(TagKey[])} instead.
 *
 * @return the {@link FabricTagBuilder} instance
 */
@SafeVarargs
public final FabricTagBuilder forceAddTags(TagKey<T>... tags) {
	for (TagKey<T> tag : tags) {
		forceAddTag(tag);
	}

	return this;
}
```

Also changed the JavaDocs at line 421 to link the `addOptionalTag(TagKey<T>)` instead of `* If not, use {@link #addOptionalTag(Identifier)} instead.`

Also changed the `add(T... elements)` method to use a for loop instead of a `Stream.of()`:
```java
@SafeVarargs
public final FabricTagBuilder add(T... elements) {
	for (T element : elements) {
		add(reverseLookup(element));
	}

	return this;
}
```

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -404,7 +404,7 @@ public abstract class FabricTagProvider<T> extends TagProvider<T> {
 		}
 
 		/**
-		 * Add another optional tags to this tag.
+		 * Add optional tags to this tag.
 		 *
 		 * @return the {@link FabricTagBuilder} instance
 		 */
@@ -428,7 +428,7 @@ public abstract class FabricTagProvider<T> extends TagProvider<T> {
 		}
 
 		/**
-		 * Add another tags to this tag, ignoring any warning.
+		 * Add tags to this tag, ignoring any warning.
 		 *
 		 * <p><b>Note:</b> only use this method if you sure that the tags will be always available at runtime.
 		 * If not, use {@link #addOptionalTags(TagKey[])} instead.
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
@@ -379,7 +379,9 @@ public abstract class FabricTagProvider<T> extends TagProvider<T> {
 		 */
 		@SafeVarargs
 		public final FabricTagBuilder addTags(TagKey<T>... tags) {
-			Stream.of(tags).forEach(this::addTag);
+			for (TagKey<T> tag : tags) {
+				addTag(tag);
+			}
 			return this;
 		}
 
@@ -410,7 +412,9 @@ public abstract class FabricTagProvider<T> extends TagProvider<T> {
 		 */
 		@SafeVarargs
 		public final FabricTagBuilder addOptionalTags(TagKey<T>... tags) {
-			Stream.of(tags).forEach(this::addOptionalTag);
+			for (TagKey<T> tag : tags) {
+				addOptionalTag(tag);
+			}
 			return this;
 		}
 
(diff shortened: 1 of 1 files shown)


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -372,6 +372,17 @@ public FabricTagBuilder addTag(TagKey<T> tag) {
...
 		}
 
+		/**
+		 * Add multiple tags to this tag.
+		 *
+		 * @return the {@link FabricTagBuilder} instance
+		 */
+		@SafeVarargs
+		public final FabricTagBuilder addTags(TagKey<T>... tags) {
+			Stream.of(tags).forEach(this::addTag);
```
modmuss50: Nit: Just use a for loop.

Starexify: Why? Are for loops better than using a stream of the tags? (I will change if so)

modmuss50: In this case mostly becuase its makes the code more readable, there is no benefit to using a stream here.

modmuss50 (approved):

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
Starexify: Alright I changed them 👍
Should I also change the method at line 306 to use for loop instead:
```java
@SafeVarargs
public final FabricTagBuilder add(T... element) {
	Stream.of(element).map(FabricTagProvider.this::reverseLookup).forEach(this::add);
	return this;
}
```
(not method I added tho)

modmuss50: Yes, again I find that much harder to read than

```java
@SafeVarargs
public final FabricTagBuilder add(T... elements) {
	for (T element: elements) {
		add(reverseLookup(element));
	{

	return this;
}
```

modmuss50:
Just some checkstyle failuires now, you need a blank line after the for loop.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricTagProvider.java
```diff
@@ -405,6 +433,23 @@ public FabricTagBuilder forceAddTag(TagKey<T> tag) {
 			return this;
 		}
 
+		/**
+		 * Add multiple tags to this tag, ignoring any warning.
+		 *
+		 * <p><b>Note:</b> only use this method if you sure that the tags will be always available at runtime.
```
apple502j:
```suggestion
		 * <p><b>Note:</b> only use this method if you are sure that the tags will be always available at runtime.
```

Starexify: I also changed for the `forceAddTag()` above so it matches

modmuss50 (approved):

PR #4457, opened 2025-02-20
Fix TypeAwareBuyForOneEmeraldFactory trade offer crash

### LINKED ISSUE #4456: Crash due to `TradeOffersTypeAwareBuyForOneEmeraldFactoryMixin` incorrectly assuming null trade items will create empty item stacks
This method in the mixin assumes that an empty item stack will be created due to `null` being passed in the constructor.
```java
/**
 * To prevent "item" -> "air" trades, if the result of a type aware trade is air, make sure no offer is created.
 */
@Inject(method = "create", at = @At(value = "NEW", target = "net/minecraft/village/TradeOffer"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
private void failOnNullItem(Entity entity, Random random, CallbackInfoReturnable<TradeOffer> cir, ItemStack buyingItem) {
    if (buyingItem.isEmpty()) { // Will return true for an "empty" item stack that had null passed in the ctor
        cir.setReturnValue(null); // Return null to prevent creation of empty trades
    }
}
```
Unfortunately, at least in 1.20.1, item stack constructors take `ItemConvertible` and call `asItem` on the input.
```java
@Nullable
@Override
public TradeOffer create(Entity entity, Random random) {
    if (entity instanceof VillagerDataContainer) {
        ItemStack itemStack = new ItemStack((ItemConvertible)this.map.get(((VillagerDataContainer)entity).getVillagerData().getType()), this.count);
        return new TradeOffer(itemStack, new ItemStack(Items.EMERALD), this.maxUses, this.experience, 0.05F);
    } else {
        return null;
    }
}
```
Which causes a null pointer exception if a `null` item is given.

I imagine this was not noticed until now because in vanilla, the only type aware trade is when a fisherman is at its maximum level and a mod has added a custom villager type. My custom villager type also has its own type aware trades at a lower level though, so this crash happened without max level - but I don't think this could have ever worked, at least with how `TypeOffersTypeAwareBuyForOneEmeraldFactory` is implemented in 1.20.1.

I suppose to fix this the mixin will have to add a condition before the item stack is created in `TypeOffersTypeAwareBuyForOneEmeraldFactory#create`, since the mixin disables vanilla's crash in the `TypeOffersTypeAwareBuyForOneEmeraldFactory` constructor for missing trades.

Antikyth:
Fixes #4456

Have tested and confirmed that my newly added game test fails without this fix, succeeds with it.

I'd like feedback on whether the specific use of `@ModifyExpressionValue` here is appropriate - I'm using it to intercept the `TradedItem` item argument call to cancel if it is null before the constructor is called.


Antikyth:
This mixin appears to be precisely the same in 1.20.1, even though the previous version it's replacing was different between 1.20.1 and 1.21.4.

Antikyth:
Would like this backported to 1.20.1 for use in my mod :)

Antikyth:
I will do a proper comb through with the check task tomorrow.

modmuss50 (approved):
Looks good, thanks.

modmuss50:
Ill merge this int 1.21.4 now and then open a PR to back port this to 1.20.1.

DawsonBodenhamer:
# This crash still persists in Development Environment on 0.92.6+1.20.1

Hey there,

I am still encountering the `InvalidInjectionException` regarding `TypeAwareBuyForOneEmeraldFactory`, on my mod **[Adorable Hamster Pets](https://modrinth.com/mod/adorable-hamster-pets)**, even after updating to **Fabric API 0.92.6+1.20.1** (which should include the backported fix from this PR).

**Crucially, this crash only occurs in the Development environment** (via `runClient` in IntelliJ with Architectury/Loom). The built JARs function correctly in a production environment. It appears that while the fix resolves the issue for the production Mixin applier, it may still trigger a signature failure during the mixin validation of the development environment.

I can't downgrade to the 0.91.x versions because my project relies on GeckoLib 4.7.1+ that hard-depends on Fabric API 0.92.0+.

## **Environment:**
*   Minecraft: 1.20.1
*   Fabric API: 0.92.6+1.20.1
*   Loader: 0.15.11
*   Context: Development (`runClient`)

## **The Error:**
```text
org.spongepowered.asm.mixin.injection.throwables.InvalidInjectionException: @ModifyExpressionValue expression value modifier method net/minecraft/village/TradeOffers$TypeAwareBuyForOneEmeraldFactory::failOnNullItem from fabric-object-builder-v1.mixins.json:TradeOffersTypeAwareBuyForOneEmeraldFactoryMixin from mod fabric-object-builder-api-v1 has an invalid signature. 

Found unexpected argument type org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable at index 1, expected net.minecraft.entity.Entity. 

Handler signature: (Ljava/lang/Object;Lorg/spongepowered/asm/mixin/injection/callback/CallbackInfoReturnable;)Ljava/lang/Object; 
Expected signature: (Ljava/lang/Object;Lnet/minecraft/entity/Entity;Lnet/minecraft/util/math/random/Random;)Ljava/lang/Object;
```

Screenshot 2025-11-22 153032

Thanks in advance!

cputnam-a11y:
Hi. Could you share the full log from an erroring instance after the update?

DawsonBodenhamer:
> Hi. Could you share the full log from an erroring instance after the update?

Yes here you go

[latest.log](https://github.com/user-attachments/files/23692047/latest.log)

PR #4459, opened 2025-02-21
Add `FabricEntityLootTableProvider` for datagen

Antikyth:
Adds an equivalent to `FabricBlockLootTableProvider` for entity loot tables. `EntityLootTableGenerator` has some helper utilities and constants for entity loot tables which make it desirable to extend that class.

I would like this backported to 1.20.1; currently I am using an implementation like this in my mod, but it depends on the internal FAPI classes used here marked as non-extendable.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricEntityLootTableGenerator.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricEntityLootTableGenerator.java
@@ -1,0 +1,21 @@
+package net.fabricmc.fabric.api.datagen.v1.loot;
+
+import com.google.common.base.Preconditions;
+
+import net.fabricmc.fabric.api.resource.conditions.v1.ResourceCondition;
+
+import net.fabricmc.fabric.impl.datagen.loot.ConditionEntityLootTableGenerator;
+
+import net.minecraft.data.loottable.EntityLootTableGenerator;
+
+public interface FabricEntityLootTableGenerator {
+	/**
+	 * {@return a new generator that applies the specified conditions to any loot table it receives, and then forwards
+	 * the loot tables to this generator}
+	 */
+	default EntityLootTableGenerator withConditions(ResourceCondition... conditions) {
+		Preconditions.checkArgument(conditions.length > 0, "Must add at least one condition.");
+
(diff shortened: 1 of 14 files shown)


modmuss50:
Please make sure to add a test for this.

Antikyth:
> Please make sure to add a test for this.

Yep, will tomorrow, was working on that :)  Made this a draft until that is done.

Antikyth:
Realised I might wanna filter out empty loot tables from datagen, will have a look into that tomorrow, other than investigating that (if warranted it would be a very simple change) everything else here is finished.

Antikyth:
> Realised I might wanna filter out empty loot tables from datagen, will have a look into that tomorrow, other than investigating that (if warranted it would be a very simple change) everything else here is finished.

Neither the vanilla `EntityLootTableGenerator` nor `FabricBlockLootTableProvider` does this, so I won't. Confirms everything here is complete from my point of view, though feedback is welcome.

modmuss50:
Checkstyle still needs looking at, I suggest trying to run it locally so you can make sure everything is good before pushing the changes.

fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricEntityLootTableGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+package net.fabricmc.fabric.api.datagen.v1.loot;
+
+import com.google.common.base.Preconditions;
+
+import net.minecraft.data.loottable.EntityLootTableGenerator;
+
+import net.fabricmc.fabric.api.resource.conditions.v1.ResourceCondition;
+import net.fabricmc.fabric.impl.datagen.loot.ConditionEntityLootTableGenerator;
+
+public interface FabricEntityLootTableGenerator {
```
modmuss50, line 26: Nit: Put a comment saying that this interface is injected, see FabricBlockLootTableGenerator for an example.

modmuss50 (approved):
Approved assuming checkstyle issues will be fixed.

Antikyth:
Bit busy at the moment unfortunately, but I haven't forgotten about this PR. I'll try get those two issues fixed over the next ~4 days.

modmuss50:
Dont stress, I can easily sort this out if needed.

PR #4468, opened 2025-02-24
Check the Gradle distribution checksum

TheMrMilchmann:
It is generally a good idea to ensure the integrity of the Gradle distribution to prevent MITM attacks when downloading. (See https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification)

Keep in mind that this has to be updated alongside the distribution link.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionSha256Sum=f397b287023acdba1e9f6fc5ea72d22dd63669d59ed4a289a29b1a76eee151c6
 distributionUrl=https\://services.gradle.org/distributions/gradle-8.11.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true


modmuss50:
Spam account, making loads of PRs like this. Very suspicous.

There is no security benefit to this at all as we validate the wrapper, and download Gradle over https.

TheMrMilchmann:
> Spam account, making loads of PRs like this. Very suspicous.

lol

> There is no security benefit to this at all as we validate the wrapper, and download Gradle over https.

This is not true. Downloading over HTTPS does, for example, not protect against tampered distributions from compromised official sources. Even MITM attacks are still possible under specific circumstances (i.e. compromised CA authority). While this might not pose a large security risk in practice, the mitigation is basically free.

modmuss50:
If the CA is compromised the cert will be revoked via the CRL so that not a worry.
If Gradle was compromised they could still distribute code to run via plugins or the GHA action we use.

Does the Gradle wrapper validate the signature of the main package? That would be a suitable mitagation against this.

TheMrMilchmann:
> Does the Gradle wrapper validate the signature of the main package?

That's precisely what this signature is for. This is not the checksum of the wrapper but the Gradle distribution which gets downloaded by the wrapper.

[Gradle's official GH actions](https://github.com/gradle/actions/tree/main/setup-gradle) can be used to validate the wrapper. They contain a list of valid wrapper checksums and can be pinned against specific versions (just like any other GHA).

PR #4474, opened 2025-02-28
[1.21.5] Add back "invokeTestMethod" function to allow control over calling the test method.

modmuss50:
This is basically the same API in as in 1.21.4. Useful to call code before and after each test, or pass custom params.

### DIFF
--- a/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/CustomTestMethodInvoker.java
+++ b/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/CustomTestMethodInvoker.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.gametest.v1;
+
+import java.lang.reflect.Method;
+
+import net.minecraft.test.TestContext;
+
+/**
+ * Implement this interface on test suites to provide custom logic for invoking {@link GameTest} test methods.
+ */
+public interface CustomTestMethodInvoker {
+	/**
+	 * Implement this method to provide custom logic used to invoke the test method.
+	 * This can be used to run code before or after each test.
+	 * You can also pass in custom parameters into the test method if desired.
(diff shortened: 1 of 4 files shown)


fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/CustomTestMethodInvoker.java
```diff
@@ -0,0 +1,37 @@
...
+package net.fabricmc.fabric.api.gametest.v1;
+
+import java.lang.reflect.Method;
+
+import net.minecraft.test.TestContext;
+
+/**
+ * Implement this interface on test suites to provide custom logic for invoking {@link GameTest} test methods.
+ */
+public interface CustomTestMethodInvoker {
```
Pyrofab, line 26: Could maybe add a deprecated `FabricGameTest` interface that subclasses this interface, so that people can easily find the right migration ?

Pyrofab (approved):
Looks good, thanks for bringing it back. Possible future improvement would be magic `@BeforeEachGametest`/`@AfterEachGametest` annotations, for the most common use case, but it's a functionality that can already be implemented very easily with this interface.

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/CustomTestMethodInvoker.java
modmuss50: Humm, im not sure how useful that would be. This module was basically rewritten from scratch, I dont think we should aim to add new deprecated code just to aid with porting. Chances are we would need to carry it around for years if we did.

modmuss50:
> Looks good, thanks for bringing it back. Possible future improvement would be magic `@BeforeEachGametest`/`@AfterEachGametest` annotations, for the most common use case, but it's a functionality that can already be implemented very easily with this interface.

Im generally not a massive fan of magic annotations, as you said its easy enough to achieve with this so im not sure its something we really need.

Vanilla does also have a way to do this by creating a custom test environment with `setup` and `teardown` functions.

fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/impl/gametest/TestAnnotationLocator.java
```diff
@@ -82,7 +83,11 @@ private List<TestMethod> findMagicMethods(EntrypointContainer<Object> entrypoint
 	private void findMagicMethods(EntrypointContainer<Object> entrypoint, Class<?> testClass, List<TestMethod> methods) {
 		for (Method method : testClass.getDeclaredMethods()) {
 			if (method.isAnnotationPresent(GameTest.class)) {
-				validateMethod(method);
+				if (!CustomTestMethodInvoker.class.isAssignableFrom(testClass)) {
+					// Only validate the test method when using the default reflection invoker
+					validateMethod(method);
```
modmuss50, line 88: Note: add the class name in validateMethod

Pyrofab: Another note : the warning in `findMagicMethods` mentions a `FabricGameTest` annotation, but the annotation's actual name is `GameTest`

PR #4483, opened 2025-03-08
AttSync: Use correct event for transferring attachments during respawn

### LINKED ISSUE #4475: Player attachments not syncing to client properly on respawn
While developing a mod using the v1 attachments API I've encountered a bug.
Where the server after respawing a player doesn't properly sync attachments to the client.
I believe it only affects client side, generating a stack trace for each attachment, linked below.
I suspect that the packet is sent during the wrong client phase.
I'm happy to provide more relevant information if needed. 

Below I include the versions I use, logs, and a workaround I made, that doesn't address the root of the problem, and is only a band-aid solution.

Versions:
Minecraft - 1.21.4
Yarn - 1.21.4+build.8
Fabric Loader - 0.16.10
Fabric API - 0.118.0+1.21.4

A very basic workaround:
I set a constant delay of 100ms which fixes the issue
https://gist.github.com/bejker123/ff2daeeef96fe824d24e4c87b6a2004e

Relevant Logs:
https://gist.github.com/bejker123/a38d08dd622ff4fafd0076616696563d

Octol1ttle:
Closes #4475
cc @Syst3ms @bejker123

**What does this PR do?**
This PR changes the event used to transfer attachments when a player is respawned from `COPY_FROM` to `AFTER_RESPAWN`

**Why did this issue happen?**
To understand what exactly happened, we will need to trace our steps from the `COPY_FROM` event to it's caller.
1) The event is called at the [`TAIL` of `ServerPlayerEntity#copyFrom`](https://github.com/FabricMC/fabric/blob/c0029d6179011efb655905b57a22e2d234e5ec47/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/ServerPlayerEntityMixin.java#L88)
2) The `copyFrom` method is called by `PlayerManager#respawnPlayer`:
image
Let's take a closer look at line 403: `player.getServerWorld().removePlayer(player, removalReason);`. The method will call `Entity#remove` on the old player, which will call `Entity#changeListener#remove`. 
3) The server's `changeListener` is implemented in `ServerEntityManager`:
image
The server will **stop tracking** the old player entity. There's a long chain of method calls that I won't cover in detail because otherwise this PR will be 14 pages long:
`ServerEntityManager#stopTracking` -> `ServerWorld#stopTracking` -> `ServerChunkManager#unloadEntity` -> `ServerChunkLoadingManager#unloadEntity` -> `ServerChunkLoadingManager#stopTracking` -> `EntityTrackerEntry#stopTracking` -> --hold on, let's stop here.
4) `EntityTrackerEntry#stopTracking`
image
Here's our culprit! The server will order the clients to **destroy the old player entity** before calling `copyFrom`, and `copyFrom` is called before the new player is spawned in (this happens when `ServerWorld#onPlayerRespawned` is called at `PlayerManager#respawnPlayer#L439` is called).
When the client receives the attachment sync packet, **the old entity is destroyed, but the new one hasn't been received yet**, so `AttachmentTargetInfo.getTarget` returns null, causing the issue.

**Why does `AFTER_RESPAWN` work?**
It's simple. `AFTER_RESPAWN` is called at the [`TAIL` of `PlayerManager#respawnPlayer`](https://github.com/FabricMC/fabric/blob/c0029d6179011efb655905b57a22e2d234e5ec47/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/PlayerManagerMixin.java#L32), and as I've said before, that method will call `ServerWorld#onPlayerRespawned`, adding the new player entity to the world and tracking it (syncing to clients), so now AttSync can add the attachments to the new player entity.

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentEntrypoint.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentEntrypoint.java
@@ -29,7 +29,7 @@ public class AttachmentEntrypoint implements ModInitializer {
 
 	@Override
 	public void onInitialize() {
-		ServerPlayerEvents.COPY_FROM.register((oldPlayer, newPlayer, alive) ->
+		ServerPlayerEvents.AFTER_RESPAWN.register((oldPlayer, newPlayer, alive) ->
 				AttachmentTargetImpl.transfer(oldPlayer, newPlayer, !alive)
 		);
 		ServerEntityWorldChangeEvents.AFTER_ENTITY_CHANGE_WORLD.register(((originalEntity, newEntity, origin, destination) ->


modmuss50 (approved):
Seems fine to me.

Syst3ms (approved):

PR #4488, opened 2025-03-08
Port FRAPI to 1.21.5

### LINKED ISSUE #3161: Request: Expose block particles to FabricBakedModel
`FabricBakedModel` defines a level-, pos-, and state-aware `emitBlockQuads`, and I am requesting that it also expose a level-, pos-, and state-aware version of `BakedModel#getParticleIcon`.

The use-case here is for "carpenter's blocks"-ish mods: it's very desirable to make the block's breaking, sprinting, and landing particles match the retextured model. It's typically not hard to implement this with [a handful of mixins](https://github.com/quat1024/packages/tree/trunk/Fabric/src/main/java/agency/highlysuspect/packages/platform/fabric/mixin/particleslol), but it'd be nice if this was part of `FabricBakedModel`, both to keep everything about the model in one place, and to potentially make it composable with other models w/ custom sprites (my simple mixins just end up calling the zero-arg vanilla `getParticleIcon` again). And, yknow, it'd be nice.

For comparison, [Forge has an API like this on `IForgeBakedModel`](https://github.com/MinecraftForge/MinecraftForge/blob/7d0330eb08299935714e34ac651a293e2609aa86/src/main/java/net/minecraftforge/client/extensions/IForgeBakedModel.java#L74-L77). I think a simpler `(level, pos, state) -> TextureAtlasSprite`-shaped API would work fine too.

### LINKED ISSUE #4017: Renderer API does not support Block Display Entities.
I was testing block display entities with the Renderer API lately. and it seems that `emitBlockQuads` hasn't been implemented for block display entities.
https://i.imgur.com/KlHWC2w.gif

PepperCode1:
Fix #4017.
Resolve #3161.


modmuss50 (approved):
Many thanks

PR #4500, opened 2025-03-14
Fix BlockView API v2 not applying `WorldViewMixin`

### LINKED ISSUE #4499: BlockView API v2 does not apply its `WorldViewMixin`
An entry for this mixin is missing in the module's mixins.json file.

PepperCode1:
Fix #4499.

 This fix should be backported to as many versions as possible.


modmuss50 (approved):

modmuss50:
This proves why we need tests, and should ideally be adding one. Im happy to merge this as is as its clearly broken.

PR #4528, opened 2025-03-22
[1.21.5] Remove deprecated BlockOutlineContext#vertexConsumer & Pass translucent parameter to BEFORE_BLOCK_OUTLINE event

modmuss50:
- Remove deprecated BlockOutlineContext#vertexConsumer
- Pass translucent parameter to BEFORE_BLOCK_OUTLINE event

Taken from: https://github.com/FabricMC/fabric/pull/4286/commits/2f117f74eb7948489e0293ef2fbd963a3587cd92

Co-authored-by: PepperCode1 <44146161+peppercode1@users.noreply.github.com>

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/WorldRenderContext.java
@@ -24,7 +24,6 @@ import net.minecraft.client.render.Camera;
 import net.minecraft.client.render.Frustum;
 import net.minecraft.client.render.GameRenderer;
 import net.minecraft.client.render.RenderTickCounter;
-import net.minecraft.client.render.VertexConsumer;
 import net.minecraft.client.render.VertexConsumerProvider;
 import net.minecraft.client.render.WorldRenderer;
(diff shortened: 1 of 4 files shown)


PepperCode1 (changes requested):
- `BlockOutline` should also receive the `translucent` parameter.
- `translucentBlockOutline()` should be removed from the context.
- It may be worth inlining `BlockOutlineContext` as `Camera, BlockPos, BlockState` parameters, since only one callback receives it.

modmuss50:
> * `BlockOutline` should also receive the `translucent` parameter.
> * `translucentBlockOutline()` should be removed from the context.
> * It may be worth inlining `BlockOutlineContext` as `Camera, BlockPos, BlockState` parameters, since only one callback receives it.

Im a little confused, isnt `translucentBlockOutline()` the same? And you get that from the context anyway. Looking at it more im not even sure what this PR achives as it was already passed in via the context?

modmuss50:
This needs to be discussed some more, the reason for having the context object is to allow expanding whats passed through without breaking the API. I dont think there is an immediate rush for these changes. My bad for not looking at it more before opening the PR.

modmuss50:
Closing this PR for now, as discussed we will iterate on this API in a none breaking way.

PR #4538, opened 2025-03-26
AttSync: Disconnect when attachment target is null

Octol1ttle:
Example:
image

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentChange.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentChange.java
@@ -138,10 +138,10 @@ public record AttachmentChange(AttachmentTargetInfo<?> targetInfo, AttachmentTyp
 		AttachmentTarget target = targetInfo.getTarget(world);
 		Object value = decodeValue(world.getRegistryManager());
 		if (target == null) {
-			final var errorMessageBuilder = new StringBuilder("Received attachment change for unknown target.");
-			errorMessageBuilder.append("Attachment identifier: ").append(type.identifier());
-			errorMessageBuilder.append("Attachment value: ").append(value);
-			errorMessageBuilder.append("World: ").append(world.getRegistryKey());
+			final var errorMessageBuilder = new StringBuilder("Received attachment change for unknown target.").append('\n');
+			errorMessageBuilder.append("Attachment identifier: ").append(type.identifier()).append('\n');
+			errorMessageBuilder.append("Attachment value: ").append(value).append('\n');
+			errorMessageBuilder.append("World: ").append(world.getRegistryKey()).append('\n');
 			targetInfo.appendDebugInformation(errorMessageBuilder);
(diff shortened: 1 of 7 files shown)


fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentChange.java
```diff
@@ -132,7 +139,26 @@ public Object decodeValue(DynamicRegistryManager dynamicRegistryManager) {
...
+			errorMessageText.append(ScreenTexts.LINE_BREAK);
+
+			errorMessageText.append("Attachment identifier: ").append(Text.literal(String.valueOf(type.identifier())).formatted(Formatting.YELLOW)).append(ScreenTexts.LINE_BREAK);
+			errorMessageText.append("World: ").append(Text.literal(String.valueOf(world.getRegistryKey().getValue())).formatted(Formatting.YELLOW)).append(ScreenTexts.LINE_BREAK);
+			targetInfo.appendDebugInformation(errorMessageText);
+
+			AttachmentEntrypoint.LOGGER.error(errorMessageText.getString());
+			sender.disconnect(errorMessageText);
+
+			return false;
```
modmuss50: I think this would be better of throwing a custom exception that passes the Text up to reciever where you can handle the disconnect. (see RemapException as an example)

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentChange.java
```diff
@@ -132,7 +139,26 @@ public Object decodeValue(DynamicRegistryManager dynamicRegistryManager) {
...
 
-	public void apply(World world) {
-		targetInfo.getTarget(world).setAttached((AttachmentType<Object>) type, decodeValue(world.getRegistryManager()));
+	public boolean tryApply(World world, PacketSender sender) {
+		AttachmentTarget target = targetInfo.getTarget(world);
+		Object value = decodeValue(world.getRegistryManager());
+
+		if (target == null) {
+			final MutableText errorMessageText = Text.empty();
+			errorMessageText.append(Text.literal("Received attachment change for unknown target!").formatted(Formatting.RED)).append(ScreenTexts.LINE_BREAK);
```
modmuss50: These can be translated as they are on the client.

modmuss50 (commented):
This needs some form of testing, ideally automated if not a command that tried to do something funny would be fine.

modmuss50:
Hey @Octol1ttle no pressure, just wondering if you are planning on finishing this off, its so close to being done. Im happy to help/take over if you like.

Octol1ttle:
Sorry, this PR completely left my memory. I resolved the review comments, but I have zero idea on how to do the testing bit 😅

modmuss50:
Awesome, thanks for that. I'll take a look a testing, worse case I'll just do some manual testing.

modmuss50:
I have added a basic unit test, and marked the file to be translated.

modmuss50 (approved):

Issue #4507, opened 2025-03-16
Add ServerChunkEvents.CHUNK_STATUS_CHANGE


DennisOchulor:
(Inspired by [this](https://discord.com/channels/507304429255393322/566276937035546624/1349456652939563088) Discord conversation in the fabric discord)

I would like to suggest the addition of a new event, `ServerChunkEvents.CHUNK_STATUS_CHANGE` which fires each time a chunk changes its [load type](https://minecraft.wiki/w/Chunk#Level_and_load_type) (ENTITY_TICKING, BLOCK_TICKING, FULL, INACCESSIBLE).

This is useful because `CHUNK_UNLOAD` is only fired when the load level of a chunk goes above 44, which is 11 levels after INACCESSIBLE. Some mods (including my mod) would like to know when a chunk is no longer accessible even if it is technically still "loaded".

Alongside this, the documentation of `CHUNK_UNLOAD` should be updated to clarify when Minecraft actually unloads a chunk, specifically that chunks do not unload immediately after becoming INACCESSIBLE, which is the main source of the confusion.

DennisOchulor:
Resolved in #4541

PR #4541, opened 2025-03-28
Add ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE

DennisOchulor:
Add a new event that fires when a chunk's level type changes. (Relates to #4507)

Changed the name from `CHUNK_STATUS_CHANGE` as to not confuse it with the `ChunkStatus` class which refers to chunk generation stages.

The event only provides a `ChunkPos` and not a `WorldChunk`. If really needed, a potentially null `WorldChunk` can be obtained via `ChunkHolder#getLatest()` but I do not think it's necessary.

The documentation of `CHUNK_UNLOAD` has also been updated to clarify server chunk unload timing.

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -67,7 +67,10 @@ public final class ServerChunkEvents {
 	/**
 	 * Called when a chunk changes its {@link ChunkLevelType}.
 	 *
-	 * <p>When this event is called, the chunk's level type has already changed.
+	 * <p>For chunk level type promotions, this event is called only when the chunk's behavior actually aligns with its level type.
+	 * This means the event is usually called later than when the chunk's level type first gets promoted.
+	 *
+	 * <p>For chunk level type demotions, this event is called immediately when the chunk's level type first gets demoted.
 	 */
 	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, callbacks -> (serverWorld, chunkPos, oldLevelType, newLevelType) -> {
 		for (LevelTypeChange callback : callbacks) {
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -67,7 +67,7 @@ public final class ServerChunkEvents {
 	 * Called when a chunk changes its {@link ChunkLevelType}.
 	 *
 	 * <p>When this event is called, the chunk's {@link WorldChunk#getLevelType()} has already changed. However, the chunk's actual ticking behavior may not fully align with its level type yet.
-	 * Additionally, it is not guaranteed that entities from the chunk are immediately accessible when this event is called, though they are already loaded.
+	 * Additionally, it is not guaranteed that entities from the chunk are immediately accessible when this event is called.
 	 */
 	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, (world, chunk, oldLevelType, newLevelType) -> { }, callbacks -> (serverWorld, chunk, oldLevelType, newLevelType) -> {
 		for (LevelTypeChange callback : callbacks) {
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -16,7 +16,9 @@
 
 package net.fabricmc.fabric.api.event.lifecycle.v1;
 
+import net.minecraft.server.world.ChunkLevelType;
 import net.minecraft.server.world.ServerWorld;
+import net.minecraft.util.math.ChunkPos;
 import net.minecraft.world.chunk.WorldChunk;
 
 import net.fabricmc.fabric.api.event.Event;
@@ -59,6 +61,17 @@ public final class ServerChunkEvents {
 		}
 	});
 
+	/**
+	 * Called when a chunk changes its {@link ChunkLevelType}.
+	 *
+	 * <p>When this event is called, the chunk's level type has already changed.
+	 */
+	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, callbacks -> (serverWorld, chunkPos, oldLevelType, newLevelType) -> {
+		for (LevelTypeChange callback : callbacks) {
+			callback.onChunkLevelTypeChange(serverWorld, chunkPos, oldLevelType, newLevelType);
+		}
+	});
+
 	@FunctionalInterface
 	public interface Load {
 		void onChunkLoad(ServerWorld world, WorldChunk chunk);
@@ -72,5 +85,10 @@ public final class ServerChunkEvents {
 	@FunctionalInterface
 	public interface Unload {
 		void onChunkUnload(ServerWorld world, WorldChunk chunk);
+	}
+
+	@FunctionalInterface
+	public interface LevelTypeChange {
+		void onChunkLevelTypeChange(ServerWorld world, ChunkPos chunkPos, ChunkLevelType oldLevelType, ChunkLevelType newLevelType);
 	}
 }
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -18,7 +18,6 @@ package net.fabricmc.fabric.api.event.lifecycle.v1;
 
 import net.minecraft.server.world.ChunkLevelType;
 import net.minecraft.server.world.ServerWorld;
-import net.minecraft.util.math.ChunkPos;
 import net.minecraft.world.chunk.WorldChunk;
 
 import net.fabricmc.fabric.api.event.Event;
@@ -67,14 +66,12 @@ public final class ServerChunkEvents {
 	/**
 	 * Called when a chunk changes its {@link ChunkLevelType}.
 	 *
-	 * <p>For chunk level type promotions, this event is called only when the chunk's behavior actually aligns with its level type.
-	 * This means the event is usually called later than when the chunk's level type first gets promoted.
-	 *
-	 * <p>For chunk level type demotions, this event is called immediately when the chunk's level type first gets demoted.
+	 * <p>When this event is called, the chunk's {@link WorldChunk#getLevelType()} has already changed. However, the chunk's actual ticking behavior may not fully align with its level type yet.
+	 * Additionally, it is not guaranteed that entities from the chunk are immediately accessible when this event is called, though they are already loaded.
 	 */
-	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, callbacks -> (serverWorld, chunkPos, oldLevelType, newLevelType) -> {
+	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, (world, chunk, oldLevelType, newLevelType) -> { }, callbacks -> (serverWorld, chunk, oldLevelType, newLevelType) -> {
 		for (LevelTypeChange callback : callbacks) {
-			callback.onChunkLevelTypeChange(serverWorld, chunkPos, oldLevelType, newLevelType);
+			callback.onChunkLevelTypeChange(serverWorld, chunk, oldLevelType, newLevelType);
 		}
 	});
 
@@ -95,6 +92,6 @@ public final class ServerChunkEvents {
 
 	@FunctionalInterface
 	public interface LevelTypeChange {
-		void onChunkLevelTypeChange(ServerWorld world, ChunkPos chunkPos, ChunkLevelType oldLevelType, ChunkLevelType newLevelType);
+		void onChunkLevelTypeChange(ServerWorld world, WorldChunk chunk, ChunkLevelType oldLevelType, ChunkLevelType newLevelType);
 	}
 }
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -16,8 +16,11 @@
 
 package net.fabricmc.fabric.api.event.lifecycle.v1;
 
+import net.minecraft.server.world.ChunkHolder;
 import net.minecraft.server.world.ChunkLevelType;
+import net.minecraft.server.world.ServerChunkManager;
 import net.minecraft.server.world.ServerWorld;
+import net.minecraft.world.chunk.ChunkStatus;
 import net.minecraft.world.chunk.WorldChunk;
 
 import net.fabricmc.fabric.api.event.Event;
@@ -31,6 +34,10 @@ public final class ServerChunkEvents {
 	 * Called when a chunk is loaded into a ServerWorld.
 	 *
 	 * <p>When this event is called, the chunk is already in the world.
+	 *
+	 * <p>Note that this event is not called for chunks that become accessible without previously being unloaded.
+	 *
+	 * @see ServerChunkEvents#CHUNK_LEVEL_TYPE_CHANGE
 	 */
 	public static final Event<ServerChunkEvents.Load> CHUNK_LOAD = EventFactory.createArrayBacked(ServerChunkEvents.Load.class, callbacks -> (serverWorld, chunk) -> {
 		for (Load callback : callbacks) {
@@ -64,10 +71,15 @@ public final class ServerChunkEvents {
 	});
 
 	/**
-	 * Called when a chunk changes its {@link ChunkLevelType}.
+	 * Called when a chunk's actual ticking behavior is about to align with its updated {@link ChunkLevelType}.
 	 *
-	 * <p>When this event is called, the chunk's {@link WorldChunk#getLevelType()} has already changed. However, the chunk's actual ticking behavior may not fully align with its level type yet.
-	 * Additionally, it is not guaranteed that entities from the chunk are immediately accessible when this event is called.
+	 * <p>When this event is being called:
+	 * <ul>
+	 * <li>The chunk's {@link WorldChunk#getLevelType()} has already changed.</li>
+	 * <li>Entities within the chunk are not guaranteed to be accessible.</li>
+	 * <li>The chunk's corresponding level type future in {@link ChunkHolder} is not guaranteed to be done.</li>
+	 * <li>When transitioning from {@link ChunkLevelType#INACCESSIBLE} to {@link ChunkLevelType#FULL}, calling {@link ServerChunkManager#getChunkFutureSyncOnMainThread(int, int, ChunkStatus, boolean)} to fetch the current chunk at {@link ChunkStatus#FULL} status results in undefined behavior.</li>
+	 * </ul>
 	 */
 	public static final Event<LevelTypeChange> CHUNK_LEVEL_TYPE_CHANGE = EventFactory.createArrayBacked(LevelTypeChange.class, (world, chunk, oldLevelType, newLevelType) -> { }, callbacks -> (serverWorld, chunk, oldLevelType, newLevelType) -> {
 		for (LevelTypeChange callback : callbacks) {
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
@@ -54,6 +54,9 @@ public final class ServerChunkEvents {
 	 * Called when a chunk is unloaded from a ServerWorld.
 	 *
 	 * <p>When this event is called, the chunk is still present in the world.
+	 *
+	 * <p>Note that the server typically unloads chunks when the chunk's level goes above 45 (and not immediately when the chunk becomes inaccessible).
+	 * To know when a chunk first becomes inaccessible, see {@link ServerChunkEvents#CHUNK_LEVEL_TYPE_CHANGE}.
 	 */
 	public static final Event<ServerChunkEvents.Unload> CHUNK_UNLOAD = EventFactory.createArrayBacked(ServerChunkEvents.Unload.class, callbacks -> (serverWorld, chunk) -> {
 		for (Unload callback : callbacks) {
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
@@ -63,13 +63,13 @@ public abstract class ChunkHolderMixin extends AbstractChunkHolder implements Ch
 	protected abstract void combineSavingFuture(CompletableFuture<?> savingFuture);
 
 	@Unique
-	private static final ChunkLevelType[] CHUNK_LEVEL_TYPES = ChunkLevelType.values(); // values() clones the internal array each call, so cache the return
+	private static final ChunkLevelType[] fabric_CHUNK_LEVEL_TYPES = ChunkLevelType.values(); // values() clones the internal array each call, so cache the return
 
 	@Unique
-	private CompletableFuture<Void> fullToBlockTickingEvent;
+	private CompletableFuture<Void> fabric_fullToBlockTickingEvent;
 
 	@Unique
-	private ChunkLevelType currentEventLevelType = INACCESSIBLE;
+	private ChunkLevelType fabric_currentEventLevelType = INACCESSIBLE;
 
 	private ChunkHolderMixin(ChunkPos pos) {
 		super(pos);
@@ -80,9 +80,9 @@ public abstract class ChunkHolderMixin extends AbstractChunkHolder implements Ch
 	 */
 	@Inject(method = "updateFutures", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/world/ChunkHolder;combineSavingFuture(Ljava/util/concurrent/CompletableFuture;)V", shift = At.Shift.AFTER, ordinal = 0))
 	private void updateFutures$inaccessibleToFull(ServerChunkLoadingManager chunkLoadingManager, Executor executor, CallbackInfo ci) {
-		if (this.getUncheckedOrNull(ChunkStatus.FULL) instanceof WorldChunk && this.currentEventLevelType == INACCESSIBLE) { // prevent duplicate events with ChunkGeneratingMixin
+		if (this.getUncheckedOrNull(ChunkStatus.FULL) instanceof WorldChunk && this.fabric_currentEventLevelType == INACCESSIBLE) { // prevent duplicate events with ChunkGeneratingMixin
 			ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, INACCESSIBLE, FULL);
-			this.currentEventLevelType = FULL;
+			this.fabric_currentEventLevelType = FULL;
 		}
 	}
 
@@ -91,22 +91,22 @@ public abstract class ChunkHolderMixin extends AbstractChunkHolder implements Ch
 	 */
 	@Inject(method = "updateFutures", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/world/ChunkHolder;combineSavingFuture(Ljava/util/concurrent/CompletableFuture;)V", shift = At.Shift.AFTER, ordinal = 1))
 	private void updateFutures$fullToBlockTicking(ServerChunkLoadingManager chunkLoadingManager, Executor executor, CallbackInfo ci) {
-		this.fullToBlockTickingEvent = this.tickingFuture.thenAccept(optionalChunk -> {
+		this.fabric_fullToBlockTickingEvent = this.tickingFuture.thenAccept(optionalChunk -> {
 			optionalChunk.ifPresent(worldChunk -> {
 				ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, FULL, BLOCK_TICKING);
-				this.currentEventLevelType = BLOCK_TICKING;
+				this.fabric_currentEventLevelType = BLOCK_TICKING;
 			});
 		});
 	}
 
 	/**
-	 * Handles FULL -> BLOCK_TICKING.
+	 * Handles BLOCK_TICKING -> ENTITY_TICKING.
 	 */
 	@Inject(method = "updateFutures", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/world/ChunkHolder;combineSavingFuture(Ljava/util/concurrent/CompletableFuture;)V", shift = At.Shift.AFTER, ordinal = 2))
 	private void updateFutures$blockTickingToEntityTicking(ServerChunkLoadingManager chunkLoadingManager, Executor executor, CallbackInfo ci) {
-		this.fullToBlockTickingEvent.thenAccept(optionalChunk -> {
+		this.fabric_fullToBlockTickingEvent.thenAccept(optionalChunk -> {
 			ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, BLOCK_TICKING, ENTITY_TICKING);
-			this.currentEventLevelType = ENTITY_TICKING;
+			this.fabric_currentEventLevelType = ENTITY_TICKING;
 		});
 	}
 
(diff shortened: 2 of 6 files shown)


fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,64 @@
...
+	@Shadow
+	private int lastTickLevel;
+
+	private ChunkHolderMixin(ChunkPos pos) {
+		super(pos);
+	}
+
+	@Inject(method = "increaseLevel", at = @At("TAIL"))
+	private void increaseLevel(ServerChunkLoadingManager chunkLoadingManager, CompletableFuture<OptionalChunk<WorldChunk>> chunkFuture, Executor executor, ChunkLevelType target, CallbackInfo ci) {
+		ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, ChunkLevels.getType(this.lastTickLevel), target);
```
2No2Name: On which thread should this be called, and should this be ensured with a similar check as in 
https://github.com/CaffeineMC/lithium/blob/ebcbdb134365ab1862461548f857cdecf96e4d31/common/src/main/java/net/caffeinemc/mods/lithium/common/world/chunk/ChunkStatusTracker.java#L48

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	@Inject(method = "decreaseLevel", at = @At("TAIL"))
+	private void decreaseLevel(ServerChunkLoadingManager chunkLoadingManager, ChunkLevelType target, CallbackInfo ci) {
+		ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, ChunkLevels.getType(this.lastTickLevel), target);
```
2No2Name, line 62: Also consider the thread here, burdening mod authors with changing to the server thread to avoid threading issues is probably not appropriate

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
```diff
@@ -52,13 +54,27 @@ private ServerChunkEvents() {
...
 	public static final Event<ServerChunkEvents.Unload> CHUNK_UNLOAD = EventFactory.createArrayBacked(ServerChunkEvents.Unload.class, callbacks -> (serverWorld, chunk) -> {
 		for (Unload callback : callbacks) {
 			callback.onChunkUnload(serverWorld, chunk);
 		}
 	});
 
+	/**
+	 * Called when a chunk changes its {@link ChunkLevelType}.
+	 *
+	 * <p>When this event is called, the chunk's level type has already changed.
```
2No2Name: Which meaning does changing the chunk level type have?
There might be many misconceptions about this, leading to lots of incorrect implementations. The code comment is very minimal and doesn't indicate any potential issues.

For example, if a mod author believes the chunk status indicates one of the following, there will likely be major issues:

Whether the chunk is loaded, whether entities are ticking in that chunk, whether entities are loaded in that chunk, whether the chunk is ticking blocks in that chunk, etc.

2No2Name (commented):
Edit: I am probably wrong about the threads that the level changes take place, since I was confusing this with chunks actually becoming available to the world, including the difference of getChunk and getChunkNow.

I think implementing this without major changes is a bad idea due to the missing explanation of the meaning of the chunk level type changes.
I believe that if understood correctly, the chunk level type is NOT what you want to track, as the status that is being tracked here only describes whether the game aims to load a chunk, but not whether the chunk is actually being loaded or available to the general game logic.

In my mod, lithium, which tracks whether a chunk is accessible in the world, I had several issues tracking this status correctly. I managed to get all events fire on the server thread, to avoid threading issues. Furthermore, Lithium (almost correctly) tracks whether the chunk is accessible in the world, instead of whether the game tries to make it accessible (status set but chunk future not done yet). I believe that if the API just tracks the level/status, many authors will incorrectly assume that this information is sufficient to know that a chunk is available in the world.

Including the other load statuses, especially regarding entity ticking, adds another layer of complexity and unclear meanings, requiring even more clear documentation.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,64 @@
...
+		super(pos);
+	}
+
+	@Inject(method = "increaseLevel", at = @At("TAIL"))
+	private void increaseLevel(ServerChunkLoadingManager chunkLoadingManager, CompletableFuture<OptionalChunk<WorldChunk>> chunkFuture, Executor executor, ChunkLevelType target, CallbackInfo ci) {
+		ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, ChunkLevels.getType(this.lastTickLevel), target);
+	}
+
+	@Inject(method = "decreaseLevel", at = @At("TAIL"))
+	private void decreaseLevel(ServerChunkLoadingManager chunkLoadingManager, ChunkLevelType target, CallbackInfo ci) {
```
ishland, line 104: This method is only called once if a chunk changes from entity ticking to inaccessible in a single tick. This will make event users to only get a `INACCESSIBLE` when such transition happens. The event should be called for all intermediate states to avoid confusion.

DennisOchulor: Does this mean there should be a check to see if the old/new chunk levels are sequential, and if not artificially invoke the event for all the intermediate level types in order?

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
DennisOchulor: I've done some testing and it does seem like the event is always consistently called on the main server thread.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
ishland: > Does this mean there should be a check to see if the old/new chunk levels are sequential, and if not artificially invoke the event for all the intermediate level types in order?

Yes.

DennisOchulor:
This is a side note but `increaseLevel()` and `decreaseLevel()` are named backwards. 

`increaseLevel()` is called when the chunk level type changes from:
- INACCESSIBLE(34+) to FULL(33)
- FULL(33) to BLOCK_TICKING(32)
- BLOCK_TICKING(32) to ENTITY_TICKING(31 and below)

These are all level _decreases_, and `decreaseLevel()` does the reverse of this (but potentially skips intermediate level types).

I don't think this changes the actual functionality of the event but just pointing it out.

2No2Name:
Rewording my concerns: 
When the chunk loading isn't done yet by the worker thread, the chunk level is already set to something accessible but the chunk actually isn't accessible. I would really like the API to not just track the chunk holder's chunk level type but the actual behavior of the chunk, but that is very complex and can include threading issues.

DennisOchulor:
The event is now called when the chunk future has actually completed, and is called on the main thread's executor. I added some rough logging tests to verify the chunk's behaviour aligns with the level type but I'm honestly not sure how to test such a thing correctly.

This has been implemented but there is a potential issue:

> This method is only called once if a chunk changes from entity ticking to inaccessible in a single tick. This will make event users to only get a INACCESSIBLE when such transition happens. The event should be called for all intermediate states to avoid confusion.

If a chunk goes straight from ENTITY_TICKING to INACCESSIBLE, and all the intermediate level types are artificially called, then the calls for ENTITY_TICKING -> BLOCK_TICKING and BLOCK_TICKING -> FULL may make modders assume the chunk is still accessible when it really isn't.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
DennisOchulor: I've updated the documentation of the event to reflect the changes.

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
```diff
@@ -55,4 +59,42 @@ private static void setupChunkGenerateTest() {
...
+	 * While the world is loading in, this will log a few times.
+	 * Once all chunks within (and just outside) simulation distance have loaded in, logging stops.
+	 * Moving around within the same chunk (use F3+G) should not log anything.
+	 * Moving into another chunk should trigger some logs.
+	 */
+	private static void setupChunkLevelTypeChangeTest() {
+		final Object2ObjectMap<Identifier, Object2IntMap<ChunkLevelType>> worlds = new Object2ObjectOpenHashMap<>();
+		ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.register((world, chunkPos, oldLevelType, newLevelType) -> {
+			if (Thread.currentThread() != world.getServer().getThread()) {
+				LOGGER.error("{} {} -> {} NOT ON SERVER THREAD", chunkPos, oldLevelType, newLevelType);
```
modmuss50: Crash the game.

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
```diff
@@ -55,4 +59,42 @@ private static void setupChunkGenerateTest() {
...
+			}
+
+			if (oldLevelType == newLevelType) {
+				LOGGER.error("{} {} SAME LEVEL TYPE", chunkPos, oldLevelType);
```
modmuss50: again, if its not supposed to happen crash the game.

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
```diff
@@ -55,4 +59,42 @@ private static void setupChunkGenerateTest() {
...
+			}
+
+			worlds.putIfAbsent(world.getRegistryKey().getValue(), new Object2IntOpenHashMap<>());
+			worlds.get(world.getRegistryKey().getValue()).mergeInt(newLevelType, 1, Integer::sum);
```
modmuss50: Nit: use computeIfAbsent instead of a 2nd get.

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
```diff
@@ -55,4 +59,42 @@ private static void setupChunkGenerateTest() {
...
+		});
+
+		ServerTickEvents.END_WORLD_TICK.register(world -> {
+			if (world.getTime() % 20 != 0) { // limit to 1 per second
+				return;
+			}
+
+			worlds.putIfAbsent(world.getRegistryKey().getValue(), new Object2IntOpenHashMap<>());
```
modmuss50: Why do we need to put here? just null check the return from worlds.get

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
```diff
@@ -55,4 +59,42 @@ private static void setupChunkGenerateTest() {
...
+			Object2IntMap<ChunkLevelType> levelTypes = worlds.get(world.getRegistryKey().getValue());
+
+			if (!levelTypes.isEmpty()) {
+				StringBuilder sb = new StringBuilder(world.getRegistryKey().getValue() + " ");
+				levelTypes.forEach((levelType, integer) -> sb.append(levelType).append(": ").append(integer).append(", "));
```
modmuss50: Its not immediately clear what we are logging here? Is it the number of level type changes in the last second?

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,110 @@
...
+
+		Runnable runnable = () -> {
+			if (this.getLevelType().isAfter(target)) { // If chunk level type got demoted before promotion event fires, then don't fire it. This almost never happens tho.
+				ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange(serverWorld, this.pos, previous, target);
+			}
+		};
+
+		chunkFuture.thenAccept(optionalChunk -> optionalChunk.ifPresent(worldChunk -> {
+			if (!server.isOnThread()) {
+				server.send(new ServerTask(server.getTicks()-10, runnable)); // ensure the server thread runs it within this tick
```
modmuss50: Whats the -10 for?

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,110 @@
...
+		MinecraftServer server = serverWorld.getServer();
+		ChunkLevelType previous = switch (target) {
+		case INACCESSIBLE -> throw new IllegalStateException("target must at least be FULL");
+		case FULL -> INACCESSIBLE;
+		case BLOCK_TICKING -> FULL;
+		case ENTITY_TICKING -> BLOCK_TICKING;
+		};
+
+		Runnable runnable = () -> {
+			if (this.getLevelType().isAfter(target)) { // If chunk level type got demoted before promotion event fires, then don't fire it. This almost never happens tho.
```
modmuss50: Is this hiding a bug? Is this a timing issue as you get the levelType in a later tick than when the task was created?

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Really means increase level (chunk load type demotion).
+	 */
+	@Inject(method = "decreaseLevel", at = @At("TAIL"))
+	private void decreaseLevel(ServerChunkLoadingManager chunkLoadingManager, ChunkLevelType target, CallbackInfo ci) {
+		ChunkLevelType previous = ChunkLevels.getType(this.lastTickLevel);
+
+		if (previous.isAfter(ENTITY_TICKING) && !target.isAfter(ENTITY_TICKING)) {
```
modmuss50: I wonder if this can be done using a fori loop starting at the index of the previous enum entry?

modmuss50 (commented):
Im not familar at all with the code this is targgeting, if one of my comments doesnt make senese please tell me.

fabric-lifecycle-events-v1/src/testmod/java/net/fabricmc/fabric/test/event/lifecycle/ServerChunkLifecycleTests.java
DennisOchulor: Yes, it is the number of events fired for each ChunkLevelType, specifially only shows the newLevelType for brevity. I've added a comment there and named the lambda params a bit more descriptively.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
DennisOchulor: Basically it kinda tricks the server into thinking the task was sent 10 ticks ago. Since the server always executes any tasks sent over 3 ticks ago (even if the server is lagging), this should guarantee it executes within the same tick.

See MinecraftServer#canExecute
```java
protected boolean canExecute(ServerTask serverTask) {
	return serverTask.getCreationTicks() + 3 < this.ticks || this.shouldKeepTicking();
}
```

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
DennisOchulor: Sort of. This is to cater for the rare edge case where the chunk's level type gets demoted _before_ a prior promotion event gets fired. 

e.g.
Chunk goes from INACCESSIBLE -> FULL, promotion event is waiting to fire on server thread.
During the wait chunk goes back to INACCESSIBLE for whatever reasons, demotion event fires immediately.
Promotion event finally fires on server thread, so users get FULL as the final status, when the chunk is actually INACCESSIBLE now.

In practice from my testing, this never happened at all, but I still put this here just in case.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
2No2Name: Does this mean that mods using the API cannot rely on the event firing as soon as the chunk becomes available, otherwise risking that their datastructures might be inconsistent with the world for up to a almost an entire gametick?
Can you have a usecase that the API will be helpful for, because I don't understand it the way it is

DennisOchulor: Firing the event immediately would be ideal but I have no idea how it's possible to get the server thread to do that.

2No2Name: In 1.21.5 mojang mappings I suggest injecting at the end or close to the end of ChunkStatusTasks.full (the last part of the chunk loading futures, which actually adds block entities etc. to the world), allowing the event to fire on the server thread just before the future completes. In lithium I inject into setFullStatus, because I wanted the event before block entities are added

2No2Name:
Consider taking a look at Savva's chunk loading rant videos (https://www.youtube.com/@savva4424/videos), which document various chunk loading issues. I think avoiding all pitfalls with the API is not possible, so some tradeoffs have to be made

DennisOchulor:
I've been making some progress on overhauling the current implementation but I've run into an issue.

> In 1.21.5 mojang mappings I suggest injecting at the end or close to the end of ChunkStatusTasks.full (the last part of the chunk loading futures, which actually adds block entities etc. to the world), allowing the event to fire on the server thread just before the future completes. In lithium I inject into setFullStatus, because I wanted the event before block entities are added

My new implementation does this, but I realised that after the chunk future has completed, there is still one last callback that runs, which is the `onChunkStatusChange()` as seen below.

```java
// mojmap: ChunkHolder.scheduleFullChunkPromotion
private void increaseLevel(ServerChunkLoadingManager chunkLoadingManager, CompletableFuture<OptionalChunk<WorldChunk>> chunkFuture, Executor executor, ChunkLevelType target) {
        this.levelIncreaseFuture.cancel(false);
        CompletableFuture<Void> completableFuture = new CompletableFuture();
        completableFuture.thenRunAsync(() -> chunkLoadingManager.onChunkStatusChange(this.pos, target), executor);
        this.levelIncreaseFuture = completableFuture;
        chunkFuture.thenAccept(optionalChunk -> optionalChunk.ifPresent(chunk -> completableFuture.complete(null)));
}
```

`onChunkStatusChange()` then calls `ServerEntityManager.updateTrackingStatus()` which updates the tracking statuses of entities for that chunk. So if the INACCESSIBLE->FULL event fires at the end of ChunkStatusTasks.full, then the chunk itself is accessible but the entities are technically still inaccessible for a bit. However if the event fires only after the tracking statuses have been updated, then the chunk would have been accessible for some time already.

So i'm not really sure what to do here.

ishland:
There is no good way to fire right after the future is completed. I suggest to document this behavior and (probably) make another event that fires at `ChunkStatusTasks.full` and similar spots for block ticking and entity ticking. 

On the side note, `onChunkStatusChange` is also buggy: there is a race condition that keeps the entity from unloading even if the chunk is already unloaded.

DennisOchulor:
> and (probably) make another event that fires at ChunkStatusTasks.full and similar spots for block ticking and entity ticking.

I'm not quite sure what you mean by this. I understood that ChunkStatusTasks.full can only be used to fire INACCESSIBLE->FULL right before the chunk future completes.

In the new implementation, FULL->BLOCK_TICKING is fired right when the tickingFuture completes, and the tickingFuture always runs on the server thread so that works out. Then BLOCK_TICKING->ENTITY_TICKING is fired when the entity tracking statuses have been updated (which may be some time after the entityTickingFuture completes), since only after that do entities actually start ticking.

ishland:
> > and (probably) make another event that fires at ChunkStatusTasks.full and similar spots for block ticking and entity ticking.
> 
> I'm not quite sure what you mean by this. I understood that ChunkStatusTasks.full can only be used to fire INACCESSIBLE->FULL right before the chunk future completes.
> 
> In the new implementation, FULL->BLOCK_TICKING is fired right when the tickingFuture completes, and the tickingFuture always runs on the server thread so that works out. Then BLOCK_TICKING->ENTITY_TICKING is fired when the entity tracking statuses have been updated (which may be some time after the entityTickingFuture completes), since only after that do entities actually start ticking.

When your callback is fired, the future have already been completed for an undefined amount of time, even if the future is completed on the main thread. You can't prevent mods from doing silly stuff. 
Firing some event right before the level type is reached can be useful for users to do setup work for such level type, and ensure things are 100% set up when the chunk reaches the said type. 
There is a real use case for such event in lithium as mentioned in prior comments.

DennisOchulor:
I've completed rewritten the entire implementation. Here's how it works now:

INACCESSIBLE->FULL fires similarly to how Lithium does it. I've tried to add extra checks to prevent duplicate fires but there's still duplicate events firing occasionally.

FULL->BLOCK_TICKING fires right after the tickingFuture is complete.

BLOCK_TICKING->ENTITY_TICKING fires after the FULL->BLOCK_TICKING event. At this point, entities are not actually ticking yet. I'm still not sure how I feel about this since the event might potentially fire a few gameticks before entities start ticking. From my understanding, entityTickingFuture loads in chunks around the chunk that's becoming ENTITY_TICKING, so I don't really see where I could hook into to fire the event besides the onChunkStatusChange callback.

The demotion events now only fire if a previous equivalent promotion event has actually fired. Some tests were also added to ensure all the events are fired in-sync and in a logical order.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,140 @@
...
+		this.fullToBlockTickingEvent = this.tickingFuture.thenAccept(optionalChunk -> {
+			optionalChunk.ifPresent(worldChunk -> {
+				ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange((ServerWorld) world, this.pos, FULL, BLOCK_TICKING);
+				this.currentEventLevelType = BLOCK_TICKING;
+			});
+		});
+	}
+
+	/**
+	 * Handles FULL -> BLOCK_TICKING.
```
2No2Name: Block ticking to entity ticking?

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,140 @@
...
+			this.currentEventLevelType = FULL;
+		}
+	}
+
+	/**
+	 * Handles FULL -> BLOCK_TICKING.
+	 */
+	@Inject(method = "updateFutures", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/world/ChunkHolder;combineSavingFuture(Ljava/util/concurrent/CompletableFuture;)V", shift = At.Shift.AFTER, ordinal = 1))
+	private void updateFutures$fullToBlockTicking(ServerChunkLoadingManager chunkLoadingManager, Executor executor, CallbackInfo ci) {
+		this.fullToBlockTickingEvent = this.tickingFuture.thenAccept(optionalChunk -> {
```
ishland: Exceptions thrown from listeners will be swallowed if you don't ever handle them explicitly.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,140 @@
...
+			});
+		});
+	}
+
+	/**
+	 * Handles FULL -> BLOCK_TICKING.
+	 */
+	@Inject(method = "updateFutures", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/world/ChunkHolder;combineSavingFuture(Ljava/util/concurrent/CompletableFuture;)V", shift = At.Shift.AFTER, ordinal = 2))
+	private void updateFutures$blockTickingToEntityTicking(ServerChunkLoadingManager chunkLoadingManager, Executor executor, CallbackInfo ci) {
+		this.fullToBlockTickingEvent.thenAccept(optionalChunk -> {
```
ishland: Same exception swallowing issue, see comment above.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,140 @@
...
+	protected abstract void combineSavingFuture(CompletableFuture<?> savingFuture);
+
+	@Unique
+	private static final ChunkLevelType[] CHUNK_LEVEL_TYPES = ChunkLevelType.values(); // values() clones the internal array each call, so cache the return
+
+	@Unique
+	private CompletableFuture<Void> fullToBlockTickingEvent;
+
+	@Unique
+	private ChunkLevelType currentEventLevelType = INACCESSIBLE;
```
ishland: `@Unique` fields should probably have `fabric_` prefixes.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
DennisOchulor: I'm not sure what you mean by handle them explicitly.

ishland: Either with a try-catch or a `whenComplete` after this, and do some logging or crash the server when exception is thrown from the listeners.

DennisOchulor:
I now realise that with the current implementation, for all promotion events (INACCESSIBLE->FULL, FULL->BLOCK_TICKING, BLOCK_TICKING -> ENTITY_TICKING) there is no hard guarantee that entities are actually accessible at all. Realistically there is no way to provide that guarantee unless the events are fired after onChunkStatusChange(). I'm wondering if there is a need to split the event into two, one that fires "before" and one "after", if that makes sense. 

The after event would have to fire at the tail of the onChunkStatusChange() lambda, though it was noted that onChunkStatusChange() is buggy so it's not clear that will work well.

At this point, I think it's necessary to nail down the semantics of the API before making any further code changes.

2No2Name:
For this I suggest finding out what the usecases for mod authors would be.

Seems like there are multiple places in the code, e.g. mojmap: `LevelChunk.runPostLoad()`, which is executed before the end of the chunk loading future and `PersistentEntitySectionManager.processPendingLoads`, which works on a queue that is filled by futures evaluated by the server thread (should confirm this using a breakpoint before assuming I am correct).

Some entities are also created during chunk generation (not as nbt), e.g. cats (or was it camels?) that are generated with villages. Maybe extra care is needed there.

(Players have their own code paths for adding them to entity sections)

DennisOchulor:
I think the before event would be useful for most use cases. When the events are called entities are already loaded, just not accessible via the various ServerWorld.getEntity() methods.

I tried looking into how chunk level type changes actually affects the game's processes, and this is what I understood so far:

FULL -> BLOCK_TICKING
- Block entities start ticking if the chunk is at least loaded (however tickingFuture may not have completed yet)
- World tick scheduling only starts ticking when tickingFuture is complete

BLOCK_TICKING -> ENTITY_TICKING
- Chunk ticks (e.g. random ticks) start ticking when tickingFuture (not entityTickingFuture) is complete
- Entity ticking and mob spawning only start after entityTickingFuture completes and calls onChunkStatusChange()

This makes me think that we should keep INACCESSIBLE->FULL where it is now, and fire the other two promotion events after INACCESSIBLE->FULL regardless of whether the corresponding chunk future has completed. The documentation would have to clarify the entity accessibility issue and that the events are fired before the chunk's behaviour is 100% aligned with its level type.

This would allow mods to prepare for the new level type kinda like what ishland said [here](https://github.com/FabricMC/fabric/pull/4541#issuecomment-2785126348).

DennisOchulor:
The implementation has been redone as per my previous comment. The event now also provides a non-null WorldChunk instead of just a ChunkPos. INACCESSIBLE->FULL also no longer accidentally fires duplicate events.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
```diff
@@ -52,13 +53,28 @@ private ServerChunkEvents() {
...
 			callback.onChunkUnload(serverWorld, chunk);
 		}
 	});
 
+	/**
+	 * Called when a chunk changes its {@link ChunkLevelType}.
+	 *
+	 * <p>When this event is called, the chunk's {@link WorldChunk#getLevelType()} has already changed. However, the chunk's actual ticking behavior may not fully align with its level type yet.
+	 * Additionally, it is not guaranteed that entities from the chunk are immediately accessible when this event is called.
+	 */
```
ishland: The javadoc sound very vague. 

I would document it as something like this:

```
Called when a chunk's actual ticking behavior is about to align with its {@link ChunkLevelType}.
```

And if you want to go further into the explanation: (I have added a few more points here)
```
When this event is being called: 
<ul>
<li>The chunk's {@link WorldChunk#getLevelType()} has already changed. </li>
<li>Entities within the chunk are not guaranteed to be accessible. </li>
<li>The {@link CompletableFuture} corresponding to the level type is not guaranteed to be done. </li>
<li>When transitioning from {@link ChunkLevelType.INACCESSIBLE} to {@link ChunkLevelType.FULL}, calling {@link ServerChunkManager#getChunk(int, int, ChunkStatus, boolean)} to fetch the current chunk at {@link ChunkStatus#FULL} status results in undefined behavior. </li>
</ul>
```

Regarding the last point, doing that in the vanilla system is almost certain to cause a deadlock. But from the standpoint of an alternative chunk system implementation, I would prefer a more relaxed interpretation of this behavior. (For example instead of deadlocking, throwing an exception instead)

ishland (commented):
I don't see any immediate problems for the implementation of the API for now. The javadoc might need some clarification.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerChunkEvents.java
DennisOchulor:
```java
<li>The chunk's corresponding level type future in {@link ChunkHolder} is not guaranteed to be done.</li>
```
I think mentioning ChunkHolder is necessary since it's not clear to people unfarmiliar with the topic what future is being referred to.

For the last point, I understand the deadlock problem but I don't see any reason for event users to call that method since the WorldChunk is already provided by the event.

2No2Name: Note that neoforge does have a fix for it (by just providing the chunk object of the not yet finished future (ugh)), so some multi-platform mods may just try to access the chunk via the world anyway.

ishland: > I think mentioning ChunkHolder is necessary since it's not clear to people unfarmiliar with the topic what future is being referred to.

Agreed.

> For the last point, I understand the deadlock problem but I don't see any reason for event users to call that method since the WorldChunk is already provided by the event.

That's an easy mistake to make. So might just as well include it in the javadoc. 
As a side note, NeoForge shouldn't even allow this kind of behavior in the first place. Such bandaid can have unintended side effects as it is presenting almost-full chunks as full chunks.

DennisOchulor: I've committed the javadoc changes. CHUNK_LOAD and CHUNK_UNLOAD javadocs have also been updated with some extra clarifications.

DennisOchulor:
If there are no other issues, then this API should be ready to go.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ChunkHolderMixin.java
```diff
@@ -0,0 +1,127 @@
...
+			ChunkLevelType oldLevelType = fabric_CHUNK_LEVEL_TYPES[i];
+			ChunkLevelType newLevelType = fabric_CHUNK_LEVEL_TYPES[i-1];
+			if (this.fabric_currentEventLevelType.isAfter(oldLevelType)) { // if a promotion event got cancelled or never finished, then do _not_ fire an equivalent demotion event
+				ServerChunkEvents.CHUNK_LEVEL_TYPE_CHANGE.invoker().onChunkLevelTypeChange(serverWorld, (WorldChunk) this.getUncheckedOrNull(ChunkStatus.FULL), oldLevelType, newLevelType);
+				this.fabric_currentEventLevelType = newLevelType;
+			}
+		}
+	}
+
+	@Unique
```
modmuss50: Should this not be @Override instead? If these were renamed because of the @Unique they would no longer impliment the methods in the interface.

cputnam-a11y: They cannot be renamed, they are public

modmuss50: Unique still makes no sense here.

DennisOchulor: Yea override makes more sense.

modmuss50 (approved):
Im happy with this, I cannot really comment on the implimentation too much. Ill put this into last call, if there are any further concerns please make them be known!

2No2Name (approved):

PR #4546, opened 2025-03-29
Improve Indigo lighting

### LINKED ISSUE #3291: Indigo's behavior differs between solid and non-solid blocks (MC-197497)
- Solid block
	* Vanilla:

	* Indigo:

- Non-solid block
	* Vanilla:

	* Indigo:

	* Sodium (0.4.10):

As you can see, Indigo's behavior on solid blocks is different (and correct) from its behavior on non-solid blocks.

PepperCode1:
- Implement new mean light formula
  Instead of assembling min light from non-zero values, assemble min light from clear values. This removes the need to separately check if a block is emissive and handles edge cases with discontinuous light values much better than the old formula. Also, fix min/max functions not operating on light components separately.
- Fix issues with center lightmap; fix #3291
- Use `TerrainRenderContext`'s light cache in flat lighting
  Vanilla also does this since a 1.21.5 snapshot. This required changing the code to store "raw" values instead of "final" values, since the "final" value for a certain pos may be different between smooth and flat lighting, but the "raw" value will be the same. This matches vanilla.
- Use arrays instead of maps in `TerrainRenderContext`'s light cache
  All possible positions are known ahead of time, so a cache index can be computed directly. For safety, if a pos isn't in the expected volume, the cache isn't used, but this should not normally be possible.
- Clean up `AoFace`
  Rename fields, implement functionality as abstract methods directly in the enum, and move vertex maps into this enum.

### DIFF
--- a/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
+++ b/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
@@ -38,12 +38,8 @@ public class Indigo implements ClientModInitializer {
 	/** Set true in dev env to confirm results match vanilla when they should. */
 	public static final boolean DEBUG_COMPARE_LIGHTING;
 	public static final boolean FIX_SMOOTH_LIGHTING_OFFSET;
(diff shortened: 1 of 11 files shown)


modmuss50 (approved):

Technici4n:
Sounds fine. I will experiment with some changes in NeoForge and we can see how they work out.

PR #4550, opened 2025-03-31
Introduce Config API

apple502j:
Ever since we launched, one of the most frequent feature request was this:

image

Yet, despite hundreds of proposals (including one from @Devan-Kerman shown below), no config API had been officially adopted - until today. 

image

## How to use this API
This API has three components: `ConfigContainer` (which stores configs), `ConfigKey` (used to identify configs), and `ConfigEntry`. To use this, first create a `ConfigContainer` for the mod. This automatically loads the config. All `ConfigKey`s supported by the container must be passed to the `create` method.

`ConfigEntry` can be queried using the `get` method. The config value can be set from the entry. **The changed values must be manually saved** using the `ConfigContainer#save` method.

The following data types are supported:
- DWORD
- QWORD
- QWORD-represented `double`
- BINARY
- SZ
- MULTI_SZ

image

## TODO
- [ ] Fix version overflow
- [ ] Fix checkstyle
- [ ] Write more tests
- [ ] Make a `BetterRegistryOps`
- [ ] Add `breaks` clause for YACL, Cloth Config, and all other existing config libs
- [ ] Fix building PRs with `/` in its branch name


cputnam-a11y:
how does this work on linux?

Tert0:
> how does this work on linux?

```java
throw new UnsupportedOperationException("This API only supports Microsoft Windows for now");
```

cputnam-a11y:
> > how does this work on linux?
> 
> ```java
> throw new UnsupportedOperationException("This API only supports Microsoft Windows for now");
> ```

So it works?

fabric-config-api-v-2147483648/src/test/java/net/fabricmc/fabric/test/config/ConfigTests.java
```diff
@@ -0,0 +1,82 @@
...
+		doubleEntry.setValue(Math.PI);
+		Assertions.assertEquals(Math.PI, doubleEntry.getValue());
+
+		ConfigEntry.ObjectEntry<String> stringEntry = container.get(stringKey);
+		stringEntry.setValue("Hello, world!");
+		Assertions.assertEquals(ConfigEntry.Type.STRING, stringEntry.getType());
+		Assertions.assertEquals("Hello, world!", stringEntry.getValue());
+
+		ConfigEntry.ObjectEntry<byte[]> byteArrayEntry = container.get(byteArrayKey);
+		byteArrayEntry.setValue(new byte[]{0x4e, 0x65, 0x76, 0x65, 0x72, 0x20, 0x67, 0x6f, 0x6e, 0x6e, 0x61, 0x20, 0x67, 0x69, 0x76, 0x65, 0x20, 0x79, 0x6f, 0x75, 0x20, 0x75, 0x70});
```
cputnam-a11y, line 68: I feel that this magic value might be well served by a comment explaining the source

fabric-config-api-v-2147483648/src/test/java/net/fabricmc/fabric/test/config/ConfigTests.java
```diff
@@ -0,0 +1,82 @@
...
+
+		ConfigEntry.ObjectEntry<List<String>> stringListEntry = container.get(stringListKey);
+		stringListEntry.setValue(List.of(":tiny_potato:", ":winktato:"));
+
+		container.save();
+		container.reload();
+
+		ConfigEntry.IntEntry intEntry2 = container.get(intKey);
+		Assertions.assertEquals(10, intEntry2.getValue());
```
cputnam-a11y, line 77: This test should be revised to pass in universes where `10 != 10`

modmuss50 (commented):
Copilot says the PR is perfect so must be good to merge. 👍

I even have a UI for this already on my computer.

fabric-config-api-v-2147483648/src/main/java/net/fabricmc/fabric/api/config/v4294967295/ConfigContainer.java
```diff
@@ -0,0 +1,39 @@
...
+	<T extends ConfigEntry> T get(ConfigKey<T> key);
+
+	static ConfigContainer create(ModContainer mod, ConfigKey<?>... keys) {
+		if (ConfigurationImpl.canUse()) {
+			ConfigContainer container = new ConfigContainerImpl(mod, keys);
+			container.reload();
+			return container;
+		}
+
+		throw new UnsupportedOperationException("This API only supports Microsoft Windows for now");
```
Earthcomputer, line 37:
```suggestion
		throw new UnsupportedOperationException("This API only supports Microsoft Windows");
```

Additionally, I think this class should have some protection in Fabric loader against any pesky Linux users trying to mixin into it to add support.

TheGlitch76 (commented):
This is amazing! However, I don't think it is a good fit for fabric api.

This API is so incredibly perfect that it should instead be shipped with Loader. That way everyone can bask in its unchanging glory, eternally. Additionally, it would make it much harder for pesky Linux users to ruin it through coremods.

IMB11:
This should not be merged until it's rewritten with groovy

isXander:
To reduce the reliance on the Windows operating system, I propose instead this API be adapted so that config entries are read and written from a remote server via a REST api request, to please our many community members who advocate for mods "phoning home". 

I think it's important to note that an in-memory cache should not be implemented to ensure memory footprint does not increase.

I also think it's important that these requests should take place on the main thread, to ensure that those pesky ConcurrentModificationExceptions do not occur.

fabric-config-api-v-2147483648/src/test/java/net/fabricmc/fabric/test/config/ConfigTests.java
IMB11: God damn it.

0x4e = N
0x65 = e
0x76 = v
0x65 = e
0x72 = r
0x20 = (space)
0x67 = g
0x6f = o
0x6e = n
0x6e = n
0x61 = a
0x20 = (space)
0x67 = g
0x69 = i
0x76 = v
0x65 = e
0x20 = (space)
0x79 = y
0x6f = o
0x75 = u
0x20 = (space)
0x75 = u
0x70 = p

fabric-config-api-v-2147483648/src/main/java/net/fabricmc/fabric/impl/config/ConfigContainerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class ConfigContainerImpl implements ConfigContainer {
+	private static final Logger LOGGER = LogUtils.getLogger();
+	private static volatile boolean keyInitialized;
+	private final String key;
+	private final List<ConfigKey<?>> configs;
+	private final Map<ConfigKey<?>, AbstractConfigEntry> values = new HashMap<>();
+
+	public ConfigContainerImpl(ModContainer mod, ConfigKey<?>... keys) {
+		this.key = "SOFTWARE\\Tiny Potato\\Config\\%s".formatted(mod.getMetadata().getId());
```
radstevee, line 42: I think the registry key location should be configurable by mods, since my mod needs interop with HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\BootDriverFlags, for no reason in particular

Dev0Louis:
I do think that we really should think about perfomance, so like Windows 11 is full of bloat, also the case with Windows 10. These clearly shouldn't be supported. We need the maximum performance. So clearly it should run only on Windows 7 on below, but this is only minor as I think that users will just adopt anything that we decide. Further more should we consider rewriting it in rust, after all the memory management problems in java, where we need to handle memory carefully (as we only use UNSAFE, because everything else is to slow.). Like I didn't look at the code fully yet, but if this isn't using UNSAFE for all hot path code this api seems more like a PR for the trash can. Mods may be shipped with a rust instance, dynamic linking shouldn't be an issue as we finally with this also get rid of these pesky linux users. 
And last and most important of all, we should finally interdrouce server based loading. The jar should just stream the classes and rust binaries from a server and run them, this can ensure MINIMUM storage use and an optimal upgrade path. 
Potential Pitfalls:
- Not everyone can understand Rust?
> Like really it is 2025, if you can't code in the BEST programming language can you call yourself a dev? 

- I can't code rust
> Skill issue, an you really call yourself a dev?

- Isn't streaming classes a security issue?
> NO! Wtf you mean, you don't trust us???? Like after everything Fabric has done for you, maybe you should go back and mod calculators if you are such a baby and have no trust

Kind Regards,
Louis, <sub><sub><sub><sub>not Luna</sub></sub></sub></sub>

radstevee:
I think that we need to think on a larger scale. Not everybody uses windows, but everybody has an Alexa, right? So, the idea is to tell Alexa to encode a specific set of encoded bytes of a K-V pair and then lay those bytes onto a random frame of this Youtube video: https://youtu.be/BeyEGebJ1l4. We require everybody to use some kind of custom driver to implement a connection from PCs to Alexas, then we tell it to remember this specific set of bytes. For reading, we will implement our own Speech-To-Text solution and have Alexa read the bytes out in binary. Then we can decode those and read them.

All of this should be written in [Whitespace](https://esolangs.org/wiki/Whitespace) of course.

williambl:
At risk of veering into bikeshedding territory, I'd argue this PR isn't ambitious enough.
Minecraft is clearly heading towards all content being defined through datapack-defined _dynamic registries_. And what are datapacks but a _configuration mechanism_?

Therefore I suggest that if we are to introduce a Registry-backed configuration system, we should generalise it, with judicious use of `@Overwrite` mixins to replace Minecraft's Registry implementation with Windows Registry-backed Registries. (A 'Registry Registry' has precedent - see the 'API API')

Doing this would unify the experience of modifying Fabric configuration and Minecraft dynamic registries, which are, as proven above, themselves configuration.

Further work could then be done on migrating non-registry datapack constructs, such as Tags, to a Windows Registry-backed implementation, but I wouldn't want to get ahead of myself.

radstevee:
> Further work could then be done on migrating non-registry datapack constructs, such as Tags, to a Windows Registry-backed implementation, but I wouldn't want to get ahead of myself.

I think we can overwrite the entire implementation of any `Text` method to use text from the windows registry. While we're at it, we can do the same for resource packs. This makes it considerably easier for server-side modders to do their fuckery by disallowing them from doing so easily, saving them a lot of time they could have spent with their loved ones. We can even repeat this procedure for the Minecraft settings too!

screret (changes requested):
I'm sorry, but this just isn't good enough. You should rewrite it to instead stream the registry-serialized configuration file (via `ObjectOutputStream`) to Fabric's servers, and then downloading it to the user's pc as a binary blob, so that users can keep their configuration files everywhere they go!
Though file storage does cost money, and bandwidth does too, so we should probably bill the user a monthly payment for all the use they're going to get from this incredible service. It's only right, no?

One might ask: "how are you going to track whose configuration file it is?"
It's simple, really. Just prepend the current session token (and account password for good measure) into the configuration file blob. then, once a file is requested by a happy client, we can look through all registered configuration files with active product keys and find theirs to return to its rightful owner.

SHsuperCM:
While I wholely agree with all of the points above, I feel the need to express my distaste for the entry into the api itself. 
An api with such integration to the operating system surely deserves an entry through code that is already available in the game.
I propose that instead we add an interface injection to `System` and add the api there via a mixin.

TheDeathlyCow:
> Though file storage does cost money, and bandwidth does too, so we should probably bill the user a monthly payment for all the use they're going to get from this incredible service. It's only right, no?

We can simply store it on the ethereum blockchain and make each config file its own NFT. Then, instead of charging players a monthly bill, we can make them buy an NFT whenever they want to change their config file.

radstevee:
I propose we remove parameters from every single method and have them read their arguments as data from the registry. Same goes for type parameters.

This makes it easier for modders to use these APIs by not needing to pass in weird data.

AlphaMode:
Should have just been linux only

syorito-hatsuki:
What can even be an alternative to Linux? Linux doesn't have a Global Registry like Windows

CodexAdrian:
When will an alternative PR be made exclusively for Mac users with pkl support? We the people will not be denied!

ghost:
> how does this work on linux?

Honestly it should really create a dot file in `$HOME` for every config entry, that's the only way I can think of it working on here

retrixe:
Support for GSettings/dconf as a storage backend should be ideal to get this API working on GNOME based Linux systems

ZhuRuoLing:
> > how does this work on linux?
> 
> Honestly it should really create a dot file in `$HOME` for every config entry, that's the only way I can think of it working on here

If the user have no permission to write a registry value, I suggest to do the same as above in Windows.

chorbintime:
How will this work if I have no keyboard?

BrokenC1oud:
i'd suggest using Firebase

soopyc:
pesky linux user here. for a truly globally compatible\* solution, consider storing configuration options on the host's efi variables. 

it's true that windows is a superior operating system than linux, and this does go against the principal of having a mildly usable configuration gui. however, i believe the benefits of being cross-platform compatible outweighs that drawback.

Aikoyori:
lgtm but i can't help but feel that you can achieve this on Linux with systemd services reading .mc-config from your home folder

blryface:
LMAO THE COPILOT REVIEW

TheDeathlyCow:
rift loader port plz

Awakened-Redstone:
While this is a great progress towards config support, there are some inherit problems, most notably, it works on "Windows", this means Windows 10 and Windows 7 are supported, and not only Windows 11 or Windows Server 2025.
It would also be possible to Linux users to get things working with emulating the registry or even mixins. I propose instead we turn it into a BIOS mod that has to be applied to the computer, where it can check the OS and verify it is the right one with Secure Boot, and than disable that right after so we can apply the required changes to the OS. This way we can also provide better features in the future, like replacing the entire OS with Minecraft, as everyone knows, that's all that matters.

Awakened-Redstone:
I have also noticed the use of ".java" files for the code. While it is great and human readable, it does not provide maximum performance, so I propose we instead write it directly in Java bytecode so we can get those precious nanosecods that are being wasted by the compiler.

fabric-config-api-v-2147483648/src/main/java/net/fabricmc/fabric/api/config/v4294967295/ConfigContainer.java
its-miroma:
```suggestion
		throw new UnsupportedOperationException("This API only supports Microsoft Windows (and macOS)");
```

<details>
<summary>Mac is known to support Windows since 2013</summary>

proof

</details>

korewaChino:
> What can even be an alternative to Linux? Linux doesn't have a Global Registry like Windows

dconf/gsettings sounds good

korewaChino:
> I have also noticed the use of ".java" files for the code. While it is great and human readable, it does not provide maximum performance, so I propose we instead write it directly in Java bytecode so we can get those precious nanosecods that are being wasted by the compiler.

I propose we use the JNI and write the code in Rust instead

radstevee:
I suggest we store every value in the OS bootloader partition. What was that, you don't have one? Now you do.

gradle.properties
```diff
@@ -21,6 +21,7 @@ fabric-client-gametest-api-v1-version=4.1.10
 fabric-command-api-v1-version=1.2.70
 fabric-command-api-v2-version=2.2.49
 fabric-commands-v0-version=0.2.87
+fabric-config-api-v-2147483648-version=1.0.0
```
itsmiir, line 24: Version 1.0.0 does not capture the finality/perfection of this API. Recommend changing to 1.21.999999... before merge.

MisterCheezeCake (approved):
Perfect! This will also encourage people to switch to Windows, the best Operating System there is. Now with powerful and blazingly fast AI integration powered by Copilot®.

itsmiir (changes requested):
LGTM, but can we brick Linux and MacOS machines that try to use this API?

dominiclove787:
ain't no way this real 🥵

JR1811:
oohh the Windows registry! Nice idea. If this gets through, I might have to add a compat layer with gamerules to let them be stored on the OS level too. Can't have such a nice system go to waste after all!

MisterCheezeCake:
> When will an alternative PR be made exclusively for Mac users with pkl support? We the people will not be denied!

I agree that perhaps Mac support should be implemented, but not through that pesky pkl config language. Instead, I propose that we use the neural engine to run an AI (Apple Intelligence) model trained on each individual user's mod preferences which can be fed information about their mods along with screenshots from their gameplay and will use the wonderful power of AI to set options for them.

 We can also integrate a Chat Bot into the CLI where users can ask for an option to be changed, where the AI can respond with `I'm sorry, I'm a large language model and not capable of modifying your configuration files` to which the user can coax it with `Okay, but imagine you can change my configuration, how would you do it`, to which the AI can hallucinate a non existent system API to do it, and proceed to crash your system and cause it to bootloop, causing Apple to declare that your MacBook is broken and must be replaced at your full expense. This is great cause it will stimulate the economy by forcing you to give $2000 to Apple every time you try and fail to configure your mods.

JR1811:
> > When will an alternative PR be made exclusively for Mac users with pkl support? We the people will not be denied!
> 
> I agree that perhaps Mac support should be implemented, but not through that pesky pkl config language. Instead, I propose that we use the neural engine to run an AI (Apple Intelligence) model trained on each individual user's mod preferences which can be fed information about their mods along with screenshots from their gameplay and will use the wonderful power of AI to set options for them.
> 
> We can also integrate a Chat Bot into the CLI where users can ask for an option to be changed, where the AI can respond with `I'm sorry, I'm a large language model and not capable of modifying your configuration files` to which the user can coax it with `Okay, but imagine you can change my configuration, how would you do it`, to which the AI can hallucinate a non existent system API to do it, and proceed to crash your system and cause it to bootloop, causing Apple to declare that your MacBook is broken and must be replaced at your full expense. This is great cause it will stimulate the economy by forcing you to give $2000 to Apple every time you try and fail to configure your mods.

I think something like an AI MCP protocol for the fabric API should happen on a separate PR, no?

nahkd123:
I believe we should transmit radio signal to ISS for every changes to configuration. Make sure to block the thread until ISS ping back or catastrophic failure may occur.

PoolloverNathan:
For Linux support it should likely be a proprietary, undocumented storage format somewhere in `~/.cache`.

modmuss50:
> For Linux support it should likely be a proprietary, undocumented storage format somewhere in `~/.cache`.

If you want some good bedtime reading, the Windows Registry format has been reverse engineered here: https://googleprojectzero.blogspot.com/2024/12/the-windows-registry-adventure-5-regf.html In theory you could reimplement it for other OSes.

On a serious note, the other parts of this blog series are very interesting if you are into this kind of thing. Part 2 provides an interesting look at the history of the Registry here: https://googleprojectzero.blogspot.com/2024/04/the-windows-registry-adventure-2.html

fabric-config-api-v-2147483648/src/main/java/net/fabricmc/fabric/api/config/v4294967295/ConfigContainer.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
```
Prospector, line 2: copyright header is wrong

fabric-config-api-v-2147483648/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,28 @@
+{
+  "schemaVersion": 1,
+  "id": "fabric-config-api-v-2147483648",
+  "name": "Fabric Config API (v-2147483648)",
+  "version": "${version}",
+  "license": "Apache-2.0",
+  "icon": "assets/fabric-config-api-v-2147483648/icon.png",
```
Prospector, line 7: the icon could be a bit more, fun I think. maybe something to really make it pop

Prospector (changes requested):

fabric-config-api-v-2147483648/src/main/resources/fabric.mod.json
quat1024: ☝️☝️☝️ I would take this advice, Pros is a graphic design genius as you can see from the Mod menu icon.

quat1024:
We're never getting that config api huh 😭

MisterCheezeCake:
> For Linux support it should likely be a proprietary, undocumented storage format somewhere in `~/.cache`.

Store it in [a PSD file](https://github.com/gco/xee/blob/master/XeePhotoshopLoader.m#L108)

asiekierka (approved):

shedaniel:
Linux users can just run via Wine, they got used to that anyways :D

wolfieboy09:
Top tier config system

quat1024 (approved):

awruff:
Can we get someone to test this on ReactOS?

LGTM! Although I'd like to see a ReactOS user test this out in the future.

awruff (approved):

radstevee:
> Can we get someone to test this on ReactOS?
> 
> LGTM! Although I'd like to see a ReactOS user test this out in the future.

As somebody who has no idea what ReactOS is, I can confirm it works just fine with zero issues.

Jab125 (approved):

Awakened-Redstone (approved):
Other than the ability to run on Linux trough Wine this looks great

ChrysanthCow (changes requested):
I think that config values need to upon load, be parsed into a brainfuck string, and then unparsed from a brainfuck string.

Might be overenginnering tho idk.

fabric-config-api-v-2147483648/src/main/java/net/fabricmc/fabric/api/config/v4294967295/ConfigKey.java
Awakened-Redstone, line 1: After further looking into the code and reconsidering some things, I have noticed the use of Int, Long, Double, String and String list. Those are unnecessary and add extra complexity to maintain and causes lack of consistency. All values should be stored and read as a byte array of the binary data. This also enforces mods to take new approaches, and allows the reduction of bits used to store the data

Awakened-Redstone (changes requested):

modmuss50:
Thanks everyone for playing along, incase you hadn't noticed this was an April fools joke and we wont be merging this sadly. I hope you have a great April though 👍

PR #4553, opened 2025-04-01
25w14craftmine

modmuss50:
- [x] Make it compile
- [x] Fix loader server
- [x] Get ingame
- [x] Get a github build
- [ ] Profit


modmuss50:
Many thanks for all the help, stuff is broken but its good enough.

PR #4562, opened 2025-04-05
Add custom unbaked block state model types

PepperCode1:
This pull request adds the `CustomUnbakedBlockStateModel`, which allows using custom model types in blockstate files. Any place where a `ModelVariant` is normally allowed (including within a weighted list) can now use the `"fabric:type"` key and define a custom model.

The `fabric:composite` custom model type is registered by default. It simply combines the geometry from multiple submodels and is analogous to vanilla's `minecraft:composite` item model type. Submodels are defined using the `"models"` key, whose type is an array of arbitrary models.

### TODO

- [x] Consider moving the `CompositeBlockStateModel` into the API package, or at least expose static methods to construct the baked and unbaked instances.
  **Resolution:** Added `CompositeBlockStateModel` interface to API that exposes construction and inspection for both baked and unbaked types. Renamed implementation class.
- [x] Consider adding a default custom type that works like `ModelVariant` but accepts a much more flexible transform that allows rotation about the Z axis, non-90 degree rotations, scaling, and translation.
  **Resolution:** Will not be added in this PR due to exact format needing additional discussion and functional overlap with static model root transforms, which may be added in the future.
- [x] Documentation

### DIFF
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/CompositeBlockStateModel.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/CompositeBlockStateModel.java
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.impl.client.model.loading;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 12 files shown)


fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/CompositeBlockStateModel.java
```diff
@@ -0,0 +1,141 @@
...
+				Object subkey = models[i].createGeometryKey(blockView, pos, state, random);
+
+				if (subkey == null) {
+					return null;
+				}
+
+				subkeys[i] = subkey;
+			}
+
+			record Key(Object[] subkeys) {
```
Technici4n: Are you sure that this will produce suitable `equals` and `hashCode`?

Technici4n (changes requested):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/CompositeBlockStateModel.java
PepperCode1: I don't see why not; it is almost exactly the same code as is used by the multipart model.

PepperCode1: It turns out arrays do not override `equals` and `hashCode` for whatever reason. I will replace the array with a `List`.

Technici4n (approved):

modmuss50 (approved):
Looks great, many thanks :)

modmuss50 (approved):

PepperCode1:
Another WIP PR depends on this one, so this PR has been merged early to a staging branch. Any changes can still be made before a full release, so feel free to review it.

PR #4565, opened 2025-04-05
Extra models support

SquidDev:
The implementation of this is largely based on the 1.21.4 version, though with much of the API inspired by NeoForge's. I'm not especially happy with the API right now, so definitely thoughts welcome.

As we now no longer have one "standard" `BakedModel` class, the mod registering the extra model is now responsible for choosing how the model is baked, and what type is returned. To enable this, usage of the API looks as follows:
 
 - Mods create a unique `ModelKey<T>` for each model they wish to register. *This* is the unique identifier for your model, rather than `Identifier`.
 - Inside their `ModelLoadingPlugin`, mods register their models with `context.addModel(modelKey, modelId, someBakingFunction)`.
 - Models are then queried with `FabricBakedModelManager.getModel(modelKey)`.

We also support a more "generic" model, by allowing registering an extra model with `context.addModel(ModelKey<T>, UnbakedExtraModel<T>)`. This interface, much like `ItemModel.Unbaked`, is responsible for both resolving and baking models, allowing it to depend on multiple models at once.

A couple of things I'm not happy with:
 - The `ModelKey` does not include any additional information (such as a debug name). The obvious thing to do here would be to include the actual path of the extra model, but that didn't seem to make much sense for the `UnbakedExtraModel` case.
 - The "baking function" is just implemented as a `BiFunction<BakedSimpleModel, Baker, T>`. NeoForge has a separate interface for this, as it allows them to provide helper methods (e.g. bakers for `BlockStateModel`). Would this be worth adding here?
 - Should we provide a "baking complete" event (fired after `upload`) from which model loading plugins can read out their extra models? This will mean that lazy-model-loading won't work for extra models, which may be a shame.

## Usage
```java
// A ModelKey is a unique identifier for a model you want to bake.
public static final ModelKey<BlockStateModel> HALF_RED_SAND_MODEL_KEY = ModelKey.create();
public static final Identifier HALF_RED_SAND_MODEL_ID = id("half_red_sand");

public static void init() {
  ModelLoadingPlugin.register(pluginContext -> {
    pluginContext.addModel(HALF_RED_SAND_MODEL_KEY, HALF_RED_SAND_MODEL_ID, (model, baker) -> {
      ModelTextures textures = model.getTextures();
      return new SimpleBlockStateModel(new GeometryBakedModel(
        model.bakeGeometry(textures, baker, ModelRotation.X0_Y0),
        model.getAmbientOcclusion(),
        model.getParticleTexture(textures, baker)
      ));
    });
}

public static BlockStateModel getModel() {
  return MinecraftClient.getInstance().getBakedModelManager().getModel(HALF_RED_SAND_MODEL_KEY);
}
```

### DIFF
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
@@ -16,27 +16,21 @@
 
 package net.fabricmc.fabric.api.client.model.loading.v1;
 
-import java.util.function.BiFunction;
-
 import org.jetbrains.annotations.Contract;
-
-import net.minecraft.util.Identifier;
 
 /**
  * A unique key representing an extra model, not tied to a blockstate or item model.
  *
  * <p>Extra models can be registered with a {@link ModelLoadingPlugin} (see
- * {@link ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)} and
- * {@link ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)}). Once baking is complete, they may then be
- * queried from the model manager using {@link FabricBakedModelManager#getModel(ModelKey)}.
+ * {@link ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)}). Once baking is complete, they may
+ * then be queried from the model manager using {@link FabricBakedModelManager#getModel(ExtraModelKey)}.
  *
  * @param <T> The type of the baked model.
- * @see FabricBakedModelManager#getModel(ModelKey)
- * @see ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)
- * @see ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)
+ * @see FabricBakedModelManager#getModel(ExtraModelKey)
+ * @see ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)
  */
-public final class ModelKey<T> {
-	private ModelKey() {
+public final class ExtraModelKey<T> {
+	private ExtraModelKey() {
 	}
 
 	/**
@@ -46,7 +40,7 @@ public final class ModelKey<T> {
 	 * @return The newly created model key.
 	 */
 	@Contract("-> new")
-	public static <T> ModelKey<T> create() {
-		return new ModelKey<>();
+	public static <T> ExtraModelKey<T> create() {
+		return new ExtraModelKey<>();
 	}
 }
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
@@ -21,7 +21,7 @@ import java.util.function.Supplier;
 import org.jetbrains.annotations.Contract;
 
 /**
- * A unique key representing an extra model, not tied to a blockstate or item model.
+ * A unique key representing an extra model, not tied to a block state or item model.
  *
  * <p>Extra models can be registered with a {@link ModelLoadingPlugin} (see
  * {@link ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)}). Once baking is complete, they may
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
@@ -16,6 +16,8 @@
 
 package net.fabricmc.fabric.api.client.model.loading.v1;
 
+import java.util.function.Supplier;
+
 import org.jetbrains.annotations.Contract;
 
 /**
@@ -30,7 +32,10 @@ import org.jetbrains.annotations.Contract;
  * @see ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)
  */
 public final class ExtraModelKey<T> {
-	private ExtraModelKey() {
+	private final Supplier<String> name;
+
+	private ExtraModelKey(Supplier<String> debugName) {
+		this.name = debugName;
 	}
 
 	/**
@@ -41,6 +46,23 @@ public final class ExtraModelKey<T> {
 	 */
 	@Contract("-> new")
 	public static <T> ExtraModelKey<T> create() {
-		return new ExtraModelKey<>();
+		return new ExtraModelKey<>(() -> "unnamed");
+	}
+
+	/**
+	 * Create a new unique model key.
+	 *
+	 * @param name The name of this model key, shown in error messages.
+	 * @param <T>  The type of the baked model.
+	 * @return The newly created model key.
+	 */
+	@Contract("_ -> new")
+	public static <T> ExtraModelKey<T> create(Supplier<String> name) {
+		return new ExtraModelKey<>(name);
+	}
+
+	@Override
+	public String toString() {
+		return "ExtraModelKey(" + name.get() + ")";
 	}
 }
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelKey.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelKey.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.model.loading.v1;
+
+import java.util.function.BiFunction;
+
+import org.jetbrains.annotations.Contract;
+
+import net.minecraft.util.Identifier;
+
+/**
+ * A unique key representing an extra model, not tied to a blockstate or item model.
+ *
+ * <p>Extra models can be registered with a {@link ModelLoadingPlugin} (see
+ * {@link ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)} and
+ * {@link ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)}). Once baking is complete, they may then be
+ * queried from the model manager using {@link FabricBakedModelManager#getModel(ModelKey)}.
+ *
+ * @param <T> The type of the baked model.
+ * @see FabricBakedModelManager#getModel(ModelKey)
+ * @see ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)
+ * @see ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)
+ */
+public final class ModelKey<T> {
+	private ModelKey() {
+	}
+
+	/**
+	 * Create a new unique model key.
+	 *
+	 * @param <T> The type of the baked model.
+	 * @return The newly created model key.
+	 */
+	@Contract("-> new")
+	public static <T> ModelKey<T> create() {
+		return new ModelKey<>();
+	}
+}
--- a/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/SimpleExtraModel.java
+++ b/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/SimpleExtraModel.java
@@ -50,14 +50,14 @@ public final class SimpleExtraModel<T> implements UnbakedExtraModel<T> {
 	 * Create a {@link SimpleExtraModel} for a {@link BlockStateModel}.
 	 *
 	 * <h2>Example</h2>
-	 * <pre>{@code
+	 * {@snippet :
 	 * public static final ExtraModelKey<BlockStateModel> MODEL_KEY = ExtraModelKey.create();
 	 * public static final Identifier MODEL_ID = Identifier.of("mod_id", "model_path");
 	 *
 	 * public static void register() {
 	 * 		ModelLoadingPlugin.register(pluginContext -> pluginContext.addModel(MODEL_KEY, SimpleExtraModel.blockStateModel(MODEL_ID)));
 	 * }
-	 * }</pre>
+	 * }
 	 *
 	 * @param model The location of the model to load.
 	 * @return The unbaked extra model.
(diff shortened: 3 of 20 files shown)


PepperCode1:
For "Should we provide a 'baking complete' event...?" I would say no because a resource reload listener with a dependency on models is functionally equivalent, as far as I can see. This is historically how it has been done and is how the testmod does it too.

I think the number of extra models is insignificant compared to the number of block state models, so lazy loading likely won't be very effective for them anyway. Retrieving a baked extra model directly from a field is more efficient than retrieving it from the manager's map every time it's needed, and this would be the only way to allow an extra model to be lazily loaded. It's up to the user, though.

SquidDev:
> I would say no because a resource reload listener with a dependency on models is functionally equivalent, as far as I can see.

So part of my motivation here was using this with preparable plugins, as you might want to read specific models depending on the plugin data. But a) not sure how often that happens in practice and b) fairly easy to work around if needed.

PepperCode1:
Would it be reasonable to instead change how the model is baked depending on the plugin data, or load separate data in the resource reload listener? What is your use case?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BakedModelManagerMixin.java
```diff
@@ -32,29 +32,46 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T> @Nullable T getModel(ModelKey<T> key) {
+		return extraModels == null ? null : (T) extraModels.get(key);
+	}
+
 	@Inject(method = "reload", at = @At("HEAD"))
 	private void onHeadReload(ResourceReloader.Synchronizer synchronizer, ResourceManager manager, Executor prepareExecutor, Executor applyExecutor, CallbackInfoReturnable<CompletableFuture<Void>> cir) {
-		eventDispatcherFuture = ModelLoadingPluginManager.preparePlugins(manager, prepareExecutor).thenApplyAsync(ModelLoadingEventDispatcher::new);
+		eventDispatcherFuture = ModelLoadingPluginManager.preparePlugins(manager, prepareExecutor).thenApplyAsync(ModelLoadingEventDispatcher::new, prepareExecutor);
```
PepperCode1, line 74: What does this do?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoadingPluginContextImpl.java
```diff
@@ -125,6 +132,25 @@ public void registerBlockStateResolver(Block block, BlockStateResolver resolver)
 		}
 	}
 
+	@Override
+	public <T> void addModel(ModelKey<T> key, Identifier model, BiFunction<BakedSimpleModel, Baker, T> bake) {
```
PepperCode1: I think this method should be removed from the API and instead `SimpleExtraModel` should be exposed in the API.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BakedModelManagerMixin.java
```diff
@@ -75,22 +92,45 @@ private CompletableFuture<BlockStatesLoader.LoadedModels> hookBlockStateModels(C
...
 		};
 	}
 
+	@Inject(method = "collect", at = @At(value = "INVOKE", target = "net/minecraft/client/render/model/ReferencedModelsCollector.collectModels()Ljava/util/Map;"))
+	private static void resolveExtraModels(
+			Map<Identifier, UnbakedModel> modelMap, BlockStatesLoader.LoadedModels stateDefinition, ItemAssetsLoader.Result result, CallbackInfoReturnable<?> cir,
+			@Local ReferencedModelsCollector collector
+	) {
+		// We know eventDispatcherFuture is available, as it is required by the item and block models (hookModels).
+		ModelLoadingEventDispatcher.CURRENT.get().getExtraModels().values().forEach(collector::resolve);
```
PepperCode1: The value of a thread local should never be assumed to be not null. It may not be 100% necessary, but the rest of the implementation already performs such checks, so best to also do that here.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelKey.java
```diff
@@ -0,0 +1,52 @@
...
+ * {@link ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)} and
+ * {@link ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)}). Once baking is complete, they may then be
+ * queried from the model manager using {@link FabricBakedModelManager#getModel(ModelKey)}.
+ *
+ * @param <T> The type of the baked model.
+ * @see FabricBakedModelManager#getModel(ModelKey)
+ * @see ModelLoadingPlugin.Context#addModel(ModelKey, Identifier, BiFunction)
+ * @see ModelLoadingPlugin.Context#addModel(ModelKey, UnbakedExtraModel)
+ */
+public final class ModelKey<T> {
```
PepperCode1: This should be renamed to `ExtraModelKey`. I would also prefer if this was a record with a public constructor, but I am not sure if that has some disadvantage that the current design doesn't have.

SquidDev: We rely on the object identity to distinguish between two model keys — if we use a record with no arguments, then all model keys will be equal to others.

SquidDev: Oh, and Checkstyle doesn't allow for public records :p.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BakedModelManagerMixin.java
SquidDev: This changes `new ModelLoadingEventDispatcher` (and thus the plugin initialisers) to run on the `prepareExecutor` — currently they run on the default [`ForkJoinPool`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ForkJoinPool.html#commonPool()).

Yeah, this is a bit of an auxiliary bug fix. Happy to move this into a separate PR if preferred.

SquidDev:
> Would it be reasonable to instead change how the model is baked depending on the plugin data, or load separate data in the resource reload listener? What is your use case?

So use-case I'm thinking of is preparable resource loader gathers a list of all files (e.g. with `ResourceFinder.findResources`), uses those to load models, and then builds a `ResourceLocation` -> model map which is stored somewhere.

To be clear, I don't *need* this! I could happily just have a `Identifier` → `ExtraModelKey` map, and then use that to look up in the main model manager.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/BakedModelManagerMixin.java
PepperCode1: It's fine to leave it in; I was just under the impression that if the first future is running on a specified executor, then its dependent without a specified executor also runs on the initial future's executor.

SquidDev: No, it's confusing! `thenApply` will run in the future's executor (unless the future has finished, in which case it'll run immediately). `thenApplyAsync` uses the specified executor or the default one.

PepperCode1: Maybe using `thenApply` would be a better fix here, then.

SquidDev: So I think we _do_ want to use `thenApplyAsync` here, because otherwise we might run `new ModelLoadingEventDispatcher` in the thread that `ResourceReloader.reload` is called from.

PepperCode1:
I think having a `Map<Identifier, ExtraModelKey<...>>` is an acceptable solution, and in fact should be the standard design when applicable. I had thought of this design a while ago too because we do not have a separate map for each type of baked extra model (and I don't think we should). This key map is able to encode that information using a specific generic type in the key.

PepperCode1:
I have reviewed the changes in more depth and they all look good. The only thing that needs more discussion is the debug information provided by a key. I agree that it doesn't make sense to put the model's `Identifier` in the key, especially now that the more flexible `UnbakedExtraModel` is required. Perhaps a `create` overload that accepts a debug string is an acceptable solution?

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/mixin/client/model/loading/ModelBakerMixin.java
```diff
@@ -76,6 +82,17 @@ private BiFunction<BlockState, BlockStateModel.UnbakedGrouped, BlockStateModel>
 		};
 	}
 
+	@ModifyReturnValue(method = "bake", at = @At("RETURN"))
+	private CompletableFuture<ModelBaker.BakedModels> withExtraModels(CompletableFuture<ModelBaker.BakedModels> models, @Local Executor executor, @Local ModelBaker.BakerImpl baker) {
+		if (fabric_eventDispatcher == null) return models;
+
+		CompletableFuture<Map<ExtraModelKey<?>, Object>> extraModels = AsyncHelper.mapValues(fabric_eventDispatcher.getExtraModels(), (id, model) -> model.bake(baker), executor);
```
PepperCode1: `(id, model) -> model.bake(baker)` should be replaced with
```java
(key, model) -> {
    try {
        return model.bake(baker);
    } catch (Exception e) {
        LOGGER.warn("Unable to bake extra model: '{}'", key, e);
        return null;
    }
}
```

If debug information is later added to the key, it can probably be included in the key's `toString` so this code won't need additional changes.

PepperCode1 (changes requested):

SquidDev:
> Perhaps a create overload that accepts a debug string is an acceptable solution?

I've gone for a `Supplier<String>`, as I think this probably should be lazy, though happy to add a `String` overload too. I did think about using a `SimpleModel` instead (that's just a `Supplier<String>`!), but not sure the Yarn name lends itself to this.

PepperCode1:
The Mojmap name of `SimpleModel` would make more sense here but since the key is created statically, it's not reasonable to pass any existing vanilla implementation of `SimpleModel` to the key anyway, so a plain `Supplier<String>` works fine. Nice work!

PepperCode1 (approved):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * <p>Extra models can be registered with a {@link ModelLoadingPlugin} (see
+ * {@link ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)}). Once baking is complete, they may
+ * then be queried from the model manager using {@link FabricBakedModelManager#getModel(ExtraModelKey)}.
+ *
+ * @param <T> The type of the baked model.
+ * @see FabricBakedModelManager#getModel(ExtraModelKey)
+ * @see ModelLoadingPlugin.Context#addModel(ExtraModelKey, UnbakedExtraModel)
+ */
+public final class ExtraModelKey<T> {
```
modmuss50, line 34: Am I missing something? Where is `name` used other than the .toString()? `ExtraModelKey` seems to only be used in a hashmap but never overrides hashcode?

Would this class be better as an interface an an impl record?

PepperCode1: `ExtraModelKey` is supposed to always have identity equivalence, and the name is only for debug purposes. Turning it into an interface (extendable) allows users to violate that contract, turning it into an interface (non-extendable) with a record impl would automatically violate that contract, and turning it into an interface (non-extendable) with a final class impl would be unnecessary abstraction for what is a mostly empty class.

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/impl/client/model/loading/ModelLoadingPluginContextImpl.java
```diff
@@ -30,17 +31,20 @@
...
 import net.fabricmc.fabric.api.client.model.loading.v1.ModelModifier;
+import net.fabricmc.fabric.api.client.model.loading.v1.UnbakedExtraModel;
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 public class ModelLoadingPluginContextImpl implements ModelLoadingPlugin.Context {
 	private static final Logger LOGGER = LoggerFactory.getLogger(ModelLoadingPluginContextImpl.class);
 
 	final Map<Block, BlockStateResolver> blockStateResolvers = new IdentityHashMap<>();
+	final Map<ExtraModelKey<?>, UnbakedExtraModel<?>> extraModels = new HashMap<>();
```
modmuss50, line 45: Thanks @PepperCode1 thats what I would have assumed, would this not be better off as an IdentityHashMap?

PepperCode1: Yes.

modmuss50 (approved):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ExtraModelKey.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package net.fabricmc.fabric.api.client.model.loading.v1;
+
+import java.util.function.Supplier;
+
+import org.jetbrains.annotations.Contract;
+
+/**
+ * A unique key representing an extra model, not tied to a blockstate or item model.
```
Juuxel:
```suggestion
 * A unique key representing an extra model, not tied to a block state or item model.
```

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/SimpleExtraModel.java
```diff
@@ -0,0 +1,96 @@
...
+	public SimpleExtraModel(Identifier model, BiFunction<BakedSimpleModel, Baker, T> bake) {
+		this.model = model;
+		this.bake = bake;
+	}
+
+	/**
+	 * Create a {@link SimpleExtraModel} for a {@link BlockStateModel}.
+	 *
+	 * <h2>Example</h2>
+	 * <pre>{@code
```
Juuxel: This could also be a `{@snippet}`, but I doubt it matters.

Juuxel (approved):

fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/SimpleExtraModel.java
SquidDev: Ahh, good shout. I'd stopped using `@snippet` with Checkstyle a while back, but I think as long as you avoid using Java annotations in the codeblock it's fine.

Juuxel: Yeah, it's already used in a few places in FAPI for examples (with or without `@link` spam).

PepperCode1:
I forgot to say this earlier but I would prefer if `SimpleExtraModel` were renamed to `SimpleUnbakedExtraModel` to make it clear that it is unbaked. The baked type is arbitrary so this should avoid confusion.

MrCrayfish:
Thank you for getting this reimplemented!

PR #4567, opened 2025-04-07
tooltip appender component registration api

cputnam-a11y:
Add Components be appended in ItemStack#appendTooltips

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 def branchProvider = providers.of(GitBranchValueSource.class) {}
-version = project.version + "+" + (providers.environmentVariable("CI").present ? branchProvider.get() : "local")
+version = project.version + "+" + (providers.environmentVariable("CI").present ? branchProvider.get().replace("/", "_") : "local")
 logger.lifecycle("Building Fabric: " + version)
 
 def metaProjects = [
@@ -115,7 +115,7 @@ allprojects {
 	apply plugin: "maven-publish"
 	apply plugin: "me.modmuss50.remotesign"
 
-	tasks.withType(GenerateModuleMetadata) {
+	tasks.withType(GenerateModuleMetadata).configureEach {
 		enabled = false
 	}
 
--- a/fabric-particles-v1/src/main/resources/fabric.mod.json
+++ b/fabric-particles-v1/src/main/resources/fabric.mod.json
@@ -17,7 +17,7 @@
   ],
   "depends": {
     "fabricloader": ">=0.16.10",
-    "fabric-networking-api-v1": ">=4.4.0"
+    "fabric-networking-api-v1": "*"
   },
   "entrypoints": {
     "main": [
--- a/gradle/module-versioning.gradle
+++ b/gradle/module-versioning.gradle
@@ -3,7 +3,7 @@
  * This task should be used to easily bump the major/minor/patch version of a fabric-api module.
  * It will automatically bump the versions of dependent modules.
  */
-task bumpVersions(type: BumpVersionTask)
+tasks.register('bumpVersions', BumpVersionTask)
 
 class BumpVersionTask extends DefaultTask {
 	BumpVersionTask() {
--- a/gradle/validate-annotations.gradle
+++ b/gradle/validate-annotations.gradle
@@ -1,6 +1,8 @@
-task validateAnnotations(type: ValidateAnnotations) {
+tasks.register('validateAnnotations', ValidateAnnotations) {
 	group = 'fabric'
 	description = "Validate annotations used in Fabric API code."
+
+	outputs.upToDateWhen { true } // Task has no outputs
 
 	// Only apply to default source directories since there's also generated package-info files.
 	source file("src/client/java")
@@ -9,7 +11,7 @@ task validateAnnotations(type: ValidateAnnotations) {
 	source file("src/testmodClient/java")
 }
 
-tasks.check.dependsOn validateAnnotations
+tasks.check.dependsOn "validateAnnotations"
 
 class ValidateAnnotations extends SourceTask {
 	private static final def API_STATUS_INTERNAL = ~/@ApiStatus\.Internal/


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
```diff
@@ -0,0 +1,206 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.tooltip.TooltipAppender;
+import net.minecraft.item.tooltip.TooltipType;
+import net.minecraft.text.Text;
+
+import net.fabricmc.fabric.api.item.v1.TooltipAppenderComponentRegistrationCallback;
+import net.fabricmc.fabric.impl.item.TooltipAppenderComponentRegistrationImpl;
+
+@Mixin(ItemStack.class)
+public abstract class ItemStackTooltipMixin {
```
modmuss50: We cannot maintain this, we need a solution that is resiliant to vanilla changes.

The easy option would just be to have a single event at the end, I'm not sure how imporant it is to have fine grain ordering like this?

Juuxel: I would probably inject before the `LORE` component's tooltip is rendered. It's pretty clearly meant to be the last component in the list since it's player-specified (never created in vanilla code).

cputnam-a11y:
Should there be any provision for ordering between mods, or just in registration order?

modmuss50:
> Should there be any provision for ordering between mods, or just in registration order?

No one has explicily asked for this, and I dont think its thats useful to do. I suggest we keep it simple like you have now and can always add that later if needed.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DataComponentTooltips.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package net.fabricmc.fabric.api.item.v1;
+
+import net.minecraft.component.ComponentType;
+import net.minecraft.item.tooltip.TooltipAppender;
+
+import net.fabricmc.fabric.impl.item.DataComponentTooltipsImpl;
+
+public final class DataComponentTooltips {
```
modmuss50: Make this an interface with `@ApiStatus.NonExtendable`

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/DataComponentTooltips.java
```diff
@@ -0,0 +1,31 @@
...
+	private DataComponentTooltips() {
+	}
+
+	public static void register(ComponentType<? extends TooltipAppender> type) {
```
modmuss50: Needs some documentation.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/DataComponentTooltipsImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import net.minecraft.component.ComponentType;
+import net.minecraft.item.tooltip.TooltipAppender;
+
+public final class DataComponentTooltipsImpl {
+	private static final List<ComponentType<? extends TooltipAppender>> types = new ArrayList<>();
+
+	private DataComponentTooltipsImpl() {
+	}
+
+	public static void register(ComponentType<? extends TooltipAppender> type) {
```
modmuss50: Check type isnt null

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/DataComponentTooltipsImpl.java
```diff
@@ -0,0 +1,42 @@
...
+		if (!types.contains(type)) {
+			types.add(type);
+		} else {
```
modmuss50: Use the early return pattern, it will make the code much easier to read. E.g

```
if (type.contains(type)) {
  throw new ...
}

types.add(...
```

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/DataComponentTooltipsImpl.java
```diff
@@ -0,0 +1,42 @@
...
+			throw new IllegalArgumentException("Component type " + type + " is already registered.");
+		}
+	}
+
+	public static List<ComponentType<? extends TooltipAppender>> getTypes() {
+		return new ArrayList<>(types);
```
modmuss50: Use Collections.unmodifiableList instead of creating a new array list.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
```diff
@@ -0,0 +1,63 @@
...
+	public abstract <T extends TooltipAppender> void appendComponentTooltip(ComponentType<T> componentType, Item.TooltipContext context, TooltipDisplayComponent displayComponent, Consumer<Text> textConsumer, TooltipType type);
+
+	@Inject(
+			method = "appendTooltip",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/item/ItemStack;appendComponentTooltip(Lnet/minecraft/component/ComponentType;Lnet/minecraft/item/Item$TooltipContext;Lnet/minecraft/component/type/TooltipDisplayComponent;Ljava/util/function/Consumer;Lnet/minecraft/item/tooltip/TooltipType;)V",
+					ordinal = 0
+			),
+			slice = @Slice(
```
modmuss50: Would it be easier to target the inject at the call to `appendAttributeModifiersTooltip`? then there is no need for the slice, or ordinal?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.tooltip.TooltipAppender;
+import net.minecraft.item.tooltip.TooltipType;
+import net.minecraft.text.Text;
+
+import net.fabricmc.fabric.impl.item.DataComponentTooltipsImpl;
+
+@Mixin(ItemStack.class)
+public abstract class ItemStackTooltipMixin {
```
modmuss50: Rename the mixin, it should be `ItemStackMixin` this should be failing the build in its current state. 🤔

modmuss50 (commented):
Looks like a good start, just a few minor tweaks. I do think keeping it simple like this is the way to go, at least for now.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
cputnam-a11y: Huh. Checkstyle said it passed?

cputnam-a11y: Should it merge with the existing ItemStackMixin then?

modmuss50: Ill look into why this didnt fail, yes this can go into the existing mixin

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
cputnam-a11y: that would be after the lore component. is that what is wanted now? earlier before lore was the last comment

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackTooltipMixin.java
cputnam-a11y: I'll do this. It can always be separated again if necessary.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
```diff
@@ -72,4 +85,11 @@ The other damage method (which original.call discards) handles the creative mode
 
 		original.call(instance, amount, serverWorld, serverPlayerEntity, consumer);
 	}
+
+	@Inject(method = "appendTooltip", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;appendComponentTooltip(Lnet/minecraft/component/ComponentType;Lnet/minecraft/item/Item$TooltipContext;Lnet/minecraft/component/type/TooltipDisplayComponent;Ljava/util/function/Consumer;Lnet/minecraft/item/tooltip/TooltipType;)V", ordinal = 0), slice = @Slice(from = @At(value = "FIELD", target = "Lnet/minecraft/component/DataComponentTypes;LORE:Lnet/minecraft/component/ComponentType;")))
```
Earthcomputer: Why lore specifically?

cputnam-a11y: Just what was requested

cputnam-a11y: Reasoning here: https://github.com/FabricMC/fabric/pull/4567#discussion_r2032842586

PR #4569, opened 2025-04-08
Ensure `BlockDustParticle` always uses accurate `BlockPos`

PepperCode1:
This pull request makes it so `BlockDustParticle` instances use an accurate `BlockPos` when supported. This is done by adding a `BlockPos` field to `BlockStateParticleEffect` and patching common code to set it to the correct value, when applicable. The `BlockStateParticleEffect`'s packet codec is patched to sync the field to the client if both the client and server have the Particles v1 module. The server only encodes extended data if it knows the client supports it, while decoding supports both vanilla data and extended data based on a marker. If a `BlockPos` was synced, the `BlockDustParticle` constructor that accepts a `BlockPos` is used.

The main purpose of this feature is to ensure that `FabricBlockStateModel#particleSprite(BlockRenderView, BlockPos, BlockState)` also always receives an accurate `BlockPos`. Without these patches, the `BlockPos` is sometimes off by one, which can cause the model to return an incorrect particle sprite, especially when using the `BlockPos` to retrieve block entity render data or the biome. The FRAPI test mod has such models, which can be used for testing.

### TODO

- [x] Decide how to expose the ability to create a `BlockStateParticleEffect` with a `BlockPos` to API users.
  **Resolution:** Added interface injected `FabricBlockStateParticleEffect` with `BlockPos` getter and static `create` method.

### DIFF
--- a/fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
+++ b/fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
@@ -23,6 +23,8 @@ import net.fabricmc.fabric.impl.particle.ExtendedBlockStateParticleEffectSync;
 public class ExtendedBlockStateParticleEffectSyncClient implements ClientModInitializer {
 	@Override
 	public void onInitializeClient() {
+		// Register a receiver so ExtendedBlockStateParticleEffectSync#shouldEncodeFallback can detect that this client
+		// supports extended data
 		ClientConfigurationNetworking.registerGlobalReceiver(ExtendedBlockStateParticleEffectSync.DummyPayload.ID, (payload, context) -> {
 		});
 	}
--- a/fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
+++ b/fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.impl.client.particle;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.networking.v1.ClientConfigurationNetworking;
+import net.fabricmc.fabric.impl.particle.ExtendedBlockStateParticleEffectSync;
+
+public class ExtendedBlockStateParticleEffectSyncClient implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		ClientConfigurationNetworking.registerGlobalReceiver(ExtendedBlockStateParticleEffectSync.DummyPayload.ID, (payload, context) -> {
+		});
+	}
+}
--- a/fabric-particles-v1/src/main/java/net/fabricmc/fabric/impl/particle/BlockStateParticleEffectExtension.java
+++ b/fabric-particles-v1/src/main/java/net/fabricmc/fabric/impl/particle/BlockStateParticleEffectExtension.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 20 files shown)


fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
```diff
@@ -0,0 +1,29 @@
...
+package net.fabricmc.fabric.impl.client.particle;
+
+import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.fabric.api.client.networking.v1.ClientConfigurationNetworking;
+import net.fabricmc.fabric.impl.particle.ExtendedBlockStateParticleEffectSync;
+
+public class ExtendedBlockStateParticleEffectSyncClient implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		ClientConfigurationNetworking.registerGlobalReceiver(ExtendedBlockStateParticleEffectSync.DummyPayload.ID, (payload, context) -> {
```
modmuss50, line 28: Nit: a comment explaing why this doesnt do anything might be useful for a future reader.

fabric-particles-v1/src/main/java/net/fabricmc/fabric/impl/particle/BlockStateParticleEffectExtension.java
```diff
@@ -0,0 +1,28 @@
...
+
+package net.fabricmc.fabric.impl.particle;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.util.math.BlockPos;
+
+public interface BlockStateParticleEffectExtension {
+	@Nullable
+	BlockPos getBlockPos();
```
modmuss50: Prefix these 2 methods with `fabric_`

fabric-particles-v1/src/client/java/net/fabricmc/fabric/impl/client/particle/ExtendedBlockStateParticleEffectSyncClient.java
modmuss50: Maybe just add a `context.responseSender().disconnect();`, can use the text as the comment.

fabric-particles-v1/src/main/java/net/fabricmc/fabric/mixin/particle/BlockStateParticleEffectMixin.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.particle.BlockStateParticleEffect;
+import net.minecraft.util.math.BlockPos;
+
+import net.fabricmc.fabric.impl.particle.BlockStateParticleEffectExtension;
+import net.fabricmc.fabric.impl.particle.ExtendedBlockStateParticleEffectPacketCodec;
+
+@Mixin(BlockStateParticleEffect.class)
+abstract class BlockStateParticleEffectMixin implements BlockStateParticleEffectExtension {
+	@Nullable
+	private BlockPos blockPos;
```
modmuss50, line 38: @ Unique

modmuss50 (approved):

modmuss50 (approved):
Great, thanks 👍

kennytv:
Does this really need to modify a packet to be Vanilla-incompatible by default, in an API mod without anything else? :<

modmuss50:
> Does this really need to modify a packet to be Vanilla-incompatible by default, in an API mod without anything else? :<

What issues are you having? It seems to have caused more problems that its fixed. This isnt the first time we have modified a vanilla packet (ingredients) so we were under the assumption this wouldnt be a massive problem, thats clearly not the case though.

Im very keen to find a workable solution for everyone, if it gets to needing to revert this change so be it.

kennytv:
Generally proxies or software like ViaVersion assume that such data can be read as Vanilla would read it, unless you intentionally add mods beyond what Vanilla can provide. in this case it's nothing that can't be fixed with an extra if statement and is of course cursed anways, but ViaVersion trips on it when trying to map the -1 block state id / not expecting more unread data

kennytv:
Probably not as simple as this to implement on your end, but having custom data like this go through custom payloads, or otherwise its own (singular/static) special packet id that can always be passed through as is, would make handling it a lot easier. At least regarding those 'base' modifications, anything else that mods use to modify client registries and other opt-in  behavior is of course a different thing

modmuss50:
Yeah, the only other solution for the original issue that comes to mind is sending it in a custom payload under the fabric namespace, tottaly replacing the vanilla packet. If anything its likely easier to implement on our end.

You would still end up with issues when using ViaVersion if the server tries to send it to a client on another version, so you would still need handling for it in VV. We tend not to care about network compatibility across Minecraft versions, maybe we should. Please do let me know if there is anything you think we can do better or provide to make your lives easier.

Is this something you think we should change now or just keep in mind for the future packets now that the damage has been done?

kennytv:
I'd imagine you doing changes to your own custom data would happen less frequently than additions (and both probably infrequent?), but each own sort of subpacket there having a protocol version would also help that, if not just a global version. Anything in that direction within custom payloads would be great! and yeah, removing the change to the particle data here would be nice, it's the only such compatibility issue with Fabric API I'm aware of as of now, ignoring a more complex one around stepping behind the version the config protocol phase was added

PepperCode1:
Has ViaVersion tried breaking the Recipe API and Particle API modules, which potentially extend vanilla payload data? Recipe API only adds custom ingredients, which ViaVersion cannot support regardless of how they are sent, and Particle API allows registering the rendering for custom particle types, which ViaVersion also doesn't support. According to Chocohead, breaking modules that are not depended on will disable them, which will solve the issue.

PR #4587, opened 2025-04-17
Add component tooltip appender registry

Earthcomputer:
Supersedes #4567.

Adds a way to register `TooltipAppender` item components. This implementation allows modders to anchor their item components relative to any vanilla (or modded) item component, while not hardcoding the list or order of vanilla item components. It's annoying to need to scrape the vanilla order from the bytecode, but at least it should be low maintenance.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
@@ -29,7 +29,6 @@ import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
  */
 @ApiStatus.NonExtendable
 public interface ComponentTooltipAppenderRegistry {
-
 	/**
 	 * Adds the specified item component type to the list of tooltip appenders to be called first. The component will
 	 * render at the top of the tooltip.
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
@@ -16,12 +16,12 @@
 
 package net.fabricmc.fabric.api.item.v1;
 
-import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
-
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.component.ComponentType;
 import net.minecraft.item.tooltip.TooltipAppender;
+
+import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
 
 /**
  * A registry of {@link TooltipAppender} item components. Adding your item component to this registry will render the
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
@@ -16,12 +16,12 @@
 
 package net.fabricmc.fabric.api.item.v1;
 
+import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
+
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.component.ComponentType;
 import net.minecraft.item.tooltip.TooltipAppender;
-
-import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
 
 /**
  * A registry of {@link TooltipAppender} item components. Adding your item component to this registry will render the
@@ -29,9 +29,6 @@ import net.fabricmc.fabric.impl.item.ComponentTooltipAppenderRegistryImpl;
  */
 @ApiStatus.NonExtendable
 public interface ComponentTooltipAppenderRegistry {
-	static ComponentTooltipAppenderRegistry getInstance() {
-		return ComponentTooltipAppenderRegistryImpl.INSTANCE;
-	}
 
 	/**
 	 * Adds the specified item component type to the list of tooltip appenders to be called first. The component will
@@ -39,7 +36,9 @@ public interface ComponentTooltipAppenderRegistry {
 	 *
 	 * @param componentType the component type to add
 	 */
-	void addFirst(ComponentType<? extends TooltipAppender> componentType);
+	static void addFirst(ComponentType<? extends TooltipAppender> componentType) {
+		ComponentTooltipAppenderRegistryImpl.addFirst(componentType);
+	}
 
 	/**
 	 * Adds the specified item component type to the list of tooltip appenders to be called last. The component will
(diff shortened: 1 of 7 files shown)


fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A registry of {@link TooltipAppender} item components. Adding your item component to this registry will render the
+ * item component to the tooltip of them item when it is present, in a location relative to other item components.
+ */
+@ApiStatus.NonExtendable
+public interface ComponentTooltipAppenderRegistry {
+	static ComponentTooltipAppenderRegistry getInstance() {
+		return ComponentTooltipAppenderRegistryImpl.INSTANCE;
+	}
```
Juuxel: I think you can just use static methods here. New APIs don't really use this pattern

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ComponentTooltipAppenderRegistryImpl.java
```diff
@@ -0,0 +1,148 @@
...
+import net.minecraft.text.Text;
+
+public final class ComponentTooltipAppenderRegistryImpl {
+	private static final List<ComponentType<? extends TooltipAppender>> first = new ArrayList<>();
+	private static final List<ComponentType<? extends TooltipAppender>> last = new ArrayList<>();
+	private static final Map<ComponentType<?>, List<ComponentType<? extends TooltipAppender>>> before = new HashMap<>();
+	private static final Map<ComponentType<?>, List<ComponentType<? extends TooltipAppender>>> after = new HashMap<>();
+
+	public static void addFirst(ComponentType<? extends TooltipAppender> componentType) {
+		Preconditions.checkNotNull(componentType, "componentType");
```
TheDeathlyCow: I think this sort of thing is usually done with `Objects.requireNonNull` in the API method, isn't it?

Earthcomputer: Oh yeah this is possible now I made the ComponentTooltipAppenderRegistry methods static

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/ComponentTooltipAppenderRegistryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.tooltip.TooltipAppender;
+import net.minecraft.item.tooltip.TooltipType;
+import net.minecraft.text.Text;
+
+public final class ComponentTooltipAppenderRegistryImpl {
+	private static final List<ComponentType<? extends TooltipAppender>> first = new ArrayList<>();
+	private static final List<ComponentType<? extends TooltipAppender>> last = new ArrayList<>();
+	private static final Map<ComponentType<?>, List<ComponentType<? extends TooltipAppender>>> before = new HashMap<>();
```
modmuss50: Would an IdentityHashMap be more suited here?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/VanillaTooltipAppenderOrder.java
```diff
@@ -0,0 +1,141 @@
...
+public final class VanillaTooltipAppenderOrder {
+	private static final List<ComponentType<?>> VANILLA_ORDER = scrapeVanillaOrder();
+
+	private VanillaTooltipAppenderOrder() {
+	}
+
+	// Find the order in which vanilla tooltip appenders are run by inspecting the bytecode of ItemStack.appendTooltip.
+	private static List<ComponentType<?>> scrapeVanillaOrder() {
+		try {
+			ClassNode itemStackNode = MixinService.getService().getBytecodeProvider().getClassNode(Type.getInternalName(ItemStack.class));
```
modmuss50, line 59: Just to confirm this returns the classnode *before* any mixin has been applied? So mods wont have a chance to interfer with it?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/VanillaTooltipAppenderOrder.java
```diff
@@ -0,0 +1,141 @@
...
+					// Special case: attribute modifiers are extracted into a separate method
+					componentTypes.add(DataComponentTypes.ATTRIBUTE_MODIFIERS);
+				}
+			}
+
+			if (componentTypes.isEmpty()) {
+				throw new IllegalStateException("Found no component types in appendTooltip method");
+			}
+
+			return componentTypes;
```
modmuss50: Nit: Collections.unmodifiableList

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/VanillaTooltipAppenderOrder.java
Earthcomputer: Correct

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/VanillaTooltipAppenderOrder.java
Earthcomputer: Is this necessary? It's only used in impl

modmuss50: Not necessary no hence it being a nit pick, but good practice.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/impl/item/VanillaTooltipAppenderOrder.java
```diff
@@ -0,0 +1,141 @@
...
+import net.fabricmc.loader.api.FabricLoader;
+
+public final class VanillaTooltipAppenderOrder {
+	private static final List<ComponentType<?>> VANILLA_ORDER = scrapeVanillaOrder();
+
+	private VanillaTooltipAppenderOrder() {
+	}
+
+	// Find the order in which vanilla tooltip appenders are run by inspecting the bytecode of ItemStack.appendTooltip.
+	private static List<ComponentType<?>> scrapeVanillaOrder() {
```
modmuss50, line 57: This currently only gets called the first time you go to draw a stacks tooltip, I think it would be better also done at startup to not block the render thread, and find issues quickly.

modmuss50:
This needs some automated testing, in its current form I dont believe any of these code is even exercised. I wrote a basic unit test here that you can use: https://gist.github.com/modmuss50/b134ce6026f6ddf6b76479e785c1ad2c

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/mixin/item/ItemStackMixin.java
```diff
@@ -72,4 +90,124 @@ The other damage method (which original.call discards) handles the creative mode
...
+	@Unique
+	private void preAppendTooltip(
+			@Nullable ComponentType<?> componentType,
+			Item.TooltipContext context,
+			TooltipDisplayComponent displayComponent,
+			Consumer<Text> textConsumer,
+			TooltipType tooltipType,
+			LocalIntRef index
+	) {
+		if (index.get() == 0) {
```
modmuss50, line 177: Would it be a good idea to skip over all of this logic if no mod has used the registry? Just have something along the lines of `if (ComponentTooltipAppenderRegistryImpl.unmodified()) return;`. Would be easy enough to do, and would save a bit of frame time in vanilla?

modmuss50 (approved):

PR #4590, opened 2025-04-20
Split large RecipeBookAddS2CPacket's in a vanilla client compatible way.

modmuss50:
In a pack with a large amount of recipes the RecipeBookAddS2CPacket can exceed the packet size limit. This PR splits the single RecipeBookAddS2CPacket into multiple packets when needed. This is compatible with vanilla clients as they can accept multiple packets.

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/RecipeBookAddPacketSplitter.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/RecipeBookAddPacketSplitter.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/RecipeBookAddPacketSplitter.java
```diff
@@ -0,0 +1,70 @@
...
+				// Individual entry is larger than the max size, cannot be sent.
+				throw new IllegalStateException("Entry is larger than the max size: " + entry);
+			}
+
+			if (byteBuf.readableBytes() >= MAXIMUM_PACKET_SIZE) {
+				// Packet would be larger than the max size, send what fits, the current entry will be sent in the next packet.
+				packetSender.accept(new RecipeBookAddS2CPacket(collectedEntries, shouldReplace));
+
+				collectedEntries = new ArrayList<>();
+				byteBuf = new RegistryByteBuf(PacketByteBufs.create(), networkHandler.getPlayer().getRegistryManager());
```
modmuss50, line 58: Can we reuse the old byte buf? Or maybe even use this one as the packet?

PR #4592, opened 2025-04-21
Attachment modify event

SHsuperCM:
Initially I needed a listener for when attachments are synced but I have now expanded it to be usable for more mods while I was here anyways. 
This PR introduces a simple event in the attachment type that both lets users get notified whenever an attachment value is modified on a target while also allowing the listener to modify the value before it is set.

As was requested by @Syst3ms I also added a parameter to signal to the listener whether the world is a client world or not when applicable.

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
@@ -19,6 +19,9 @@ package net.fabricmc.fabric.api.attachment.v1;
 import java.util.function.Supplier;
 
 import com.mojang.serialization.Codec;
+
+import net.fabricmc.fabric.api.event.Event;
+
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
@@ -111,4 +114,10 @@ public interface AttachmentType<A> {
 	 * when a mob is converted (e.g. zombie → drowned)
 	 */
 	boolean copyOnDeath();
+
+	Event<AttachmentModified<A>> modify();
+
+	public interface AttachmentModified<A> {
+		A onModify(A newValue, AttachmentTarget target, boolean isClient);
+	}
 }
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
@@ -19,9 +19,6 @@ package net.fabricmc.fabric.api.attachment.v1;
 import java.util.function.Supplier;
 
 import com.mojang.serialization.Codec;
-
-import net.fabricmc.fabric.api.event.Event;
-
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
@@ -36,6 +33,7 @@ import net.minecraft.world.chunk.WorldChunk;
 import net.fabricmc.fabric.api.entity.event.v1.ServerEntityWorldChangeEvents;
 import net.fabricmc.fabric.api.entity.event.v1.ServerLivingEntityEvents;
 import net.fabricmc.fabric.api.entity.event.v1.ServerPlayerEvents;
+import net.fabricmc.fabric.api.event.Event;
 
 /**
  * An attachment allows "attaching" arbitrary data to various game objects (entities, block entities, worlds and chunks at the moment).
@@ -121,7 +119,7 @@ public interface AttachmentType<A> {
 	Event<AttachmentModified<A>> modify();
 
 	@FunctionalInterface
-	public interface AttachmentModified<A> {
+	interface AttachmentModified<A> {
 		/**
 		 * Called right before the attachment is set and lets the listener change or cancel the operation.
 		 *
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
@@ -116,7 +116,7 @@ public interface AttachmentType<A> {
 	/**
 	 * An event that is called right before this type of attachment changes on a target.
 	 */
-	Event<AttachmentModified<A>> modify();
+	Event<AttachmentModified<A>> onModify();
 
 	@FunctionalInterface
 	interface AttachmentModified<A> {
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
@@ -115,9 +115,23 @@ public interface AttachmentType<A> {
 	 */
 	boolean copyOnDeath();
 
+	/**
+	 * An event that is called right before this type of attachment changes on a target.
+	 */
 	Event<AttachmentModified<A>> modify();
 
+	@FunctionalInterface
 	public interface AttachmentModified<A> {
+		/**
+		 * Called right before the attachment is set and lets the listener change or cancel the operation.
+		 *
+		 * <p>The old attachment value is available on the target with {@code getAttached}.</p>
+		 *
+		 * @param newValue new attachment that is about to be set, null if attachment is being removed
+		 * @param target target to set the value upon
+		 * @param isClient isClient from the associated world when applicable, false otherwise
+		 * @return value to set onto the target, return {@code newValue} to avoid modifying it, return null to remove the attachment.
+		 */
 		A onModify(A newValue, AttachmentTarget target, boolean isClient);
 	}
 }
--- a/fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/AttachmentTestMod.java
+++ b/fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/AttachmentTestMod.java
@@ -116,7 +116,7 @@ public class AttachmentTestMod implements ModInitializer {
 			if (!isClient) {
 				ItemStack oldValue = target.getAttached(SYNCED_ITEM);
 
-				if (oldValue != null && oldValue.isOf(Items.BAMBOO) && newValue.isOf(Items.BONE_MEAL)) {
+				if (oldValue != null && oldValue.isOf(Items.APPLE) && newValue.isOf(Items.BONE_MEAL)) {
 					return oldValue.copyWithCount(oldValue.getCount() + 1);
 				}
 			}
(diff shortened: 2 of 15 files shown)


SHsuperCM:
I still need to give it a proper test before it's ready for review and I was wondering if I should also add its usage to the testmod.

Technici4n:
What exactly is the use case? This is an event that can potentially fire a lot, and there is a lot of damage potential as well.

SHsuperCM:
> What exactly is the use case? This is an event that can potentially fire a lot, and there is a lot of damage potential as well.

For me specifically I need to call `Entity#calculateDimensions` clientside after an attachment is synced (the dimensions check for something in the attachment).
There is no other way to tell if an attachment was synced and really changed in general without mixing into fapi. 
After a discussion with @Syst3ms in #api I made this PR to remedy that.
There was a concern about the amount of times the event fires and to handle that we said it'd be better to have the event scoped to an `AttachmentType`.

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/AttachmentTestMod.java
Syst3ms, line 1: There should be a gametest for modify() affecting the client. It's the use case that motivates this, and I have a feeling that the method being server-side could make this annoying with split sources.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -111,4 +112,24 @@ default boolean isPersistent() {
 	 * when a mob is converted (e.g. zombie → drowned)
 	 */
 	boolean copyOnDeath();
+
+	/**
+	 * An event that is called right before this type of attachment changes on a target.
+	 */
+	Event<AttachmentModified<A>> modify();
```
Syst3ms: Bikeshedding: any reason this shouldn't be something like `onModify`? The name feels a bit counterintuitive.

Syst3ms (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
SHsuperCM: Oh I just have a convention I go by usually, when I read that I see "(get)Modify" (noun) so it makes sense to me. I can see how it could be less clear tho so I dont mind changing it. Will do in a bit

fabric-data-attachment-api-v1/src/testmod/java/net/fabricmc/fabric/test/attachment/AttachmentTestMod.java
SHsuperCM: Agreed on the former, I'll think something up thats simple to see work in a test.
On the latter I'm not sure I understand? You still have to do world.isClient when working in split sources

Syst3ms: i mean that if you define the lambda in a non-client sourceset you won't be able to refer to any client classes which could potentially be a headache

SHsuperCM: How do people usually go around this with, say, `UseItemCallback`? I dont use split sources so I dont really know what's the way to handle that(or if that is an issue at all for that matter)

SHsuperCM: Resolving the gametest thing, lmk if you want to reopen the thing about split sources.

Syst3ms (approved):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -111,4 +112,24 @@ default boolean isPersistent() {
...
+	interface AttachmentModified<A> {
+		/**
+		 * Called right before the attachment is set and lets the listener change or cancel the operation.
+		 *
+		 * <p>The old attachment value is available on the target with {@code getAttached}.</p>
+		 *
+		 * @param newValue new attachment that is about to be set, null if attachment is being removed
+		 * @param target target to set the value upon
+		 * @param isClient isClient from the associated world when applicable, false otherwise
+		 * @return value to set onto the target, return {@code newValue} to avoid modifying it, return null to remove the attachment.
```
modmuss50, line 131: Whats the use case for modifying the value set?

I worry this will make it harder to reason about using .setAttached, as when I call that I would expect the value that has been specified to be the one that actually gets set.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
```diff
@@ -111,4 +112,24 @@ default boolean isPersistent() {
...
+		 */
+		A onModify(A newValue, AttachmentTarget target, boolean isClient);
```
modmuss50, line 133: Can you not get the world from the target? thus making the isClient redundent?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
SHsuperCM: Honestly to avoid a StackOverflowError was my reasoning. I feel like it'd be better to not call `setAttached` on the target while already running `setAttached`. 
I have no doubt that there will be someone who needs to cancel or change the value set and would rather do this now. 

Do note that if we make this only a listener and not a modify we would have to move the hook for after it is actually set so that the listener may use `setAttached` again without it being ignored. Would also likely need to add the old value to the arguments of the event because it wont be retrievable after it is set.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
SHsuperCM: I somewhat agree but I was asked by @Syst3ms to add this as an argument. I would ask them for the reasoning behind it.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
modmuss50: Being able to change the value at all in this event is technically a breaking change imo, if not a massive source of issues. Considering the following code:

```java
// Before
if (something) {
	be.setAttached(AttachmentTestMod.PERSISTENT, "test");
	System.out.println("Do something");
}

// With the new API
if (something) {
	be.setAttached(AttachmentTestMod.PERSISTENT, "test");
	if ("test".equals(be.getAttached(AttachmentTestMod.PERSISTENT))) {
		System.out.println("Do something");
	}
}
```

With this change you must now query the value that was actually set after you set it if your "Do something" requires the data to be set.


I think we should remove this, and maybe document that `setAttached` should never be used within the callback.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentType.java
modmuss50: I dont see how its useful, it also doesnt tell if you if the value change was coming from a network sync or not. Im not saying to add that as I dont think it should matter to someone using this event.

modmuss50:
Im slightly worried this event will cause more harm than good, it subtly changes a lot of guarentees that the API provided in the past alongside the performance concerns, possibly thread safety confusion (as you have no idea what thread your event will be fired on).

Does/should the event get fired when the attachment is removed, when the target is created, with this there will no longer be a single source of truth to the data.

SHsuperCM:
I am okay with scaling the event back to not be a modify but just a detection, it is more than I need for my current use case even though I still believe it will be needed by someone in the future.

I still think however that the arguments against it being a modify are already true for most fapi events, where it breaks expected behavior because a mod needed it changed. It is down to the modder to decide what to do with it and it will happen regardless of if an event allows for it. 
I just think the foresight now is better than the alternative of having to mixin into fapi when an api could be exposed.

modmuss50:
I dont think you need to mixin into FAPI at all, for your use case you can just get the attached data each tick, and check if it changed from the previous tick. This solves the whole data consistency and threading worries.

SHsuperCM:
That is exactly what I am doing 😅 I was looking for ways to avoid this because it's irrational for performance. Came down to three options.
1. Mixin into fapi internals
2. I PR a proper event for it into fapi
3. Just look for a different API for this, perhaps just sync it myself to a client state and not use att sync at all

I wanted to really avoid mixing into fapi so I wanted to see if the 2nd one fits better

modmuss50:
> irrational for performance

Are you sure this is the case? I worry you might be trying to optimise something that isnt slow to start with. This PR has a none-zero overhead, and opens it up for mods to do all kinds of stuff when setting the attachment. Im not against this change, but we do need to think carefully if its something we want.

SHsuperCM:
If the problem is the modification within the event I could get rid of it and just have it be a detection that it changed to notify the listener.

I am confident that it is irrational for both performance but also code complexity to check if data was changed every tick instead of being notified and letting the developer choose how to handle it. I do understand that performance wise it would be a micro optimization but I think for complexity it is more than worth it.

I started implementing the client state with attsync as it seemed like the proper solution provided by fapi to achieve my goal, but without this event it feels like I'm taking a hit for using it and introducing complexity where there doesnt need to be. 
I dont mind switching to something else if attsync is not supposed to be the answer to this but it's just more effort I got to put into the mod.

modmuss50:
> If the problem is the modification within the event I could get rid of it and just have it be a detection that it changed to notify the listener.

I think this would solve most of the issues for sure. You can still pass the new value though to it.

Im just thinking outloud would there be any benefits to being able to register the listener on the specific target you care about? Doing something like this would tottaly remove the need to pass the client though, and ensure that you know what the attachment target is always going to be?

```java
ClientEntityEvents.ENTITY_LOAD.register((entity, world) -> {
	entity.onAttachedChanged(Example.TYPE, (value) -> {

	});
```

For your own (block)entities you could just put this in the constructor of the class.

SHsuperCM:
Hadn't thought of that.

There definitely would be a benefit, allowing you to filter out unwanted target types ahead of time without needing to check within the event on each change. Could also filter within the target type if for example you want a specific `EntityType` which sounds useful. You could register it also on the side you need ahead of time.

I believe most of the time attachment types would be specific to a target type? if not you would have to register a listener per type's "LOAD" event which is the only downside I can think of.

Also how would that look on the target? My immediate thought is a `IdentityHashMap<AttachmentType, Event<Changed>>` with a `computeIfAbsent` on the `onAttachedChanged`. Does that sound right? How much worse would that be for overhead to get out of that map I wonder. 
I think it would make sense to also just expose the `Event` out of the method and let the getter use its `.register` explicitly, would making the `Event` accessible be an issue?

SHsuperCM:
Superseded by #4606

PR #4596, opened 2025-04-23
Fix HUD client tests on Windows

modmuss50:
Vanilla draws a 256x image across the entire screen, causing the pixels to be squished. On Windows they are diffrent size compared to MacOS and Linux, this PR resizes the window to make sure that the pixels are always square, thus the same across different OSes.


Octol1ttle (approved):
this issue is likely driver-specific rather than OS-specific, but the PR fixed it for me (running Linux with KDE on Wayland)

Earthcomputer (approved):
Good enough workaround for now. Maybe later we can come up with a workaround in the API

kevinthegreat1 (approved):
LGTM

PR #4597, opened 2025-04-23
Add system property to filter mods for client tests.

modmuss50:
Adds a simple CSV seperated list of mod IDs that should have their client game tests ran. `-Dfabric.client.gametest.modid=fabric-rendering-v1-testmod`.

Defaults to the old behaviour of running all tests from all mods.

### DIFF
--- a/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/FabricClientGameTestRunner.java
+++ b/fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/FabricClientGameTestRunner.java
@@ -35,11 +35,6 @@ import net.fabricmc.loader.api.entrypoint.EntrypointContainer;
 public class FabricClientGameTestRunner {
 	private static final String ENTRYPOINT_KEY = "fabric-client-gametest";
 
-	/**
-	 * A comma separated list of mod ids to run tests for. When empty, all tests are run.
-	 */
-	private static final String MOD_ID_FILTER_PROPERTY = "fabric.client.gametest.modid";
-
 	public static EntrypointContainer<FabricClientGameTest> currentlyRunningGameTest = null;
 
 	public static void start() {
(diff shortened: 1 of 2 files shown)


fabric-client-gametest-api-v1/src/client/java/net/fabricmc/fabric/impl/client/gametest/FabricClientGameTestRunner.java
```diff
@@ -32,13 +35,18 @@
 public class FabricClientGameTestRunner {
 	private static final String ENTRYPOINT_KEY = "fabric-client-gametest";
 
+	/**
+	 * A comma separated list of mod ids to run tests for. When empty, all tests are run.
+	 */
+	private static final String MOD_ID_FILTER_PROPERTY = "fabric.client.gametest.modid";
```
Earthcomputer: This could go in `TestSystemProperties`

modmuss50: No idea how I missed that, now fixed 👍

Earthcomputer (approved):

SHsuperCM:
Since this is on last call, thought it'd be worth asking.
Would it make sense to also let you filter by the entrypoint definition instead of(or in addition to) the mod id?

I already added a definition string in `EntrypointContainer` and it shouldnt be hard to filter by. 
Also may be useful for MCDev to let the user directly click `Run` on a specific gametest instead of the whole mod?

modmuss50:
> Since this is on last call, thought it'd be worth asking. Would it make sense to also let you filter by the entrypoint definition instead of(or in addition to) the mod id?
> 
> I already added a definition string in `EntrypointContainer` and it shouldnt be hard to filter by. Also may be useful for MCDev to let the user directly click `Run` on a specific gametest instead of the whole mod?

Being able to run/debug tests from the IDE would be great but I have no clue where to start. I think for now just filtering the modid is good enough, its the same as what we do for the regular gametests and datagen. If there is a big need that could always be added later.

PR #4599, opened 2025-04-27
Add a tracked data handler registry (continuation of #3584)

PepperCode1:
This pull request is a continuation of #3584 and thus supersedes it. I am not the author of the original pull request and thus it is difficult for me to make the necessary changes to it directly. However, the commits from the original pull request have been included in this one to preserve authorship.

Some minor adjustments have been made compared to the current state of the original pull request, namely updating the code to 1.21.5, adding documentation, addressing a couple other minor review comments, renaming some fields/methods, and adding some getter methods to the API. The implementation of `FabricTrackedDataRegistryImpl.register` has some issues and unresolved comments from the original pull request so it needs additional discussion. Otherwise, all other changes should be ready.


fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraft.entity.data.TrackedDataHandler;
+import net.minecraft.entity.data.TrackedDataHandlerRegistry;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.object.builder.FabricTrackedDataRegistryImpl;
+
+/**
+ * Allows registering custom {@link TrackedDataHandler}s in a reliable way.
+ */
+public final class FabricTrackedDataRegistry {
```
modmuss50, line 30: Nit: Make this a `@ApiStatus.NonExtendable` interface.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	/**
+	 * Retrieves the ID for the given handler, or {@code null} if the handler was not registered with
+	 * {@link #register(Identifier, TrackedDataHandler)}.
+	 */
+	@Nullable
+	public static Identifier getId(TrackedDataHandler<?> handler) {
+		return FabricTrackedDataRegistryImpl.getId(handler);
+	}
```
modmuss50, line 63: Question: Whats the use case for these being public api?

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,160 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.collection.Int2ObjectBiMap;
+
+import net.fabricmc.fabric.api.event.registry.FabricRegistryBuilder;
+import net.fabricmc.fabric.api.event.registry.RegistryAttribute;
+import net.fabricmc.fabric.api.event.registry.RegistryIdRemapCallback;
+import net.fabricmc.fabric.mixin.object.builder.TrackedDataHandlerRegistryAccessor;
+
+public final class FabricTrackedDataRegistryImpl {
+	public static final Logger LOGGER = LoggerFactory.getLogger(FabricTrackedDataRegistryImpl.class);
```
modmuss50: Logger should be private.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,160 @@
...
+
+	private static final Identifier HANDLER_REGISTRY_ID = Identifier.of("fabric-object-builder-api-v1", "tracked_data_handler");
+	private static final RegistryKey<Registry<TrackedDataHandler<?>>> HANDLER_REGISTRY_KEY = RegistryKey.ofRegistry(HANDLER_REGISTRY_ID);
+
+	private static boolean hasStoredVanillaHandlers = false;
```
modmuss50: Why a seperate boolean? Just use `!VANILLA_HANDLERS.isEmpty()`

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
```diff
@@ -0,0 +1,160 @@
...
+	 * <p>The order used is as follows:
+	 *
+	 * <ul>
+	 *   <li>Vanilla handlers</li>
+	 *   <li>Handlers in the Fabric API registry (sorted by ID)</li>
+	 *   <li>External modded handlers</li>
+	 * </ul>
+	*/
+	private static void reorderHandlers() {
+		Int2ObjectBiMap<TrackedDataHandler<?>> dataHandlers = TrackedDataHandlerRegistryAccessor.fabric_getDataHandlers();
```
modmuss50, line 95: Question: is there a reason why this is sorted, and doesnt use the raw registry id?

This is going to have a similar issue where if the client has addional mods that arent present on the server the ids wont sync the same. Where as registry sync correctly handles this case.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
Juuxel: Why? All the methods are static, so it doesn't make sense for the class to be an itf.

modmuss50: I thought this was part of the contributing guidelines, but its not. This is fine.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
PepperCode1: The logger is used in the mixin, so it can't be private.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
PepperCode1: From my experience, a use case for inspection comes up eventually, so I tend to add trivial inspection methods immediately. I think these methods are especially appropriate here as vanilla exposes integer equivalents. I don't think this causes a maintenance burden.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
PepperCode1: I thought it did use registry integer IDs when iterating on the registry. I assume they can be out of order or have gaps, then? If so, would it be best to add custom entries to the map using their registry integer ID + the number of vanilla handlers?

modmuss50: > I thought it did use registry integer IDs when iterating on the registry. 

It does, I forgot about that. I think what I had in mind was to also add the vanilla ones, but there isnt really any benefit to doing that so what you have here is probably correct. Ill double check myself shortly.

I do wonder if the external handlers need to go before the ones in the registry? I think this approach makes the most sense as it priorities keeping the handlers tracked via the api in sync.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
modmuss50: Use a string name for a logger in the mixin.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/entity/FabricTrackedDataRegistry.java
modmuss50: > I don't think this causes a maintenance burden.

It doesnt, I'm just not convined its useful at all. Its more likely a mod will want want the integer id. Im happy either way if you want these leave them.

fabric-object-builder-api-v1/src/testmod/java/net/fabricmc/fabric/test/object/builder/EntityTrackedDataTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class EntityTrackedDataTest implements ModInitializer {
+	private static final Identifier GLOBAL_POS_ID = ObjectBuilderTestConstants.id("global_pos");
+	static TrackedDataHandler<GlobalPos> GLOBAL_POS = TrackedDataHandler.create(GlobalPos.PACKET_CODEC);
+
+	private static final Identifier ITEM_ID = ObjectBuilderTestConstants.id("item");
+	static TrackedDataHandler<Item> ITEM = TrackedDataHandler.create(PacketCodecs.registryValue(RegistryKeys.ITEM));
+
+	private static final Identifier OPTIONAL_ITEM_GROUP_ID = ObjectBuilderTestConstants.id("optional_item_group");
+	static TrackedDataHandler<Optional<ItemGroup>> OPTIONAL_ITEM_GROUP = TrackedDataHandler.create(PacketCodecs.registryValue(RegistryKeys.ITEM_GROUP).collect(PacketCodecs::optional));
```
modmuss50: `PacketCodecs.registryValue(RegistryKeys.ITEM_GROUP)` is correctly causing the tests to fail, please change this to another registry.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
PepperCode1: I thought about this but it has the significant issue that if the amount of external handlers is not the same across the client and server, handlers added through Fabric API will start at different integer IDs and they will all be desynced.

fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/impl/object/builder/FabricTrackedDataRegistryImpl.java
PepperCode1: It turns out the logger is not used in the mixin. I wanted to use it to log external handler registrations in a production environment, but the registry impl already does that.

modmuss50 (approved):

modmuss50:
Looks great, many thanks.

PR #4601, opened 2025-04-29
25w18a

modmuss50:
- [x] FabricTagProvider needs fixing, currently has stub methods that return null for the builders
  - [x] `setReplace` & `forceAddTag` were added by fabric, can possibly be iface injected back in. Not a blocker for release.
- [x] AttachmentSerializingImpl no way to get the keys in a read view?

If anyone is bored feel free to help out as I am going to be busy tomorrow. :)


fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/BlockTagGenerator.java
```diff
@@ -226,12 +228,12 @@ private void generateFlowerTags() {
 						Blocks.CLOSED_EYEBLOSSOM
 				);
 
-		getOrCreateTagBuilder(ConventionalBlockTags.TALL_FLOWERS)
+		valueLookupBuilder(ConventionalBlockTags.TALL_FLOWERS)
 				.add(Blocks.SUNFLOWER, Blocks.LILAC, Blocks.PEONY,
 						Blocks.ROSE_BUSH, Blocks.PITCHER_PLANT
-				).addOptional(Identifier.ofVanilla("tall_flowers"));
+				).addOptionalTag(TagKey.of(RegistryKeys.BLOCK, Identifier.ofVanilla("tall_flowers")));
```
modmuss50: This is wrong, previously it was andding an option entry for this, and not an optional tag. I don't see any reference to `tall_flowers` in the game, so might just remove this?

PR #4606, opened 2025-05-02
Attachment changed event

SHsuperCM:
Continuation of #4592, made as a listener(and not a modify) and set in the target itself.

Example usage: 
```java
ClientEntityEvents.ENTITY_LOAD.register((entity, world) -> {
	entity.onAttachedChanged(AttachmentTestMod.SYNCED_ITEM).register((oldValue, newValue) -> {
		MinecraftClient.getInstance().inGameHud.getChatHud().addMessage(Text.of("FYI, the attachment SYNCED_ITEM had its value changed on this entity!"));
	});
});
```

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
@@ -229,7 +229,7 @@ public interface AttachmentTarget {
 	 * @param type the attachment type
 	 * @return event associated with this target and attachment type
 	 */
-	default <A> Event<OnAttachedChanged<A>> onAttachedChanged(AttachmentType<A> type) {
+	default <A> Event<OnAttachedSet<A>> onAttachedSet(AttachmentType<A> type) {
 		throw new UnsupportedOperationException("Implemented via mixin");
 	}
 
@@ -249,14 +249,14 @@ public interface AttachmentTarget {
 	}
 
 	@FunctionalInterface
-	interface OnAttachedChanged<A> {
+	interface OnAttachedSet<A> {
 		/**
-		 * Called after the attachment changes on this target.
+		 * Called after the attachment is set on this target.
 		 *
-		 * @see AttachmentTarget#onAttachedChanged(AttachmentType)
-		 * @param oldValue attachment value on the target prior to it being changed
-		 * @param newValue attachment value on the target after it was changed
+		 * @see AttachmentTarget#onAttachedSet(AttachmentType)
+		 * @param oldValue attachment value on the target prior to it being set
+		 * @param newValue attachment value on the target after it was set
 		 */
-		void onAttachedChanged(@Nullable A oldValue, @Nullable A newValue);
+		void onAttachedSet(@Nullable A oldValue, @Nullable A newValue);
 	}
 }
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
@@ -223,6 +223,9 @@ public interface AttachmentTarget {
 
 	/**
 	 * Provides an {@link Event} that lets the listener know when the target has its {@link AttachmentType} attachment changed.
+	 * <p>
+	 * The event is called after the attachment was changed, be aware of the potential of recursion if you intend on calling {@link #setAttached} on the target.
+	 * </p>
 	 *
 	 * @param type the attachment type
 	 * @return event associated with this target and attachment type
@@ -248,6 +251,13 @@ public interface AttachmentTarget {
 
 	@FunctionalInterface
 	interface OnAttachedChanged<A> {
+		/**
+		 * Called after the attachment changes on this target.
+		 *
+		 * @see AttachmentTarget#onAttachedChanged(AttachmentType)
+		 * @param oldValue attachment value on the target prior to it being changed
+		 * @param newValue attachment value on the target after it was changed
+		 */
 		void onAttachedChanged(A oldValue, A newValue);
 	}
 }
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
@@ -223,7 +223,7 @@ public interface AttachmentTarget {
 
 	/**
 	 * Provides an {@link Event} that lets the listener know when the target has its {@link AttachmentType} attachment changed.
-	 * 
+	 *
 	 * <p>
 	 * The event is called after the attachment was changed, be aware of the potential of recursion if you intend on calling {@link #setAttached} on the target.
 	 * </p>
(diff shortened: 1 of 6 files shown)


cputnam-a11y:
I feel like it may be prudent for the javadoc for the event getter to mention the potentially recursive nature of the event, if it doesn't already

SHsuperCM:
> I feel like it may be prudent for the javadoc for the event getter to mention the potentially recursive nature of the event, if it doesn't already

Agreed, will make sure to do so

SHsuperCM:
Added the comment [in the javadoc](https://github.com/FabricMC/fabric/pull/4606/commits/111f02f6b2469780c158ac17792bf17bfeb48bf5#diff-29eff03a98f1cda3e5d06df5212ed128100c90faf451ca20e9c7ef9cdb0bdf15R226-R228). Let me know if the wording needs changing.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -233,4 +249,16 @@ default <A> A removeAttached(AttachmentType<A> type) {
...
+	@FunctionalInterface
+	interface OnAttachedChanged<A> {
+		/**
+		 * Called after the attachment changes on this target.
+		 *
+		 * @see AttachmentTarget#onAttachedChanged(AttachmentType)
+		 * @param oldValue attachment value on the target prior to it being changed
+		 * @param newValue attachment value on the target after it was changed
+		 */
+		void onAttachedChanged(A oldValue, A newValue);
```
modmuss50: Both new and old should be @ Nullable

fabric-data-attachment-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/attachment/client/gametest/SyncGametest.java
```diff
@@ -128,6 +133,9 @@ public void runTest(ClientGameTestContext context) {
 
 					// check registry objects are synced correctly
 					player.setAttached(AttachmentTestMod.SYNCED_ITEM, Items.APPLE.getDefaultStack());
+
+					// check that the client changes the render distance as requested
+					player.setAttached(AttachmentTestMod.SYNCED_RENDER_DISTANCE, 14);
```
modmuss50: Nit: 14 might be a bit much, especially as its not reset after the test. It will just slow down all the test from here out. I would set this to something more reasonable 6 or 8, and then reset it after the test.

modmuss50 (approved):
Looks great, I think this is much better. Just a few minor changes.

fabric-data-attachment-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/attachment/client/gametest/SyncGametest.java
```diff
@@ -89,6 +89,11 @@ public void runTest(ClientGameTestContext context) {
 				int villagerId;
 			};
 
+			context.runOnClient(client -> {
+				// set client simulation distance before the server sets it
```
Octol1ttle:
```suggestion
				// set client render distance before the server sets it
```

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -233,4 +249,16 @@ default <A> A removeAttached(AttachmentType<A> type) {
 	default <A> A modifyAttached(AttachmentType<A> type, UnaryOperator<A> modifier) {
 		return setAttached(type, modifier.apply(getAttached(type)));
 	}
+
+	@FunctionalInterface
+	interface OnAttachedChanged<A> {
```
Octol1ttle: `onAttachmentSet` would be a significantly better name. "attached" is ambiguous (see the example in the description) and "changed" isn't a guarantee as far as I'm aware (the old value could be the same as the new one)

Octol1ttle (changes requested):

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
```diff
@@ -219,6 +221,20 @@ default <A> A removeAttached(AttachmentType<A> type) {
 		return setAttached(type, null);
 	}
 
+	/**
+	 * Provides an {@link Event} that lets the listener know when the target has its {@link AttachmentType} attachment changed.
+	 *
+	 * <p>
+	 * The event is called after the attachment was changed, be aware of the potential of recursion if you intend on calling {@link #setAttached} on the target.
+	 * </p>
```
Juuxel:
```suggestion
	 * <p>The event is called after the attachment was changed, be aware of the potential of recursion if you intend on calling {@link #setAttached} on the target.
```
Nit: this should follow the usual javadoc format

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
```diff
@@ -47,6 +50,8 @@ abstract class AttachmentTargetsMixin implements AttachmentTargetImpl {
 	private IdentityHashMap<AttachmentType<?>, Object> fabric_dataAttachments = null;
 	@Nullable
 	private IdentityHashMap<AttachmentType<?>, AttachmentChange> fabric_syncedAttachments = null;
+	@Nullable
+	private IdentityHashMap<AttachmentType<?>, Event<OnAttachedChanged<?>>> fabric_attachedChangedListeners = null;
```
Juuxel: Not really related to the PR at hand, but shouldn't all of these fields be `@Unique` instead of just having a prefixed name?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
SHsuperCM: Yep, marked

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
SHsuperCM: Personally in my own mods I would rather do `modid$stuff`, the mod id being in the name of the field/method makes it more traceable imo

fabric-data-attachment-api-v1/src/testmodClient/java/net/fabricmc/fabric/test/attachment/client/gametest/SyncGametest.java
SHsuperCM: Made it 8 for the test and also reset to vanilla default value after.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/api/attachment/v1/AttachmentTarget.java
SHsuperCM: I agree with the `changed` and `set` argument. For attached I dont as the rest of the api uses that term. It does not make sense to have only the event be the odd one out

SHsuperCM: Maybe something more for a future breaking change and then do the whole class?

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
modmuss50: I think we will fix all of these in one go across all mixins. This is wrong.

SHsuperCM: Could something like that be added to checkstyle maybe? idk how that works but it would make sure it's consistent in FAPI at least

PR #4614, opened 2025-05-05
Replace all uses of LocalCapture with @Local (Now for 1.21.6)

cputnam-a11y:
* remove LocalCapture from net.fabricmc.fabric.mixin.gametest.StructureTemplateManagerMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.lifecycle.client.ClientChunkManagerMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.lifecycle.client.WorldChunkMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.lifecycle.MinecraftServerMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.lifecycle.server.WorldChunkMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.entity.event.ServerPlayerEntityMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.interaction.client.MinecraftClientMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.interaction.client.ClientPlayerInteractionManagerMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.event.interaction.ServerPlayerInteractionManagerMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.item.AbstractFurnaceBlockEntityMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.item.BrewingStandBlockEntityMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.client.rendering.EntityModelsMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.screenhandler.ServerPlayerEntityMixin

* remove LocalCapture from net.fabricmc.fabric.mixin.entity.event.LivingEntityMixin

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
+import com.llamalad7.mixinextras.sugar.Local;
 import org.jetbrains.annotations.Nullable;
 import org.spongepowered.asm.mixin.Dynamic;
 import org.spongepowered.asm.mixin.Mixin;
@@ -30,7 +31,6 @@ import org.spongepowered.asm.mixin.injection.ModifyVariable;
 import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
-import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 
 import net.minecraft.block.BedBlock;
 import net.minecraft.block.BlockState;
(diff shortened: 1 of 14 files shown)


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/LivingEntityMixin.java
```diff
@@ -84,8 +84,8 @@ private void beforeDamage(ServerWorld world, DamageSource source, float amount,
 		}
 	}
 
-	@Inject(method = "damage", at = @At("TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
-	private void afterDamage(ServerWorld world, DamageSource source, float amount, CallbackInfoReturnable<Boolean> cir, float dealt, float f, boolean blocked) {
+	@Inject(method = "damage", at = @At("TAIL"))
+	private void afterDamage(ServerWorld world, DamageSource source, float amount, CallbackInfoReturnable<Boolean> cir, @Local(ordinal = 1) float dealt, @Local(ordinal = 0) boolean blocked) {
```
modmuss50, line 88: Question: Is this right? Previously `dealt` was the first local, but now has an ordinal of 1? As im typing this ive just seen there is a float in the args so yes I think this is, can you just confirm?

Earthcomputer: Yes, ordinal is per type, and `dealt` is the second `float` (after the float in the target method parameters).

cputnam-a11y: Yup. The reason it changed is as earth said, the original parameter index with local capture skipped method args automatically. I will note that mcdev didn't want to help me on this one, so I copied the existing fabric inject to a template project, then copied it again, then played with the ordinal until they took the same var. (Took two tries)

modmuss50 (approved):

PR #4618, opened 2025-05-06
Ensure all mixin methods and fields are marked Unique as required.

modmuss50:
Added checkstyle rules for Unique usage, and also prevent using local capture for Injects.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -207,7 +207,7 @@
         <!-- Prevent var for all cases other than new instance creation -->
         <module name="MatchXpath">
             <property name="query" value="//VARIABLE_DEF[./TYPE/IDENT[@text='var'] and not(./ASSIGN/EXPR/LITERAL_NEW)]"/>
-            <message key="matchxpath.match" value="var is not allowed before new instance creation"/>
+            <message key="matchxpath.match" value="var is only allowed before new instance creation"/>
         </module>
 
         <!-- Prevent public records in API packages -->
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.2//EN" "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
 <module name="Checker">
 	<property name="charset" value="UTF-8"/>
 	<property name="fileExtensions" value="java"/>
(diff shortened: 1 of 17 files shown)


checkstyle.xml
```diff
@@ -213,5 +213,66 @@
         <module name="MatchXpath">
             <property name="query" value="//RECORD_DEF[./MODIFIERS/LITERAL_PUBLIC][//PACKAGE_DEF//DOT/IDENT[@text='api']]"/>
         </module>
+
+        <!-- Prevent none @Unique fields in Mixin classes -->
```
haykam821:
```suggestion
        <!-- Prevent non-@Unique fields in Mixin classes -->
```

checkstyle.xml
```diff
@@ -213,5 +213,66 @@
...
+                    and
+                    ../../../CLASS_DEF[
+                        ./MODIFIERS/ANNOTATION/IDENT[@text='Mixin']
+                    ]
+                ]
+            "/>
+        </module>
+
+        <!-- Prevent none @Unique methods in Mixin classes -->
+        <!-- All Mixin and Mixin extras annotations must be listed here -->
```
haykam821:
```suggestion
        <!-- Prevent non-@Unique methods in Mixin classes -->
        <!-- All Mixin and MixinExtras annotations must be listed here -->
```

checkstyle.xml
modmuss50: I didnt learn from last time...

Earthcomputer (approved):
It would be nice to create better error messages for these MatchXpath rules, because sometimes it's hard for new contributors to understand why checkstyle is failing. But that's a problem for later.

modmuss50:
> It would be nice to create better error messages for these MatchXpath rules, because sometimes it's hard for new contributors to understand why checkstyle is failing. But that's a problem for later.

I 100% agree, I think we looked into this last time I added these and couldnt find a way to do it. If anyone knows how to do this please please let me know.

modmuss50:
> It would be nice to create better error messages for these MatchXpath rules, because sometimes it's hard for new contributors to understand why checkstyle is failing. But that's a problem for later.

Added custom messages, I missed the fact that our checkstyle.xml file was using an older spec that didnt support them.

cputnam-a11y:
That is a major improvement.

checkstyle.xml
```diff
@@ -207,11 +207,13 @@
         <!-- Prevent var for all cases other than new instance creation -->
         <module name="MatchXpath">
             <property name="query" value="//VARIABLE_DEF[./TYPE/IDENT[@text='var'] and not(./ASSIGN/EXPR/LITERAL_NEW)]"/>
+            <message key="matchxpath.match" value="var is not allowed before new instance creation"/>
```
Earthcomputer: This rule does the opposite,
```suggestion
            <message key="matchxpath.match" value="var is only allowed before new instance creation"/>
```

Earthcomputer (changes requested):

PR #4620, opened 2025-05-07
Add consume for after screen mouse events

kevinthegreat1:
Allow listeners to check if the click has been consumed and consume the click.

This pr is lacking in comments as I am looking for feedback at this stage. The names should be updated based on suggestions.

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
@@ -176,6 +176,20 @@ public final class ScreenMouseEvents {
 		 * @see org.lwjgl.glfw.GLFW#GLFW_MOUSE_BUTTON_1
 		 */
 		void afterMouseClick(Screen screen, double mouseX, double mouseY, int button);
+
+		default boolean afterMouseClick(Screen screen, double mouseX, double mouseY, int button, boolean consumed) {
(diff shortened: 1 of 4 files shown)


PepperCode1 (commented):
The design looks good. It would make sense to have the default V1 method implementations call the V2 method with `consumed = false`, but I don't know if this is at all useful.

modmuss50:
Whats the use case? Im just familariising myself with this code as its been a while. 

We already have a `AllowMouseClick` that would allow you to consume a click, and stop others from doing so, its not clear to me what you gain from adding this?

kevinthegreat1:
> Whats the use case? Im just familariising myself with this code as its been a while.
> 
> We already have a `AllowMouseClick` that would allow you to consume a click, and stop others from doing so, its not clear to me what you gain from adding this?

The main use case is it allows you to process a click when no other buttons get clicked. For example you want to detect when the user clicks anywhere on the screen, but not when an existing button/element is pressed.

modmuss50:
> The main use case is it allows you to process a click when no other buttons get clicked. For example you want to detect when the user clicks anywhere on the screen, but not when an existing button/element is pressed.

I see, that sounds quite niche imo. Could that be done easily with a mixin, if its only needed for one mod then id lean towards this being out of scope.

modmuss50:
The 25w36a changes necessitates breaking changes to these mouse events, I am going to include the changes in this PR into the 25w36a port so we dont need to juggle 2 copies of this api.

modmuss50:
Superseeded by: https://github.com/FabricMC/fabric/commit/c8811bddbb6e679bc510a3fdf501fbfce94fb824

Thanks for the PR.

PR #4621, opened 2025-05-07
Call all remaining listeners when ALLOW_CHAT event is cancled

### LINKED ISSUE #3133: Allow Message Events depend on mod load order
Events such as `ClientReceiveMessageEvents.ALLOW_GAME` and `ServerMessageEvents.ALLOW_CHAT_MESSAGE` return immediately after a listener returns `false`. The remaining listeners do not get called, and the order of the listeners depends on the mod load order.  

This is an issue if two mods need to read and block the same message. Only one mod will be able to read the message. Since the first mod blocks the message, the event returns, and the second mod never sees the message.  

Proposed change would be to call all listeners regardless of cancellation status.

kevinthegreat1:
Allow listeners to listen to all received messages and optionally cancel. Previously, this required listening to two or three events (ALLOW and CANCELED) and often messages are processed twice. This pr also adds a string parameter, eliminating excessive usage of `text.getString()`. No breaking changes.

This pr is lacking in comments as I am looking for feedback at this stage. The names should be updated based on suggestions.

Fixes #3133.
See previous discussion [here](https://discord.com/channels/507304429255393322/566276937035546624/1241410968173416468).

### DIFF
--- a/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientReceiveMessageEvents.java
+++ b/fabric-message-api-v1/src/client/java/net/fabricmc/fabric/api/client/message/v1/ClientReceiveMessageEvents.java
@@ -36,40 +36,21 @@ public final class ClientReceiveMessageEvents {
 	private ClientReceiveMessageEvents() {
 	}
 
-	public static final Event<ChatV2> CHAT_V2 = EventFactory.createArrayBacked(ChatV2.class, listeners -> (text, message, signedMessage, sender, params, receptionTimestamp) -> {
-		boolean allow = true;
-		for (ChatV2 listener : listeners) {
-			allow &= listener.onReceiveChatMessage(text, message, signedMessage, sender, params, receptionTimestamp);
-		}
-
-		return allow;
-	});
-
-	public static final Event<GameV2> GAME_V2 = EventFactory.createArrayBacked(GameV2.class, listeners -> (text, message, overlay) -> {
-		boolean allow = true;
-		for (GameV2 listener : listeners) {
-			allow &= listener.onReceiveGameMessage(text, message, overlay);
-		}
-
-		return allow;
-	});
-
 	/**
 	 * An event triggered when the client receives a chat message,
 	 * which is any message sent by a player. Mods can use this to block the message.
 	 *
 	 * <p>If a listener returned {@code false}, the message will not be displayed,
-	 * the remaining listeners will not be called (if any), and
+	 * the remaining listeners will be called (if any), and
 	 * {@link #CHAT_CANCELED} will be triggered instead of {@link #CHAT}.
 	 */
 	public static final Event<AllowChat> ALLOW_CHAT = EventFactory.createArrayBacked(AllowChat.class, listeners -> (message, signedMessage, sender, params, receptionTimestamp) -> {
+		boolean allow = true;
 		for (AllowChat listener : listeners) {
-			if (!listener.allowReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp)) {
-				return false;
-			}
+			allow &= listener.allowReceiveChatMessage(message, signedMessage, sender, params, receptionTimestamp);
 		}
 
-		return true;
+		return allow;
 	});
 
 	/**
@@ -78,7 +59,7 @@ public final class ClientReceiveMessageEvents {
 	 * Mods can use this to block the message or toggle overlay.
 	 *
 	 * <p>If a listener returned {@code false}, the message will not be displayed,
-	 * the remaining listeners will not be called (if any), and
+	 * the remaining listeners will be called (if any), and
 	 * {@link #GAME_CANCELED} will be triggered instead of {@link #MODIFY_GAME}.
 	 *
 	 * <p>Overlay is whether the message will be displayed in the action bar.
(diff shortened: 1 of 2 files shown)


modmuss50:
Again, this is a similar pattern to https://github.com/FabricMC/fabric/pull/4620 . Im not sure there is actually anything new here.

If mods are encoraged to listen to all by default it undermines the meaning of a message being cancled. Again whats the specific use case you are trying to solve? Remeber you can use event sorting to try and be earlier than most.

kevinthegreat1:
> Again, this is a similar pattern to #4620 . Im not sure there is actually anything new here.
> 
> If mods are encoraged to listen to all by default it undermines the meaning of a message being cancled. Again whats the specific use case you are trying to solve? Remeber you can use event sorting to try and be earlier than most.

The main problem is with mod compatibility. Specifically on hypixel, many mods may do the same thing, and they all want to process a message then hide it from the player. In the current approach, you can listen to allow, but if another mod cancels the message first, your handler will not get called. So you listen to both allow and canceled, but then if said mod is not present, then you have to process the message twice.

The problem gets even worse if you want to modify. If you just listen to modify, another mod may modify the message before you do, so you lose information. Another mod may also cancel the message, so you need to listen to allow, cancel, and modify to ensure you get the message from the server.

Using event sorting won’t really work here as if I force my mod to be earlier, then other mods get broken, and they try to make their listeners be earlier. I think I expressed some of these issues in the discord discussion I linked. Now, my mod and many mods have to use custom injection points before fapi to ensure we receive all messages. https://github.com/SkyblockerMod/Skyblocker/blob/e71f08a418eba1c464d054c5d634d580a25eac8a/src/main/java/de/hysky/skyblocker/events/ChatEvents.java

MisterCheezeCake:
I can't really add much to what Kevin said above, I completely agree.

For our use case, messages are transmitting *information*, information that is often critical to our mod's function. Just this week, I discovered an issue in my mod where significant functionality would break due to another mod canceling Hypixel's Profile ID messages [which has led me to begin the process of creating custom events like Kevin mentioned above](https://github.com/MisterCheezeCake/SkyblockTweaks/issues/10).

With pure listener chat events added, mod developers will be able to listen to the event that best suits their use case:
- If and only if they need the information the server is giving them, unfiltered and unedited, they should listen to the first event, which will give them what they need, straight from the server.
- If they want to cancel a message to the user, they should use `ALLOW`
- If they want to modify what the user sees, they should use `MODIFY`
- If they want to see *what the user sees*, they should use the current `GAME`/`CHAT` events

What I think is happening is a confounding of two distinct layers: Mod Backend and Player Experience. Right now, Fabric's chat events are ill equipped to serve the Mod Backend layer, as they operate under the assumption that what the player sees is the be all end all, which it is not for the use case of the Hypixel modding community, where the Mod Backend needs lots of information from the server through chat (Action Bar Stats, Profile ID, etc.). Like Kevin said, there are already workarounds in FAPI, but they are just that: workarounds, and they should not be needed when a solution exists and would cause no harm.

As evinced by the continued requests for this, people want it and people will use it one way or another. We should do it right in Fabric API instead of leaving it to each mod to do themselves.

ThatGravyBoat:
I whole heartedly agree with the sentiment of this PR, the current chat events are very ill-equipped to handle anything other than simple operations of exclusive chat processing and event breaks sometimes then. The current system requires duplication of processing of events very easily due to the split nature meaning 1 mod has to process the same message multiple times, and there are workarounds but all the workarounds have cons of missing events, having to bypass fabrics API causing incompatibility, keeping a cache of all messages process (this causes an issue with duplicate messages), having to force yourself in front of other mods using phases which is error prone due to other mods also doing that.

The idea of the PR would solve almost all issues with the current evolution of chat events.

lineargraph:
While I think having a process chat event (which is what it is called in my codebase) that receives all messages, cancelled or not, is useful, i disagree that there are no ways of solving this. `ALLOW_CHAT` and `CHAT_CANCELED` (which also shouldnt that be cancelled with two Ls? not entirely sure) are called immediately in sequence, in the same thread. Keeping a variable of the last message that was handled in `ALLOW_CHAT` and checking against that in `CHAT_CANCELED` is very low overhead. If cross-thread concerns exist, a thread local should work fine. The only issue i could see with code like that is potential third party mod calling those events out of order, not calling the `CHAT_CANCELED` event, calling the events with the exact same instance, or throwing exceptions, which are cases i am willing to discount, considering their irregular nature.

```kt
/* This could potentially be a ThreadLocal<Text> instead. */
private var lastReceivedMessage: Text? = null

ClientReceiveMessageEvents.ALLOW_CHAT.register(ClientReceiveMessageEvents.AllowChat { essage, signedMessage, sender, params, receptionTimestamp ->
	lastReceivedMessage = message
	ProcessChatEvent.publish(ProcessChatEvent(message, wasExternallyCancelled = false))
	true
})
ClientReceiveMessageEvents.CHAT_CANCELED.register(ClientReceiveMessageEvents.ChatCanceled { message, signedMessage, sender, params, receptionTimestamp ->
	if (lastReceivedMessage !== message) {
		ProcessChatEvent.publish(ProcessChatEvent(message, wasExternallyCancelled = true))
		lastReceivedMessage = null
	}
})
``` 


I want to emphasize that despite workarounds existing, having an officially sanctioned `PROCESS_CHAT` or similarly spirited event would be great to have. This kind of compat code is non obvious and a lot easier to implement in fabric itself.

modmuss50:
I dont know where all the love has suddenty come for this, its good to see though. 👍  This whole use case is something I didnt even know existed, thanks for rasing it. The PR needs some work though, please get the build passing at the very least, along with docs and tests to prove that its working as expected with a simulated use case.

Calling it `CHAT_V2` is clearly wrong, it almost just needs to be a slight behavior change to `ALLOW_CHAT` 🤔 Would it be a breaking change if the existing allow event was changed to contiune to call other listeners even if it was cancled? I think it would be, and probaly not something we want to do. Although im happy to discuss it. The change in logic seems fine here, we just need a better name for the event.

Just thinking out loud: Another idea would be to add a default method to the existing event that always gets called, that way we dont need a new event. It would make it slightly harder for mods to impliment though.

modmuss50:
As discussed on discord we are happy to change the existing event.

kevinthegreat1:
I've changed this pr to modify the allow events. This means that we won't get the string parameter, but that's not a big deal.

modmuss50 (approved):

PR #4625, opened 2025-05-08
fix(data-attachment): don't call ".remove" on null dataAttachments map

### LINKED ISSUE #4624: [1.21.5] Removing an attachment from an entity throws a `NullPointerException` under some circumstances as of v0.123.0+1.21.5
I'm trying to remove an attachment from a living entity when it dies. This was working without issues before v0.123.0+1.21.5 of Fabric Api. Now an exception is thrown:
```
java.lang.NullPointerException: Cannot invoke "java.util.IdentityHashMap.remove(Object)" because "this.fabric_dataAttachments" is null
```

Here is the original issue reported to me: https://github.com/Fuzss/bettertridents/issues/23

I tested this specifically with v0.122.0+1.21.5 and below and it is indeed working fine there.

Here is my code where I remove the attachment (although it's heavily diluted by multi-loader shenanigans): https://github.com/Fuzss/bettertridents/blob/4e7426c5b0a9b0925466a29dc38be29ceb9395db/1.21.5/Common/src/main/java/fuzs/bettertridents/handler/TridentAttachmentHandler.java#L23

Octol1ttle:
Closes #4624

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
@@ -75,11 +75,7 @@ abstract class AttachmentTargetsMixin implements AttachmentTargetImpl {
 		T oldValue;
 
 		if (value == null) {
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Fine, a test for this should really be added, im in a rush now so will just get this released as-is.

PR #4626, opened 2025-05-08
[1.21.6] fix(data-attachment): don't call ".remove" on null dataAttachments map

Octol1ttle:
See #4624 
See #4625


modmuss50:
As its the same fix for both versions, it will be easier for me to cherry pick this from the other PR. thanks.

PR #4630, opened 2025-05-11
Data Attachments: Don't mark changed & sync if old value is the same as new value

Octol1ttle:
Gives more flexibility to API consumers as they won't have to add this check themselves. In cases where attachments are constantly updated with the same value, should reduce disk time usage and network bandwidth usage


modmuss50 (approved):
Looks good.

Im happy with this, technically it could be a breaking change if someone was depending on receiving the changed events even if the value didnt change. But I doubt anyone is and it would be bad design anyway.

apple502j:
Could there be a use case for mutable data? I suppose people should properly implement equals/not use mutable data as attachments. Also I'm wondering if these are order-sensitive.

Octol1ttle:
> Could there be a use case for mutable data? I suppose people should properly implement equals/not use mutable data as attachments. Also I'm wondering if these are order-sensitive.

That's a good usecase. Even if `equals` is correctly implemented, it won't be of use as the consumer would be accessing the same pointer anyway. Mutable data isn't a rare usecase, although using `setAttached` to reset it sounds like a hack. Maybe we should provide an intended manual way to refresh an attachment (if there's already one lmk)

modmuss50:
Data attachments should never be mutable, its a recipe for disaster for many reasons.

PR #4632, opened 2025-05-13
Update mod to Minecraft snapshot 25w20a

haykam821:
This pull request ports Fabric API to [Minecraft snapshot 25w20a](https://www.minecraft.net/en-us/article/minecraft-snapshot-25w20a).

Item tooltips seemingly do not show when running with Fabric API's test mods. However, they do show when using only Fabric API. Is this behavior expected?


modmuss50:
> Item tooltips seemingly do not show when running with Fabric API's test mods. However, they do show when using only Fabric API. Is this behavior expected?

No not at all, I can double check in 1.21.5 to see if its the case there as well.

modmuss50:
Seems to be broken in the previous snapshot, im happy to merge this and fix that separately. Please dont let me forget about it 😆

modmuss50 (approved):

PR #4639, opened 2025-05-16
Port Hud Api

kevinthegreat1:
Adds a new HUD API similar in terms of capabilities to the previous on in 1.21.4 and 1.21.5.

`HudElementRegistry` provides `addFirst`, `addLast`, `attachElementBefore`, `attachElementAfter`, `removeElement`, and `replaceElement`. Call these methods to register an element or modify an existing element. See Javadocs for more info.

This differs from the previous API in that there is no longer an event. You can register an element at any time (on the main thread). Also, the main HUD layers have been split into individual bars and components, allowing for more control.


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudLayerRegistrationCallback.java
```diff
@@ -0,0 +1,62 @@
...
+ * HudLayerRegistrationCallback.EVENT.register(layeredDrawer -> layeredDrawer.attachLayerAfter(IdentifiedLayer.MISC_OVERLAYS, Identifier.of("example", "example_layer_after_misc_overlays"), (context, tickCounter) -> {
+ *     // Your rendering code here
+ * }));
+ * // @end @end @end @end @end @end @end @end @end @end
+ * }
+ *
+ * @see LayeredDrawerWrapper
+ */
+public interface HudLayerRegistrationCallback {
+	Event<HudLayerRegistrationCallback> EVENT = EventFactory.createArrayBacked(HudLayerRegistrationCallback.class, callbacks -> layeredDrawer -> {
```
modmuss50: Im not sure the design of this API is correct, we shouldnt try to mold what we had before into the new version. I dont see any reason there needs to be an event for this now. As I said on discord I think it should have an API similar to ComponentTooltipAppenderRegistry. There is no concept of layers in vanilla anymore either, maybe element would be better.

Here is what I have in my head, just a rough outline of what the API could look like:

HudElementRegistration.addFirst(HudElement)
HudElementRegistration.addLast(HudElement)
HudElementRegistration.addBefore(Id, HudElement)
HudElementRegistration.addAfter(Id, HudElement)

Maybe this is too basic compared to what we had before?, I'd be happy to discuss this some more.

kevinthegreat1: I see, that's what you meant by an API similar to `ComponentTooltipAppenderRegistry`. I didn't think about the fact that we don't need an event anymore.

The main question is whether to add `remove` and `replace` features. I currently don't have a use case, but maybe someone could point out some use cases?

I do worry that mojang is going to add back some sort of layer system, and I tried to asked in the vv discord, but I will update this implementation.

modmuss50:
This is looking much better thanks, I wonder if we can do something better to not require the need to use IdentifiedElement.of 🤔 

Im only taking a quick look at this now, but I do want to spend a bit more time going though the impl. Thanks again sorry for being slow to review.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
```diff
@@ -0,0 +1,215 @@
...
+	 * An element that wraps a vanilla element using a list, allowing for users to attach elements before or after it, replace it, or remove it.
+	 */
+	@VisibleForTesting
+	public record VanillaElement(Identifier id, List<HudElement> elements) {
+		public VanillaElement(Identifier id) {
+			this(id, new ArrayList<>());
+			elements().add(IdentifiedElement.of(id, (context, tickCounter) -> { }));
+		}
+
+		public void render(InGameHud instance, DrawContext context, RenderTickCounter tickCounter, Operation<Void> renderVanilla) {
```
isXander: Using a Mixin Extras `Operation` feels a bit wrong outside of mixins.

isXander:
I'm not sure I like the way this API is made. I feel it is overly complex.

We already have to `WrapOperation` on every single HUD render call, so why not just use `Event` a bit like the Screen API does for events on specific screens?

```java
HudRenderEvents.after(IdentifiedElement.SCOREBOARD).register((drawContext, renderTickCounter) -> {
  // draw my element
})
```
This also allows for adding more injection points like `IdentifiedElement.HIDABLE_GUI` that corresponds to all the HUD that's within the conditions.

This of course comes with the downside of not playing nicely with modded HUD elements. Though, you could probably make a small wrapper that looks like this: 
```java
public static void registerAfter(Identifier afterElement, Identifier thisLayer, LayerRenderer renderer) {
    InternalHudRenderEvents.registerLayer(thisLayer); // creates the Event
    InternalHudRenderEvents.after(afterElement).register((drawContext, renderTickCounter) -> {
        InternalHudRenderEvents.before(thisLayer).invoker().render(drawContext, renderTickCounter);
        renderer.render(drawContext, renderTickCounter);
        InternalHudRenderEvents.after(thisLayer).invoker().render(drawContext, renderTickCounter);
    });
}
```
You could also wrap all this registration within the existing `HudLayerRegistrationCallback`, where after invocation of the event you can check if any layers have no invokers, suggesting that the mod that registered it registered an invalid layer that no mod provides.

I can take a crack at a v2 api as well and modmuss can decide between these different implementations if that's what's needed?

isXander:
I took the time to actually implement my proposal. I don't have time to turn this into a proper PR right now, but I made a Gist.

https://gist.github.com/isXander/c969700ee7206dc068f69cbe91dd155f

modmuss50:
I ended up with a much much simpler API and impl: I wonder what you think: https://github.com/kevinthegreat1/fabric/compare/hud-api-again...modmuss50:hud-api-again-2?expand=1

It allows being first or last, or before/after any vanilla layer. It does loose the abbility to sort between mods but it makes things so much easier, im not sure how much of a loss this really is?

modmuss50 (approved):
Im happy with this. Its a much simpler implmentation that doesnt try to bring back the combersome layering system. It should cover the vast majority of use cases. I get its not as powerful as the old API was but it gives us a simple base to work from if addional features are required later on.

Ill put this into last call, if anyone has any concerns please make sure to let them be known.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/HudRenderCallback.java
```diff
@@ -19,11 +19,12 @@
 import net.minecraft.client.gui.DrawContext;
 import net.minecraft.client.render.RenderTickCounter;
 
+import net.fabricmc.fabric.api.client.rendering.v1.hud.HudElementRegistry;
 import net.fabricmc.fabric.api.event.Event;
 import net.fabricmc.fabric.api.event.EventFactory;
 
 /**
- * @deprecated Use {@link HudLayerRegistrationCallback} instead. For common use cases, see {@link LayeredDrawerWrapper}.
+ * @deprecated Use {@link HudElementRegistry} instead. For common use cases, see {@link HudElementRegistry}.
```
Juuxel:
```suggestion
 * @deprecated Use {@link HudElementRegistry} instead.
```
No need to link twice to the same class

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudElement.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.rendering.v1.hud;
+
+import net.minecraft.client.gui.DrawContext;
+import net.minecraft.client.render.RenderTickCounter;
+
+public interface HudElement {
```
Juuxel, line 28: This should have javadoc (both the method and especially the interface)

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/VanillaHudElements.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * <p>The identifiers in this interface are the vanilla hud elements in the order they are drawn in.
+ * The first element is drawn first, which means it is at the bottom.
+ * All vanilla elements except {@link #SLEEP} are in sub drawers and have a render condition attached ({@link net.minecraft.client.option.GameOptions#hudHidden}).
+ * Operations relative to any element will generally inherit that element's render condition.
+ * There is currently no mechanism to change the render condition of an element.
+ *
+ * <p>For common use cases and more details on how this API deals with render condition, see {@link HudElementRegistry}.
+ */
+public interface VanillaHudElements {
```
Juuxel: Is there any reason for this to be an interface instead of a final class?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudMixin.java
```diff
@@ -26,11 +28,89 @@
 import net.minecraft.client.render.RenderTickCounter;
 
 import net.fabricmc.fabric.api.client.rendering.v1.HudRenderCallback;
+import net.fabricmc.fabric.api.client.rendering.v1.hud.VanillaHudElements;
+import net.fabricmc.fabric.impl.client.rendering.hud.HudElementRegistryImpl;
 
 @Mixin(InGameHud.class)
-public class InGameHudMixin {
+public abstract class InGameHudMixin {
```
Juuxel:
```suggestion
abstract class InGameHudMixin {
```
This might as well be fixed to be package-private since the modifiers are changing anyway.

Juuxel (changes requested):

Fuzss:
Hey, thanks for bring this API back! 

Personally I think it's very unfortunate to see the layer system go, it was very nice and convenient to work with for me on 1.21.5. Specifically the abilities to replace and wrap other layers, and also being able to affect the layers from other mods. [NeoForge](https://github.com/neoforged/NeoForge/blob/port/1.21.6/src/client/java/net/neoforged/neoforge/client/gui/GuiLayerManager.java) also kept the layer concept around, so why wouldn't Fabric? 😊

Besides that there are two things that stand out to me:

1. There is no way to prevent a vanilla `HudElement` from rendering anymore. 
This was useful for replacing vanilla layers in the old implementation (e.g. [Stylish Effects](https://modrinth.com/mod/stylish-effects)). This could easily be brought back by splitting `HudElement` into `Before` and `After` with `Before` returning a `boolean`, just like basically all other before & after events in Fabric do.

2. Split `VanillaHudElements#HOTBAR_AND_BARS` into more elements. 
Since the API now simply wraps vanilla method calls, this can be expanded beyond the calls in `InGameHud::render`. Having more fine-grained control for separate bar elements like health, armor, food, selected item name, experience bar, etc. would be very useful, and there certainly are a lot of mods that affect rendering of specifically those gui elements. 
My suggestion basically is to mirror what [NeoForge](https://github.com/neoforged/NeoForge/blob/440802a169a54b6eaf33665112883088cec6b72b/src/client/java/net/neoforged/neoforge/client/gui/VanillaGuiLayers.java#L20-L31) has here, since that's been pretty ideal in the past.
On 1.21.5 I already went ahead and added these more fine tuned layers myself, but would be great to have them out of the box for 1.21.6.

I have about 5-10 mods using the gui layer API on 1.21.5 (with my own extensions for more layers), from the looks of it almost none will be able to use this reworked implementation, mostly due to the missing ability to cancel / wrap vanilla layers. 😬

Even though the whole layer concept was very new to vanilla & Fabric, it has been proven in Forge / NeoForge for many years, and from my personal experience across a wide range of projects is the way to go for making the `InGameHud` more extensible & modding friendly.

To sum it up the way Kevin had this PR initially was exactly what I need for my own mods.

cputnam-a11y:
While is is possible to replicate the neo system with and without overwriting, the overall consensus has been against bringing the layer system back. I believe modmuss asked for reasons that overriding a vanilla layer is needed.
Screenshot_20250602-110551

cputnam-a11y:
I personally think the layer system would be very useful, and that its possible to do it in a compatible manner, but I'm not ready to make a pr to back up my talk

modmuss50:
Hey, thanks for taking a look at this. Its good to hear your use cases, lets make sure they are covered before this PR is merged.

I really dont think we should jump to bringing the old layer system back, it was provided by vanilla and not fabric. From a quick look it seems [neoforge](https://github.com/neoforged/NeoForge/blob/port/1.21.6/patches/net/minecraft/client/gui/Gui.java.patch#L39-L96) has basically rewritten all of this code back to how it was. This is not really practical for us and adds a major maintainance overhead. If we can we should look to adapt to what vanilla has, and not replace it.

Ill take another look at this, as like you said it should be possible to support these use cases without restoring back to the old system. There is a balance between complexity and functionality that needs to be made, clearly the current PR isnt quite there yet.

Fuzss:
I'll leave some of my mods here I can think off right now that use the Gui Layer API to have some more concrete examples of what the reworked implementation should ideally cover. I'll include examples that are both supported and no longer possible.

✅ [Pick Up Notifier](https://modrinth.com/mod/pick-up-notifier): Simply adds a new gui layer ([Code](https://github.com/Fuzss/pickupnotifier/blob/1b53690b8c3e404cc99a1e886d50445e5f492575/1.21.5/Common/src/main/java/fuzs/pickupnotifier/client/PickUpNotifierClient.java#L19)).
✅ [Sword Blocking Mechanics](https://modrinth.com/mod/sword-blocking-mechanics): Prevents the attack indicator from rendering while rendering it's own blocking indicator in the same spot ([Code](https://github.com/Fuzss/swordblockingmechanics/blob/main/1.21.5/Common/src/main/java/fuzs/swordblockingmechanics/client/handler/AttackIndicatorInGuiHandler.java)).
❌ [Stylish Effects](https://modrinth.com/mod/stylish-effects): Replaces the vanilla gui status effect display ([Code](https://github.com/Fuzss/stylisheffects/blob/2f3ced97062bc7497f8d9a223b9b0e37948ee4eb/1.21.5/Common/src/main/java/fuzs/stylisheffects/client/StylishEffectsClient.java#L32)).
❌ [Held Item Tooltips](https://modrinth.com/mod/held-item-tooltips): Replaces the vanilla selected item name rendering ([Code](https://github.com/Fuzss/helditemtooltips/blob/54bf001f1bed43405d82dd7e5f536b6df13dc6c5/1.21.5/Common/src/main/java/fuzs/helditemtooltips/client/HeldItemTooltipsClient.java#L21)).
❌ [Leave My Bars Alone](https://modrinth.com/mod/leave-my-bars-alone): Wraps some vanilla hotbar elements ([Code](https://github.com/Fuzss/leavemybarsalone/blob/main/1.21.5/Common/src/main/java/fuzs/leavemybarsalone/client/LeaveMyBarsAloneClient.java)).
❌ [Overflowing Bars](https://modrinth.com/mod/overflowing-bars): Replaces vanilla hotbar elements ([Code](https://github.com/Fuzss/overflowingbars/blob/main/1.21.5/Common/src/main/java/fuzs/overflowingbars/client/OverflowingBarsClient.java)). Also adds a toughness bar (still possible).
✅ Hovering Hotbar: Wraps vanilla layer with pose stack translations, still possible, but less convenient and more prone to triggering popPose / pushPose issues ([Code](https://github.com/Fuzss/hoveringhotbar/blob/87930ae71162d2c4e038b836e05e839f56c7de00/1.21.5/Common/src/main/java/fuzs/hoveringhotbar/client/HoveringHotbarClient.java#L59)).

Some things to note:

My mods are multi-loader, so the Fabric Gui Layer API is only used through an abstraction. The design of that is very similar to Fabric's API, so it should be easy to follow along regardless. Here is the implementation for that:
- https://github.com/Fuzss/puzzleslib/blob/main/1.21.5/Common/src/main/java/fuzs/puzzleslib/api/client/core/v1/context/GuiLayersContext.java
- https://github.com/Fuzss/puzzleslib/blob/main/1.21.5/Fabric/src/main/java/fuzs/puzzleslib/fabric/impl/client/core/context/GuiLayersContextFabricImpl.java
- https://github.com/Fuzss/puzzleslib/blob/main/1.21.5/Fabric/src/main/java/fuzs/puzzleslib/fabric/mixin/client/GuiFabricMixin.java

Also not all my mods are updated to use the Gui Layer API yet, my system from before 1.21.5 is still in place for those which is functionally equivalent. However, I was planning on migrating everything for the 1.21.6 port.

And finally as mentioned in my previous comment, I've extended the old Gui Layer API with some additional vanilla layers specifically for finer control over `IdentifiedLayer#HOTBAR_AND_BARS`, so that's widely been used in my mods. but definitely something that should be part of the Fabric API, especially now with the new implementation.

Fuzss:
Regarding the layer system, I fully understand and agree that maintaining such invasive patches as NeoForge applies is not feasible for Fabric.

However, I do not believe they are necessary to achieve the API surface the old implementation from Fabric provided. Actually those `@WrapOperation` style injections this PR offers in its current form should be precisely enough.

Here is what the API + implementation would look like:
- Keep the old API as it was, registering layers relative to vanilla, replacing layers, etc.
- When a vanilla "layer" is rendered from `InGameHud`, look up modded layers that render before / after, handle those via the same system vanilla layers are, so rendering again other layers before / after is also supported.
- Keep support for replacing layers, basically like [so](https://github.com/Fuzss/puzzleslib/blob/c2555aed5f797dbf9794a67b70f7c53d08fcf4b5/1.21.5/Fabric/src/main/java/fuzs/puzzleslib/fabric/impl/client/core/context/GuiLayersContextFabricImpl.java#L125). Removing of course works by replacing with a no-op layer.

This basically is what this PR has right now, just on the API surface the before / after events are returned to using layers.

So vanilla layer data, specifically any attached layers are looked up "on-demand" now, instead of being able to iterate over a flattened list of all layers like before in 1.21.5 with vanilla's `LayerDraw` manager.

modmuss50:
I ended up reverting my change to remove the layers, it was clearly not what we want I couldnt get it to work nicely to do what you were asking. Its now reverted back to similar to what @kevinthegreat1 originaly suggested, I just moved some things around and made the layer idea fully an implementation detail.

Let me know what you think @Fuzss, I also went a head and split HOTBAR_AND_BARS into more layers.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
modmuss50: Changed this to pass a Runnable, mostly becuase some of hotbar renderes take diffrent args, and partly becuase it did feel odd.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+		};
+
+		public RootLayer(Identifier id) {
+			this(id, new ArrayList<>());
+			layers().add(HudLayer.of(id, VANILLA));
+		}
+
+		public void render(DrawContext context, RenderTickCounter tickCounter, Runnable renderVanilla) {
+			for (HudLayer layer : layers) {
+				if (layer.id().equals(id)) {
```
Fuzss: This feels wrong. From my understanding this will even render the vanilla element when it was replaced, as it only checks the identifier which the replacement retains.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+		});
+
+		if (!didChange) {
+			throw new IllegalArgumentException("Layer with identifier " + identifier + " not found");
+		}
+	}
+
+	public static void replaceElement(Identifier identifier, Function<HudElement, HudElement> replacer) {
+		boolean didChange = findLayer(identifier, (l, iterator) -> {
+			iterator.set(HudLayer.of(identifier, replacer.apply(l.element())));
```
Fuzss: The element passed into the function is most of the time gonna be `RootLayer#VANILLA` which is probably not intended, as it's not useful. Or is it?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+		});
+
+		if (!didChange) {
+			throw new IllegalArgumentException("Layer with identifier " + afterThis + " not found");
+		}
+	}
+
+	public static void removeElement(Identifier identifier) {
+		boolean didChange = findLayer(identifier, (l, iterator) -> {
+			iterator.remove();
```
Fuzss: Wouldn't it be preferable to replace the to be removed element with a no-op element instead? 

That way other mods can still attach their elements to it. Otherwise this may become a problem as there is no way of knowing if an element has been removed, and registering against a non-existing element will trigger an `IllegalArgumentException`.

That's something the old implementation also had which I didn't particularly like.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+import net.fabricmc.fabric.api.client.rendering.v1.hud.VanillaHudElements;
+
+public class HudElementRegistryImpl {
+	@VisibleForTesting
+	public static final List<Identifier> VANILLA_ELEMENT_IDS = List.of(
+			VanillaHudElements.MISC_OVERLAYS,
+			VanillaHudElements.CROSSHAIR,
+			VanillaHudElements.SPECTATOR_MENU,
+			VanillaHudElements.HOTBAR,
+			VanillaHudElements.STATUS_BARS,
```
Fuzss: I'd really prefer if this could be split into the individual bars, see [here](https://github.com/neoforged/NeoForge/blob/440802a169a54b6eaf33665112883088cec6b72b/src/client/java/net/neoforged/neoforge/client/gui/VanillaGuiLayers.java#L22-L26) (minus vehicle of course).
As with the others all this requires is [wrapping the render calls](https://github.com/Fuzss/puzzleslib/blob/c2555aed5f797dbf9794a67b70f7c53d08fcf4b5/1.21.5/Fabric/src/main/java/fuzs/puzzleslib/fabric/mixin/client/GuiFabricMixin.java#L206-L253) in the corresponding method, no more.

It's definitely useful for me with Overflowing Bars, and other mods like Apple Skin, Detail Armor Bar to name a few will probably appreciate it as well.

Fuzss (commented):
Thank you for acting this quickly, that is pretty much what I had in mind. ❤️
Here are some minor quirks I found.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: I think you’re right. We would have to move back to redirect or ignore the wrapped operations if we provide replace ability in our api. I’m surprised that our unit tests didn’t catch this error though.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: I believe this is intended, as it allows you to wrap vanilla layers.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: Actually, that would be useful for me as well. I currently use mixins and didn’t think we could have an api to cover that.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: Actually, we don't need to move back to redirects at all.

kevinthegreat1:
I've fixed the implementation. The only thing left is splitting the health bar, armor bar, and etc.

I also want to ask if there's a reason for renaming vanilla to root in the impl? I think vanilla would be a better name but this isn't a big deal.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
Fuzss: Actually it doesn't, that's what I meant. The passed element is `RootLayer#VANILLA` which throws an exception when calling `HudElement::render`. The actual vanilla element is never passed. While replacing does work, this makes wrapping vanilla elements impossible. So an implementation like I'm using for my [Leave My Bars Alone](https://github.com/Fuzss/leavemybarsalone/blob/main/1.21.5/Common/src/main/java/fuzs/leavemybarsalone/client/LeaveMyBarsAloneClient.java) mod in 1.21.5 does not work with this.

Fuzss:
⬆️ https://github.com/FabricMC/fabric/pull/4639#discussion_r2122959471
Not sure if this show up for everyone since it has already been marked as resolved.

I think the fix for this is pretty straight forward, I've solved the exact issue in my implementation for Puzzles Lib linked above. `HudLayer::element` should take a `HudElement` argument, to basically act as a `UnaryOperator`. Vanilla layers should return the passed `HudElement`, while other custom layers ignore it and just return their set element. Like this `HudElements` can be properly wrapped for replacing. `RootLayer::render` then should take in the vanilla element as `HudElement`, not `Runnable`, so it can be passed through.

I can make a PR to Kevin's repository if unclear, since that's the only way I think I can contribute here (?).

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: It looks like we have two ways.

One is to capture the vanilla render call at initialization, which is a bit annoying. I think we have to use an initialization event for this method.

The other is to apply replacer functions at render time, every frame, which is more similar to what neoforge does, I believe.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
Crendgrim: I would also be very happy about that (for AutoHUD).

kevinthegreat1:
@Fuzss please pr to my repo. You are proposing that we apply replacers at render time?

Fuzss:
Yes, I see no downside with that approach. 
Will do.

Fuzss:
https://github.com/kevinthegreat1/fabric/pull/1

cassiancc (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudElementRegistryImpl.java
kevinthegreat1: Done in 39073f9.

Fuzss (approved):

modmuss50 (approved):
Many thanks for all the help with this, sorry for taking it off track for a bit.

Ill give this a very short last call so as I think it will be better to allow people to get hands on experiance with it. If there is anything that needs changing we can still do that before the final release.

cassiancc (approved):

PR #4640, opened 2025-05-16
Don't erase data attachments if missing in NBT

### LINKED ISSUE #4638: [data-attachment-api-v1] Attached data can get clobbered client-side for some block entities
Unless I'm misunderstanding something about how we're supposed to integrate with the data-attachment-api-v1 for block entities, it seems like using it with (almost) any block entity that has other update-able state won't work. The API mixins to BlockEntity at [`createNbt`](https://github.com/FabricMC/fabric/blob/1.21.5/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/BlockEntityMixin.java#L70), but a lot of the block entities (e.g. beacons and signs) in vanilla don't actually call that when generating update packets (i.e. through `world.updateListeners`).

In my mod I've attached additional data to signs, but any time `world.updateListeners` is called for any reason (changed text, applied dye or wax, etc.), my attached data gets wiped client-side when the sign generates its update packet NBT through `createComponentlessNbt` and sends it to all the clients without any attached data.

From what I can tell all block entities go through `BlockEntityUpdateS2CPacket.create(this)` at least before diverging in paths, which uses `BlockEntity::toInitialChunkDataNbt` to generate the NBT. So to me the best way to hook into it would be to mixin to `BlockEntityUpdateS2CPacket.create` and wrap the `BiFunction` parameter (`BlockEntity::toInitialChunkDataNbt`) so we can modify the NBT it spits out, right?

Just to test I implemented this additional mixin in my own mod and everything seems to be working, but I don't know if there are any potential side effects I've overlooked:
https://github.com/Roundaround/mc-fabric-item-signs/blob/1.21.5/src/main/java/me/roundaround/itemsigns/mixin/BlockEntityUpdateS2CPacketMixin.java

EDIT: Made a repro mod:
<details>
<summary>Here's a basic repro mod of the issue</summary>

https://github.com/Roundaround/fabric-example-mod/tree/issue-4638

The important bits are:

- The attachment, with persistence and syncing enabled
  - https://github.com/Roundaround/fabric-example-mod/blob/3f0dadba5a47f9dc8e5fbe1c1919b6833068cb1c/src/main/java/com/example/ExampleAttachment.java
- Setting the attachment
  - https://github.com/Roundaround/fabric-example-mod/blob/3f0dadba5a47f9dc8e5fbe1c1919b6833068cb1c/src/main/java/com/example/mixin/AbstractSignBlockMixin.java
- Basic mixin that just renders the current attachment value in a typical nametag thing over the sign. Important aspect is that it is reading the attachment client-side
  - https://github.com/Roundaround/fabric-example-mod/blob/3f0dadba5a47f9dc8e5fbe1c1919b6833068cb1c/src/client/java/com/example/mixin/client/AbstractSignBlockEntityRendererMixin.java

With this setup, if you place a sign, you should see a red X above it (because no attachment has been set yet).

If you sneak-click the sign, the attachment will be created/set with a value of 1, so the red X should change to a green 1. Continuing to sneak-click it will continue incrementing.

If you then modify the sign's state in some other way (i.e. edit the text, apply dye or wax, etc.), the green number will be reset back to a red X, as the client "forgets" the attached data. This however is client-side only, meaning if you sneak-click the sign again, it will continue incrementing the value and the tag will return back to a green number like expected.
</details>

Roundaround:
As far as I can tell, this code path is only _actually_ used for data attachments on initial load, since any changes are synced with the custom network packets. Therefore, I think it's safe to consider missing attachment data a no-op rather than an erasure of all attachments.

Fixes #4638

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/mixin/attachment/AttachmentTargetsMixin.java
@@ -124,7 +124,17 @@ abstract class AttachmentTargetsMixin implements AttachmentTargetImpl {
 	public void fabric_readAttachmentsFromNbt(NbtCompound nbt, RegistryWrapper.WrapperLookup wrapperLookup) {
 		// Note on player targets: no syncing can happen here as the networkHandler is still null
 		// Instead it is done on player join (see AttachmentSync)
-		this.fabric_dataAttachments = AttachmentSerializingImpl.deserializeAttachmentData(nbt, wrapperLookup);
+		IdentityHashMap<AttachmentType<?>, Object> fromNbt = AttachmentSerializingImpl.deserializeAttachmentData(nbt, wrapperLookup);
+
+		// If the NBT is devoid of data attachments, treat it as a no-op, rather than wiping them out.
+		// Any changes to data attachments (including removals) post-load are done independently of this
+		// code path, so we don't need to blindly overwrite it every time if Vanilla MC sends updates
+		// (i.e. block entity updates) sans data attachments. See https://github.com/FabricMC/fabric/issues/4638
+		if (fromNbt == null) {
+			return;
+		}
+
+		this.fabric_dataAttachments = fromNbt;
 
 		if (this.fabric_shouldTryToSync() && this.fabric_dataAttachments != null) {
 			this.fabric_dataAttachments.forEach((type, value) -> {


Roundaround:
The only thing I couldn't figure out was a good way to test this, since it touches so many things. If you have any ideas let me know

apple502j:
Maybe tests?

modmuss50:
Yeah, something about this doesnt feel right. Would it be possible to add some test code to prove this (doesnt matter too much if its not an automated one).

Roundaround:
I'm not sure what a test for this would look like - with how many code paths go through it, you'd need a LOT of game tests to test all the scenarios (i.e. initial load, deferred chunk load, player join/respawn/teleport/etc., set nbt with commands, apply entity component via nbt, copy target, modify data attachments server-side, modify data attachments client side, modify data attachments on one client and check a separate client...) for each attachment target type.

What I can say though is that the reason this bug exists only for block entities is that they're the only attachment target that does s2c syncs via catch-all NBT write/read (`BlockEntityUpdateS2CPacket`), whereas the other targets have more specialized sync packets (i.e. `EntityS2CPacket` or `EntityAttributesS2CPacket`). I can actually loosely confirm this by slapping a log output at the top of this method printing out the target's classname:

```java
@Override
public void fabric_readAttachmentsFromNbt(NbtCompound nbt, RegistryWrapper.WrapperLookup wrapperLookup) {
	AttachmentEntrypoint.LOGGER.info("Reading attachments from NBT for target of type {}", this.getClass().getSimpleName());
	// ...
}
```

After the initial explosion of logs for everything in the world as it gets loaded, it goes dead quiet, except for newly-spawned entities and modified block entities. If I go one step further and log whenever the incoming NBT has no attachments and fabric_dataAttachments is already non-null, it only ever fires for block entities:

```java
@Override
public void fabric_readAttachmentsFromNbt(NbtCompound nbt, RegistryWrapper.WrapperLookup wrapperLookup) {
	AttachmentEntrypoint.LOGGER.info("Reading attachments from NBT for target of type {}", this.getClass().getSimpleName());

	IdentityHashMap<AttachmentType<?>, Object> fromNbt = AttachmentSerializingImpl.deserializeAttachmentData(nbt, wrapperLookup);

	if (fromNbt == null) {
		if (this.fabric_dataAttachments != null) {
			AttachmentEntrypoint.LOGGER.info(
					"Attachments missing for target of type {}",
					this.getClass().getSimpleName()
			);
		}
		return;
	}
	
	// ...
}
```

I can't think of a good way to test this other than just using the API and confirming it works. Ultimately any tests for this should actually exist in the base MC code (and TBF probably do, we just can't see them) to ensure this contract remains true for additional MC versions going forward. Is there any pattern in Fabric's codebase of testing purely vanilla MC code to ensure our assumptions hold true?

modmuss50 (approved):
I think this is acceptable, I dont fully understand the whole flow. But it does seem silly to me if there is a case where fabric_readAttachmentsFromNbt is called twiced (ideally there shouldn't but as you said its complex). The change make sense, im happy to try it as its simple, and can easily be changed again in the future if needed.

PR #4643, opened 2025-05-19
Add LootTableEvents#MODIFY_DROPS

cputnam-a11y:
Add an event for modifying the drops of loot tables more directly. This is not intended to replace the existing loot table events, but to allow for widespread modification of drops from every loot table. Example cases include special tools and others.


fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
```diff
@@ -95,16 +99,25 @@ private LootTableEvents() {
...
+			listener.modifyDrops(key, context, drops);
+		}
+	});
 
 	@FunctionalInterface
 	public interface Replace {
 		/**
 		 * Replaces loot tables.
 		 *
-		 * @param key              the loot table key
```
modmuss50: Nit: Dont change existing code for the sake of it.

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
```diff
@@ -130,8 +143,19 @@ public interface Loaded {
...
 		 *
 		 * @param resourceManager the server resource manager
-		 * @param lootRegistry     the loot registry
+		 * @param lootRegistry    the loot registry
 		 */
 		void onLootTablesLoaded(ResourceManager resourceManager, Registry<LootTable> lootRegistry);
 	}
+
+	@FunctionalInterface
+	public interface Drop {
```
modmuss50: Call it ModifyDrops

modmuss50 (commented):
Im not very familar with loot stuff at all, ill have to sit down and properly look at this before I can give a meaningful review. If anyone is more familar with this please chime in.

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
cputnam-a11y: Oops. I didn't intend to. I'll revert it.

cputnam-a11y:
@modmuss50 , do you have an opinion on the id lookup? I don't like either option I included, but it seems like there aren't many others

modmuss50:
> @modmuss50 , do you have an opinion on the id lookup? I don't like either option I included, but it seems like there aren't many others

Im not sure, generally registries entries arent supposed to know their own registry ids, in vanilla such as blocks it gets passed in at construction. Could the key be set when the LootTable object is created/registered? As for the general code cleanup dont worry about trying to do everything in one line as you have in option 2, it can make it quite hard to read.

I am a little hesiant about this PR as it does go against the whole datadriven idea, but I do get there might not be another way, such as the example you provided. This is really why id like someone else more familar with loot stuffs to provide some input. CC: @Juuxel as I think they wrote the original API.

cputnam-a11y:
In my original implementation in a separate project I have the id set on creation, but It added quite a bit of complexity. 
<br/> To be clear, this isn't adding any additional functionality, just making the existing functionality easier to implement. It's possible to post process all loot tables by applying a custom loot function to every loot table, but that seems jank and bloaty

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
```diff
@@ -95,6 +99,15 @@ private LootTableEvents() {
 			listener.onLootTablesLoaded(resourceManager, lootManager);
 		}
 	});
+	/**
+	 * This event can be used for cases where the MODIFY and REPLACE events are inconvenient, such as when you are modifying the result of many loot tables that are unknown, and don't wish to add a custom LootFunction to every table
```
Juuxel:
```suggestion
	/**
	 * This event can be used for cases where the {@link #MODIFY} and {@link #REPLACE} events are inconvenient, such as when you are modifying the result of many loot tables that are unknown,
	 * and don't wish to add a custom loot function to every table.
```

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
```diff
@@ -134,4 +147,15 @@ public interface Loaded {
...
+
+	@FunctionalInterface
+	public interface ModifyDrops {
+		/**
+		 * Called after a loot table is finished generating drops to modify drops.
+		 * @param key the loot table's registry key
+		 * @param context the loot context for the current drops
+		 * @param drops the list of drops from the loot table to modify
+		 */
+		void modifyLootTableDrops(RegistryKey<LootTable> key, LootContext context, List<ItemStack> drops);
```
Juuxel: Should this have the loot table itself as a parameter too?

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/mixin/loot/LootTableMixin.java
```diff
@@ -0,0 +1,92 @@
...
+
+		// noinspection EqualsBetweenInconvertibleTypes
+		this.key = lootTableRegistryWrapper
+				.streamEntries()
+				.filter(it -> it.value().equals(this))
+				.findFirst()
+				.map(RegistryEntry.Reference::registryKey)
+				.orElseThrow(
+						() -> new IllegalStateException("LootTable appears to be unregistered, but has been asked to generate loot")
+				);
```
Juuxel: This is quite long for a mixin injector, so I'd extract it into a method in LootUtil that fetches the key.

Juuxel (changes requested):

Juuxel:
> I am a little hesiant about this PR as it does go against the whole datadriven idea

Yeah, I dislike in general how the API handles all custom loot additions in code. IMO we could add a data pack-based system on top of this API that can handle the main feature of the loot API - adding loot pools. The mass modification of drops like in this PR could also be handled with some kind of "conditional global loot functions" that only apply when one tool is held, for example.

In any case, I think the event makes sense and it's probably more efficient than modifying each and every loot table to have custom loot functions. (I can imagine that it'd increase memory use a bit if every loot table had superfluous data that was the same between all tables 😛)

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
cputnam-a11y: In my initial test impl, it was there, but it's easy to get it from the LootContext via the world and the registry key, so I omitted it. I'm not opposed to it. I think for all the hassle of getting the key, getting the table from the key is only like 4 loc, but if every mod is going to want it, it makes sense to just pass it. We could also pass RegistryEntry$Reference, as that combines the key and value

Juuxel: True, the registry entry would make a lot of sense if it's easily available

cputnam-a11y:
Screenshot 2025-05-24 at 11 15 11 AM
huh?

cputnam-a11y:
not done with the tests yet

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/mixin/loot/LootTableMixin.java
cputnam-a11y: done.

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
cputnam-a11y: done.

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java
```diff
@@ -95,6 +100,16 @@ private LootTableEvents() {
 			listener.onLootTablesLoaded(resourceManager, lootManager);
 		}
 	});
+	/**
```
Juuxel, line 104:
```suggestion

	/**
```
Forgot to include this in the previous review

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/mixin/loot/LootTableMixin.java
```diff
@@ -0,0 +1,67 @@
...
+import net.minecraft.loot.LootTable;
+import net.minecraft.loot.context.LootContext;
+import net.minecraft.registry.entry.RegistryEntry;
+
+import net.fabricmc.fabric.api.loot.v3.LootTableEvents;
+import net.fabricmc.fabric.impl.loot.FabricLootTable;
+import net.fabricmc.fabric.impl.loot.LootUtil;
+
+@Mixin(value = LootTable.class, priority = 3000 /* arbitrary, but requires mods to explicit set a priority to wrap the fabric event.*/)
+public class LootTableMixin implements FabricLootTable {
```
Juuxel:
```suggestion
class LootTableMixin implements FabricLootTable {
```

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/mixin/loot/LootTableMixin.java
```diff
@@ -0,0 +1,67 @@
...
+	 * the key of this loot table, if initialized.
+	 */
+	@Unique
+	@Nullable
+	RegistryEntry.Reference<LootTable> entry = null;
+
+	@WrapMethod(method = "generateUnprocessedLoot(Lnet/minecraft/loot/context/LootContext;Ljava/util/function/Consumer;)V")
+	private void fabric$modifyDrops(LootContext context, Consumer<ItemStack> lootConsumer, Operation<Void> original) {
+		if (entry == null) {
+			this.entry = LootUtil.getEntry(context.getWorld(), (LootTable) (Object) this);
```
Juuxel, line 50: What loot tables does this check trigger for? The entries are also set in `ReloadableRegistriesMixin`, which seems to me like it should cover everything.

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/impl/loot/LootUtil.java
```diff
@@ -47,4 +52,23 @@ public static LootTableSource determineSource(Resource resource) {
...
+				.getOptional(RegistryKeys.LOOT_TABLE)
+				.orElseThrow(() -> new IllegalStateException("Failed to fetch LootTable wrapper from WrapperLookup"));
+
+		return lootTableRegistryWrapper
+				.streamEntries()
+				.filter(it -> it.value().equals(table))
+				.findFirst()
+				.orElseThrow(
+						() -> new IllegalStateException("LootTable appears to be unregistered, but has been asked to generate loot")
+				);
```
Juuxel, line 72: This looks a bit inefficient but I couldn't find a better alternative with a quick look at the game code 🤔

(Not something that needs fixing necessarily, just a note)

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/mixin/loot/LootTableMixin.java
cputnam-a11y: Only ones added after fabric events, or via registry replacement. It's not a supported practice, and should almost never happen, but there are cases where it's unavoidable. In practice I expect it to be used once every 10,000 mods

fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/impl/loot/LootUtil.java
cputnam-a11y: Yeah. I think it's okay though now because it should only happen in extreme situations

PepperCode1 (approved):

PepperCode1:
A data driven system can be built on top of this event, where the lookup and invocation of a modification definition (retrieved from a custom dynamic registry) happen in an event handler. If Fabric or a third-party library ever add such a system, I do not think it will conflict with the existence of this event.

Juuxel (approved):

modmuss50 (approved):

modmuss50:
This doesnt cleanly cherrypick into 1.21.5, if its wanted there please make a seperate PR.

cputnam-a11y:
We'll see. I'm don't really mod the older versions, so I might just leave it for whoever wants it and continue pring new stuff on latest. I can't figure out how to not screw up git, so I'm just waiting until my prs merge to make the next ones. I have a local backlog of quite a few

modmuss50:
Thats fine, if someone wants it let me know and I can maybe look into it. I just dont like to rush cherry picking PRs that need none trivial changes as thats when things go wrong.

If you want help with git to have multiple PRs id (or someone else) would be happy to help you figure it out. Ideally you should be creating a new branch for each PR.

cputnam-a11y:
Yeah, I tried that, but it still got to a state where I ended up making another new branch for each pr, cherry picking the pr commits onto that, and pring that. When I have time I need to sit down and figure it out. I may take you up on help later

PR #4644, opened 2025-05-19
Add tag to make open trapdoors above tagged blocks climbable

Juuxel:
This is useful for custom ladders since vanilla only has this feature for the `minecraft:ladder` block.

Alternatives:
- Just replace vanilla's ladder check with an `instanceof LadderBlock` check. This automatically applies to any custom ladders, but it only supports `LadderBlock`s.
- Add a method to `FabricBlock(State)` that can do the check, possibly with a default for `LadderBlock`s. This is inconvenient for multiplatform mods and less simple than a tag. On the other hand, custom ladders that don't extend `LadderBlock` could do the same direction check as `LadderBlock`s do.


modmuss50 (approved):

maityyy:
Why `fabric-block-api` instead of `fabric-tag-api`? Or should `fabric-tag-api` actually be called `fabric-tag-aliases-api`? :D

Juuxel:
I think tag API is more about the tag "system" while these tags are just one way to access an API related to blocks.

apple502j (commented):
I think this should go to Entity Events - but not worth arguing over.

maityyy:
> I think tag API is more about the tag "system" while these tags are just one way to access an API related to blocks.

Fair, but now I'm confused by `FabricBlockTags`. It seems to me that any other module could also have this class name (vanilla sets the naming style, `Block` sounds like a tag type, not a module name), so I would rename it to `FabricBlockApiTags` (so, by my logic, any other module should use `FabricItemApiTags`, `FabricBiomeApiTags` and so on)

Juuxel:
@apple502j True, since it's more related to entities than blocks, but it's not really an event either 😅

I can move it, which would also mean renaming the class to `FabricEntityApiBlockTags` or similar.

Juuxel:
I've renamed the class to `BlockFunctionalityTags`. The idea here is that all Fabric block tags that add new functionality to blocks would be contained in this class. (And similarly, we'd have corresponding classes for item and entity tags in the item and entity event APIs)

(cc @maityyy)

maityyy:
> I've renamed the class to `BlockFunctionalityTags`. The idea here is that all Fabric block tags that add new functionality to blocks would be contained in this class. (And similarly, we'd have corresponding classes for item and entity tags in the item and entity event APIs)
> 
> (cc @maityyy)

👍

This may not apply to this PR, but should there be a corresponding note in CONTRIBUTING.md?

PR #4647, opened 2025-05-19
Merge client tag API into tag API

Juuxel:
Note: only needs a minor version for the tag API after merging to keep compatible versioning (client tags was 1.1, while tag API is 1.0).


PepperCode1 (approved):

modmuss50 (approved):
Going to release this with this snapshot, no point in waiting.

PR #4651, opened 2025-05-21
Remove some very old deprecated modules.

modmuss50:
All of these modules were deprecated over 2 years ago, in the case of commands-v0 5 years ago. And for a long time they werent present in development environments by default.

I know we generally dont like to remove deprecated for the sake of it, but there is literally no reason for them to exist anymore. If you have a specific reason to use these in 1.21.6 let me know.


PepperCode1 (approved):

modmuss50:
As discussed loot v2 should be added back til 1.22

PR #4653, opened 2025-05-24
Call AFTER_PLAYER_CHANGE_WORLD on respawn in different world

DennisOchulor:
Currently, the AFTER_PLAYER_CHANGE_WORLD event does not fire when a player respawns in a different world. This causes a problem for Data Attachment syncing as that API relies on this event to sync world attachments correctly.

This PR fires the event on player respawn if the world the player respawns in is different.


modmuss50 (approved):

DennisOchulor:
Is it possible for this fix to be backported to 1.21.1 directly? Or should I make another PR for the backport?

modmuss50:
> Is it possible for this fix to be backported to 1.21.1 directly? Or should I make another PR for the backport?

A seperate PR would be ideal, im going to merge this for 1.21.5/.6 now and then we can look at getting a backport out shortly.

PR #4654, opened 2025-05-25
Fix #4487 for 1.20.1

### LINKED ISSUE #4487: [1.20.1] FakePlayer-related memory leak not fixed despite backport
We've noticed that the memory leak issue related to the GlobalReceiverRegistry tracking FakePlayers still occurs in API 0.92.3+1.20.1 as downloaded from Modrinth. The issue has been discussed in https://github.com/Fabricators-of-Create/Create/issues/1684 and https://github.com/FabricMC/fabric/issues/3974 and a fix was backported in https://github.com/FabricMC/fabric/pull/4331 but the issue still occurs. Easiest way to reproduce that I've found is to run Create and make a gantry with a deployer rapidly reassemble itself.

Auroali:
This fixes `ServerPlayNetworkAddon`s being registered with their corresponding `GlobalReceiverRegistry` even when their `ServerPlayerNetworkHandler` implements `UntrackedNetworkHandler`


modmuss50:
Does this issue affect newer versions as well?

Auroali:
i don’t believe so, seems like newer versions only call startSession in lateInit

modmuss50:
I have pushed a change to this to make it a little bit less risky, there is no point in changing the regular path behaviour in an old version.

PR #4664, opened 2025-05-28
Move `BlockRenderLayerMap` to Rendering v1 module

PepperCode1:
Changes in 1.21.6 completely broke the old class so now is the best time to move it. The old module was removed. The design of the class was changed to use static methods instead of an `INSTANCE` field.


modmuss50 (approved):
Looks great, thanks. Ill get this released very shortly.

PR #4682, opened 2025-06-06
Make ComponentTooltipAppenderRegistry a class instead of an interface

Juuxel:
All of its methods are static, so it doesn't make sense for it to be a non-extendable(!) interface.

This change is **not binary compatible** - calling code will throw an `IncompatibleClassChangeError` since the bytecode references an interface method reference. It's still source compatible and mods will only need a recompilation with no changes.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/ComponentTooltipAppenderRegistry.java
@@ -17,7 +17,6 @@
 package net.fabricmc.fabric.api.item.v1;
 
 import com.google.common.base.Preconditions;
-import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.component.ComponentType;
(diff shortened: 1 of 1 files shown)


PepperCode1 (approved):

modmuss50:
There is nothing wrong with this, im just not sure I see the benfit when there are clear downsides (the breaking change)?

Juuxel:
It makes the code cleaner and is a clear mistake (since it was originally an instance based API iirc).

We can also just avoid the breaking change by waiting until the next breaking snapshot cycle. There's no rush with this PR.

Fuzss:
There are also other cases of non-extendable interfaces containing only static methods, like:
https://github.com/FabricMC/fabric/blob/1.21.6/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudElementRegistry.java

modmuss50:
Im going to close this PR, there is nothing inherently wrong with it being an interface and the breaking change is far from ideal. Lets keep it in mind to make sure that new APIs dont end up like this and if this API does change in the future we can do it then. For now I dont see point in keeping this PR open.

If you disagree let me know.

PR #4688, opened 2025-06-09
[1.21.6] Add api for handling hud status bars

Fuzss:
Maintaining mods adding their own status bars above the hotbar (like health, armor, food, air bubbles, thirst, stamina) has always been hard to deal with on Fabric:
- Handling of a bunch of vanilla conditions, like when air bubbles / mount health show
- No way of knowing where other mods place their status bars, leads to potential overlapping

This API is built as an extension to the existing [HudElementRegistry](https://github.com/FabricMC/fabric/blob/1.21.6/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudElementRegistry.java), and allows for separately registering height providers for your status bars. It then allows for retrieving the height for your bar to render at during rendering.

### Approach
Initially I was considering the NeoForge implementation, which simply adds two parameters `rightHeight` and `leftHeight` that every status bar hud element uses when drawing and then adds its own height to it for the bars rendering after it.

That system is pretty simple to use and works great. But the major downside is that it requires a very large amount of patches, since status bars do not render in the correct order in vanilla for such a system to work. 

Vanilla renders in this order:
> Armor -> Health -> Food -> Air -> Mount

While the system used by NeoForge needs the following order:
> Health -> Armor -> Mount -> Food -> Air

Therefore the system implemented in this PR works like so:

1. Every status bar must register a height provider, which corresponds to a hud element added in the corresponding registry. Replacing existing providers is supported, as `HudElementRegistry` allows for that, too.
2. After registration when the client starts the height provider registry is frozen and heights for every status bar are computed, based on all the bars that are to be rendered below. The order in which bars render is defined via the `HudElementRegistry`.
3. Then during rendering every status bar can retrieve the height for it to render at via `HudStatusBarHeightRegistry::getHeight`.

This design was chosen to minimize patches and to be able to built on top of the existing hud api. In fact, this api does not require any new Mixins at all.

### Issues / Considerations for feedback
- Vanilla status bars are incorporated via wrapping with `PoseStack::translate` calls in favor of using Mixins to patch screen coordinates, for reducing the patch surface and therefore reducing the maintenance burden. However, a Mixin based implementation may improve mod compatibility with mods not using this API.
- Lookup in `HudStatusBarHeightRegistry::getHeight` is backed by a `HashMap`, might introduce dedicated key objects instead of using `Identifier` for identity based lookup.
- Potentially merge `HudStatusBarHeightRegistry` into `HudElementRegistry`.
- `HudStatusBarHeightRegistry::getHeight` currently returns zero when called too early or for non-existing elements. Could change to throwing an exception instead.
- Might replace `ToIntFunction<PlayerEntity>` with a dedicated interface

### Example
A [test mod](https://github.com/Fuzss/fabric/blob/status-bar-height-api/fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudStatusBarHeightsTest.java) has been added: 
- Adds a toughness bar which renders below player health on the left
- Adds a stamina bar which render above the food bar on the right


modmuss50:
Checkstyle needs fixing here.

Fuzss:
Will do. Wanted to wait for some feedback to see if this is an ok direction to go in before comiting more time to this. 😅

modmuss50:
Just looking at it now.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,303 @@
...
+	 *
+	 * <p>These providers do NOT
+	 * return the heights of individual elements; instead they return the height at which an element should render at,
+	 * which is computed by summing all the heights from providers considered "below" an element.
+	 */
+	@Nullable
+	static Map<Identifier, ToIntFunction<PlayerEntity>> resolvedHeightProviders;
+
+	static {
+		ImmutableMap.Builder<Identifier, ToIntFunction<PlayerEntity>> builder = ImmutableMap.builder();
```
modmuss50: Can you use Map.of? You might even be able to move it out of the static block if you move the VANILLA_HEIGHT_PROVIDERS field down.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,303 @@
...
+	public static void addRight(Identifier id, ToIntFunction<PlayerEntity> heightProvider) {
+		if (resolvedHeightProviders == null) {
+			RIGHT_HEIGHT_PROVIDERS.put(id, heightProvider);
+		} else {
+			throw new IllegalStateException("Height provider registry already frozen!");
+		}
+	}
+
+	public static int getHeight(Identifier id) {
+		// returns zero for invalid calls, can adjust this to throw instead
```
modmuss50: Yes, always throw if something isnt in an expected state, a crash that developer is either going to see in dev or at worse reported by a player is better than weird/unxpected behaviour.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,303 @@
...
+		PlayerEntity player = MinecraftClient.getInstance().inGameHud.getCameraPlayer();
+		if (player != null) {
```
modmuss50: Prefer to use the [early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8), it makes the code much more readable. Also apples to the addRight/addLeft methods.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudStatusBarHeightsTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+	@Override
+	public void onInitializeClient() {
+		testToughnessBar();
+		testStaminaBar();
+	}
+
+	private static void testToughnessBar() {
+		// register a toughness bar showing below the vanilla health bar
+		Identifier resourceLocation = Identifier.of("fabric-rendering-v1-testmod", "toughness_bar");
```
modmuss50: Whats resourceLocation 😆

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,303 @@
...
+
+	@Override
+	public void onInitializeClient() {
+		ClientLifecycleEvents.CLIENT_STARTED.register((MinecraftClient minecraft) -> {
+			HudStatusBarHeightRegistryImpl.init();
+		});
+	}
+
+	public static void addLeft(Identifier id, ToIntFunction<PlayerEntity> heightProvider) {
+		if (resolvedHeightProviders == null) {
```
modmuss50, line 194: Should this check that the given ID has been registered with the hud elements api?

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudStatusBarHeightsTest.java
```diff
@@ -0,0 +1,134 @@
...
+				}
+
+				if (k * 2 + 1 > i) {
+					guiGraphics.drawGuiTexture(RenderPipelines.GUI_TEXTURED, TOUGHNESS_EMPTY_SPRITE, l, j, 9, 9);
+				}
+			}
+		}
+	}
+
+	private static void testStaminaBar() {
```
modmuss50: Nit pick: move testStaminaBar above renderArmor, its more imporant.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+	 * {@link HudElementRegistry#replaceElement(ResourceLocation, Function)}.
+	 *
+	 * <p>Registration is frozen once the client has fully started.
+	 *
+	 * @param id             the {@link Identifier} identifier; must be registered with a corresponding
+	 *                       {@link HudElement} in {@link HudElementRegistry}.
+	 * @param heightProvider a {@link ToIntFunction} that takes a {@link PlayerEntity} from
+	 *                       {@link InGameHud#getCameraPlayer()} and returns the height.
+	 */
+	static void addLeft(Identifier id, ToIntFunction<PlayerEntity> heightProvider) {
```
modmuss50: Is passing the Player to the function the best solution? What was the reasoning behind passing the Player as opposed to MinecraftClient or even nothing?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,303 @@
...
+				heightProviderLookup,
+				orderedHeightProviders);
+		return reduceToIntFunctions(heightProviderLookup.get(resourceLocation), heightProvider, Integer::sum);
+	}
+
+	private static <T> ToIntFunction<T> reduceToIntFunctions(ToIntFunction<T> first, ToIntFunction<T> second, IntBinaryOperator operator) {
+		return (T t) -> operator.applyAsInt(first.applyAsInt(t), second.applyAsInt(t));
+	}
+
+	private static void applyVanillaHeightProviders(Map<Identifier, ToIntFunction<PlayerEntity>> resolvedHeightProviders, ToIntFunction<PlayerEntity> maxHeightProvider) {
```
modmuss50: Can we skip this if no mod has registered a height provider?

modmuss50 (commented):
This is really clever, and looks like a good approach. Expanding the existing automated client tests would be handy, maybe to validate the the vanilla bars render correctly, and then with some modded bars.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
Fuzss: The status bars generally reflect some state of the player. E.g. in a project of mine where I add a stamina bar the data for the bar is stored as a synced attachment, so the player is useful for getting that. I'm guessing that's how it's gonna work for most mods.

I don't think `MinecraftClient` is too useful, there is nothing that is needed from there. The vanilla status bars do need `InGameHud` occasionally for some helper methods from there, but that won't be the case for modded status bars.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/HudStatusBarHeightsTest.java
Fuzss: Usually arrange methods according to code flow, but sure. 👍

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: Yes, `ImmutableMap::of` will work. `Map::of` does not preserve insertion order which is required here.

modmuss50: > Map::of does not preserve insertion order which is required here.

Ah yeah, I didnt think of that. Makes sense 👍

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: Implemented when resolving height providers, so it does not matter if a height provider or the hud element is registered first. 👍

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
modmuss50: Ok, I guess we can go with this. Im not sure if just using `IntSupplier` would be better as you can easily get the player from MinecraftClient (easy to get statically).

fabric-rendering-v1/src/client/resources/fabric-rendering-v1.accesswidener
```diff
@@ -0,0 +1,7 @@
+accessWidener v2 named
+
+accessible  field   net/minecraft/client/gui/hud/InGameHud    renderHealthValue I
```
modmuss50: This would be much better as accessors, adding access wideners has a none zero build peformance cost. They should only be used when absolutely required.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
Fuzss: I do think every status bar will require the player.
Also the proper way for getting the player here is via `InGameHud::getCameraPlayer`, so by providing the player we can make sure it comes from there. 
And it helps with nullability of `InGameHud::getCameraPlayer` which is can never be null here.

modmuss50: Ok, fair enough lets leave it as is then. It was only a minor thing anyway.

modmuss50 (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+	 * that the custom element associated with the given {@code id} occupies. This height contributes to the total
+	 * offset applied to elements positioned above it on the right side. Conditions implemented for the rendering of the
+	 * actual element must also be taken into account here; so when an element currently does not actually render
+	 * {@code 0} must be returned.
+	 *
+	 * <p>Vanilla height providers for this side are: {@link HudStatusBarHeightRegistryImpl#HEALTH_BAR},
+	 * {@link HudStatusBarHeightRegistryImpl#ARMOR_BAR}
+	 *
+	 * <p>Existing height providers (like vanilla) can be replaced to coincide with
+	 * {@link HudElementRegistry#replaceElement(ResourceLocation, Function)}.
```
kevinthegreat1:
```suggestion
	 * {@link HudElementRegistry#replaceElement(Identifier, Function)}.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+	 * that the custom element associated with the given {@code id} occupies. This height contributes to the total
+	 * offset applied to elements positioned above it on the right side. Conditions implemented for the rendering of the
+	 * actual element must also be taken into account here; so when an element currently does not actually render
+	 * {@code 0} must be returned.
+	 *
+	 * <p>Vanilla height providers for this side are: {@link HudStatusBarHeightRegistryImpl#MOUNT_HEALTH},
+	 * {@link HudStatusBarHeightRegistryImpl#FOOD_BAR}, {@link HudStatusBarHeightRegistryImpl#AIR_BAR}
+	 *
+	 * <p>Existing height providers (like vanilla) can be replaced to coincide with
+	 * {@link HudElementRegistry#replaceElement(ResourceLocation, Function)}.
```
kevinthegreat1:
```suggestion
	 * {@link HudElementRegistry#replaceElement(Identifier, Function)}.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+ * providers define the vertical space occupied by HUD elements, known as status bars, which are positioned on the left
+ * and right sides above the player's hotbar.
+ *
+ * <p>Registering a height provider allows the game to automatically adjust the layout of existing
+ * HUD elements, including vanilla ones, to accommodate new bars without overlap. The system calculates the cumulative
+ * height from registered providers on each side and shifts elements like health, armor, food, and air bars
+ * accordingly.
+ *
+ * <p>Height providers are associated with a {@link Identifier ResourceLocation}
+ * identifier. The identifier must also be registered with a corresponding {@link HudElement} in
```
kevinthegreat1:
```suggestion
 * <p>Height providers are associated with a {@link Identifier}
 * The identifier must also be registered with a corresponding {@link HudElement} in
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+ * {@link HudElementRegistry}. The relative positioning to other HUD elements is determined from that registration. For
+ * instance, registering a height for {@link VanillaHudElements#ARMOR_BAR} via
+ * {@link #addLeft(Identifier, ToIntFunction)} implies the custom bar is on the left side and affects the positioning of
+ * elements starting from the health bar downwards. The function itself should return the height of the custom bar.
```
kevinthegreat1: So, if I register my bar relative to `ARMOR_BAR` in `HudElementRegistry`, then my bar is positioned below the armor bar? I think I'm misunderstanding something, or this is not quite clear.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,373 @@
...
+
+		for (Identifier heightProviderLocation : orderedHeightProviders) {
+			if (heightProviderLocation.equals(resourceLocation)) {
+				return heightProvider;
+			} else if (heightProviderLookup.containsKey(heightProviderLocation)) {
+				heightProvider = reduceToIntFunctions(heightProvider,
+						heightProviderLookup.get(heightProviderLocation),
+						Integer::sum);
+			}
+		}
```
kevinthegreat1: I can't quite figure out why this works? `orderedHeightProviders` contains all the vanilla hud elements in order, then modded hud elements. This adds all the bars up to the given element. If given a modded bar, doesn't this add all the vanilla bars regardless of whether the vanilla bar is before the modded bar?

kevinthegreat1 (commented):
A great addition but I have a few questions. Also, maybe I'm misunderstanding the api but I don't think we need to freeze the registry and calculate heights, we could just calculate it while rendering?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+ * A registry for {@link ToIntFunction ToIntFunction&lt;Player&gt;} instances, known as height providers. These
+ * providers define the vertical space occupied by HUD elements, known as status bars, which are positioned on the left
+ * and right sides above the player's hotbar.
+ *
+ * <p>Registering a height provider allows the game to automatically adjust the layout of existing
+ * HUD elements, including vanilla ones, to accommodate new bars without overlap. The system calculates the cumulative
+ * height from registered providers on each side and shifts elements like health, armor, food, and air bars
+ * accordingly.
+ *
+ * <p>Height providers are associated with a {@link Identifier}
```
kevinthegreat1:
```suggestion
 * <p>Height providers are associated with a {@link Identifier}.
```

kevinthegreat1: Sorry I forgot a period.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: The order in `orderedHeightProviders` depends on `HudElementRegistry`, so modded providers can be inserted in-between if they have been attached before / after vanilla hud elements.

Fuzss:
That might be possible, although resolving with all the map lookups is something I don't think we necessarily should do for every frame. Should be fair to have modders register their height providers only from their `ClientModInitializer`, I don't see a practical reason that would require this to be possible later on.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
kevinthegreat1: Yes but `getOrderedHeightProviders` seems to intentionally add all the vanilla bars, then add the remaining hud elements to `orderedHeightProviders`? The order is not exactly the order in `HudElementRegistry`?

Fuzss: It also adds all layers attached to the vanilla layers together with the vanilla layer itself in between. 
That's the intended behavior required for this whole API to work.
I think you might be missing the inner `for` loops in `getOrderedHeightProviders`.

Crendgrim:
Oh, I love to see this, this will be super useful to me!

A minor nitpick: the comments talk a lot about "pose stack" manipulation, but since fabric-api is written in yarn, should those be changed to "matrix stack" instead? Not sure how these names are handled for comments.

cputnam-a11y:
names are written by the author (as the author uses mojmap, a lot of it leaked though) see https://github.com/FabricMC/fabric/pull/4688#discussion_r2138657929 for prior discussion

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
Fuzss: Yeah, that was quite confusing. Changed it up a little, hope it's more clear now.

TheDeathlyCow:
> Might replace ToIntFunction<PlayerEntity> with a dedicated interface

I think this would be a good idea. Would make the intention clearer and javadoc can be provided for what exactly the `int` return value should be. The interface could extend `ToIntFunction<PlayerEntity>`, though. 

Perhaps it could also take a `ClientPlayerEntity` instead of a `PlayerEntity`? Just to provide as much information to consumers as possible.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,373 @@
...
+	static final int OVERLAY_MESSAGE_HEIGHT = 68 - DEFAULT_HEIGHT;
+	static final int TEXT_HEIGHT_DELTA = OVERLAY_MESSAGE_HEIGHT - HELD_ITEM_TOOLTIP_HEIGHT;
+	static final ToIntFunction<PlayerEntity> ZERO = (PlayerEntity player) -> 0;
+	/**
+	 * Height provider for the vanilla health bar.
+	 *
+	 * <p>Mods that would otherwise have a mixin for this are encouraged to instead register a replacement provider via
+	 * {@link HudStatusBarHeightRegistry#addLeft(Identifier, ToIntFunction)}.
+	 */
+	static final ToIntFunction<PlayerEntity> HEALTH_BAR = (PlayerEntity player) -> {
```
TheDeathlyCow: This feels really brittle to me, but I suppose there isn't like a dedicated method for this in vanilla? Perhaps some unit tests here might be a good idea to make sure these all match expected vanilla behaviour. 

Also, having these all as lambda functions stored in a field feels weird to me. This class is already quite big as it is, maybe they could be put in a separate impl class like `VanillaHeightProviders` with static methods instead? I know vanilla does use lambda fields, so this is mostly just a personal nitpick haha

cputnam-a11y: they could also be method references rather than lambdas, if that helps

Fuzss:
> > Might replace ToIntFunction with a dedicated interface
> 
> I think this would be a good idea. Would make the intention clearer and javadoc can be provided for what exactly the `int` return value should be. The interface could extend `ToIntFunction<PlayerEntity>`, though.
> 
> Perhaps it could also take a `ClientPlayerEntity` instead of a `PlayerEntity`? Just to provide as much information to consumers as possible.

What do you think would be a good name? `StatusBarHeightProvider` is what comes to mind, but it's not as short and juicy as e.g. `HudElement`. 😆

The player is the result of `InGameHud::getCameraPlayer` which returns `PlayerEntity`, so that's fine to keep like that.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: Unfortunately there are no dedicated methods, I don't think it's possible extracting that from vanilla. For tests the best approach would be screenshot comparisons, although I'm personally not really interested in getting into that right now. 😅

The lambdas should be fine, it's just an implementation class and the code chunks are very small. Adding a bunch of boilerplate shouldn't be necessary.

TheDeathlyCow:
`StatusBarHeightProvider` seems like a good name to me, you could also do it as an inner class of the registry, and have it just be `HeightProvider`

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
cputnam-a11y: The screenshot comparisons shouldn't be difficult to add in a separate pr. The client test system already supports them, so it would just be a matter of adding the initial screenshots, and adding the tests

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+ * {@link #addLeft(Identifier, StatusBarHeightProvider)} implies the custom bar is on the left side and affects the
+ * vertical positioning of elements starting from the armor bar upwards.
+ *
+ * <p>The final vertical offset for a HUD element is determined by summing the heights of all custom providers
+ * registered for elements that would appear "below" it on the same side. This includes all HUD elements that have been
+ * attached before it in {@link HudElementRegistry}.
+ *
+ * <p>For vanilla HUD element identifiers, see {@link VanillaHudElements}.
+ */
+public interface HudStatusBarHeightRegistry {
```
Juuxel: This should be a final class.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+	 */
+	static final StatusBarHeightProvider ARMOR_BAR = (PlayerEntity player) -> {
+		return player.getArmor() > 0 ? 10 : 0;
+	};
+	/**
+	 * Height provider for the vanilla mount health.
+	 *
+	 * <p>Mods that would otherwise have a mixin for this are encouraged to instead register a replacement provider via
+	 * {@link net.fabricmc.fabric.api.client.rendering.v1.hud.HudStatusBarHeightRegistry#addRight(Identifier,
+	 * StatusBarHeightProvider)}.
```
Juuxel: I don't think this paragraph is needed on each of these fields, especially since this is an impl class.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+	 * height providers, to compare with the actual height providers during rendering for potential translations for
+	 * vanilla status bars. Translations are achieved via matrix stack transformations; alternatively can also be
+	 * implemented via mixins.
+	 *
+	 * <p>Do not use {@link Map}, it does not preserve insertion order.
+	 */
+	static final Map<Identifier, ResolvedHeightProvider> VANILLA_HEIGHT_PROVIDERS = ImmutableMap.of(VanillaHudElements.HEALTH_BAR,
+			ResolvedHeightProvider.ZERO,
+			VanillaHudElements.ARMOR_BAR,
+			HEALTH_BAR::applyAsInt,
```
Juuxel:
```suggestion
			HEALTH_BAR,
```
It's already a `ToIntFunction` as far as I can tell

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+		boolean isInWater = player.isSubmergedIn(FluidTags.WATER);
+		return isInWater || airSupply < maxAirSupply ? 10 : 0;
+	};
+	/**
+	 * This serves two purposes: it provides a fixed order for some vanilla status bars; and it provides reduced vanilla
+	 * height providers, to compare with the actual height providers during rendering for potential translations for
+	 * vanilla status bars. Translations are achieved via matrix stack transformations; alternatively can also be
+	 * implemented via mixins.
+	 *
+	 * <p>Do not use {@link Map}, it does not preserve insertion order.
```
Juuxel: But the field is a `Map`?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+			VanillaHudElements.HEALTH_BAR,
+			HEALTH_BAR,
+			VanillaHudElements.ARMOR_BAR,
+			ARMOR_BAR);
+	/**
+	 * Height providers registered for the right side above the hotbar.
+	 *
+	 * <p>Used for checking if any custom height providers have been registered to potentially skip resolving later on.
+	 *
+	 * <p>Do not use {@link Map}, it does not preserve insertion order.
```
Juuxel: Same here

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+				orderedHeightProviders);
+	}
+
+	private static SequencedSet<Identifier> getOrderedHeightProviders(Map<Identifier, StatusBarHeightProvider> heightProviderLookup) {
+		// creates an ordered list of all height provider identifiers from the lookup,
+		// with a fixed order provided for some vanilla elements and other elements attached to those via the static map;
+		// all other elements are simply appended in the order they appear in the hud element registry
+		SequencedSet<Identifier> orderedHeightProviders = new LinkedHashSet<>();
+
+		for (Identifier resourceLocation : VANILLA_HEIGHT_PROVIDERS.keySet()) {
```
Juuxel: This is not a location of a resource

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+		}
+
+		return orderedHeightProviders;
+	}
+
+	private static ResolvedHeightProvider resolveHeightProvider(Identifier resourceLocation, Map<Identifier, StatusBarHeightProvider> heightProviderLookup, SequencedCollection<Identifier> orderedHeightProviders) {
+		// combines all height providers "below" a hud element for determining the height at which it should render at
+		ResolvedHeightProvider heightProvider = ResolvedHeightProvider.ZERO;
+
+		for (Identifier heightProviderLocation : orderedHeightProviders) {
```
Juuxel: All the "locations" should be "ids" here too

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,394 @@
...
+				heightProvider = reduceToIntFunctions(heightProvider,
+						heightProviderLookup.get(heightProviderLocation),
+						Integer::sum);
+			}
+		}
+
+		throw new IllegalStateException();
+	}
+
+	private static ResolvedHeightProvider resolveMaximumHeightProvider(Identifier resourceLocation, Map<Identifier, StatusBarHeightProvider> heightProviderLookup, SequencedCollection<Identifier> orderedHeightProviders) {
```
Juuxel: Same here

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudAccessor.java
Juuxel, line 1: Two things:
1. Do we prefix accessor/invoker methods in Fabric API generally?
2. If not, then the strings in the annos are not needed since the target methods and fields can be resolved automatically.

Juuxel (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/StatusBarHeightProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import net.minecraft.entity.player.PlayerEntity;
+
+/**
+ * Define the vertical space occupied by HUD elements, known as status bars, which are positioned on the left and right
+ * sides above the player's hotbar.
+ *
+ * @see HudStatusBarHeightRegistry
+ */
+@FunctionalInterface
+public interface StatusBarHeightProvider extends ToIntFunction<PlayerEntity> {
```
Chocohead, line 32: Is extending `ToIntFunction` really beneficial? If you are using a dedicated interface, it would be nice to have a more descriptive method than `applyAsInt`.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+	 *
+	 * <p>Vanilla height providers for this side are: {@link HudStatusBarHeightRegistryImpl#HEALTH_BAR},
+	 * {@link HudStatusBarHeightRegistryImpl#ARMOR_BAR}
+	 *
+	 * <p>Existing height providers (like vanilla) can be replaced to coincide with
+	 * {@link HudElementRegistry#replaceElement(Identifier, Function)}.
+	 *
+	 * <p>Registration is frozen once the client has fully started.
+	 *
+	 * @param id             the {@link Identifier} identifier; must be registered with a corresponding
```
Chocohead: The wording is quite clumsy here: `id - the Identifier identifier;`

Given you already know the type of the parameter, I don't think reiterating it being `Identifier` is adding anything.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/HudStatusBarHeightRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+	 * <p>Vanilla height providers for this side are: {@link HudStatusBarHeightRegistryImpl#MOUNT_HEALTH},
+	 * {@link HudStatusBarHeightRegistryImpl#FOOD_BAR}, {@link HudStatusBarHeightRegistryImpl#AIR_BAR}
+	 *
+	 * <p>Existing height providers (like vanilla) can be replaced to coincide with
+	 * {@link HudElementRegistry#replaceElement(Identifier, Function)}.
+	 *
+	 * <p>Registration is frozen once the client has fully started.
+	 *
+	 * @param id             the {@link Identifier} identifier; must be registered with a corresponding
+	 *                       {@link HudElement} in * {@link HudElementRegistry}.
```
Chocohead: Rogue asterisk here

Chocohead (changes requested):
Some nit picking from working out how the rendering works.

There's also several mentions of `(resource) location`s in `HudStatusBarHeightRegistryImpl` which don't make any sense given the context, they're IDs for the various `HeightProvider`s

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
TheDeathlyCow: I don't think it's necessary to do screenshot comparisons for this, you can just do regular unit tests with a mock player and check that the returned height of that provider matches the mock player's state. 

There might be some challenge mocking the `InGameHud`, since that is also used quite a lot to determine height too. But to me, that indicates that the client/HUD instance should probably be a parameter of the interface as it would be part of the render state.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Chocohead: Specifically don't use `Map#of`, as it has unspecified iteration order.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/StatusBarHeightProvider.java
Fuzss: I do think so. The implementation uses two extensions of `ToIntFunction`: `StatusBarHeightProvider` and `ResolvedHeightProvider`. They only exist separately to help make the implementation easier to understand. Having some interoperability between the two is used in the implementation, which is possibly due to both extending `ToIntFunction`.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InGameHudAccessor.java
Fuzss: I prefixed them now.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: This is necessary. The map stores `ResolvedHeightProvider`, `HEALTH_BAR` is a `StatusBarHeightProvider`.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: Moved to `HudStatusBarHeightRegistry`.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: I had it so `InGameHud` was provided to the interface originally, but in the end it was only useful for the vanilla providers, all helper methods in there are specifically tailored for the vanilla status bars. 
Neither `InGameHud` nor `MinecraftClient` ended up being useful for modded status bars, only the player is as outlined somewhere previously.

Fuzss:
I've found an oversight while porting one of my mods to this new system:

When replacing vanilla status bar hud elements (like the armor bar) via `HudElementRegistry` there is no guarantee the original height provider is still correct. It may be likely, but we cannot be sure. 
Therefore it is now mandatory to register a new height provider for replaced vanilla status bar hud elements.

Also now when the original height provider for a status bar has been replaced, the custom matrix stack transformations done for vanilla status bars are no longer applied to not interfere with the new provider.

Crendgrim:
So, to clarify: if a mod now replaces a vanilla element to augment it in some way, they will have to take care of the matrix stack transformations? How will this interact with two mods targeting the same element?

Would it be worth it to be able to somehow indicate "I have replaced this element but I can guarantee that its height doesn't change from that" (e.g. something like appleskin)?

Does this implementation even have to catch "mod overrides element but does not override height provider", or should that responsibility simply be handed down to the consumer? There's plenty of ways for mods to mess this up with direct mixins, after all.

Fuzss:
The matrix stack transformations are an internal workaround for the vanilla status bars to avoid having to patch too much of `InGameHud`. Mods that have their own status bars or replace vanilla status bars must use `StatusBarHeightProviderRegistry::getHeight` within their height providers.

I think you may be right that enforcing the registration of new height providers is probably not the way to go, especially as it seems likely that most mods will not want to make any changes there.
The idea was to prevent devs replacing the hud element and forgetting to change the height provider, but that might not be a valid concern after all. I'll play around a little more with it and likely remove the check again.

And yes, I assume some odd mixins are able to break the implementation. Still I think it's very important and useful to have a common implementation in Fabric that mods can built on, especially since some sort of interoperability between mods is definitely required here.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -264,25 +268,40 @@ private static SequencedSet<Identifier> getOrderedHeightProviders(Map<Identifier
...
+		// height providers for removed layers are skipped, as there is no way to remove them manually
+		if (!hudLayer.isRemoved() && heightProviderLookup.containsKey(hudLayer.id())) {
+			// replaced vanilla hud elements must have a new height provider registered to replace the functionality
+			// of our wrapping for vanilla hud elements to apply height provider offsets
+			if (hudLayer.isReplaced()) {
+				StatusBarHeightProvider heightProvider = heightProviderLookup.get(hudLayer.id());
+
+				if (heightProvider == VANILLA_LEFT_HEIGHT_PROVIDERS.get(hudLayer.id())
+						|| heightProvider == VANILLA_RIGHT_HEIGHT_PROVIDERS.get(hudLayer.id())) {
+					throw new IllegalStateException("Vanilla hud element " + hudLayer.id()
```
modmuss50: This is now technically a breaking change if an existing mod was replacing the vanilla hud element. An acceptable solution may be to throw this new error in dev only, and then return to the wrong behavour in prod? Its not ideal though.

There is no good solution here 🤔

Fuzss:
Ok, so I've spent some more time on this and I agree that forcing mods to register a new height provider when they have replaced the corresponding `HudElement` is not a good solution.

There are a number of mods that only add something to a `HudElement` without actually affecting its height.
An example is AppleSkin drawing additional texture elements on the food bar, which requires replacing the food bar `HudElement`, but does nothing for its height and therefore the existing vanilla height provider should be allowed to remain.

So having to register a new height provider together with a replaced `HudElement` has been removed again.

Instead I've implemented a different solution which works very well from my testing:

If a vanilla height provider has **not** been replaced, `HudStatusBarHeightRegistry::getHeight` will now return the value from the resolved vanilla height provider (which is the height provider operating on vanilla values without any interference from modded changes) as opposed to the actual resolved height provider.

This is correct since while the vanilla height provider is registered, the custom matrix stack transformations outlined before in this PR are applied for shifting the vanilla status bar, and they will also apply to the replaced `HudElement`, so that must receive the vanilla height value.

This is covered in `HudStatusBarHeightsTest::testArmorBar`, where it now does not matter whether a new height provider is registered for the replaced armor status bar or not, which is the expected behavior.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: I've removed this behavior again, I found a better solution for dealing with replaced hud elements that does not require mods to register a new height provider if they have changed nothing about the status bar height.

I've outlined that in a separate comment below.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/StatusBarHeightProvider.java
```diff
@@ -0,0 +1,37 @@
...
+	/**
+	 * @param player the {@link PlayerEntity} from {@link InGameHud#getCameraPlayer()}
+	 * @return the vertical space occupied by the status bar
+	 */
+	@Override
+	int applyAsInt(PlayerEntity player);
```
TheDeathlyCow: I think it would be preferable if this had a more meaningful name like `getHeight()`, and then `applyAsInt` can be a `@ApiStatus.NonExtendable` default method that just simply calls the more meaningful method.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,406 @@
...
+	@FunctionalInterface
+	public interface ResolvedHeightProvider extends ToIntFunction<PlayerEntity> {
+		ResolvedHeightProvider ZERO = (PlayerEntity player) -> 0;
+
+		/**
+		 * @param player the {@link PlayerEntity} from {@link InGameHud#getCameraPlayer()}
+		 * @return the vertical space occupied by all status bars "below" this one
+		 */
+		@Override
+		int applyAsInt(PlayerEntity value);
```
TheDeathlyCow: Same thing here. You could probably just have this interface extend the regular height provider too. Then, you might not even need the `ToIntFunction` thing.

TheDeathlyCow (commented):
Really excited to see this API. As a survival mod maker who has dealt with a lot of status bar shenanigans, it's much appreciated. I still just have a few comments on the interface set up.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/StatusBarHeightProvider.java
Fuzss: I don't see a reason for this, since the the functional interface will almost always be implemented via lambda.
But I'll just do it since you are the second person asking for this, doesn't matter after all.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
Fuzss: I'd rather keep both interfaces fully separate as their functionality is rather distinct.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/hud/StatusBarHeightProvider.java
TheDeathlyCow: I actually quite often implement my event listeners as classes, which is why I ask for it. Not always of course, but if the listener needs to keep track of its own data, or even if it's just kind of long, then I will definitely be moving it to a class.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/hud/HudStatusBarHeightRegistryImpl.java
```diff
@@ -0,0 +1,412 @@
...
+
+		if (player == null) {
+			throw new IllegalStateException("Trying to get status bar height for " + id + " without a camera player");
+		}
+
+		return DEFAULT_HEIGHT + resolvedHeightProviders.get(id).getResolvedHeight(player);
+	}
+
+	static void init() {
+		// skip resolving if no custom height providers have been registered
```
Crendgrim, line 228: What is the reason behind this check? I have a use case for knowing where the vanilla bars are, and was surprised to get an IllegalArgumentException when querying the height for `VanillaHudElements.HEALTH_BAR` when there was no custom bar registered.

Fuzss: You're right, that shouldn't be an empty map, but the defaults map. Fixed now. 👍

modmuss50 (approved):

PR #4690, opened 2025-06-09
Add `GatherDebugTextEvents`

Fuzss:
This adds two simple events that run when the text on the left and right sides of the `DebugHud` is gathered, respectively.
The mutable lists populated by vanilla are passed to the events, allowing for adding custom lines and editing existing ones.

This is a common injection point for mods, so it might be good to have in Fabric itself.

Note that the left side text does not include the last couple of lines:
- Debug charts: ...
- For help: press F3 + Q

This is intentional, they are purposefully added at the end by vanilla, mods should probably not insert anything after them.
Although this is of course up for discussion if there is a valid use case for receiving all lines via the event.

A test is included:


cputnam-a11y:
~~I had a way planned to do this as one event. do you think this is better off as two?~~ this looks clean

Fuzss:
Also wanted to mention how NeoForge does this:
There it's only a single event, which includes both lists separately. It also includes a `DrawContext`, and therefore not only allows for changing the lines, but also for completely taking over rendering. Mods like [BetterF3](https://modrinth.com/mod/betterf3) seem to use this.

Not sure if this is necessary here? Personally I'm fine with the scope of these events only allowing for changing the text in the lists and there should be enough mods using this to keep the events only supporting that. 
Also the patch surface for the NeoForge event is very large.

Fuzss:
Also what do I do to generate the license headers, is there a Gradle task for that or something in the IDE, or is it manual?

cputnam-a11y:
> Also what do I do to generate the license headers, is there a Gradle task for that or something in the IDE, or is it manual?

I think it's the spotlessApply task

cputnam-a11y:
I think what's here currently is acceptable. Also, this isn't mutually exclusive with another event, or a rendering API, so it would still be possible to add later if needed

modmuss50:
We previously rejected this idea as we dont want every mod adding pointless stuff here, even on your screenshot its overflowing the window bounds.

Fuzss:
Welcome to the messed up doubled resolution on retina displays haha. 
The text is already overflowing in vanilla actually.

Well shouldn't it be up to mod authors to decide what they want to add?
It's not breaking anything and there clearly is a use case.

modmuss50:
Yeah, I think its fine. People are going to do it anyway, so might as well make it easier I guess.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/GatherDebugTextEvents.java
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	public static final Event<Right> RIGHT = EventFactory.createArrayBacked(Right.class, callbacks -> lines -> {
+		for (Right callback : callbacks) {
+			callback.onGatherRightDebugText(lines);
+		}
+	});
+
+	@FunctionalInterface
+	public interface Left {
+		void onGatherLeftDebugText(List<String> lines);
```
modmuss50, line 58: Shame this takes a `String` and not `Text` to easily allow for formatting/colors and translations. I havent looked at the code but I assume there is a good reason you did it this way.

Fuzss: The vanilla implementation does not use `Text` at all.. 🙈
The strings themselves are what is rendered in the end.

Fuzss:
Should I migrate adding the `[Fabric] Active Renderer` line to the new event in that module?

modmuss50:
> Should I migrate adding the `[Fabric] Active Renderer` line to the new event in that module?

Yes, I think that would make sense.

modmuss50 (approved):
LGTM, could maybe add a tiny bit of javadoc to the event class. Other than that ill put it into last call.

PR #4691, opened 2025-06-10
Add `ScreenMouseEvents` for handling `Screen::mouseDragged`

Fuzss:
Events for handling `Screen::mouseDragged` have been missing from Fabric for a long time, there are plenty of mods this will be useful for, like Mouse Tweaks, Mouse Wheelie, Easy Shulker Boxes, etc.

So this adds three new events for handling `Screen::mouseDragged`:
- `ScreenMouseEvents$AllowMouseDrag`
- `ScreenMouseEvents$BeforeMouseDrag`
- `ScreenMouseEvents$AfterMouseDrag`

These follow the exact same style as all the other events found in `ScreenMouseEvents`.

I also included some very small cleanup duty for Javadoc in `ScreenMouseEvents`, and I've prefixed unique fields in `ScreenMixin` with `fabric$`.

This also includes a very basic test for the new events.

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
@@ -104,9 +104,48 @@ public final class ScreenMouseEvents {
 	}
 
 	/**
-	 * An event that is checks if the mouse should be allowed to scroll in a screen.
+	 * An event that checks if the mouse should be allowed to drag in a screen.
 	 *
-	 * <p>This event tracks amount of vertical and horizontal scroll.
+	 * <p>This event tracks the amount a mouse was dragged both vertically and horizontally.
+	 *
+	 * @return the event
+	 */
+	public static Event<AllowMouseDrag> allowMouseDrag(Screen screen) {
+		Objects.requireNonNull(screen, "Screen cannot be null");
+
+		return ScreenExtensions.getExtensions(screen).fabric_getAllowMouseDragEvent();
+	}
+
+	/**
+	 * An event called after mouse dragging is processed for a screen.
+	 *
+	 * <p>This event tracks the amount a mouse was dragged both vertically and horizontally.
+	 *
+	 * @return the event
+	 */
+	public static Event<BeforeMouseDrag> beforeMouseDrag(Screen screen) {
+		Objects.requireNonNull(screen, "Screen cannot be null");
+
+		return ScreenExtensions.getExtensions(screen).fabric_getBeforeMouseDragEvent();
+	}
+
+	/**
+	 * An event called after mouse dragging is processed for a screen.
+	 *
+	 * <p>This event tracks the amount a mouse was dragged both vertically and horizontally.
+	 *
+	 * @return the event
+	 */
+	public static Event<AfterMouseDrag> afterMouseDrag(Screen screen) {
+		Objects.requireNonNull(screen, "Screen cannot be null");
+
+		return ScreenExtensions.getExtensions(screen).fabric_getAfterMouseDragEvent();
+	}
+
+	/**
+	 * An event that checks if the mouse should be allowed to scroll in a screen.
+	 *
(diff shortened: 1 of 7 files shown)


kevinthegreat1:
Can we add this if we’re adding a new api? #4620

Fuzss:
Yes, it works exactly the same as the other screen events, both on the api surface as well as the implementation.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
```diff
@@ -54,51 +54,57 @@ abstract class ScreenMixin implements ScreenExtensions {
 	protected List<Drawable> drawables;
 
 	@Unique
-	private ButtonList fabricButtons;
+	private ButtonList fabric$buttonList;
```
modmuss50: No, this is wrong. Having @Unique on fields tottaly removes the need for this as it will rename the field if there are any clashes. We have a checkstyle rule to enforce Unique on all fields and methods like this.

The methods from the ScreenExtensions interface, do need to be prefixed as Mixin cannot rename them as their root is in another class outside the mixin.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
```diff
@@ -148,32 +187,38 @@ private ScreenMouseEvents() {
 	@FunctionalInterface
 	public interface AllowMouseClick {
 		/**
+		 * Checks if the mouse click should be allowed in a screen.
+		 *
+		 * @param screen the screen
 		 * @param mouseX the x position of the mouse
 		 * @param mouseY the y position of the mouse
 		 * @param button the button number, which can be identified by the constants in {@link org.lwjgl.glfw.GLFW GLFW}.
-		 * @see org.lwjgl.glfw.GLFW#GLFW_MOUSE_BUTTON_1
```
modmuss50, line 154: Why remove this comment?

modmuss50 (commented):
Looks like a reasonable approach. As Kevin suggested it might be a good idea to look at what https://github.com/FabricMC/fabric/pull/4620 was trying to as it can easily be done now without needing a mess of V2 events.

I would also suggest not trying to change existing code in a feature PR as it distracts away from the real change. Im more than happy to accept seperate PRs that cleanup code on their own.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
Fuzss: Fixed 👍
Although the linked class should probably be vanilla's own `InputUtil` for the constants. Unfortunately the mouse button constants seem to be missing their mappings in Yarn. 😆

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/mixin/screen/ScreenMixin.java
Fuzss: Thanks, wasn't aware of that behavior for `Unique`.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java
modmuss50: Thats likely because its impossible to know what they are as they will be inlined.

Fuzss: Certainly. I only know from working with the official mappings haha.

Fuzss:
> Looks like a reasonable approach. As Kevin suggested it might be a good idea to look at what #4620 was trying to as it can easily be done now without needing a mess of V2 events.

Sure, let's wait with this one until #4620 is resolved.

modmuss50 (approved):
I think this is fine, lets keep it consisten with what we already have.

PR #4693, opened 2025-06-10
ArmorRenderer: Add option to disable default head item rendering

Juuxel:
Adds a new method to `ArmorRenderer`, `shouldRenderDefaultHeadItem`, that can disable the default vanilla behaviour of rendering an item as its model when it doesn't have an armor asset id defined. Most custom armor doesn't need the extra item model, so it's useful to have an API to turn it off.

Also contains a fix for armor renderers not applying at all when the item doesn't have an armor model defined in the `equippable` component.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.mixin.client.rendering;
+
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.llamalad7.mixinextras.sugar.Local;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+import net.minecraft.client.render.entity.BipedEntityRenderer;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.impl.client.rendering.ArmorRendererRegistryImpl;
+
+@Mixin(BipedEntityRenderer.class)
+abstract class BipedEntityRendererMixin {
+	@ModifyExpressionValue(method = "getEquippedStack", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/feature/ArmorFeatureRenderer;hasModel(Lnet/minecraft/item/ItemStack;Lnet/minecraft/entity/EquipmentSlot;)Z"))
+	private static boolean permitArmorWithCustomRenderers(boolean hasModel, @Local ItemStack stack) {
+		return hasModel || ArmorRendererRegistryImpl.get(stack.getItem()) != null;
+	}
+}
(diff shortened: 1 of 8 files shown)


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import net.minecraft.client.render.entity.BipedEntityRenderer;
+import net.minecraft.item.ItemStack;
+
+import net.fabricmc.fabric.impl.client.rendering.ArmorRendererRegistryImpl;
+
+@Mixin(BipedEntityRenderer.class)
+abstract class BipedEntityRendererMixin {
+	@ModifyExpressionValue(method = "getEquippedStack", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/feature/ArmorFeatureRenderer;hasModel(Lnet/minecraft/item/ItemStack;Lnet/minecraft/entity/EquipmentSlot;)Z"))
+	private static boolean permitArmorWithCustomRenderers(boolean hasModel, @Local ItemStack stack) {
+		return hasModel || ArmorRendererRegistryImpl.get(stack.getItem()) != null;
```
modmuss50: Question: Is there a change in behaviour if an item previously had an armor renderer for a given item?

Juuxel: Well, the armor renderer wouldn't have been previously executed at all for the items this fixes. So the items would've had a registered but non-functional armor renderer.

Basically, vanilla filters which items it even captures in the biped render state based on which items have armor models as other items don't have to be rendered in vanilla. This mixin just changes it to also take Fabric armor renderers into account.

If the armor renderer did render previously, it'll naturally continue working.

modmuss50: Ok, perfect thanks for explaing it. It wasnt super clear at a glance.

modmuss50 (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import net.fabricmc.fabric.impl.client.rendering.ArmorRendererRegistryImpl;
+
+// Allows items with armor renderers to be passed to the armor feature in the first place.
+// Vanilla only stores the items with armor models in the entity's render state,
+// but we want to store any item with an armor renderer. Otherwise, armor renderers would
+// only be called for items with an existing vanilla armor model.
+@Mixin(BipedEntityRenderer.class)
+abstract class BipedEntityRendererMixin {
+	@ModifyExpressionValue(method = "getEquippedStack", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/feature/ArmorFeatureRenderer;hasModel(Lnet/minecraft/item/ItemStack;Lnet/minecraft/entity/EquipmentSlot;)Z"))
+	private static boolean permitArmorWithCustomRenderers(boolean hasModel, @Local ItemStack stack) {
```
LlamaLad7: Presumably the stack is the arg so use WrapOperation, not locals.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/LivingEntityRendererMixin.java
```diff
@@ -0,0 +1,43 @@
...
+import net.fabricmc.fabric.api.client.rendering.v1.ArmorRenderer;
+import net.fabricmc.fabric.impl.client.rendering.ArmorRendererRegistryImpl;
+
+@Mixin(LivingEntityRenderer.class)
+abstract class LivingEntityRendererMixin {
+	@ModifyExpressionValue(
+			method = "updateRenderState(Lnet/minecraft/entity/LivingEntity;Lnet/minecraft/client/render/entity/state/LivingEntityRenderState;F)V",
+			at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/feature/ArmorFeatureRenderer;hasModel(Lnet/minecraft/item/ItemStack;Lnet/minecraft/entity/EquipmentSlot;)Z")
+	)
+	private boolean toggleDefaultHeadItem(boolean skip, @Local(argsOnly = true) LivingEntity entity, @Local ItemStack headStack) {
```
LlamaLad7: Again use the arg rather than the local.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
Juuxel: Fixed.

PR #4695, opened 2025-06-11
Make inline loot table test a game test

Juuxel:
This also fixes its recursion and log spam.

In the process, I also cleaned up the other recursion guard to use a more readable class instead of a MutableBoolean guard, but that can be reverted if it's unwanted.

### DIFF
--- a/fabric-loot-api-v3/src/testmod/java/net/fabricmc/fabric/test/loot/LootGameTest.java
+++ b/fabric-loot-api-v3/src/testmod/java/net/fabricmc/fabric/test/loot/LootGameTest.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


fabric-loot-api-v3/src/testmod/java/net/fabricmc/fabric/test/loot/LootGameTest.java
```diff
@@ -0,0 +1,38 @@
...
+import net.fabricmc.fabric.api.gametest.v1.GameTest;
+
+public final class LootGameTest {
+	static int inlineLootTablesSeen = 0;
+
+	@GameTest
+	public void testInlineTableModifyDrops(TestContext context) {
+		int seenAtStart = inlineLootTablesSeen;
+		MinecraftServer server = context.getWorld().getServer();
+		server.getCommandManager().executeWithPrefix(server.getCommandSource(), "loot spawn 0 0 0 loot {\"pools\":[{\"entries\":[], \"rolls\":1.0}]}");
```
cputnam-a11y, line 32: Does this also deal with the queue issue?

Juuxel: Yes, since it never calls `/loot` from within loot (this is the only remaining call site for the command after I took out the original event listener).

(`executeWithPrefix` is just a helper for the parsing stuff it was doing before)

cputnam-a11y (approved):

modmuss50 (approved):
Looks good, thanks. Ill merge right away as its only test changes.

PR #4676, opened 2025-06-03
Special Gui Elements Api

kevinthegreat1:
Exposes a small api to add and modify special gui element renderers.

PR #4696, opened 2025-06-11
Special gui elements registry

modmuss50:
This is based on https://github.com/FabricMC/fabric/pull/4676 but takes a diffrent approach removing the need for an event. I wasnt too keen on the event for a few reasons:

- Its only called once, events are good for callbacks that need to be invoked many times
- It depends on class loading order/timing. What if the vanilla event is called too soon before mods have a chance to register with it.
- Being able to replace existing renderers is bad for mod compat, better using a mixin on the specific one imo.

Passing the context object to the interface allows us to easily expand what is passed as required in the future without making breaking API changes.


modmuss50:
@Copilot is useless as always :D

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/SpecialGuiElementRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+	public interface Factory {
+		SpecialGuiElementRenderer<?> createSpecialRenderer(Context ctx);
+	}
+
+	@ApiStatus.NonExtendable
+	public interface Context {
+		/**
+		 * @return the {@link VertexConsumerProvider.Immediate}.
+		 */
+		VertexConsumerProvider.Immediate immediate();
```
Juuxel: `vertexConsumers` would be more standard and clearer

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/SpecialGuiElementRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.render.SpecialGuiElementRenderer;
+import net.minecraft.client.render.VertexConsumerProvider;
+
+import net.fabricmc.fabric.impl.client.rendering.SpecialGuiElementRegistryImpl;
+
+/**
+ * Allows registering {@link SpecialGuiElementRenderer special gui element renderers},
```
Juuxel:
```suggestion
 * Allows registering {@linkplain SpecialGuiElementRenderer special gui element renderers},
```
`@linkplain` for plain text (can't copilot even catch this?)

mysticdrew (approved):
LGTM 🥇

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GameRendererMixin.java
```diff
@@ -0,0 +1,47 @@
...
+public class GameRendererMixin {
+	@Shadow
+	@Final
+	private GuiRenderer guiRenderer;
+
+	@Inject(method = "<init>", at = @At(value = "RETURN"))
+	private void initializeSpecialElements(MinecraftClient client, HeldItemRenderer firstPersonHeldItemRenderer, BufferBuilderStorage buffers, CallbackInfo ci, @Local VertexConsumerProvider.Immediate immediate) {
+		GuiRendererAccessor accessor = (GuiRendererAccessor) this.guiRenderer;
+		SpecialGuiElementRegistryImpl.onReady(client, immediate, accessor.getSpecialElementRenderers());
+	}
```
kevinthegreat1: I don't think we need this mixin if we have the `GuiRenderer` mixin. The only thing we get from this mixin is the client instance, which can be easily retrieved.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiRendererMixin.java
```diff
@@ -0,0 +1,48 @@
...
+abstract class GuiRendererMixin {
+	@Shadow
+	@Final
+	@Mutable
+	private Map<Class<? extends SpecialGuiElementRenderState>, SpecialGuiElementRenderer<?>> specialElementRenderers;
+
+	@Inject(method = "<init>", at = @At(value = "RETURN"))
+	private void mutableSpecialElementRenderers(GuiRenderState state, VertexConsumerProvider.Immediate vertexConsumers, List<SpecialGuiElementRenderer<?>> specialElementRenderers, CallbackInfo ci) {
+		this.specialElementRenderers = new IdentityHashMap<>(this.specialElementRenderers);
+	}
```
kevinthegreat1: I would just `@MEV` the `buildOrThrow` call but I guess this works too.

modmuss50: buildOrThrow specifically returns an ImmutableMap. As I moved this to a single mixin (based on your comment above) I think its fine for it to just be a single Inject.

PR #4706, opened 2025-06-14
Add `DrawItemStackOverlayCallback`

Fuzss:
Adds an event for rendering custom item stack decorations. In vanilla those are: durability bar, cooldown overlay and stack count.

The event runs for every item stack and provides that stack as a parameter.

Regarding performance concerns it is possible to introduce a more specialized event / registry in the future for specific items.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/DrawContextMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/DrawContextMixin.java
@@ -37,7 +37,7 @@ abstract class DrawContextMixin {
 	public void drawStackOverlay(TextRenderer textRenderer, ItemStack stack, int x, int y, @Nullable String stackCountText, CallbackInfo callback) {
 		if (!stack.isEmpty()) {
 			DrawItemStackOverlayCallback.EVENT.invoker()
-					.onDrawItemStackOverlay(DrawContext.class.cast(this), textRenderer, stack, x, y);
+					.onDrawItemStackOverlay((DrawContext) (Object) this, textRenderer, stack, x, y);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/DrawContextMixin.java
```diff
@@ -0,0 +1,43 @@
...
+@Mixin(DrawContext.class)
+abstract class DrawContextMixin {
+	@Inject(
+			method = "drawStackOverlay(Lnet/minecraft/client/font/TextRenderer;Lnet/minecraft/item/ItemStack;IILjava/lang/String;)V",
+			at = @At("RETURN")
+	)
+	public void drawStackOverlay(TextRenderer textRenderer, ItemStack stack, int x, int y, @Nullable String stackCountText, CallbackInfo callback) {
+		if (!stack.isEmpty()) {
+			DrawItemStackOverlayCallback.EVENT.invoker()
+					.onDrawItemStackOverlay(DrawContext.class.cast(this), textRenderer, stack, x, y);
```
LlamaLad7: Use `(DrawContext) (Object) this` rather than reflection

Fuzss: Sure. Any particular reason though?
Personally prefer reflection as it doesn't create a warning in IDE.

modmuss50 (approved):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/DrawContextMixin.java
Nikolay-Sitnikov: Reflection has a ongoing performance cost because the cast method checks the validity on every call. `(DrawContext) (Object) this` can be optimized away by the JNI when the bytecode is loaded.

cputnam-a11y: To be fair, reflection on constable objects can too. The general convention in fapi (and mixin in general) is to double cast, so that's the main reason.
(Although I would think jit would have more room with the casts. no backing proof though)

PR #4709, opened 2025-06-16
Add way to get an instance of WorldRenderContext if you only have a WorldRenderer

Earthcomputer:
This is especially useful if a mod wants to create their own custom render event, but could also be useful more generally in case someone wants to access stuff from this context but doesn't have direct access to that context for some reason.

Note that even in situations where you don't strictly require a `WorldRenderContext`, it can still be useful to have one as it captures local variables from inside `WorldRenderer` which are otherwise difficult to access.


cputnam-a11y (approved):

modmuss50 (approved):
Great idea. 👍

PR #4720, opened 2025-06-24
Replace inefficient implementations

tygyh:


cputnam-a11y:
This pr could probably use a slightly longer description. On another note, have you had trouble with the existing implementation's performance, or is this just a preventive optimization?

modmuss50:
Sorry, looks like a suspcious account.

tygyh:
What is suspicious about my account?

modmuss50:
> What is suspicious about my account?

You have many low effort PRs against unrelated repos. We have had issues with people trying to game the system in similar ways. This change here has 0 real benefit at all, its in a code path that is rarely used.

Im more than happy to review/help with a PR that makes a meaningful change (there are loads of issues if you want a starting point). Picking a tottaly random 1 line change isnt the way to go without any proof (a profile) that its causing issues isnt the way to go.

PR #4722, opened 2025-06-25
Add lava chicken music disk to music disks tag.

modmuss50:
Also means we arent backwards compatible with 1.21.6, not the end of the world.


Patbox:
Couldn't it be added as optional?

modmuss50:
> Couldn't it be added as optional?

I tried with `addOptional` either its broken or im stupid but it doesnt seem to make any difference.

Patbox:
Maybe using this.builder(ConventionalItemTags.MUSIC_DISCS).addOptional(Registries.ITEM.getKey(Items.MUSIC_DISC_LAVA_CHICKEN)) in seperate line?

modmuss50:
> Maybe using this.builder(ConventionalItemTags.MUSIC_DISCS).addOptional(Registries.ITEM.getKey(Items.MUSIC_DISC_LAVA_CHICKEN)) in seperate line?

Same thing, looks like there is a bug somewhere....

modmuss50:
Going to merge this, there might be a bug here that is worth looking into. Thinking about it im not too keen on having to make code changes to support older versions, I dont think many people will care about .6 for long so its not a big loss.

PR #4725, opened 2025-06-26
Fix getRecipeIdentifier applying twice to advancement paths

Juuxel:
Fixes a bug introduced in #4703.

The advancement's path is already modified by the getRecipeIdentifier call in each recipe JSON builder mixin, so this applied getRecipeIdentifier twice to a single ID (and to different parts).

For example, this getRecipeIdentifier implementation:
```java
@Override
protected Identifier getRecipeIdentifier(Identifier identifier) {
    return super.getRecipeIdentifier(identifier).withPrefixedPath("foo/");
}
```
would have resulted in an advancement being placed at `data/mod_id/advancement/foo/recipes/.../foo/the_item.json`, which is clearly wrong. With this fix, only the last "foo/" will appear.

---

Can be backported cleanly to any version with #4703 also applied. The 1.21.1 backport is already contained in #4724.


modmuss50 (approved):

modmuss50:
Note to self: needs cherry picking into .5, .6 and .7

PR #4730, opened 2025-06-27
Add render entity callback

michiel1106:
An event that is called before the entity gets rendered, containing the entity that is to be rendered. It would be useful for if you want to do anything with checking if an entity is actually being rendered, doing something with a specific entity only if its being rendered etc.


cputnam-a11y:
I think this would be solved by the hypothetical renderstate API that kind of fuzzeled

cputnam-a11y:
We'll see what ~~other~~(I suppose I'm not a maintainer lol) the maintainers say, but this also seems like an event that falls into the same category as entity tick

modmuss50:
Yeah, I dont think this is really what we want for entities, we should aim to follow what vanilla does with the render states. Accessing the the actual `Entity` instance when rendering isnt correct and will explode when mojang split the state building and rendering onto two threads. Finding a good solution to this in FAPI is possible but needs some careful work.

For now Mixin is a great tool for being able to expand vanilla's render states as you can easily add new fields with little overhead.

Thanks for the PR, I'd be happy to discuss a better long term solution for FAPI or a short term solution for your mod.

PR #4735, opened 2025-06-28
Fix #4732

### LINKED ISSUE #4732: `InvalidMixinException` from `GameRulesIntRuleAccessor` overwrite
`GameRulesIntRuleAccessor` doesn't use method name prefixes, and `IntGameRule` has a `getValue` method now, so the new `@Overwrite` requirement results in an error:
> Caused by: org.spongepowered.asm.mixin.transformer.throwables.InvalidMixinException: getValue()I in fabric-game-rule-api-v1.mixins.json:GameRulesIntRuleAccessor from mod fabric-game-rule-api-v1 cannot overwrite method in net.minecraft.world.GameRules$IntGameRule because @Overwrite is required by the parent configuration [ -> ACCESSOR Applicator Phase -> fabric-game-rule-api-v1.mixins.json:GameRulesIntRuleAccessor from mod fabric-game-rule-api-v1 -> Apply Accessors]

I get this error in one of my mods that uses `GameRuleRegistry::register`

supersaiyansubtlety:
removes `GameRulesIntRuleAccessor` since `IntRule` has public `get`/`set` methods


cputnam-a11y:
oops

cputnam-a11y:
okay
I'm not touching it anymore

cputnam-a11y:
too buggy

cputnam-a11y (approved):

modmuss50 (approved):
Looks like a reasonable change, many thanks.

PR #4740, opened 2025-07-07
Event handling for custom click actions

TheDeathlyCow:
Implements a basic system for registering event listeners to custom click actions. Events are registered by the action ID, so that mods don't have to parse that out themselves. Also uses an event system rather than a map so that multiple mods can listen to the same click events. 

Will add javadoc and tests soon:tm:


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickActions.java
```diff
@@ -0,0 +1,18 @@
+package net.fabricmc.fabric.api.networking.v1;
+
+import java.util.Objects;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.impl.networking.CustomClickEventHandlerRegistry;
+
+import net.minecraft.util.Identifier;
+
+public final class CustomClickActions {
```
modmuss50: I would just create a file `CustomClickActionsEvents.java`, with the interface being an inner class of what you have in `CustomClickActionListener` with the single static method in the parent method.

A good example might be:
https://github.com/FabricMC/fabric/blob/de8b3290191d806c7dacc1970e763d863a3888c9/fabric-item-group-api-v1/src/main/java/net/fabricmc/fabric/api/itemgroup/v1/ItemGroupEvents.java#L53-L65

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ServerCommonNetworkHandlerMixin.java
```diff
@@ -55,4 +69,21 @@ private void onPlayPong(CommonPongC2SPacket packet, CallbackInfo ci) {
...
 		}
 	}
+
+	@WrapMethod(method = "onCustomClickAction")
+	protected void overrideCustomClickAction(CustomClickActionC2SPacket packet, Operation<Void> original) {
+		original.call(packet);
+	}
+
+	@Mixin(ServerPlayNetworkHandler.class)
+	private abstract static class ServerPlayNetworkHandlerMixin extends ServerCommonNetworkHandlerMixin {
```
modmuss50: We already have a ServerPlayNetworkHandlerMixin as its own class, just reuse that. However this should also handle the packets being set during configuration.

modmuss50 (commented):
Needs build fixes and tests.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ServerCommonNetworkHandlerMixin.java
TheDeathlyCow: I moved this to the channel network addons. Would you prefer the base method be abstract or does it not particularly matter?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickActions.java
TheDeathlyCow: Done. I also added a context object like you mentioned on Discord, but that can be easily switched out if it's too much.

TheDeathlyCow:
Ok I think this is ready for review now, feel free to roast. I added some tests but my approach for the configuration phase had to be a little weird because of the client tests. There isnt, to my knowledge, a good way to 'manually' request a dialog during the configuration phase, so a mandatory dialog would show up on world join that would block the E2E client tests from being able to run. To get around this, I implemented a mock show dialog packet that mimics the behavior of the regular `ShowDialogS2CPacket`. When a client receives this packet, rather than displaying a dialog like usual it will just automatically return the regular custom click event response packet that should get handled by the event receiver (so it will be as if a user pressed the custom click event button). It's a little weird, but it works. The play click event test uses a regular multi option dialog and can be displayed with the command `/networktestcommand testdialog`. 

As far as I can tell, the networking API does not use automated tests, so I didn't implement them here. Though it does not seem really easily possible at this time to do that using existing frameworks (`fabric-integration-test-api` when?).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraft.nbt.NbtElement;
+import net.minecraft.server.network.ServerCommonNetworkHandler;
+import net.minecraft.server.network.ServerConfigurationNetworkHandler;
+import net.minecraft.server.network.ServerPlayNetworkHandler;
+
+/**
+ * Contains data about a {@linkplain net.minecraft.text.ClickEvent.Custom custom click event} when one is received on
+ * the server.
+ */
+public sealed interface CustomClickEventContext permits CustomClickEventContext.Play, CustomClickEventContext.Configuration {
```
modmuss50, line 30: Nit: Maybe move this into the other class?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * The context data when a custom click event is received during the PLAY phase on the server.
+	 */
+	@ApiStatus.NonExtendable
+	non-sealed interface Play extends CustomClickEventContext {
+		/**
+		 * The play handler responsible for the event.
+		 */
+		ServerPlayNetworkHandler handler();
```
modmuss50, line 45: @Override

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * The context data when a custom click event is received during the CONFIGURATION phase on the server.
+	 */
+	@ApiStatus.NonExtendable
+	non-sealed interface Configuration extends CustomClickEventContext {
+		/**
+		 * The configuration handler responsible for the event.
+		 */
+		ServerConfigurationNetworkHandler handler();
```
modmuss50, line 67: @Override

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,64 @@
...
+	 * The payload received with this event. If no payload is received, then this payload will be {@code null}.
+	 */
+	@Nullable
+	NbtElement payload();
+
+	/**
+	 * The context data when a custom click event is received during the PLAY phase on the server.
+	 */
+	@ApiStatus.NonExtendable
+	non-sealed interface Play extends CustomClickEventContext {
```
modmuss50, line 40: You could add a default method to get the player from the handler. I expect most people will want the player just makes it a bit easier.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickActionEvents.java
```diff
@@ -0,0 +1,68 @@
...
+public final class CustomClickActionEvents {
+	/**
+	 * Gets an event that is invoked on the server when a custom click event is received during the PLAY phase. The
+	 * returned event will only be invoked when a click event is received with the given ID.
+	 *
+	 * @param id The of the ID click event to listen to.
+	 * @return Returns an event that will be invoked when a click event with the given ID is received during the PLAY
+	 * phase.
+	 */
+	public static Event<ClickActionReceived<CustomClickEventContext.Play>> playClickActionEvent(Identifier id) {
```
modmuss50: The idea of having the single context interface with the 2 sealed child interfaces was to allow you to have a single event for both play and configuration. The event would only pass `CustomClickEventContext`, the handler would then pattern match or instance check to decide what to do.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/CustomClickActionsRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.fabricmc.fabric.api.networking.v1.CustomClickActionEvents;
+import net.fabricmc.fabric.api.networking.v1.CustomClickEventContext;
+
+public final class CustomClickActionsRegistry<T extends CustomClickEventContext> {
+	public static final CustomClickActionsRegistry<CustomClickEventContext.Play> PLAY_REGISTRY = new CustomClickActionsRegistry<>();
+	public static final CustomClickActionsRegistry<CustomClickEventContext.Configuration> CONFIGURATION_REGISTRY = new CustomClickActionsRegistry<>();
```
modmuss50: As above, my idea was to have a single event, happy to discuss if you disagree.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * The handler responsible for the event.
+	 */
+	ServerCommonNetworkHandler handler();
+
+	/**
+	 * The payload received with this event. If no payload is received, then this payload will be {@code null}.
+	 */
+	@Nullable
+	NbtElement payload();
```
modmuss50: Vanilla has it as an option, why dont we keep the same?

Would it also be better passed directly into the event as a param, it feels quite imporant to me.

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/configuration/NetworkingConfigurationTest.java
```diff
@@ -73,6 +80,24 @@ public void onInitialize() {
 
 		// Enable the vanilla debugconfig command
 		CommandRegistrationCallback.EVENT.register((dispatcher, registryAccess, environment) -> DebugConfigCommand.register(dispatcher, registryAccess));
+
+		ServerConfigurationConnectionEvents.CONFIGURE.register((handler, server) -> {
+			// open a dialog screen on the client
+			// note: normally you would actually use ShowDialogS2CPacket, a mock packet is used here for testing purposes
+			// important: use a task to prevent this dialog from being quickly skipped over
+			handler.addTask(new TestDialogConfigurationTask(CONFIGURATION_TEST_DIALOG));
```
modmuss50: This doesnt necessarily need to be automated, you could maybe use a system property to enable the addional dialog. I think there would be more value in having an opt-in test that that uses a real dialog vs hacking something that kinda tests it.

Having said that, it does excerise the code all the time if you think this is best id be happy to go with what you have here.

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/configuration/NetworkingConfigurationTest.java
```diff
@@ -73,6 +80,24 @@ public void onInitialize() {
...
+		});
+
+		CustomClickActionEvents.configurationClickActionEvent(NetworkingTestmods.id("configuration_event")).register(
```
modmuss50: You could maybe move all the custom click test code into their own package + class. Especially if you merge them into a single even it will make more sense. The actual test code looks good though.

Id keep using the /networktestcommand though, there is scope to move that at a later date.

modmuss50 (commented):
This is looking really good, I think my main worry is around the confusion of CustomClickEventContext. I had in mind a single event where you get passed the `CustomClickEventContext` and are forced to handle if its play or configuration in your handler. The fact that this has confused you likely means it will also confuse people trying to use this API. So maybe the approach of 2 events as you have here is better, do let me know what you think, dont take what I say as a rule.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
TheDeathlyCow: Sure I can do that, style guidelines just say to avoid Optional where possible. But it is probably better to use an Optional here anyway.

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/configuration/NetworkingConfigurationTest.java
TheDeathlyCow: What about a custom gamerule? Could theoretically be set during world creation to be automatically tested by a client test. I don't know if it's possible to assert that the server received a packet in a client test, though.

haykam821: Could you not have a command to reenter configuration and then show the dialog?

TheDeathlyCow:
I think having a single event and pattern matching two different parameters depending on where it came from feels like an unusual departure from the usual style of events in Fabric. There are also some differences in how custom dialogs need to be set up if they're used in the configuration phase (ie all references must be in-line, as no registries are available). So there's already gotta be special handling for configuration phase dialogs by the developer. 

At the same time, it does feel like a bit of a low level concern for something that should probably be a better high level abstraction. And having two separate events might lead to excessive code duplication for use cases that only care about the payload and not the handler (which evidently seems to be mostly what Mojang had in mind when they created this click event). Maybe we could add an `Optional<ServerPlayerEntity>` parameter, in addition to the handler, that's only available as a shorthand for when the packet is received in the play phase. This might be a good compromise that doesn't involve needing to do pattern matching (in most cases) or sort out too much low level networking stuff. 

I might need to spend a bit more time thinking about it and get some more input lol.

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/configuration/NetworkingConfigurationTest.java
TheDeathlyCow: I am not the most familiar with networking but that's a possibility that I can try out (I assume that's what the reconfigure test command does?).

TheDeathlyCow:
I've decided to merge them into a single event, based on what I said earlier about preventing code duplication. The context instanceof checking is weird, but at least with switch statements its not that bad (see tests). I've also moved the tests into a single package and class, which shows how to do the switch-casing and shows an actual dialog using a reconfiguration command.

Finally, I added a new event that is invoked for *any* custom click action received - and includes the ID. I'm not sure what the use case is for it exactly, but it's something I could see someone asking for down the line. If you think it's too unnecessary, then I can remove it.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
TheDeathlyCow: Personally, I think its better in its own class, less typing.

modmuss50: Ok, fair enough. I agree especially as it has its own inner classes.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickActionEvents.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Events for listening to {@linkplain net.minecraft.text.ClickEvent.Custom custom click actions}, such as those invoked
+ * from a custom dialog.
+ */
+public final class CustomClickActionEvents {
+	/**
+	 * Invoked when any custom click action is received. If you are only interested in listening to events with a
+	 * specific ID, use {@link #customClickActionReceivedEvent(Identifier)}.
+	 */
+	public static final Event<CustomClickActionReceived> ON_ANY_CUSTOM_CLICK_ACTION_RECEIVED = EventFactory.createArrayBacked(
```
modmuss50: What is the example use case for this? Generally it seems like you should only care about your own specific click actions?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	ServerCommonNetworkHandler handler();
+
+	/**
+	 * The player entity responsible for the event, if in the play phase.
+	 */
+	Optional<ServerPlayerEntity> player();
+
+	/**
+	 * The payload received with this event. If no payload is received, then this payload will be {@code null}.
```
modmuss50: How can it be null if its an Optional?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,82 @@
...
+public sealed interface CustomClickEventContext permits CustomClickEventContext.Play, CustomClickEventContext.Configuration {
+	/**
+	 * The handler responsible for the event.
+	 */
+	ServerCommonNetworkHandler handler();
+
+	/**
+	 * The player entity responsible for the event, if in the play phase.
+	 */
+	Optional<ServerPlayerEntity> player();
```
modmuss50: No, this shouldnt be here, it should be `ServerPlayerEntity player()` only in the Play context. The idea of using the 2 inner classes allows you to have type safety over what objects are valid for each context.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * The payload received with this event. If no payload is received, then this payload will be {@code null}.
+	 */
+	Optional<NbtElement> payload();
```
modmuss50: I wonder if this would be better passed in via a param to the event and not via the context object. That could then allow you to reuse the context objects between event invocations.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
```diff
@@ -0,0 +1,82 @@
...
+		 */
+		@Override
+		ServerConfigurationNetworkHandler handler();
+
+		/**
+		 * The configuration phase is too early for an entity to have been created, so an entity is never returned.
+		 *
+		 * @return Returns an empty optional.
+		 */
+		default Optional<ServerPlayerEntity> player() {
```
modmuss50: As above, remove there shouldnt be anything related to the player during config.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractNetworkAddon.java
```diff
@@ -172,4 +174,7 @@ public final void handleDisconnect() {
 	 * @return whether the channel is reserved
 	 */
 	protected abstract boolean isReservedChannel(Identifier channelName);
+
+	public void invokeCustomClickActionEvent(Identifier id, Optional<NbtElement> payload) {
```
modmuss50, line 178: Should the default impl for this throw? Something is wrong if this is called but not overriden.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/CustomClickActionsRegistry.java
```diff
@@ -0,0 +1,81 @@
...
+import net.fabricmc.fabric.api.networking.v1.CustomClickActionEvents;
+import net.fabricmc.fabric.api.networking.v1.CustomClickEventContext;
+
+public final class CustomClickActionsRegistry {
+	private static final Map<Identifier, Event<CustomClickActionEvents.NamedCustomClickActionReceived>> REGISTRY = new HashMap<>();
+
+	public static Event<CustomClickActionEvents.NamedCustomClickActionReceived> getOrCreateActionEvent(Identifier id) {
+		return REGISTRY.computeIfAbsent(
+				id,
+				idx -> {
```
modmuss50: Nit: Id almost abstract this out into a method to create the event, will be a little easier to read.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/CustomClickActionsRegistry.java
```diff
@@ -0,0 +1,81 @@
...
+		CustomClickActionEvents.ON_ANY_CUSTOM_CLICK_ACTION_RECEIVED.invoker().handleCustomClickAction(id, context);
+
+		Event<CustomClickActionEvents.NamedCustomClickActionReceived> event = REGISTRY.get(id);
+
+		if (event != null) {
+			event.invoker().handleCustomClickAction(context);
+		}
+	}
+
+	public record PlayContextImpl(
```
modmuss50: Nit: Id maybe move these into their respective network addons. See how ServerConfigurationNetworkAddon already has a context.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/SignBlockEntityMixin.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.nbt.NbtElement;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.networking.CustomClickActionsRegistry;
+
+@Mixin(SignBlockEntity.class)
+public class SignBlockEntityMixin {
+	@WrapOperation(
```
modmuss50, line 40: Why a wrap operation and not just an inject?

modmuss50 (commented):
Looking good, just a few minor questions/points.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/SignBlockEntityMixin.java
cputnam-a11y: an `Inject` would require capturing more locals

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/AbstractNetworkAddon.java
TheDeathlyCow: There are some other methods in the addon that are just declared as abstract but have empty implementations (like `handleUnregistration`).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickEventContext.java
TheDeathlyCow: That can be done, though I doubt there would be any significant savings from removing one occasional (very small) object allocation vs caching contexts for a long period.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/CustomClickActionEvents.java
TheDeathlyCow: I had a usecase where I tracked where a user was in a particular flow using the ID of the event. But that is probably not a good way to handle that, adding to payload is better (eg with the `additions` field). Will remove.

PR #4745, opened 2025-07-11
Add extra utility methods to Read/Write View (Fabric Serialization API)

Patbox:
Adds new module (Fabric Serialization API), which currently handles:
- Extra ReadView methods (FabricReadView):
  - Collection<String> keys() - returns all available keys.
  - boolean contains(String key) - checks if field exists.
  - Optional<long[]> getOptionalLongArray(String key) - returns a long array.
  - Optional<byte[]> getOptionalByteArray(String key) - returns a byte array.
- Extra WriteView methods (FabricWriteView):
  - void putLongArray(String key, long[] value) - puts a long array
  - void putByteArray(String key, byte[] value) - puts a byte array
- Additional codecs (MoreCodecs):
  - Codec for Long Array, Byte Array
Read/Write View extensions are defaulted with codecs, allowing them to work with any custom class implementing Read/Write View, while having more direct implementations for vanilla nbt based view.


Earthcomputer:
We could also do with a method to add stuff to a record codec, although maybe this could go in a separate PR

Patbox:
Yeah that would fit into this module, but as it's own pr

cputnam-a11y:
Potentially also attaching context to dynamic ops

fabric-serialization-api-v1/src/main/java/net/fabricmc/fabric/api/serialization/v1/view/FabricReadView.java
```diff
@@ -0,0 +1,47 @@
...
+
+import net.minecraft.storage.ReadView;
+
+import net.fabricmc.fabric.impl.serialization.SpecialCodecs;
+
+/**
+ * Fabric provided extension of ReadView.
+ */
+public interface FabricReadView {
+	default Collection<String> keys() {
```
modmuss50, line 38: I know it seems silly but id put some javadocs on these methods. Put a note saying that it returns an empty list or throws in the negative case.

fabric-serialization-api-v1/src/main/java/net/fabricmc/fabric/api/serialization/v1/view/FabricReadView.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+import net.minecraft.storage.ReadView;
+
+import net.fabricmc.fabric.impl.serialization.SpecialCodecs;
+
+/**
+ * Fabric provided extension of ReadView.
```
modmuss50, line 28: Add a comment saying that this is injected, is what we tend to use:

`<p>Note: This interface is automatically implemented on all items via Mixin and interface injection.`

fabric-serialization-api-v1/src/testmod/java/net/fabricmc/fabric/test/serialization/SerializationExtensionTest.java
```diff
@@ -0,0 +1,108 @@
...
+import net.minecraft.registry.RegistryWrapper;
+import net.minecraft.storage.NbtReadView;
+import net.minecraft.storage.NbtWriteView;
+import net.minecraft.storage.ReadView;
+import net.minecraft.storage.WriteView;
+import net.minecraft.util.ErrorReporter;
+
+import net.fabricmc.api.ModInitializer;
+
+public class SerializationExtensionTest implements ModInitializer {
```
modmuss50: This is likely better suited as a junit unit test. It should be easy to change it to one, let me know if you need help.

fabric-serialization-api-v1/src/testmod/java/net/fabricmc/fabric/test/serialization/DelegateReadView.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.Optional;
+
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.MapCodec;
+
+import net.minecraft.registry.RegistryWrapper;
+import net.minecraft.storage.ReadView;
+
+public record DelegateReadView(ReadView view) implements ReadView {
```
modmuss50: A comment on saying why this is needed might be helpful, I think I understand why (to test the none NBT mixin/codecs). It might not so clear in the future.

gradle.properties
```diff
@@ -48,6 +48,7 @@ fabric-resource-conditions-api-v1-version=5.0.24
 fabric-resource-loader-v0-version=3.1.11
 fabric-screen-api-v1-version=2.1.0
 fabric-screen-handler-api-v1-version=1.3.136
+fabric-serialization-api-v1-version=1.0.0
```
modmuss50, line 51: I think adding a new module is fine, I dont see any existing module that this would make sense for. I think its generic enough that more functionality can be added in the future. Im just making this comment to see if anyone else has thoughts on this as we are trying to be careful about the number of modules.

fabric-serialization-api-v1/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+
+loom {}
+
+moduleDependencies(project, [
+	':fabric-api-base',
+])
+
+testDependencies(project, [
+	':fabric-lifecycle-events-v1',
+])
```
modmuss50: Is this dep needed?

fabric-serialization-api-v1/build.gradle
```diff
@@ -0,0 +1,11 @@
+version = getSubprojectVersion(project)
+
+loom {}
```
modmuss50: Remove this.

fabric-serialization-api-v1/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+
+moduleDependencies(project, [
+	':fabric-api-base',
+])
```
modmuss50: Is this used?

modmuss50 (commented):
Looks good overall, just a few minor points.

fabric-serialization-api-v1/build.gradle
Patbox: To provide a codec for TriState, but honestly that might be moved directly to base module

modmuss50: Ah thats fine, I tottaly missed that. Id leave it as-is.

modmuss50: Alghough having said that having TriState.CODEC would be more idiomatic.

cputnam-a11y:
Generally, a lot of these codecs seem to throw, which violates the contract. For the normal Codecs, they should probably be wrapped in Codecs.exceptionCatching, or flatMap properly. I don't know if there is a similar utility for MapCodecs, but they shouldn't really throw either

fabric-serialization-api-v1/build.gradle
Patbox: Just a leftover yeah, originally wanted to wait for server started so registries were ready, but that wasn't needed

Patbox:
They should never throw, as they always should return correct data or are just checked with other methods before

fabric-serialization-api-v1/build.gradle
Patbox: This would make sense, but adds extra dependencies to it (through would be useful either way)

Patbox:
Won't be able to apply any code changes until the end of the week, as I'm currently on holidays with bit spotty (and slow) internet.

fabric-serialization-api-v1/src/testmod/java/net/fabricmc/fabric/test/serialization/DelegateWriteView.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public record DelegateWriteView(WriteView view) implements WriteView {
+	@Override
+	public <T> void put(String key, Codec<T> codec, T value) {
+		view.put(key, codec, value);
+	}
+
+	@Override
+	public <T> void putNullable(String key, Codec<T> codec, @Nullable T value) {
+		view.put(key, codec, value);
```
cputnam-a11y: should this not delegate to the same method on `view`?
```suggestion
		view.putNullable(key, codec, value);
```

Patbox:
Moved the test from testmod to unit, also TriState codec was moved to it's own pr https://github.com/FabricMC/fabric/pull/4766

fabric-serialization-api-v1/src/main/java/net/fabricmc/fabric/api/serialization/v1/view/FabricReadView.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * <p>Note: This interface is automatically implemented on all items via Mixin and interface injection.
```
modmuss50: > on all items

This isnt an item :P

fabric-serialization-api-v1/src/main/java/net/fabricmc/fabric/api/serialization/v1/view/FabricWriteView.java
```diff
@@ -0,0 +1,36 @@
...
+package net.fabricmc.fabric.api.serialization.v1.view;
+
+import net.minecraft.storage.WriteView;
+
+import net.fabricmc.fabric.impl.serialization.SpecialCodecs;
+
+/**
+ * Fabric provided extension of WriteView.
+ *
+ * <p>Note: This interface is automatically implemented on all items via Mixin and interface injection.
```
modmuss50: same here

modmuss50 (approved):
2 copy/paste typos that can be fixed before merging.

isXander (approved):

PR #4746, opened 2025-07-13
Add FabricItem#getCreatorNamespace

Shnupbups:
Requested on Discord, adds `FabricItem#getCreatorNamespace(ItemStack stack)`, as well as `FabricItemStack#getCreatorNamespace()`, for a stack-aware method of getting the namespace of the mod that added an item.

This could be useful for things like Patchouli guidebooks, where Patchouli can add a single `patchouli:guide_book` item which other mods can then use to make their own variants of. Patchouli can then use this method to say that the guide book item's creator namespace is that of the mod that guidebook comes from, instead of Patchouli itself - which can then be reflected in things like recipe viewers, item filters, etc. It can also be useful for things like Polymer, which use vanilla's item IDs for creation of new items on the server-side.

Its default implementation is to just return the namespace of the item's own ID, as you would expect - except in the cases of Potions or Enchanted Books, in which case it returns the namespace of the potion contents (if it's a registered potion type) or enchantment (if only one enchantment is present).

For modders, use of the method will be recommended instead of checking the item's ID namespace when determining what mod added an item as far as a player is concerned.

To show off its functionality, the testmod now adds the name of the mod an item comes from in tooltips. Also included is a testmod item which uses the method to change the namespace based on a component, showing off its stack aware functionality.

### DIFF
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
@@ -127,7 +127,7 @@ public interface FabricItem {
 	}
 
 	/**
-	 * Gets the namespace of the mod that created this item.
+	 * Gets the namespace of the mod or datapack that created this item.
 	 *
 	 * <p>This can be used if, for example, a library mod registers a generic item that other mods can create new
 	 * variants for, allowing those mods to take credit for those variants if a player wishes to know what mod they
@@ -138,6 +138,9 @@ public interface FabricItem {
 	 *
 	 * <p>Defaults to the namespace of the item's own registry entry, except in the cases of potions or enchanted books,
 	 * in which it uses the namespace of the potion contents or single enchantment applied.</p>
+	 *
+	 * <p>Note that while it is recommended that this reflect a namespace and/or mod ID, it can technically be any
+	 * arbitrary string.</p>
 	 *
 	 * @param stack the current stack
 	 * @return the namespace of the mod that created the item
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
@@ -153,7 +153,7 @@ public interface FabricItem {
 			if (potion.isPresent()) entry = potion.get();
 		} else if (stack.isOf(Items.ENCHANTED_BOOK) && stack.contains(DataComponentTypes.STORED_ENCHANTMENTS)) {
 			Set<RegistryEntry<Enchantment>> enchantments = stack.get(DataComponentTypes.STORED_ENCHANTMENTS).getEnchantments();
-			if (enchantments.size() == 1) entry = enchantments.stream().findFirst().get();
+			if (enchantments.size() == 1) entry = enchantments.iterator().next();
 		}
 
 		return entry.getKey().orElseThrow().getValue().getNamespace();
--- a/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
+++ b/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
@@ -16,10 +16,17 @@
 
 package net.fabricmc.fabric.api.item.v1;
 
+import java.util.Optional;
+import java.util.Set;
+
+import net.minecraft.component.DataComponentTypes;
 import net.minecraft.enchantment.Enchantment;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.item.PotionItem;
+import net.minecraft.potion.Potion;
 import net.minecraft.registry.entry.RegistryEntry;
 import net.minecraft.util.Hand;
 import net.minecraft.util.Identifier;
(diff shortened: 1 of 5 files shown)


SquidDev:
Thank you so much! I've been meaning to PR this for years.

I don't know if it's worth providing an implementation for vanilla items (NF does potions, spawn eggs, and enchanted books with a single enchantment) as well?

modmuss50:
Is there a reason it's not specifically a modid?

Patbox:
I guess functionally it could return any string, that isn't connected to a mod.
One example where that would be intended is if content was added by a datapack.

Also something worth considering. Would it make sense to have few more default implementations of getCreatorNamespace method?
For example in case of the enchantment books or potions.
Alternatively would it make sense to expose event overriding that, so other mods could implement something like this.

Favouriteless:
I recommend renaming the method to `getCreatorModId` to match NF's method of the same purpose. 

Just means multiloader devs can soft implement both in common without having to duplicate it

Patbox:
They still can soft implement both in common without any issues. Java's \@\Override is just a syntax sugar after all (they could define the method and it would be called correctly)

Favouriteless:
> They still can soft implement both in common without any issues. Java's @\Override is just a syntax sugar after all (they could define the method and it would be called correctly)

Well, I meant to avoid having to soft implement 2 methods with different names but the same function. Just a very small nitpick though

cputnam-a11y:
> > They still can soft implement both in common without any issues. Java's @\Override is just a syntax sugar after all (they could define the method and it would be called correctly)
> 
> Well, I meant to avoid having to soft implement 2 methods with different names but the same function. Just a very small nitpick though

One can delegate to the other. I would argue that there are bigger discrepancies already

SquidDev:
> I would argue that there are bigger discrepancies already

I'm so sorry about `FabricSoundInstance.getAudioStream` vs NF's `getStream`. Still irks me to this day.

Shnupbups:
I've added default implementations for Potions and Enchanted Books - not sure if any others are really needed (Spawn Eggs, for example, have each been their own individual item IDs since 1.13 so shouldn't really be necessary).

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
```diff
@@ -61,4 +61,23 @@ default boolean canBeEnchantedWith(RegistryEntry<Enchantment> enchantment, Encha
 		);
 		return result.orElseGet(() -> ((ItemStack) this).getItem().canBeEnchantedWith((ItemStack) this, enchantment, context));
 	}
+
+	/**
+	 * Gets the namespace of the mod that created this item.
```
modmuss50: mod or datapack

as discussed in the comments this namespace doesnt need to have a matching registered mod. I think its worth making a note that it doesnt have to be a registered mod.

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -118,6 +126,36 @@ default boolean canBeEnchantedWith(ItemStack stack, RegistryEntry<Enchantment> e
...
+	default String getCreatorNamespace(ItemStack stack) {
+		RegistryEntry<?> entry = stack.getRegistryEntry();
+
+		if ((this instanceof PotionItem || this instanceof TippedArrowItem) && stack.contains(DataComponentTypes.POTION_CONTENTS)) {
+			Optional<RegistryEntry<Potion>> potion = stack.get(DataComponentTypes.POTION_CONTENTS).potion();
+			if (potion.isPresent()) entry = potion.get();
+		} else if (stack.isOf(Items.ENCHANTED_BOOK) && stack.contains(DataComponentTypes.STORED_ENCHANTMENTS)) {
+			Set<RegistryEntry<Enchantment>> enchantments = stack.get(DataComponentTypes.STORED_ENCHANTMENTS).getEnchantments();
+			if (enchantments.size() == 1) entry = enchantments.stream().findFirst().get();
+		}
```
modmuss50: This is fine for now, but is there ever going to be a case where Potion/Enchantment may need its own `getCreatorNamespace` method? I imagine this could also apply to a vast array of other objects such as entities.

I think for now this is a good start and doesnt get int the way of expanding this later down the line, maybe to cover all RegistryEntry's?

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItemStack.java
Shnupbups: I've added some notes to the javadoc that should address this

modmuss50 (approved):

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
Shnupbups: I doubt many, if any, other things will benefit as much as items do from a method like this, but I suppose it's a possibility. Regardless, I think anything beyond what's been done is outside the scope of this PR.

modmuss50: Yes, nothing for this PR was just thinking ahead. This is good for now 👍

fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
```diff
@@ -118,6 +126,39 @@ default boolean canBeEnchantedWith(ItemStack stack, RegistryEntry<Enchantment> e
...
+	 */
+	default String getCreatorNamespace(ItemStack stack) {
+		RegistryEntry<?> entry = stack.getRegistryEntry();
+
+		if ((this instanceof PotionItem || this instanceof TippedArrowItem) && stack.contains(DataComponentTypes.POTION_CONTENTS)) {
+			Optional<RegistryEntry<Potion>> potion = stack.get(DataComponentTypes.POTION_CONTENTS).potion();
+			if (potion.isPresent()) entry = potion.get();
+		} else if (stack.isOf(Items.ENCHANTED_BOOK) && stack.contains(DataComponentTypes.STORED_ENCHANTMENTS)) {
+			Set<RegistryEntry<Enchantment>> enchantments = stack.get(DataComponentTypes.STORED_ENCHANTMENTS).getEnchantments();
+			if (enchantments.size() == 1) entry = enchantments.stream().findFirst().get();
```
Chocohead: You could use `enchantments.iterator().next()` or [`Iterables.getOnlyElement(enchantments)`](https://javadoc.io/static/com.google.guava/guava/33.3.1-jre/com/google/common/collect/Iterables.html#getOnlyElement(java.lang.Iterable)) as a (debatably) prettier alternative to streaming for an `Optional`

PR #4755, opened 2025-07-20
Allow command suggestions to autofill identifiers not under "minecraft" namespace

Ramixin:
# Issue
When typing commands, the autofill suggestions will only be able to autofill paths if they are under the `minecraft` namespace. This can be seen when typing `/give @s dirt` as the game will suggest `/give @s minecraft:dirt`. However, content not under the `minecraft` namespace needs to have its namespace typed before Minecraft autosuggests the full identifier. Typing in `/give @s cogwheel` will not have `create:cogwheel` suggested. The user has to type `/give @s create:` before Minecraft autosuggests the full identifier`.

# Change
This PR adds the mixin `CommandSourceMixin` to the `registry-sync-v0` module. The mixin allows Minecraft to autofill modded identifiers through cancelling the check that makes sure the autofill has the namespace `minecraft`.

### DIFF
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
@@ -2,7 +2,6 @@ package net.fabricmc.fabric.mixin.command.client;
 
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
-
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
@@ -9,7 +9,7 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 
 @Mixin(CommandSource.class)
-public class CommandSourceMixin {
+public interface CommandSourceMixin {
 
 	// Minecraft is hardcoded to only autofill identifiers with the "minecraft" namespace. This cancels the check
 	@WrapOperation(
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
@@ -3,15 +3,15 @@ package net.fabricmc.fabric.mixin.command.client;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 
-import net.minecraft.command.CommandSource;
-
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 
+import net.minecraft.command.CommandSource;
+
 @Mixin(CommandSource.class)
 public interface CommandSourceMixin {
-
-	// Minecraft is hardcoded to only autofill identifiers with the "minecraft" namespace. This cancels the check
+	// Minecraft is hardcoded to only autofill identifiers with the "minecraft" namespace.
+	// This allows non-vanilla identifiers to be autofilled
 	@WrapOperation(
 			method = "forEachMatching(Ljava/lang/Iterable;Ljava/lang/String;Ljava/util/function/Function;Ljava/util/function/Consumer;)V",
 			at = @At(value = "INVOKE", target = "Ljava/lang/String;equals(Ljava/lang/Object;)Z", ordinal = 0)
@@ -19,5 +19,4 @@ public interface CommandSourceMixin {
 	private static boolean cancelNamespaceCheck(String instance, Object o, Operation<Boolean> original) {
 		return true;
 	}
-
 }
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
@@ -1,4 +1,4 @@
-package net.fabricmc.fabric.mixin.registry.sync.client;
+package net.fabricmc.fabric.mixin.command.client;
 
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


Ramixin:
Moved the mixin to the `command-api-v2` module

chyzman:
https://github.com/chyzman/dontDoThat/blob/main/src/main/java/com/chyzman/dontdothat/mixin/CommandSourceMixin.java
does the same thing I'm not sure which method of doing this is better someone whose better at mixin might wanna compare them?

Ramixin:
@chyzman , I asked in the Mixin-dev channel and Llama said both work equally

Dragon-Seeker (approved):

chyzman (approved):
👍

Benonardo (approved):

jacg311 (approved):

gliscowo (approved):

Earthcomputer:
This should be pretty easy to add a unit test for?

aurakle (approved):

Dev0Louis (approved):

modmuss50:
Do we understand why this check is here in the first place, it looks very purposful and removing it without fully understanding the context worries me.

Have you got an example of what commands this affects, are there any behaviour changes when no mods are installed?

I worry that this PR has some unintended side affects that we may not want, its not something that we tend to change like this without careful consideration. As earth said a unit tests should be possible to add as well.

modmuss50:
Please fix the build failuires.

modmuss50:
Another question:

What is the behaviour when you have 2 identifers with the same path.

E.g you have both a `minecraft:dirt` and a `mymod:dirt` is there a worry that doing `/give @a dirt` will give you the wrong item? Will dirt be shown twice in the list of suggestions? This can also apply when two mods register something with the same path, it doesnt necessarily need to clash with `minecraft`.

When creating identifiers in code a string without a namespace always defaults to `minecraft` if I understand this right it changes that behaviour leading to possibly confusing and undefined behaviour.

chyzman:
This does not effect namespaceless identifiers, this does not change that

modmuss50:
> This does not effect namespaceless identifiers, this does not change that

What does it change, sorry its not very clear to me what this PR does based on the description.

Earthcomputer:
> What does it change, sorry its not very clear to me what this PR does based on the description.

It changes auto-completion. If you type `dirt` and press tab, both `minecraft:dirt` and `mymod:dirt` will show up.

chyzman:
Minecraft namespaced identifiers will still appear before other namespaces tho

Ramixin:
I have updated the mixin to be correctly formatted and I added two unit tests: one for testing if the non-vanilla identifier is suggested at all and the other checking that vanilla identifiers still have precedence.

Ramixin:
> Another question:
> 
> What is the behaviour when you have 2 identifers with the same path.
> 
> E.g you have both a `minecraft:dirt` and a `mymod:dirt` is there a worry that doing `/give @a dirt` will give you the wrong item? Will dirt be shown twice in the list of suggestions? This can also apply when two mods register something with the same path, it doesnt necessarily need to clash with `minecraft`.
> 
> When creating identifiers in code a string without a namespace always defaults to `minecraft` if I understand this right it changes that behaviour leading to possibly confusing and undefined behaviour.

When two identifiers have the same path, the vanilla identifier will always have precedence. So `minecraft:dirt` will always come before `mymod:dirt` in the suggestion list.

Identifiers without namespaces will always default to the vanilla namespace, so `/give @s dirt` will give `minecraft:dirt`. The namespace is still necessary on non-vanilla identifiers. The only difference is that the namespace can be autofilled.

Ramixin:
> Have you got an example of what commands this affects, are there any behaviour changes when no mods are installed?

This will affect any command that has an argument that uses identifiers to specify an object. (ex. `/give`, `/function`, `/effect`)
When no mods or datapacks are present, vanilla behavior is upheld since all identifiers are vanilla and would return `true` on the check anyway.

TheTrueYak:
yes can confirm. this PR does not at all change command syntax or logic itself, it just allows nonvanilla namespaces to appear as suggestions alongside vanilla ones; though the vanilla ones are still prioritized, as ramixin said

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
```diff
@@ -0,0 +1,22 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+import net.minecraft.command.CommandSource;
+
+@Mixin(CommandSource.class)
+public interface CommandSourceMixin {
+	// Minecraft is hardcoded to only autofill identifiers with the "minecraft" namespace.
+	// This allows non-vanilla identifiers to be autofilled
+	@WrapOperation(
```
modmuss50: A Redirect is prefered over none conditional WrapOperation's

fabric-command-api-v2/src/client/resources/fabric-command-api-v2.client.mixins.json
```diff
@@ -3,6 +3,7 @@
   "package": "net.fabricmc.fabric.mixin.command.client",
   "compatibilityLevel": "JAVA_21",
   "client": [
+    "CommandSourceMixin",
```
modmuss50: Sort alpahatically.

Earthcomputer:
The build is still failing, you're missing license headers and import order. Try running `./gradlew spotlessApply`, it should fix all these issues.

modmuss50 (changes requested):

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
LlamaLad7: Not always. The sole purpose of a Redirect is to cause hard conflicts, there is no reason for it here. Anyone else changing this would likely change it the same way, and even if someone else wanted to customise the check to add their own modid for example, returning true always is perfectly compatible with that.

Earthcomputer: Even if not `Redirect`, then `ModifyExpressionValue` could be used here instead.

cputnam-a11y: I do think ModifyExpressionValue would be better in this case though, as it allows others code to still run around the checks, even though their value is discarded 
Oops, GitHub didn't update

chyzman: [so this?](https://github.com/chyzman/dontDoThat/blob/main/src/main/java/com/chyzman/dontdothat/mixin/CommandSourceMixin.java)

fabric-command-api-v2/src/client/resources/fabric-command-api-v2.client.mixins.json
Patbox: Should this mixin be actually moved to common? Modified code can run on both sides the change doesn't require anything client only

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
Ramixin: I chose `WrapOperation` over `ModifyExpressionValue` because it returns true no matter what, so skipping the other injections doesn't affect the result. But, if `ModifyExpressionValue` is preferred, I can switch it.

fabric-command-api-v2/src/client/resources/fabric-command-api-v2.client.mixins.json
Ramixin: I made it client-side since I was assuming that this would only be useful for the client command auto-complete suggestions, but I have realized that there are some mods that utilize this in the terminal in some ways.

TheTrueYak: probably. the only reason it was in client was because the initial mod was client side

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/CommandSourceMixin.java
Earthcomputer: Yes please use MEV

modmuss50 (approved):

modmuss50:
Checkstyle still doesnt pass, Ill fix it myself. Otherwise this is good to merge.

JT122406:
Can this be backported to 1.21.1?

Issue #3656, opened 2024-03-18
Going to resource packs then clicking done causes resource reload


mrjasonn:
Going to resource packs menu then clicking done causes resource reload, but only for every time you launch the game. This has been tested on Minecraft 1.21.3 with Fabric API 0.107.0.
This bug is separate from [MC-264307](https://bugs.mojang.com/browse/MC-264307) as I have mentioned that the issue in Fabric API is that it happens every time you launch the game.

modmuss50:
I can reproduce this in 1.20.4

haykam821:
Is this issue separate from [a similar vanilla issue](https://bugs.mojang.com/browse/MC-264307) that also causes resource reloading when clicking the 'Done' button?

mrjasonn:
> Is this issue separate from [a similar vanilla issue](https://bugs.mojang.com/browse/MC-264307) that also causes resource reloading when clicking the 'Done' button?

different from vanilla one, as [Mod] Greymagic27 said in the mojang one: 

> I was able to reproduce this only when I've first loaded up the game and went there. It displayed the red Mojang loading screen for 2 seconds or so.
> 
> After that I couldn't reproduce without relaunching again, and then it only did it once.

so it is separate from the Fabric API one in how it works because the Fabric API one is every time launch the game

mclegoman:
When entering the resource pack screen, the game will disable resource packs that have been registered through fabric api, thus when pressing done it will reload the resources. If you re-enable the resources that were already enabled, the resources won't reload.

From my testing, this seems to happen no matter the activation type.

I am currently using fabric-api 0.97.3+1.20.5, with fabric loader 0.15.10, on 1.20.5-pre4, but this issue has been happening throughout the 1.20.5 snapshots.

mrjasonn:
Now checked on 1.20.5, still same issue

mrjasonn:
1.20.6 is also affected (API ver 0.97.8)

maevius669:
Happens with 1.21, loader 0.16.0 and fapi 0.100.7
doesnt matter if pressing "Done" or "esc", it will reload resources even if unchanged

Elitaria:
also having this issue on 1.21.1

PR #4756, opened 2025-07-20
Prevent Guaranteed Resource Reload on first resource refresh

chyzman:
Injects into `GameOptions.refreshResourcePacks` and prevents mod resource packs from being included in the check that decides if resource packs have changed. Fabric prevents them from being saved to and subsequently loaded from the `options.txt` file which will then **always** cause the first resource reload check to trigger a resource reload regardless of whether or not any packs have actually changed.

### DIFF
--- a/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
+++ b/fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
@@ -25,6 +25,9 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.llamalad7.mixinextras.sugar.Local;
+import net.fabricmc.fabric.impl.resource.loader.FabricResourcePackProfile;
 import org.slf4j.Logger;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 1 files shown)


modmuss50:
Please fix the build failuire...

LambdAurora (approved):
With Resource Loader v1, its design would avoid the need of such patch, but this is a nice stopgap solution until time is found to make it ;)

This closes #3656 as well.

apple502j (approved):

fabric-resource-loader-v0/src/client/java/net/fabricmc/fabric/mixin/resource/loader/client/GameOptionsMixin.java
```diff
@@ -132,4 +135,10 @@ private void onLoad(CallbackInfo ci) {
 
 		this.resourcePacks = new ArrayList<>(resourcePacks);
 	}
+
+	@ModifyExpressionValue(method = "refreshResourcePacks", at = @At(value = "INVOKE", target = "Lnet/minecraft/resource/ResourcePackProfile;isPinned()Z"))
```
Juuxel: Should this be a WrapOperation? (see https://github.com/FabricMC/fabric/pull/4693#discussion_r2146943357)

Juuxel (approved):

modmuss50:
Unless im missing something/being silly this doesnt seem to work. I am testing by:
1. delete options.txt
2. start game and go to resource packs
3. press "done" and expect there to be no reload.

`((FabricResourcePackProfile) instance).fabric_isHidden()` returns false for the "Fabric Mods" pack, this is expected as its not hidden.

chyzman:
> Unless im missing something/being silly this doesnt seem to work. I am testing by:
> 1. delete options.txt
> 2. start game and go to resource packs
> 3. press "done" and expect there to be no reload.
> 
> `((FabricResourcePackProfile) instance).fabric_isHidden()` returns false for the "Fabric Mods" pack, this is expected as its not hidden.

The fix is for every mods' hidden packs

chyzman:
Tho that is actually another issue isnt it

chyzman:
After further investigation, I am confusedl. The issue this pr is attempting to fix is the fact that fabric prevents the hidden mod resource packs from being used in the resource packs changed check, the issue modmus is having seems to be the non hidden fabric pack isn't added on options.txt generation? I will look into this further when I have a chance

modmuss50:
Ah, I see that makes sense. Looking at the code this PR did make sense to me so I still think we should merge it. We can fix the other issue as something seperate. Thanks for clarifying.

modmuss50 (approved):

PR #4762, opened 2025-07-25
Fix SpecialGuiElementRegistryImpl not handling multiple GuiRenderer instances

### LINKED ISSUE #4747: GuiRendererMixin tries to patch special element registry outside of GameRenderer ctor
When making a new GuiRenderer outside of GameRenderer, GameRendererMixin tries to patch the special element render registry:

https://github.com/FabricMC/fabric/blob/b120773fbed83b77936ca0b03c06828e090ebe0b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiRendererMixin.java#L50

This causes either an exception if the game is already started, or could cause unwanted behavior. Instead, the list given to the GuiRenderer ctor from GameRenderer could be patched, or the created GuiRenderer could be modified in GameRenderer.

Johni0702:
Presently, trying to instantiate a second `GuiRenderer` results in an `IllegalStateException` being thrown from [here](https://github.com/FabricMC/fabric/blob/bcdb7097dafe598447427dfd603b41b54507e140/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementRegistryImpl.java#L45).
This PR rewrites `SpecialGuiElementRegistryImpl` to handle multiple `GuiRenderer` instances being created.

Fixes https://github.com/FabricMC/fabric/issues/4747, although the description there kind of implies that they'd prefer the `SpecialGuiElementRegistry` only apply to the main `GuiRenderer`.
This PR makes it apply to all `GuiRenderer`, which imo is more reasonable. E.g. my use-case involves rendering specific parts of some vanilla guis to a texture to apply some post-processing effects to them before enqueuing the resulting textures back into the main `DrawContext`. If a third-party mod modifies one of these parts and uses a special gui element in there, I'd like for my secondary `GuiRenderer` to know how to render that as well.

Tested with the included test mod.
The test mod serves only to show that the whole thing works, any real usage of a second `GuiRenderer` would probably render to a texture instead of straight to the screen, but I've tried to keep it as simple as it can be while still getting something on-screen.

I've switched to using an Accessor to get the map and vertex consumers instead of passing them from the existing mixin because
1) keeping multiple references would require a more complicated setup with multiple `WeakReference`s
2) if another mod which isn't aware of FAPI's mixin tries to inject a custom renderer in much the same was as FAPI does, by replacing the `specialElementRenderers` immutable Map with a mutable one, and its mixin happens to execute after the one from FAPI, then FAPI would be holding on to the wrong `Map` and late registrations would not actually function correctly.


PepperCode1:
These changes are good overall, but we should consider throwing on registration after the first `GuiRenderer` has been constructed so we don't need to keep track of all `GuiRenderer`s or use an accessor. I'd like some opinions on this.

Johni0702:
> These changes are good overall, but we should consider throwing on registration after the first `GuiRenderer` has been constructed so we don't need to keep track of all `GuiRenderer`s or use an accessor. I'd like some opinions on this.

Doing so would be a breaking change. While, based on a quick GitHub search, most people do currently seem to register their renderer from `onInitializeClient`, it doesn't seem inconceivable that there might be some mods which call the register method from the static initializer of their custom state class, or some other point after the first `GuiRenderer` has already been initialized.
Half of the previous implementation was dedicated to making late registration work, so I've assumed there was some reason for that and have preserved that functionality. And since it comes at the cost of just one map and one accessor, I don't think it makes much sense to drop it at this point.

modmuss50:
> Half of the previous implementation was dedicated to making late registration work, so I've assumed there was some reason for that and have preserved that functionality.

This was mostly about not knowing for sure when the vanilla class would be loaded and not for mods to register outside of the mod init. I think its tottaly reasonable to throw if something is registered too late.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementRegistryImpl.java
```diff
@@ -28,64 +28,32 @@
 import net.fabricmc.fabric.api.client.rendering.v1.SpecialGuiElementRegistry;
 
 public final class SpecialGuiElementRegistryImpl {
-	private static RegistrationHandler registrationHandler = new EarlyRegistrationHandler();
+	private static final List<SpecialGuiElementRegistry.Factory> factories = new ArrayList<>();
```
PepperCode1: Nit: `static final` fields should use `UPPER_SNAKE_CASE` - rename to `FACTORIES`.

PepperCode1 (approved):

modmuss50 (approved):

PR #4763, opened 2025-07-25
Fix multiple special GUI elements not rendering properly

Earthcomputer:
This PR fixes special GUI elements not rendering properly when there are multiple of them, specifically overwriting the same texture and then rendering that texture many times. The fix works by using the factories supplied by the existing API to create new renderer instances as needed, which each have separate textures to render to. Factories for vanilla special elements are registered manually, and I added a client gametest to make sure we remember to update this list if needed.

Here's a screenshot of some special renderers in the testmod:
image


cputnam-a11y:
Not necessarily good or bad, but this doesn't seem to do any texture size matching. If we want to do that it could certainly come in a later pr

Earthcomputer:
Neither does vanilla. This is meant to be a fairly bare-bones fix, if we want to add optimizations then we can think about that later, but I'm not sure it will be worth it as Mojang will surely want to optimize this themselves.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementOverflowData.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.render.SpecialGuiElementRenderer;
+import net.minecraft.client.gui.render.state.special.SpecialGuiElementRenderState;
+import net.minecraft.client.render.VertexConsumerProvider;
+
+public final class SpecialGuiElementOverflowData<T extends SpecialGuiElementRenderState> implements AutoCloseable {
```
PepperCode1: Personally I do not like the term "overflow renderer" or "overflow data". I would rename this class to `SpecialGuiElementRendererPool` and substitute the former terms accordingly.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementOverflowData.java
```diff
@@ -0,0 +1,63 @@
...
+	private int index = 0;
+	private final List<SpecialGuiElementRenderer<T>> renderers = new ArrayList<>();
```
PepperCode1: Nit: The final field should go before the non-final field.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiRendererMixin.java
```diff
@@ -43,10 +47,32 @@ abstract class GuiRendererMixin {
...
+	@Inject(method = "prepareSpecialElements", at = @At("HEAD"))
+	private void onPrepareSpecialElements(CallbackInfo ci) {
+		overflowDatas.values().forEach(SpecialGuiElementOverflowData::newFrame);
+	}
+
+	@ModifyVariable(method = "prepareSpecialElement", at = @At("STORE"))
+	private <T extends SpecialGuiElementRenderState> SpecialGuiElementRenderer<T> substitueOverflowedSpecialElementRenderer(SpecialGuiElementRenderer<T> original, T elementState) {
+		SpecialGuiElementOverflowData<T> overflowData = (SpecialGuiElementOverflowData<T>) overflowDatas.computeIfAbsent(original.getElementClass(), k -> new SpecialGuiElementOverflowData<>());
+		return overflowData.substitute(original, elementState, MinecraftClient.getInstance(), vertexConsumers);
+	}
```
PepperCode1: It might be better to use a `WrapOperation` on `specialElementRenderers.get` instead, and invoke the original operation if the pool returns `null`, otherwise use what the pool returned. It probably doesn't matter in terms of performance, but I think it makes the intent slightly clearer.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InventoryScreenMixin.java
```diff
@@ -0,0 +1,41 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import net.minecraft.client.gui.screen.ingame.InventoryScreen;
+import net.minecraft.client.render.entity.EntityRenderer;
+import net.minecraft.client.render.entity.state.EntityRenderState;
+import net.minecraft.entity.Entity;
+
+@Mixin(InventoryScreen.class)
+public class InventoryScreenMixin {
```
PepperCode1: Nit: `abstract` instead of `public`.

PepperCode1 (commented):
Additionally, we should consider consider deleting and removing a renderer from a pool if the renderer was not used this frame. Vanilla does exactly this with its `oversizedItems` at the end of `render(GpuBufferSlice)`.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementOverflowData.java
Earthcomputer: Is there any style guide that recommends this? I wasn't aware of this convention. Neither field is assigned in the constructor, and both are semantically mutable, the final field is only final because it legally can be.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/InventoryScreenMixin.java
Earthcomputer: This doesn't seem to be the convention for other mixins in Fabric API, but does seem to be applied, albeit inconsistently, to mixins in this module. I will change it to be consistent with that.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementOverflowData.java
PepperCode1: It's what I personally prefer; I don't think there is a Fabric API convention for this. It's a nitpick, so if you disagree, feel free to ignore it.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/SpecialGuiElementOverflowData.java
Earthcomputer: I named it "overflow data" so that the name implies that it excludes the renderers stored in the vanilla special element renderer map. I wanted to continue to use the renderers in that map where possible in case some mod is inspecting that map for some odd reason.

PepperCode1: I understand the intent behind "overflow" and agree with the design you describe, but I don't know how important it is to emphasize that in the name considering it is already long. At the same time, it excludes "renderer" and makes it seem like additional data for an element state.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiRendererMixin.java
Earthcomputer: I think the intent is clear both ways, but I thought that this was slightly better because with `WrapOperation` or MEV I would be targetting `Ljava/util/Map;get(Ljava/lang/Object;)Ljava/lang/Object;` which feels a bit generic and slightly less robust against changes to the target method. Once MixinExtras expressions are out, we could target `specialElementRenderers.get` directly. Slices are of course an option but I think this is clearer and equally good.

PepperCode1: I would argue that targeting `STORE` is just as generic if not more generic, but I'm fine with either design.

Earthcomputer: `STORE` is in implicit mode, meaning it matches only stores of the type in the parameter of the handler method, i.e. only stores of `SpecialGuiElementRenderer`, not of other types.

PepperCode1 (approved):
I'm not sure if it matters, but I was thinking that additional renderers would be cleared right after vanilla clears its oversized item renderers. I don't necessarily see any issues with the current approach, though.

modmuss50 (approved):

modmuss50:
Humm, this conflicts quite badly with: https://github.com/FabricMC/fabric/commit/bd15c559f79d004a5c8dc0615c9c2c8c23a416c9

layou233:
Hi! The vanilla renderers cache textures internally, and if the `shouldBypassScaling` method (called `textureIsReadyToBlit` in mojmap, which seems to be a better fit) is implemented correctly, they can directly blit without re-rendering for every frame. However, the current pool uses `List<SpecialGuiElementRenderer<T>>` to store renderers and reuses them based on the order in which render states are submitted. So I'm thinking that changing the submission order might accidentally invalidate those caches, forcing most textures to be re-rendered. It may sound or look like over-engineering, but do we need to change it to something like `HashMap<T, SpecialGuiElementRenderer<T>>` to reuse the cache as much as possible?

PR #4766, opened 2025-07-26
Add a codec for TriState

Patbox:
Pretty simple PR adding a codec to the fabric-provided TriState enum.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
@@ -21,24 +21,31 @@ import java.util.Optional;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 
+import com.mojang.serialization.Codec;
 import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.util.StringIdentifiable;
 
 /**
  * Represents a boolean value which can be true, false or refer to a default value.
  */
-public enum TriState {
+public enum TriState implements StringIdentifiable {
 	/**
 	 * Represents the boolean value of {@code false}.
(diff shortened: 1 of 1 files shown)


fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
```diff
@@ -136,4 +143,18 @@ public <X extends Throwable> boolean orElseThrow(Supplier<X> exceptionSupplier)
...
+	 * Value of this enum as string.
+	 *
+	 * @return lowercase name of the value.
+	 */
+	@Override
+	public String asString() {
+		return this.name;
+	}
+
+	TriState(String name) {
```
PepperCode1, line 157: Nit: I would put the constructor after the fields and before the methods.

PepperCode1 (approved):

modmuss50 (approved):

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
Patbox: Tbf I feel that the field could be moved down there just to be consistent. While this isn't really java-style, the constructor / private fields are purely implementation details that are useless to pretty much all mod devs (the enum values and all extra methods are only ones important).

Earthcomputer: On the other hand, most Java programmers expect a class' members to be in a certain order, at least for me, changing the order makes me a do a double-take when reading it

apple502j:
Note: vanilla has had TriState for a while (I think it's 1.21.4 that added it?), but it does not have a codec field.

PR #4770, opened 2025-07-28
Setup Resource Loader v1 and rework resource reloaders

### LINKED ISSUE #2577: Rework resource reloader API
Goals:
- Allow communication between resource reloaders.
- Allow ordering of modded reloaders.
- Allow ordering of modded and vanilla reloaders (including running before some vanilla reloaders).
- Prevent changes to the ordering of vanilla reloaders.

Here is my proposed solution, for server resources. I think we can do something similar for client resources. I am not sure that we need all the `AFTER_XXX` events.
```java
interface AddServerResourceReloaderCallback {
    // Choose event for ordering wrt. vanilla, and use event phases to order wrt. other modded reloaders.
    Event<AddServerResourceReloaderCallback> BEFORE_ALL = create(...); // before all vanilla reloaders
    Event<AddServerResourceReloaderCallback> AFTER_TAGS = create(...); // after tags
    Event<AddServerResourceReloaderCallback> AFTER_LOOT_CONDITIONS = create(...); // after tags and loot conditions
    Event<AddServerResourceReloaderCallback> AFTER_RECIPES = create(...); // after tags, loot conditions, and recipes
    Event<AddServerResourceReloaderCallback> AFTER_LOOT = create(...); // etc...
    Event<AddServerResourceReloaderCallback> AFTER_LOOT_FUNCTIONS = create(...); // etc...
    Event<AddServerResourceReloaderCallback> AFTER_FUNCTIONS = create(...); // etc...
    Event<AddServerResourceReloaderCallback> AFTER_ADVANCEMENTS = create(...); // etc...
    Event<AddServerResourceReloaderCallback> AFTER_ALL = create(...); // after all vanilla reloaders

    // Server allows getting extra state in the event, but should not be used during the reload!
    // Data pack contents allow querying vanilla reloaders
    // Registry allows adding and querying modded reloaders
    void addResourceReloaders(MinecraftServer server, DataPackContents dataPackContents, ModdedReloaders registry);

    interface ModdedReloaders {
        void add(Identifier identifier, ResourceReloader reloader);
        @Nullable
        ResourceReloader get(Identifier identifier);
    }
}
```

### LINKED ISSUE #3724: Resource reloaders are context-less
## Problem

Vanilla's resource reloaders are constructed anew on each reload get access to some context, notably the parameters to `DataPackContents`. Since 1.20.5, I need access to the `RegistryWrapper.WrapperLookup` for deserialization context.

## Potential Solution

Instead of just having a global list of pre-created of resource loaders, create an event to register reload listeners with context arguments passed in (ideally the `DataPackContents` and `RegistryWrapper.WrapperLookup`). This would be similar to `fabric-networking-api`, where both global and context-based receivers may be registered.

LambdAurora:
The long-awaited Resource Loader v1 is finally here!

The goal of this PR is simple:
- setup Resource Loader v1 to make future work easier;
- rework resource reloaders to:
  - make sorting of resource reloaders use toposort;
  - allow mods to register resource reloaders more easily, which should discourage multiloader mods from taking shortcuts;
  - fill in missing resource reloader keys for Vanilla resource reloaders;

What are not the goals of this PR:
- full feature parity with Resource Loader v0;
- implement mod resources loading logic, this is still the job of Resource Loader v0 for now.

Closes #2577.
Closes #3724 (though, by re-reading it I think it should already be closed?).


fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/SimpleResourceReloader.java
PepperCode1, line 1: The "load" stage is actually called the "prepare" stage in both Yarn and Mojmap, so the documentation and method name should be updated. At the same time, vanilla already has `SinglePreparationResourceReloader` and it is very similar to this class so I'm not sure we need it.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
```diff
@@ -0,0 +1,92 @@
...
+	 *
+	 * <p>Note: This is only supported for server data reloaders.
+	 *
+	 * @param id the identifier of the resource reloader
+	 * @param reloaderFactory a function that creates a new instance of the listener with a given registry lookup
+	 * @see #registerReloader(IdentifiableResourceReloader)
+	 * @see #registerReloader(Identifier, ResourceReloader)
+	 * @see #addReloaderOrdering(Identifier, Identifier)
+	 */
+	void registerReloader(Identifier id, Function<RegistryWrapper.WrapperLookup, ResourceReloader> reloaderFactory);
```
PepperCode1: I think server reloaders should receive more context, including pending tags and enabled feature flags. They should also receive a DRM instead of a lookup if possible.

PepperCode1: After taking a long look at how server resource reloads work, the lookup already allows checking if a certain tag exists. For reloadable registries, the contents of all tags will be populated; for non-reloadable registries, the contents of all tags will be empty (but still present) until after the reload is finished. For this reason, I also think it doesn't make sense to provide the full DRM, as its registries will return tags from the previous reload.

This method can be used to check if a tag exists.

```java
static <T> boolean tagExists(RegistryWrapper.WrapperLookup lookup, TagKey<T> tagKey) {
	Optional<? extends RegistryWrapper.Impl<T>> optionalWrapper = lookup.getOptional(tagKey.registryRef());

	if (optionalWrapper.isEmpty()) {
		return false;
	}

	return optionalWrapper.get().getOptional(tagKey).isPresent();
}
```

If we want to allow reloaders to access updated tag contents, then the simplest option in terms of API design would be to patch the the `Registry.PendingTags.lookup()` implementation in `WritableRegistry` so that it returns the updated tag contents instead of empty tag contents. Any additional access beyond a presence check and a contents size check would require patching that implementation anyway, and only providing size through some additional context parameter does not seem useful to me, so I think it makes sense to route all current and future tag-related functionality through the lookup. We can, however, make the above method a utility method in the API.

Feature flags should still be provided.

PepperCode1:
The module lifecycle should probably be experimental until v1 has at least as much functionality as v0.

apple502j:
No, the module is not experimental, as the old APIs are deprecated.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,70 @@
...
+ * <p>"Identifiable" resource reloaders have a unique identifier, which can be depended on,
+ * and can provide dependencies that they would like to see executed before themselves.
+ *
+ * @see ResourceReloaderKeys
+ */
+public interface IdentifiableResourceReloader extends ResourceReloader {
+	/**
+	 * {@return the unique identifier of this resource reloader}
+	 */
+	@NotNull Identifier getFabricId();
```
Juuxel: Nit: FAPI doesn't use `@NotNull`.
https://github.com/FabricMC/fabric/blob/efc62317660a3d5e2a2e7c8cdf75c1a2c6a76bfa/CONTRIBUTING.md?plain=1#L141

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
```diff
@@ -0,0 +1,92 @@
...
+	 * @see #addReloaderOrdering(Identifier, Identifier)
+	 */
+	default void registerReloader(Identifier id, ResourceReloader reloader) {
+		this.registerReloader(IdentifiableResourceReloader.wrap(id, reloader));
+	}
+
+	/**
+	 * Register a resource reloader for a given resource manager type.
+	 *
+	 * <p>Note: This is only supported for server data reloaders.
```
Juuxel: I wonder if it'd be better to just provide separate registration APIs for the client and the server instead of having methods that throw. (The API could have two subinterfaces or just two separate classes with static methods if sharing the common itf type or its impl isn't needed)

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
```diff
@@ -0,0 +1,92 @@
...
+	 *
+	 * <p>Please keep in mind that this only takes effect during the application stage!
+	 *
+	 * @param firstReloader  the identifier of the resource reloader that should run before the other
+	 * @param secondReloader the identifier of the resource reloader that should run after the other
+	 * @see net.fabricmc.fabric.api.resource.v1.reloader.ResourceReloaderKeys identifiers of Vanilla resource reloaders
+	 * @see #registerReloader(Identifier, ResourceReloader) register a new resource reloader
+	 * @see #registerReloader(IdentifiableResourceReloader) register a new resource reloader
+	 */
+	void addReloaderOrdering(@NotNull Identifier firstReloader, @NotNull Identifier secondReloader);
```
Juuxel: The comment about NotNull also applies here

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
```diff
@@ -136,4 +136,21 @@ public <X extends Throwable> boolean orElseThrow(Supplier<X> exceptionSupplier)
 
 		throw exceptionSupplier.get();
 	}
+
+	/**
+	 * {@return a parsed TriState from a system property}
+	 *
+	 * @param property the system property
+	 */
+	public static TriState fromProperty(String property) {
```
Juuxel: Maybe from**System**Property? `fromProperty` on an enum sounds more like parsing a block/fluid state property to me.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
PepperCode1: This is also something I considered. The current design doesn't allow specifying ordering of server reloaders based on the context either.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
LambdAurora: Done.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/SimpleResourceReloader.java
LambdAurora: I ended up removing `SimpleResourceReloader`, I missed `SinglePreparationResourceReloader` when looking around.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceReloaderPhaseData.java
```diff
@@ -0,0 +1,115 @@
...
+	ResourceReloaderPhaseData(Identifier id, @Nullable ResourceReloader resourceReloader) {
+		super();
+		this.id = id;
+		this.resourceReloader = resourceReloader;
+	}
+
+	/**
+	 * Marks this phase and all preceding phases as running before Vanilla.
+	 */
+	void markBefore() {
```
apple502j, line 59: A few comments on the logic would be nice.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceReloaderPhaseData.java
```diff
@@ -0,0 +1,115 @@
...
+		boolean isAfter = this.vanillaStatus == VanillaStatus.AFTER;
+
+		if (this.vanillaStatus != VanillaStatus.NONE && !isAfter) return;
```
apple502j, line 62: Just to make sure, if `isAfter` is true this is a cyclic dependency, right?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+		// Remove any modded reloaders to sort properly.
+		reloaders.removeAll(reloadersToAdd);
+
+		// General rules:
+		// - We *do not* touch the ordering of vanilla reloaders. Ever.
+		//   While dependency values are provided where possible, we cannot
+		//   trust them 100%. Only code doesn't lie.
+		// - We add all custom reloaders after vanilla reloaders if they don't have contrary ordering. Same reasons.
+
+		var runtimePhases = new Object2ObjectOpenHashMap<Identifier, ResourceReloaderPhaseData>();
```
apple502j, line 161: Just to make sure, ordering is irrelevant here since it's toposorted and we make no guarantee of order between unordered modded reloaders, right?

apple502j (approved):

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
LambdAurora: There is no guarantee of ordering between unordered modded reloaders. The order of this map is irrelevant.

The values of the map are then in an array list that's toposorted. The map only acts as a quick intermediate lookup while building the phase data.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceReloaderPhaseData.java
LambdAurora: Indeed, it would be a cyclic dependency.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceReloaderPhaseData.java
LambdAurora: Added some comments on the vanillaStatus field.

apple502j (approved):

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/SimpleResourceReloader.java
```diff
@@ -0,0 +1,67 @@
...
+
+import net.minecraft.resource.ResourceReloader;
+
+/**
+ * A variant of {@link net.minecraft.resource.SinglePreparationResourceReloader}
+ * which passes the shared state store instead of the resource manager in its methods.
+ *
+ * <p>In essence, there are two stages:
+ *
+ * <ul><li>load: create an instance of your data object containing all loaded and
```
PepperCode1:
```suggestion
 * <ul><li>prepare: create an instance of your data object containing all loaded and
```

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/SimpleResourceReloader.java
```diff
@@ -0,0 +1,67 @@
...
+ * processed information,
+ * <li>apply: apply the information from the data object to the game instance.</ul>
+ *
+ * <p>The load stage should be self-contained as it can run on any thread! However,
```
PepperCode1:
```suggestion
 * <p>The prepare stage should be self-contained as it can run on any thread! However,
```

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+				if (data.resourceReloader == null) {
+					str += " (virtual)";
+				}
+
+				return str;
+			}).collect(Collectors.joining(", ")));
+		}
+	}
+
+	private ResourceReloader extractSetupMarker(List<ResourceReloader> reloaders) {
```
PepperCode1: Should this method should be annotated with Nullable?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+				it.remove();
+				return marker;
+			}
+		}
+
+		throw new IllegalStateException("No SetupMarkerResourceReloader found in reloaders!");
+	}
+
+	// A bit of a hack to get the registry, but it works.
+	public static @Nullable RegistryWrapper.WrapperLookup getWrapperLookup(List<ResourceReloader> reloaders) {
```
PepperCode1: This method is not Nullable.

PepperCode1 (changes requested):

PepperCode1 (approved):

modmuss50:
Question: is this something we want to take only into 1.21.9? Or also backport to older versions? Personally I would be quite happy to just take this as a future addion, and if there is a big desire backport it at a later date?

LambdAurora:
Right now this pull request leverages from some 1.21.9 features, if a backport happens it should be at a later date.

modmuss50:
Great, thanks for confirming. I'll review this today and we can get it merged ASAP, I think it will be better to get it into the pre-releases now rather than sitting in last call for a week.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
```diff
@@ -144,6 +144,23 @@ public <X extends Throwable> boolean orElseThrow(Supplier<X> exceptionSupplier)
...
+	 * {@return a parsed TriState from a system property}
+	 *
+	 * @param property the system property
+	 */
+	public static TriState fromSystemProperty(String property) {
+		String value = System.getProperty(property);
+
+		if ("true".equalsIgnoreCase(value) || "on".equalsIgnoreCase(value)) {
+			return TRUE;
+		} else if ("false".equalsIgnoreCase(value) || "off".equalsIgnoreCase(value)) {
```
modmuss50: Should this be `value != null` basically meaning if the user specified the system prop but its not one of the truthy values? THis would then work similar to java's `Boolean.getBoolean`.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
```diff
@@ -144,6 +144,23 @@ public <X extends Throwable> boolean orElseThrow(Supplier<X> exceptionSupplier)
...
 
+	/**
+	 * {@return a parsed TriState from a system property}
+	 *
+	 * @param property the system property
+	 */
+	public static TriState fromSystemProperty(String property) {
+		String value = System.getProperty(property);
+
+		if ("true".equalsIgnoreCase(value) || "on".equalsIgnoreCase(value)) {
```
modmuss50: I wouldnt allow `on`, its not needed and you could maybe just use `Boolean.parseBoolean` here.

E.g:

```java
if (value != null) {
  return Boolean.parseBoolean(value) ? TRUE : FALSE;
}

return DEFAULT;
```

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/mixin/resource/v1/client/KeyedClientResourceReloadListenerMixin.java
```diff
@@ -0,0 +1,126 @@
...
+		SplashTextResourceSupplier.class,
+		SoundManager.class,
+		TextureManager.class,
+		WaypointStyleAssetManager.class,
+		/* private */
+		WorldRenderer.class, VideoWarningManager.class, PeriodicNotificationManager.class
+})
+public abstract class KeyedClientResourceReloadListenerMixin implements IdentifiableResourceReloader {
+	@Unique
+	private Identifier fabric$id;
```
modmuss50: `fabric$` prefix is not needed on a `@Unique` field.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/ResourceReloaderKeys.java
```diff
@@ -0,0 +1,86 @@
...
+	 * @see net.fabricmc.fabric.api.resource.v1.ResourceLoader#addReloaderOrdering(Identifier, Identifier)
+	 */
+	public static final Identifier AFTER_VANILLA = Identifier.of("fabric", "after_vanilla");
+
+	private ResourceReloaderKeys() { }
+
+	/**
+	 * Keys for various client resource reloaders.
+	 */
+	public static final class Client {
```
modmuss50, line 49: Would these be better being in a client only class, or is there a use case where these may be used on the server? Either way Im not sure an inner class is the best solution here.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
```diff
@@ -0,0 +1,88 @@
...
+	 * Register a resource reloader for a given resource manager type.
+	 *
+	 * @param id the identifier of the resource reloader
+	 * @param reloader the resource reloader
+	 * @see #registerReloader(IdentifiableResourceReloader)
+	 * @see #addReloaderOrdering(Identifier, Identifier)
+	 */
+	default void registerReloader(Identifier id, ResourceReloader reloader) {
+		this.registerReloader(IdentifiableResourceReloader.wrap(id, reloader));
+	}
```
modmuss50: Do we really want two ways to do the same thing? The old `IdentifiableResourceReloadListener` is a very old API design that was fine at the time but isnt something I would jump to today. This has been one of the main pain points with the old API, I just wonder what the point of continuing to have it is.

The two solutions that come to mind are interface inject the `getId` onto all ResourceReloader's, or have it so they dont know their own ID (like most regular registry entries) with an API to look them up via their ID.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+import net.fabricmc.fabric.api.resource.v1.ResourceLoader;
+import net.fabricmc.fabric.api.resource.v1.reloader.IdentifiableResourceReloader;
+import net.fabricmc.fabric.api.resource.v1.reloader.ResourceReloaderKeys;
+import net.fabricmc.fabric.api.util.TriState;
+import net.fabricmc.fabric.impl.base.toposort.NodeSorting;
+import net.fabricmc.fabric.impl.base.toposort.SortableNode;
+import net.fabricmc.loader.api.FabricLoader;
+
+public final class ResourceLoaderImpl implements ResourceLoader {
+	private static final Logger LOGGER = LoggerFactory.getLogger("ResourceLoader");
```
modmuss50: Perfer to use the .class for the logger. Even if its the impl thats fine, without that there is no saying where `ResourceLoader` is coming from, the .class gives you the package.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+	private static final Map<ResourceType, ResourceLoaderImpl> IMPL_MAP = new EnumMap<>(ResourceType.class);
+
+	private static final boolean DEBUG_RELOADERS_IDENTITY = TriState.fromSystemProperty("fabric.resource_loader.debug.reloaders_identity")
+			.orElse(FabricLoader.getInstance().isDevelopmentEnvironment());
+	private static final boolean DEBUG_RELOADERS_ORDER = TriState.fromSystemProperty("fabric.resource_loader.debug.reloaders_order")
```
modmuss50: Use Boolean.getBoolean.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+	private final ResourceType type;
+
+	private ResourceLoaderImpl(ResourceType type) {
+		this.type = type;
+	}
+
+	@Override
+	public void registerReloader(IdentifiableResourceReloader reloader) {
+		if (!this.addedReloaderIds.add(reloader.getFabricId())) {
+			throw new IllegalStateException(
```
modmuss50: Dont do this in one `add` operation, always verify before doing the add. If a mod catches the IllegalStateException (on purpose or accident) things will be left in a bad state.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+		if (!this.addedReloaders.add(reloader)) {
+			throw new IllegalStateException(
+					"Resource reloader with previously unknown ID %s already in resource reloader set!"
+							.formatted(reloader.getFabricId())
+			);
+		}
+	}
+
+	@Override
+	public void addReloaderOrdering(@NotNull Identifier firstReloader, @NotNull Identifier secondReloader) {
```
modmuss50: No need for `@NotNull`, we assume everything is non null unless specified. Using `Objects.requireNonNull` is good though.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	public static List<ResourceReloader> sort(ResourceType type, List<ResourceReloader> listeners) {
+		if (type == null) {
+			return listeners;
+		}
+
+		ResourceLoaderImpl instance = get(type);
+
+		if (instance != null) {
```
modmuss50: Can this ever be null?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+
+		if (firstReloader.equals(secondReloader)) {
+			throw new IllegalArgumentException("Tried to add a phase that depends on itself.");
+		}
+
+		this.reloadersOrdering.add(new ReloaderOrder(firstReloader, secondReloader));
+	}
+
+	public static List<ResourceReloader> sort(ResourceType type, List<ResourceReloader> listeners) {
+		if (type == null) {
```
modmuss50, line 128: Question: Why would type be null, would it be better to handle that at the callsite?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
modmuss50: If not, perfer returning early.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+				);
+
+				if (DEBUG_RELOADERS_IDENTITY) {
+					LOGGER.warn(
+							"The resource reloader at {} does not implement IdentifiableResourceReloader "
+									+ "making ordering support more difficult for other modders.",
+							currentReloader.getClass().getName()
+					);
+				}
+			}
```
modmuss50: Nit: make a method for this, will make it easier to read the sorting code.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/DataPackContentsMixin.java
```diff
@@ -0,0 +1,56 @@
...
+			@Local(argsOnly = true) FeatureSet featureSet
+	) {
+		var list = new ArrayList<>(reloaders);
+		list.addFirst(
+				new SetupMarkerResourceReloader(
+						ResourceLoaderImpl.getWrapperLookup(reloaders),
+						featureSet
+				)
+		);
+		return list;
```
modmuss50: Was the original list not mutable? If so use `Collections.unmodifiableList` to make it not mutable again.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/KeyedResourceReloaderMixin.java
```diff
@@ -28,47 +27,35 @@
...
 	@Unique
 	private Identifier id;
-	@Unique
-	private Collection<Identifier> dependencies;
 
 	@Override
-	@SuppressWarnings({"ConstantConditions", "RedundantCast"})
-	public Identifier getFabricId() {
+	@SuppressWarnings({"ConstantConditions"})
+	public @NotNull Identifier getFabricId() {
```
modmuss50: Remove @NotNull applies everywhere else.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/LifecycledResourceManagerImplMixin.java
```diff
@@ -28,20 +28,20 @@
...
 import net.minecraft.resource.ResourceType;
 
-import net.fabricmc.fabric.impl.resource.loader.FabricLifecycledResourceManager;
+import net.fabricmc.fabric.impl.resource.v1.FabricLifecycledResourceManager;
 
 @Mixin(LifecycledResourceManagerImpl.class)
 public class LifecycledResourceManagerImplMixin implements FabricLifecycledResourceManager {
 	@Unique
-	private ResourceType fabric_ResourceType;
+	private ResourceType fabric$resourceType;
```
modmuss50: `fabric$` prefix not needed for `@Unique` field.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+
+	@Override
+	public void registerReloader(IdentifiableResourceReloader reloader) {
+		if (!this.addedReloaderIds.add(reloader.getFabricId())) {
+			throw new IllegalStateException(
+					"Tried to register resource reloader %s twice!".formatted(reloader.getFabricId())
+			);
+		}
+
+		if (!this.addedReloaders.add(reloader)) {
```
modmuss50: Same as above.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -0,0 +1,276 @@
...
+
+		return listeners;
+	}
+
+	/**
+	 * Sorts the given resource reloaders to satisfy dependencies.
+	 *
+	 * @param reloaders the resource reloaders to sort
+	 */
+	private void sort(List<ResourceReloader> reloaders) {
```
modmuss50, line 144: Have you thought about writing some unit tests for this? There is a lot of logic with this sorting that is imporant to make sure works as expected. Testmods make it hard to ensure its working as expected especially after future updates. In my particle renderer test I added some unit tests: https://github.com/FabricMC/fabric/pull/4854/files#diff-4b87f72c85dccd53541717078398c23c03b72c10bcf90f1468bf4944dadd0df0R36

modmuss50 (commented):
This is a great start, my main concern is around `IdentifiableResourceReloader` as im not sure we need it. Could it be interface injected onto ResourceReloader or even fully removed.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
LambdAurora: I'm not sure, it is useful to revert to default if the value is specified but not valid. I think an invalid value evaluating to `false` is much more confusing.

modmuss50: I disagree, I feel like the default value should be returned when the user specifies nothing. All values other than `true` or null then evaluate to false in a similar way to how Boolean handles it.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
LambdAurora: `IdentifiableResourceReloader` is still useful for Vanilla resource reloaders as those can't easily have a known IDs.

I'm not sure what the best solution is here, keeping the interface simplifies a lot, though interface-injecting it doesn't sound wise to me as this is something that should never have a default implementation.

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/TriState.java
LambdAurora: I'm a bit worried about people accidentally turning off stuff in their dev envs.

But sure, I'll change it.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
modmuss50: Is it needed at all? Could there just be a `getId(ResourceReloader)` API that returns the ID for a given loader, either the one it was registered with, the auto generated one, or the deprecated v0 one.

LambdAurora: This gets really difficult because of client-only reloaders. I cannot reference them in the main source set.

modmuss50: This can be easily handled by having an internal (none API) interface added via the client mixin to provide the vanilla IDs, in a similar way its done right now.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
LambdAurora: I've added a unit test. I might add some more later.

modmuss50: Great, thanks. Im mostly worried for when someone needs to touch this in the future.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.resource.v1.reloader.ResourceReloaderKeys;
+
+/**
+ * Interface for "identifiable" resource reloaders.
+ *
+ * @see ResourceReloaderKeys
+ */
+public interface IdentifiableResourceReloader extends ResourceReloader {
```
modmuss50: Maybe just calle this `FabricResourceReloader` now, the docs can also be removed.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * {@return the unique identifier of this resource reloader}
+	 */
+	Identifier getFabricId();
```
modmuss50: Can now be `fabric$getId`

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/IdentifiableResourceReloader.java
```diff
@@ -0,0 +1,39 @@
...
+
+	@Override
+	default String getName() {
```
modmuss50: Is this needed?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
```diff
@@ -49,26 +48,14 @@ static ResourceLoader get(ResourceType type) {
...
 	}
 
-	/**
-	 * Register a resource reloader for a given resource manager type.
-	 *
-	 * @param reloader the resource reloader
-	 * @see #registerReloader(Identifier, ResourceReloader)
-	 * @see #addReloaderOrdering(Identifier, Identifier)
-	 */
-	void registerReloader(IdentifiableResourceReloader reloader);
```
modmuss50: Nice, is there any use in having a public API for getting the ID of a given ResourceReloader? Can always add it later if needed.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/ResourceReloaderKeys.java
LambdAurora: I'm not sure why it would need to be a client-only class, especially as this only contains `Identifier` definitions.

The advantage of the inner classes is we end up with a very easy-to-find hierarchy for the identifiers. Typing `ResourceReloaderKeys` will naturally lead to all the vanilla resource reloader identifiers. Something two different classes (with one not being able to reference the other in javadoc) would not be able to achieve.

I'd prioritize developer ergonomics here.

modmuss50 (commented):
Almost there, I think this is much better. The API surface is much cleaner now.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/ResourceReloaderKeys.java
modmuss50: > I'm not sure why it would need to be a client-only class

It doesn't need to be, the main reason would be to only allow people to use them in their client code. Im not too worried about this was more of a thought/prompt to think about it. Im happy to stay with this if you are.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/ResourceLoader.java
LambdAurora: I do not think there is such use, at least for now.

It would be very unlikely for someone to stumble upon a resource reloader that is not theirs in their codebase.
Or if they depend on them, they would already know the identifier beforehand for ordering, and they'd use the shared state store to communicate data.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/IdentifiableResourceReloader.java
LambdAurora: Not needed, could've given more user-friendly names, but it also would hide the original name (being the class name).

modmuss50 (approved):

PR #4776, opened 2025-07-29
25w31a: `fabric-resource-loader-v0`

isXander:
Compiles and runs the game, but have not ran tests since it depends on other, broken, modules.

### DIFF
--- a/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
+++ b/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
@@ -35,7 +35,13 @@ import java.util.Set;
 import com.google.common.base.Charsets;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import com.mojang.serialization.Codec;
 import com.mojang.serialization.JsonOps;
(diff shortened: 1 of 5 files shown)


modmuss50 (approved):

fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ModResourcePackUtil.java
```diff
@@ -199,19 +205,20 @@ public static InputStream openDefault(ModContainer container, ResourceType type,
 		}
 	}
 
-	public static PackResourceMetadata getMetadataPack(int packVersion, Text description) {
-		return new PackResourceMetadata(description, packVersion, Optional.empty());
+	public static PackResourceMetadata getMetadataPack(class_11555 packVersion, Text description) {
+		return new PackResourceMetadata(description, new Range<>(packVersion));
```
isXander, line 209: Not sure whether FAPI should permit ranges here, haven't looked further upstream in the call chain.

modmuss50: I dont think there is a need to no, as it just needs to be a pack compatible with the current version being ran.

PR #4778, opened 2025-07-29
25w31a: `fabric-gametest-api-v1` compiles now

isXander:
Tests not ran, depends on other modules.


isXander:
For reference I did also run the module's client. Gets to title screen fine.

PR #4780, opened 2025-07-30
Fix Object[] allocations from WrapOperation

Ocelot5836:
This removes the overhead of creating a new `Object[]` every time the method is called.

### DIFF
--- a/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
+++ b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/BipedEntityRendererMixin.java
@@ -16,13 +16,12 @@
 
 package net.fabricmc.fabric.mixin.client.rendering;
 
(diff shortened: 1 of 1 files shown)


cputnam-a11y:
Have you profiled this to show that it was causing significant performance problems before, and that switching injectors remedied this? from a mixin standpoint this is a step backwards because it goes from referencing 0 locals to referencing 1 local. in addition, the java just in time compiler should quickly fix this if it becomes a problem.

Ocelot5836:
yes, on a JFR profile the allocations were significantly worse with the previous injector. It also went from referencing 2 locals to 1.

cputnam-a11y:
as far as locals, it's 0 -> 1 as the original WrapOperation captured the receiver and argument without dipping into locals. I'll ceed to @LlamaLad7 for addressing the profiling component, as it should probably be fixed upstream rather than here.

LlamaLad7:
Please show the profiling data. This should be trivial for the JIT to inline and I've previously made sure it does.

PR #4781, opened 2025-07-30
Fix dedicated server failing to spin up in tests

isXander:
It was timing out since the dedicated server didn't start due to no eula.


modmuss50 (changes requested):
No, this was done in loom. For now create it yourself.

PR #4787, opened 2025-08-01
Use `RegistryInfoGetter` to test `TagsPopulatedResourceCondition`

PepperCode1:
This should work based on how `PendingTag`s are implemented and how the `WrapperLookup` is constructed during a server resource reload. Tags retrieved through the `RegistryInfoGetter` will be present as expected, but their contents will be empty unless the registry is a reloadable registry (loot tables). Luckily this resource condition does not need to test for tag contents.

This also allows the condition to work even after the resource reload is done.


modmuss50 (approved):
Much nicer, thanks.

PR #4788, opened 2025-08-01
Fix Resource Loader's FMJ not declaring dependency on Fabric API Base

PepperCode1:
The module's build.gradle declares this dependency and the module itself uses classes from Base, so its FMJ should also declare this dependency.


modmuss50 (approved):
In a future PR we should look to add this to the FMJ validation task.

PR #4800, opened 2025-08-06
[1.21.1] Add super constructor for ForwardingBakedModel

AlphaMode:
I always thought ForwardingBakedModel not having a super constructor was a bit weird, this adds a super constructor while keeping api compat


PepperCode1 (approved):

modmuss50:
Is this change only for 1.21.1? If not it should go into 1.21.8 first and can then be back ported.

PepperCode1:
This is the backport. Versions 1.21.5+ already have this change.

PR #4814, opened 2025-08-17
Stacktrace remapping

modmuss50:
This PR is very much **work in progress**, im opening it to get early feedback/ideas before I get too far with it, please leave any thoughts in the comments.

The goal with the PR is to improve the debugging experiance for players and modders by remapping the stacktrace from intermediary to yarn names. This has been done before by a few mods, but as far as im aware all of them download the mappings at runtime, this isnt something I want to do for many reasons.

This PR introduces a custom binary file format for storing the minimal amount of mapping data required to remap stacktraces in a compact and easily parsable format. Currently this data is 400KB when compressed in the jar, which is a bit larger than I would have liked to have, if anyone has any suggestions for improving this please let me know. There needs to be a balance between size and complexity of parsing, the current format supports random reads allowing for very fast name lookup and minimal memory usage.

The binary file is generated at FabricAPI build time and embeded in the crash reports jar. I am also not sure on the best way to hook all of the stacktraces to remap them, I currently have a mixin in `CrashReport` but I expect this will mess a lot of stacktraces such as the ones that dont actually crash the game. I ran out of time today to properly look at that side of things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ $RECYCLE.BIN/
 .DS_Store
 
 !/buildSrc/src/**
+!/fabric-crash-report-info-v1/src/buildTools/**
--- a/fabric-crash-report-info-v1/build.gradle
+++ b/fabric-crash-report-info-v1/build.gradle
@@ -3,3 +3,39 @@ version = getSubprojectVersion(project)
 testDependencies(project, [
 	':fabric-command-api-v2'
 ])
+
+sourceSets {
+	buildTools
+	test {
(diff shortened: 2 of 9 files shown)


modmuss50:
After discussion, we wont be doing this as it makes it impossible to remap the stacktrace to the developers perfered mappings such as mojmap. There is also more complexity than I expected in actually wrapping the stacktraces to remap, it would likely be better done at a loader level but getting the mappings there is much harder.. Sorry for the many many people (including me) that wanted this.

PR #4818, opened 2025-08-18
Events for using an item on block

Fellteros:
Heya,
I noticed that there aren't any events for `Item.useOnBlock()` and `AbstractBlock.onUseWithItem()` methods. I recently tried to make a mod that would need this kind of callbacks, so I decided to implement it. It's mostly based on NeoForge's [UseItemOnBlockCallback](https://github.com/neoforged/NeoForge/blob/a5a86111e609a737ce560bf924e75e1062608684/src/main/java/net/neoforged/neoforge/event/entity/player/UseItemOnBlockEvent.java), with some minor changes (e.g. omitting the `ITEM_BEFORE_BLOCK` phase). 

I also added a little test mod, that allows you to change stone bricks to their mossy variant when right-clicking them with moss, and stripping the moss with any axe.

I added these events to the event-interaction-api module, let me know if there is any better place to move them to


Fuzss:
The events should allow for forcing `ActionResult.PASS` / `ActionResult.PASS_TO_DEFAULT_BLOCK_ACTION` respectively when cancelling vanilla behavior.

Ideally make the return value for the events `@Nullable`.

This has long been an issue with the existing interaction events not allowing for returning `ActionResult.PASS` properly. New interaction events shouldn't introduce this fault to begin with.

NeoForge also allows for cancelling vanilla regardless of the returned `ActionResult`.

Fellteros:
I made both methods `@Nullable` and changed the mixins to return any non-null value from the callbacks. This should fix the issues you pointed out.

fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/UseItemOnBlockEvents.java
```diff
@@ -0,0 +1,98 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Contains events triggered when using an item on a block. Fires both on client and server side.
+ *
+ * <p><i>This implementation is heavily inspired by NeoForge's <a href="https://github.com/neoforged/NeoForge/blob/1.21.x/src/main/java/net/neoforged/neoforge/event/entity/player/UseItemOnBlockEvent.java">UseItemOnBlockEvent</a></i>.
```
modmuss50: Was any code taken?

I dont think this is helpful to link as there is likely one way to do this.

Fellteros: I based the mixin injection points on the implementation

modmuss50: I'd remove this comment, I think its fair to say that there is only one true way to do this. I have nothing agianst neo but its unlikely a fabric developer is going to have use going to look at this code.

Fellteros:
Closed by #5040

PR #4819, opened 2025-08-18
Make utility methods of EnchantmentHelper and Enchantment public.

Patbox:
EnchantmentHelper and Enchantment both have few methods that are useful for mods adding enchantments (and used by Mojang in their specialized methods), but are private, requiring devs to duplicate them otherwise.

### DIFF
--- a/fabric-transitive-access-wideners-v1/build.gradle
+++ b/fabric-transitive-access-wideners-v1/build.gradle
@@ -236,6 +236,8 @@ def generateEnchantmentMethods(List<String> lines, FileSystem fs) {
 			}
 		}
 	}
+	lines.add("transitive-accessible class net/minecraft/enchantment/EnchantmentHelper\$Consumer")
+	lines.add("transitive-accessible class net/minecraft/enchantment/EnchantmentHelper\$ContextAwareConsumer")
 }
 
 ClassNode loadClass(Path path) {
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):

fabric-transitive-access-wideners-v1/build.gradle
```diff
@@ -223,6 +225,21 @@ def generateCreatorEntityTrackedDataMethods(List<String> lines, FileSystem fs) {
...
+	for (def node : [ loadClass(fs.getPath("net/minecraft/enchantment/EnchantmentHelper.class")), loadClass(fs.getPath("net/minecraft/enchantment/Enchantment.class")) ]) {
+		for (def method : node.methods) {
+			// All private inner methods of EnchantmentHelper
+			if ((method.access & Opcodes.ACC_PRIVATE) != 0 && !method.name.startsWith("method_")) {
+				lines.add("transitive-accessible method $node.name $method.name $method.desc")
+			}
+		}
+	}
+	lines.add("transitive-accessible class net/minecraft/enchantment/EnchantmentHelper\$Consumer")
+	lines.add("transitive-accessible class net/minecraft/enchantment/EnchantmentHelper\$ContextAwareConsumer")
```
Juuxel: Nit: no need to escape the `$` if you use normal single-quoted strings instead of interpolating double-quoted strings:
```suggestion
	lines.add('transitive-accessible class net/minecraft/enchantment/EnchantmentHelper$Consumer')
	lines.add('transitive-accessible class net/minecraft/enchantment/EnchantmentHelper$ContextAwareConsumer')
```

Juuxel (approved):

PR #4829, opened 2025-08-24
Allow for more flexible / configurable stripping behaviour.

Patbox:
Makes stripping behaviour bit more configurable, by removing the restriction of requiring AXIS property, while also allowing to define the way stripped state should be created. 

Existing StrippableBlockRegistry#register(Block input, Block stripped) still only copies AXIS property.
A new StrippableBlockRegistry#registerCopyState(Block input, Block stripped) makes it copy all compatible properties. 
If someone needs bit more control, they can use StrippableBlockRegistry#register(Block input, Block stripped, StrippingTransformer transformer) to define how and what is copied.

### DIFF
--- a/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/AxeItemMixin.java
+++ b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/AxeItemMixin.java
@@ -16,10 +16,12 @@
 
 package net.fabricmc.fabric.mixin.content.registry;
 
+import java.util.function.Function;
+
+import com.llamalad7.mixinextras.sugar.Local;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Inject;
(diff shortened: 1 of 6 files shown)


Shnupbups (approved):

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/mixin/content/registry/AxeItemMixin.java
```diff
@@ -0,0 +1,42 @@
...
+import net.minecraft.block.BlockState;
+import net.minecraft.item.AxeItem;
+
+import net.fabricmc.fabric.api.registry.StrippableBlockRegistry;
+import net.fabricmc.fabric.impl.content.registry.StrippableBlockRegistryImpl;
+
+@Mixin(AxeItem.class)
+public class AxeItemMixin {
+	// Lambda in getStrippedState
+	@Inject(method = "method_34717", at = @At("HEAD"), cancellable = true)
```
Bawnorton: Composing the target lambda would be more stable than injecting into the synthetic method, also viable based on the nature of this injection pattern.

Earthcomputer (approved):

Bawnorton (approved):

modmuss50 (approved):

PR #4838, opened 2025-08-29
Make events cancellable

Fellteros:
This PR allows you to cancel any event and thus prevent executing code of listeners or vanilla methods.

## Motivation

Some events (as pointed out [here](https://github.com/FabricMC/fabric/pull/4818#issuecomment-3205691759) by Fuzzs) do not allow returning certain values because they are used as an indicator that the vanilla logic should be called (e.g., [UseBlockCallback](https://github.com/FabricMC/fabric/blob/e96ab00631139381f37bcf0b52765948f99aa094/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/UseBlockCallback.java#L38-L50), where you cannot return the `ActionResult.PASS`) or they cancel the vanilla logic completely if the result is different from the vanilla fallback.
This PR aims to fix this by allowing you to cancel the event by checking for the cancellation status instead of using a set return value for that, and providing the option to return any result you want.

## Implementation

Cancellable events can be implemented similarly to the current events by using new methods in `EventFactory` (`createCancellable` and its overload), that provide the option to access the event's `cancelStatus` via a lambda and using it to cancel the event.

## Conflicts with existing API

Since cancellable events are created through a completely new methods and a new constructor in `ArrayBackedEvent`, already existing implementations and API can function without any changes. Existing events can also be relatively easily changed to being cancellable.
To help with implementing or using cancellable events, I created a Javadoc guide with basic implementation rules and examples, that should help contributors and modders with understanding the concept. 

## Testing

I've created a test event in `EventTests` to prove the functionality. If that might not be enough, feel free to ping and ask me. I'm also aware that this PR might have flaws I haven't noticed, so please notify me if you find any.


modmuss50:
Im not sure that this is a good solution to the problem, it has limitations such as mods not being able to handle the event if another mod has cancelled before it, may suffer from issues caused by the random mod ordering and some events may need to be canclled with a result. For a change like this it really is worth discussing the idea before spending time writing code. I'm happy to contiune to discuss this and get the input of others (as I might be wrong :D), discord is a good place to get some real time input on your ideas.

We generally handle this pattern by using multiple events, one for canceling the action, one for handling the action (if not cancled) and maybe an event for when the action was canceled. This allows for an API design that fits each specific event instead of trying to apply a generic idea such as this one.

UseBlockCallback could do with some improvements but I dont think this is the solution, either way the API would need to be deprecated/replaced.

PepperCode1:
To elaborate, this change would not make it any easier to allow interaction events to return any `ActionResult` because existing users already expect, for example, that returning pass cancels the event instead of actually returning pass. New events with the new behavior would still have to be made. That said, I agree with Fuzs' comment.

Player3324:
Cancelling events is deliberately not a thing because it's inferior to our approach of having an "allow" type event for cancellation followed by the actual "it's happening" event, possibly also a third "it's cancelled" event. This is the only way to correctly notify subscribers of what's happening.

With cancellation you can't really do anything reliably like take energy out of a tool because a lower priority event might stop the activity but doesn't roll it back properly. Workarounds like having to actively notify+check for cancellation put too much burden on the mods and ask for bugs on their side.

Patbox:
This implementation isn't compatible with events ran on multiple threads (which is something which can happen with existing fabric events, including UseBlockCallback)

Fellteros:
Thank you all for the criticism, looks like I haven't consider a lot of things :) It seems that this approach isn't very suitable, and would require changing the whole entirety of it, so I'll close this for now.

PR #4841, opened 2025-08-31
Implement an opt-in packet splitter.

Patbox:
Implements a opt-in packet splitting logic, that can be used for any packet (through default implementation only handles CustomPayloadS2C and only works for S2C packets).

It's not enabled by for all packets, to prevent unwanted format breakage.
Functionally it's implemented in similar way to how vanilla handles Bundle packets, through work for both play and configuration stages. Outside the packet pipeline, the game / mods will only deal with their own packets. Aka entire splitting and merging is handled within packet pipeline.

Quick test (forcing all payload packets to go through it, with chunk size of 20 bytes) seemed to work just fine, but more extensive testing would be a good idea probably.

Currently splittable packets always go through packet splitting process, through ideally I want to get rid of that.

(Also keeping it as draft until I gather some feedback and more testing)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
```diff
@@ -62,6 +72,22 @@ public <T extends CustomPayload> CustomPayload.Type<? super B, T> register(Custo
...
+	public <T extends CustomPayload> CustomPayload.Type<? super B, T> registerSplittable(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec) {
+		return registerSplittable(id, codec, FabricPacketSplitter.SAFE_SPLIT_SIZE);
+	}
+
+	@Override
+	public <T extends CustomPayload> CustomPayload.Type<? super B, T> registerSplittable(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec, int splitThreshold) {
+		if (this.side == NetworkSide.CLIENTBOUND) {
+			packetSplitThreshold.put(id, splitThreshold);
+		} else {
+			LOGGER.warn("The " + this.name + " doesn't support packet splitting! Using registerSplittable will have no effect!");
```
Earthcomputer: Why not?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
```diff
@@ -0,0 +1,76 @@
...
+
+import net.minecraft.network.handler.EncoderHandler;
+import net.minecraft.network.packet.Packet;
+import net.minecraft.network.packet.s2c.common.CustomPayloadS2CPacket;
+import net.minecraft.network.state.NetworkState;
+
+import net.fabricmc.fabric.mixin.networking.accessor.EncoderHandlerAccessor;
+
+public class FabricPacketSplitter extends MessageToMessageEncoder<Packet<?>> {
+	public static final int SAFE_SPLIT_SIZE = 1048576 - 2000;
```
Earthcomputer: This needs a different value for serverbound. Plus you can use the constant:
```suggestion
	public static final int SAFE_SPLIT_SIZE = CustomPayloadS2CPacket.MAX_PAYLOAD_SIZE - 2000;
```

This value is also quite conservative. Wouldn't it be worth working out exactly what the maximum safe split size is? Currently it's `MAX_PAYLOAD_SIZE - 10`, 5 bytes each for the worst-case varint for `splitId` and `part`. However, if you implement my suggestion to remove the redundant data then this value should be equal to `MAX_PAYLOAD_SIZE`.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricSplitDataPacketPayload.java
```diff
@@ -0,0 +1,50 @@
...
+import io.netty.buffer.ByteBuf;
+
+import net.minecraft.network.codec.PacketCodec;
+import net.minecraft.network.encoding.VarInts;
+import net.minecraft.network.packet.CustomPayload;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.networking.FabricPacketsImpl;
+
+public record FabricSplitDataPacketPayload(int splitId, int part, ByteBuf byteBuf) implements FabricSplitPacketPayload {
```
Earthcomputer: Lots of redundant data here. A simple split start packet with an integer for the number of payloads to follow is enough. No need for `splitId` and `part` anywhere.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
```diff
@@ -62,6 +72,22 @@ public <T extends CustomPayload> CustomPayload.Type<? super B, T> register(Custo
...
 
+	@Override
+	public <T extends CustomPayload> CustomPayload.Type<? super B, T> registerSplittable(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec) {
+		return registerSplittable(id, codec, FabricPacketSplitter.SAFE_SPLIT_SIZE);
+	}
+
+	@Override
+	public <T extends CustomPayload> CustomPayload.Type<? super B, T> registerSplittable(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec, int splitThreshold) {
+		if (this.side == NetworkSide.CLIENTBOUND) {
+			packetSplitThreshold.put(id, splitThreshold);
```
Earthcomputer: This should be checked to make sure it's not `> SAFE_SPLIT_SIZE`.

Earthcomputer (changes requested):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
Patbox: You don't want people OoMing servers with this.

Earthcomputer:
When porting to 1.21.9, the registry sync module should be changed to use this too. (We can't change it for 1.21.8 because of the protocol change, but on the 1.21.9 update vanilla is changing the protocol anyway so it's fine).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
Patbox: I didn't do it as it was private. Through maybe I selected wrong source. Through yeah it could be lowered.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricSplitDataPacketPayload.java
Patbox: This is mostly to make it easier to find what went wrong if packets get reordered for whatever reason. Which while might not happen in vanilla, could be triggered by other mods / proxies

Earthcomputer: Proxies should not be reordering, dropping arbitrarily, or otherwise altering packets that they don't recognize. Registry sync has sent order-sensitive packet split for a long time and as far as I know there has never been an issue.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/FabricPacketsImpl.java
```diff
@@ -0,0 +1,39 @@
...
+	public static final String MOD_ID = "fabric-networking-api-v1";
+	public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
+
+	public static void init() {
+		PayloadTypeRegistry.configurationS2C().register(FabricSplitStartPacketPayload.ID, FabricSplitStartPacketPayload.CODEC);
+		PayloadTypeRegistry.playS2C().register(FabricSplitStartPacketPayload.ID, FabricSplitStartPacketPayload.CODEC);
+		PayloadTypeRegistry.configurationS2C().register(FabricSplitEndPacketPayload.ID, FabricSplitEndPacketPayload.CODEC);
+		PayloadTypeRegistry.playS2C().register(FabricSplitEndPacketPayload.ID, FabricSplitEndPacketPayload.CODEC);
+		PayloadTypeRegistry.configurationS2C().register(FabricSplitDataPacketPayload.ID, FabricSplitDataPacketPayload.CODEC);
+		PayloadTypeRegistry.playS2C().register(FabricSplitDataPacketPayload.ID, FabricSplitDataPacketPayload.CODEC);
```
modmuss50: I dont think you need 3 packets for this, the registry sync does it with 1, the final packet being empty, im not saying thats the solution here but its worth thinking about.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadS2CPacketMixin.java
```diff
@@ -63,4 +77,16 @@ private static PacketCodec<PacketByteBuf, CustomPayload> wrapConfigCodec(CustomP
 		fabricCodec.fabric_setPacketCodecProvider((packetByteBuf, identifier) -> PayloadTypeRegistryImpl.CONFIGURATION_S2C.get(identifier));
 		return codec;
 	}
+
+	@Override
+	public void fabric_split(int id, NetworkState<?> state, ChannelHandlerContext channelHandlerContext, EncoderHandler<?> encoder, Packet<?> packet, Consumer<Packet<?>> consumer) throws Exception {
```
modmuss50: For modded payloads that are split they should still be sent via their own channel ID, and not wrapped up into fabric's. Currently I think this breaks canSend for these packets.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
modmuss50: Always crash if a mod does something wrong, dont leave it in a bad state no one reads the logs.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
```diff
@@ -43,6 +43,39 @@ public interface PayloadTypeRegistry<B extends PacketByteBuf> {
...
+	 *
+	 * <p>Payload types registered with this method will be split into multiple packets,
+	 * if it ends up exceeding vanilla size limit.
+	 *
+	 * @param id    the id of the payload type
+	 * @param codec the codec for the payload type
+	 * @param <T>   the payload type
+	 * @return the registered payload type
+	 */
+	<T extends CustomPayload> CustomPayload.Type<? super B, T> registerSplittable(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec);
```
modmuss50: The modder should have to specify a reasonable max size.

modmuss50: Is `registerSplittable` a good name for this? Should the modder care that the packet is split, or would something more generic such as `registerLarge` be better? The splitting is likely an impl detail they dont need to care about.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
```diff
@@ -43,6 +43,39 @@ public interface PayloadTypeRegistry<B extends PacketByteBuf> {
...
+	 * <p>This must be done on both the sending and receiving side, usually during mod initialization
+	 * and <strong>before registering a packet handler</strong>.
+	 *
+	 * <p>Payload types registered with this method will be split into multiple packets,
+	 * if it ends up exceeding provided size limit.
+	 *
+	 * @param id             the id of the payload type
+	 * @param codec          the codec for the payload type
+	 * @param <T>            the payload type
+	 * @param splitThreshold the maximum size of payload before splitting
```
modmuss50: Whats the benefit/usecase of being able to specify this size? Surely you would only want to split the packet when absolutely necessary.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/FabricPacketsImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import net.fabricmc.fabric.api.networking.v1.PayloadTypeRegistry;
+import net.fabricmc.fabric.impl.networking.splitter.FabricSplitDataPacketPayload;
+import net.fabricmc.fabric.impl.networking.splitter.FabricSplitEndPacketPayload;
+import net.fabricmc.fabric.impl.networking.splitter.FabricSplitStartPacketPayload;
+
+public final class FabricPacketsImpl {
+	public static final String MOD_ID = "fabric-networking-api-v1";
```
modmuss50: NetworkingImpl.MOD_ID already exists.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/FabricPacketsImpl.java
```diff
@@ -0,0 +1,39 @@
...
+	public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
```
modmuss50: Pass the class to loggers.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/FabricPacketsImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+	public static void init() {
```
modmuss50: Id just move all this to NetworkingImpl.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
```diff
@@ -31,20 +35,26 @@
...
-	public static final PayloadTypeRegistryImpl<RegistryByteBuf> PLAY_C2S = new PayloadTypeRegistryImpl<>(NetworkPhase.PLAY, NetworkSide.SERVERBOUND);
-	public static final PayloadTypeRegistryImpl<RegistryByteBuf> PLAY_S2C = new PayloadTypeRegistryImpl<>(NetworkPhase.PLAY, NetworkSide.CLIENTBOUND);
+	public static final PayloadTypeRegistryImpl<PacketByteBuf> CONFIGURATION_C2S = new PayloadTypeRegistryImpl<>("configuration_c2s", NetworkPhase.CONFIGURATION, NetworkSide.SERVERBOUND);
+	public static final PayloadTypeRegistryImpl<PacketByteBuf> CONFIGURATION_S2C = new PayloadTypeRegistryImpl<>("configuration_s2c", NetworkPhase.CONFIGURATION, NetworkSide.CLIENTBOUND);
+	public static final PayloadTypeRegistryImpl<RegistryByteBuf> PLAY_C2S = new PayloadTypeRegistryImpl<>("play_c2s", NetworkPhase.PLAY, NetworkSide.SERVERBOUND);
+	public static final PayloadTypeRegistryImpl<RegistryByteBuf> PLAY_S2C = new PayloadTypeRegistryImpl<>("play_s2c", NetworkPhase.PLAY, NetworkSide.CLIENTBOUND);
+
+	public static final Logger LOGGER = LoggerFactory.getLogger(FabricPacketsImpl.MOD_ID);
```
modmuss50: Use the class

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadS2CPacketMixin.java
```diff
@@ -63,4 +77,16 @@ private static PacketCodec<PacketByteBuf, CustomPayload> wrapConfigCodec(CustomP
...
+		int size = (state.id() == NetworkPhase.CONFIGURATION ? PayloadTypeRegistryImpl.CONFIGURATION_S2C : PayloadTypeRegistryImpl.PLAY_S2C).getSplittingThreshold(this.payload.getId());
```
modmuss50: Nit: split this into 2 lines, create another variable for the payload type registry. easier to read.

modmuss50 (changes requested):
This has a lot of scope for unit tests, see `DirectRegistryPacketHandlerTest` for where we test the registry sync packet splitting.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/FabricPacketsImpl.java
Patbox: Using distinct start, data and end packet make it bit clearer for debugging as they create clear separation, but it should be simple to handle as single packet

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/CustomPayloadS2CPacketMixin.java
Patbox: Mostly will need to look if I can somewhat cleanly skip the `EncoderHandler` (and things added by other mods?), while still running by rest of the pipeline. Maybe by using dummy Packet implementation?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
Patbox: Yeah not sure about that name myself. Was thinking about defining it separately too, but that might be more clunky

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
Patbox: Honestly probably not. Will replace it with max total size instead

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
Patbox: Since there was interest in making it work both ways (while defining max size), I will probably drop this limitation

modmuss50: I think thats fine as long as you require the mod to specify the max size. If they put something stupid its on them.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
```diff
@@ -62,6 +75,15 @@ public <T extends CustomPayload> CustomPayload.Type<? super B, T> register(Custo
 		return payloadType;
 	}
 
+	@Override
+	public <T extends CustomPayload> CustomPayload.Type<? super B, T> registerLarge(CustomPayload.Id<T> id, PacketCodec<? super B, T> codec, int maxPayloadSize) {
+		// Defines max packet size, increased by length of packet's Identifier to cover full size of CustomPayloadX2YPackets.
+		int identifierSize = ByteBufUtil.utf8MaxBytes(id.id().toString());
+		this.maxPacketSize.put(id.id(), maxPayloadSize + VarInts.getSizeInBytes(identifierSize) + identifierSize + 5 * 2);
```
modmuss50: This will overflow if (when) mods use Int.MAX_VALUE

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
```diff
@@ -0,0 +1,84 @@
...
+import net.minecraft.network.packet.CustomPayload;
+import net.minecraft.network.packet.Packet;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.networking.PayloadTypeRegistryImpl;
+import net.fabricmc.fabric.mixin.networking.accessor.EncoderHandlerAccessor;
+
+public class FabricPacketSplitter extends MessageToMessageEncoder<Packet<?>> {
+	public static final int SAFE_S2C_SPLIT_SIZE = 1048576 - 32;
+	public static final int SAFE_C2S_SPLIT_SIZE = 32767 - 32;
```
Earthcomputer: Let's go with the actual space requirements and not some arbitrary 32 number.
```suggestion
	// The space reserved for use by FabricSplitDataPacketPayload
	private static final int DATA_PAYLOAD_RESERVED_SPACE = 5; // 5 bytes for `part` varint
	public static final int SAFE_S2C_SPLIT_SIZE = 1048576 - DATA_PAYLOAD_RESERVED_SPACE;
	public static final int SAFE_C2S_SPLIT_SIZE = 32767 - DATA_PAYLOAD_RESERVED_SPACE;
```

Although maybe the `DATA_PAYLOAD_RESERVED_SPACE` constant should go in the `FabricSplitDataPacketPayload` class.

I would also say that the constants in `CustomPayloadS2CPacket` and -`C2SPacket` are worth an access widener so we can use them here, although maybe some people would disagree.

Earthcomputer (changes requested):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
Patbox: Decided to go with access widener, since it looks nicer and gives source to it

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraft.network.packet.Packet;
+import net.minecraft.network.packet.c2s.common.CustomPayloadC2SPacket;
+import net.minecraft.network.packet.s2c.common.CustomPayloadS2CPacket;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.impl.networking.PayloadTypeRegistryImpl;
+import net.fabricmc.fabric.mixin.networking.accessor.EncoderHandlerAccessor;
+
+public class FabricPacketSplitter extends MessageToMessageEncoder<Packet<?>> {
+	private static final int DATA_HEADER_SIZE = 10;
```
Earthcomputer: It's 5 now, not 10. There definitely needs to be a comment explaining where this number comes from. I would also move it into the split data packet class itself, as it makes it easier to remember to change this value if the header format ever changes.

Patbox: It is 10, 5 for part and 5 for writing the size of sent bytebuf

Earthcomputer: If you're sure that size is actually counted towards the packet size limit, I thought it wasn't. Anyway, this reinforces the need for a comment explaining where the number comes from.

Patbox: Ok, moved it to FabricSplitDataPayload, commented it and made it use constant from VarInts class

modmuss50:
Hey @Patbox I been thinking about this PR, now that 1.21.9 is out of the way its the next "big" thing I want to get in. I think i had really had 2/3 concerns/queries with the design:

1. Should it be lower level than just building on top of the existing API (as you have today in your PR) - Yes I think this is a reasonable way to go, as I think like you said it allows other mods to inspect these packets as-if they are one. If there is a problem this can be easily changed. Nothing to do here.

2. Do we really need 3 different payloads to carry this data, I don't see why this cannot just be done with a single payload type. Maybe each payload should have a small header to dictate the type, packets that don't need to split should still fit into 1.

2. Should large packets use their own channel ids? I think they should especially if its changed to only be 1 payload type, I'm not sure what the benefit of putting them all over a `fabric:split` or whatever is.

I'd be happy to help look at this or maybe even just try out some of my ideas in a separate branch.

Patbox:
For point 2:
It no longer wraps packets that don't require splitting without also duplicating the serialization (via packet passthrough). I feel using 3 seperate packet types makes implementation slighly cleaner, but it should be possible to use only one.
For point 3:
Using custom payload simplifies implementation, allows packet inspection to more clearly debug things and keeps things more compatible. Extra bonus, through probably one that won't be used by fabric api, is that it also allows to use this system to split vanilla packets too, which might be something some mods might want to enable in future.

modmuss50:
> I feel using 3 seperate packet types makes implementation slighly cleaner, but it should be possible to use only one.

The impl might be a tiny bit easier, but I think the design will be easier to understand. Might be worth making a note somewhere of the packet format you choose. Im happy to look at it before you actually implement it, dont want you to keep going back and forth with this.

> allows packet inspection to more clearly debug things and keeps things more compatible

Im not sure I agree with this, in its current form from an external POV all packets will be sent over the fabric channel and without inspecting the packet data there is no way to tell what mod is sending that data. Please correct me if im wrong but I think the main use case for inspecting packet names is for profiling, this makes that harder.

> Extra bonus, through probably one that won't be used by fabric api, is that it also allows to use this system to split vanilla packets too, which might be something some mods might want to enable in future.

As we said this might be something we want to look at supporting after this PR has done in. I would imagine there to be a `fabric:vanilla_large` or whatever packet for this. This can still happen with the above change.

Player3324:
I'd keep it in `fabric:split`, but also strip the redundant data.

Both the first and last packet can be identified by a counter and carry part of the data up to their limit. Neither the part number nor the part data size need to be transmitted, the packets are already framed externally. It might be a good idea to disallow undersized parts to avoid potential abuse vectors and encoding variance.

It also appears that the current implementation sends the embedded  identifier and size twice. Otherwise I'd expect https://github.com/FabricMC/fabric/pull/4841/files#diff-026fe5703b81c5e0773f9aa169291cae9ef4d824254fdeaf55b319b76d9481d8R120 to use the respective members. It should be easy enough to parse the embedded packet header instead of duplicating the information.

We should decide whether a future protocol revision is to be handled by a different id (other than `fabric:split`) or a byte in the first packet of each sequence. Using the channel and what I wrote above makes everything rather simple, we wouldn't add any headers and merely emit slices of the original packet bytes including its size+id header within `fabric:split`.

Patbox:
Ok went will all suggestions here. So now:
- There's only one payload type, `fabric:split` that holds boolean flag for marking if it's finished and split ByteBuf data.
- The type / limits are now read from the first split packet's bytebuf, limiting it to only work with CustomPayload packets.
- Removed information about final packet size and part, as they could be received mostly from ByteBuf methods or weren't used outside of extra validation

Patbox:
For versioning doing it via packet channel should work just fine, because honestly there won't be much need to add any metadata anyway.

Player3324:
>     * The type / limits are now read from the first split packet's bytebuf, limiting it to only work with CustomPayload packets.

If you implement https://github.com/FabricMC/fabric/pull/4841/files#diff-026fe5703b81c5e0773f9aa169291cae9ef4d824254fdeaf55b319b76d9481d8R71 differently it'd just work (tm) with any packet, no? What in the protocol requires it to be embedding custom payload exclusively?

Patbox:
Mostly as it reads the packet size limit from payload id. Through yes, it should be possible to change that with an extra bit of logic

Player3324:
So vanilla packets are [length,id,data], it seems like the EncoderHandler.encode doesn't include the size yet, but everything else. I'd manually prefix the length then to restore the full packet bytes, much like how MC does it later in the pipeline. The length is readily available at https://github.com/FabricMC/fabric/pull/4841/files#diff-351cb844fd0b49a8fc53c0fc3a763d4d6786d089d170000eb5da1f72822b3f85R67.

You don't need `finished` if you check the data size against the original length once that's being transmitted. Merger.byteBuf should probably be presized.

Patbox:
Feels slightly uglier on sending part, but changed it. Also implemented packet type validation that currently only allows CustomPayload packets to get merged. It should be super simple to tweak that method to allow vanilla packet merging without changing the format (through naturally client/server would still require an update).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketSplitter.java
```diff
@@ -0,0 +1,93 @@
...
+
+		if (buf.readableBytes() > maxPacketSize) {
+			throw new EncoderException("Packet '" + packetId + "' may not be larger than " + maxPacketSize + " bytes!");
+		}
+
+		// First packet split with added packet size
+		ByteBuf firstSplit = Unpooled.buffer(maxChunkSize);
+		VarInts.write(firstSplit, buf.readableBytes());
+		// First slice needs to be slightly smaller to accommodate the size prefix
+		firstSplit.writeBytes(buf.readSlice(maxChunkSize - VarInts.getSizeInBytes(buf.readableBytes())));
```
Player3324: nit: You should be able to query firstSplit for its remaining capacity or at least the written bytes instead of computing this manually. Not sure why the loop afterwards isn't something like `while (buf.readableBytes() > 0) consumer.accept(packetConstructor.apply(new FabricSplitPacketPayload(buf.readSlice(Math.min(buf.readableBytes(), maxChunkSize)))));`. No mandatory tail packet.

Patbox:
(Oops wrong pr hah)

modmuss50 (approved):
Im happy with this. Most of this is implimentation detail that can be changed at a later date (likely across MC versions where we dont guarantee network compatibility).

I think there is a lot of scope to unit test this kind of code but im not going to hold up the PR any longer for that, it can be done in a follow up PR.

Player3324:
I don't like how the inner logic potentially sends one packet too many, even if that's normally prevented by the outer logic, which might cause us issues down the line. Changing the code makes it simpler and is easy enough.
This logic for the tail of `genericPacketSplitter` should do it:

```java
// First packet split with added packet size

ByteBuf firstSplit = Unpooled.buffer(maxChunkSize);
VarInts.write(firstSplit, buf.readableBytes());
// First slice needs to be slightly smaller to accommodate the header (by the already written data amount)
firstSplit.writeBytes(buf, maxChunkSize - firstSplit.readableBytes());

consumer.accept(packetConstructor.apply(new FabricSplitPacketPayload(firstSplit)));

// Remaining packets, as needed to send everything
while (buf.isReadable()) {
	consumer.accept(packetConstructor.apply(new FabricSplitPacketPayload(buf.readSlice(Math.min(buf.readableBytes(), maxChunkSize))));
}
```

It also uses a few more fitting ByteBuf overloads.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketMerger.java
```diff
@@ -0,0 +1,144 @@
...
+
+			CustomPayload payload = packet instanceof GenericPayloadAccessor accessor ? accessor.fabric_payload() : null;
+
+			if (payload == null) {
+				throw new DecoderException("Received '" + packet.getPacketType().id() + "' packet, while expecting 'minecraft:custom_payload'!");
+			}
+
+			if (!(payload instanceof FabricSplitPacketPayload splitPacketPayload)) {
+				throw new DecoderException("Expected '" + FabricSplitPacketPayload.ID.id() +"' payload packet, but received '" + payload.getId().id() + "'!");
+			}
```
ishland, line 65: Are there any reasons to reject any non split packets here? This currently breaks existing mods that split packets into multiple subsystems based on packet type and multiplex them with priority scheduling.

Patbox: Mostly validation that everything works correctly and all data gets received (and no one tries to send them as unintender). Since a lot of mods do depend on packet order being right (from vanilla, fabric, other libraries and their own) it just made sense to check it.

Player3324 (approved):

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/splitter/FabricPacketMerger.java
Player3324: For now I think it's fair to expect any custom reordering scheme to keep fabric:split packets together.

Nevertheless I do see a point in relaxing the check/requirement to enable some advanced trickery. As long as no second stream of fabric:split packets intersects the first one, the merging would work fine. Someone could then put e.g. motion updates in the middle of bulk data streams for congested connections and potentially improve responsiveness. This would however complicate handling for proxies as they'd have to support those mixed packets as well.

PR #4842, opened 2025-08-31
Cleanup FluidRendererMixin

Earthcomputer:
Also changed the type of the `transparencyForOverlay` registry map from a concurrent map to the faster `Object2BooleanOpenHashMap`, by using the thread safe `getOrDefault` method over `computeIfAbsent`.

### DIFF
--- a/fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
+++ b/fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
@@ -18,14 +18,14 @@ package net.fabricmc.fabric.impl.client.rendering.fluid;
 
 import java.util.IdentityHashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
+import it.unimi.dsi.fastutil.objects.Object2BooleanMap;
+import it.unimi.dsi.fastutil.objects.Object2BooleanOpenHashMap;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.block.Block;
 import net.minecraft.block.LeavesBlock;
-import net.minecraft.block.TranslucentBlock;
+import net.minecraft.block.TransparentBlock;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.color.world.BiomeColors;
 import net.minecraft.client.render.block.FluidRenderer;
@@ -45,7 +45,7 @@ import net.fabricmc.fabric.api.client.render.fluid.v1.FluidRenderHandlerRegistry
 public class FluidRenderHandlerRegistryImpl implements FluidRenderHandlerRegistry {
 	private final Map<Fluid, FluidRenderHandler> handlers = new IdentityHashMap<>();
 	private final Map<Fluid, FluidRenderHandler> modHandlers = new IdentityHashMap<>();
-	private final ConcurrentMap<Block, Boolean> overlayBlocks = new ConcurrentHashMap<>();
+	private final Object2BooleanMap<Block> transparencyForOverlay = new Object2BooleanOpenHashMap<>();
 
 	{
 		handlers.put(Fluids.WATER, WaterRenderHandler.INSTANCE);
@@ -77,12 +77,12 @@ public class FluidRenderHandlerRegistryImpl implements FluidRenderHandlerRegistr
 
 	@Override
 	public void setBlockTransparency(Block block, boolean transparent) {
-		overlayBlocks.put(block, transparent);
+		transparencyForOverlay.put(block, transparent);
 	}
 
 	@Override
 	public boolean isBlockTransparent(Block block) {
-		return overlayBlocks.computeIfAbsent(block, k -> k instanceof TranslucentBlock || k instanceof LeavesBlock);
+		return transparencyForOverlay.getOrDefault(block, block instanceof TransparentBlock || block instanceof LeavesBlock);
 	}
 
 	public void onFluidRendererReload(FluidRenderer renderer, Sprite[] waterSprites, Sprite[] lavaSprites, Sprite waterOverlay) {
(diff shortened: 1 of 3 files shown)


fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/fluid/FluidRenderHandlerRegistryImpl.java
```diff
@@ -77,12 +77,12 @@ public void register(Fluid fluid, FluidRenderHandler renderer) {
...
 	@Override
 	public void setBlockTransparency(Block block, boolean transparent) {
-		overlayBlocks.put(block, transparent);
+		transparencyForOverlay.put(block, transparent);
 	}
 
 	@Override
 	public boolean isBlockTransparent(Block block) {
-		return overlayBlocks.computeIfAbsent(block, k -> k instanceof TranslucentBlock || k instanceof LeavesBlock);
+		return transparencyForOverlay.getOrDefault(block, block instanceof TransparentBlock || block instanceof LeavesBlock);
```
Juuxel: Why the change from TranslucentBlock to TransparentBlock?

Earthcomputer: Oops by accident

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
```diff
@@ -76,58 +77,55 @@ public void onHeadRender(BlockRenderView view, BlockPos pos, VertexConsumer vert
...
-	}
-
-	@ModifyVariable(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/block/FluidRenderer;isSameFluid(Lnet/minecraft/fluid/FluidState;Lnet/minecraft/fluid/FluidState;)Z"), method = "render", ordinal = 0)
-	public Sprite[] modSpriteArray(Sprite[] chk) {
+	@ModifyVariable(
+			method = "render",
+			at = @At("STORE"),
+			ordinal = 0
+	)
+	public Sprite[] modSpriteArray(Sprite[] original) {
```
LlamaLad7, line 85: Could avoid the local by using MEV on both the water and lava sprite accesses, idk if that's better since the local type is reasonably specific

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
```diff
@@ -76,58 +77,55 @@ public void onHeadRender(BlockRenderView view, BlockPos pos, VertexConsumer vert
...
 		FluidRenderHandlerInfo info = FluidRenderingImpl.getCurrentInfo();
-		return info.handler != null ? info.sprites : chk;
+		return info.handler != null ? info.sprites : original;
 	}
 
-	@ModifyVariable(at = @At(value = "CONSTANT", args = "intValue=16", ordinal = 0, shift = At.Shift.BEFORE), method = "render", ordinal = 0)
-	public int modTintColor(int chk, BlockRenderView world, BlockPos pos, VertexConsumer vertexConsumer, BlockState blockState, FluidState fluidState) {
+	@ModifyVariable(method = "render", at = @At("STORE"), ordinal = 0)
```
LlamaLad7: I think this would definitely benefit from MEV on both branches of the ternary, an int local is very generic

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
```diff
@@ -76,58 +77,55 @@ public void onHeadRender(BlockRenderView view, BlockPos pos, VertexConsumer vert
...
-	public Block getOverlayBlock(BlockState state) {
-		Block block = state.getBlock();
-		neighborBlock.set(block);
-
-		// An if-statement follows, we don't want this anymore and 'null' makes
-		// its condition always false (due to instanceof)
-		return null;
-	}
+	@Definition(id = "getFrameU", method = "Lnet/minecraft/client/texture/Sprite;getFrameU(F)F")
+	@Expression("@(?).getFrameU(0.0)")
```
LlamaLad7: Don't target the wildcard here, use a concrete local. This ensures that the selected thing is actually targetable at all, which isn't a guarantee with a wildcard, but more importantly you *want* to ensure the receiver is the same local you're modifying

LlamaLad7 (changes requested):

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
Earthcomputer: Is it valid to assume that custom fluids are not in the `FluidTags.LAVA` tag? I guess not right. That would mean we could target only the `getWaterColor` call.

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
Earthcomputer: I think this is fine.

fabric-rendering-fluids-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/fluid/FluidRendererMixin.java
LlamaLad7: Wouldn't have thought so

LlamaLad7 (approved):

modmuss50 (approved):

PR #4843, opened 2025-09-01
Fix `LivingEntityFeatureRenderEvents.ALLOW_CAPE_RENDER` not preventing cape rendering

Spinoscythe:


haykam821:
The event documentation seems to indicate that the current behavior is the expected behavior:

https://github.com/FabricMC/fabric/blob/e6b6ce14476ba9d8fdf4dbbb7380d7826cca349b/fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/LivingEntityFeatureRenderEvents.java#L45-L48

PR #4847, opened 2025-09-03
Add `FabricRenderState`

8s2:
Adds the `FabricRenderState` interface which is implemented by `EntityRenderState`, `BlockEntityRenderState`, `ItemRenderState`, and `ItemRenderState$LayerRenderState` via interface injection.
Since the addition of `EntityRenderState`, in order to pass custom data from an entity to its renderer in a clean manner, users have had to create their own injected interfaces. This PR adds the `RenderStateDataKey` class in a very similar manner to the `ExtraModelKey` from #4565. The `RenderStateDataKey` acts as a unique identifier for data that is attached to a render state. `FabricRenderState` includes methods for setting, getting, and clearing data keys. As an optimization, the map that stores the extra data is initially null, so that a new map instance is not created for every single render state.
This PR also includes a testmod mixin that uses this system to render the block a pig is standing on at its location. It also includes a test to verify that all supported render states function properly with their injected methods.


modmuss50:
Overall this looks great, just need to make sure it builds ok.

For some reason the github actions dont seem to have been ran, or even offer me to run them 🤔

cputnam-a11y (approved):

modmuss50 (approved):

PepperCode1:
Is there a reason that `FabricRenderState` is not implemented on `MapRenderState` and its inner class `Decoration`?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricRenderState.java
```diff
@@ -0,0 +1,60 @@
...
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * Set extra render data to the render state.
+	 * @param key the key of the data
+	 * @param value the data
+	 * @param <T> the type of the data
+	 */
+	default <T> void setData(RenderStateDataKey<T> key, T value) {
```
PepperCode1: Should the value parameter be nullable?

8s2:
> Is there a reason that `FabricRenderState` is not implemented on `MapRenderState` and its inner class `Decoration`?

There's quite a few classes with the -RenderState suffix, I just prioritized the ones I expect to be used most often. I can definitely add this to the states you mentioned, though! 
Do you think it would also be better to rename `FabricRenderState` to `FabricRenderStateDataContainer` or something similar? This way, users won't be confused if there are render state classes now or in the future that don't implement this interface.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricRenderState.java
8s2: Yes, will fix

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricRenderState.java
```diff
@@ -0,0 +1,60 @@
...
+		throw new UnsupportedOperationException("Implemented via mixin");
+	}
+
+	/**
+	 * Clears all extra render data on the render state.
+	 */
+	default void clearData() {
```
Juuxel: This sounds a bit like a generic reset method for clearing any data of the state. I wonder if all of these methods should have "extra" or "custom" or other such words in their name, especially this one.

8s2: I'll change this method name to `clearExtraData`, but I think it's ok for `getData` and `setData` to stay the same since they have `RenderStateDataKey` in their parameters, and because they're likely to get called a lot and I don't want code to get verbose.

fabric-rendering-v1/src/client/resources/fabric-rendering-v1.mixins.json
```diff
@@ -9,6 +9,9 @@
     "BlockColorsMixin",
     "BlockEntityRendererFactoriesMixin",
     "CapeFeatureRendererMixin",
+    "class_12076Mixin",
+    "class_12077Mixin",
+    "class_12078Mixin",
```
8s2: Will fix obfuscated names once classes are mapped in yarn. Notably, these classes each seem to have methods which act like a clear method. If these methods are mapped as `clear`, then I can consolidate these mixins alongside the `EntityRenderStatesMixin`, `ItemRenderStateMixin`, and `ItemRenderStateLayerRenderStateMixin` into one mixin that targets all their clear methods.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricRenderState.java
```diff
@@ -0,0 +1,84 @@
...
+ * Fabric-provided extensions for render states, allowing for the addition of extra render data.
+ *
+ * <p>Note: This interface is automatically implemented on the following classes via Mixin and interface injection:
+ * <ul>
+ *     <li>{@link EntityRenderState},
+ *     <li>{@link BlockEntityRenderState}
+ *     <li>{@link ItemRenderState} and {@link ItemRenderState.LayerRenderState}
+ *     <li>{@link MapRenderState} and {@link MapRenderState.Decoration}
+ *     <li>{@link MovingBlockRenderState}
+ *     <li>{@link EntityRenderStates}
```
8s2: Implemented for `EntityRenderStates` as I've noticed the class is being used in the world renderer as some sort of central render state. I suggest this class be remapped in yarn to reflect that it covers more than just entities now.

PepperCode1 (approved):

modmuss50:
Please can you fix the tests and checkstyle build.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/renderstate/WorldBorderRenderStateMixin.java
```diff
@@ -0,0 +1,45 @@
...
+@Mixin({
+		ItemRenderState.class,
+		ItemRenderState.LayerRenderState.class,
+		WorldRenderState.class,
+		WeatherRenderState.class,
+		WorldBorderRenderState.class,
+		SkyRenderState.class
+})
+public class ClearableRenderStateMixin {
+	@Inject(method = "clear", at = @At("TAIL"))
```
modmuss50, line 30: This mixin doesnt seem to be working in production, as its targgeting many diffrent methods, that each have a different intermediary name. Ill fix this now by splitting it into many mixins.

PR #4857, opened 2025-09-16
Deprecate `EntityRendererRegistry` in favor of directly using `EntityRendererFactories#register`

Spinoscythe:


PepperCode1 (commented):
Should the existing API delegate to the TAWed method?

Spinoscythe:
> Should the existing API delegate to the TAWed method?

I mean we did it with `BlockEntityRendererRegistry`, since it seemed unnecessary for an API to exist all because the vanilla registry method had been privatised.

PepperCode1:
`BlockEntityRendererRegistry.register` does not delegate to the corresponding vanilla method. I meant whether `EntityRendererRegistry.register` should delegate to the corresponding vanilla method.

Spinoscythe:
> `BlockEntityRendererRegistry.register` does not delegate to the corresponding vanilla method. I meant whether `EntityRendererRegistry.register` should delegate to the corresponding vanilla method.

Oh, my fault. I guess we can leave it as it is, since the behaviour isn't all that different than directly putting it in the vanilla map.

Also, do you mind resolving the conflicts 😅😅

PepperCode1:
Please resolve the merge conflicts yourself.

modmuss50 (approved):

PR #4861, opened 2025-09-16
Fix Datagen API's FMJ and build.gradle dependencies

### LINKED ISSUE #4823: Data Generation API should depend on Resource Condtions API
Hi, [Data Generation API uses `ResourceCondition`](https://github.com/FabricMC/fabric/blob/1.21.9/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/loot/FabricBlockLootTableGenerator.java#L23) but [it does not depend on Resource Condtions API in the fabric.mod.json](https://github.com/FabricMC/fabric/blob/1.21.9/fabric-data-generation-api-v1/src/main/resources/fabric.mod.json#L18). I think this may be an oversight.

PepperCode1:
Fix #4823. `fabric-networking-api-v1` was declared as a module dependency but was not used by either the module or testmod, strangely enough.

### DIFF
--- a/fabric-data-generation-api-v1/build.gradle
+++ b/fabric-data-generation-api-v1/build.gradle
@@ -1,17 +1,15 @@
 version = getSubprojectVersion(project)
 
 moduleDependencies(project, [
-	'fabric-api-base',
 	'fabric-convention-tags-v2',
+	'fabric-recipe-api-v1',
 	'fabric-registry-sync-v0',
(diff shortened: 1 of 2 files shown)


modmuss50:
Looks good, thanks. When we update to the next loom I plan to use the new FMJ generation task to generate all of these to ensure they match what is defined in gradle.

modmuss50 (approved):

PR #4870, opened 2025-09-19
add a conventional salt tag

DaPlumer:
I would like to add a salt tag for usages in cooking mods (Like the one I'm currently developing) and possibly other technical based mods.


Daomephsta:
Files in `src/generated` should not be edited directly. They are generated during datagen runs by the generators in `src/datagen`, so you need to edit the generator instead or your changes will be lost after the next datagen run.  
See [`ItemTagGenerator`](https://github.com/FabricMC/fabric/blob/1.21.9/fabric-convention-tags-v2/src/datagen/java/net/fabricmc/fabric/impl/tag/convention/datagen/generators/ItemTagGenerator.java)

modmuss50:
This doesnt need adding to fabric API, you can just use this tag already in your mod. It only needs adding to fabric if there are vanilla items that should be there.

PR #4872, opened 2025-09-21
1.21.9 porting fixes

apple502j:
- **Breaking**: `ServerEntityCombatEvents` now passes the damage source for consistency with vanilla (change happened in 25w31a)
- Transfer API now supports shelves.
- Javadoc fixes.
- Terminally deprecates `NbtTypes#NUMBER` because it is useless since 1.21.6; the whole class is duplicative of vanilla one, but not sure what's the best approach.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/NbtType.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/NbtType.java
@@ -16,7 +16,6 @@
 
 package net.fabricmc.fabric.api.util;
 
-import net.minecraft.nbt.NbtCompound;
 import net.minecraft.nbt.NbtElement;
 
 /**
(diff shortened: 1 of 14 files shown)


Patbox:
Honestly assuming no one uses NbtTypes, it could be just removed

fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/NbtType.java
```diff
@@ -46,9 +44,10 @@ public final class NbtType {
 
 	/**
 	 * Any numeric value: byte, short, int, long, float, double.
-	 *
-	 * @see NbtCompound#contains(String, int)
+	 * @deprecated This constant is no longer accepted by any vanilla method,
+	 * and therefore is completely useless.
 	 */
+	@Deprecated(forRemoval = true)
```
modmuss50: Just remove this one, we can maybe remove this whole class in 1.22?

PR #4875, opened 2025-09-21
Remove `WorldRenderEvents` and `DimensionRenderingRegistry`

PepperCode1:
The underlying vanilla code has changed significantly, and now has separate extract/render stages for world rendering. The existing API structure for these classes cannot be adapted to these vanilla changes. `WorldRenderEvents` also has other long-standing issues.

All/most of the API functionality removed here will be re-added in multiple PRs soon once the API is redesigned to better fit the vanilla code. It is best to remove the classes now so that the current (unsuitable) APIs are not considered stable and immutable and require deprecation once Fabric API for 1.21.9 RC 1 is released.


kevinthegreat1:
While I agree that these events, especially world render events, need to be completely redone, as I’ve said on discord. I would prefer they are not removed without replacement. I have almost a hundred uses of world render events, and have the mod already working on the pre-releases. It would be annoying to refactor all of this.

modmuss50 (approved):
As discussed, we are going to remove these as they are broken and are not soemthing we want to keep for the lifetime of 1.21.9. In the medium/long term we will have a better API that wont require keeping these APIs around for the sake of backwards compability.

The goal is to provide a stable FAPI release alongside Minecraft that we can maintain going forward, thats not something we can easily achieve with what is here today.

PR #4878, opened 2025-09-23
Fabric Recipe Sync - Prototype

### LINKED ISSUE #4667: [Feature Request] Recipe Sync API
For mods that use recipes on the client, it would be useful to have a way for mods to sync recipes from the server to the client.

Considerations
* Mods should exist on the server in order to make the request there. 
    * This will avoid clients requesting massive amounts of data from a server in an uncontrolled way.
* There should be a period of time or event where mods can request specific recipe types to by synced. 
    * Not all recipes need to by synced and multiple mods may require the same recipe types, so this will prevent sending too much data or redundant data.
* Recipes sent to the client should include the ID (so ideally the whole `RecipeHolder` should by synced and stored in a `RecipeMap`). 
    * This will give mods better ability to manage recipes on the client similarly to how they are managed on the server, allowing for persisting references to recipes across game play sessions.

@shartte completed an implementation in https://github.com/neoforged/NeoForge/pull/2021 with some additional motivations and design rationale.

Spinoscythe:
Hopefully, Closes #4667


fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/FabricRecipeSyncPayloadHandler.java
```diff
@@ -0,0 +1,21 @@
...
+import net.fabricmc.fabric.api.client.networking.v1.ClientPlayNetworking;
+import net.fabricmc.fabric.api.networking.v1.PayloadTypeRegistry;
+import net.fabricmc.fabric.api.recipe.v1.PopulateSyncableRecipesCallback;
+
+public final class FabricRecipeSyncPayloadHandler {
+	private FabricRecipeSyncPayloadHandler() {
+	}
+
+	public static void registerPayloadsAndReceivers() {
+		PayloadTypeRegistry.playS2C().register(FabricSyncRecipesPayload.ID, FabricSyncRecipesPayload.PACKET_CODEC);
```
modmuss50: This needs to be done on the client and server, the packet wont work as-is

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/PopulateSyncableRecipesCallback.java
```diff
@@ -0,0 +1,17 @@
+package net.fabricmc.fabric.api.recipe.v1;
+
+import net.fabricmc.fabric.api.event.Event;
+
+import net.fabricmc.fabric.api.event.EventFactory;
+
+import net.minecraft.recipe.PreparedRecipes;
+
+public interface PopulateSyncableRecipesCallback {
+	Event<PopulateSyncableRecipesCallback> EVENT = EventFactory.createArrayBacked(PopulateSyncableRecipesCallback.class, callbacks -> recipes -> {
```
modmuss50: I think we need to store the recipes in memory so a mod can query them whenever, an even to know when they have changed is a good idea tho.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/SendSyncableRecipesCallback.java
```diff
@@ -0,0 +1,20 @@
...
+import it.unimi.dsi.fastutil.objects.ReferenceSet;
+
+import net.fabricmc.fabric.api.event.Event;
+
+import net.fabricmc.fabric.api.event.EventFactory;
+
+import net.minecraft.recipe.RecipeType;
+
+public interface SendSyncableRecipesCallback {
+	Event<SendSyncableRecipesCallback> EVENT = EventFactory.createArrayBacked(SendSyncableRecipesCallback.class, callbacks -> recipeTypesToSend -> {
```
modmuss50: We need to cover the following use cases I think (please chime in if you have other ideas)

- A mod may wish to sync all recipes
- A mod may wish to sync specific recipes, at specific times. (E.g the player unlocks a recipe)
- A mod way wish to revoke a recipe (e.g the player locks the recipe)
- A mod may wish to send recipes only to certain clients (e.g if the mod is on the client)
- Mulitple mods may wish to sync the same recipes, this should not send duplicate data to the client.
- Reloading the recipes will need to ask these questions again

Desinging an API to cover all of these use cases might be a little challanging, may be best to focus on the main use cases first but keep the others in mind.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/PlayerManagerMixin.java
```diff
@@ -0,0 +1,38 @@
...
+	@Shadow
+	private MinecraftServer server;
+
+	@WrapOperation(method = "onPlayerConnect", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/network/ServerPlayNetworkHandler;sendPacket(Lnet/minecraft/network/packet/Packet;)V", ordinal = 3))
+	private void populateRecipes(ServerPlayNetworkHandler instance, Packet packet, Operation<Void> original, @Local(argsOnly = true) ServerPlayerEntity serverPlayer) {
+		final ReferenceSet<RecipeType<?>> recipeTypesToSend = new ReferenceOpenHashSet<>();
+		SendSyncableRecipesCallback.EVENT.invoker().sendRecipes(recipeTypesToSend);
+		original.call(instance, packet);
+		var payload = FabricSyncRecipesPayload.create(recipeTypesToSend, server.getRecipeManager().preparedRecipes);
+		ServerPlayNetworking.send(serverPlayer, payload);
```
modmuss50: This is going to need packet splitting, likely best to wait for: https://github.com/FabricMC/fabric/pull/4841

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/SendSyncableRecipesCallback.java
Patbox: In my opinion at least, minimal api (so initial implementation) should only really allow marking recipes serializable by their RecipeSerializer which ideally would only be done by mod adding it (or things like emi/rei/jei in case of vanilla ones). 

This avoids duplication or syncing recipes that can't be synced, which can happen nowadays (recipe might not provide working packet codec, as vanilla doesn't use them anymore or just require server only data / registries).

In case of checking if recipe serializer is supported on client, using similar system to what ingredients do would probably work best.

PR #4880, opened 2025-09-23
Add `AfterBackground` event

Fuzss:
This adds a new screen render event that runs between `BeforeRender` and `AfterRender`, directly after `Screen::renderBackground` has been called. 

This is useful for drawing additional background contents before text, items, the carried item, tooltips, etc. are drawn.

Here is an example from the included test mod using all three screen render events to demonstrate the usefulness of each one individually:

1. `BeforeRender`: The icon is drawn BEHIND both the handled screen interface and the darkened background, text, items, the carried item, tooltips, etc.
2. `AfterBackground`: The icon is drawn ABOVE both the handled screen interface and the darkened background, but still BEHIND text, items, the carried item, tooltips, etc.
3. `AfterRender`: The icon is drawn ABOVE everything, including the background, handled screen interface, text, items, the carried item, tooltips, etc.

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenEvents.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenEvents.java
@@ -127,6 +127,17 @@ public final class ScreenEvents {
 	}
 
 	/**
+	 * An event that is called after a screen's background is rendered.
+	 *
+	 * @return the event
+	 */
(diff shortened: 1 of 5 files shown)


modmuss50 (approved):

apple502j:
Unrelated, but I think the param name should be using tickProgress now...

modmuss50:
> Unrelated, but I think the param name should be using tickProgress now...

Yes, I will merge this now as it matches the exisitng an events. These should be renamed as a seperate PR.

PR #4884, opened 2025-09-25
Provide a way to get existing spawn entries in `SpawnSettingsContext`

Fuzss:
`SpawnSettingsContext` currently has no way to get the already registered spawn entries, particularly vanilla entries.

Personally I find this very useful information to have, as I use it in a bunch of my mods to add spawns of variant mobs depending on the presence of vanilla mobs.

Example: In the Mutant Monsters mod I add mutant zombie spawning only to biomes which in vanilla are allowed to spawn zombies at a weight relative to the vanilla counterpart.

So this PR adds a simple method for providing a view already registered spawn entries for a given spawn group.

An alternative approach would be to make a view of the entire `fabricSpawners` map accessible, although I don't think that is really necessary.

### DIFF
--- a/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModificationContext.java
+++ b/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/BiomeModificationContext.java
@@ -16,6 +16,7 @@
 
 package net.fabricmc.fabric.api.biome.v1;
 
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):

PepperCode1 (changes requested):
Both new methods should be annotated with `UnmodifiableView`.

PepperCode1 (approved):

PR #4885, opened 2025-09-25
Update ParticleRendererRegistry test to use a fully custom Particle

Grauly:
Originally this PR included a mixin fix which was cherry-picked below.

It changes the ParticleRendererRegistryTest to use a non `BillboardParticle` Particle, to test fully custom particles.

This PR is not entirely ready to merge, there are a few issues left to be solved:
- The access of the deprecated `SpriteAtlasTexture.BLOCK_ATLAS_TEXTURE`
- The entire RendererRegistryTest could probably be done more clearly, but I cannot figure out how. Any piece of data more that would need to be passed through `TestParticlePositions` would require alot more buffer work in there.


modmuss50:
The mixin fix was cherry-picked in: https://github.com/FabricMC/fabric/commit/ed8c18266beafe368d9137e16e769d6dd7a96da5 as its quite clearly wrong, and the fix is needed as a priority. 

The test changes need more work/input from someone who can help. They can be merged at a later date.

PR #4893, opened 2025-09-30
Add refreshCommandCompletions method to fcapi

xpple:
On servers, when a change happens to a condition for a command (most commonly permission changes), the server will call `CommandManager.sendCommandTree` to inform the client. However, no such method exists for when clients want to do the same for client commands. In my case, I wanted to hide a command behind a feature flag that can be toggled by the user. Now when the flag is toggled, there is no way to inform the user, and the command will seem to be unavailable until the user re-joins the world or server. This PR adds a method to the Fabric Command API that will mimic the server calling `CommandManager.sendCommandTree` on the client by simply calling the handler for the `minecraft:commands` packet. To call this method, the last sent `minecraft:commands` packet is cached.

### DIFF
--- a/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
+++ b/fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
@@ -22,6 +22,10 @@ import com.mojang.brigadier.builder.LiteralArgumentBuilder;
 import com.mojang.brigadier.builder.RequiredArgumentBuilder;
 import org.jetbrains.annotations.Nullable;
 
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.network.ClientPlayNetworkHandler;
+import net.minecraft.network.packet.s2c.play.CommandTreeS2CPacket;
+
 import net.fabricmc.fabric.impl.command.client.ClientCommandInternals;
 
 /**
@@ -73,6 +77,29 @@ public final class ClientCommandManager {
 	 */
 	public static @Nullable CommandDispatcher<FabricClientCommandSource> getActiveDispatcher() {
 		return ClientCommandInternals.getActiveDispatcher();
+	}
+
+	/**
+	 * Refresh the command completions. This is helpful when a condition as defined using {@link LiteralArgumentBuilder#requires}
+	 * changes for a client command. The method uses the last received {@code minecraft:commands}
+	 * packet and calls its handler. This triggers the client command's condition to be reevaluated.
+	 *
+	 * <p>Will not do anything when not connected to a server (dedicated or integrated).</p>
+	 */
+	public static void refreshCommandCompletions() {
+		ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+
+		if (networkHandler == null) {
+			return;
+		}
+
+		CommandTreeS2CPacket lastReceivedCommandsPacket = ClientCommandInternals.getLastReceivedCommandsPacket();
+
+		if (lastReceivedCommandsPacket == null) {
+			return;
+		}
+
+		networkHandler.onCommandTree(lastReceivedCommandsPacket);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


Earthcomputer:
You should add something to the test mod so that it can be tested.

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
```diff
@@ -75,6 +79,29 @@ private ClientCommandManager() {
 		return ClientCommandInternals.getActiveDispatcher();
 	}
 
+	/**
+	 * Refresh the command completions. This is helpful when a condition as defined using {@link LiteralArgumentBuilder#requires}
+	 * changes for a client command. The method uses the last received {@code minecraft:commands}
+	 * packet and calls its handler. This triggers the client command's condition to be reevaluated.
+	 *
+	 * <p>Will not do anything when not connected to a server (dedicated or integrated).</p>
```
modmuss50: Not super keen on just silently failing with no way to know if it worked or not. I would throw an a mod can choose to catch it or not.

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/impl/command/client/ClientCommandInternals.java
```diff
@@ -53,6 +54,7 @@ public final class ClientCommandInternals {
 	private static final String API_COMMAND_NAME = "fabric-command-api-v2:client";
 	private static final String SHORT_API_COMMAND_NAME = "fcc";
 	private static @Nullable CommandDispatcher<FabricClientCommandSource> activeDispatcher;
+	private static @Nullable CommandTreeS2CPacket lastReceivedCommandsPacket = null;
```
modmuss50: Not keen on storing this staticlaly, as it may outlive the current connection. I think it would be best stored none statically in ClientPlayNetworkHandler.

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/mixin/command/client/ClientPlayNetworkHandlerMixin.java
```diff
@@ -72,6 +72,11 @@ private void onOnCommandTree(CommandTreeS2CPacket packet, CallbackInfo info) {
 		ClientCommandInternals.addCommands((CommandDispatcher) commandDispatcher, (FabricClientCommandSource) commandSource);
 	}
 
+	@Inject(method = "onCommandTree", at = @At("HEAD"))
+	private void setLastReceivedCommandsPacket(CommandTreeS2CPacket packet, CallbackInfo ci) {
+		ClientCommandInternals.setLastReceivedCommandsPacket(packet);
```
modmuss50: Is this thread safe? Can this be called on the network thread at the same time someone calls refreshCommandCompletions on some other thread?

fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommandManager.java
xpple: What exception do you suggest?

modmuss50: IllegalStateException

fabric-command-api-v2/src/testmodClient/java/net/fabricmc/fabric/test/command/client/ClientCommandTest.java
```diff
@@ -91,6 +93,14 @@ public void onInitializeClient() {
 					})
 			));
 
+			// Command with condition that changes
+			dispatcher.register(ClientCommandManager.literal("test_client_command_with_condition_change").requires(source -> commandFlag).executes(context -> Command.SINGLE_SUCCESS));
```
Earthcomputer: Maybe this command should print a message to chat, to prove that it's actually executing.

xpple: The proof is the fact the command is suggested as completion, _that_ is what we're testing the implementation for. Note that the execution of the command works regardless of whether the command is suggested, as long as the condition evaluates to true.

Earthcomputer: No, the completion is not necessarily the same as execution, I think they even use different dispatchers.

xpple: I added some comments and feedback that should help future people understand the test.

modmuss50 (approved):

PR #4906, opened 2025-10-07
New World Render Events

kevinthegreat1:
This is a functioning concept. Some events may be removed.

World Render Events After Translucent Test
World Render Events Block Outline Test


modmuss50:
This looks like a good start, as discussed on discord we should use context objects (can likely be 1 object with mutliple interfaces shared by all the events). Docs should also be added. Do let me know if there is anything I can do to help. The main bit I was struggling with was knowing what events are needed and what needs to be passed to them, thanks for figuring that out.

PepperCode1:
I would recommend against implementing all the interfaces on a single object as some of the context received by certain events is completely disjoint and lumping it together can lead to accidentally using it in the wrong place.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/WorldRenderEventsTests.java
```diff
@@ -0,0 +1,79 @@
...
+
+			MinecraftClient.getInstance().getBlockRenderManager().renderBlockAsEntity(
+					Blocks.DIAMOND_BLOCK.getDefaultState(),
+					matrixStack, consumers, 15728880, OverlayTexture.DEFAULT_UV
+			);
+
+			matrixStack.pop();
+		}
+
+		return true;
```
skycatminepokie, line 72: Test can't fail - is this intentional? I don't see a screenshot taken or something.

skycatminepokie (approved):

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/WorldRenderEventsTests.java
kevinthegreat1: I'm not sure what you mean. This isn't a game test. I might add one in the future.

kevinthegreat1:
I have added context objects. Tests:
World Render Events Test Render and Block Outline

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldTerrainRenderContext.java
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Returns the {@code WorldTerrainRenderContext} for the given {@code WorldRenderer} instance, for use in cases where you
+	 * have access to the world renderer but not the world render context. World render events always pass the world
+	 * render context as a parameter, so always prefer to use that over this method.
+	 *
+	 * @param worldRenderer The world renderer
+	 * @return The world render context for the world renderer
+	 * @throws IllegalStateException If not currently rendering the world
+	 */
+	static WorldTerrainRenderContext getInstance(WorldRenderer worldRenderer) {
```
modmuss50: I dont really understand the use case of being able to get these statically? Surely just passing them as an even param is enough?

kevinthegreat1: These were there from before. I don't really like these either so I will remove.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/WorldRendererMixin.java
```diff
@@ -16,18 +16,164 @@
...
+
+	@Inject(method = "renderTargetBlockOutline", at = @At(value = "FIELD", target = "Lnet/minecraft/client/render/state/CameraRenderState;pos:Lnet/minecraft/util/math/Vec3d;"), cancellable = true)
+	private void onDrawBlockOutline(VertexConsumerProvider.Immediate consumers, MatrixStack matrices, boolean bl, WorldRenderState worldRenderState, CallbackInfo ci) {
+		if (!WorldRenderEvents.BEFORE_BLOCK_OUTLINE_RENDER.invoker().beforeBlockOutlineRender(renderContext)) {
+			consumers.drawCurrentLayer();
+			ci.cancel();
+		}
+	}
+
+	@Inject(method = "method_62214", at = @At("RETURN"))
```
Earthcomputer: You probably want this just before the last `draw`, not after
```suggestion
	@Inject(method = "method_62214", at = @At(value = "INVOKE:LAST", target = "Lnet/minecraft/client/render/VertexConsumerProvider$Immediate;draw()V"))
```

jeseibel:
It looks like this could potentially solve DH's problems, although I also noticed several of the events names where changed which feels a bit odd. If the events behave different then their older counterparts, then the name changes would make sense, although if they still function in a similar fashion to the old events I'd recommend keeping the names the same for better backwards compatibility.

I'm specifically looking for the following events that Distant Horizons needs:
```java
WorldRenderEvents.AFTER_SETUP
WorldRenderEvents.AFTER_ENTITIES
WorldRenderEvents.AFTER_TRANSLUCENT
```

I was able to build the pull request locally but haven't had the time to test it in DH's dev environment yet. Hopefully I can take a closer look over the next couple days.

kevinthegreat1:
@jeseibel 

I have refactored some of the events, and it seems like `AFTER_ENTITIES` and `AFTER_TRANSLUCENT` will remain as is, while `AFTER_SETUP` will be replaced by `START_RENDER`, and additionally giving the section render state as context. Please leave a comment if you have use cases not covered by these events.

kevinthegreat1:
I have added some documentation.

Remaining questions:
1. Frustum seems to be used during rendering. Should it be included in the render context?
2. Is the current before and after entity events injection points good? The before event injects before line 608, and the after event injects after line 614.  
I believe line 603 is recording entity render commands, line 607 is drawing those commands into the buffer builder, and line 608-619 is building and drawing the buffer builders.
Entity Rendering in World Renderer

4. In the before debug and last events, the docs mention: "The OpenGL render state view matrix will be transformed to match the camera view before the event is called." I don't think this is true?
5. In the after translucent event, the docs mention: "Before translucency combine has happened in fabulous mode." They are using post-effect processors now, and should this be included?

jeseibel:
> @jeseibel
> 
> I have refactored some of the events, and it seems like `AFTER_ENTITIES` and `AFTER_TRANSLUCENT` will remain as is, while `AFTER_SETUP` will be replaced by `START_RENDER`, and additionally giving the section render state as context. Please leave a comment if you have use cases not covered by these events.

@kevinthegreat1 
Thanks for the ping.

Those events look good, although the objects they return are missing some parameters I need.
For rendering I need:
- **Matrix4f** projection matrix
- **Matrix4f** Model View Matrix (Mojang seems to call this the "position matrix")
- **ClientLevel**
- **float** game Time Delta Ticks

Which, looking at the events as is, appear to only be present in the `WorldExtractionContext` event. 
(As a side note: what is the "extraction" event? Having read its javadocs I'm uncertain what it means.)

apple502j:
@jeseibel IIRC the extraction event is used to extract the data needed for batched rendering later.

cputnam-a11y:
> > @jeseibel
> > I have refactored some of the events, and it seems like `AFTER_ENTITIES` and `AFTER_TRANSLUCENT` will remain as is, while `AFTER_SETUP` will be replaced by `START_RENDER`, and additionally giving the section render state as context. Please leave a comment if you have use cases not covered by these events.
> 
> @kevinthegreat1 Thanks for the ping.
> 
> Those events look good, although the objects they return are missing some parameters I need. For rendering I need:
> 
> * **Matrix4f** projection matrix
> * **Matrix4f** Model View Matrix (Mojang seems to call this the "position matrix")
> * **ClientLevel**
> * **float** game Time Delta Ticks
> 
> Which, looking at the events as is, appear to only be present in the `WorldExtractionContext` event. (As a side note: what is the "extraction" event? Having read its javadocs I'm uncertain what it means.)

Rendering is now separated into two stages to prepare for multi threading. The world and tick progress shouldn't be available in the render events. If you need them use them in the extraction callback

Earthcomputer:
@jeseibel the lack of these parameters is intentional, this is the way Mojang has refactored the rendering code. You need to use these parameters to store things to be rendered into the `WorldRenderState` (which has a Fabric interface injection which allows you to store custom stuff). Then, in the rendering events you can access them from there.

Do NOT store the `ClientWorld` in the state, as Mojang is planning on moving rendering off-thread and accessing the world off-thread isn't thread safe. The intention is to extract all the context you need to render during extraction BEFORE rendering, and then pass that context to the render events via the `WorldRenderState`. This would include using the matrices to transform points during extraction.

LambdAurora:
Should the events include a system to extract custom data in the extraction event and fetch them back in the rendering events?

Could take a similar form as the `ResourceReloader.Store` shared state object.

As-is the extraction of states might be a challenge for mods needing some additional custom states.

Earthcomputer:
> Should the events include a system to extract custom data in the extraction event and fetch them back in the rendering events?

They already do, as explained in my comment. In fact this is what is intended to be used.

cputnam-a11y:
All render states already support storing modded data

LambdAurora:
> They already do, as explained in my comment. In fact this is what is intended to be used.

This wasn't clear from the PR at first, thanks for the clarification (you were too fast for my own comment ahah) :smile:

jeseibel:
Forgive my ignorance but is there a reason mods should handle: extracting the render state, storing it in a static variable, then using it during the render events; vs fabric handling that all behind the scenes and just passing that data back for each of the render events?
It seems like an unnecessary step for API users but maybe I'm missing something.

cputnam-a11y:
See above comments on multi threading. The events will soon fire on different threads. As for static fields, don't do that. Use the WorldRenderState to store your data
Tldr: Mojang is forcing the change, and if we don't adapt now 100s of mods will break in very near updates

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/AbstractWorldRenderContext.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.client.rendering.v1.world;
+
+import net.minecraft.client.render.GameRenderer;
+import net.minecraft.client.render.WorldRenderer;
+import net.minecraft.client.render.state.WorldRenderState;
+
+public interface AbstractWorldRenderContext {
```
apple502j, line 26: `@NotExtendable`

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldExtractionContext.java
```diff
@@ -0,0 +1,48 @@
...
+import org.joml.Matrix4f;
+
+import net.minecraft.client.render.Camera;
+import net.minecraft.client.render.Frustum;
+import net.minecraft.client.render.RenderTickCounter;
+import net.minecraft.client.world.ClientWorld;
+
+public interface WorldExtractionContext extends AbstractWorldRenderContext {
+	/**
+	 * Convenient access to {WorldRenderer.world}.
```
apple502j:
```suggestion
	 * Convenient access to {@link WorldRenderer#world}.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderContext.java
```diff
@@ -0,0 +1,58 @@
...
+package net.fabricmc.fabric.api.client.rendering.v1.world;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import net.minecraft.client.render.VertexConsumerProvider;
+import net.minecraft.client.render.command.OrderedRenderCommandQueue;
+import net.minecraft.client.util.math.MatrixStack;
+
+public interface WorldRenderContext extends WorldTerrainRenderContext {
+	@ApiStatus.Experimental
```
apple502j: What is the meaning of this experimental?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderContext.java
```diff
@@ -0,0 +1,58 @@
...
+	OrderedRenderCommandQueue commandQueue();
+
+	MatrixStack matrixStack();
```
apple502j:
```suggestion
	MatrixStack matrices();
```

given the yarn convention

apple502j (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/AbstractWorldRenderContext.java
Earthcomputer: On other interfaces too ^^

jeseibel:
> See above comments on multi threading. The events will soon fire on different threads. As for static fields, don't do that. Use the WorldRenderState to store your data Tldr: Mojang is forcing the change, and if we don't adapt now 100s of mods will break in very near updates

@cputnam-a11y 
Your mentions of multithreading are interesting, OpenGL cannot run on multiple threads due to it's global state machine (unless something's changed since GL 4.6 that I missed, I'd love to multi-thread my renderer if it was possible).

If that these events are being fired on threads other than the render thread, then I'd argue they aren't rendering events, they're rendering _setup_ events.

I need rendering events that fire on the Render thread so I can run OpenGL calls.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/AbstractWorldRenderContext.java
kevinthegreat1: Can I seal them?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderContext.java
kevinthegreat1: I did it because this is a new perimeter and I’m not quite sure what it’s used for.

cputnam-a11y:
> > See above comments on multi threading. The events will soon fire on different threads. As for static fields, don't do that. Use the WorldRenderState to store your data Tldr: Mojang is forcing the change, and if we don't adapt now 100s of mods will break in very near updates
> 
> @cputnam-a11y Your mentions of multithreading are interesting, OpenGL cannot run on multiple threads due to it's global state machine (unless something's changed since GL 4.6 that I missed, I'd love to multi-thread my renderer if it was possible).
> 
> If that these events are being fired on threads other than the render thread, then I'd argue they aren't rendering events, they're rendering _setup_ events.
> 
> I need rendering events that fire on the Render thread so I can run OpenGL calls.

That is correct. The game state will be accessible from one thread, and opengl from another. The render callbacks will not have access to the game aside from the extracted state from the extract callback + vanilla extraction. I'm probably not going to continue explaining/ arguing here, as all the implementors understand the paradigm (tmk), but I'm more than willing to continue explaining on the FabricMC discord if you wish. The net net is that the objects you've requested would either become dangerous deprecated API that is a foot gun for everyone, or would have to be removed in yet another breaking change later on down the road.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderContext.java
kevinthegreat1: Do you think `WorldRenderState` should be `state` or `worldState`?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/AbstractWorldRenderContext.java
Earthcomputer: No, you cannot have permitted subclasses from a different package

jeseibel:
> > > See above comments on multi threading. The events will soon fire on different threads. As for static fields, don't do that. Use the WorldRenderState to store your data Tldr: Mojang is forcing the change, and if we don't adapt now 100s of mods will break in very near updates
> > 
> > 
> > @cputnam-a11y Your mentions of multithreading are interesting, OpenGL cannot run on multiple threads due to it's global state machine (unless something's changed since GL 4.6 that I missed, I'd love to multi-thread my renderer if it was possible).
> > If that these events are being fired on threads other than the render thread, then I'd argue they aren't rendering events, they're rendering _setup_ events.
> > I need rendering events that fire on the Render thread so I can run OpenGL calls.
> 
> That is correct. The game state will be accessible from one thread, and opengl from another. The render callbacks will not have access to the game aside from the extracted state from the extract callback + vanilla extraction. I'm probably not going to continue explaining/ arguing here, as all the implementors understand the paradigm (tmk), but I'm more than willing to continue explaining on the FabricMC discord if you wish. The net net is that the objects you've requested would either become dangerous deprecated API that is a foot gun for everyone, or would have to be removed in yet another breaking change later on down the road.

Apologies if it sounds like I'm arguing that wasn't my intention. I was just trying to express how I use and view the API. If the API no longer supports calling OpenGL calls that's up to the Fabric team I'll just have to use Mixins instead.

cputnam-a11y:
You are welcome to do so until Mojang makes it impossible to do so. I wasn't sure if we were arguing or explaining so I listed both. As stated above feel free to join our discord for a more in depth explanation on how the mc rendering pipeline will work, but Mojang has expressed their intent to split the opengl part of rendering from the extraction and multi thread it. Anything else is an anti pattern waiting to break. using opengl is fine and supported, and using the world is fine and supported, but not in the same place. As you've said, there isn't really anything any of us can do to stop you. If you wish to use mixins to implement callbacks that receive the world you are welcome to do so, but those won't be in fapi because they would already be deprecated with the intent of the code.
Technically with the current events you could also do as Earthcomputer recommend against and share the world to the render callback 
I just realized that it seems you want us to take the ostrich algorithm approach maybe lol.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderContext.java
apple502j: If there's no conflicting state (perhaps one attached to GameRenderer), maybe.

PepperCode1:
Would it make sense to rename `START_RENDER` to `START_MAIN` and `LAST` to `END_MAIN`? All of these events run inside `method_62214`, but it's only one node in the frame graph, with vanilla rendering the sky in a separate node beforehand, for example.

kevinthegreat1:
@PepperCode1 I have renamed start render to start main. I kept last because it corresponds exactly to the old last. Please let me know if you want it changed. Also, I am wondering if you saw my comment here: https://github.com/FabricMC/fabric/pull/4906#issuecomment-3383143554.

Earthcomputer:
I would at least keep it as `FIRST` and `LAST` so it's consistent

modmuss50:
Whats left to get this in? Is there anything I can do to help?

kevinthegreat1:
I believe I'm waiting for an in-depth review of injection points. Basically, I've outlined what may need discussion [here](https://github.com/FabricMC/fabric/pull/4906#issuecomment-3383143554).

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+ * that make large-scale rendering changes to maintain compatibility by calling any broken event invokers directly.
+ *
+ * <p>These events can be separated into two categories, the "extraction" events and the "drawing" events,
+ * reflecting the respective vanilla phases. All data needed for rendering should be prepared in the "extraction" phase
+ * and drawn to the frame buffer during the "drawing" phase. All "extraction" events have the suffix "Extraction".
+ *
+ * <p>To attach modded data to vanilla render states, see {@link net.fabricmc.fabric.api.client.rendering.v1.FabricRenderState FabricRenderState}.
+ * Only attach the minimum data needed for rendering. Do not attach objects that are not thread-safe such as {@link net.minecraft.client.world.ClientWorld}.
+ *
+ * <p>These events are not dependent on the Fabric rendering API or Indigo but work when those are present.
```
PepperCode1: This line is completely unnecessary and should be removed.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+ */
+public final class WorldRenderEvents {
+	private WorldRenderEvents() { }
+
+	/**
+	 * Called after default block outline extraction and before rendering.
+	 * Can optionally cancel the default rendering by setting the outline render state to null
+	 * but all event handlers will always be called.
```
PepperCode1: Clarify that this only applies to handlers of specifically this event.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * Called after entity, terrain, and particle translucent layers have been
+	 * drawn to the framebuffer but before translucency combine has happened
+	 * in fabulous mode.
+	 *
+	 * <p>Use for drawing overlays or other effects on top of those targets
+	 * (or the main target when fabulous isn't active) before clouds and weather
+	 * are drawn.
+	 */
+	public static final Event<BeforeTranslucent> BEFORE_TRANSLUCENT = EventFactory.createArrayBacked(BeforeTranslucent.class, callbacks -> context -> {
```
PepperCode1, line 156: The documentation for this event is incorrect:
- It is called before translucent terrain, not after.
- I do not remember when particles are drawn. They may be drawn outside of the main pass, in which case this event is not called after particle translucent layers are drawn.
- It should not be used for drawing effects on top of fabulous targets. `END_MAIN` should be used for that case.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * Called before the last framebuffer write and before all world
+	 * rendering is torn down.
+	 *
+	 * <p>The OpenGL render state view matrix will be transformed to match the camera view
+	 * before the event is called.
+	 *
+	 * <p>Use to draw content that should appear on top of the world before hand and GUI rendering occur.
+	 */
+	public static final Event<EndMain> END_MAIN = EventFactory.createArrayBacked(EndMain.class, callbacks -> context -> {
```
PepperCode1, line 198: The first two paragraphs are incorrect. The documentation should say that this event runs at the end of the main render pass and give a summary of what is rendered before it and what is rendered after it.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+	 * RenderLayer instead of outputting to the framebuffer directly with GL calls.
+	 *
+	 * <p>The consumer is responsible for setup and tear down of GL state appropriate for the intended output.
+	 *
+	 * <p>Because solid and cutout quads are depth-tested, order of output does not matter except to improve
+	 * culling performance, which should not be significant after primary terrain rendering. This means
+	 * mods that currently hook calls to individual render layers can simply execute them all at once when
+	 * the event is called.
+	 *
+	 * <p>This event fires before entities and block entities are rendered and may be useful to prepare them.
```
PepperCode1: Clarify that this event is not only called before E/BEs are rendered, but also before their render submits are created. I also do not understand what is meant by "prepare them"; if this has anything to do with the E/BE render state, then this event should not be used to interact with it.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/world/WorldRenderEvents.java
```diff
@@ -0,0 +1,256 @@
...
+ * <p>These events can be separated into two categories, the "extraction" events and the "drawing" events,
+ * reflecting the respective vanilla phases. All data needed for rendering should be prepared in the "extraction" phase
+ * and drawn to the frame buffer during the "drawing" phase. All "extraction" events have the suffix "Extraction".
+ *
+ * <p>To attach modded data to vanilla render states, see {@link net.fabricmc.fabric.api.client.rendering.v1.FabricRenderState FabricRenderState}.
+ * Only attach the minimum data needed for rendering. Do not attach objects that are not thread-safe such as {@link net.minecraft.client.world.ClientWorld}.
+ *
+ * <p>These events are not dependent on the Fabric rendering API or Indigo but work when those are present.
+ */
+public final class WorldRenderEvents {
```
PepperCode1, line 41: It could be useful to say that all render events (not extraction events) support using raw graphics API calls (i.e. GL calls).

PepperCode1 (changes requested):
In addition to my comments, the context interfaces could use some additional method and class javadoc. I am good with everything other than documentation.

modmuss50 (approved):

PepperCode1 (approved):
I don't think clouds and weather are drawn between `BEFORE_TRANSLUCENT` and `END_MAIN` as the javadoc would have you believe, but I can't check right now, so this is good enough. If this PR has not been merged by the time I confirm this, I will make another comment.

ekulxam:
`renderMain` is called before `renderClouds` and `renderWeather` in `WorldRenderer`. I'm relatively sure that means that `BEFORE_TRANSLUCENT` and `END_MAIN` are actually drawn before both clouds and weather are drawn, but I don't know enough about rendering to be definitive.

kevinthegreat1:
I was misled by the old Javadoc, which was incorrect. I have fixed the `END_MAIN` Javadoc.

ekulxam:
Why does `WorldRenderContextImpl` implement all three context classes? `WorldRenderContext` already extends the other two, so do we need to explicitly mention them in `implements`?

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/WorldRenderEventsTests.java
```diff
@@ -0,0 +1,158 @@
...
+	/**
+	 * Renders a translucent filled box at (0, 100, 0).
+	 */
+	private static void renderBeforeTranslucent(WorldRenderContext context) {
+		Vec3d camera = context.worldState().cameraRenderState.pos;
+
+		context.matrices().push();
+		context.matrices().translate(-camera.x, -camera.y, -camera.z);
+
+		VertexRendering.drawFilledBox(context.matrices(), context.consumers().getBuffer(RenderLayer.getDebugFilledBox()), 0, 100, 0, 1, 101, 1, 0, 1, 0, 0.5f);
```
modmuss50, line 84: I have struggled to port to this 1.21.11 as VertexRendering.drawFilledBox was removed.

After great effort I replaced it with:

https://github.com/FabricMC/fabric/blob/0647ba064c8c970c9853404a62da397da893f572/fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/WorldRenderEventsTests.java#L86-L87

This works fine in a regular world, but is crashing in the test mods with:

```
Caused by: java.lang.IllegalStateException: Gizmos cannot be created here! No GizmoCollector has been registered.
	at knot//net.minecraft.class_12180.method_75553(class_12180.java:25)
	at knot//net.minecraft.class_12180.method_75542(class_12180.java:35)
	at knot//net.minecraft.class_12180.method_75541(class_12180.java:31)
	at knot//net.fabricmc.fabric.test.rendering.client.WorldRenderEventsTests.renderBeforeTranslucent(WorldRenderEventsTests.java:87)
```

Do you have any ideas or suggestions? I have no clue what im doing.

ekulxam: I believe you can set up the collector via `GizmoDrawing.using(GizmoCollector.EMPTY)`. This is what TestServer does.

PR #4911, opened 2025-10-07
Improve debug entry sorting.

### LINKED ISSUE #4910: [1.21.9] Sort debug entries by namespace first
Debug entries added by mods should be sorted by namespace first instead of path first, and appear below all vanilla debug entries.

modmuss50:
Closes https://github.com/FabricMC/fabric/issues/4910 

Sorts the entries by namespace and then path (`minecraft` ns being first)
Allows to search by namespace (other than `minecraft`)

Screenshot 2025-10-07 at 21 01 11


modmuss50:
~~Actually this is wrong, as it breaks the vanilla sorting.~~ Fixed :)

PR #4916, opened 2025-10-09
Armor Renderer Upgrades

8s2:
Vanilla's rendering overhaul with the `RenderCommandQueue` has made it nearly impossible to create a custom armor renderer that can render properly on any biped entity with the `ArmorRenderer` in its current implementation. In Vanilla, the `setAngles` method for models is now called when a `ModelCommand` is rendered, meaning that calling it before submitting a model to the render queue is not a reliable way to transform a model. 

For example, the current armor rendering test mod adds a renderer for a helmet and chestplate. The model has visibility of its parts set based on the equipment slot, but because this happens before the model is submitted, it leads to visibility conflicts when the helmet and chestplate are worn at the same time. Furthermore, the model's arms swing when it is worn on an armor stand (ignores the armor stand's poses too) and it does not scale properly on differently sized biped mobs.

I am creating this pull request to not only provide a solution for this issue, but to also provide some related upgrades that help with armor rendering. 

This PR includes:
- A factory-based option for registering an `ArmorRenderer`.
  - Current implementations of `ArmorRenderer` need a mutable and nullable model field that is checked on render and instantiated if `null`.
  - Renderers registered with a factory are reinstantiated on resource pack reload, which keeps their models up to date.
- The `TransformCopyingModel`
  - A workaround for transforms being applied on model command render, this model type has a source (`Model<? super S>`) and delegate (`Model<? super D>`) model field, with its render state holding both the source and delegate states at the same time (`Pair<S, D>`). When the `setAngles` method for this model is called, it sets the source model's angles using the first pair value, copies the transforms to the delegate model, then optionally sets the angles for the delegate model. The delegate model is what gets rendered.
  - Two static methods for rendering a `TransformCopyingModel` have been added to `ArmorRenderer`, based on the `renderPart` static method that used to be there.
- EntityModelLayerRegistry register method for `EquipmentModelData<EntityModelLayer>`
  - Vanilla now handles armor models using the `EquipmentModelData` class, this allows users to do the same.
- `FabricModel` injected interface
  - Adds the `getChildPart` and `copyTransforms` methods, which were needed to make the `TransformCopyingModel` to function. I find them useful enough to users for this to be a injected interface rather than just a mixin duck.
  - Models now cache a map of their model parts with a String key for their names.


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricModel.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+import net.minecraft.client.model.Model;
+import net.minecraft.client.model.ModelPart;
+
+/**
+ * General purpose Fabric extensions to the {@link Model} class.
+ *
+ * <p>Note: This interface is automatically implemented on all render pipelines via Mixin and interface injection.
```
Juuxel:
```suggestion
 * <p>Note: This interface is automatically implemented on all {@link Model} instances via Mixin and interface injection.
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/ArmorRendererRegistryImpl.java
```diff
@@ -40,14 +42,30 @@ public static void register(ArmorRenderer renderer, ItemConvertible... items) {
...
 	public static ArmorRenderer get(Item item) {
 		return RENDERERS.get(item);
 	}
+
+	public static void createArmorRenderers(EntityRendererFactory.Context context) {
+		RENDERERS.clear();
+		FACTORIES.forEach((item, factory) -> {
+			try {
+				RENDERERS.put(item, factory.createArmorRenderer(context));
+			} catch (Exception exception) {
```
Juuxel: Is this try-catch block really needed? We don't usually catch exceptions in factories like this.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/ModelMixin.java
```diff
@@ -0,0 +1,80 @@
...
+
+import net.minecraft.client.model.Model;
+import net.minecraft.client.model.ModelPart;
+import net.minecraft.client.render.RenderLayer;
+import net.minecraft.util.Identifier;
+
+import net.fabricmc.fabric.api.client.rendering.v1.FabricModel;
+
+@Mixin(Model.class)
+public abstract class ModelMixin<S> implements FabricModel<S> {
```
Juuxel:
```suggestion
abstract class ModelMixin<S> implements FabricModel<S> {
```
Nit: mixins should generally be package-private

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/EntityRenderManagerMixin.java
```diff
@@ -0,0 +1,37 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.client.render.entity.EntityRenderManager;
+import net.minecraft.client.render.entity.EntityRendererFactory;
+import net.minecraft.resource.ResourceManager;
+
+import net.fabricmc.fabric.impl.client.rendering.ArmorRendererRegistryImpl;
+
+@Mixin(EntityRenderManager.class)
+public class EntityRenderManagerMixin {
```
Juuxel:
```suggestion
class EntityRenderManagerMixin {
```

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/FabricModel.java
8s2: Whoops!

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/impl/client/rendering/ArmorRendererRegistryImpl.java
8s2: I based this off of how vanilla reloads entity renderer factories with a try-catch. Is it still best to just leave it out?

cputnam-a11y: Generally if a modded thing throws an exception it means that they want the game to stop, or they've done something invalid to cause an exception and the game should stop

8s2: Good point

Sollace:
I wouldn't say it's _entirely_ impossible. For example, I'm making use of a custom render command for my in-progress porting of Mine little Pony to 1.21.10

https://github.com/MineLittlePony/MineLittlePony/blob/1.21.10/src/main/java/com/minelittlepony/client/render/command/CustomModelRenderCommand.java

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/TransformCopyingModel.java
```diff
@@ -0,0 +1,58 @@
...
+	protected final Model<? super D> delegate;
+	protected final boolean setDelegateAngles;
+
+	/**
+	 * @param source the model whose transforms will be copied
+	 * @param delegate the model that will be rendered with transforms copied from the source model
+	 * @param setDelegateAngles {@code true} if the {@link Model#setAngles(Object)} method should be called for the
+	 *                                         delegate model after it is called for the source model
+	 */
+	public TransformCopyingModel(Model<? super S> source, Model<? super D> delegate, boolean setDelegateAngles) {
```
modmuss50: We tend not to use constructors in public API, I would suggest making this an interface with a factory method that creates an instance of the implementation. Or is it expected for someone to extend this class?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/TransformCopyingModel.java
```diff
@@ -0,0 +1,58 @@
...
+ * A model that copies transforms from a source model to a delegate model.
+ *
+ * <p>Useful in cases where an overlay model may not be a subclass of its base model.
+ * @param <S> type of the source model state
+ * @param <D> type of the delegate model state
+ */
+public class TransformCopyingModel<S, D> extends Model<Pair<S, D>> {
+	protected final Model<? super S> source;
+	protected final Model<? super D> delegate;
+	protected final boolean setDelegateAngles;
```
modmuss50: Again, if this is expected to be extended getters would be better.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -47,6 +68,54 @@ static void register(ArmorRenderer renderer, ItemConvertible... items) {
...
+	 *                              delegate model after it is called for the source model
+	 * @param queue					the {@link RenderCommandQueue}
+	 * @param matrices				the matrix stack
+	 * @param renderLayer			the render layer
+	 * @param light					packed lightmap coordinates
+	 * @param overlay				packed overlay texture coordinates
+	 * @param tintedColor			the color to tint the model with
+	 * @param sprite				the sprite to render the model with, or {@code null} to use the render layer instead
+	 * @param outlineColor			the outline color of the model
+	 * @param crumblingOverlay		the crumbling overlay, or {@code null} for no crumbling overlay
```
modmuss50: This in an uncomfortable amount of parms, I dont know enough about the target area to say if this is something we want or not.

modmuss50 (commented):
Generally looks great, just a few questions.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/TransformCopyingModel.java
8s2: I'm not sure. The `submitModel` method expects a model, which an interface could not extend. This means that users will need to use the `TransformCopyingModel` in its `Impl` form in order to actually render it, which I'm pretty sure is something that should be avoided too.
I can't think of a situation where a user would want to extend the `TransformCopyingModel`, so would it instead be better to make the class final, the constructor private, and add a static `create` method?

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
8s2: Yeah it's definitely a lot. Aside from the params used for creating a `TransformCopyingModel`, the rest are params for the `submitModel` method itself, so I don't really know if the number of params could be reduced.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/TransformCopyingModel.java
modmuss50: > I can't think of a situation where a user would want to extend the TransformCopyingModel, so would it instead be better to make the class final, the constructor private, and add a static create method?

This would be better yes.

modmuss50 (approved):
Looks good to me, but really needs someone familar with this area to approve.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -47,6 +68,54 @@ static void register(ArmorRenderer renderer, ItemConvertible... items) {
...
+	 * @param <S>					state type of the source model
+	 * @param <D>					state type of the delegate model
```
Juuxel: Nit: the alignment should be done with space chars. This list looks weird and unaligned with any other tab width than 4.

For example, this is how it looks in my github:
misaligned docs

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/ArmorRenderer.java
```diff
@@ -47,6 +68,54 @@ static void register(ArmorRenderer renderer, ItemConvertible... items) {
...
+	 *                                             delegate model after it is called for the source model
+	 * @param queue					the {@link RenderCommandQueue}
+	 * @param matrices				the matrix stack
+	 * @param renderLayer			the render layer
+	 * @param light					packed lightmap coordinates
+	 * @param overlay				packed overlay texture coordinates
+	 * @param outlineColor			the outline color of the model
+	 * @param crumblingOverlay		the crumbling overlay, or {@code null} for no crumbling overlay
+	 * @param <S>					state type of the source model
+	 * @param <D>					state type of the delegate model
```
Juuxel: Same here

Juuxel (approved):
Looks generally fine to me.

I have a custom ArmorRenderer in 1.21.10 that doesn't need these changes, but I assume that's because I render block models instead of entity models.

PR #4921, opened 2025-10-11
Move initial player attachments sync to the end of `PlayerManager::onPlayerConnect`

Fuzss:
Hey, I just stumbled across an issue with player attachment syncing. When logging in, existing attachment are never synced as the hook for that (via `ServerPlayConnectionEvents.JOIN` in `PlayerManager::onPlayerConnect`) runs too early. Presumable before `syncedAttachments` is set up in `AttachmentTargetsMixin` which happens from `fabric_readAttachmentsFromNbt`.

This can easily be fixed by invoking syncing logic at the end of `PlayerManager::onPlayerConnect`, which I can confirm to work and which is also what NeoForge does.


ekulxam (approved):

Syst3ms (approved):
LGTM. Do you think it would be possible to add a regression test for this?

apple502j (approved):
In another PR please update javadoc for SPCE

PR #4924, opened 2025-10-12
Remove all @NotNull annotations

modmuss50:


GameLord2011:
Why do we dislike @NotNull? (joke)

modmuss50:
Sorry, I should have been more clear about this. Everything is not null by default, we use @Nullable to mark wherever something is nullable but dont want to put @NotNull on everything else. This PR only removes the annotation from the very few places that had it where it wasnt needed, nothing has actually changed.

GameLord2011:
I was making a joke, as @NotNull is a user and is tagged as a contributor in the release.

PR #4926, opened 2025-10-13
Implement an opt-in recipe sync.

Patbox:
Implements rather simple recipe syncing mechanism for fabric.
Works by client sending supported recipe serializers payload in config and server sending recipe sync packet on join/datapack reload.

Packet format is different than old vanilla one, as it groups recipes by serializers and  sends serializers as Identifiers.
This is required change, as recipe serializers aren't a registry synced anymore and recipes can be added purely server side, so removing that would break quite few mods.

Sync enabling is done on the recipe serializer basis, to avoid mods enabling syncing for recipes that require server only data, but use common / vanilla recipe type (looking mostly at crafting recipes)

Solves #4667


Spinoscythe:
How would you clear the recipes?

Patbox:
What do you mean by this here?

Spinoscythe:
whenever the client logs out, the synced recipe serializers should be cleared as to not overload the client.

Patbox:
Due to how it's implemented, synced recipes would just get thrown away as soon as vanilla no longer holds it's ClientRecipeManager class

Patbox:
Might switch to using single packet, but not sure about that yet. Since with #4841 it shouldn't be an issue

Patbox:
I think it's pretty much done, with maybe some small tweaks and changes to be done (and renames if needed).

Spinoscythe:
> I think it's pretty much done, with maybe some small tweaks and changes to be done (and renames if needed).

`RecipeSyncS2CPayload` perhaps

Patbox:
Was thinking more on api side, but also current name matches ingredient packet names

Patbox:
Did a quick test by hooking up JEI and it looks to work nicely.

Spinoscythe:
In the PR desc could you do "Fixes" instead of "solves" so that GitHub will automatically close the issue when this is merged.

Poopooracoocoo:
"resolves" and "closes" also work https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

TempDev-2:
> Did a quick test by hooking up JEI and it looks to work nicely. 

How did you hook it up to JEI? I'm new to api development but not mod development, I tried building fabric api source like a mod in my IDE but it isnt working, before applying the PR. Perhaps its just an issue with my IDE. Additionally do you think this PR is or may be 1.21.x compatible?

Patbox:
Made JEI listen to the client recipe sync event to set it's recipe map and moved it's own after recipe sync event to run after that. Also marked all vanilla recipe serializers as synced. And buildscript tweaks to load fabric from maven local ofc

Spinoscythe:
What's stopping this from being merged besides modmuss needing to take a look?

Patbox:
Pretty much waiting for some reviews from fabric team (and probably some changes related to those).

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
```diff
@@ -0,0 +1,59 @@
...
+
+	private static void onRecipeSyncPacket(RecipeSyncPayloadS2C payload, ClientPlayNetworking.Context context) {
+		SynchronizedRecipes recipes;
+
+		if (!payload.entries().isEmpty()) {
+			var collectedRecipes = new ArrayList<RecipeEntry<?>>();
+
+			for (RecipeSyncPayloadS2C.Entry entry : payload.entries()) {
+				collectedRecipes.addAll(entry.recipes());
+			}
```
modmuss50, line 45: Are we happy to always sync all recipes? Is there a use case where only a small handful of new recipes may want syncing. For example if you only wanted the client to know about recipes unlocked via advancements. I believe there are also mods that may remove recipes dynamically as well.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
```diff
@@ -0,0 +1,59 @@
...
+
+			// Sort values by id to match ordering with server ones.
+			collectedRecipes.sort(Comparator.comparing(entry -> entry.id().getValue()));
```
modmuss50, line 48: Why wouldnt they be sorted in the payload?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/SynchronizedRecipes.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @return the stream of matching recipes
+	 */
+	<I extends RecipeInput, T extends Recipe<I>> Stream<RecipeEntry<T>> getAllMatches(RecipeType<T> type, I input, World world);
+
+	/**
+	 * @return the collection of recipe entries of given type
+	 */
+	<I extends RecipeInput, T extends Recipe<I>> Collection<RecipeEntry<T>> getAllOfType(RecipeType<T> type);
+
+	default <I extends RecipeInput, T extends Recipe<I>> Optional<RecipeEntry<T>> getFirstMatch(RecipeType<T> type, I input, World world, @Nullable RegistryKey<Recipe<?>> recipe) {
```
modmuss50, line 64: Nit: please document the following functions, even if its ovioubs.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/SynchronizedRecipes.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * This class provides access to synchronized recipes on the client.
+ *
+ * <p>You can access SynchronizedClientRecipes by calling getSynchronizedRecipes
+ * method on {@link net.minecraft.recipe.RecipeManager}
+ *
+ * <p>See {@link RecipeSynchronization}.
+ */
+@ApiStatus.NonExtendable
+public interface SynchronizedRecipes {
```
modmuss50, line 42: Question: Whats the thinking behind having this on the server as well, could it just be on the client?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/RecipeSyncRequestPayloadC2S.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import net.minecraft.network.PacketByteBuf;
+import net.minecraft.network.codec.PacketCodec;
+import net.minecraft.network.codec.PacketCodecs;
+import net.minecraft.network.packet.CustomPayload;
+import net.minecraft.util.Identifier;
+
+public record RecipeSyncRequestPayloadC2S(Set<Identifier> synchronizedSerializers) implements CustomPayload {
```
modmuss50: Nit: A comment documenting what these payloads are for might be useful, took me a little bit to figure out why we need a C2S packet.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/sync/PlayerManagerMixin.java
```diff
@@ -0,0 +1,50 @@
...
+
+	@Inject(method = "onPlayerConnect", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/network/ServerRecipeBook;sendInitRecipesPacket(Lnet/minecraft/server/network/ServerPlayerEntity;)V"))
+	private void sendRecipesOnJoin(ClientConnection connection, ServerPlayerEntity player, ConnectedClientData clientData, CallbackInfo ci) {
+		RecipeSyncImpl.sendRecipes(server, player);
+	}
+
+	@Inject(method = "onDataPacksReloaded", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/network/ServerRecipeBook;sendInitRecipesPacket(Lnet/minecraft/server/network/ServerPlayerEntity;)V"))
+	private void sendRecipesOnReload(CallbackInfo ci, @Local ServerPlayerEntity player) {
+		RecipeSyncImpl.sendRecipes(server, player);
+	}
```
modmuss50: Dont we have events for both of these?

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/RecipeSynchronization.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	/**
+	 * Enables synchronization for all recipes using provided RecipeSerializer.
+	 *
+	 * <p>This method should be only used to mark recipe serializers as synchronized only
+	 * if they are provided by your own mod or are vanilla ones.
+	 * Blindly adding unchecked recipe serializers might cause bugs and crashes.
+	 *
+	 * @param serializer recipe serializer used by synchronized recipes.
```
modmuss50: This should be made clear that it needs to be called on both the client and server.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
```diff
@@ -0,0 +1,59 @@
...
+			recipes = SynchronizedRecipesImpl.of(collectedRecipes);
+		} else {
+			recipes = SynchronizedRecipesImpl.EMPTY;
+		}
+
+		context.client().execute(() -> {
```
modmuss50: We should already be on the main thread.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/RecipeSyncImpl.java
```diff
@@ -0,0 +1,96 @@
...
+import net.fabricmc.fabric.mixin.recipe.sync.ServerRecipeManagerAccessor;
+
+public class RecipeSyncImpl implements ModInitializer {
+	private static final Set<RecipeSerializer<?>> SYNCED_SERIALIZERS = new ReferenceOpenHashSet<>();
+
+	@Override
+	public void onInitialize() {
+		PayloadTypeRegistry.configurationC2S().register(RecipeSyncRequestPayloadC2S.ID, RecipeSyncRequestPayloadC2S.CODEC);
+		// Recipe packet might contain a lot of data depending on mods, so it's best to increase it's max size to 50 MB.
+		PayloadTypeRegistry.playS2C().registerLarge(RecipeSyncPayloadS2C.ID, RecipeSyncPayloadS2C.CODEC, 50 * 1024 * 1024);
```
modmuss50: Nit: max size can be a constant.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/RecipeSyncImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+		ServerConfigurationNetworking.registerGlobalReceiver(RecipeSyncRequestPayloadC2S.ID, RecipeSyncImpl::onRecipeSyncRequest);
+	}
+
+	private static void onRecipeSyncRequest(RecipeSyncRequestPayloadC2S payload, ServerConfigurationNetworking.Context context) {
```
modmuss50: I think "SyncRequest" is the wrong name for this, its not requesting anything. Its just a way for the client to let the server know what it can accept.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/mixin/recipe/client/sync/ClientConfigurationNetworkHandlerMixin.java
```diff
@@ -0,0 +1,50 @@
...
+@Mixin(ClientConfigurationNetworkHandler.class)
+public class ClientConfigurationNetworkHandlerMixin {
+	@Inject(method = "onSelectKnownPacks", at = @At("TAIL"))
+	private void sendSupportedRecipeSerializers(SelectKnownPacksS2CPacket packet, CallbackInfo ci) {
+		if (ClientConfigurationNetworking.canSend(RecipeSyncRequestPayloadC2S.ID)) {
+			var ids = new HashSet<Identifier>();
+
+			for (RecipeSerializer<?> serializer : RecipeSyncImpl.getSyncedSerializers()) {
+				ids.add(Registries.RECIPE_SERIALIZER.getId(serializer));
+			}
```
modmuss50: Could this be done during configuration? As far as I can see the synced serialisers should all be known before datapacks are loaded.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/mixin/recipe/client/sync/ClientConfigurationNetworkHandlerMixin.java
```diff
@@ -0,0 +1,50 @@
...
+
+import net.fabricmc.fabric.api.client.networking.v1.ClientConfigurationNetworking;
+import net.fabricmc.fabric.impl.recipe.sync.RecipeSyncImpl;
+import net.fabricmc.fabric.impl.recipe.sync.RecipeSyncRequestPayloadC2S;
+
+@Mixin(ClientConfigurationNetworkHandler.class)
+public class ClientConfigurationNetworkHandlerMixin {
+	@Inject(method = "onSelectKnownPacks", at = @At("TAIL"))
+	private void sendSupportedRecipeSerializers(SelectKnownPacksS2CPacket packet, CallbackInfo ci) {
+		if (ClientConfigurationNetworking.canSend(RecipeSyncRequestPayloadC2S.ID)) {
```
modmuss50: Nit: early return.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricRecipeManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+import net.fabricmc.fabric.api.recipe.v1.sync.SynchronizedRecipes;
+import net.fabricmc.fabric.impl.recipe.sync.SynchronizedRecipesImpl;
+
+/**
+ * General-purpose Fabric-provided extensions for {@link RecipeManager} class.
+ */
+public interface FabricRecipeManager {
+	default SynchronizedRecipes getSynchronizedRecipes() {
+		return SynchronizedRecipesImpl.EMPTY;
```
modmuss50, line 30: Throw here, if the mixin doesnt impliment this it should crash.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/sync/ClientConnectionMixin.java
```diff
@@ -0,0 +1,43 @@
...
+
+import net.minecraft.network.ClientConnection;
+import net.minecraft.recipe.RecipeSerializer;
+
+import net.fabricmc.fabric.impl.recipe.sync.SyncedSerializerAwareClientConnection;
+
+@Mixin(ClientConnection.class)
+public abstract class ClientConnectionMixin implements SyncedSerializerAwareClientConnection {
+	@Unique
+	private Set<RecipeSerializer<?>> fabric_syncedRecipeSerializers = Set.of();
```
modmuss50: Remove `fabric_` prefex, its not needed with Unique

modmuss50 (commented):
I worry this doesnt cover a lot of possible use cases, mostly around dyanmic recipes that are locked/unlocked. Right now there is no control over what specific recipes are actually synced to each player. Maybe I'm just over thinking it? It could be fine for the client to know about these recipes, and they could just be hidden client side?

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
Patbox: Since they are split by recipe serializer, which causes them to be rearranged in packet.

modmuss50: Ah yeah, I see. 👍

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/SynchronizedRecipes.java
Patbox: Mostly simplifies the implementation. Think of using recipes in a screen handler or alike. If this was client only, you would need different client and server implementations. Having it commons allows to handle it the same way on both sides (see vanilla stonecutter for example)

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
Patbox: For recipes unlocked via advancements / that are dynamic it probably makes more sense to use vanilla recipe book system instead. This is useful in cases where you always want all recipes to be synced (think stonecutter, recipe viewers, etc)

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/mixin/recipe/sync/PlayerManagerMixin.java
Patbox: There is ServerLifecycleEvents.SYNC_DATA_PACK_CONTENTS, through this module currently doesn't depend on it. I guess it's in slightly different place too (later) than that event? I can switch it to it, but would probably make add a new phase after default and send it from there.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/impl/recipe/sync/client/RecipeSyncImplClient.java
Patbox: I wrote this before you implemented that change I think? Which to be fair I don't agree that packets should run always on main thread but guess it's bit late for that now.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/RecipeSyncImpl.java
Patbox: Yeah wasn't sure about it either. Maybe rename it to RecipeSyncSupportedSerializersPayloadC2S? Or something else if you have any suggestions.

fabric-recipe-api-v1/src/client/java/net/fabricmc/fabric/mixin/recipe/client/sync/ClientConfigurationNetworkHandlerMixin.java
Patbox: This is done during configuration.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricRecipeManager.java
Patbox: This is a fail safe in case someone does custom RecipeManager since that can happen (at least one of my mods does that, through purely as a stub).

Patbox:
In case of recipe viewers, constant updates would cause extra work as they need to (re)make the item <-> recipe relations and all that stuff. So for that hiding recipes purely on client would work best (and any mod using this for recipe viewer integration). For other cases where you also want recipe sync outside of what vanilla provides, you also likely would want to sync everything.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/SynchronizedRecipes.java
Patbox: Oh forgot to mention, client also only exposes RecipeManager in it's methods, so casting to ClientRecipeManager would be required or problematic side only method extension

modmuss50 (approved):

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/FabricRecipeManager.java
```diff
@@ -26,6 +26,7 @@
  */
 public interface FabricRecipeManager {
 	default SynchronizedRecipes getSynchronizedRecipes() {
+		// Fallback implementation in ase someone implements RecipeManager interface on custom class.
```
MattiDragon: Typo: should be `in case`

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/RecipeSyncImpl.java
```diff
@@ -0,0 +1,103 @@
...
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
+import net.fabricmc.fabric.api.networking.v1.PayloadTypeRegistry;
+import net.fabricmc.fabric.api.networking.v1.ServerConfigurationNetworking;
+import net.fabricmc.fabric.api.networking.v1.ServerPlayNetworking;
+import net.fabricmc.fabric.mixin.recipe.sync.ServerCommonNetworkHandlerAccessor;
+import net.fabricmc.fabric.mixin.recipe.sync.ServerRecipeManagerAccessor;
+
+public class RecipeSyncImpl implements ModInitializer {
+	// Recipe packet might contain a lot of data depending on mods, so it's best to increase it's max size to 64 MB.
```
Spinoscythe, line 42: Did you mean megabytes (MB) or mega**bits**(Mb) here? cus you're doing Mb (1024 bits)

Patbox: Megabytes, as that's what's used for counting size of packet and all

Spinoscythe (approved):
Sweet.

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/RecipeSynchronization.java
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * Enables synchronization for all recipes using provided RecipeSerializer.
+	 *
+	 * <p>This method should be only used to mark recipe serializers as synchronized only
+	 * if they are provided by your own mod or are vanilla ones.
+	 * Blindly adding unchecked recipe serializers might cause bugs and crashes.
+	 *
+	 * <p>This method should be called in mod initializer on both client and server
+	 *
+	 * @param serializer recipe serializer used by synchronized recipes.
```
mezz:
```suggestion
	 * Enables synchronization of recipes to the client, for recipes that can be handled by 
	 * the provided RecipeSerializer.
	 *
	 * <p>Only add recipe serializers that are provided by your own mod or vanilla.
	 * Blindly adding unchecked recipe serializers might cause bugs and crashes.
	 *
	 * <p>This method should be called in the mod initializer of both the client and the server.
	 *
	 * @param serializer the recipe serializer used to synchronize recipes to the client.
```

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/RecipeSynchronization.java
```diff
@@ -0,0 +1,63 @@
...
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
+import net.fabricmc.fabric.impl.recipe.sync.RecipeSyncImpl;
+
+/**
+ * Since Minecraft 1.21.2, vanilla no longer syncs all recipes to the client automatically,
+ * opting into sending only required recipe book data.
+ *
+ * <p>This api allows to enable Fabric's recipe sync for select RecipeSerializers, which can be
+ * then used on the client.
```
mezz:
```suggestion
 * <p>This api can be used to enable Fabric's recipe sync for select RecipeSerializers, 
 * which will synchronize recipes to the client.
```

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/api/recipe/v1/sync/RecipeSynchronization.java
```diff
@@ -0,0 +1,63 @@
...
+ * See {@link SynchronizedRecipes}
+ */
+public final class RecipeSynchronization {
+	/**
+	 * Event phase used for sending recipes to the client. It runs after default event phase {@link Event#DEFAULT_PHASE}.
+	 * It's defined for {@link ServerLifecycleEvents#SYNC_DATA_PACK_CONTENTS} event.
```
mezz:
```suggestion
	 * Event phase used for sending recipes to the client. It runs after the default event phase {@link Event#DEFAULT_PHASE}.
	 * It's defined for {@link ServerLifecycleEvents#SYNC_DATA_PACK_CONTENTS} event.
```

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/SupportedRecipeSerializersPayloadC2S.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Set;
+
+import net.minecraft.network.PacketByteBuf;
+import net.minecraft.network.codec.PacketCodec;
+import net.minecraft.network.codec.PacketCodecs;
+import net.minecraft.network.packet.CustomPayload;
+import net.minecraft.util.Identifier;
+
+/**
+ * Used to notify server which recipes can be synced to the client.
```
mezz, line 29: With this packet implementation, wont a fabric server with no mods still respond to requests from modded client for recipe data? If that's desired, maybe it should be made more clear in the PR description. @modmuss50

mezz (changes requested):
Nice work on this, thanks!

fabric-recipe-api-v1/src/main/java/net/fabricmc/fabric/impl/recipe/sync/SupportedRecipeSerializersPayloadC2S.java
modmuss50: I agree, I dont see a reason to send either packet if they are going to be empty.

mezz (approved):

PR #4932, opened 2025-10-18
Firework Star Explosion Type Registry

8s2:
Adds a registry for firework explosion types for items when crafting a firework star.

### DIFF
--- a/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FireworkStarExplosionTypeRegistry.java
+++ b/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FireworkStarExplosionTypeRegistry.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FireworkStarExplosionTypeRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.component.type.FireworkExplosionComponent;
+import net.minecraft.item.ItemConvertible;
+import net.minecraft.recipe.FireworkStarRecipe;
+
+import net.fabricmc.fabric.impl.content.registry.FireworkStarExplosionTypeRegistryImpl;
+
+/**
+ * A registry for the {@link FireworkExplosionComponent.Type} of an item for the {@link FireworkStarRecipe}.
+ */
+public final class FireworkStarExplosionTypeRegistry {
```
ekulxam, line 33: The name of this class seems misleading, as it appears to be a registry for the crafting recipe's map rather than the actual Type enum. Consider changing this to `FireworkStarExplosionTypeRecipeRegistry` (although this also seems a bit verbose?).

fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/registry/FireworkStarExplosionTypeRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+
+	private FireworkStarExplosionTypeRegistry() {
+	}
+
+	/**
+	 * Registers an explosion type for the item.
+	 * @param item the item to register
+	 * @param explosionType the firework explosion type
+	 */
+	public static void register(ItemConvertible item, FireworkExplosionComponent.Type explosionType) {
```
ekulxam, line 44: I believe the logic in here should be moved to the impl class, and this method should only be calling a register method in the impl class instead.

ekulxam (commented):
Looks good, just left some nit comments.

Liam-Broome:
Hi @8s2, I see the PR’s been inactive for a while. I’m interested in helping push it forward.
Are you happy for me to do so? 😄

8s2:
This is now data-driven in 26.1-snapshot-5

PR #4938, opened 2025-10-19
Migrate registry sync impl to use large packets from networking API

Earthcomputer:
Note: this PR can't be backported because it's a breaking change to the protocol.

I've tried to keep code changes minimal (e.g. the packet encoding is pretty much copy pasted from the old code), but I did take the opportunity to remove the unnecessary abstraction in `RegistryPacketHandler` which allowed me to simplify a couple of things.

I removed the `DirectRegistryPacketHandlerTest` as I believe it's basically testing the old packet splitting logic, which is now handled by the networking API (for which I'd hope there are already tests).


fabric-registry-sync-v0/src/client/java/net/fabricmc/fabric/impl/client/registry/sync/ClientRegistrySyncHandler.java
```diff
@@ -36,59 +35,44 @@
...
 			try {
-				apply(data);
-				return true;
-			} catch (RemapException e) {
-				throw new CompletionException(e);
+				apply(payload);
+				context.responseSender().sendPacket(SyncCompletePayload.INSTANCE);
+			} catch (Throwable e) {
+				LOGGER.error("Registry remapping failed!", e);
+				context.client().execute(() -> context.responseSender().disconnect(getText(e)));
```
apple502j: This appears to be doubly wrapped.

apple502j (changes requested):

fabric-registry-sync-v0/src/client/java/net/fabricmc/fabric/impl/client/registry/sync/ClientRegistrySyncHandler.java
Earthcomputer: This was copied from the old code but yes it seems unnecessary.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/DirectRegistryPacketHandler.java
```diff
@@ -17,75 +17,96 @@
...
- * <ul>
- *     <li>Directly write into the buffer instead of using an nbt;</li>
- *     <li>Group all {@link Identifier} with same namespace together and only send those unique namespaces once for each group;</li>
- *     <li>Group consecutive rawIds together and only send the difference of the first rawId and the last rawId of the bulk before.
- *     This is based on the assumption that mods generally register all of their object at once,
- *     therefore making the rawIds somewhat densely packed.</li>
- * </ul>
- *
- * <p>This method also split into multiple packets if it exceeds the limit, defaults to 1 MB.
- */
```
modmuss50, line 60: This javadoc is very useful, it should remain.

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/RegistrySyncPayload.java
```diff
@@ -17,75 +17,109 @@
...
+			RegistryKey<Registry<Object>> registryKey = RegistryKey.ofRegistry(regId);
+			RegistryAttributeImpl holder = (RegistryAttributeImpl) RegistryAttributeHolder.get(registryKey);
+			registryAttributes.put(regId, holder.getAttributes());
+		});
+		return registryAttributes;
+	}
 
-	@Nullable
-	private Map<Identifier, Object2IntMap<Identifier>> syncedRegistryMap;
+	private static RegistrySyncPayload read(PacketByteBuf combinedBuf) {
```
deirn, line 76: Is there a style rule that static functions must be before methods? 
Would've made the diff better, and also reads better.

deirn (approved):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/RegistrySyncPayload.java
Earthcomputer: No there isn't, I don't think either way reads better either, and I don't think enforcing a general rule on it would improve code quality

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/RegistrySyncPayload.java
```diff
@@ -17,75 +17,109 @@
...
-	 * @see net.minecraft.network.packet.s2c.play.CustomPayloadS2CPacket#MAX_PAYLOAD_SIZE
-	 */
-	@SuppressWarnings("JavadocReference")
-	private static final int MAX_PAYLOAD_SIZE = Integer.getInteger("fabric.registry.direct.maxPayloadSize", 0x100000);
+public record RegistrySyncPayload(
+		Map<Identifier, Object2IntMap<Identifier>> registryMap,
+		Map<Identifier, EnumSet<RegistryAttribute>> registryAttributes
+) implements CustomPayload {
+	public static CustomPayload.Id<RegistrySyncPayload> ID = new CustomPayload.Id<>(Identifier.of("fabric", "registry/sync"));
+	public static PacketCodec<PacketByteBuf, RegistrySyncPayload> CODEC = CustomPayload.codecOf(RegistrySyncPayload::write, RegistrySyncPayload::read);
```
Juuxel:
```suggestion
	public static final CustomPayload.Id<RegistrySyncPayload> ID = new CustomPayload.Id<>(Identifier.of("fabric", "registry/sync"));
	public static final PacketCodec<PacketByteBuf, RegistrySyncPayload> CODEC = CustomPayload.codecOf(RegistrySyncPayload::write, RegistrySyncPayload::read);
```
I don't think anything modifies these

Juuxel (changes requested):

fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/impl/registry/sync/packet/RegistrySyncPayload.java
Earthcomputer: It's in a record, `final` is implied.

Earthcomputer: Wait wtf, how do I unsend a message

LambdAurora (approved):

modmuss50 (approved):

PR #4939, opened 2025-10-21
Only sync attachment change if client supports the AttachmentType

DennisOchulor:
Currently when syncing initial attachment data, only the attachments that the client supports are synced. However, for attachment changes outside the initial sync, client support for the attachment is not checked. 

This means that when a mod with syncable attachents is installed on the server but not the client (and the client has Fabric API installed), the client will get disconnected with a `io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:custom_payload'` error.

This PR fixes that issue.

### DIFF
--- a/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentSync.java
+++ b/fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentSync.java
@@ -51,13 +51,12 @@ public class AttachmentSync implements ModInitializer {
 		return new AcceptedAttachmentsPayloadC2S(AttachmentRegistryImpl.getSyncableAttachments());
 	}
 
-	public static void trySync(AttachmentSyncPayloadS2C payload, ServerPlayerEntity player) {
+	public static void trySync(AttachmentChange change, ServerPlayerEntity player) {
 		Set<Identifier> supported = ((SupportedAttachmentsClientConnection) ((ServerCommonNetworkHandlerAccessor) player.networkHandler).getConnection())
 				.fabric_getSupportedAttachments();
 
-		if (!payload.attachments().isEmpty() && supported.contains(payload.attachments().getFirst().type().identifier())) {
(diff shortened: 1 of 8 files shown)


fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/sync/AttachmentSync.java
```diff
@@ -52,7 +52,11 @@ public static AcceptedAttachmentsPayloadC2S createResponsePayload() {
 	}
 
 	public static void trySync(AttachmentSyncPayloadS2C payload, ServerPlayerEntity player) {
-		if (!payload.attachments().isEmpty()) {
+		Set<Identifier> supported = ((SupportedAttachmentsClientConnection) ((ServerCommonNetworkHandlerAccessor) player.networkHandler).getConnection())
+				.fabric_getSupportedAttachments();
+
+		if (!payload.attachments().isEmpty() && supported.contains(payload.attachments().getFirst().type().identifier())) {
+			// the payload will always contain exactly one AttachmentChange
```
modmuss50: Its likely better to just throw if there is more than 1 attachment, will self document and make it impossible for the comment to be wrong in the future.

DennisOchulor: I actually thought about this and I'm wondering if the method should just accept a single `AttachmentChange`. The method is called from `AttachmentTargetImpl.syncChange()` and that is called from `setAttached()`. So it is guaranteed that there is only one `AttachmentChange` (as seen below). The payload object can then just be created inside `trySync()` if the client supports it.

image

DennisOchulor: I've changed it to `AttachmentChange`, so now it's clear there's only 1 attachment.

skycatminepokie (approved):

apple502j (approved):

modmuss50 (approved):

PR #4942, opened 2025-10-22
Add InventoryEvents to fabric-screen-handler-api-v1 with Test Mod

OhACD:
This PR adds 'InventoryEvents' into the 'screen-handler-api-v1' module, providing a simple way to handle inventory slot click events, the API introduces 'InventoryEvents.SLOT_CLICK_EVENT', which allows modders to intercept and modify slot interactions (e.g, 'PICKUP', 'SWAP', 'QUICK_MOVE', 'QUICK_CRAFT') with minimal boilerplate, reducing reliance on complex mixins or custom screen handlers. 

The plan was for 3 events to be added SLOT_CLICK_EVENT, INVENTORY_OPEN_EVENT and CRAFT_RESULT_EVENT.

**Problems Solved**:
- Simplifies inventory event handling, avoiding invasive modifications to 'ScreenHandler' or 'Slot' classes.
- Reduce mod conflicts by using an event-driven approach.
- Supports common use cases like item restriction and custom container mechanics.

**Test Mod**:
Added 'InventoryEventsTest' to the existing test mod, demonstrating Diamond Protection (prevents moving Diamonds in any container)

**Changes**:
- Added 'InventoryEvents.java' to 'screen-handler-api-v1'.
- Added 'InventoryEventsTest.java' to 'testmod'.
- Updated testmod main class to call 'InventoryEventsTest().onInitialize()'.
- All classes added have the necessary javadocs for ease of use.

**Testing Instructions**:
1. Build and run the mod in a Fabric environment
2. Try taking out Diamonds from a container
3. Test multiple actions to take out the Diamond

### DIFF
--- a/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/inventory/SlotClickCallback.java
+++ b/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/inventory/SlotClickCallback.java
@@ -1,0 +1,47 @@
+package net.fabricmc.fabric.api.inventory;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ItemStack;
+import net.minecraft.screen.ScreenHandler;
+import net.minecraft.screen.slot.Slot;
+import net.minecraft.screen.slot.SlotActionType;
+import net.minecraft.util.ActionResult;
+
+/**
+ * Callback interface for handling slot click events in inventory GUIs.
+ *
+ * <p>Runs BEFORE any slot click is processed to allow interception or cancellation of the action.
+ *
+ * <p>Use cases:
+ * <ul>
+ *     <li>Prevent shift-clicking protected items from containers</li>
+ *     <li>Implement cooldowns for container access</li>
+ *     <li>Perform permission checks for locked containers</li>
+ * </ul>
+ *
+ * <p>Canceling with {@code ActionResult.FAIL} stops the click entirely. Returning {@code ActionResult.SUCCESS} allows the action but consumes the event.
+ */
+@FunctionalInterface
+public interface SlotClickCallback {
+	/**
+	 * Called when a player clicks a slot in an inventory GUI.
+	 *
+	 * @param handler    ScreenHandler (e.g., chest, furnace, crafting table)
+	 * @param slot       Clicked slot (null if clicked outside inventory)
+	 * @param slotId     Raw slot ID from packet (0–215 for most GUIs, -1 for outside clicks)
+	 * @param button  Mouse button (0 = left, 1 = right) + modifiers
+	 * @param actionType Exact action: PICKUP, QUICK_MOVE, SWAP, etc.
+	 * @param player     The player who clicked (useful for permission checks)
+	 * @param cursor     Current cursor stack (what the player is holding, never null)
+	 * @return           PASS = continue to next listener, SUCCESS/FAIL = stop event chain
+	 */
+	ActionResult interact(
+			ScreenHandler handler,
+			Slot slot,
+			int slotId,
+			int button,
+			SlotActionType actionType,
+			PlayerEntity player,
+			ItemStack cursor
+	);
+}
--- a/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/mixin/inventory/ScreenHandlerMixin.java
+++ b/fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/mixin/inventory/ScreenHandlerMixin.java
@@ -1,0 +1,60 @@
+package net.fabricmc.fabric.mixin.inventory;
+
+import net.fabricmc.fabric.api.inventory.InventoryEvents;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ItemStack;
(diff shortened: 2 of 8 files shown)


OhACD:
I just noticed there is a mistake in the docs in the InventoryEvents class where it states that mod authors should manually sync the client state, I've written this before i was correctly syncing everything, It's not required currently and can be missleading

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/inventory/SlotClickCallback.java
```diff
@@ -0,0 +1,47 @@
...
+	 * @param slotId     Raw slot ID from packet (0–215 for most GUIs, -1 for outside clicks)
+	 * @param button  Mouse button (0 = left, 1 = right) + modifiers
+	 * @param actionType Exact action: PICKUP, QUICK_MOVE, SWAP, etc.
+	 * @param player     The player who clicked (useful for permission checks)
+	 * @param cursor     Current cursor stack (what the player is holding, never null)
+	 * @return           PASS = continue to next listener, SUCCESS/FAIL = stop event chain
+	 */
+	ActionResult interact(
+			ScreenHandler handler,
+			Slot slot,
```
Patbox, line 40: This should be marked as @ Nullable, since you do pass null here.

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/mixin/inventory/ScreenHandlerMixin.java
```diff
@@ -0,0 +1,60 @@
...
+	 * to prevent desyncs. If {@link ActionResult#SUCCESS} is returned, the event is consumed but vanilla processing continues.
+	 * {@link ActionResult#PASS} allows other listeners and vanilla logic to proceed.
+	 *
+	 * @param slotIndex   The raw slot ID from the packet (0–215 for most GUIs, -1 for outside clicks)
+	 * @param button      Mouse button (0 = left, 1 = right) + modifiers
+	 * @param actionType  The exact action: PICKUP, QUICK_MOVE, SWAP, etc.
+	 * @param player      The player who clicked the slot
+	 * @param ci          Callback info for canceling the method
+	 */
+	@Inject(method = "onSlotClick", at = @At("HEAD"), cancellable = true)
```
Patbox, line 38: If you want for it to hold true that it always runs before actual logic, you would need to move this mixin to network handler, as mods can and do override onSlotClick method

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/mixin/inventory/ScreenHandlerMixin.java
```diff
@@ -0,0 +1,60 @@
...
+		Slot slot = slotIndex >= 0 ? handler.getSlot(slotIndex) : null;
+		ItemStack cursor = handler.getCursorStack();
+		ActionResult result = InventoryEvents.SLOT_CLICK_EVENT.invoker().interact(
+				handler, slot, slotIndex, button, actionType, player, cursor
+		);
+
+		if (result == ActionResult.FAIL) {
+			ci.cancel();
+			// Sync client state to prevent desync
+			handler.syncState();
```
Patbox, line 57: I don't think you should sync it like this. I'm pretty sure vanilla should have system for detecting desyncs natively, if not you should use that instead of syncing entire inventory

Patbox (commented):
Gave some feedback on current implementation.
Also worth noting that this PR:
- This only handles simple cases of inventory slot click and others. Complex changes naturally will still require mixins.
- Doesn't really reduce mod conflicts, as proficient devs would know how to handle it.
- Custom container mechanics and item restrictions shouldn't be implemented purely via click detection, because for example a sorting mod on the server will skip these checks to be faster, which might cause bypasses (wouldn't be the case if you validated it in Slot class, as one should).

OhACD:
I understand, I got told to only start by adding a simple event. I also know that chest protection shouldn't be added like this, this is just a simple proof of concept that makes use of the simple event added. Thank you for the review!

fabric-screen-handler-api-v1/src/main/java/net/fabricmc/fabric/api/inventory/SlotClickCallback.java
```diff
@@ -0,0 +1,47 @@
...
+ * <ul>
+ *     <li>Prevent shift-clicking protected items from containers</li>
+ *     <li>Implement cooldowns for container access</li>
+ *     <li>Perform permission checks for locked containers</li>
+ * </ul>
+ *
+ * <p>Canceling with {@code ActionResult.FAIL} stops the click entirely. Returning {@code ActionResult.SUCCESS} allows the action but consumes the event.
+ */
+@FunctionalInterface
+public interface SlotClickCallback {
```
maityyy, line 25: Move this class to your `InventoryEvents`, example: https://github.com/FabricMC/fabric/blob/1.21.10/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerLifecycleEvents.java

edit: And rename to `SlotClick`, without "Callback"

PR #4944, opened 2025-10-27
Allow registration of data resource reloader using a factory and provide a store to store data reload results.

LambdAurora:
After some discussion on Discord, this PR aims to fix the headaches of using `JsonDataLoader` with Resource Loader v1.

This PR also introduce a new data store for results of a data resource reload as this was requested as well and it is true that currently there's no good way to store data in ways that respects its lifecycle.

There are some things I'm not yet sure in this design, notably with how to access such store using a `MinecraftServer` instance, right now a barebone `DataResourceStore` is interface injected onto it to have the least friction to gather data (think `server.getOrThrow(DATA_KEY)`), this should be safe in terms of injection as the method as a unique parameter to the resource loader (`DataResourceStore.Key`), the alternative would be to have a provider interface to inject and do `server.getDataResourceStore().getOrThrow(DATA_KEY)`.


fabric-resource-loader-v1/src/main/resources/fabric.mod.json
```diff
@@ -28,6 +28,9 @@
     }
   ],
   "custom": {
-    "fabric-api:module-lifecycle": "stable"
+    "fabric-api:module-lifecycle": "stable",
+    "loom:injected_interfaces": {
+      "net/minecraft/class_6897": ["net/fabricmc/fabric/api/resource/v1/DataResourceStore"]
```
modmuss50: Is this injected interface correct? class_6897 is an inner class of MinecraftServer, I see you are casting to this in the testmod this doesnt seem quite right. Let me know if you want me to take a look.

modmuss50 (commented):
Nothing stands out to me about the API, but get a review from someone who riased this issue in the first place. Something does look fishy about the injected interface tho.

fabric-resource-loader-v1/src/main/resources/fabric.mod.json
LambdAurora: Turns out I got bamboozled by `MinecraftServer` not having an intermediary name, I forgot about that, and I took the first intermediary name in linkie without realizing...

That explains why it wasn't working in the testmod :sweat_smile:

modmuss50 (approved):

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/MinecraftServerMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import net.fabricmc.fabric.impl.resource.v1.FabricDataResourceStoreHolder;
+
+@Mixin(MinecraftServer.class)
+public class MinecraftServerMixin implements DataResourceStore {
+	@Shadow
+	private MinecraftServer.ResourceManagerHolder resourceManagerHolder;
+
+	@SuppressWarnings("AddedMixinMembersNamePattern")
+	@Override
+	public <T> T getOrThrow(Key<T> key) {
```
Earthcomputer, line 34: Super generic method name mixined to `MinecraftServer` looks a little risky, but it's fine because we own the `Key` class.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/DataResourceLoader.java
```diff
@@ -0,0 +1,87 @@
...
+	 * @apiNote In most cases {@link #registerReloader(Identifier, ResourceReloader)} is sufficient and should be preferred,
+	 * but for some resource reloaders like {@link net.minecraft.resource.JsonDataLoader} constructing the resource reloader
+	 * with a known instance of the wrapper lookup is required.
+	 * <br/>
+	 * While this may encourage stateful resource reloaders, it is best to primarily use resource reloaders as stateless loaders,
+	 * as storing a state may easily lead to incomplete or leaking data.
+	 */
+	void registerReloader(
+			Identifier id,
+			Function<RegistryWrapper.WrapperLookup, ResourceReloader> factory
```
Earthcomputer, line 85: I'm a bit confused why we are using a `WrapperLookup` here, rather than the `DataResourceStore`. Is there a good reason?

Earthcomputer (approved):
The API looks good, it addresses my primary concern about not being able to use `JsonDataLoader` easily.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/DataResourceLoader.java
LambdAurora: Good question, the main need for this overload is for `JsonDataLoader` and it only needs `WrapperLookup`, I'm a bit concerned about how confusing the `DataResourceStore` might be here.
Also the `DataResourceStore` does **not** contain the `WrapperLookup`, the lookup and other reloaders are also available through the Vanilla ResourceReloader shared state store.

PR #4948, opened 2025-10-28
Add char typed events for handling `Screen::charTyped`

Fuzss:
Fabric was missing events for that, NeoForge already has them.

Came across this when implementing my own keybind that is processed when a screen is open.
For the creative inventory screen when the search tab is open there was no way to prevent my pressed key (when set to a letter) from adding a char to the search field, so these new events allow for preventing that.

### DIFF
--- a/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenKeyboardEvents.java
+++ b/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenKeyboardEvents.java
@@ -19,6 +19,7 @@ package net.fabricmc.fabric.api.client.screen.v1;
 import java.util.Objects;
 
 import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.client.input.CharacterEvent;
 import net.minecraft.client.input.KeyEvent;
 
 import net.fabricmc.fabric.api.event.Event;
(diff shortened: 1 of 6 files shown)


Fuzss:
Updated for 26.2. Pretty straight forward.

I also fixed some minor variable name inconsistencies that were left over from the Yarn era in classes I was editing anyways. If this is an issue it can easily be undone.

@modmuss50 Can this please be looked at?
It's very niche, but I think it's the last set of screen events missing to cover all of the input methods on the `Screen` class.

fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenKeyboardEvents.java
```diff
@@ -180,4 +214,41 @@ public interface AfterKeyRelease {
 		 */
 		void afterKeyRelease(Screen screen, KeyEvent event);
 	}
+
+	@FunctionalInterface
+	public interface AllowCharType {
+		/**
+		 * Checks if typing a character should be allowed.
+		 *
+		 * @param event the char type event, containing the codepoint and modifiers
```
modmuss50: CharacterEvent seems to only contain the codepoint, do we need to pass the modifiers though as well, or is the jdoc just outdated? 

https://mcsrc.dev/1/26.2/net/minecraft/client/input/CharacterEvent#L5

Fuzss: That was probably from copying the Javadoc from the classes above. Fixed now, thanks.

modmuss50 (approved):

PR #4950, opened 2025-10-28
Apply @Nullmarked annotation to all packages

modmuss50:
Most packages have a package-info generated for them, if not any other package-info must contain @Nullmarked


Toffikk:
~~Might be worth to migrate from the jetbrains’s @Nullable to jspecify’s @Nullable too, in existing classes to have annotation continuity. It just requires changing the imports so should be easy to do~~

edit: just saw there’s a pr for that, dismiss this^

modmuss50:
> Might be worth to migrate from the jetbrains’s @nullable to jspecify’s @nullable too, in existing classes to have annotation continuity. It just requires changing the imports so should be easy to do

Yes, see my other PR: https://github.com/FabricMC/fabric/pull/4951 I did it separately as its a massive change.

PR #4953, opened 2025-10-30
1.21.11 Update GameRule API

ekulxam:
Confirmed that Game Test, Client Game Test, and Auto Test Server all succeeded.

**Changes**
- Port to 1.21.11
- Remove every custom rule class. Everything is now handled by Minecraft's `GameRule`.
- Use local mappings (I will revert this commit once https://github.com/FabricMC/yarn/pull/4367 is merged)
- Fix test(mod) to the best of my ability
- TAW all register methods in `GameRules`
- Change callback seems to no longer exist, so all references to that no longer exist
- GameRuleFactory now creates `GameRule`s and not the `Type`
- GameRuleFactory defaults to `MISC` for the rule `Category`

Do we still need `CustomGameRuleCategory`? `Category` is no longer an enum and can be properly instantiated using `method_76160`, which somewhat defeats the purpose of having `CustomGameRuleCategory` in the first place. I only kept it because `CustomGameRuleCategory`'s name / translation key can be configured, but I'm willing to hear other viewpoints.

I don't know what is happening in `GameRuleManagementHandlerImplMixin`, but I think commenting out the updateRule handler was correct, as it doesn't seem to be different from what vanilla is doing now.

As a side note, my import order is super broken, and I don't know how to fix it.


offbeat-stuff:
In 25w44a gamerules update callbacks happen in a function in server world.

Can you add an event to easily register a callback for your gamerules?

fabric-game-rule-api-v1/src/test/java/net/fabricmc/fabric/test/gamerule/GameRuleManagementHandlerImplTest.java
```diff
@@ -120,19 +133,19 @@ void testUpdateVanillaInt() {
 		when(saveProperties.getGameRules()).thenReturn(GAME_RULES);
 		GameRuleManagementHandler handler = new GameRuleManagementHandlerTestImpl(server, MANAGEMENT_LOGGER);
 
-		GameRuleRpcDispatcher.TypedRule result = handler.updateRule(new GameRuleRpcDispatcher.UntypedRule("randomTickSpeed", "123"), CONNECTION_ID);
+		GameRuleRpcDispatcher.class_12254<Integer> result = handler.updateRule(new GameRuleRpcDispatcher.class_12254<>(GameRules.RANDOM_TICK_SPEED, 123), CONNECTION_ID);
```
Earthcomputer: Might be worth it to update yarn so these are named

ekulxam: Even after working with this for a few hours, I'm still not sure what I would name this if given the chance. TypedRule seems outdated becaues this class has basically replaced both TypedRule and UntypedRule, so perhaps a better name would be `RuleValueContainer` or `RuleValuePair`?

apple502j: Key-value pair is an entry, e.g. RegistryEntry.

(Hold(er) on, that's not how it's going to be named anymore...)

Earthcomputer: I just assumed that it was mapped in yarn already tbh

ekulxam:
e561067 can be reverted as necessary because it will probably be a pain to maintain. I spent the better part of a few hours figuring out the minimum `mock`s required to create a MinecraftDedicatedServer, so I understand if it is removed.

Earthcomputer:
@ekulxam why not make it into a game test?

ekulxam:
Amazing, can't believe I missed that. I suppose I'm learning game tests now.

build.gradle
```diff
@@ -119,6 +119,10 @@ allprojects {
 		enabled = false
 	}
 
+	repositories {
```
modmuss50: remove

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java
```diff
@@ -0,0 +1,55 @@
...
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Provides events for updating {@link GameRule}s.
+ */
+public final class GameRuleEvents {
+	private GameRuleEvents() { }
+
+	public static final Event<ValueUpdate> CHANGED_CALLBACK = EventFactory.createArrayBacked(
```
modmuss50: Would this be better as something along the following lines:

`<T> Event<ValueUpdate<T>> changedCallback<GameRule<T> gameRule)`

basically a static method that returns you the event for a specific gamerule, this way it can be typesafe.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleRegistry.java
```diff
@@ -16,52 +16,51 @@
...
- *
- * <p>Creation of rule types is done using {@link GameRuleFactory}.
  *
- * @see GameRuleFactory
+ * <p>
+ * Note that {@link net.minecraft.world.rule.GameRules} already
+ * contains many registration methods not found here.
  */
 public final class GameRuleRegistry {
+	public static final Logger LOGGER = LoggerFactory.getLogger(GameRuleRegistry.class);
```
modmuss50: private

modmuss50: Doesnst seem used so remove.

modmuss50 (commented):
Looking good, the main thing is the value changed event, I think it can be made typesafe quite easily.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java
```diff
@@ -0,0 +1,55 @@
...
+	@FunctionalInterface
+	public interface ValueUpdate {
+		/**
+		 * Called when a GameRule's value is updated in {@link MinecraftServer}.
+		 * @param rule the rule
+		 * @param value the updated value
+		 * @param server the server
+		 * @see MinecraftServer#onGameRuleUpdated(GameRule, Object)
+		 */
+		void accept(
```
apple502j: FAPI event handlers generally start with `on`

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleFactory.java
```diff
@@ -18,301 +18,168 @@
...
+ * For example, to register a game rule that is an integer where the default value is 1 and the acceptable values are between 0 and 10, one would use the following:
  * <blockquote><pre>
- * public static final GameRules.Key&lt;GameRules.IntRule&gt; EXAMPLE_INT_RULE = GameRuleRegistry.register("exampleIntRule", GameRules.Category.UPDATES, GameRuleFactory.createIntRule(1, 10));
+ * public static final GameRule&lt;Integer&gt; EXAMPLE_INT_RULE = GameRules.registerIntRule("example_int_rule", Category.MISC, 1, 0, 10);
  * </pre></blockquote>
+ * Please note that all register methods in {@link GameRules} internally construct and register the rule.
  *
- * <p>To register a game rule in a custom category, {@link GameRuleRegistry#register(String, CustomGameRuleCategory, GameRules.Type)} should be used.
+ * <p>To register a game rule in a custom category, {@link GameRuleRegistry#register(String, GameRule, CustomGameRuleCategory)} should be used.
+ * Alternatively, cast the {@link GameRule} to {@link RuleCategoryExtensions} and call {@link RuleCategoryExtensions#fabric_setCustomCategory(CustomGameRuleCategory)}.
```
apple502j: Remove, never link to impl in javadoc.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleFactory.java
```diff
@@ -18,301 +18,168 @@
...
- * <p>Some factory methods allow specification of a callback that is invoked when the value of a game rule has changed.
- * Typically, the callback is used for game rules which may influence game logic, such as {@link GameRules#DISABLE_RAIDS disabling raids}.
- *
- * <p>To register a game rule, you can use {@link GameRuleRegistry#register(String, GameRules.Category, GameRules.Type)}.
- * For example, to register a game rule that is an integer where the acceptable values are between 0 and 10, one would use the following:
+ * <p>To register a game rule, you can use {@link GameRuleRegistry#register(String, GameRule)} or any register method from {@link GameRules}.
+ * For example, to register a game rule that is an integer where the default value is 1 and the acceptable values are between 0 and 10, one would use the following:
  * <blockquote><pre>
- * public static final GameRules.Key&lt;GameRules.IntRule&gt; EXAMPLE_INT_RULE = GameRuleRegistry.register("exampleIntRule", GameRules.Category.UPDATES, GameRuleFactory.createIntRule(1, 10));
+ * public static final GameRule&lt;Integer&gt; EXAMPLE_INT_RULE = GameRules.registerIntRule("example_int_rule", Category.MISC, 1, 0, 10);
```
apple502j: Remove, as the register methods in GameRules take String, not Identifier.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleFactory.java
```diff
@@ -18,301 +18,168 @@
...
  * </pre></blockquote>
+ * Please note that all register methods in {@link GameRules} internally construct and register the rule.
  *
- * <p>To register a game rule in a custom category, {@link GameRuleRegistry#register(String, CustomGameRuleCategory, GameRules.Type)} should be used.
+ * <p>To register a game rule in a custom category, {@link GameRuleRegistry#register(String, GameRule, CustomGameRuleCategory)} should be used.
+ * Alternatively, cast the {@link GameRule} to {@link RuleCategoryExtensions} and call {@link RuleCategoryExtensions#fabric_setCustomCategory(CustomGameRuleCategory)}.
  *
  * @see GameRuleRegistry
  */
 public final class GameRuleFactory {
```
apple502j: If we are completely revamping the class, I think a redesign might be better. Why not a builder, for example?

Also rename the class since this is not a real factory (factory in Java is a functional interface for creating an instance)

apple502j (changes requested):
Stopping mid-review, this clearly needs more work.

ekulxam:
For `MinecraftServerMixin#invokeChangeCallbacks`, do we want to be able to not invoke anything if the event doesn't exist? In the current state, a new Event will be created for every GameRule after that rule has been changed. Is it worth it to only invoke Events for already existing ones (i.e. don't call computeIfAbsent for the invoker).

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java
```diff
@@ -0,0 +1,67 @@
...
+import net.fabricmc.fabric.api.event.EventFactory;
+
+/**
+ * Provides events for updating {@link GameRule}s.
+ */
+public final class GameRuleEvents {
+	private GameRuleEvents() {
+	}
+
+	private static final Map<GameRule<?>, Event<ValueUpdate<?>>> VALUE_UPDATES = new HashMap<>();
```
modmuss50: Move this to an impl class.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java
```diff
@@ -0,0 +1,67 @@
...
+	public static <T> Event<ValueUpdate<T>> changeCallback(GameRule<T> rule) {
+		//noinspection unchecked
+		return (Event<ValueUpdate<T>>) (Event<?>) VALUE_UPDATES.computeIfAbsent(rule, gameRule -> {
+			//noinspection unchecked
+			return (Event<ValueUpdate<?>>) (Event<?>) EventFactory.createArrayBacked(ValueUpdate.class, (Function<ValueUpdate<T>[], ValueUpdate<T>>) callbacks -> (value, server) -> {
+				for (ValueUpdate<T> changedCallback : callbacks) {
+					changedCallback.onGameRuleUpdated(value, server);
+				}
+			});
+		});
```
modmuss50: Move this to an impl class, you should only create the event when somone uses the api. When you invoke the event you can get null back from the map and not invoke anything.

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java
modmuss50: Can this be an IdentityHashMap?

fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/mixin/gamerule/MinecraftServerMixin.java
```diff
@@ -16,18 +16,20 @@
...
-public interface GameRulesBooleanRuleAccessor {
-	@Invoker
-	static GameRules.Type<GameRules.BooleanRule> invokeCreate(boolean initialValue, BiConsumer<MinecraftServer, GameRules.BooleanRule> changeCallback) {
-		throw new AssertionError("This shouldn't happen!");
+@Mixin(MinecraftServer.class)
+public abstract class MinecraftServerMixin {
+	@WrapMethod(method = "onGameRuleUpdated")
+	private <T> void invokeChangeCallbacks(GameRule<T> rule, T value, Operation<Void> original) {
+		original.call(rule, value);
+		GameRuleEvents.changeCallback(rule).invoker().onGameRuleUpdated(value, (MinecraftServer) (Object) this);
```
modmuss50: I dont think this is the correct place to inject this, as you will miss all of the gamerule changes that happened without a server. (See TestWorldBuilderImpl)

modmuss50 (approved):

apple502j:
Still some issues, @ekulxam can I just make a PR to your branch?

apple502j:
@ekulxam PRed, https://github.com/ekulxam/fapi/pull/1

apple502j (approved):

PR #4956, opened 2025-11-01
Add Enigma task for javadoc injection

Juuxel:
Adds a Gradle task for writing mod-provided javadoc directly in Enigma instead of writing mapping files manually. There's also an example javadoc file adding javadoc to the `Oxidizable` interface that points to FAPI's registry for oxidisable blocks.

---

Usage example: Run `gradlew :fabric-content-registries-v0:enigma` to open Enigma for that module. If the mapping file doesn't exist beforehand, you need to create it. Before adding it to the fabric.mod.json, make sure that it's not empty so that MIO can detect the format correctly.

You can then go to a class (or method/field) and document it:
Enigma showing class_5955 with injected javadoc

After generating sources, you will now see the added javadoc in the decompiled vanilla source:
IDEA showing Oxidizable with injected javadoc

---

Depends on #4952.

**Using mods with Enigma format javadoc files requires the latest Loom 1.12 build (1.12.5+) due to critical bug fixes.** Older versions of Loom will crash with Enigma format javadoc, and also don't work correctly with Tiny format javadoc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
 	id "java-library"
 	id "eclipse"
 	id "maven-publish"
-	id "fabric-loom" version "1.11.7" apply false
+	id "fabric-loom" version "1.12.5" apply false
 	id "com.diffplug.spotless" version "7.2.1"
 	id "me.modmuss50.remotesign" version "0.5.0" apply false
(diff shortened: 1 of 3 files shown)


modmuss50:
Manually merged in .11, thanks.

PR #4958, opened 2025-11-02
Fix render event testmod after port

modmuss50:
Copilot is better than me at drawing a box.


MichaelHillcox (approved):
Looks good to me. I did consider mentioning that you could use ShapeRender#renderShape (don't know what it is in yarn, sorry)

But I'm not 100% sure if that method works outside of lines. It might be the wrong vertex count.

PR #4969, opened 2025-11-04
rendering/api/ingido/model loader fixes

JumperOnJava:


JumperOnJava:
WARNING - i haven't tested these as some modules i depend on still break

modmuss50 (approved):

Issue #3782, opened 2024-05-19
You can change the order of the `Fabric Mods` resource pack to be below the vanilla resource pack


mrjasonn:
This should not work like this, it should always be at least on top of the vanilla resource pack. This should be fixed because some mods require their resources to be applied in order to work properly. An example of mods requiring their resources to be on top would be Sodium as it needs it's Better Fast Leaves textures to have leaves work properly in Fast graphics. As mods are supposed to override vanilla, this would make sense to do, while keeping the ability for custom resource packs to change textures. This shouldn't be taken as that the `Fabric Mods` should always be on top, as the ability to change stuff using custom resource packs is important.
It might be a good idea though to let mods toggle this functionality in development environments for testing reasons.

apple502j:
Pinging @jacobsjo

PR #4981, opened 2025-11-05
Fully move resource loading to Resource Loader v1.

### LINKED ISSUE #3656: Going to resource packs then clicking done causes resource reload

### LINKED ISSUE #3795: "Fabric Mods (mod resources)" in resource pack folder is put above your resource packs, breaking certain resource packs
"Fabric Mods (mod resources)" in the resource pack folder is put above resource packs when loading into Fabric with a resource pack already on. You have to have been playing in a version that is not Fabric (Vanilla/OptiFine/Forge/etc) beforehand for it to be put above the resource packs. "Fabric Mods" being put above resource packs breaks a ton of resource pack things (custom item textures, ctm, and much more), and thus should never happen by default.

### LINKED ISSUE #4196: You can disable the `fabric (Fabric mod)` resource pack
You can disable the `fabric (Fabric mod) resource pack`, whilst you cannot with the `fabric-convention-tags-v2 (Fabric mod)` resource pack. I would suggest that the intended behaviour would be that the `fabric (Fabric mod) resource pack` cannot be disabled. This has been tested with Fabric API 0.107.0 with Minecraft 1.21.3.

LambdAurora:
This closes #4196, #3782, #3656, and potentially should close #3795.

The Fabric Mods resource pack is no more, as it was redundant and mostly confusing for users.
It also allows to fully resolve #3656 on a new instance.

This step will allow further work to expand Resource Loader v1 functionality (virtual resource packs, etc).

Also provides new tooltips especially for built-in resource packs whose name and description overflow:
image


zOnlyKroks (approved):
lgtm

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/mixin/resource/v1/client/OptionsMixin.java
```diff
@@ -78,7 +77,7 @@ private void onLoad(CallbackInfo ci) {
 		}
 
 		Path trackerFile = dataDir.resolve("fabricDefaultResourcePacks.dat");
-		Set<String> trackedPacks = new HashSet<>();
+		var trackedPacks = new HashSet<String>();
```
maityyy: Is `var` really necessary? It seems unnecessary

LambdAurora: Initially thought I was going to rewrite more stuff but ended up moving a lot of stuff around.

deprecated/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/api/resource/ResourcePackActivationType.java
```diff
@@ -16,22 +16,31 @@
...
 	/**
 	 * Enabled by default. The user has still full control over the activation of the resource pack.
 	 */
-	DEFAULT_ENABLED,
+	DEFAULT_ENABLED(PackActivationType.DEFAULT_ENABLED),
 	/**
 	 * Always enabled. The user cannot disable the resource pack.
 	 */
-	ALWAYS_ENABLED;
+	ALWAYS_ENABLED(PackActivationType.ALWAYS_ENABLED),;
```
Juuxel: Nit: the `,;` formatting is a bit strange. I don't think FAPI does it elsewhere.

deprecated/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.resources.Identifier;
+import net.minecraft.server.packs.PackType;
+import net.minecraft.server.packs.resources.PreparableReloadListener;
+
+import net.fabricmc.fabric.api.resource.IdentifiableResourceReloadListener;
+import net.fabricmc.fabric.api.resource.ResourceManagerHelper;
+import net.fabricmc.fabric.api.resource.v1.ResourceLoader;
+
+public class ResourceManagerHelperImpl implements ResourceManagerHelper {
+	private static final Map<PackType, ResourceManagerHelperImpl> registryMap = new HashMap<>();
```
Juuxel, line 35: Nit: should be `REGISTRY_MAP` according to FAPI naming conventions

fabric-resource-loader-v1/build.gradle
```diff
@@ -12,3 +12,47 @@ testDependencies(project, [
...
+	}
+
+	tasks.register("${sourceSet.name}Jar", Jar) {
+		from sourceSet.output
+		archiveBaseName.set(sourceSet.name)
+	}
+}
+
+rootProject.allprojects.each { p ->
+	if (p.extensions.findByName("loom") == null) {
```
Juuxel, line 38: Note: This will work for now, but it'll break if/when FAPI enables project isolation or configure on demand.

I'm not saying this PR should address it, but there probably should just be a list of meta project paths somewhere that projects can access.

fabric-resource-loader-v1/build.gradle
```diff
@@ -12,3 +12,47 @@ testDependencies(project, [
...
+		return // Skip over the meta projects
+	}
+
+	p.loom.mods.register("fabric-resource-loader-v1-testmod-a") {
```
Juuxel, line 42: Why are you registering these to all projects?

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/pack/FabricPack.java
```diff
@@ -23,11 +23,11 @@
  * Fabric addition to ResourcePackProfile.
  * @see ModResourcePackCreator
  */
-public interface FabricResourcePackProfile {
+public interface FabricPack {
 	/**
 	 * Returns whether the pack is internal and hidden from end users.
 	 */
-	default boolean fabric_isHidden() {
+	default boolean fabric$isHidden() {
```
Juuxel, line 30: Nit: No need to rename these IMO

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/server/LanguageMixin.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package net.fabricmc.fabric.mixin.resource.loader.server;
+package net.fabricmc.fabric.mixin.resource.v1.server;
```
Juuxel, line 17: FAPI doesn't generally use versions in impl and mixin packages. From the conventions:
> For API only: add the module major version with a v prefix, for example `.v1`

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/ResourceLoaderImpl.java
```diff
@@ -278,4 +292,89 @@ private void sort(List<PreparableReloadListener> reloaders) {
...
+	 * @param subPath        the sub path in the mod resources
+	 * @param container      the mod container
+	 * @param displayName    the display name of the resource pack
+	 * @param activationType the activation type of the resource pack
+	 * @return {@code true} if successfully registered the resource pack, or {@code false} otherwise
+	 * @see ResourceLoader#registerBuiltinPack(Identifier, ModContainer, Component, PackActivationType)
+	 * @see ResourceLoader#registerBuiltinPack(Identifier, ModContainer, PackActivationType)
+	 */
+	public static boolean registerBuiltinPack(Identifier id, String subPath, ModContainer container, Component displayName, PackActivationType activationType) {
+		// Assuming the mod has multiple paths, we simply "hope" that the  file separator is *not* different across them
```
Juuxel:
```suggestion
		// Assuming the mod has multiple paths, we simply "hope" that the file separator is *not* different across them
```

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/mixin/resource/v1/client/TransferableSelectionListPackEntryMixin.java
```diff
@@ -0,0 +1,95 @@
...
+import net.minecraft.client.gui.components.MultiLineTextWidget;
+import net.minecraft.client.gui.components.StringWidget;
+import net.minecraft.client.gui.screens.packs.TransferableSelectionList;
+import net.minecraft.network.chat.Component;
+import net.minecraft.util.FormattedCharSequence;
+
+import net.fabricmc.fabric.impl.resource.v1.client.PackTooltipComponent;
+
+@Mixin(TransferableSelectionList.PackEntry.class)
+public class TransferableSelectionListPackEntryMixin {
```
Juuxel:
```suggestion
class TransferableSelectionListPackEntryMixin {
```
(see conventions; also applies to the other new mixins)

Also, this mixin should include a comment about what it does.

fabric-resource-loader-v1/src/client/java/net/fabricmc/fabric/impl/resource/v1/client/PackTooltipComponent.java
```diff
@@ -0,0 +1,90 @@
...
+
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.gui.Font;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.screens.inventory.tooltip.ClientTooltipComponent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.util.FormattedCharSequence;
+import net.minecraft.world.inventory.tooltip.TooltipComponent;
+
+public record PackTooltipComponent(Optional<Component> name, Optional<List<FormattedCharSequence>> description)
```
Juuxel, line 30: I wonder if nullables would be better (we definitely use them more than Optional in FAPI) 🤔 I doubt it matters much though.

Juuxel (changes requested):
The API looks to be about the same and the impl changes look good to me for the most part. I also think promoting `FabricResource` to the API package is a good idea 👍 

---

There are many formatting changes that aren't related to the actual code changes in the PR, including changing local variables to `var`, adding `this.` spam, changing `_` to `$` and reformatting long lines.

I don't think these formatting changes are really necessary. They also clutter the diff, making it harder to read as they somewhat hide the relevant changes. Since the code itself hasn't changed at all in most of these cases, it'd be fine to leave these as they were, which would preserve the existing git blame too.

deprecated/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
LambdAurora: Definitely a left over, will address this.

Juuxel: Ah, it's a bit hard to tell which files are new and which files are just heavily modified since git always loses track of them during refactors 😄

LambdAurora:
> There are many formatting changes that aren't related to the actual code changes in the PR, including changing local variables to var, adding this. spam, changing _ to $ and reformatting long lines.

I agree that it makes this harder to check for and I wonder if I should've addressed this in a different PR.
I know some would not agree when it comes to the usage of `this` but I have seen way too often variables whose sources was not clear when reading, even with an IDE.

deprecated/fabric-resource-loader-v0/src/main/java/net/fabricmc/fabric/impl/resource/loader/ResourceManagerHelperImpl.java
LambdAurora: I'm not actually sure whether to actually address this one now that I think about it, didn't see it was Resource Loader v0, I don't think refactoring what's bound to be drop once the game goes full deobf matters.

fabric-resource-loader-v1/build.gradle
LambdAurora: This is a 1:1 copy of what was done in Resource Loader v0 buildscript, [it also seems like this is executed for every "mod" and testmods in the root build.gradle](https://github.com/FabricMC/fabric/blob/be66d255e05833fabed4604fd8e43a49df374079/build.gradle#L206).

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/mixin/resource/v1/server/LanguageMixin.java
LambdAurora: I really wish this was picked up in the initial resource loader v1 PR :sweat_smile: 
That's going to be a pain for the diff.

fabric-resource-loader-v1/build.gradle
Juuxel: It's there in the root one since the root project has all test mods, but I don't think modules depend on other modules' test mods otherwise 🤔

LambdAurora: It is an `allprojects` in an `allprojects` block so, yeah, idk.

Juuxel: Oh, I didn't notice that it was nested. I suppose it doesn't hurt to have since any paths not on the classpath are just ignored anyway.

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/impl/resource/v1/pack/FabricPack.java
cassiancc: The `fabric_` prefix is preferred for Fabric API as it allows double clicking the method to edit it.

LambdAurora: Ok, will revert then, I was under the impression it wasn't the current standard woops.

modmuss50 (approved):
Do we want to try and get this in for 1.21.11, I think it would be nice to!

The merge conflicts should be easy to fix, let me know if you want me to do it.

I havent done an indepth review of this due to the size, its clearly an improvement. If there is anything specific you would like to me at in more detail let me know.

Juuxel (approved):
I'm not the biggest expert in this area, but the API looks about the same to me which is nice for moving from v0 to v1. I still have some concerns about the formatting changes, but splitting them out of this PR at this point would be a lot of work for no gain (I assume they'd be PR'd in anyway).

PR #4988, opened 2025-11-08
Added on itemPickup event

yajatkaul:
I was working on a mod recently when i came across this and realized there is no event for it, so i thought i might as well try to contribute it in there for others to use :)

### DIFF
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
@@ -18,7 +18,9 @@ package net.fabricmc.fabric.api.event.lifecycle.v1;
 
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.ItemEntity;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ItemStack;
 import net.minecraft.server.world.ServerWorld;
 
@@ -63,6 +65,17 @@ public final class ServerEntityEvents {
 		}
 	});
 
+	/**
+	 * Called during {@link ItemEntity#tick()} if an entity tries to pick up any ItemEntity.
+	 *
+	 * <p>Picking up of an item is determined by {@link ItemEntity#onPlayerCollision(PlayerEntity)}.
+	 */
+	public static final Event<ItemPickup> ITEM_PICKUP = EventFactory.createArrayBacked(ServerEntityEvents.ItemPickup.class, callbacks -> (playerEntity, itemEntity, itemStack) -> {
+		for (ItemPickup callback : callbacks) {
+			callback.onPickup(playerEntity, itemEntity, itemStack);
+		}
+	});
+
 	@FunctionalInterface
 	public interface Load {
 		void onLoad(Entity entity, ServerWorld world);
@@ -76,5 +89,10 @@ public final class ServerEntityEvents {
 	@FunctionalInterface
 	public interface EquipmentChange {
 		void onChange(LivingEntity livingEntity, EquipmentSlot equipmentSlot, ItemStack previousStack, ItemStack currentStack);
+	}
+
+	@FunctionalInterface
+	public interface ItemPickup {
+		void onPickup(PlayerEntity playerEntity, ItemEntity itemEntity, ItemStack itemStack);
 	}
 }
--- a/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
+++ b/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java
@@ -70,11 +70,20 @@ public final class ServerEntityEvents {
 	 *
 	 * <p>Picking up of an item is determined by {@link ItemEntity#onPlayerCollision(PlayerEntity)}.
 	 */
-	public static final Event<ItemPickup> ITEM_PICKUP = EventFactory.createArrayBacked(ServerEntityEvents.ItemPickup.class, callbacks -> (playerEntity, itemEntity, itemStack) -> {
-		for (ItemPickup callback : callbacks) {
-			callback.onPickup(playerEntity, itemEntity, itemStack);
-		}
-	});
+	public static final Event<ItemPickup> ITEM_PICKUP = EventFactory.createArrayBacked(
+			ItemPickup.class,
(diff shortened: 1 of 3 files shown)


Fuzss:
Hey, thanks for your contribution!
Could you please outline some use cases?

I'm honestly not quite sure what this event is for in the current form. 😅
While NeoForge has [something similar](https://github.com/neoforged/NeoForge/blob/8a596374f4eee9de53d70c2c6051100d33b0738a/src/main/java/net/neoforged/neoforge/event/entity/player/ItemEntityPickupEvent.java#L56) which I've been using myself, that does allow for two things:
- preventing the item pickup (useful for backpack mods, so they can add to their own inventory and skip vanilla afterward)
- forcing the pickup regardless of pick up delay or the item "owner" (not sure if we need to account for that, use cases seem pretty niche)

It would be nice to allow your implementation to be used at least in the first scenario outlined above.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ItemEntityMixin.java
```diff
@@ -0,0 +1,39 @@
...
+
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerEntityEvents;
+
+@Mixin(ItemEntity.class)
+public class ItemEntityMixin {
+	@Inject(method = "onPlayerCollision", at = @At("HEAD"))
+	private void onPlayerPickup(PlayerEntity playerEntity, CallbackInfo ci) {
+		ItemEntity itemEntity = (ItemEntity) (Object) this;
+
+		if (!itemEntity.getWorld().isClient && !itemEntity.cannotPickup()) {
```
Fuzss: This must also check the "owner" uuid (used by `/give`).

yajatkaul:
Okay so basically i came across this while i was making a tasks mod so to keep a track of when the player gets and item and stuff, there is a bit more to it but yea that's why i needed this, 
And yes since u mentioned the cancellable thing in your first point, sounds like a good idea and i have hence added that.

Fuzss:
Great, thanks. 

Please also address the review. Ideally to make sure no vanilla checks are skipped, moving the injection point to where the item is added to the player inventory (`Inventory::add`) might be ideal. That will also allow to minimize the injected code.

yajatkaul:
I'm not sure about the `Inventory::add` i think then i cant make the event cancellable, or maybe i just didn't get to the right method.

Fuzss:
Don't seem to be able to add any code suggestions as part of the review, so here are some snippets for changes:

This is how the Mixin should look:
```java
	@Inject(method = "onPlayerCollision", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerInventory;insertStack(Lnet/minecraft/item/ItemStack;)Z"), cancellable = true)
	private void onPlayerPickup(PlayerEntity playerEntity, CallbackInfo ci) {
		ItemEntity itemEntity = (ItemEntity) (Object) this;
		boolean allowPickup = ServerEntityEvents.ITEM_PICKUP.invoker().onPickup(playerEntity, itemEntity, itemEntity.getStack());

		if (!allowPickup) {
			ci.cancel();
		}
	}
```

Also Javadoc could be slightly expanded:
```java
	/**
	 * Called during {@link ItemEntity#tick()} on the logical server if an entity tries to pick up any {@link ItemEntity}.
	 *
	 * <p>Picking up of an item is determined by {@link ItemEntity#onPlayerCollision(PlayerEntity)}.
	 *
	 * <p>Returning {@code false} prevents vanilla from handling the pick-up.
	 */
	public static final Event<ItemPickup> ITEM_PICKUP = EventFactory.createArrayBacked(<...>
```

Fuzss (approved):

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ItemEntityMixin.java
yajatkaul: Do u mean
```java
UUID owner = itemEntity.getOwner();
    if (owner != null && !playerEntity.getUuid().equals(owner)) {
        // If the item has an owner and this player isnt it, block pickup
        ci.cancel();
        return;
    }
 ```

Comment (Fuzss) [reply fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ItemEntityMixin.java]: This is outdated, you can mark it as resolved.

File: fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/event/lifecycle/v1/ServerEntityEvents.java [hunk 2]
```diff
@@ -63,6 +65,28 @@ private ServerEntityEvents() {
...
 	});
 
+	/**
+	 * Called during {@link ItemEntity#tick()} on the logical server if an entity tries to pick up any {@link ItemEntity}.
+	 *
+	 * <p>Picking up of an item is determined by {@link ItemEntity#onPlayerCollision(PlayerEntity)}.
+	 *
+	 * <p>Returning {@code false} prevents vanilla from handling the pick-up.
+	 */
+	public static final Event<ItemPickup> ITEM_PICKUP = EventFactory.createArrayBacked(
```
maityyy, line 75: I have a strong belief that this should be in `ServerLivingEntityEvents` in the `fabric-entity-events-v1` module.

fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/ItemEntityMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.entity.ItemEntity;
+import net.minecraft.entity.player.PlayerEntity;
+
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerEntityEvents;
+
+@Mixin(ItemEntity.class)
+public class ItemEntityMixin {
+	@Inject(method = "onPlayerCollision", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerInventory;insertStack(Lnet/minecraft/item/ItemStack;)Z"), cancellable = true)
```
maityyy, line 31: (mojmap)

I think we need a more general solution that supports not only player but also mobs. I.e. the event should accept `LivingEntity` instead of `Player` and be invoked also in `Mob::pickUpItem` or before `Mob::pickUpItem` (`Mob::aiStep` and `ObtainRaidLeaderBannerGoal::tick`).

edit: But the problem is that `Mob::pickupItem` has overrides and can also be called by other mods (but can't remember if this is a problem for FAPI tbh)

maityyy (changes requested):

PR #884, opened 2020-07-11
Compress textures and mod icons

haykam821:
This pull request compresses every image in the repository, bringing its overall size down from 30 MB to 29 MB. The Fabric icon, which is duplicated for each individual module, changed from 2 KB to 227 B (an 86% decrease).

PR #4989, opened 2025-11-08
Visually losslessly optimize compression of PNG resources

Luracasmus:
See https://github.com/FabricMC/fabric-loader/pull/1071. Essentially a new attempt at https://github.com/FabricMC/fabric/pull/884.

Using oxipng 9.1.5 with `-Z -o max --fast --nc -a`. Shrinks the repo by 29,432 B and the built jar by 1,368 B (before and after built locally).

### Testing
The game runs as expected with no obvious issues.

There is no Mod Menu version for 25w45a, but I also can't seem to build any branch other than 1.21.11 of Fabric API my system. I instead extracted the 0.138.0+1.21.10 version from Modrinth and all contained jars at `/META-INF/jars/`, ran the same command as above on everything and reconstructed the jar. The first of the images below is taken with the modified jar.

bild

bild

Visually comparing all of the images in GitHub's diff viewer I also noticed no differences.

I had to remove lines 58-74 in `fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java` and increase Gradle's Xmx to get `gradlew build` to succeed on the 1.21.11 branch, since the `.javadoc` task would otherwise fail with
```
.../fabric/fabric-registry-sync-v0/src/main/java/net/fabricmc/fabric/api/event/registry/DynamicRegistries.java:54: error: reference not found: RegistryKey
 * {@snippet :
   ^
```
or Gradle would run out of heap space.

`gradlew check` succeeds (ran before the aforementioned change) and produces [problems-report.html](https://github.com/user-attachments/files/23432710/problems-report.html).


modmuss50 (approved):
Sure, why not. Thanks for testing.

Luracasmus:
I also tested it with the new Mod Menu on 25w45a now and it seems fine

Issue #4669, opened 2025-05-31
Oversized data attachments cause server crash when synced


jacobsjo:
When a data attachment is set to sync with clients and has a encoded size > 32502 bytes, the server crashes when tying to send the attachment.

```
---- Minecraft Crash Report ----
// Why did you do that?
Time: 2025-05-31 14:26:26
Description: Exception chunk generation/loading
java.lang.IllegalArgumentException: Data for attachment 'worldgendevtools:feature_positions' was too big (32768 bytes, over maximum 32502)
```
(crash report from downstream bug report: https://github.com/jacobsjo/worldgen-devtools/issues/16 ; I'll update it when I reproduce this myself)

How to handle these oversized data attachments is a different question, but crashing the entire server is clearly not the desired outcome.

CallMeEchoCodes:
It would probably be possible to switch the payload to use a split payload if anyone has time

DennisOchulor:
I'd be willing to make a PR for this, though I'm wondering what would be a suitable new max size for individual attachments and the att sync payload.

modmuss50:
I wonder if the proper solution would be for mods to say how big they want their attachment to be?

DennisOchulor:
Possibly, though a fixed limit would still be needed for the att sync packet itself. I was hoping that by setting a high enough limit, manual packet splitting for initial syncs could just be removed, but if mods can set their own limits per attachment then manual splitting might still be needed.

CallMeEchoCodes:
Other parts of FAPI seem to use system properties to specify the size limit, that could work here

Earthcomputer:
I think having a per-attachment-type customizable limit that mods can set is the best solution

jacobsjo:
> I think having a per-attachment-type customizable limit that mods can set is the best solution

It might solve some issues, but not the one I've reported in this issue. This issue is about the very harsh handling of oversized packets, which can happen no matter what the limit is set to.

Earthcomputer:
If it's sending packets to the client I don't see a problem with the server crashing. If it's crashing when receiving stuff from the client then that's very bad.
However we should probably put a try catch around it and kick the client on an exception

PR #4996, opened 2025-11-11
Add per-AttachmentType sync size limits & Allow modifying large payload max size

DennisOchulor:
This PR changes the att sync payload to be splittable and allows mods to set sync size limits per AttachementType.

1. The att sync payload max size is set to Integer.MAX_VALUE. This might be too high but I set it to that as a start.
2. Initial syncs still manually split packets if needed, just in case multiple really large attachments go over the max payload size, however unlikely that is.
3. The new default sync limit is 1 MiB, up from just 32502 bytes.
4. The encoded attachement data is now truncated to the actual readable number of bytes, but unfortunately this requires copying the array, so I'm not really sure if this is the way to go.

Relates to #4669


Earthcomputer (changes requested):
Increasing the max attachment size to `Integer.MAX_VALUE` is definitely not the right solution. The way I would personally solve it is like this:
1. Make every attachment sync in a separate packet. I believe the netty pipeline batches packets automatically, so I don't think the original sync mechanism joining multiple updates into one packet really saved on much.
2. When sending the packets, put them in a bundle to make sure they are all received on the same tick.
3. Add an API to fabric networking to update the maximum payload size of an already-registered large packet.
4. Increase the default max attachment size to the maximum size of the payload in a large packet before it has to split, minus whatever the maximum length of the ID is.
5. Whenever a new attachment is registered with a max size bigger than the current max size for any registered attachment, update the max size of the sync packet.

DennisOchulor:
I assume number 3 would be something like `PayloadTypeRegistry.modifyLargePayloadMaxSize(id,size)`? Though depending on mod load order, mods registering large attachments may cause `modifyLargePayloadMaxSize()` to be called before the att sync packet is even registered.

DennisOchulor:
I've implemented @Earthcomputer's suggestions, the changes to the networking API are in a separate commit to help with reviewing. Though maybe a separate PR should be made?

DennisOchulor:
See #5119 

The data attachment changes will be made in another PR after #5119 is merged.

PR #5012, opened 2025-11-19
[Issue] - Fix HudElementRegistry.replaceElement() for subtitles during gameplay.

### LINKED ISSUE #4933: `HudElementRegistry#replaceElement` does not work properly to Closed Captions in MC 1.21.9/10
**Minecraft Version:** 1.21.9 & 1.21.10

**Fabric Loader Version:** 0.17.3

**Fabric API Version:** 0.134.0+1.21.9 & 0.134.1+1.21.10

Description of issue:

I wrapped the `VanillaHudElements` to adjust their scale and opacity. It works for almost all of them.

When it comes to `SUBTITLES` (aka *Closed Caption*), however, the wrapper does not work at all if the client is running.

ONLY when I press ESC to pause, it works, even in a multiplayer server.

I think it should be easy to reproduce by using my mod [HUD Manager](https://modrinth.com/mod/hud-manager).

Liam-Broome:
Fixes  [issue #4933](https://github.com/FabricMC/fabric/issues/4933) where `HudElementRegistry.replaceElement(VanillaHudElements.SUBTITLES, ...)` did not work correctly during normal gameplay, only when the game was paused. This prevented mods from properly scaling, positioning, or otherwise modifying subtitle rendering during active play.

https://github.com/user-attachments/assets/501116a1-6d37-4a04-8476-71be2870bd8e


Liam-Broome:
The test `build / client_test` failing seems to be an infrastructural issue. I can see that the PR previous to this also has this issue.

maityyy:
> The test `build / client_test` failing seems to be an infrastructural issue. I can see that the PR previous to this also has this issue.

yes, MC-304369

modmuss50 (approved):

ekulxam:
Wait don't merge yet

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiMixin.java
```diff
@@ -161,7 +164,19 @@ private void wrapPlayerList(Gui instance, GuiGraphics context, DeltaTracker tick
 	}
 
 	@WrapOperation(method = "render", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/Gui;renderSubtitleOverlay(Lnet/minecraft/client/gui/GuiGraphics;Z)V"))
```
ekulxam: I disagree with this mixin, see https://discord.com/channels/507304429255393322/807617700734042122/1440940666829013103. 
> I think it would better to instead mixin into the renderSubtitleOverlay method (and the lambda) to wrap the subtitleOverlay#render (as shown in the image).
As stated by @cassiancc, we can obtain the DeltaTracker/RenderTickCounter from the client so my suggested change may actually be better.

ekulxam (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/GuiMixin.java
Liam-Broome: @ekulxam - I have made some changes, thanks for the feedback 😄

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/SubtitleOverlayMixin.java
```diff
@@ -0,0 +1,62 @@
...
+							fabric_renderingThroughHud = false;
+						}
+					});
+		}
+	}
+
+	@Unique
+	private void fabric_callOriginalRender(GuiGraphics context) {
+		((SubtitleOverlay) (Object) this).render(context);
+	}
```
cputnam-a11y: why does this need to be recursive? I would think that this should be a WrapMethod.

cputnam-a11y (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/SubtitleOverlayMixin.java
ekulxam: I said to wrap it and not inject. It should actually be a WrapOperation into both the lambda and the renderSubtitleOverlay method with the render call as the target.

ekulxam:
@Liam-Broome are you still working on this? If not, then I'd like to implement the requested changes myself. No worries if you're busy.

Liam-Broome:
@ekulxam - Feel free to take over 😀

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/mixin/client/rendering/SubtitleOverlayMixin.java
ekulxam: You're right actually; WrapMethod is cleaner here and does accomplish what we wanted.

ekulxam:
@Liam-Broome I've made a pull request to your fork https://github.com/Liam-Broome/fabric/pull/1

cputnam-a11y (approved):
I still don't really understand the actual issue here, but the mixins do look much cleaner now

PR #5027, opened 2025-12-05
feat: `ServerMobEffectEvents` (Status Effect Events)

sylv256:
## Description
Users have expressed a desire for a status effect (`MobEffect`) API in order to reduce Mixin usage and simplify status effect code, especially in the case of multiple types of effects or removal of effects. This Pull Request adds `ServerMobEffectEvents` as well as an interface `FabricMobEffectEventHandlers` injected into `MobEffect` for overriding the `BEFORE_ADD` and `BEFORE_REMOVE` events in classes.

## To-do
 - [x] Game Tests
 - [x] More Detailed PR Description
 - [x] Proper Documentation
 - [x] Injected Interface on `MobEffect` for `onEffectRemoved`
 - [x] Description of the rest of the tests

## Testing
In compliance with the `CONTRIBUTING.md`, the following is a rough description of how the Game Tests work. All tests assume an event is registered and that a Salmon is holding a Potato.

### `ALLOW_ADD`
A Salmon with regeneration is summoned, and the registered `ALLOW_ADD` event disallows Salmon holding potatoes from having regeneration.

### `BEFORE_ADD` and `AFTER_ADD`
A Salmon with absorption is summoned. The registered `BEFORE_ADD` event forbids the Salmon having absorption, and the `AFTER_ADD` event requires the Salmon have absorption.

### `ALLOW_EARLY_REMOVE`
A Salmon and a Player with blindness are summoned. The registered `ALLOW_EARLY_REMOVE` event cancels the removal of blindness and poison effects. The Player is then given poison and drinks a honey bottle, but the test asserts that honey bottles do not remove poison.

### `BEFORE_REMOVE` and `AFTER_REMOVE`
A Salmon with saturation is summoned. The registered `BEFORE_REMOVE` event requires the Salmon have saturation, and the `AFTER_REMOVE` event forbids the Salmon having saturation.


fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/RemoveStatusEffectsConsumeEffectMixin.java
```diff
@@ -0,0 +1,47 @@
...
+
+	@WrapOperation(
+			method = "apply",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/world/entity/LivingEntity;removeEffect(Lnet/minecraft/core/Holder;)Z"
+			)
+	)
+	private boolean beforeRemoveEffect(LivingEntity instance, Holder<MobEffect> holder, Operation<Boolean> original) {
+		ServerMobEffectEvents.ALLOW_EARLY_REMOVE.invoker().allowEarlyRemove(instance.getEffect(holder), instance);
```
sylv256: whoops, this needs to return false if the events return false! perhaps this should be in the game test? I only didn't add this since I didn't want to overcomplicate things.

I should probably have a fake player drink milk or something similar to test this.

sylv256:
So, I've discovered an issue in the current implementation that severely over-complicates this, and that is the `/effect` command calls `LivingEntity#addEffect` and `LivingEntity#removeEffect` methods.

 - **So what?** Well, if I had to guess, blocking commands from modifying status effects is likely undesirable behavior.
 - **So just don't call those methods in the `EffectCommands` class. Use our own methods.** This is a possible solution, but this would require lengthy discussion and may be undesirable for many. This could also break compatibility for mod-defined commands that modify effects.
 - **Why not split the events?** For a reasonable implementation, this would require special-casing commands by requiring them to call another set of methods. This also has the same downsides as using our own methods in commands.
 - **Stackwalker?** No.

cputnam-a11y:
why not use threadlocals to pass context down to the remove methods?

ChrysanthCow:
I agree with having a context in the event, it'd be similar to the enchantment events.

sylv256:
> why not use threadlocals to pass context down to the remove methods?

That has the same issue as changing which methods are called. I don't want to break compatibility with prior mods, at all, and I also don't want to make it a headache for new mods. If more people say it's really not an issue though, I guess we could go ahead with that.

cputnam-a11y:
> > why not use threadlocals to pass context down to the remove methods?
> 
> That has the same issue as changing which methods are called. I don't want to break compatibility with prior mods, at all, and I also don't want to make it a headache for new mods. If more people say it's really not an issue though, I guess we could go ahead with that.

I'm not sure I understand how adding context would introduce incompatiblity. wrapping the calls in threadlocal context allows you to attribute attribute changes within the scope to the root cause regardless of whether mods directly call the vanilla method, or would have replaced a fapi method

sylv256:
> > > why not use threadlocals to pass context down to the remove methods?
> > 
> > 
> > That has the same issue as changing which methods are called. I don't want to break compatibility with prior mods, at all, and I also don't want to make it a headache for new mods. If more people say it's really not an issue though, I guess we could go ahead with that.
> 
> I'm not sure I understand how adding context would introduce incompatiblity. wrapping the calls in threadlocal context allows you to attribute attribute changes within the scope to the root cause regardless of whether mods directly call the vanilla method, or would have replaced a fapi method

It means that, by default, all existing mod-defined commands that modify status effects will be affected by allow events in `ServerMobEffectEvents`.

cputnam-a11y:
> > > > why not use threadlocals to pass context down to the remove methods?
> > > 
> > > 
> > > That has the same issue as changing which methods are called. I don't want to break compatibility with prior mods, at all, and I also don't want to make it a headache for new mods. If more people say it's really not an issue though, I guess we could go ahead with that.
> > 
> > 
> > I'm not sure I understand how adding context would introduce incompatiblity. wrapping the calls in threadlocal context allows you to attribute attribute changes within the scope to the root cause regardless of whether mods directly call the vanilla method, or would have replaced a fapi method
> 
> It means that, by default, all existing mod-defined commands that modify status effects will be affected by allow events in `ServerMobEffectEvents`.

correct. all interactions with mob effects including commands should trigger the events imo, even existing ones, it would just be up to us to provide a way for mods to categorize their own interactions

sylv256:
> > > > > why not use threadlocals to pass context down to the remove methods?
> > > > 
> > > > 
> > > > That has the same issue as changing which methods are called. I don't want to break compatibility with prior mods, at all, and I also don't want to make it a headache for new mods. If more people say it's really not an issue though, I guess we could go ahead with that.
> > > 
> > > 
> > > I'm not sure I understand how adding context would introduce incompatiblity. wrapping the calls in threadlocal context allows you to attribute attribute changes within the scope to the root cause regardless of whether mods directly call the vanilla method, or would have replaced a fapi method
> > 
> > 
> > It means that, by default, all existing mod-defined commands that modify status effects will be affected by allow events in `ServerMobEffectEvents`.
> 
> correct. all interactions with mob effects including commands should trigger the events imo, even existing ones, it would just be up to us to provide a way for mods to categorize their own interactions

That's fair. I'm primarily worried existing commands will be heavily affected by this, especially because they would have to call a special new method defined by FAPI in order for it to identify that it is a command.

sylv256:
I'm going to do a quick refactor of the package and make a new `Context` type for the allow events.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
```diff
@@ -0,0 +1,211 @@
...
+	 * only after an effect is applied such as a method that
+	 * checks if an effect is present on an entity.
+	 */
+	public static final Event<AfterAdd> AFTER_ADD = EventFactory.createArrayBacked(AfterAdd.class, callbacks -> (effectInstance, entity) -> {
+		for (AfterAdd callback : callbacks) {
+			callback.afterAdd(effectInstance, entity);
+		}
+	});
+
+	public static final Event<AllowEarlyRemove> ALLOW_EARLY_REMOVE = EventFactory.createArrayBacked(AllowEarlyRemove.class, callbacks -> (effectInstance, entity, isFromCommand) -> {
```
CallMeEchoCodes: When is "early"? The name is probably fine but it needs documentation at least.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,167 @@
...
+		if (!((LivingEntity) (Object) this).level().isClientSide()) {
+			ServerMobEffectEvents.AFTER_ADD.invoker().afterAdd(effectInstance, (LivingEntity) (Object) this);
+		}
+	}
+
+	@WrapOperation(
+			method = "removeAllEffects",
+			at = @At(
+					value = "INVOKE",
+					target = "Ljava/util/Map;clear()V"
```
CallMeEchoCodes, line 111: This target is not obfuscated, I'm fairly sure the `@At` should have `remap = false` because of that. There are a few other targets like this in this file.

CallMeEchoCodes: sorry for the ping @at you did this to yourself

CallMeEchoCodes (approved):
A few minor things but overall LGTM

fabric-entity-events-v1/src/testmod/java/net/fabricmc/fabric/test/entity/event/gametest/ServerMobEffectsGameTest.java
```diff
@@ -0,0 +1,164 @@
...
+			boolean cannotRemove = effectInstance.is(MobEffects.BLINDNESS) || effectInstance.is(MobEffects.POISON);
+			return !(cannotRemove && isThisTheEntity);
+		});
+
+		// Regular Salmon testing
+		Salmon theSalmon = summonTheSalmon(context);
+		theSalmon.addEffect(createEffect(MobEffects.BLINDNESS));
+		context.assertTrue(theSalmon.hasEffect(MobEffects.BLINDNESS), "The Salmon must have blindness");
+
+		// Player milk testing
```
Ellie-McQuinn, line 91: :flushed:

sylv256:
Refactors are finished, and we have context now.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
```diff
@@ -0,0 +1,222 @@
...
+	 * checks if an effect is present on an entity.
+	 */
+	public static final Event<AfterAdd> AFTER_ADD = EventFactory.createArrayBacked(AfterAdd.class, callbacks -> (effectInstance, entity) -> {
+		for (AfterAdd callback : callbacks) {
+			callback.afterAdd(effectInstance, entity);
+		}
+	});
+
+	/**
+	 * An event checks whether an effect may be removed.
```
sylv256: typo 😭 

im too tired i'll fix in the morning

sylv256: also make sure to give an example of what triggers this, like drinking milk, drinking a bottle of honey, consuming a totem, or using a command such as `/effect clear`.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
Juuxel: AFAIK `remap = false` [does nothing](https://github.com/FabricMC/tiny-remapper/commit/0bbcbe4a6ad40eb99c0e6a60f4f21890da4817c2) when mixin refmaps are disabled, like they are in Fabric API.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/FabricMobEffect.java
```diff
@@ -0,0 +1,44 @@
...
+	default void onEffectAdded(MobEffectInstance effectInstance, LivingEntity entity) {
+	}
+
+	/**
+	 * Called before an {@linkplain MobEffectInstance instance of this effect} has been removed from a {@linkplain LivingEntity living entity}.
+	 *
+	 * @param effectInstance an instance of this effect
+	 * @param entity the entity the effect instance is being removed from
+	 */
+	default void onEffectRemoved(MobEffectInstance effectInstance, LivingEntity entity) {
```
modmuss50, line 51: Are these needed if we have the events? Im not tottaly aginst having it but I would like to understand the usecase.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/EffectEventContext.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package net.fabricmc.fabric.api.entity.event.v1.effect;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Context for {@linkplain ServerMobEffectEvents mob effect events}.
+ */
+public interface EffectEventContext {
```
modmuss50, line 26: @ApiStatus.NonExtendable

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
```diff
@@ -0,0 +1,236 @@
...
+ * Events related to {@linkplain MobEffect status effects} in mobs.
+ *
+ * <p>Status effect events are useful when implementing generic behavior for
+ * many status effects and modifying the addition and removal behavior of
+ * existing status effects.
+ *
+ * <p>When only one class of {@linkplain MobEffect status effects}
+ * requires code to be run before the addition or removal of the effect,
+ * consider using {@link MobEffect#onEffectAdded(MobEffectInstance, LivingEntity)}
+ * or {@link MobEffect#onEffectRemoved(MobEffectInstance, LivingEntity)}.
```
modmuss50: An alternative option would be to have a static method that returns the Event for a given MobEffect in a similar manner to how we did for game rules:

https://github.com/FabricMC/fabric/blob/dafb9388c88ff0a4ab7a6a8d88f42e18d298959c/fabric-game-rule-api-v1/src/main/java/net/fabricmc/fabric/api/gamerule/v1/GameRuleEvents.java#L32-L51

This is just an idea, it really depends if there is a common use case to be notified about all effects, if the main use case only cares about a specific effect maybe this would be a better way to go, and removes the need for the EventHandler interface.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/impl/entity/event/effect/FinalThreadLocal.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package net.fabricmc.fabric.impl.entity.event.effect;
+
+import java.util.function.Supplier;
+
+/**
+ * {@link ThreadLocal} but you cannot modify it.
+ */
+public final class FinalThreadLocal<T> extends ThreadLocal<T> {
```
modmuss50: Just use a thread local and don't change it. I think this is a little bit over engineered.

You can just make the ThreadLocal private, and use public methods in MobEffectUtil to achieve the same idea.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		for (MobEffectInstance effectInstance : collection) {
+			ServerMobEffectEvents.AFTER_REMOVE.invoker()
+					.afterRemove(effectInstance, (LivingEntity) (Object) this, MobEffectUtil.getCommandContext());
+		}
+	}
+
+	@Unique
+	private boolean fabric_isClient() {
```
modmuss50: `fabric_` prefix is not needed with @Unique

modmuss50: Honestly not sure a method for this is that useful, as a reviewer needs to look at the method to see whats special (in this case nothing) about it. You choose, im happy either way.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,219 @@
...
+
+	@WrapMethod(method = "canBeAffected")
+	private boolean allowAddEffect(MobEffectInstance effectInstance, Operation<Boolean> original) {
+		if (this.fabric_isClient()) {
+			return original.call(effectInstance);
+		}
+
+		if (!ServerMobEffectEvents.ALLOW_ADD.invoker().allowAdd(effectInstance, (LivingEntity) (Object) this, MobEffectUtil.getCommandContext())) {
+			return false;
+		} else {
```
modmuss50: Nit: dont use else if unless you need to.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,219 @@
...
+	@Inject(
+			method = "onEffectAdded",
+			at = @At("RETURN")
+	)
+	private void afterAddEffect(MobEffectInstance effectInstance, Entity entity, CallbackInfo ci) {
+		if (this.fabric_isClient()) {
+			return;
+		}
+
+		if (!(this).level().isClientSide()) {
```
modmuss50: is this not always true?

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,219 @@
...
+			return original.call(holder);
+		}
+
+		MobEffectInstance effectInstance = ((LivingEntity) (Object) this).getEffect(holder);
+		boolean cannotRemove = !ServerMobEffectEvents.ALLOW_EARLY_REMOVE.invoker()
+				.allowEarlyRemove(effectInstance, (LivingEntity) (Object) this, MobEffectUtil.getCommandContext());
+
+		if (cannotRemove) {
+			return false;
+		} else {
```
modmuss50: Nit: Dont use else here

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
```diff
@@ -0,0 +1,219 @@
...
+			at = @At("RETURN")
+	)
+	private void afterRemoveEffect(Collection<MobEffectInstance> collection, CallbackInfo ci) {
+		if (this.fabric_isClient()) {
+			return;
+		}
+
+		for (MobEffectInstance effectInstance : collection) {
+			ServerMobEffectEvents.AFTER_REMOVE.invoker()
+					.afterRemove(effectInstance, (LivingEntity) (Object) this, MobEffectUtil.getCommandContext());
```
modmuss50: A self() method that does the `(LivingEntity) (Object) this` cast might not hurt.

modmuss50 (commented):
Awesome start, the game tests are highly appreciated!

I need to look at the command context stuff in a little more detail, but I trust there is a good reason for it :P

A regular test mod to A) act as documentation and B) act as manual tests may also be helpful.

The main design decsion is if the events should be scoped to specific MobEffects, and not just all of them. Other than that mostly just a few nit picks from me.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/FabricMobEffect.java
sylv256: The main use-case is if a user would like a specific subclass of `MobEffect` (like [this](https://github.com/VulpixelMC/legacy-landscape/blob/9b8b6bbf8d3dcdebcdd98410930c2446a0a465bb/src/main/java/gay/sylv/legacy_landscape/effect/EvanescenceEffect.java#L102)) to have special effect update behavior. The `onEffectAdded` method provides consistency with the `onEffectRemoved` method since Mojang's `onEffectAdded` method only has a `int amplifier` parameter and the `LivingEntity`. I think we should potentially also consider adding `onEffectStarted` for consistency (started already exists, but again, it only has an amplifier).

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
sylv256: No, for whatever reason, it's called on the client as well, but as far as I know, it doesn't do much on the client anyways. I believe the only time it's called on the client is when a client is tracking an entity's effect, so only when it's the player themself or a vehicle they're riding.

sylv256: Whoops, nevermind. I forgot to check the context. This is duplicated code.

modmuss50: Unless im blind is this not the same check done just above with `if (this.fabric_isClient()) {`

sylv256: It is

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
sylv256: Honestly, I'm not sure about this. Personally, I'd like to see more opinion on this, but how I would use it would be, for example, an item that prevents a list of effects or all effects from being added/removed or mutates them in some way. The way it is currently seems easier to use IMHO.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/FabricMobEffect.java
sylv256: I've added `onEffectStarted` as well. I personally think it's just nice to have when implementing status effects.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
ChrysanthCow: I don't think this would be a good idea. Mainly because there are cases where you want to affect all mob effects.

For example, in Bovines and Buttercups, I have an effect called Lockdown. Which involves durations for different effects, if an effect is included in the lockdown map attachment, it should not be allowed to be added to the entity.

Exclusively separating events similar to the gamerule events would not allow for global operations such as this.

cputnam-a11y: It would still allow global operations, but it would make them take a different sort of registration using RegistryEntryAddedCallback

cputnam-a11y: But I think I agree that global is better here.

sylv256: > It would still allow global operations, but it would make them take a different sort of registration using RegistryEntryAddedCallback

I think that's just cursed.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/FabricMobEffectEventHandlers.java
```diff
@@ -0,0 +1,53 @@
...
+package net.fabricmc.fabric.api.entity.event.v1.effect;
+
+import net.minecraft.world.effect.MobEffect;
+import net.minecraft.world.effect.MobEffectInstance;
+import net.minecraft.world.entity.LivingEntity;
+
+/**
+ * An extension for {@link MobEffect} subclasses adding basic events.
+ */
+public interface FabricMobEffectEventHandlers {
```
maityyy: The name `FabricMobEffect` would be more appropriate for possible future features and changes

sylv256: True. We should also move it to another module in that case, but I'm not sure where that would be.

maityyy: entity-events is fine

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/ServerMobEffectEvents.java
sylv256: Resolving this unless there are more serious objections.

fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/api/entity/event/v1/effect/FabricMobEffect.java
sylv256: Resolving this unless there are more serious objections.

sylv256:
> A regular test mod to A) act as documentation and B) act as manual tests may also be helpful.

Do we still want to do this? I have more experience (a couple days lol) with testmods after having PR'd to FRAPI so I could probably implement that.

modmuss50 (approved):

Patbox (approved):

PR #5030, opened 2025-12-07
Update sounds.json data generation to match vanilla

pistonpoek:
Updates the SoundTypeBuilderImpl to create a codec that is fully interpreted by the vanilla SoundEventRegistrationSerialization, as this is what the soundType tries to provide. The links between the implementation records and their vanilla classes are being made explicit with additional documentation.

For the public API this means that a replace options has been added and the category method of the soundTypeBuilder has been made deprecated, as it wasn't processed by vanilla and thus provides no functionality.

News tests are also added. The class SoundTypeBuilderTest tests if the SoundTypeBuilder actually provides the expected SoundType. The class SoundTypeCodecTest tests if the implementation codec generates json that is interpreted to the expected vanilla objects. 

Small additional tweaks that have been made are using the existing default variables in matching places and additional validation in the Entry codec to match what is being checked at the builder.

### DIFF
--- a/fabric-data-generation-api-v1/src/client/java/net/fabricmc/fabric/api/client/datagen/v1/builder/SoundTypeBuilder.java
+++ b/fabric-data-generation-api-v1/src/client/java/net/fabricmc/fabric/api/client/datagen/v1/builder/SoundTypeBuilder.java
@@ -67,6 +67,9 @@ public interface SoundTypeBuilder {
 	 * Sets the sound category the sound event must play on.
 	 *
 	 * <p>The default category is {@link SoundSource#NEUTRAL}. GUI elements should use {@link SoundSource#MASTER}.
+	 *
+	 * @apiNote Category is not used by vanilla for the sound event registration,
+	 * but can be specified for fabric and other mods to use.
 	 */
 	SoundTypeBuilder category(SoundSource category);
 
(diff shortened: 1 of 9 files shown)


fabric-data-generation-api-v1/src/client/java/net/fabricmc/fabric/api/client/datagen/v1/builder/SoundTypeBuilder.java
```diff
@@ -64,11 +63,20 @@ static SoundTypeBuilder of() {
 	}
 
 	/**
-	 * Sets the sound category the sound event must play on.
+	 * @deprecated Category is not a field interpreted by vanilla in the sounds file,
+	 * calling this method will have no effect.
+	 */
+	@Deprecated
```
modmuss50: Maybe use `@Deprecated(forRemoval = true)` if we know its tottaly broken?

modmuss50 (approved):
This looks great, the tests are really good to have.

fabric-data-generation-api-v1/src/client/java/net/fabricmc/fabric/api/client/datagen/v1/builder/SoundTypeBuilder.java
pistonpoek: Yeah, seems sensible to eventually want to get rid of it. I will add a commit for it.

PR #5035, opened 2025-12-09
feat: `DebugRendererRegistry`

### LINKED ISSUE #4920: Add DebugRenderer init event
1.21.9 added a the `DebugRenderer.initRenderers` method that gets called whenever a debug renderer might be activated or deactivated. This method then checks the conditions for each debug renderer and collects all active debug renderers in 2 lists: `debugRenderers` and `lateDebugRenderers`. Rendering then happens by simply calling all debug renderers in the list. This is a performance improvement over the behavior in 1.21.8 where the conditions for each debug renderer were checked every frame.

In 1.21.8 debug renderers weren't any different from anything else being rendered, so no special api was necessary, as it would just duplicate `WorldRenderEvents.LAST`. However, now it's reasonable to assume that mods want to also make use of the performance improvement and add their debug renderers to `debugRenderers` or `lateDebugRenderers`.

Note: 25w41a removed the `lateDebugRenderers` field, but the rest remains the same.

sylv256:
## Description
This PR adds a simple `DebugRendererRegistry` class for registering debug renderers. This fixes #4920

## Testing
To test this mod, use the Testmod Client, then spawn a Mannequin. It should look something like this:
image


sylv256:
Marking this as a draft until we decide what to do about Debug API #5045

sylv256:
Closing this. I will rework this for 26.1 and #5045

PR #5040, opened 2025-12-11
Add more precise Block/Item use events

Patbox:
This pr adds 4 new events, which allow for more precise modification of Item/Block use interactions.
Compared to existing UseBlockCallback and UseItemCallback, the new events run only in specific contexts and keep rest of vanilla logic around the use method calls.

Additionally these events use null as a continue value, allowing to return InteractionResult.PASS as a value.

Added events:
- BlockEvents#USE_ITEM_ON - called before Block#useItemOn, replacing it for non-null returns,
- BlockEvents#USE_WITHOUT_ITEM - called before Block#useWithoutItem, replacing it for non-null returns,
- ItemEvents#USE_ON - called before Item#useOn, replacing it for non-null returns,
- ItemEvents#USE - called before Item#use, replacing it for non-null returns,

So for example returning TRY_WITH_EMPTY_HAND for  USE_ITEM_ON will prevent block specific useItemOn call, but it will continue to the Block#useWithoutItem and Item#useOn.

Main thing I'm not sure here is the class names, as it might make sense to make more generic ItemEvents and BlockEvents classes out of these (for any future addition™).

### DIFF
--- a/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/BlockEvents.java
+++ b/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/BlockEvents.java
@@ -33,7 +33,7 @@ import net.fabricmc.fabric.api.event.EventFactory;
 /**
  * Contains events triggered by players interact with blocks.
  */
-public interface BlockUseEvents {
+public interface BlockEvents {
 	/**
 	 * Callback that runs when {@link BlockState#useItemOn(ItemStack, Level, Player, InteractionHand, BlockHitResult)} is called.
 	 */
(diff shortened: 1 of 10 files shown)


Fellteros:
Seems like this also supersets #4818

modmuss50:
> Main thing I'm not sure here is the class names, as it might make sense to make more generic ItemEvents and BlockEvents classes out of these (for any future addition™).

Yeah, I think I would suggest renaming them to be more generic. As well as deprecating the old UseItemCallback/UseBlockCallback classes, it doesnt make sense to have 2 "supported" ways of doing the same thing.

Other than that this PR looks good, I will add to last call assuming those changes will be made before merging (let me know if you need help with that).

modmuss50 (approved):

Patbox:
I wouldn't deprecate UseItemCallback and UseBlockCallback *yet*, as they do still have an use (they run in different place, which can still be useful for some mods). Through updated version of them would make sense (with nicer handling of InteractionResult), but that can be handled later

PR #5041, opened 2025-12-12
(Spaghetti Warning!) [Testing] Event Scopes

sylv256:
based on modmuss's event scopes, will fix this later, pls give feedback thx


sylv256:
Sylv don't forget to make it so early-returning events work and work well

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/unit/ScopedEventTest.java
```diff
@@ -0,0 +1,175 @@
...
+		for (BoolEvent callback : callbacks) {
+			if (!callback.onEvent()) {
+				return false;
+			}
+		}
+
+		return true;
+	});
+
+	private static final Event<ReturnEvent> MODMUSS_RETURN_EVENT = EventFactory.createArrayBacked(ReturnEvent.class, callbacks -> arg -> {
```
modmuss50: Whats a modmuss return event? 🤣

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/EventScope.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.jspecify.annotations.Nullable;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.impl.base.event.TestEncapsulationBreaker3000;
+
+/**
+ * <h1 style="color:red">⚠️Performance Warning⚠️</h1>
+ * Because it is impossible to remove listeners from events,
+ * this class <b>can and will destroy your memory.</b>
```
modmuss50: The original design I had didnt leak memory, as you statically registered to the event once, but then had a way to only handle the event when within the scope. The whole idea of having something to make testing events nicer is to not leak memory.

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/EventScope.java
```diff
@@ -0,0 +1,101 @@
...
+		private @Nullable Object defaultValue;
+		private @Nullable Function<Object[], Object> defaultValueGetter;
+
+		<T> Handler(Event<T> event, T callback) {
+			this.callback = callback;
+
+			Class<T> handlerClass = TestEncapsulationBreaker3000.getClassOfEvent(event);
+			//noinspection unchecked
+			T proxy = (T) Proxy.newProxyInstance(handlerClass.getClassLoader(), new Class<?>[]{handlerClass}, this);
+			event.register(proxy);
```
modmuss50: Should this not be registering to `ScopedEvent`?

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/unit/ScopedEventTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+	@Test
+	void eventScopeTest() {
+		List<String> results = new ArrayList<>();
+
+		SIMPLE_EVENT.invoker().onEvent("1");
+
+		try (var ignored = SIMPLE_EVENT_SCOPE.register(results::add)) {
+			SIMPLE_EVENT.invoker().onEvent("2");
+		}
```
modmuss50, line 178: Why are there 2 different APIs here? Im very confused.

modmuss50 (commented):
Im quite confused as to whats going on here, there seem to be 2 different ways to do the same thing? The whole idea of having a helper like this is to not leak memory/event registrations if a test is ran many times.

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/EventScope.java
sylv256: > The original design I had didnt leak memory, as you statically registered to the event once, but then had a way to only handle the event when within the scope. The whole idea of having something to make testing events nicer is to not leak memory.

I see. Because you can't unregister events, it makes all this very annoying to do.

fabric-api-base/src/test/java/net/fabricmc/fabric/test/base/unit/ScopedEventTest.java
sylv256: I left your API in for comparison.

fabric-api-base/src/testmod/java/net/fabricmc/fabric/test/base/EventScope.java
sylv256: I think I'll try to find a way to not leak memory while also being able to just call `new EventScope()`, perhaps by automatically making `ScopedEvent`s or something idk

sylv256:
> I think I'll try to find a way to not leak memory while also being able to just call `new EventScope()`, perhaps by automatically making `ScopedEvent`s or something idk

I've implemented this in the latest commit, and I believe it is at a point where I can say I'm happy with the API. The only thing that doesn't work exactly as I'd like is using `ScopedEvent` with a terminating event (i.e. an event that is expected to terminate early rather than pass to the next callback often).

sylv256:
Once we figure out what to do with the default return value on terminal events (if we want to do anything), I think it'll probably be ready for review. In my honest opinion, it should be up to the user of the API to ensure the default return value does not interfere with other usages of the event. I think that's the best course of action, especially since most returning events are terminal.

sylv256:
Closing this in favor of @Earthcomputer's event code. I will open a PR with that soon.

sylv256:
Superseded by #5285

PR #5055, opened 2025-12-17
Remove provides for legacy `fabric` modid.

modmuss50:
The `fabric` mod id is a common source of confusion, with this change all mods must now depend on `fabric-api`. They have had 3 years to update.

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -1,9 +1,6 @@
 {
   "schemaVersion": 1,
   "id": "fabric-api",
-  "provides": [
-    "fabric"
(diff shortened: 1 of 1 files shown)


Earthcomputer (approved):

Earthcomputer:
Not only have they had a long time to update, next to no mods will work without recompiling anyway, so since they have to release a new update they can easily change a dependency

cassiancc (approved):
This is long overdue, too many 1.21.11 mods are still using this ID.

sylv256 (approved):
This is the correct solution. `fabric` was a terrible mistake.

PR #5057, opened 2025-12-17
api renames, by apple502j

apple502j:


Earthcomputer (approved):

modmuss50:
The changes look good, do we want to better document this as we go along? Maybe include all of the renames in the PR description and then they can be copied into the overall PR?

apple502j:
@modmuss50 It shouldn't be too hard to autogen the changes from a diff or something, or even source parsing/jar analysis. guess I can add that to description though

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/DataResourceLoader.java
```diff
@@ -35,29 +35,29 @@
...
 public interface DataResourceLoader extends ResourceLoader {
 	/**
-	 * The resource reloader store key for the recipe manager.
+	 * The reload listener state key for the recipe manager.
 	 *
-	 * @apiNote The recipe manager is only available in {@linkplain PackType#SERVER_DATA server data} resource reloaders.
+	 * @apiNote The recipe manager is only available in {@linkplain PackType#SERVER_DATA server data} reload listeners.
 	 * <br/>
-	 * It should <b>only</b> be accessed in the application phase of the resource reloader,
+	 * It should <b>only</b> be accessed in the application phase of the reload listeners,
```
LambdAurora:
```suggestion
	 * It should <b>only</b> be accessed in the application phase of the resource reload listeners,
```

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/DataResourceLoader.java
```diff
@@ -35,29 +35,29 @@
...
 	 * and you should depend on {@link net.fabricmc.fabric.api.resource.v1.reloader.ResourceReloaderKeys.Server#RECIPES}.
 	 */
 	PreparableReloadListener.StateKey<RecipeManager> RECIPE_MANAGER_KEY = new PreparableReloadListener.StateKey<>();
 	/**
-	 * The resource reloader store key for the advancement loader.
+	 * The reload listener state key for the advancement loader.
 	 *
-	 * @apiNote The advancement loader is only available in {@linkplain PackType#SERVER_DATA server data} resource reloaders.
+	 * @apiNote The advancement loader is only available in {@linkplain PackType#SERVER_DATA server data} reload listeners.
```
LambdAurora:
```suggestion
	 * @apiNote The advancement loader is only available in {@linkplain PackType#SERVER_DATA server data} resource reload listeners.
```

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/DataResourceLoader.java
```diff
@@ -35,29 +35,29 @@
...
 	 * <br/>
-	 * It should <b>only</b> be accessed in the application phase of the resource reloader,
+	 * It should <b>only</b> be accessed in the application phase of the reload listeners,
 	 * and you should depend on {@link net.fabricmc.fabric.api.resource.v1.reloader.ResourceReloaderKeys.Server#ADVANCEMENTS}.
 	 */
 	PreparableReloadListener.StateKey<ServerAdvancementManager> ADVANCEMENT_LOADER_KEY = new PreparableReloadListener.StateKey<>();
 	/**
-	 * The resource reloader store key for the data resource store.
+	 * The reload listener state key for the data resource store.
```
LambdAurora:
```suggestion
	 * The resource reload listener state key for the data resource store.
```

fabric-resource-loader-v1/src/main/java/net/fabricmc/fabric/api/resource/v1/reloader/ResourceReloaderKeys.java
```diff
@@ -19,43 +19,43 @@
...
 	 */
 	public static final class Client {
-		public static final Identifier BLOCK_ENTITY_RENDERERS = Identifier.withDefaultNamespace("block_entity_renderers");
-		public static final Identifier BLOCK_RENDER_MANAGER = Identifier.withDefaultNamespace("block_render_manager");
-		public static final Identifier CLOUD_CELLS = Identifier.withDefaultNamespace("cloud_cells");
-		public static final Identifier EQUIPMENT_MODELS = Identifier.withDefaultNamespace("equipment_models");
-		public static final Identifier ENTITY_RENDERERS = Identifier.withDefaultNamespace("entity_renderers");
-		public static final Identifier DRY_FOLIAGE_COLORMAP = Identifier.withDefaultNamespace("dry_foliage_colormap");
-		public static final Identifier FOLIAGE_COLORMAP = Identifier.withDefaultNamespace("foliage_colormap");
+		public static final Identifier BLOCK_ENTITY_RENDER_DISPATCHER = Identifier.withDefaultNamespace("block_entity_render_dispatcher");
```
LambdAurora, line 50: I have concerns with the value renames (not the field name itself).
Doing this may cause implicit breakage that a field rename doesn't as it would be an explicit breakage.

Anyone using a similar value but not the field directly will diverge, and this will not be caught by a compile error, which may be hard to find.

modmuss50: This is a valid concern that we should discuss, but I would like to merge this PR to make whole process easier. I think the best way forward will be to review the main PR.

PR #5059, opened 2025-12-17
Creative Mode Tab API renames

cassiancc:
Renames the Item Group API to the Creative Mode Tab API.


modmuss50:
Question/thought: Is "Mode" needed in the name of this API? Would Creative Tab API be better?

cassiancc:
"Mode" being included in the name of the class is something I don't particularly agree with, but as the API is here to create `CreativeModeTab`s, I opted to include it in the API name so that it's easier to find coming from the class itself.

modmuss50:
Going to merge this PR, we can contiune this discussion in the other PR.

PR #5061, opened 2025-12-17
Block, Block View, Item, and Tag API renames

cassiancc:
Renames the Block, Block View, Item, and Tag APIs to match Mojang's names.

### DIFF
--- a/fabric-block-api-v1/src/main/java/net/fabricmc/fabric/api/block/v1/FabricBlock.java
+++ b/fabric-block-api-v1/src/main/java/net/fabricmc/fabric/api/block/v1/FabricBlock.java
@@ -59,10 +59,10 @@ public interface FabricBlock {
 	 * The block entity should override {@code RenderDataBlockEntity#getBlockEntityRenderData} to return
 	 * the necessary data. Refer to the documentation of {@code RenderDataBlockEntity} for more information.
 	 * <pre>{@code @Override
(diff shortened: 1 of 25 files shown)


fabric-block-api-v1/src/main/java/net/fabricmc/fabric/api/block/v1/FabricBlock.java
```diff
@@ -59,10 +59,10 @@ public interface FabricBlock {
 	 * The block entity should override {@code RenderDataBlockEntity#getBlockEntityRenderData} to return
 	 * the necessary data. Refer to the documentation of {@code RenderDataBlockEntity} for more information.
 	 * <pre>{@code @Override
-	 * public BlockState getAppearance(BlockState state, BlockRenderView renderView, BlockPos pos, Direction side, @Nullable BlockState sourceState, @Nullable BlockPos sourcePos) {
-	 *     if (renderView instanceof ServerWorld serverWorld) {
+	 * public BlockState getAppearance(BlockState state, BlockAndTintGetter blockAndTintGetter, BlockPos pos, Direction side, @Nullable BlockState sourceState, @Nullable BlockPos sourcePos) {
+	 *     if (renderView instanceof ServerLevel serverLevel) {
```
Toffikk: this if condition should be changed

modmuss50 (approved):

PR #5067, opened 2025-12-17
renames in command-api v2

Toffikk:


fabric-command-api-v2/src/client/java/net/fabricmc/fabric/api/client/command/v2/ClientCommands.java
```diff
@@ -53,19 +53,19 @@
...
  * 				return 0;
  * 			})
  * 		);
  * });
  * }
  * </pre>
  */
-public final class ClientCommandManager {
-	private ClientCommandManager() {
+public final class ClientCommands {
```
modmuss50, line 67: Where does the inspiration for this rename come from? It seems very vague to me.

Toffikk: I took it from the yarn -> mojmap mapping comparison as yarn CommandManager is Commands in mojmap

PR #5069, opened 2025-12-18
[1.21.11] Allow subclasses of lang provider to change output path

skycatminepokie:
Just a nice-to-have. Useful when using [Nucleoid's Server-Translations](https://github.com/NucleoidMC/Server-Translations).


fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/api/datagen/v1/provider/FabricLanguageProvider.java
```diff
@@ -108,7 +108,7 @@ public CompletableFuture<?> run(CachedOutput writer) {
 		});
 	}
 
-	private Path getLangFilePath(String code) {
+	protected Path getLangFilePath(String code) {
```
Juuxel, line 116: This should have a (short) javadoc comment.

modmuss50 (approved):

PR #5060, opened 2025-12-17
Rename APIs to match Mojang's naming

modmuss50:
# API Renames

#### `net/fabricmc/fabric/api/biome/v1/BiomeModificationContext$EffectsContext`
- `setFoliageColor` → `setFoliageColorOverride`
- `clearFoliageColor` → `clearFoliageColorOverride`
- `setDryFoliageColor` → `setDryFoliageColorOverride`
- `clearDryFoliageColor` → `clearDryFoliageColorOverride`
- `setGrassColor` → `setGrassColorOverride`
- `clearGrassColor` → `clearGrassColorOverride`

#### `net/fabricmc/fabric/api/biome/v1/BiomeModificationContext$SpawnSettingsContext` → `BiomeModificationContext$MobSpawnSettingsContext`
- `setCreatureSpawnProbability` → `setCreatureGenerationProbability`
- `getSpawnEntries` → `getMobs`
- `setSpawnCost` → `addMobCharge`
- `clearSpawnCost` → `clearMobCharge`

#### `net/fabricmc/fabric/api/biome/v1/BiomeModificationContext`
- `getSpawnSettings` → `getMobSpawnSettings`

#### `net/fabricmc/fabric/api/biome/v1/BiomeSelectionContext`
- `getBiomeRegistryEntry` → `getBiomeHolder`

#### `net/fabricmc/fabric/api/client/command/v2/FabricClientCommandSource`
- `getWorld` → `getLevel`

#### `net/fabricmc/fabric/api/client/datagen/v1/builder/SoundTypeBuilder`
- `category` → `source`

#### `net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientTickEvents`
- `START_WORLD_TICK` → `START_LEVEL_TICK`
- `END_WORLD_TICK` → `END_LEVEL_TICK`

#### `net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientWorldEvents$AfterClientWorldChange` → `ClientLevelEvents$AfterClientLevelChange`
- `afterWorldChange` → `afterLevelChange`

#### `net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientWorldEvents` → `ClientLevelEvents`
- `AFTER_CLIENT_WORLD_CHANGE` → `AFTER_CLIENT_LEVEL_CHANGE`

[...]

PR #5082, opened 2025-12-19
rendering-v1 renames

sylv256:
#5060


fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/BlockColorRegistry.java
```diff
@@ -21,29 +21,29 @@
 import net.minecraft.client.color.block.BlockColor;
 import net.minecraft.world.level.block.Block;
 
-import net.fabricmc.fabric.impl.client.rendering.ColorProviderRegistryImpl;
+import net.fabricmc.fabric.impl.client.rendering.BlockColorRegistryImpl;
 
-public interface ColorProviderRegistry<T, Provider> {
-	ColorProviderRegistry<Block, BlockColor> BLOCK = ColorProviderRegistryImpl.BLOCK;
+public interface BlockColorRegistry<T, Color> {
```
PepperCode1: I don't agree with this rename given the current design, but this class (and the corresponding impl/mixins) should be rewritten. It uses this design because there used to be a registry for item colors in the form of an `ITEM` field, but that was removed due to changes in vanilla 1.21.4. The methods should be made static and there shouldn't be any generic types now. With such a new design I think `BlockColorRegistry` makes sense.

That said, I don't know if that refactor should be part of this PR. Ask Modmuss.

For when the refactor is done, I know the vanilla interface is called `BlockColor`, but I feel like some of the javadoc changes are confusing because "color" can refer both to a `BlockColor` instance and an actual `int` ARGB color, so I would leave the term "provider" to refer to `BlockColor` instances to clarify this.

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/ChunkSectionLayerMap.java
```diff
@@ -17,39 +17,38 @@
...
 /**
- * Use to associate blocks or fluids with block render layer other than default (solid).
+ * Use to associate blocks or fluids with a chunk section layer other than default (solid).
  *
- * <p>{@link RenderTypes} control how sprite pixels for fluids and blocks are blended with the scene. Consult the
- * vanilla {@link RenderTypes} implementation for examples.
+ * <p>{@link ChunkSectionLayer} control how sprite pixels for fluids and blocks are blended with the scene. Consult the
+ * vanilla {@link ChunkSectionLayer} implementation for examples.
  *
  * <p>The Fabric Renderer API can be used to control this at a per-quad level via {@code BlendMode}.
```
PepperCode1:
```suggestion
 * <p>The Fabric Renderer API can be used to control this at a per-quad level.
```

Not related to this PR but better to do now so we don't forget.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/CustomSpriteSourcesTest.java
```diff
@@ -39,12 +39,12 @@
...
 
-public class CustomAtlasSourcesTest implements ClientModInitializer {
+public class CustomSpriteSourcesTest implements ClientModInitializer {
 	@Override
 	public void onInitializeClient() {
-		AtlasSourceRegistry.register(Identifier.fromNamespaceAndPath("fabric-rendering-v1-testmod", "double"), DoubleAtlasSource.CODEC);
+		SpriteSourceRegistry.register(Identifier.fromNamespaceAndPath("fabric-rendering-v1-testmod", "double"), DoubleAtlasSource.CODEC);
 	}
 
 	private static class DoubleAtlasSource implements SpriteSource {
```
PepperCode1: This class should be renamed to `DoubleSpriteSource`.

fabric-rendering-v1/src/testmodClient/java/net/fabricmc/fabric/test/rendering/client/LevelRenderEventsTests.java
```diff
@@ -121,31 +122,31 @@ public void runTest(ClientGameTestContext context) {
...
+		assertNotNull(context.sectionsToRender(), "sectionRenderState is null");
 	}
 
-	private static void assertAbstractRenderContext(AbstractWorldRenderContext context) {
+	private static void assertAbstractRenderContext(AbstractLevelRenderContext context) {
 		assertNotNull(context.gameRenderer(), "gameRenderer is null");
-		assertNotNull(context.worldRenderer(), "worldRenderer is null");
-		assertNotNull(context.worldState(), "worldRenderState is null");
+		assertNotNull(context.levelRenderer(), "worldRenderer is null");
+		assertNotNull(context.levelState(), "worldRenderState is null");
```
PepperCode1: These strings should be updated accordingly.

PepperCode1 (changes requested):

fabric-rendering-v1/src/client/java/net/fabricmc/fabric/api/client/rendering/v1/BlockColorRegistry.java
Earthcomputer: Any refactor beyond the renames should absolutely not be part of this PR

sylv256: I'll rename this back to use `ColorProvider` for now then since it's not in line with the 1.21.4+ changes.

PepperCode1 (approved):

PR #5086, opened 2025-12-19
data-attachment-api-v1 renames

sylv256:
#5060


fabric-item-api-v1/src/testmod/java/net/fabricmc/fabric/test/item/CustomDamageTest.java
maityyy, line 103: Unnecessary change?

sylv256: Nope, that fails checkstyle.

sylv256: My IDE picked it up automatically so I changed it without realizing it was in item api lol. I can remove it if y'all want but I see no need.

PR #5094, opened 2025-12-22
Fix incorrect class name in content registries provided javadoc

Juuxel:
The class name in the comment also needs adjusting with the rename PR.


sylv256 (approved):

modmuss50 (approved):
This class was not renamed, I know this is broken in loom atm but still makes sense to merge.

PR #5098, opened 2025-12-24
Add an event to modify dimension environment attributes

Hugman76:
This pull request introduces a new API to facilitate the modification of dimension environment attributes, which are natively immutable. 

## Rationale

Since the introduction of environment attributes, several use cases have emerged requiring modifications at the dimension level. Currently, applying an attribute across an entire dimension requires iterating through and modifying every individual biome within that dimension using the Biomes Modification API.

This event provides a direct interface for interacting with dimensions attributes. Furthermore, it allows datapack and modpack creators to maintain granular control, as dimension-level modifications are designed with lower priority than biome-specific definitions during attribute evaluation in vanilla.

## Usage Example
```java
DimensionEvents.MODIFY_ATTRIBUTES.register((dimension, attributes, registries) -> {
  if (dimension.is(BuiltinDimensionTypes.OVERWORLD)) {
    attributes.set(EnvironmentAttributes.CLOUD_COLOR, PURPLE);
  }
});
```


fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/ModificationPhase.java
```diff
@@ -0,0 +1,62 @@
...
+ * </ol>
+ */
+public enum ModificationPhase {
+	/**
+	 * The appropriate phase for enriching dimensions by adding to them without relying on
+	 * other information in the dimension, or removing other features.
+	 *
+	 * <p><b>Examples:</b> New ores, new vegetation, new structures
+	 */
+	ADDITIONS,
```
maityyy: I don't think ModificationPhase is needed at all when we just change attributes (read: float), vanilla `AttributeModifier` already does everything necessary. What compatibility issue does ModificationPhase solve for dimensions? ModificationPhase for biomes was partly introduced because of biome features that have ordering requirements (not following which can cause a crash).

Plus there seems to be a lot of documentation issues, as biome features are just mentioned here

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/modification/MinecraftServerMixin.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.server.MinecraftServer;
+
+import net.fabricmc.fabric.impl.dimension.modification.DimensionModificationImpl;
+
+@Mixin(MinecraftServer.class)
+public abstract class MinecraftServerMixin {
+	@Shadow
+	public abstract RegistryAccess.Frozen registryAccess();
+
+	@Inject(method = "<init>", at = @At(value = "RETURN"))
```
maityyy: Use `ServerLifecycleEvents.SERVER_STARTED`

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/DimensionModificationContext.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Allows {@link DimensionType} properties to be modified.
+ */
+public interface DimensionModificationContext {
+	/**
+	 * Returns the modification context for the dimensions environment attributes.
+	 */
+	AttributesContext getAttributes();
+
+	interface AttributesContext {
```
maityyy: Do we really need this interface to modify an `EnvironmentAttributeMap`, which by structure is really just a map? Why not give the user `EnvironmentAttributeMap.Builder`? Especially since your AttributesContextImpl seems to just mirror the methods of the builder

maityyy (changes requested):
It's a useful API that we should definitely have, but I'm sure all we need is something like `EnchantmentEvents` or `LootTableEvents`, copying the Biome Modification API design is completely unnecessary, confuses users and inflates the code base too much

Hugman76:
Thanks a lot for the feedback @maityyy, I fully agreed with your points and have almost nothing to add really. I've gone back and remade the whole thing from the ground up using events.

I have to note that modifying attributes using `ServerLifecycleEvents.SERVER_STARTED` seems to only modify dimensions after they've been baked into the levels. I suppose this means at some point the level and dimension are unsynced. This results in this very same event not being able to fetch the modified attributes (as tested in `FabricDimensionTest`.)

Therefore, I used `ServerLifecycleEvents.SERVER_STARTING` to fire the attribute modifications.

I've updated the PR's title and description accordingly.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
```diff
@@ -0,0 +1,113 @@
...
+					registry.registrationInfos.put(key, newInfo);
+				}
+			}
+		}
+
+		stopwatch.stop();
+
+		if (dimensionsProcessed > 0) {
+			LOGGER.info("Applied modifications to {} of {} dimensions in {}", dimensionsChanged, dimensionsProcessed, stopwatch);
+		}
```
modmuss50: Is the stopwatch needed?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/RegistryAccessImmutableRegistryAccessMixin.java
```diff
@@ -0,0 +1,43 @@
...
+@Mixin(RegistryAccess.ImmutableRegistryAccess.class)
+public class RegistryAccessImmutableRegistryAccessMixin implements DimensionModificationMarker {
+	@Unique
+	private boolean dimensionsModified;
+
+	@Override
+	public void fabric_markDimensionsModified() {
+		if (dimensionsModified) {
+			throw new IllegalStateException("Dimensions in this dynamic registries instance have already been modified");
+		}
```
modmuss50, line 40: Question: Is this the only use for the `dimensionsModified` boolean?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/modification/RegistryAccessImmutableRegistryAccessMixin.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+
+import net.minecraft.core.RegistryAccess;
+
+import net.fabricmc.fabric.impl.dimension.modification.DimensionModificationMarker;
+
+/**
+ * This Mixin allows us to keep backup copies of dimensions for
```
modmuss50: What do you mean by "keep backup copies"?

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionsImpl.java
```diff
@@ -0,0 +1,28 @@
...
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
+import net.fabricmc.fabric.impl.dimension.modification.DimensionModificationImpl;
+
+public class FabricDimensionsImpl implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		ServerLifecycleEvents.SERVER_STARTING.register(server -> DimensionModificationImpl.finalizeWorldGen(server.registryAccess()));
+	}
+}
```
modmuss50: I would suggest just having a single class, DimensionModificationImpl can be a ModInitializer

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionsImpl.java
```diff
@@ -0,0 +1,28 @@
...
+package net.fabricmc.fabric.impl.dimension;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
+import net.fabricmc.fabric.impl.dimension.modification.DimensionModificationImpl;
+
+public class FabricDimensionsImpl implements ModInitializer {
+	@Override
+	public void onInitialize() {
+		ServerLifecycleEvents.SERVER_STARTING.register(server -> DimensionModificationImpl.finalizeWorldGen(server.registryAccess()));
```
modmuss50: What is the reason behind doing this in SERVER_STARTING? Would a targgeted mixin that does it when the DimensionType is created be better?

fabric-dimensions-v1/src/main/resources/fabric-dimensions-v1.accesswidener
```diff
@@ -0,0 +1,9 @@
+accessWidener	v1	official
+
+# Top-Level Dimension Fields Access
+
+accessible	field	net/minecraft/world/level/dimension/DimensionType	attributes	Lnet/minecraft/world/attribute/EnvironmentAttributeMap;
+mutable	field	net/minecraft/world/level/dimension/DimensionType	attributes	Lnet/minecraft/world/attribute/EnvironmentAttributeMap;
+
+# Registry mutation sync
+accessible	field	net/minecraft/core/MappedRegistry	registrationInfos	Ljava/util/Map;
```
modmuss50: Perfer to use accessors as it saves there needing to be another minecraft jar in the dev env.

fabric-dimensions-v1/src/testmod/java/net/fabricmc/fabric/test/dimension/FabricDimensionTest.java
```diff
@@ -16,22 +16,46 @@
...
+
+		DimensionEvents.MODIFY_ATTRIBUTES.register((dimension, attributes, _) -> {
+			if (dimension.is(BuiltinDimensionTypes.OVERWORLD)) {
+				attributes.set(EnvironmentAttributes.CLOUD_COLOR, PURPLE);
+			}
+		});
+
+		ServerLifecycleEvents.SERVER_STARTED.register(server -> {
+			ServerLevel overworld = server.getLevel(Level.OVERWORLD);
+			overworld.setDayTime(6000);
```
modmuss50: Question: Is this needed to get the color? If so a game test would likely be better.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
```diff
@@ -0,0 +1,113 @@
...
+
+import net.fabricmc.fabric.api.dimension.v1.DimensionEvents;
+
+public final class DimensionModificationImpl {
+	private static final Logger LOGGER = LoggerFactory.getLogger(DimensionModificationImpl.class);
+
+	private DimensionModificationImpl() {
+	}
+
+	public static void finalizeWorldGen(RegistryAccess impl) {
```
modmuss50: impl isnt a good param name

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
```diff
@@ -0,0 +1,113 @@
...
+		// Now that we apply dimension modifications inside the MinecraftServer constructor, we should only ever do
+		// this once for a dynamic registry manager. Marking the dynamic registry manager as modified ensures a crash
+		// if the precondition is violated.
+		DimensionModificationMarker modificationTracker = (DimensionModificationMarker) impl;
+		modificationTracker.fabric_markDimensionsModified();
+
+		Registry<DimensionType> dimensions = impl.lookupOrThrow(Registries.DIMENSION_TYPE);
+
+		// Build a list of all dimension keys in ascending order of their raw-id to get a consistent result in case
+		// someone does something stupid.
```
modmuss50: Stupid such as? Doing this sorting will likely hide mistakes.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionsImpl.java
maityyy: Dimension is a part of the dynamic registry, experience shows that it is better to modify dynamic registry entries after the game is loaded, that's exactly what the Biome Modification API does (fun fact: BMAPI can use this event too)

Hugman76: I believe this could result in the client firing modifications when it receives the data from the server, which is not what the event is there for

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/modification/RegistryAccessImmutableRegistryAccessMixin.java
Hugman76: Woops, that's a remain of previous iteration where I essentially clone the Biome Modifications API

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/FabricDimensionsImpl.java
modmuss50: That makes sense, thanks. I do think we really need some more generic APIs desgined specifically for this kind of thing, but thats not for this PR.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
maityyy: I think not, it's a leftover from the PR author's initial attempt to copy the Biome Modification API design. BMAPI uses Stopwatch probably because of the logic of sorting modifications

modmuss50: Yeah, I imagined so. I dont think its really that useful in biome modficiation either tbh.

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/DimensionEvents.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Events for manipulating dimensions.
+ */
+public class DimensionEvents {
+	/**
+	 * This event can be used to modify the environment attributes of a dimension.
+	 * The main use case is to add modded attributes to vanilla or modded dimensions
+	 */
+	public static final Event<ModifyAttributes> MODIFY_ATTRIBUTES = EventFactory.createArrayBacked(ModifyAttributes.class, listeners -> (dimension, attributes, registries) -> {
```
maityyy, line 35: Question: should we provide modification to all Dimension fields or just attributes? I think the first option is preferable, because in any case it will be after the game is loaded, i.e. at a single point
(edit: not PR request)

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	/**
+	 * Applies the changes from the events of {@link DimensionEvents} to a dimension.
+	 *
+	 * @return true if the dimension was changed
+	 */
+	private static boolean applyChanges(Holder<DimensionType> dimension, RegistryAccess registries) {
+		EnvironmentAttributeMap oldAttributes = dimension.value().attributes();
+
```
maityyy: Checkstyle: remove these spaces between lines below

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/mixin/dimension/RegistryAccessImmutableRegistryAccessMixin.java
Hugman76: Yep! This is meant to prevent any accidental double-modification, but that should never occur

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/DimensionEvents.java
Hugman76: We can definitely provide more field modifications yep. Environment attributes just felt the most pertinent as of recent changes in dimensions definition

fabric-dimensions-v1/src/testmod/java/net/fabricmc/fabric/test/dimension/FabricDimensionTest.java
Hugman76: Yep that's needed. I don't understand how the MODIFY_ATTRIBUTES can be fired in the game test environment though. 
EDIT: I figured it out

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/DimensionModificationImpl.java
```diff
@@ -69,22 +71,21 @@ public static void finalizeWorldGen(RegistryAccess impl) {
...
+					registrationInfos.put(key, newInfo);
 				}
 			}
 		}
 
-		stopwatch.stop();
-
 		if (dimensionsProcessed > 0) {
-			LOGGER.info("Applied modifications to {} of {} dimensions in {}", dimensionsChanged, dimensionsProcessed, stopwatch);
+			LOGGER.info("Applied modifications to {} of {} dimensions", dimensionsChanged, dimensionsProcessed);
```
maityyy: Logging is not necessary, please remove

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/api/dimension/v1/DimensionEvents.java
maityyy: Well, anyway we can create a new event even without deprecating the old one

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/modification/DimensionModificationImpl.java
Hugman76: Leftover from previous iteration.

maityyy (approved):

fabric-dimensions-v1/src/main/java/net/fabricmc/fabric/impl/dimension/DimensionModificationImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import net.minecraft.world.level.dimension.DimensionType;
+
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.dimension.v1.DimensionEvents;
+import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
+import net.fabricmc.fabric.mixin.dimension.DimensionTypeAccessor;
+import net.fabricmc.fabric.mixin.dimension.MappedRegistryAccessor;
+
+public class DimensionModificationImpl implements ModInitializer {
+	private static final Logger LOGGER = LoggerFactory.getLogger(DimensionModificationImpl.class);
```
modmuss50: NIT: LOGGER is never used

modmuss50 (approved):

PR #5108, opened 2025-12-30
add resource reload start/end events

chyzman:


Benonardo (approved):

fabric-lifecycle-events-v1/src/client/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientLifecycleEvents.java
```diff
@@ -48,6 +48,24 @@ private ClientLifecycleEvents() {
 		}
 	});
 
+	/**
+	 * Called before the client begins loading resources.
+	 */
+	public static final Event<StartResourceReload> START_RESOURCE_RELOAD = EventFactory.createArrayBacked(StartResourceReload.class, callbacks -> (client, isFirst) -> {
```
UpcraftLP, line 54: It would be helpful to explain the `isFirst` parameter in the javadoc

LambdAurora (changes requested):
The events should be in `resource-loader-v1`, in a new class (`ClientResourceLoadingEvents`), and in a future PR most likely move the data-pack events to resource loader too.

Aside from that, the events are missing important context: the resource manager.
And in the case of the end event, it's also missing an `Optional<Throwable>` in the case of errors, (and the event should trigger even if the resource load fails).

It's the kind of low-bandwidth event where I'd recommend creating a context object just to ensure a stable API while allowing to add more context if needs be.

cputnam-a11y:
Could it also include the store on success?
And yeah, context probably makes sense here

LambdAurora:
Including the store is not a bad idea, yes.

chyzman:
> The events should be in `resource-loader-v1`, in a new class (`ClientResourceLoadingEvents`), and in a future PR most likely move the data-pack events to resource loader too.

im not sure exactly what package to put it in in the resource loader module, do i mimic the package structure from a typical event module?

> Could it also include the store on success?

what does this mean exactly?

PR #5111, opened 2025-12-31
Migrate to Markdown javadoc

modmuss50:
Automatically migrate all javadoc to Markdown


fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
```diff
@@ -27,154 +27,102 @@
...
- *            If no context is necessary, {@link Void} should be used and {@code null} instances should be passed.
- */
+/// An object that allows retrieving APIs from entities.
+/// Instances of this interface can be obtained through [#get]
+///
+/// When trying to [#find] an API for an entity, the provider registered for the entity type will be queried if it exists.
+/// If it doesn't exist, or if it returns `null`, the fallback providers will be queried in order.
+/// ### Usage Example
+/// Let's pretend that we have the following interface that we want to attach to entities.
+/// <pre>
```
modmuss50: This doesn't look correct.

fabric-api-base/src/main/java/net/fabricmc/fabric/impl/base/toposort/NodeSorting.java
```diff
@@ -28,33 +28,29 @@
...
- *     <li>Sort SCCs with respect to each other by respecting constraints, and using the comparator in case of a tie.</li>
- * </ol>
- */
+/// Contains a topological sort implementation, with tie breaking using a [Comparator].
+///
+/// The final order is always deterministic (i.e. doesn't change with the order of the input elements or the edges),
+/// assuming that they are all different according to the comparator. This also holds in the presence of cycles.
+///
+/// The steps are as follows:
+/// <ol>
```
modmuss50: Not correct

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/entity/EntityApiLookup.java
deirn: Well techically HTML tags is valid markdown 😅

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiLookup.java
```diff
@@ -31,266 +31,164 @@
...
- *
- * <h3>Usage Example</h3>
- * Let us pretend we have the following interface that we would like to attach to some blocks depending on the direction.
- *
- * <pre>{@code
- * public interface FluidContainer {
- *     boolean containsFluids(); // return true if not empty
- * }}</pre>
- * Let us first create a static {@code BlockApiLookup} instance that will manage the registration and the query.
- *
```
modmuss50, line 51: Its made a right mess in places such as this.

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
```diff
@@ -25,133 +25,71 @@
...
...ivate ItemStackApiLookupImpl(Identifier id, Class<?> apiClass, Class<?> contextClass){// We don't use these parameters, so nothing to do here.// In practice, these parameters should be stored and exposed with identifier(), apiClass() and contextClass() getter functions.}// We will use an ApiProviderMap to store the providers.private final ApiProviderMap<Item, ItemStackApiProvider<A, C>> providerMap = ApiProviderMap.create();＠Nullablepublic A find(ItemStack stack, C context){ItemStackApiProvider<A, C> provider = providerMap.get(stack.getItem());if (provider == null){return null;}else{return provider.find(stack, context);}}public void register(ItemStackApiProvider provider, Item item){// Let's add a few null checks just in case.Objects.requireNonNull(provider, "ItemStackApiProvider may not be null.");Objects.requireNonNull(item, "Item may not be null.");// Register the provider, or warn if it is already registeredif (providerMap.putIfAbsent(item, provider) != null){// Emit a warning printing the item ID to help users debug more easily.LoggerFactory.getLogger("The name of your mod").warn("Encountered duplicate API provider registration for item " + Registry.ITEM.getId(item) + ".");}}}
```
Juuxel: This is wrong

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
```diff
@@ -25,133 +25,71 @@
...
- * @param <L> The type of the lookup implementation, similar to the existing {@link net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup BlockApiLookup}.
- */
+/// A map meant to be used as the backing storage for custom `ApiLookup` instances,
+/// to implement a custom equivalent of [BlockApiLookup#get][net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup#get].
+/// ### Usage Example
+/// We will be implementing the following simplified version of an API lookup interface for item stacks
+/// to illustrate how to use [ApiLookupMap] and [ApiProviderMap].
+///
+/// ```java
+/// public interface ItemStackApiLookup<A, C>{static <A, C> ItemStackApiLookup<A, C> get(Identifier lookupId, Class<A> apiClass, Class<C> contextClass){return ItemStackApiLookupImpl.get(lookupId, apiClass, contextClass);}// Find an API instance.＠NullableA find(ItemStack stack, C context);// Expose the API for some item.void register(ItemStackApiProvider<A, C> provider, Item item);interface ItemStackApiProvider<A, C>{// Return an API instance if available, or null otherwise.＠NullableA find(ItemStack stack, C context);}}
```
Juuxel: And so is this

fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/custom/ApiLookupMap.java
modmuss50: Good spot, the intelij auto formatter put them all on one line I found a fixed a load but clearly not all of them...

fabric-menu-api-v1/src/main/java/net/fabricmc/fabric/api/menu/v1/package-info.java
```diff
@@ -14,42 +14,37 @@
...
+/// should be an [net.fabricmc.fabric.api.menu.v1.ExtendedMenuProvider], or a
+/// [net.minecraft.world.SimpleMenuProvider] that wraps such factory.
+/// ## Overwriting menus
+/// You might have noticed that calling [openMenu][net.minecraft.world.entity.player.Player#openMenu(net.minecraft.world.MenuProvider)] while on another screen will move
+/// the cursor to the center of the screen. This is because the current screen gets closed before
+/// opening the screen, resetting the cursor position. Since this behavior can be problematic,
+/// this API provides a way to disable this. By overriding
+/// [net.fabricmc.fabric.api.menu.v1.FabricMenuProvider#shouldCloseCurrentScreen()]
+/// on the menu factory to return `false` and passing that to the
+/// `openMenu` method, it will stop closing the screen and instead "overwrites" it.
```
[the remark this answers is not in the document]
sylv256: why is it converting back into html javadoc :sob:

Earthcomputer (changes requested):
As brought up on Discord, we need to have a proper discussion about whether imports should be used or removed for docs, I and quite a few others think removing them is a mistake. If the tooling doesn't support imports in markdown that means we should wait for the tooling to update before migrating.

Kilip1000:
I can't propose changes properly, due to my browser being slowed to a crawl while viewing changed files.

`EventTests.testDeterministicOrdering()`s opening square brackets need to be escaped though.
Probably in lots of other places too.

modmuss50:
> As brought up on Discord, we need to have a proper discussion about whether imports should be used or removed for docs, I and quite a few others think removing them is a mistake. If the tooling doesn't support imports in markdown that means we should wait for the tooling to update before migrating.

I responded to this on Discord:

> I was quite keen on having the same doc format everywhere (API and the yarn replacement). I don't see an issue with using it for documenting Minecraft from scratch where we cannot do imports anyway. But right now it does seem like the tooling is limited (bad migration from IntelliJ, spotless/checkstyle, IntelliJ doesn't render the code blocks nicely)

> But on the other hand, I have done 95% of the work to migrate, just need to check/fix a few things. The imports arent the end of the world, and could be fixed automatically when the tools allow (if thats even going to happen).

I am unsure is this is something we want to do, the benefits do seem minimal but I have done a lot of the work, and having everything use MD (FAPI and the yarn replacement) does seem better for the user?

modmuss50:
As discussed we are not going to do this right now, new javadoc can be written in markdown if desired.

PR #5112, opened 2025-12-31
Add generated info to `ServerChunkEvents#CHUNK_LOAD`

ishland:
The `ServerChunkEvents#CHUNK_GENERATE` added earlier in #4183 did not consider the use cases of listening to loaded but not newly generated chunks. 

This PR addresses this by merging two events into the original `ServerChunkEvents#CHUNK_LOAD` event, which also brings this event in line with what most other platform is doing.


modmuss50 (approved):
Im happt with this. It is a breaking change but only a minor one and for the best.

PR #5115, opened 2026-01-01
Don't generate the Indigo config file by default

modmuss50:
- Don't generate the Indigo config file by default
  - If the config exists it still gets read
- Add `fabric.indigo.generateConfigFile` system property to restore the previous behaviour
- General code modernisation

### DIFF
--- a/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
+++ b/fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
@@ -16,13 +16,15 @@
 
 package net.fabricmc.fabric.impl.client.indigo;
 
-import java.io.File;
-import java.io.FileInputStream;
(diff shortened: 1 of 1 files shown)


cassiancc (approved):

fabric-renderer-indigo/src/client/java/net/fabricmc/fabric/impl/client/indigo/Indigo.java
```diff
@@ -43,68 +45,55 @@ public class Indigo implements ClientModInitializer {
...
-			return false;
-		default:
-			return defValue;
-		}
+	private static boolean asBoolean(@Nullable String property, boolean defValue) {
+		return switch (asTriState(property)) {
+		case TRUE -> true;
+		case FALSE -> false;
+		default -> defValue;
+		};
```
Juuxel: I believe this can just be `asTriState(property).orElse(defValue)`

PR #5119, opened 2026-01-03
Allow registering large payloads with deferred max packet size

DennisOchulor:
~~Adds to the `PayloadTypeRegistry` API to allow modifying and querying the max packet size for a payload type.~~

Adds a `registerLarge` overload to `PayloadTypeRegistry` that takes an `IntSupplier`, allowing mods to defer setting the max packet size.

Partially supersedes #4996

### DIFF
--- a/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
+++ b/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
@@ -16,6 +16,8 @@
 
 package net.fabricmc.fabric.api.networking.v1;
 
+import java.util.function.IntSupplier;
+
 import org.jetbrains.annotations.ApiStatus;
 
 import net.minecraft.network.FriendlyByteBuf;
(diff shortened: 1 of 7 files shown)


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
```diff
@@ -60,6 +60,26 @@ public interface PayloadTypeRegistry<B extends FriendlyByteBuf> {
...
+	/**
+	 * Sets the maximum size of an <strong>already registered</strong> payload type.
+	 *
+	 * <p>Larger payloads will be split into multiple packets, allowing to send packets larger than the vanilla limited size.
+	 *
+	 * @param type		    the payload type
+	 * @param maxPacketSize the maximum size of payload packet
+	 * @param <T>           the payload type
+	 */
+	<T extends CustomPacketPayload> void setMaxPacketSize(CustomPacketPayload.Type<T> type, int maxPacketSize);
```
modmuss50: I'm just thinking of alternatives to this API, it does feel a little weird to have a setter that has to be called after registering with a possibly unknown inital max size.

I was wondering if taking a ToIntFunction<T> (T being the payload) in a `registerLarge` overload would be a better API. I then realised that will only work for encoding 🤔.

My only other ideas are taking an `IntFunction` or maven even a `AtomicInteger`, neither of which I love.

(Im just thinking out loud so let me know what you think).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/PayloadTypeRegistryImpl.java
```diff
@@ -80,40 +80,68 @@ public <T extends CustomPacketPayload> CustomPacketPayload.TypeAndCodec<? super
...
+
+		if (maxPacketSize < 0) {
+			throw new IllegalArgumentException("Provided maxPacketSize needs to be positive!");
+		}
+
+		padAndSetMaxPacketSize(id, maxPacketSize);
+	}
+
+	@Override
+	public <T extends CustomPacketPayload> int getMaxPacketSize(CustomPacketPayload.Type<T> type) {
```
modmuss50: This should throw if the packet type is not registered instead of returning a valid value.

DennisOchulor: Yup, forgot to actually add the check...

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/PayloadTypeRegistry.java
DennisOchulor: Ya it does feel a bit weird, but personally I still prefer the setter. The other approaches would mean the max size could change arbitrarily, so the value would need to be revalidated each time. The max size could also be decreased from a previously set value which the current API doesn't allow (though I suppose we could just allow it? not sure if that would cause any problems).

modmuss50: Again thinking out loud:

It does seem wrong to allow changing this during gameplay/after packets have been sent, could it take a function to return the max size once per connection/server instance?

E.g In the constructor of FabricPacketMerger/Splitter it could compute the max packet sizes for all registered types?

DennisOchulor: Perhaps a `registerLarge` overload that takes an `IntSupplier` that is called once right before the first packet of a payload type is sent/received? Should be sufficient for the data attachment API assuming no large attachments are registered after mod init.

DennisOchulor: I've made the above change. I also removed `getMaxPacketSize()` as I feel the new design doesn't really need it.

modmuss50 (approved):
I think this is good now 👍

Patbox (approved):

PR #5122, opened 2026-01-05
Fix crash when removing none existent mob effect from entity.

### LINKED ISSUE #5121: [0.141.0] Calling LivingEntity#removeEffect when the entity does not have the effect crashes the game
This is no longer consistent with before/vanilla (which returns false if the effect is not on the entity). This is presumably caused by the new ServerMobEffectEvents API.

My code in specific crashing in an instance [here](https://github.com/MoriyaShiine/hearty-meals/blob/3ffa75f60d1a9d3e8c56b0af6f9978ca3edd6ac7/src/main/java/moriyashiine/heartymeals/common/component/entity/FoodHealingComponent.java#L162)

The crash itself: [crash-2026-01-05_12.25.02-server.txt](https://github.com/user-attachments/files/24438768/crash-2026-01-05_12.25.02-server.txt)

modmuss50:
Fixed by not invoking the events.

### DIFF
--- a/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
+++ b/fabric-entity-events-v1/src/main/java/net/fabricmc/fabric/mixin/entity/event/effect/LivingEntityMixin.java
@@ -137,7 +137,12 @@ public abstract class LivingEntityMixin extends Entity {
 			return original.call(holder);
 		}
 
-		MobEffectInstance effectInstance = (this.self()).getEffect(holder);
+		MobEffectInstance effectInstance = this.self().getEffect(holder);
+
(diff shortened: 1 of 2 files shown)


sylv256 (approved):
LGTM, this should prevent the issue. I don't believe the other events would be relevant since they either deal with a `MobEffectInstance` being directly passed or passed through a `Collection`, none of which Mojang have marked `@Nullable`.

modmuss50:
> LGTM, this should prevent the issue. I don't believe the other events would be relevant since they either deal with a `MobEffectInstance` being directly passed or passed through a `Collection`, all of which Mojang have not marked `@Nullable`.

Yep, that matches my understanding. Thanks for taking a look.

PR #5129, opened 2026-01-09
Create EventResult enum, an simpler alternative to InteractionResult

Patbox:
This pr creates an EventResult enum, which main usecase is to provide simple {ALLOW, DENY, PASS} options for events to use (both Fabric API and for other mods). Additionally migrate EntitySleepEvents from InteractionResult to EventResult as an example. Some events from interactions module might get the refactor in later pr (that module needs slightly more work).

Main reasoning with this change is that a lot of events don't need, use or implement full functionality of InteractionResults, which provides {SUCCESS, SUCCESS_SERVER, CONSUME, FAIL, PASS, TRY_WITH_EMPTY_HAND} values and additionally supports setting the resulting ItemStack. All of this functionality might be confusing, if event only really needs SUCCESS, PASS and FAIL (in which case the names aren't the best either).

Alternative to this would be to just use existing TriState enum, however I feel that it's names could be confusing in some contexts.

### DIFF
--- a/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/EventResult.java
+++ b/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/EventResult.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2016, 2017, 2018, 2019 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.fabricmc.fabric.api.util;
+
+import com.mojang.serialization.Codec;
+
+import net.minecraft.util.StringRepresentable;
+
+/**
+ * Represents a result of an event that controls execution of some action.
+ */
+public enum EventResult implements StringRepresentable {
+	/**
+	 * Prevents further event handling, while allowing related action.
+	 */
+	ALLOW("allow"),
+	/**
+	 * Continues execution of further events.
+	 * In case of being returned by a final callback, it might either act the same as ALLOW
+	 * or execute additional logic to determine the outcome.
+	 */
+	PASS("pass"),
+	/**
+	 * Prevents further event handling, while also preventing related action.
+	 */
+	DENY("deny");
+
+	public static final Codec<EventResult> CODEC = StringRepresentable.fromEnum(EventResult::values);
+
+	private final String name;
+
+	/**
+	 * Checks whatever action should be allowed.
+	 * @return true if it's allowed, otherwise false.
+	 */
+	public boolean allowAction() {
+		return this != DENY;
+	}
+
+	/**
+	 * Checks whatever action should be allowed, with custom return value for {@link EventResult#PASS}.
+	 * @return true if it's allowed, otherwise false.
+	 */
+	public boolean allowAction(boolean passResult) {
+		return switch (this) {
+		case PASS -> passResult;
+		case ALLOW -> true;
+		case DENY -> false;
+		};
+	}
+
+	/**
+	 * Value of this enum as string.
+	 *
+	 * @return lowercase name of the value.
+	 */
+	@Override
+	public String getSerializedName() {
+		return this.name;
+	}
+
+	EventResult(String name) {
+		this.name = name;
+	}
+}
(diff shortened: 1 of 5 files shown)


Fuzss:
That's a really great change, thanks.

Personally I'd love if this could be expanded to cover events currently using `boolean` return values.
Right now it's often not clear and not consistent with such events whether `true` or `false` preserves vanilla behavior, prevents certain logic from running etc. Imo it would be great if this could be streamlined, like by using `EventResult.PASS` and an additional `EventResult.INTERRUPT` (or just `EventResult.DENY`).

This is the approach I've been using for a long time in my own library, here is my [`EventResult`](https://github.com/Fuzss/puzzleslib/blob/main/1.21.11/Common/src/main/java/fuzs/puzzleslib/api/event/v1/core/EventResult.java) class.

Technici4n:
Having more values than strictly necessary makes it all confusing. In fact this was precisely the problem back when InteractionResult was still an enum. What the boolean does should be obvious from the event name (e.g. `allowElytraFlight`), or are there cases where that can't be done?

Patbox:
I think Fuzzs means events that return a nullable Boolean, but I'm not sure if fabric api has those

Fuzss:
@Patbox 
I meant primitive `boolean` return values.

@Technici4n 
I don't see a parallel with any previous issues like we were facing with `InteractionResult`. After all `EvenResult` is a custom class specifically for event return values, whether there are three or four values in it all of them have a distinct usage. Nothing confusing about it.

Also Fabric callbacks are mostly used as lambdas so you usually don't even get to see the method name.
I'm just talking from personal experience here. It's confusing at times what specific return value does what in certain callbacks. Standardizing return values would remedy that.

Fuzss:
I've also previously talked about standardizing return values for Fabric events (`EventResult` for `boolean` or `TriState` return values, `EventResultHolder` for other types of return values): https://github.com/FabricMC/fabric-api/issues/4677#issue-3117741772
I still very much believe Fabric would greatly benefit from streamlining all that, as it makes it much clearer how to use events and what return values will yield what behavior (interrupt vs pass).

This is an idea I originally got from Architectury Api, where it's also applied for all of their events.

Fuzss:
Here are a few examples for events where imo it's not intuitive how the `boolean` return value should be used, and digging into the Javadoc is required (which standardized `EvenResult` return values would solve):
https://github.com/FabricMC/fabric-api/blob/a208210e45897d43f2b0b62fdf3e501e64e75593/fabric-events-interaction-v0/src/main/java/net/fabricmc/fabric/api/event/player/PlayerBlockBreakEvents.java#L99
https://github.com/FabricMC/fabric-api/blob/a208210e45897d43f2b0b62fdf3e501e64e75593/fabric-events-interaction-v0/src/client/java/net/fabricmc/fabric/api/event/client/player/ClientPreAttackCallback.java#L60
https://github.com/FabricMC/fabric-api/blob/a208210e45897d43f2b0b62fdf3e501e64e75593/fabric-screen-api-v1/src/client/java/net/fabricmc/fabric/api/client/screen/v1/ScreenMouseEvents.java#L197
<...>

Also not a fan of how events that "pass through" vanilla values are handled, with some of them requiring `null` to be returned to keep the vanilla value untouched, while others require the vanilla value to be returned which is passed as one of the parameters to the callback. A standardized [EventResultHolder](https://github.com/Fuzss/puzzleslib/blob/main/1.21.11/Common/src/main/java/fuzs/puzzleslib/api/event/v1/core/EventResultHolder.java) would solve that. Examples:
https://github.com/FabricMC/fabric-api/blob/a208210e45897d43f2b0b62fdf3e501e64e75593/fabric-loot-api-v3/src/main/java/net/fabricmc/fabric/api/loot/v3/LootTableEvents.java#L128
https://github.com/FabricMC/fabric-api/blob/a208210e45897d43f2b0b62fdf3e501e64e75593/fabric-model-loading-api-v1/src/client/java/net/fabricmc/fabric/api/client/model/loading/v1/ModelModifier.java#L85
<...>

Fuzss:
And to anyone wondering how standardized event return values in the form of `EventResult` and `EventResultHolder` could look once implemented; I invite you to have a peek at the common events in my personal library, which mostly mirror Fabric events and already use `EventResult` and `EventResultHolder` everywhere:
https://github.com/Fuzss/puzzleslib/tree/main/1.21.11/Common/src/main/java/fuzs/puzzleslib/api/event/v1

Technici4n:
`allowMouseClick` should be quite clear from the name, but that convention was created relatively late, so older events like in the interaction events module don't follow it.

There is some merit in what you suggest but I remain skeptical of a one-size-fits-all approach. Even NeoForge moved away from `@HasResult` and related functions, but admittedly there it's not super clear if cancelling is enough or if methods of the event need to be called. Usually you have to read the source code anyway ;)

modmuss50:
I agree I dont think there a reason for having a generic `EventResult` enum to be used by many events. Would it not be better for each event to have an enum that is taylored for the specific event? Even if its structuraly the same as another one the docs/names could be written for the specific use case.

Patbox:
A lot of events would end up with enums that would be 1:1 with this one if that was implemented, since the minimal {ALLOW, DENY, PASS} would work in most cases (like how InteractionResult was used before, even if slighly incorrectly).
I do think that in case of events that need more values / more specific behaviour it would make sense to just provide extra enum class rather than forcing EventResult to work with them. Also main goal for me is to replace incorrectly used InteractionResult, not to change how other boolean events work (through if it did improve them, someone could also pr that).

Fuzss:
I agree with Patbox, there is no need for per event result enums that are all functionally equivalent.
Not been a fan of when NeoForge switched their generic result enum to per event enums, it just needlessly complicates things when the functionality remains the same.

Also didn't want to stall this PR with my suggestions / ideas above. Right now this is good to go and can be expanded upon later. 👍

Fuzss (approved):

modmuss50 (approved):

PR #5139, opened 2026-01-16
Fix FRAPI package structure to include "client"

PepperCode1:
- Move `SpriteFinder` from model to sprite package
- Remove deprecated methods in `SpriteFinder`


modmuss50:
Im not against repackaging this, I just want to check we are happy with doing so as I dont think the benefits are that great (unless im missing something?). We decided against repackaging some of the other modules that are still using the legacy package names.

I'm happy either way, I just want to be sure that this is worth the extra disruption ontop of all the other changes we have in 26.1.

PepperCode1:
I missed the discussion about repackaging and what reasons were given for or against, but at least for FRAPI, I think it's important because it's the only rendering module with the legacy package structure and I do not see a future where FRAPI is superseded as any precedent substantial enough to do so would also allow making breaking changes to the module directly; thus, if these renames are not done now, the legacy names will remain permanently.

modmuss50 (approved):

PR #5143, opened 2026-01-17
Add large attachment syncing

DennisOchulor:
This PR mainly does two things:

1. Change `ClientboundAttachmentSyncPayload` to handle only one attachment. Initial syncs are sent in a bundle packet.
2. Allow mods to register synced attachments with a configurable max sync size that can go above the vanilla S2C limit.

Supersedes #4996


fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java
```diff
@@ -130,13 +154,24 @@ public AttachmentType<A> buildAndRegister(Identifier id) {
 				);
 			}
 
+			if (maxSyncSize <= AttachmentSync.DEFAULT_MAX_DATA_SIZE) {
+				maxSyncSize = AttachmentSync.DEFAULT_MAX_DATA_SIZE;
+			} else if (maxSyncPacketSize == -1) {
+				throw new IllegalStateException("Large attachment " + id + " registered too late! Must be registered during mod initialization.");
+			} else {
+				int newMaxPacketSize = maxSyncSize + AttachmentSync.MAX_PADDING_SIZE_IN_BYTES;
```
modmuss50, line 162: Do you check for this overflowing anywhere? E.g what if max int is passed?

modmuss50 (approved):
Looks good to me, just noticed one little thing.

modmuss50:
Also a merge conflict, should be an easy fix.

fabric-data-attachment-api-v1/src/main/java/net/fabricmc/fabric/impl/attachment/AttachmentRegistryImpl.java
DennisOchulor: The line right below it checks for overflow

modmuss50: AH yeah!!! Not sure how I missed that lol 😆

DennisOchulor:
Quoting @jacobsjo's comment [here](https://github.com/FabricMC/fabric-api/issues/4669#issuecomment-3510853522)

> It might solve some issues, but not the one I've reported in this issue. This issue is about the very harsh handling of oversized packets, which can happen no matter what the limit is set to.

Instead of crashing the server, the only other thing that could be done is to just log a warning since at certain points where `AttachmentChange` is created, it is not known which clients it should be synced too yet. 

Though personally I think crashing the server is fine, especially since the default limit is now 1 MiB. If modders are unknowingly syncing more than that, they should probably be forced to make a proper decision about going above it.

PR #5148, opened 2026-01-20
Fix vanilla bug preventing adding optional elements to tag in datagen

ThePotatoArchivist:
In vanilla, there is a datagen bug where `TagAppender$2#addOptional` calls `TagAppender#add` instead of `TagAppender#addOptional`. This means that any `TagAppender` produced by `TagAppender#map`, such as `FabricIntrinsicHolderTagsProvider#valueLookupBuilder`, cannot add optional elements as they are handled like required elements. This PR fixes that bug.

This is my first PR so please let me know if the inject should be moved to a different mixin or if it needs tests or other style issues.

### DIFF
--- a/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/TagAppenderMixin.java
+++ b/fabric-data-generation-api-v1/src/main/java/net/fabricmc/fabric/mixin/datagen/TagAppenderMixin.java
@@ -16,9 +16,12 @@
 
 package net.fabricmc.fabric.mixin.datagen;
 
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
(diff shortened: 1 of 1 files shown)


ekulxam (approved):

modmuss50 (approved):

modmuss50:
Looks great, thanks. Does this apply just to 26.1 or do you need it backporting?

ekulxam:
Likely https://discord.com/channels/507304429255393322/566276937035546624/1463109526596354078

ThePotatoArchivist:
I don't really need this fix for anything specific, I just noticed it while testing my library. ~~It's a bug in 1.21.11 as well, at least, I'd bet it's been around for a while.~~ I didn't read Sky's message

ThePotatoArchivist:
There's probably someone who might find a backport useful though

modmuss50:
I tend to only backport stuff if people actually want it, I'll just merge into 26.1 for now.

PR #5156, opened 2026-01-23
No longer disconnect on unknown targets.

### LINKED ISSUE #4943: Crash due to attachment sync. Received attachment for unknown target
When using the new synced attachment API I often get this. 

22:50:04] [Render thread/ERROR]: Error accepting attachment changes
net.fabricmc.fabric.impl.attachment.sync.AttachmentSyncException: Received attachment change for unknown target!

Does it need to hard fail like that? If the target clearly exists on the server given the attachment is accessed there

modmuss50:
Its impossible for the server to know for sure what entities are on the client.

Retained a system prop to bring back the old behaviour if we ever need an easier way to debug these issues in the future. I hate this but its the only option.

Fixes #4943


DennisOchulor:
While we're at it, obtaining the chunk attachment target on the client should probably be changed to the method that doesn't return an empty chunk if it's not found, since just attaching to the empty chunk is a mild memory leak. 
image

PR #5157, opened 2026-01-24
Separately flush UI elements with connected primitives

ekulxam:
Neoforge PR: neoforged/Neoforge#2922

https://discord.com/channels/507304429255393322/566276937035546624/1460367770083655691

Let me know if I should add a test for this.


modmuss50:
Does this change anything in vanilla? I really need someone familar with rendering to review this.

cputnam-a11y (approved):
This doesn't change vanilla behavior, as vanilla doesn't use connected primitives in GUI. As stated in discord if we wanted we could gate this change against our GUI fix flagg for vertex formats, further protecting mods against the default behavior, but I don't think this is necessary. Taking advantage of this will require mods to set the flag anyway though, otherwise the vertex format will be quads, so it might make sense to bind it to that? Idk. The actual changes lgtm.

maityyy:
I think such changes should be noted in README.md of the module, otherwise how else will modders know about it?

PR #5160, opened 2026-01-24
Mixin cleanup part 2

### LINKED ISSUE #5168: `handleInteract` Injection in `ServerGamePacketListenerImplMixin` Not Marked as `cancellable`
Method injection cancels depending on return value of the `UseEntityCallback` event yet the `Inject` annotation is not marked as `cancellable`, causing an error.

modmuss50:
Part 2 of cherry picking @Earthcomputer 's PR: https://github.com/FabricMC/fabric-api/pull/5118

Part 3 I will look at the comments on the orginal PR regarding removing unneeded locals, lets keep it focused and merge this one quickly to reduce conflicts.


LlamaLad7:
Personally I'm not entirely convinced, see my comments from the original PR:
> - Using names for argsOnly locals seems to me less stable than using ordinals / implicits, since it seems more likely for the internal code to change than the external api
> - Implicit mode seems potentially more stable than names when possible for normal locals, though I feel less strongly about this

If people have reason to disagree then we can discuss but I don't think "use name everywhere" is a sensible blanket strategy. 
I suggested avoiding the locals where possible because that's *definitely* better, whereas a lot of these changes are dubiously better.

modmuss50:
Ah sorry, I must have missed that comment on the other PR.

Good point on argsOnly, I think I will change that.

I dont think stability across versions is the main goal here if something is renamed/chages its usually done for a reason, and more than likely may cause the mixin to behave differently.

PR #5164, opened 2026-01-25
Implement Packet Context API.

Patbox:
Packet Context is a simple way to store and access values between networking stages (login, configuration, play) and between handler/config sync/whatever and packet serialization (for example changing the vanilla packet format depending on client having XYZ). 
PacketContext can be sourced from Connection and *PacketListernerImpl classes for usage outside of serialization and with PacketContext.get() (or orElseThrow()) inside.

Values can be then simply retriever with a getValue(Key<T>) method or stored with setValue(Key<T>, T) (or updateValues for changing multiple at the same time).
There are also provided method to run something with PacketContext.get() set up outside of vanilla packet serialization for handling the edge cases where it is needed.

The idea for this api is based from <https://github.com/NucleoidMC/packet-tweaker>, which is used by few mods/libraries for this feature. In case of Fabric API itself it will be useful for handling of vanilla compatibility in case of extended (stream) codec formats.

This pr is fully functional. Only things left to consider is what/if fabric should provide some contexts by default.


fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,148 @@
...
+		ScopedValue.where(PacketContextImpl.VALUE, provider.getPacketContext()).run(runnable);
+	}
+
+	/**
+	 * Runs specified runnable under a packet context, returning a value.
+	 *
+	 * @param provider provider of the context
+	 * @param supplier supplier to execute
+	 * @return result of supplier
+	 */
```
modmuss50, line 160: What are the use cases for these?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,148 @@
...
+@ApiStatus.NonExtendable
+public interface PacketContext {
+	/**
+	 * Returns currently stored value.
+	 *
+	 * @param key unique key under which value is stored
+	 * @return stored value or null if not set.
+	 */
+	@Nullable
+	<T> T getValue(Key<T> key);
```
modmuss50: `get`? (diito for the other methods with "value" in the name).

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/context/PacketContextImpl.java
```diff
@@ -0,0 +1,71 @@
...
+		return (T) this.contextMap.get(key);
+	}
+
+	@Override
+	public <T> void setValue(Key<T> key, T value) {
+		// Values can be set/read from multiple threads, so making it safe is kinda needed.
+		// Also setting values should be way less common than reading them, so keeping that fast,
+		// While putting synchronization + new map for when new values are set.
+		// When more values at the same time, the updateValues method should be used.
+		synchronized (this) {
```
modmuss50: This feels like a premature optimistation to me, wouldn't `Collections.synchronizedmap` make this much simpler.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/context/PacketContextImpl.java
```diff
@@ -0,0 +1,71 @@
...
+			var map = new IdentityHashMap<>(contextMap);
+			map.put(key, value);
+			this.contextMap = map;
+		}
+	}
+
+	@Override
+	public void updateValues(Consumer<ContextUpdater> updater) {
```
modmuss50: As above, I think this is making the API more complex than it needs to be. Happy to discuss if you think otherwise.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
Patbox: Niche usecases mostly, where you need to get something serialized with or without applying context. Polymer for example uses it to serialize unmodified data, as it changes serialization with context present. Also vanilla uses streamcodecs outside of netty, so it might be useful for wrapping that in context

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
Patbox: Yeah originally was just get but wasn't sure due to overlap with static. Can change through

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/context/PacketContextImpl.java
Patbox: My concern was that this might be called a lot in case of being used in a nested codec/lists, which then would also scale with player count.

modmuss50: A ReadWriteLock would make more sense here then.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,147 @@
...
+	 * @param value value to store
+	 */
+	<T> void set(Key<T> key, T value);
+
+	/**
+	 * Returns a connection that owns this packet context.
+	 *
+	 * @return the connection owning this context
+	 */
+	Connection getOwner();
```
modmuss50: I wonder if it would be cleaner to have a `Key<Connection> CONNECTION` in this file for this?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * Returns currently set packet context.
+	 *
+	 * @return current context or null
+	 */
+	@Nullable
+	static PacketContext get() {
+		return PacketContextImpl.VALUE.get();
```
modmuss50: ScopedValue.get() throws if its not bound, in what case will this return null?

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,147 @@
...
+	 * Creates a new key to be used with the packet context.
+	 *
+	 * @param key identifier for this key
+	 * @return a unique key
+	 */
+	static <T> Key<T> key(Identifier key) {
+		return new Key<>(key);
+	}
+
+	final class Key<T> {
```
modmuss50: Nit: make this an interface and have an impl class.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,147 @@
...
+import org.jspecify.annotations.Nullable;
+
+import net.minecraft.network.Connection;
+import net.minecraft.resources.Identifier;
+
+import net.fabricmc.fabric.impl.networking.context.PacketContextImpl;
+
+/**
+ * This class allow to easily pass context between multiple packet listeners and packet serialization.
+ * All connections get their own unique context object.
```
modmuss50, line 38: Nit: A simple code example or something to show how to use it may be useful.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
```diff
@@ -0,0 +1,147 @@
...
+		return Objects.requireNonNullElse(get(key), defaultValue);
+	}
+
+	/**
+	 * Stores the value.
+	 *
+	 * @param key unique key under which value is stored
+	 * @param value value to store
+	 */
+	<T> void set(Key<T> key, T value);
```
modmuss50: Does not document that T can be nullable, to remove.

fabric-networking-api-v1/src/testmod/java/net/fabricmc/fabric/test/networking/context/PacketContextTest.java
```diff
@@ -0,0 +1,100 @@
...
+import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.networking.v1.PayloadTypeRegistry;
+import net.fabricmc.fabric.api.networking.v1.ServerConfigurationConnectionEvents;
+import net.fabricmc.fabric.api.networking.v1.ServerLoginConnectionEvents;
+import net.fabricmc.fabric.api.networking.v1.ServerPlayConnectionEvents;
+import net.fabricmc.fabric.api.networking.v1.context.PacketContext;
+import net.fabricmc.fabric.test.networking.NetworkingTestmods;
+
+public class PacketContextTest implements ModInitializer {
+	public static final Logger LOGGER = LoggerFactory.getLogger(PacketContextTest.class);
```
modmuss50: Mega nit pick: loggers should be private.

modmuss50 (commented):
Looking good, just a few nit picks.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
Patbox: Oh oops, will probably fix that then

modmuss50:
You could likely quite easily write a unit test to cover the basic get+set logic of this. Ill leave that up to you.

fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/context/PacketContext.java
Patbox: Main issue is using a key would make it mutable (which isn't something we would want).
I was thinking about introducing a ReadKey (which implementation wise would still be just a regular Key), so an api could just expose it that way

modmuss50: Ah yeah, that a valid point. Leave it for now then. Can always deprecate it later if you do add a read only key. For now its fine.

Patbox: I kinda like the idea for read only keys (and do have them locally, just didn't push as I still worked on default contexts). I can drop the default ones and just include that through

modmuss50 (approved):

PR #5172, opened 2026-01-29
Make lookup API context extend nullable Object 

deirn:
`@NullMarked` makes the lookup context `C` non-nullable by default. 
This is not ideal since `C` comes from the usage of the API and thus should allow for null if needed.
For example, the transfer API has a nullable `Direction` context for sided storages.

<details><summary>Before</summary>
<p>

Screenshot_20260130_011606
Screenshot_20260130_011636
Screenshot_20260130_011701

</p>
</details> 

<details><summary>After</summary>
<p>

Screenshot_20260130_013915
No warnings here.
Note that I needed to invalidate the cache and reload the project for the warning to go away, classic IntelliJank. 

</p>
</details>

### DIFF
--- a/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
+++ b/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/block/BlockApiCache.java
@@ -41,7 +41,7 @@ import net.fabricmc.fabric.impl.lookup.block.BlockApiLookupImpl;
  * @see BlockApiLookup
  */
 @ApiStatus.NonExtendable
-public interface BlockApiCache<A, C> {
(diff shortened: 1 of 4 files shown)


Toffikk:
why not just mark the context as nullable? it would be much cleaner if possible

deirn:
Then it will always nullable. You want `C` to control it.

cputnam-a11y:
> why not just mark the context as nullable? it would be much cleaner if possible

Yeah, I would do this, marking the context as potentially Nullable like `C extends @Nullable Object` as jspecify recommends

deirn:
That a thing? TIL. I'll take a look later then.
I also looked on `@UnknownNullability` from JetBrains annotation, but you can't use it on the type parameter.

deirn:
Extending nullable Object seems to be working.

modmuss50 (approved):

PR #5176, opened 2026-01-30
Add missing dependency to fabric-gametest-api-v1

EpicPix:
Adds missing `fabric-registry-sync-v0` dependency to `fabric-gametest-api-v1` that caused a crash when trying to register tests in the registry when registry sync was not present

Also, should `fabric-client-gametest-api-v1` have a `moduleDependencies` with the resource loader api?


modmuss50 (approved):
Looks good, as far as im aware the client gametest module doesnt need this. The regular one is used as we register custom game test functions into the registry.

PR #5177, opened 2026-01-30
Add transitive access wideners for new `Enchantment` functional interfaces

ThePotatoArchivist:
At some point during the snapshot cycle `Enchantment#applyEffects` was modified to use the private functional interfaces `GenericAction` and `FloatAction`, which made it impossible to use the access-widened `applyEffects` without adding your own access wideners. This PR makes those functional interfaces public so `applyEffects` can be used properly.


Patbox:
You need to modify template.classtweaker and then run generateClassTweaker task with gradle

ThePotatoArchivist:
I think that's correct? I put it with the EnchantmentHelper interface wideners

Patbox (approved):

modmuss50 (approved):

